PR #23461, opened 2023-03-22
feat(bigquery): `get_catalog_names`

betodealmeida:
### SUMMARY

Implement `get_catalog_names` for BigQuery. This is needed to support catalog-level permissions, and also for https://github.com/apache/superset/issues/22862.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N/A

### TESTING INSTRUCTIONS

I configured a BigQuery database and ran this script:

```python
from superset.app import create_app

app = create_app()
with app.app_context():
    from superset import db
    from superset.db_engine_specs.bigquery import BigQueryEngineSpec
    from superset.models.core import Database

    database = db.session.query(Database).first()
    with database.get_inspector_with_context() as inspector:
        print(BigQueryEngineSpec.get_catalog_names(database, inspector))
```

When ran, it produced the expected output:

```python
['covid19-293320']
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

Issue #13351, opened 2021-02-25
[SIP-59] Proposal for Database migration standards


robdiciuccio:
### Motivation

Reduce pain around metadata database migrations by ensuring standards are followed and appropriate reviews are obtained before merging.

### Proposed Change

[SIP-57](https://github.com/apache/superset/issues/12566) (Semantic Versioning) introduced standards for avoiding breaking changes and general best practices for database migrations. The proposed changes below are in addition to those standards:

- All migrations must support rollbacks. Migrations must have a functional `downgrade` method to effectively rollback schema changes introduced in the `upgrade` method. If a migration makes changes to data that are not easily undone (e.g. #12960), the changes introduced must be non-breaking and idempotent.
- Migrations should be atomic and configured to complete fully in a single run, using a single transaction where appropriate. Any failures should trigger a rollback to the previous state. Partial migrations should be avoided.
- Any [constraints](https://docs.sqlalchemy.org/en/13/core/constraints.html) added within a migration should include an explicit [name](https://alembic.sqlalchemy.org/en/latest/naming.html), e.g. `sa.ForeignKeyConstraint(["user_id"], ["ab_user.id"], name='fk_user_id')`.
- PRs introducing database migrations must include runtime estimates and downtime expectations.
- Care should be taken to not introduce expensive DDL operations such as adding unnecessary constraints/indexes or setting column default values on tables potentially containing thousands of rows. [[1](https://www.craigkerstiens.com/2017/09/10/better-postgres-migrations/)][[2](https://severalnines.com/database-blog/schema-management-tips-mysql-mariadb)]
  - Indexes in Postgres tables should be added and removed [CONCURRENTLY](https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#indexes-with-concurrently).
- Migrations for breaking changes and cleanup (e.g. removal of columns) that should be held for the next major version, per the guidelines in [SIP-57](https://github.com/apache/superset/issues/12566), should be accumulated in `/superset/migrations/next/` for evaluation and inclusion in a future release.
- Establish Github [code owners](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) on the `superset/migrations` directory to ensure PMC members are notified of new or updated migrations.
- Require **two** approvals for PRs that include database migrations, including committers from multiple organizations.
- PRs including database migrations should be kept open for a minimum review period of **two business days** to allow for adequate review, unless circumstances such as a critical vulnerability or breakage require faster turnaround.

[...]

PR #25845, opened 2023-11-02
fix(bigquery): fix exception when bigquery import fails

drummerwolli:
this code was introduced with https://github.com/apache/superset/pull/23461. but in the tests, its not checked what happens if the google packages are not installed. i tried running Superset 3.0.1 without the bigquery packages installed. it failed to start, because it actually throws another exception.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/bigquery.py
+++ b/superset/db_engine_specs/bigquery.py
@@ -52,14 +52,14 @@ try:
     from google.oauth2 import service_account
 
     dependencies_installed = True
-except ModuleNotFoundError:
+except ImportError:
     dependencies_installed = False
 
 try:
     import pandas_gbq
 
     can_upload = True
-except ModuleNotFoundError:
+except ImportError:
     can_upload = False
 
 if TYPE_CHECKING:


drummerwolli:
are there tests without bigquery packages installed, which could validate this?

drummerwolli:
maybe @betodealmeida can have a look as original author of the code introduction in https://github.com/apache/superset/pull/23461

superset/db_engine_specs/bigquery.py
```diff
@@ -52,14 +52,14 @@
     from google.oauth2 import service_account
 
     dependencies_installed = True
-except ModuleNotFoundError:
```
john-bodley, line 55: @drummerwolli out of interest why is `ModuleNotFoundError` not suffice? According to [this](https://stackoverflow.com/questions/62854761/python3-8-whats-the-difference-between-importerror-and-modulenotfounderror#:~:text=The%20ImportError%20is%20raised%20when,Python%203.6%20and%20newer%20versions.) StackOverflow post it states,

> TL;DR You should catch ModuleNotFoundError to detect when a package is not installed.

drummerwolli:
the error message i got was misleading, and actually cause by a dependency which we install on top of superset. this dependency had some google package included, and seems to cause problems. i think its not a problem with superset itself. this can be closed, sry for the noise

PR #25847, opened 2023-11-02
109 - refactor(logging): Clean up `statsd` logging

sebastianliebscher:
### SUMMARY

This PR removes the `log_to_statsd=` functionality from the `@event_logger` and replaces usage with already existing `@statsd_metrics` decorator. I basically moved `BaseSupersetApiMixin` from `base_api.py` --> `base.py` so that the flask views can use @statsd_metrics too.

Currently, there are two decorators for basically the same use case:
- `@event_logger.log_this_with_context(log_to_statsd=True)`
- `@statsd_metrics`

The `@event_logger` just increments a statsd counter, while `@statsd_metrics` additionally measures the execution time of a function.

### TESTING INSTRUCTIONS
- run unit + integration tests
- manually checked for new statsd metrics

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/async_events/api.py
+++ b/superset/async_events/api.py
@@ -23,6 +23,7 @@ from flask_appbuilder.security.decorators import permission_name, protect
 
 from superset.async_events.async_query_manager import AsyncQueryTokenException
 from superset.extensions import async_query_manager, event_logger
+from superset.views.base import statsd_metrics
 from superset.views.base_api import BaseSupersetApi
 
 logger = logging.getLogger(__name__)
@@ -33,6 +34,7 @@ class AsyncEventsRestApi(BaseSupersetApi):
     allow_browser_login = True
 
     @expose("/", methods=("GET",))
+    @statsd_metrics
     @event_logger.log_this
     @protect()
     @safe
--- a/superset/advanced_data_type/api.py
+++ b/superset/advanced_data_type/api.py
@@ -57,7 +57,6 @@ class AdvancedDataTypeRestApi(BaseSupersetApi):
     @permission_name("read")
     @event_logger.log_this_with_context(
         action=lambda self, *args, **kwargs: f"{self.__class__.__name__}.get",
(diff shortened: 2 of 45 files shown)


superset/async_events/api.py
```diff
@@ -33,6 +34,7 @@ class AsyncEventsRestApi(BaseSupersetApi):
     allow_browser_login = True
 
     @expose("/", methods=("GET",))
+    @statsd_metrics
     @event_logger.log_this
     @protect()
```
sebastianliebscher, line 39: Most API endpoint already use `@statsd_metrics`. In a case where this decorator is added to replace implicit `log_to_statsd=True`, the statsd metric 

- `superset_<Endpoint>` 

will be replaced by

- `superset_<Class>_<Endpoint>_success|warning|error`
- `superset_<Class>_<Endpoint>_time`
- `superset_<Class>_<Endpoint>_time_count`
- `superset_<Class>_<Endpoint>_time_sum`

john-bodley:
Thanks @sebastianliebscher for the refactor/cleanup. Given (as you mentioned) this is a breaking change, i.e., the logged statsD metric name will change it seems like we would have to wait until the breaking window opens for 4.0 (circa Q1 2024) before merging.

In the interim I was wondering whether there is a workaround? One could double log the metric (which does pollute the metric space) and add deprecation logic in 3.1 (slated for late Q4 2023) and then remove the double logging in 4.0.

cc @michael-s-molina

michael-s-molina:
> One could double log the metric (which does pollute the metric space) and add deprecation logic in 3.1 (slated for late Q4 2023) and then remove the double logging in 4.0.

@john-bodley This seems like too much for something that is not urgent. I think it would be better to just wait for the breaking window.

rusackas:
This PR hasn't seen any activity/events in roughly seven months. I'll reluctantly close it, but would love to see it reopened/revisited if anyone here has interest or bandwidth.

PR #25855, opened 2023-11-03
feat(sqllab): Dynamic query limit dropdown

giftig:
### SUMMARY

Previously, this was a constant of powers of ten from 10 to 100000 and then we tacked on maxRow on the end, being the SQL_MAX_ROW config. This unfortunately results is very confusing behaviour if SQL_MAX_ROW is less than 100,000 as it still let you select 100,000, though wouldn't acctually use that LIMIT, and the messaging would not acknowledge the setting for SQL_MAX_ROW either.

Instead, construct the list by ascending through powers of ten until we reach the configured SQL_MAX_ROW.

#### Before
BEFORE-static

#### After
AFTER-dynamic

### TESTING INSTRUCTIONS

1. Set `SQL_MAX_ROW` to a value less than 100,000 or greater than 1,000,000
2. Use the LIMIT dropdown in SQL Lab
3. Observe that the drop down no longer displays values higher than `SQL_MAX_ROW`, and will display every power of 10 up to `SQL_MAX_ROW`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/QueryLimitSelect/QueryLimitSelect.test.tsx
+++ b/superset-frontend/src/SqlLab/components/QueryLimitSelect/QueryLimitSelect.test.tsx
@@ -25,7 +25,6 @@ import { render, fireEvent, waitFor } from 'spec/helpers/testing-library';
 import userEvent from '@testing-library/user-event';
 import { initialState, defaultQueryEditor } from 'src/SqlLab/fixtures';
 import QueryLimitSelect, {
-  LIMIT_DROPDOWN,
   QueryLimitSelectProps,
   convertToNumWithSpaces,
 } from 'src/SqlLab/components/QueryLimitSelect';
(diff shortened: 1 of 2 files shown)


giftig:
Tagging @justinpark since I see you were the author of this file.

superset-frontend/src/SqlLab/components/QueryLimitSelect/QueryLimitSelect.test.tsx
```diff
@@ -106,13 +105,25 @@ describe('QueryLimitSelect', () => {
...
+      mockStore(initialState),
+    );
     const dropdown = baseElement.getElementsByClassName(
       'ant-dropdown-trigger',
     )[0];
 
     userEvent.click(dropdown);
     await waitFor(() => expect(getByRole('menu')).toBeInTheDocument());
+
+    const expectedLabels = [10, 100, 1000, 10000, 50000].map(i =>
```
john-bodley, line 119: @giftig would you also mind adding a test when `maxRow` is a factor of 10?

giftig: Added.

john-bodley (approved):

PR #24158, opened 2023-05-21
fix(table): percentage metric should use verbose map

Always-prog:
### SUMMARY
Percentage metrics don't using verboseMap, this PR fixing it.

#### Before
Screenshot_52

#### After 
Screenshot_53

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25857, opened 2023-11-03
fix(table): Double percenting ad-hoc percentage metrics

john-bodley:
### SUMMARY

This PR fixes an issue which was a byproduct of another fix—https://github.com/apache/superset/pull/24158. 

Specifically the previous logic to use the verbose name (if the key sans `%` was present) would fallback to the key (containing the `%`) if not present (which is the case for ad-hoc metrics) and then (re)apply the `%`, i.e., rather than `%SUM(num)` the label would be `%%SUM(num)`. The fix is simply to only include the `%` prefix in the label if the key (sans `%`) is present in the map.

#### BEFORE 

Screenshot 2023-11-03 at 3 17 53 PM

#### AFTER 

Screenshot 2023-11-03 at 3 14 02 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


justinpark (approved):
LGTM

PR #24272, opened 2023-06-02
chore(deps): bump typescript to 4.8.4

jansule:
### SUMMARY

This bumps `typescript` from `4.5.4` to `4.8.4`. `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` were upgraded to `5.42.1` to match the typescript version.

New TSERRORs were fixed where possible. Unfortunately, I had to add a few `@ts-ignore`s here and there. Maybe someone with a deeper understanding of the code base is able to fix these errors properly.

I decided to bump to typescript 4.8.4 as work related to https://github.com/apache/superset/discussions/21758 introduced an upgrade to this version anyways. In consultation with @villebro we decided to create a dedicated PR for this.

### TESTING INSTRUCTIONS

Simply run the tests and build the application to see if any TSERRORs occur.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25869, opened 2023-11-06
feat(plugin): add plugin-chart-cartodiagram

jansule:
### SUMMARY

This adds a new plugin `plugin-chart-cartodiagram` that allows creating cartodiagrams, i.e. charts on a map. See also https://github.com/apache/superset/discussions/21758.

## How does it work?

In order to support as many charts as possible, we created some kind of a meta plugin that takes configurations of any existing chart plugin supported by superset and places it on a map. To do so, we make all charts with a common data source selectable, and inject a group-by property based on a provided geometry column. Then, we use the `buildQuery()` and `transformProps()` functions of each chart in combination with the respective chart component (via the `ChartComponentRegistry`) to render them on a map. Through this, all configuration options of any chart are supported.

Additional configuration options for cartodiagrams are:

- **Map extent**: Either fit the view to the extent of the chart data (every item should be visible), or defined a custom extent by moving the map to the desired location.
- **Background layers**: Add an arbitrary number of background layers to the map. Currently the formats WMS (raster), XYZ (raster) and WFS (vector) in multiple versions are supported. Vector layers can be styled in the explore view.
- **Background color of chart**: The background color and opacity of the chart can be configured in order to improve the readability of a chart depending on the configured background layers.
- **Corner radius of chart**: Set the corner radius of the wrapping `<div>` element of a chart.
- **Zoom based chart size**: Set the chart size for each zoom level in order to avoid cluttering. Predefined functions to compute all values are provided (`fixed`, `linear`, `exponential`).

Default layers can be configured in `MainPreset.js` which will be shown by default for every configured cartodiagram panel. Users are able to edit/remove these layers for a single panel in the explore view. Default layers currently include a WMS showing OpenStreetMap data in grayscale.

*Disclaimer*: The service providing the OpenStreetMap data is a demo service hosted by terrestris GmbH & Co. KG, which is my employer. On a higher zoomlevel, it shows a watermark for our paid service. The demo server is not recommended to be used in production and we recommend replacing this service. It was just added for others to have a default background map while getting familiar with the plugin and probably should be replaced in the future.

## Technical Details/Requirements

Additional libraries:
- **[OpenLayers](https://openlayers.org/)** as mapping library
- **[GeoStyler](https://geostyler.org/)** for styling of vector data (disclaimer: I am one of the maintainers of GeoStyler)

Both libraries are part of the [Open Source Geospatial Foundation](https://www.osgeo.org/), so the chances of future development of those libraries are quite certain.

---

In order to make use of above libraries, we had to update some dependencies (e.g. typescript, see also https://github.com/apache/superset/pull/24272), and we also had to make use of `npm dedupe` in order to resolve version conflicts. Unfortunately, this resulted in a nearly complete rewrite of the `package-lock.json`.

Due to some version upgrades, we also had to apply some linting fixes, which results in more changed files than actually directly related to this PR. Sorry for that.

---

The `<SelectControl>` had to be changed a little to be able to add a customized list of options. We use it to display all charts with a common dataset and also check, if a previously selected chart was updated after the selection.

## Limitations

There are some limitations of the plugin, which probably should be addressed in the future:

- We currently only support geometries in the coordinate reference system (CRS) ~`EPSG:3857` (WGS 84 / Pseudo-Mercator)~ _EPSG:4326 (WGS 84)_. Additional CRSs might be added in the future.
- For now, we expect the geometry column to return a GeoJSON point geometry for placing the charts on the map. Additional geometries (e.g. lines and polygons), as well as additional formats (WellKnownText, lat/lon, etc.) might be added in the future.
- Rendering performance can be improved and varies strongly depending on the used chart type.
- ~Currently, filters applied to a dashboard are not reflected in the cartodiagram plugin. We would be grateful, if someone can hint us to the right direction for supporting this.~ Dashboard filters are now supported.

---

Since the background layers point to external services, the CSP have to be adjusted. Additional information can be found in the README of the plugin.

## How can this be improved?

This PR is just the first step into the direction for additional geospatial support. So there are many things that can be added and improved. Besides the points mentioned above, following points might also be useful:

- Adding a FeatureFlag to chart plugins that should be allowed to place on a map is probably a nice feature that was also already discussed with @villebro. To do so, the filter options of the REST API need to be extended.
- Improved caching of the charts in the client, in order to reduce the number of rerenderings of a chart.
- Adding the geometries of the actual dataset as a background layer and allow styling it.

**Example dashboard with some cartodiagrams:**

superset-plugin-demo-dashboard

**Explore view of cartodiagram plugin (configuration):**

superset-plugin-demo-configuration

**Explore view of cartodiagram plugin (customization):**

superset-plugin-demo-customize

**Explore view of cartodiagram plugin (layer configuration):**

superset-plugin-demo-layer-configuration

**Explore view of cartodiagram plugin (layer styling):**

superset-plugin-demo-layer-styling

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/discussions/21758
- [ ] Required feature flags:
- [ ] Changes UI


jansule:
I'm not sure why the linting in the `Frontend/frontend-build` pipeline fails here. On my local setup, I do not get any errors (using node v16.20.2).

justinpark:
@jansule Thanks for contributing a new chart plugin. This PR is quite large and includes TypeScript changes. Could you please detach all changes not related to the newly introduced chart plugin?

superset-frontend/package.json
```diff
@@ -138,12 +139,18 @@
...
     "fast-glob": "^3.2.7",
     "fontsource-fira-code": "^4.0.0",
     "fs-extra": "^10.0.0",
     "fuse.js": "^6.4.6",
     "geolib": "^2.0.24",
+    "geostyler": "^12.0.2",
+    "geostyler-data": "^1.0.0",
+    "geostyler-openlayers-parser": "^4.2.1",
+    "geostyler-style": "^7.4.0",
+    "geostyler-wfs-parser": "^2.0.0",
```
justinpark, line 153: Please add these libraries to [your plugin directory](https://github.com/apache/superset/pull/25869/files#diff-2a826b513161e937d6993e06b6525786055cf86cc4e43408b7170a5c58a3beef), not to the root directory.

superset-frontend/package.json
```diff
@@ -138,12 +139,18 @@
     "d3-scale": "^2.1.2",
     "dom-to-image-more": "^2.10.1",
     "dom-to-pdf": "^0.3.2",
+    "echarts": "^5.4.1",
```
justinpark, line 142: i.e. `echarts` is included in the plugin-chart-echarts package, and all components that depend on ECharts should also be included in that package.
```suggestion
```

superset-frontend/babel.config.js
```diff
@@ -22,6 +22,12 @@ module.exports = {
   sourceMaps: true,
   sourceType: 'module',
   retainLines: true,
+  // After adding transformIgnorePatterns to jest.config.js
+  // the tests failed. Following line fixed the problem. See also
+  // https://github.com/babel/babel/issues/8731#issuecomment-423845498
+  // core-js seems to be packaged as commonjs modules, so it should
+  // be (more or less) save to ignore them with babel.
+  ignore: [/^core-js$/],
```
justinpark, line 30: These are likely side effects of changes to package.json. Please clean up package.json as suggested, and these changes should no longer be necessary.

```suggestion
```

jansule:
> @jansule Thanks for contributing a new chart plugin. This PR is quite large and includes TypeScript changes. Could you please detach all changes not related to the newly introduced chart plugin?

Thanks for the first review @justinpark. I created a PR with only the TypeScript changes a while ago. See https://github.com/apache/superset/pull/24272. I guess it just got lost in the amount of PRs you are receiving.

superset-frontend/package.json
jansule: I added these to the root directory as they are needed by the `<LayerConfigsControl />` which was added to `superset-frontend/src/explore` and not to the plugin itself. 

The only reason the control is not included in the plugin directory is, that the `<ControlFormItem />` was moved from `@superset-ui/core` to `superset-frontend/src/explore`, so I cannot import it in the plugin directory. This component, however, is needed for a consistent UI of the controls.

jansule:
@villebro I finally managed to update the PR so that the pipelines succeed. The failing Dependency Review pipeline fails due to a missing license entry in one of our transitive dependencies (BSD-3 Clause). I added a PR there a while ago that fixes the problem.

So could you take another look at this PR?

lcalisto:
Any update on this PR? This would be a very cool feature for Superset.

pesekon2:
I would also like to have this PR merged (mapbox is a joke). All my support to @jansule for the patience with updating and conflict solving, etc.

I've tried to test it locally, although not being eligible as a possible reviewer. An item in the geometry column looks the following way: `{'type': 'Point', 'coordinates': [-4.908446676, 48.373637064]}`. However, I get the following error:

```
Unexpected error

Error: could not identify an equality operator for type json
LINE 31:   coords3,
           ^
```

Is there an issue in my data or could it be in the code? 

image

pesekon2:
> I would also like to have this PR merged (mapbox is a joke). All my support to @jansule for the patience with updating and conflict solving, etc.
> 
> I've tried to test it locally, although not being eligible as a possible reviewer. It seems to work. I am not able to load any WMS as the background but this could possibly be due to some of the proxy/reverseproxy black magic here.

Not a proxy magic. Made it work locally with the default terreatris WMS but not being able to change the WMS to anything else. See the example in the screenshot (no WMS visible, no error message in logs; the service URL is https://eagri.cz/public/app/wms/public_zp.fcgi):

image

rusackas:
Looks like `css-font-parser` is making into `package-lock.json` and has an incompatible license. Any way to get rid of that?

jansule:
Thanks for the feedback! I will try to get some time to resolve the merge conflicts soon. 

> Looks like css-font-parser is making into package-lock.json and has an incompatible license. Any way to get rid of that?

@rusackas There is an invalid spdx identifier in `css-font-parser`, but the license itself is compatible (BSD-3-Clause). I created a PR a while ago to solve the issue https://github.com/bramstein/css-font-parser/pull/11 but did not get any feedback.

> I am not able to load any WMS as the background but this could possibly be due to some of the proxy/reverseproxy black magic here.

@lcalisto So far I had no problems with loading other WMS. You might want to ensure that your Talisman settings are correct. I will double check as soon as I get time to work on this PR again.

pesekon2:
> > I am not able to load any WMS as the background but this could possibly be due to some of the proxy/reverseproxy black magic here.
> 
> @lcalisto So far I had no problems with loading other WMS. You might want to ensure that your Talisman settings are correct. I will double check as soon as I get time to work on this PR again.

Solved it today (I guess it was on me and not on @lcalisto ). It was indeed a problem with Talisman. I didn't make it work with the Talisman so simply turning it off helped but this could very probably be connected to some security issues at the server or some obscure thing about the WMS in use. 

Therefore: Big thumbs up. Tested and working for me. Thanks very much and I hope it will get merged one day.

rusackas:
> @rusackas There is an invalid spdx identifier in css-font-parser, but the license itself is compatible (BSD-3-Clause). I created a PR a while ago to solve the issue https://github.com/bramstein/css-font-parser/pull/11 but did not get any feedback.

Ahh, ok, if it's all good, you can get the Dependency Review CI check to pass by adding the package to `allow-dependencies-licenses` in the `.github/workflows/dependency-review.yml` file

lcalisto:
I just saw that @jansule PR https://github.com/bramstein/css-font-parser/pull/11 has been merged. Perhaps that also helps. 🙂

jansule:
> I just saw that @jansule PR [bramstein/css-font-parser#11](https://github.com/bramstein/css-font-parser/pull/11) has been merged. Perhaps that also helps. 🙂

That's great. I justed pinged the owner to create a new release. Hopefully this will happen soon. Otherwise I will add the dependeny to the `allow-dependencies-licenses` as @rusackas mentioned.

> Solved it today (I guess it was on me and not on @lcalisto ). It was indeed a problem with Talisman. I didn't make it work with the Talisman so simply turning it off helped but this could very probably be connected to some security issues at the server or some obscure thing about the WMS in use.

Happy to hear that it's working for you now!

jansule:
@rusackas @villebro I just rebased the PR. Could you take another look?

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/transformProps.ts
```diff
@@ -0,0 +1,92 @@
...
+   *     the column/alias to its value. Example:
+   *     `[{ col1: 'abc', metric1: 10 }, { col1: 'xyz', metric1: 20 }]`
+   *   - `rowcount`: the number of rows in `data`
+   *   - `query`: the query that was issued.
+   *
+   * Please note: the transformProps function gets cached when the
+   * application loads. When making changes to the `transformProps`
+   * function during development with hot reloading, changes won't
+   * be seen until restarting the development server.
+   */
```
villebro, line 54: Let's remove the boilerplate comments
```suggestion
```

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/buildQuery.ts
```diff
@@ -0,0 +1,68 @@
...
+ * format (e.g. CSV or JSON) of the result and whether or not to force refresh the data from
+ * the datasource as opposed to using a cached copy of the data, if available.
+ *
+ * More importantly though, QueryContext contains a property `queries`, which is an array of
+ * QueryObjects specifying individual data requests to be made. A QueryObject specifies which
+ * columns, metrics and filters, among others, to use during the query. Usually it will be enough
+ * to specify just one query based on the baseQueryObject, but for some more advanced use cases
+ * it is possible to define post processing operations in the QueryObject, or multiple queries
+ * if a viz needs multiple different result sets.
+ */
```
villebro, line 34:
```suggestion
```

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/controlPanel.ts
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * The control panel is split into two tabs: "Data" and
+   * "Customize". The controls that define the inputs to
+   * the chart data request, such as columns and metrics, usually
+   * reside within "Data", while controls that affect the visual
+   * appearance or functionality of the chart are under the
+   * "Customize" section.
+   */
+
+  // For control input types, see: superset-frontend/src/explore/components/controls/index.js
```
villebro, line 35:
```suggestion
```

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * The constructor is used to pass relevant metadata and callbacks that get
+   * registered in respective registries that are used throughout the library
+   * and application. A more thorough description of each property is given in
+   * the respective imported file.
+   *
+   * It is worth noting that `buildQuery` and is optional, and only needed for
+   * advanced visualizations that require either post processing operations
+   * (pivoting, rolling aggregations, sorting etc) or submitting multiple queries.
+   */
```
villebro, line 39:
```suggestion
```

superset-frontend/plugins/plugin-chart-echarts/package.json
```diff
@@ -25,13 +25,13 @@
...
   "dependencies": {
     "d3-array": "^1.2.0",
-    "echarts": "^5.4.1",
     "lodash": "^4.17.21",
     "moment": "^2.30.1"
   },
   "peerDependencies": {
     "@superset-ui/chart-controls": "*",
     "@superset-ui/core": "*",
+    "echarts": "*",
```
villebro, line 34: I vaguely remember there having been a discussion about why this was needed as a core dependency. Is this really necessary, or can we leave `echarts` as a dependency only on the ECharts plugin?

superset-frontend/plugins/plugin-chart-table/src/utils/extent.ts
```diff
@@ -16,7 +16,6 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
```
villebro, line 19: Nit: I assume this is a mistake - let's avoid touching unrelated files.

villebro (changes requested):
Thanks for keeping this PR alive @jansule ❤️ If you can rebase the PR and work with us for a few rounds to iron out the last kinds I think we should be able to get this merged fairly before the new year. FYI: @michael-s-molina @rusackas this is a good example of a plugin that would benefit from the improved FE plugin architecture we've been discussing.

superset-frontend/plugins/plugin-chart-table/src/utils/extent.ts
jansule: Good catch. Removed changes

superset-frontend/plugins/plugin-chart-echarts/package.json
jansule: We use it in one of the new controls and thus also need it as a core dependency.

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/index.ts
jansule: done

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/controlPanel.ts
jansule: done

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/buildQuery.ts
jansule: done

superset-frontend/plugins/plugin-chart-cartodiagram/src/plugin/transformProps.ts
jansule: done

superset-frontend/plugins/plugin-chart-cartodiagram/src/CartodiagramPlugin.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export default function CartodiagramPlugin(props: CartodiagramPluginProps) {
+  const { height, width } = props;
+  const theme = useTheme();
+
+  const rootElem = createRef<HTMLDivElement>();
+
+  const [mapId] = useState(
+    `cartodiagram-plugin-${Math.floor(Math.random() * 1000)}`,
+  );
```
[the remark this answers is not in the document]
jansule: good point. changed.

superset-frontend/packages/superset-ui-core/src/chart/types/QueryResponse.ts
```diff
@@ -31,6 +31,14 @@ export interface TimeseriesDataRecord extends DataRecord {
...
 }
 
+export const isTimeseriesDataRecord = (
+  item: any,
+): item is TimeseriesDataRecord => Object.keys(item).includes('__timestamp');
+
+export const isTimeseriesDataRecordList = (
+  items: any[],
+): items is TimeseriesDataRecord[] => items.every(isTimeseriesDataRecord);
+
```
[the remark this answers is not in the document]
jansule: Good catch. Removed it.

jansule:
Thanks for the review @villebro! I applied your suggestions. It would be great if you could have another look at it

villebro:
> The frontend-build fails due to `@types/geojson` not being listed as a dependency of `plugins/legacy-preset-chart-deckgl`. Is this something that should be addressed within this PR?

@jansule hmm, this is curious - let's satisfy the linter and see if it all makes sense when CI is green.

jansule:
> @jansule hmm, this is curious - let's satisfy the linter and see if it all makes sense when CI is green.

Done. Looks like the dependency was actually missing. I'm wondering why this check did not fail before.

rusackas:
I'm excited we can likely get this into 5.0!

villebro (approved):
LGTM - let's 🚢 it!

kgabryje:
Hi @jansule!
I noticed that this PR has added a huge chunk to the bundle (2.14MB parsed, visible in the top left corner in the first screenshot). For some reason it contains all antd icons (700kB) and pages/Chart module (500kB), which makes this chunk the largest one in the bundle.
For comparison, second screenshot shows the bundle analysis after reverting this PR.
It probably won't affect the loading times of the dashboards which don't contain the Cartodiagram chart, but we might want to solve this issue for the sake of loading performance of dashboards that use this plugin (especially for users with slower Internet)

image

image

jansule:
Thanks for the analysis @kgabryje. I will take a close look at that

kgabryje:
Awesome, thank you!

Issue #22960, opened 2023-02-02
Missing Configuration Options to Improve Security for Embedded Dashboards


kmzs:
**Please note:** We came across these problems in a recent penetration test of a customer's application using the Superset embedded SDK. We think these problems are not vulnerabilities in the Superset framework itself. Therefore, we did not choose the process of responsible disclosure to communicate these issues to the Apache Superset team.

However, we do think this is an important issue that should be addressed because it causes security vulnerabilities in common scenarios such as the example below.

# Example Scenario
The embedded SDK is used to embed a dashboard into a web portal. This portal allows customers to book different services. The dashboard shows statistics regarding the customer's booked services. For each customer the dashboard shows only data for their own services. A customer must not be able to access statistics/data for other customers.

We think similar scenarios are a common use case for the Superset embedded SDK.

## Possible Attack
In this example a customer could act maliciously and try to access the data of another customer (e.g., a competitor). The attacker is registered in the customer portal as a non-privileged user. They are allowed to access the embedded dashboard, but only to display data for their own services.

# Problems
The embedded SDK does not provide sufficient security for scenarios such as the example.
In particular the following problems arise:

## 1. API Requests Allow to Manipulate the Database Queries
The POST request to the `/api/v1/chart/data` API allows to manipulate the database query in various ways. The body of the request contains a complex JSON object. The values of different fields in this object are directly injected into the database query. This complex object provides a large attack surface for an attacker.

A few examples:

- An attacker can manipulate the query to access other user's data. One example could be adding new conditions (e.g., using the `where` field). Another example is if the content of the guest token can be manipulated. Parts of the guest token's body are directly injected into the database query.
- An attacker can add arbitrary function provided by the database to the query. These functions are executed when the database query is executed. The impact relies on the functions provided by the specific database. However, the functions could be useful for a great variety of attacks, such as Denial of Service (DoS), Server-side Request Forgery (SSRF), and unauthorized access to other user's data, to the file system, or to internal information.
- An attacker could add multiple queries to the request by adding multiple JSON object to the `queries` array. This could lead to or increase the effect of DoS attacks against the database, for example.

To our knowledge, there is no detailed documentation of the complex JSON object in the body of the request. This makes it difficult to try to defend against all possible injections.

In general, user input should never be injected directly into database queries to protect against [SQL injection](https://owasp.org/www-community/attacks/SQL_Injection) attacks.

## 2. API Responses Reveal Internal Information
[...]

PR #25878, opened 2023-11-06
fix(embedded): Hide sensitive payload data from guest users

jfrag1:
### SUMMARY

Currently, guest users can get access to more data than they need to when viewing an embedded dashboard.  The `api/v1/dashboard/{id}` and `api/v1/dashboard/{id}/datasets` endpoints would return owners' first and last names. The latter would return information about the underlying database(s) the datasets belong to, which could be useful for an attacker.  None of this data is necessary for the embedded dashboard to render.

I recognize this PR does break typical API patterns by stripping out certain data depending on the type of user requesting, which is pretty ugly.  However, this approach avoids making breaking changes for regular API use, while making the embedded experience more secure.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/22960
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -33,7 +33,7 @@ from marshmallow import ValidationError
 from werkzeug.wrappers import Response as WerkzeugResponse
 from werkzeug.wsgi import FileWrapper
 
-from superset import is_feature_enabled, security_manager, thumbnail_cache
+from superset import is_feature_enabled, thumbnail_cache
 from superset.charts.schemas import ChartEntityResponseSchema
 from superset.commands.importers.exceptions import NoValidFilesFoundError
 from superset.commands.importers.v1.utils import get_contents_from_bundle
@@ -66,9 +66,7 @@ from superset.dashboards.filters import (
 from superset.dashboards.schemas import (
     DashboardCopySchema,
     DashboardDatasetSchema,
-    DashboardDatasetSchemaGuest,
     DashboardGetResponseSchema,
-    DashboardGetResponseSchemaGuest,
     DashboardPostSchema,
     DashboardPutSchema,
     EmbeddedDashboardConfigSchema,
@@ -240,9 +238,7 @@ class DashboardRestApi(BaseSupersetModelRestApi):
     chart_entity_response_schema = ChartEntityResponseSchema()
     dashboard_get_response_schema = DashboardGetResponseSchema()
     # pylint: disable=invalid-name
-    dashboard_get_response_schema_guest = DashboardGetResponseSchemaGuest()
     dashboard_dataset_schema = DashboardDatasetSchema()
-    dashboard_dataset_schema_guest = DashboardDatasetSchemaGuest()
     embedded_response_schema = EmbeddedDashboardResponseSchema()
     embedded_config_schema = EmbeddedDashboardConfigSchema()
 
@@ -346,12 +342,7 @@ class DashboardRestApi(BaseSupersetModelRestApi):
             404:
               $ref: '#/components/responses/404'
         """
-        schema = (
-            self.dashboard_get_response_schema_guest
-            if security_manager.is_guest_user()
-            else self.dashboard_get_response_schema
-        )
-        result = schema.dump(dash)
+        result = self.dashboard_get_response_schema.dump(dash)
 
         add_extra_log_payload(
             dashboard_id=dash.id, action=f"{self.__class__.__name__}.get"
@@ -413,12 +404,9 @@ class DashboardRestApi(BaseSupersetModelRestApi):
         """
         try:
             datasets = DashboardDAO.get_datasets_for_dashboard(id_or_slug)
-            schema = (
-                self.dashboard_dataset_schema_guest
-                if security_manager.is_guest_user()
-                else self.dashboard_dataset_schema
-            )
-            result = [schema.dump(dataset) for dataset in datasets]
+            result = [
+                self.dashboard_dataset_schema.dump(dataset) for dataset in datasets
+            ]
             return self.response(200, result=result)
         except (TypeError, ValueError) as err:
             return self.response_400(
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -239,6 +239,7 @@ class DashboardRestApi(BaseSupersetModelRestApi):
     edit_model_schema = DashboardPutSchema()
     chart_entity_response_schema = ChartEntityResponseSchema()
     dashboard_get_response_schema = DashboardGetResponseSchema()
+    # pylint: disable=invalid-name
     dashboard_get_response_schema_guest = DashboardGetResponseSchemaGuest()
     dashboard_dataset_schema = DashboardDatasetSchema()
     dashboard_dataset_schema_guest = DashboardDatasetSchemaGuest()
@@ -413,13 +414,11 @@ class DashboardRestApi(BaseSupersetModelRestApi):
         try:
             datasets = DashboardDAO.get_datasets_for_dashboard(id_or_slug)
             schema = (
-              self.dashboard_dataset_schema_guest
-              if security_manager.is_guest_user()
-              else self.dashboard_dataset_schema
-          )
-            result = [
-                schema.dump(dataset) for dataset in datasets
-            ]
+                self.dashboard_dataset_schema_guest
+                if security_manager.is_guest_user()
+                else self.dashboard_dataset_schema
+            )
+            result = [schema.dump(dataset) for dataset in datasets]
             return self.response(200, result=result)
         except (TypeError, ValueError) as err:
             return self.response_400(
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -237,7 +237,6 @@ class DashboardRestApi(BaseSupersetModelRestApi):
     edit_model_schema = DashboardPutSchema()
     chart_entity_response_schema = ChartEntityResponseSchema()
     dashboard_get_response_schema = DashboardGetResponseSchema()
-    # pylint: disable=invalid-name
     dashboard_dataset_schema = DashboardDatasetSchema()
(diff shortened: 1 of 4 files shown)


superset/dashboards/api.py
```diff
@@ -342,6 +342,11 @@ def get(
               $ref: '#/components/responses/404'
         """
         result = self.dashboard_get_response_schema.dump(dash)
+        if security_manager.is_guest_user():
```
john-bodley, line 345: @jfrag1 we should adhere to the "shift left" mentality and filter out inaccessible fields at the DAO layer as opposed to the API. This ensures we adhere to the DRY principle avoiding the need to repeat (and/or miss) said logic.

jfrag1: @john-bodley Thanks for reviewing, but I'm not sure how we're supposed to properly filter out certain fields of a record in the DAO layer.  As far as I'm aware the security logic in the DAO layer is only "row-level", protecting entire objects, there's no column-level protections.  Are you suggesting in the appropriate DAO method do something like
```python
dashboard = db.session.query(...
if is_guest:
    dashboard.owners = []
    dashboard.changed_by = None
```
I don't like this since it introduces the possibility of these changes accidentally being committed, but please let me know if you had something else in mind or if this pattern does exist currently but I'm just not aware of it.

john-bodley: @betodealmeida @dpgaspar @michael-s-molina @villebro what are your thoughts on where sanitization of the data response should occur? Note this doesn't seem to be API specific and thus I'm not sure we should explicitly be defining said logic there.

michael-s-molina: Implementing this at the API layer is problematic because there's no single point of verification and also because you will query data that you don't have access to and filter it out later, wasting resources (imagine a table with many columns and you only have access to one column). This will also be true for the suggested code below:

```
dashboard = db.session.query(...
if is_guest:
    dashboard.owners = []
    dashboard.changed_by = None
```

Considering our current architecture, where we offer a [RLS (Row Level Security)](https://docs.preset.io/docs/row-level-security-rls) solution instead of depending on the database layer for these types of controls, the correct way of implementing this is to augment our security model to also support [Column Level Security (CLS)](https://blog.fabric.microsoft.com/en-us/blog/announcing-column-level-row-level-security-for-fabric-warehouse-sql-endpoint/) or even [Object Level Security (OLS)](https://learn.microsoft.com/en-us/power-bi/enterprise/service-admin-ols).

> I don't like this since it introduces the possibility of these changes accidentally being committed

Ideally, you won't query for what you don't have access to, eliminating this concern because the values won't be in the session.

Pinging @villebro for awareness given his current work on an improved security model.

jfrag1: I do appreciate the discussion here, but this really feels like premature optimization to me.  The changes in this PR only affect two endpoints, and I don't predict that this kind of logic will needed in a widespread manner.

If I'm wrong and there is a larger need/desire for this kind of security measure, I definitely agree we should re-assess, but I personally don't see a need to come up with a more "perfect" solution at this point.

jfrag1: > Ideally, you won't query for what you don't have access to, eliminating this concern because the values won't be in the session.

I'm not sure this is the right direction, imagine a situation where the backend needs to read a column for some reason, even though we don't want to surface that field to the logged-in user.  It seems wrong to tie the server's hands behind its back and not allow it to access the data just because of the current logged-in user.  I'd argue the API layer is where that kind of security check belongs, since it's sort of the gateway from the backend to the user.

eschutho: Here's a slight shift left solution, that's not quite to the DAO layer. Currently the DAO fetches the entire row, and then we're using Marshmallow to decide what is sent back to the client. Can we let Marshmallow decide what should/shouldn't be returned in the response schema since that is the mechanism that currently handles this responsibility?

john-bodley: Yes. Ideally Marshmallow should be used to define both the request and response payload.

Additionally if this is going to reside in the API layer (at least in the interim) could we at least factor the code to adhere to the DRY principle?

michael-s-molina: > I do appreciate the discussion here, but this really feels like premature optimization to me. The changes in this PR only affect two endpoints, and I don't predict that this kind of logic will needed in a widespread manner.

I thought that by referencing the concept of [Column Level Security (CLS)](https://blog.fabric.microsoft.com/en-us/blog/announcing-column-level-row-level-security-for-fabric-warehouse-sql-endpoint/) it would be more clear why this is not a premature optimization and why we will have the same need in a widespread manner.

> I'd argue the API layer is where that kind of security check belongs

If you research a little bit about CLS you'll see that many of these security checks are implemented at the database layer, even further from the API layer.

villebro: I agree it's risky to implement this fix in a single interface, as it gives no guarantees that this information won't be exposed via some other interface, either pre-existing, or one that's added later. Mind you, this is not the only security concern regarding guest tokens - the issue of exposing granular information about the underlying query has been raised multiple times in the community, and is also something I feel we need to tackle soon.

Could we setup a meeting to discuss viable long-term solutions to these issues? I propose meeting to agree on the the following:
- A long term vision for fixing this and other similar security concerns. I agree with @michael-s-molina that baking this into the security work I've started sounds like a good idea. I can share what I have so far and my thoughts of how this can be taken forward.
- Agreeing on a quick fix to unblock heavy users of guest tokens to bridge the gap between now and when we have a proper security mechanism in place for handling this and other similar security issues.

jfrag1: > I thought that by referencing the concept of [Object Level Security (OLS)](https://learn.microsoft.com/en-us/power-bi/enterprise/service-admin-ols) it would be more clear why this is not a premature optimization and why we will have the same need in a widespread manner.

This is referring to a security feature similar to row-level security where Superset could apply security filters to queries against connected databases, while the security issue at hand in this PR has to do with data from the Superset metadata DB.

> If you research a little bit about CLS you'll see that many of these security checks are implemented at the database layer, even further from the API layer.

My understanding is that CLS is for a security feature commonly implemented in data warehouses/BI tools, not a guide for how a server should implement security checks for data it owns

superset/dashboards/api.py
```diff
@@ -341,7 +346,13 @@ def get(
             404:
               $ref: '#/components/responses/404'
         """
-        result = self.dashboard_get_response_schema.dump(dash)
+        schema = (
+            self.dashboard_get_response_schema_guest
+            if security_manager.is_guest_user()
+            else self.dashboard_get_response_schema
```
eschutho, line 352: @jfrag1 I think we can even move the schema logic into Marshmallow dynamically with either a type function or a   @marshmallow.post_dump decorator. That way the logic isn't in the api layer and you won't need two schemas.

eschutho (commented):
One more suggestion.

superset/dashboards/api.py
eschutho: I think it's a great idea to meet on long term solutions. If we were to move the logic for this endpoint into the schema (I left one more suggestion that would remove this from the api altogether) should we move forward with this for now as @villebro suggested?

villebro: > I think it's a great idea to meet on long term solutions. If we were to move the logic for this endpoint into the schema (I left one more suggestion that would remove this from the api altogether) should we move forward with this for now as @villebro suggested?

@eschutho I also think moving this into the Marshmallow schema is probably the best solution for now.

superset/dashboards/schemas.py
```diff
@@ -198,6 +199,15 @@ class DashboardGetResponseSchema(Schema):
     changed_on_humanized = fields.String(data_key="changed_on_delta_humanized")
     is_managed_externally = fields.Boolean(allow_none=True, dump_default=False)
 
+    # pylint: disable=unused-argument
+    @post_dump()
+    def post_dump(self, serialized: dict[str, Any], **kwargs: Any) -> dict[str, Any]:
+        if security_manager.is_guest_user():
+            serialized["owners"] = []
```
villebro, line 206: I would prefer to fully remove the property `owners`, as replacing it with an empty list is IMO conveying incorrect info

superset/dashboards/schemas.py
```diff
@@ -247,6 +257,14 @@ class DashboardDatasetSchema(Schema):
     normalize_columns = fields.Bool()
     always_filter_main_dttm = fields.Bool()
 
+    # pylint: disable=unused-argument
+    @post_dump()
+    def post_dump(self, serialized: dict[str, Any], **kwargs: Any) -> dict[str, Any]:
+        if security_manager.is_guest_user():
+            serialized["owners"] = []
```
villebro, line 264: Same here

villebro (changes requested):
@jfrag1 I would prefer it if we remove the owners property all together if guest tokens aren't supposed to know who the owners are. Or is there some other reason we want to keep the property there? Also, I feel we really should add tests for both cases here (hence "Request changes").

superset/dashboards/schemas.py
john-bodley: @villebro I agree. The only caveat would be if the owners field is required.

villebro: @john-bodley ideally I think we should relax that requirement if we want to make this data conditional on the permissions of the logged in user. Based on the proposed changes, I think it should be as follows:
- regular user: all relevant owner info is conveyed (current behavior)
- guest token: no owner property is present in the response

jfrag1: @villebro I initially opted to keep the owners property in the payload because the dashboard frontend expects it to always be there, and fails to render if it's not.

I found a way to exclude it from the response payload and have the client tack the property on after the dashboard is requested, which allows the embedded dashboard to render successfully.

jfrag1:
@villebro Are you able to take another look at this?

villebro (approved):
LGTM, thanks for all the iterations @jfrag1 !

PR #25657, opened 2023-10-16
fix(dremio): Fixes issue with Dremio SQL generation for Charts with Series Limit

OskarNS:
### SUMMARY
#23872 released in Superset 3.0.0 introduced a regression to Dremio SQL generation as described by my issue here: #25554. 

The regression causes Superset to not give aliases to SELECT'ed columns, which means that when Superset tries to join a table onto itself using a subquery, it tries to join using an alias that has not been defined. See the difference between the generated SQL queries here:
image

The error the author describes in #23872 is not an issue with the Dremio functionality in Superset, rather it was due to an error in Dremio that has been fixed in Dremio version 24.1.0. See the fixes introduced by 24.1.0 in the [Dremio changelog](https://docs.dremio.com/current/release-notes/version-240-release/):
image

To verify that it actually works I have executed the exact query reported to be problematic in an updated Dremio instance, which works as expected:
image

This PR reverts the changes introduced in #23872 

### TESTING INSTRUCTIONS

1. Create a new chart that has the options of Series Limit against a table in Dremio. Eg. Time-Series Chart, Area Chart etc.
2. Set required fields and Series Limit
3. Click Create Chart.
4. The chart is rendered correctly since the generated SQL is now valid and can be executed successfully by Dremio.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes: #25554 
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25882, opened 2023-11-06
feat: method for dynamic `allows_alias_in_select`

betodealmeida:
### SUMMARY

DB engine specs have an attribute called `allows_alias_in_select`, indicating if they support aliases in `SELECT` statements. For Dremio, this attribute is version dependent — see the fix in [their release notes](https://docs.dremio.com/current/release-notes/version-240-release/#2410-release-notes-june-2023).

Because of that, we've been ping-ponging on the attribute. First I disabled it in https://github.com/apache/superset/pull/23872, and later it was re-enabled in https://github.com/apache/superset/pull/25657 when the fix came out. Since we have Superset users running older versions, I added a new class method `get_allows_alias_in_select`, that uses the `version` field in the database extra configuration.

Note that this requires users running Dremio < 24.1.0 to explicitly add the `version` field when configuring their databases. I'm not sure if we could determine the version at runtime, but even if we could it would require us to run a pre-query every time, which is needlessly expensive.

Also note that the `version` field is already used in Presto for version-dependent features.

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -765,7 +765,7 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
         raise NotImplementedError()
 
     @property
-    def database(self) -> builtins.type["Database"]:
+    def database(self) -> "Database":
         raise NotImplementedError()
 
     @property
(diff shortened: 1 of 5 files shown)


superset/models/helpers.py
```diff
@@ -765,7 +765,7 @@ def db_engine_spec(self) -> builtins.type["BaseEngineSpec"]:
         raise NotImplementedError()
 
     @property
-    def database(self) -> builtins.type["Database"]:
```
betodealmeida, line 768: This was wrong, but was hidden by a bunch of `# type: ignore` comments. ☹️

john-bodley (approved):
LGTM

PR #25883, opened 2023-11-06
fix(horizontal filter bar filter labels): Increase max-width to 96px

rtexelm:
### SUMMARY

Allow the size of the label to represent more of the filter title before truncation. Size changed from 60px to 96px

BEFORE:
Previous size of label

AFTER:
New size of label

### TESTING INSTRUCTIONS

Create a filter with a title long enough to overflow 96px

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [X] Required feature flags: HORIZONTAL_FILTER_BAR
- [X] Changes UI


lilykuang (approved):

john-bodley:
Thanks @rtexelm for the PR. Isn't the readability a function of screen size and thus should the maximum width be a relative (as opposed to absolute) measure?

rtexelm:
@john-bodley Yes I think it should be relative, however I determined that would be a more significant style change that might be better suited to a separate PR after consulting with the designers while this was in response to a particular request for it to be set to 96px

eschutho:
@kasiazjc, in light of @john-bodley's comment what do you think about us just removing the max-width on this element altogether. The filters do a pretty good job of adjusting their width on their own. Here's what it would look like with no max width on the label.

https://github.com/apache/superset/assets/5186919/da49224a-6623-4355-908f-ac271c296e82

villebro:
I tend to agree with @eschutho here. In my experience dashboard designers will usually tend to want the whole label to be visible, and if there's need for additional context that's not needed in the label, that can always be placed in the description to be displayed in the tooltip. So I vote for always showing the full label.

kasiazjc:
> @kasiazjc, in light of @john-bodley's comment what do you think about us just removing the max-width on this element altogether. The filters do a pretty good job of adjusting their width on their own. Here's what it would look like with no max width on the label.
> 
>  Screen.Recording.2023-11-16.at.5.54.35.PM.mov

I think removing max makes sense in this case, and good point @villebro about using description if needed. Let's do it :)

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -49,7 +49,7 @@ const VerticalFilterControlTitle = styled.h4`
 const HorizontalFilterControlTitle = styled(VerticalFilterControlTitle)`
   font-weight: ${({ theme }) => theme.typography.weights.normal};
   color: ${({ theme }) => theme.colors.grayscale.base};
-  max-width: ${({ theme }) => theme.gridUnit * 15}px;
+  max-width: ${({ theme }) => theme.gridUnit * 24}px;
```
eschutho, line 52:
```suggestion
```

eschutho (changes requested):
Suggest to remove the max-width altogether

lilykuang (approved):

eschutho (approved):

PR #25897, opened 2023-11-07
fix: trino cursor

betodealmeida:
### SUMMARY

This PR improves the Trino `execute_with_cursor` method:

1. It uses a `threading.Event` to indicate that the query running in a separate thread has finished, removing the overhead of context switching that happens with `sleep()` calls.
2. More importantly, it removes the call to `query.id` from inside the thread. The `Query` object is loaded lazily, so when `query.id` is accessed we need to use the session to retrieve the ID. In certain deployments the call to get the session inside the thread might fail (at Preset it fails because the thread is missing additional context that the application has).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/trino.py
+++ b/superset/db_engine_specs/trino.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import contextlib
 import logging
 import threading
+import time
 from typing import Any, TYPE_CHECKING
 
 import simplejson as json
@@ -151,13 +152,7 @@ class TrinoEngineSpec(PrestoBaseEngineSpec):
         return None
 
     @classmethod
-    def handle_cursor_with_query_id(
-        cls,
-        cursor: Cursor,
-        query: Query,
-        session: Session,
-        cancel_query_id: str,
-    ) -> None:
+    def handle_cursor(cls, cursor: Cursor, query: Query, session: Session) -> None:
         """
         Handle a trino client cursor.
 
@@ -167,6 +162,7 @@ class TrinoEngineSpec(PrestoBaseEngineSpec):
         """
 
         # Adds the executed query id to the extra payload so the query can be cancelled
+        cancel_query_id = cursor.query_id
         logger.debug("Query %d: queryId %s found in cursor", query.id, cancel_query_id)
         query.set_extra_json_key(key=QUERY_CANCEL_KEY, value=cancel_query_id)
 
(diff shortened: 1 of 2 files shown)


villebro (approved):
I like the `Event` based refactor here 👍 LGTM, but let's leave it open for a day or so to be able to get additional eyes on it.

villebro:
@giftig as you wrote the initial implementation, can you also take a look?

superset/db_engine_specs/trino.py
```diff
@@ -193,34 +201,32 @@ def execute_with_cursor(
         in another thread and invoke `handle_cursor` to poll for the query ID
         to appear on the cursor in parallel.
         """
+        # Fetch the query ID before hand, since it might fail inside the thread due to
+        # how the SQLAlchemy session is handled.
+        query_id = query.id
```
giftig, line 206: Ah, this explains what sqlite was complaining about then; I never did get to the bottom of that.

superset/db_engine_specs/trino.py
```diff
@@ -193,34 +201,32 @@ def execute_with_cursor(
...
-
-        logger.debug("Query %d: Handling cursor", query.id)
-        cls.handle_cursor(cursor, query, session)
-
-        # Block until the query completes; same behaviour as the client itself
-        logger.debug("Query %d: Waiting for query to complete", query.id)
-        while not execute_result.get("complete"):
-            time.sleep(0.5)
+        logger.debug("Query %d: Handling cursor", query_id)
+        cls.handle_cursor_with_query_id(cursor, query, session, query_id)
```
giftig, line 229: You've also passed `query_id` as `query.id`, i.e. the ID of the superset `Query` object, when it should be `cursor.query_id`, the query ID within trino, which we can use to cancel the query.

superset/db_engine_specs/trino.py
```diff
@@ -193,34 +201,32 @@ def execute_with_cursor(
...
+            finally:
+                event.set()
 
-        execute_thread = threading.Thread(target=_execute, args=(execute_result,))
+        execute_thread = threading.Thread(
+            target=_execute,
+            args=(execute_result, execute_event),
+        )
         execute_thread.start()
+        execute_event.wait()
```
giftig, line 226: This waits too long; the point of doing this in a separate thread is that the `cursor.execute` call is only going to return when the query has finished executing, way too late to be calling `handle_cursor`. We poll for the presence of `cursor.query_id` because the `cursor.execute` call is going to make it available once it's received it, but will then block until the query completes.

giftig (changes requested):
This won't achieve what the fix originally aimed to do for a couple of reasons. Can you take a look at my comments, and test that you're able to cancel a running trino query from SQL Lab?

superset/db_engine_specs/trino.py
betodealmeida: Ah, I see. I was confused by the double negative logic `not A and not B`. Let me fix this.

superset/db_engine_specs/trino.py
betodealmeida: Ah, I didn't know these were different, let me fix this.

giftig:
Thanks @betodealmeida, LGTM now.

giftig (approved):

michael-s-molina:
Thanks @betodealmeida @villebro @giftig @justinpark @sadpandajoe for the investigation and fix. I closed my revert in favor of this PR.

PR #25282, opened 2023-09-13
fix(nativeFilters): Speed up native filters by removing unnecessary rerenders

Always-prog:
### SUMMARY
Currently, rendering 20+ filters takes not normal lot of time due to two issues:
1. The issue with the Select filter. After rendering, Select Filter adds the `__cache` key to the dataMask, and changing dataMask triggering a redraw of all filters again. This happens for each Select filter in sequence. This can be resolved by adding `__cache` to the default dataMask.
2. The issue with the `dataMaskSelected` in the hooks is that it is specified as an object, which always triggers a redraw. This can be resolved by setting `dataMaskSelected` to `JSON.stringify`.

#### Before
https://github.com/apache/superset/assets/66589759/4c3b4ff0-bb12-49eb-aa45-7ff1d8add39c

#### After
https://github.com/apache/superset/assets/66589759/db05070e-7dad-478f-a392-276a15848f63

### TESTING INSTRUCTIONS
1. Go to any dashboard
2. Add 30 select filters
3. Reload dashboard
4. See that filters take a lot of time for before render.
5 (optional). You can add console log to [FilterBarScrollContext](https://github.com/apache/superset/blob/master/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx#L120C1-L120C1) such as `console.log('rerendered filters bar!')` to see the number of times it logs.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #14869, opened 2021-05-27
feat(native-filters): Support default to first value in select filter

simcha90:
### SUMMARY

This PR fixes support for default first value in Select native filter. It includes next logic:
- if one of control panel items contains `requiredFirst` flag we cancel delay of native filters loading, delay charts loading until this filter will be initialized - we do assumption if `filterState.value === undefined`, it means that filter not initialized
- if it has `defaultToFirstValue` option enabled - we save on `filterState.value` undefined value and on filter loading we on the fly put first value

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

https://user-images.githubusercontent.com/56388545/119837538-5b2aa900-bf0b-11eb-9eee-83c90402477d.mov

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25901, opened 2023-11-08
fix(native filters): rendering performance improvement by reduce overrendering

justinpark:
### SUMMARY
Following up #25282, rendering performance can still be improved by reducing unnecessary rerenders (also known as over-rendering).
This commit improves performance in four areas.

#### 1. Inefficient uniqWith and object equality check in uniqueOptions

https://github.com/apache/superset/blob/d6fde3cdd5987e3bf9a37d1fcde2dc82ef695747/superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx#L221-L224

Screenshot 2023-11-01 at 1 48 34 PM

To get the list of unique values, it compares the entire data list (`arrayIncludesWith`) using object equality comparison (`isEqual`). This is slow because the data contains several attributes, and object equality comparison is slow.

As it shown in the above metric, `uniqueOptions` operation triggers four times every time a filter component re-renders, which takes more than 2 seconds per re-render for 10 filters (i.e. 50ms * 4 * 10 filters = 2s).

Using a hashmap approach (with unnecessary equality check)  reduced the delay from ~200m(=50ms * 4) to 10ms (95% reduction)
Screenshot 2023-11-01 at 1 47 15 PM

#### 2. Redundant filter re-renders due to the dynamic sizing

|Initially renders at a minimum width and then expands to its full width.|
|--|
|native-filter-width-overrendering|

Since the native filter's allocated to 100% of its width using [SuperChart](https://github.com/apache/superset/blob/e7a187680713867f22b082f3bb0a57296d2a331c/superset-frontend/packages/superset-ui-core/src/chart/components/SuperChart.tsx#L245-L258), which is wrapped by `@vx/responsive`, it triggers re-renders every time the screen size is updated.
This commit removes the resize observer (by assigning the static width) and updates the filter component wrapper to the full width using CSS instead. As a result, the filters no longer re-render when the page size changes.
native-filter-css-width

#### 3. descendants over-rendering

React's inherent behavior of rerendering all descendant components, coupled with the Redux container's rerendering upon state changes, leads to excessive rerendering of dashboard components, including filter components, whenever the chart fetch status updates the Redux store.
This commit replaces DashboardBuilder to the identical descendant (`props.children`) which prevents the rerenders when the container component rerenders.
To further optimize performance, the dashboard state synchronization logic was moved to a separate component, effectively decoupling it from the DashboardPage component and preventing unnecessary rerenders of the DashboardPage.

#### 4. Redundant `__cache` attribute

|Due to superfluous setDataMask calls triggered by __cache update, each setState update causes four unnecessary uniqWith executions|
|--|
|fix_native_filters___rendering_performance_improvement_by_reduce_overrendering_by_justinpark_·_Pull_Request__25901_·_apache_superset|

An unnecessary __cache value was introduced in the SelectFilterPlugin in [PR#14869](https://github.com/apache/superset/pull/14869/commits/1749ed542a4f4cedc6847316bf65fc2b82eb1fbd#diff-b4d2de68c8a3b04879772237085ede1b49a6908b6f2890a9ab1aaeae9e18b854R65), but it was never utilized and resulted in superfluous rerenders of filter components.
This commit removes this redundant `__cache` value to resolve the problem.

Successfully optimized the application's performance, resulting in a 10% decrease in rendering count (255 to 228) and a remarkable 60% reduction in loading time (10.9 seconds to 4.5 seconds).

|Before|After|
|--|--|
|Screenshot 2023-11-07 at 3 57 28 PM|Screenshot 2023-11-07 at 3 58 18 PM|

### TESTING INSTRUCTIONS

1. Go to any dashboard
2. Add more than 10 select filters
3. Reload dashboard
4. See that filters take a lot of time for before render.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/SyncDashboardState/index.tsx
+++ b/superset-frontend/src/dashboard/components/SyncDashboardState/index.tsx
@@ -1,0 +1,103 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import React, { useEffect } from 'react';
+import pick from 'lodash/pick';
+import { shallowEqual, useSelector } from 'react-redux';
+import { DashboardContextForExplore } from 'src/types/DashboardContextForExplore';
+import {
+  getItem,
+  LocalStorageKeys,
+  setItem,
+} from 'src/utils/localStorageHelpers';
+import { RootState } from 'src/dashboard/types';
+import { getActiveFilters } from 'src/dashboard/util/activeDashboardFilters';
+
+type Props = { dashboardPageId: string };
+
+const EMPTY_OBJECT = {};
+
+export const getDashboardContextLocalStorage = () => {
+  const dashboardsContexts = getItem(
+    LocalStorageKeys.dashboard__explore_context,
+    {},
+  );
+  // A new dashboard tab id is generated on each dashboard page opening.
+  // We mark ids as redundant when user leaves the dashboard, because they won't be reused.
+  // Then we remove redundant dashboard contexts from local storage in order not to clutter it
+  return Object.fromEntries(
+    Object.entries(dashboardsContexts).filter(
+      ([, value]) => !value.isRedundant,
+    ),
+  );
+};
+
+const updateDashboardTabLocalStorage = (
+  dashboardPageId: string,
+  dashboardContext: DashboardContextForExplore,
+) => {
+  const dashboardsContexts = getDashboardContextLocalStorage();
+  setItem(LocalStorageKeys.dashboard__explore_context, {
+    ...dashboardsContexts,
+    [dashboardPageId]: dashboardContext,
+  });
+};
+
+const SyncDashboardState: React.FC<Props> = ({ dashboardPageId }) => {
+  const dashboardContextForExplore = useSelector<
+    RootState,
+    DashboardContextForExplore
+  >(
+    ({ dashboardInfo, dashboardState, nativeFilters, dataMask }) => ({
+      labelColors: dashboardInfo.metadata?.label_colors || EMPTY_OBJECT,
+      sharedLabelColors:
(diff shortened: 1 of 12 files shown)


rusackas:
CC @kgabryje who has looked into this sort of thing in this area before.

superset-frontend/src/dashboard/components/SyncDashboardState/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  dashboardContext: DashboardContextForExplore,
+) => {
+  const dashboardsContexts = getDashboardContextLocalStorage();
+  setItem(LocalStorageKeys.dashboard__explore_context, {
+    ...dashboardsContexts,
+    [dashboardPageId]: dashboardContext,
+  });
+};
+
+const SyncDashboardState: React.FC<Props> = ({ dashboardPageId }) => {
```
kgabryje, line 61: Shouldn't this be a hook rather than a component?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
```diff
@@ -322,7 +323,7 @@ const FilterValue: React.FC<FilterControlProps> = ({
       ) : (
         <SuperChart
           height={HEIGHT}
-          width="100%"
+          width={RESPONSIVE_WIDTH}
```
kgabryje, line 326: What's the advantage of using this `RESPONSIVE_WIDTH` instead of `100%`?

jinghua-qa:
> Amazing improvement! Left 2 minor comments for the code. As often is the case in such refactors, there's a high risk of introducing regressions. Is there any chance we could get some QA help @sadpandajoe @jinghua-qa ?

I will take a look for the regression

jinghua-qa:
Do some regression test around native filter , did not find issues, LGTM

kgabryje:
Thanks @jinghua-qa! @justinpark Would you rather address my comments in this PR or merge it and do a follow up if needed?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
justinpark: The RESPONSE_WIDTH change is related to the topic no.2 (`2. Redundant filter re-renders due to the dynamic sizing`)

superset-frontend/src/dashboard/components/SyncDashboardState/index.tsx
justinpark: The reason is related to the topic no.3 `3. descendants over-rendering`

> To further optimize performance, the dashboard state synchronization logic was moved to a separate component, effectively decoupling it from the DashboardPage component and preventing unnecessary rerenders of the DashboardPage.

+ In order to observe state updates and sync them to local storage, it is necessary to include a component with the hook. The most effective approach would be to implement a middleware. In this case, it is skipped to minimize any potential regression.

justinpark:
> Thanks jinghua-qa! justinpark Would you rather address my comments in this PR or merge it and do a follow up if needed?

@kgabryje please see my comments.

kgabryje (approved):
LGTM

PR #25903, opened 2023-11-08
fix(sqllab): invalid sanitization on comparison symbol

justinpark:
### SUMMARY
The current regular expression for HTML detection is inadequate, as it sanitizes the text containing comparison symbols. (i.e. `a <= 10 and b > 10`)
This commit updates the isProbablyHTML regex expression to fix the issue.

Before:
Screenshot 2023-11-07 at 6 18 15 PM

After:
Screenshot 2023-11-07 at 6 17 58 PM

### TESTING INSTRUCTIONS
Go to SQL Lab and run `select 'a <= 10 and b > 10'`
Check the result table

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/utils/html.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/utils/html.tsx
@@ -28,7 +28,7 @@ export function sanitizeHtml(htmlString: string) {
 }
 
 export function isProbablyHTML(text: string) {
-  return /<[^>]+>/.test(text);
+  return /<\/?[a-z][\s\S]*>/i.test(text);
 }
 
 export function sanitizeHtmlIfNeeded(htmlString: string) {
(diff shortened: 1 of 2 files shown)


superset-frontend/packages/superset-ui-core/src/utils/html.tsx
```diff
@@ -28,7 +28,7 @@ export function sanitizeHtml(htmlString: string) {
 }
 
 export function isProbablyHTML(text: string) {
-  return /<[^>]+>/.test(text);
+  return /<\/?[a-z][\s\S]*>/i.test(text);
```
john-bodley, line 31: @justinpark is there merit in using a non-regex implementation, i.e., something like [this](https://stackoverflow.com/a/55680987)?

justinpark: that looks good.let me try with the alternative solution.

john-bodley (approved):

PR #25919, opened 2023-11-08
fix: always denorm column value before querying values

hughhhh:
### SUMMARY

currently with some databases/engines the `/<datasource_type>/<int:datasource_id>/column/<column_name>/values/` endpoint values due to malformatted queries. One of the edge cases has to do with column names not being denormalized. To fix this we'll denormalize the column name before querying for values

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/datasource/api.py
+++ b/superset/datasource/api.py
@@ -116,14 +116,8 @@ class DatasourceRestApi(BaseSupersetApi):
 
         row_limit = apply_max_row_limit(app.config["FILTER_SELECT_ROW_LIMIT"])
         try:
-            # always denormalize column name before querying for values
-            db_engine_spec = datasource.database.db_engine_spec
-            db_dialect = datasource.database.get_dialect()
-            denomalized_col_name = db_engine_spec.denormalize_name(
-                db_dialect, column_name
-            )
             payload = datasource.values_for_column(
-                column_name=denomalized_col_name, limit=row_limit
+                column_name=column_name, limit=row_limit
             )
             return self.response(200, result=payload)
         except KeyError:
--- a/superset/datasource/api.py
+++ b/superset/datasource/api.py
@@ -116,8 +116,14 @@ class DatasourceRestApi(BaseSupersetApi):
 
         row_limit = apply_max_row_limit(app.config["FILTER_SELECT_ROW_LIMIT"])
         try:
+            # always denormalize column name before querying for values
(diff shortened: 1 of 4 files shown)


superset/datasource/api.py
```diff
@@ -116,8 +116,14 @@ def get_column_values(
 
         row_limit = apply_max_row_limit(app.config["FILTER_SELECT_ROW_LIMIT"])
         try:
+            # always denormalize column name before querying for values
+            db_engine_spec = datasource.database.db_engine_spec
+            db_dialect = datasource.database.get_dialect()
+            denomalized_col_name = db_engine_spec.denormalize_name(
+                db_dialect, column_name
+            )
```
betodealmeida, line 124: It's better to move this logic to the `datasource.values_for_column` method, so that the API doesn't have any business logic, and so that calls to `datasource.values_for_column` from other places also work as expected.

betodealmeida (dismissed):

superset/models/helpers.py
```diff
@@ -1341,36 +1342,34 @@ def get_time_filter(  # pylint: disable=too-many-arguments
...
-            if isinstance(col, dict):
-                cols[col.get("column_name")] = col
-            else:
-                cols[col.column_name] = col
-
-        target_col = cols[column_name]
-        tp = None  # todo(hughhhh): add back self.get_template_processor()
+        # always denormalize column name before querying for values
+        db_dialect = self.database.get_dialect()
+        denomalized_col_name = self.database.db_engine_spec.denormalize_name(
```
hughhhh, line 1347: denomarlized name will return the original value if it can't be adjusted by the engine:
[preset-io/superset@60e1526/superset/db_engine_specs/base.py#L1950](https://github.com/preset-io/superset/blob/60e1526f6acc5cbfd126a29ac7d6af9420f8a648/superset/db_engine_specs/base.py#L1950)

superset/models/helpers.py
```diff
@@ -1341,36 +1341,34 @@ def get_time_filter(  # pylint: disable=too-many-arguments
...
```
jfrag1, line 1346: nit: denomalized -> denormalized

jfrag1 (approved):
LGTM

eschutho (approved):

PR #25192, opened 2023-09-05
feat: Add slice_id to query object

hwmarkcheng:
### SUMMARY

Hi team,

We're looking to make a few changes that would allow folks to populate the `sql_query_mutator` with more useful params. Ultimately we're hoping to get chart_id and dataset_id into query comments, to automate our data masking with Satori. Open to discussion on the best practices to do so.

**These are the three items we're looking to change:**
- Change [mutate_query_from_config](https://github.com/hwmarkcheng/superset/pull/1/files#diff-af5e15a55ff9c81f441eb6b9b10dae13b2ee17fec614d3e3ad0057ed0dc814aeR819) to take in kwargs (Consistent with intended purpose of sql_query_mutator, see this [PR](https://github.com/apache/superset/pull/19083))
- Add `slice_id` to `query_obj `:  As addition context for `query_obj`. Having slice_id is versatile and can open doors for powerful integrations/automations. We originally wanted to add in the whole slice object for more options, but some integration tests failed since the object cannot be converted to a dictionary. Happy to discuss the best way to do this going forward.
- Add `chart_id` and `dataset_id` as default parameters for [mutate_query_from_config](https://github.com/hwmarkcheng/superset/pull/1/files)

**Example Usecase, Automated Data Access Requests via Satori:**
Context: We use Satori for our data access management.
- Users posts chart link to request access to specific schemas -> DAR team approves request -> We grab chart/dataset id from query -> Integrate with Satori to automatically identify all schemas in the chart and grant the user access to them

### TESTING INSTRUCTIONS

- Change [sql_query_mutator](https://github.com/hwmarkcheng/superset/pull/1/files#diff-c99ae4b2b09b756ab2189a99a9685229f9d12633fc2616c368ea869770f603bfR1212) to add kwargs as comments. **Note this will be reverted before committing (we don't want to change this)**
  - Click view query on any chart in a dashboard. You should now see the chart_id, as well as the dataset_id
  - Screenshot 2023-08-23 at 4 34 57 PM

Changed query_obj tests to account for additional parameters:
- [model_tests.py](https://github.com/hwmarkcheng/superset/pull/1/files#diff-2ec392cd10be9b443aedbb76af31cd47b9aaa773c34df9353c8d854cefcd9d07R591)
- [query_object_test.py](https://github.com/hwmarkcheng/superset/pull/1/files#diff-9a8c5397a1fe57ae5a2ac0df4f3c1f687a5fef7e6f1c6f2ad2c25ccc124dd46eR62)

[...]

PR #23114, opened 2023-02-17
chore: add execution_id to any available report logs

eschutho:
### SUMMARY
To improve logging, we're looking to be able to search for an execution ID to tie together logs from the entire alert/report Celery job. To do this, I have added an execution_id to the Screenshot classes (dashboard and chart), the slack and email send methods and the webdriver screenshot method. The report task is already passing the execution id through to the commands, etc, so I also passed the celery execution id through the thumbnail generation to the screenshot logs as well.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #20903, opened 2022-07-28
feat: add header_data into emails

AAfghahi:
### SUMMARY

This adds a new config mutator that adds logging data into a notification email should a user desire that. 

The mutator works similarly to the SQL mutator that is already in position, accepting kwargs so that the logging data can be customized. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25920, opened 2023-11-09
feat: global context for logging

eschutho:
### SUMMARY
We have a few examples in the codebase where for logging purposes we need to pass data down to parts of the codebase that have no need to know about this data. It creates very coupled and bloated modules that often are just passing data down without even acting on it. This change proposes the use of a global context dictionary that is used for read-only purposes for logging. 

There are three places where we can use this functionality. One is an [open PR](https://github.com/apache/superset/pull/23114) that I wrote that helps us to log a report execution id to tie together multiple logs to the same report execution. 
Another example is [passing report header data](https://github.com/apache/superset/pull/20903) to an smtp event for logging in a mail server.
And the third use case is to pass [dashboard_id and chart_id](https://github.com/apache/superset/pull/25192) into the sql mutator for logging. 

I have rewritten the first two examples each in one commit in this PR for diffing the implementations for each with a global context. I'll add how the third example (sql mutator) either in this PR or in a subsequent one. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/reports/commands/execute.py
+++ b/superset/reports/commands/execute.py
@@ -63,7 +63,6 @@ from superset.reports.models import (
     ReportRecipientType,
     ReportSchedule,
     ReportScheduleType,
-    ReportSourceFormat,
     ReportState,
 )
 from superset.reports.notifications import create_notification
@@ -71,7 +70,7 @@ from superset.reports.notifications.base import NotificationContent
 from superset.reports.notifications.exceptions import NotificationError
 from superset.tasks.utils import get_executor
 from superset.utils.celery import session_scope
-from superset.utils.core import HeaderDataType, override_user
+from superset.utils.core import override_user
 from superset.utils.csv import get_chart_csv_data, get_chart_dataframe
 from superset.utils.decorators import context
 from superset.utils.screenshots import ChartScreenshot, DashboardScreenshot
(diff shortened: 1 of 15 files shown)


john-bodley:
@eschutho out of interest what's the criteria for logging additional context and is there any concern about how much additional data we're passing over the wire? How do we resolve situations where different organizations have different requirements?

eschutho:
Thanks @john-bodley I'd love to get some early feedback on this idea. As I'm refactoring existing code, I've been targeting that one constraint would be to only store ids in the global context. When logging, the module responsible for logging can fetch the global context id and then any relevant information from the metadata that it needs. The context doesn't need to store or know about all the data that the logger needs, just the minimum requirements. 

My main concern is more around the global context getting messy and overused. I don't think there are issues about sending information over a network request, since the context should not be returned in any api responses, but I don't want it to become a dumping ground for passing around any kind of state because that would be an antipattern. Instead, if we limit it to single resource ids, or perhaps a list of ids, I think that could get us where we need for now, and we can reevaluate if other situations arise where we feel we would need this.

eschutho:
I'm closing this and splitting it into smaller PRs.

PR #25934, opened 2023-11-09
fix(tag): update state to clear form on success

hughhhh:
### SUMMARY

Currently the state in the tags form isn't being cleared. So the values from a previous entry are staying whenever a user wants to create a new tag. To fix this I refactored the component to have a `clearForm` function that will clear the state in the component whenever called.

https://github.com/apache/superset/assets/27827808/4c731cd9-3ffc-4cdc-a218-46561cf8f5e4

### TESTING INSTRUCTIONS

1. Goto Tags Page
2. Create a new tag (enter name + description)
3. Click Save
4. Click `+ Tag` form should be empty

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/tags/TagModal.tsx
+++ b/superset-frontend/src/features/tags/TagModal.tsx
@@ -263,13 +263,7 @@ const TagModal: React.FC<TagModalProps> = ({
     <Modal
       title={modalTitle}
       onHide={() => {
-        if (clearOnHide) {
-          setTagName('');
-          setDescription('');
-          setDashboardsToTag([]);
-          setChartsToTag([]);
-          setSavedQueriesToTag([]);
-        }
+        if (clearOnHide) clearTagForm();
         onHide();
       }}
       show={show}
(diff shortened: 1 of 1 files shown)


lilykuang (approved):

john-bodley (dismissed):
@hughhhh would you mind adding some testing instructions and unit tests (if possible). I've also added @justinpark as a reviewer as it seems unclear why we need to clear the form, i.e., this seems like an atypical pattern.

hughhhh:
> @hughhhh would you mind adding some testing instructions and unit tests (if possible). I've also added @justinpark as a reviewer as it seems unclear why we need to clear the form, i.e., this seems like an atypical pattern.

Added test instructions, but the problem is when users are creating a Tag and save the form. The values are kept on the next instantiation of the modal, when they should it should be cleared for new data.

john-bodley:
@justinpark would you mind reviewing this as I'm somewhat naive when it comes to frontend.

PR #25935, opened 2023-11-09
Update installing-superset-from-scratch.mdx

lgharibpmu:
Add FLASK_APP environnement variable prior to any superset execution and SUPERSET_CONFIG_PATH.

### ADDITIONAL INFORMATION

- [x] Has associated issue: [25936](https://github.com/apache/superset/issues/25936)
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/installation/installing-superset-from-scratch.mdx
+++ b/docs/docs/installation/installing-superset-from-scratch.mdx
@@ -123,6 +123,12 @@ First, start by installing `apache-superset`:
 pip install apache-superset
 ```
 
+Export the FLASK_APP environnement variable in order to be able to execute any `superset` command:
+
+```
+export FLASK_APP=superset
+```
+
 Then, you need to initialize the database:
 
 ```
@@ -133,11 +139,16 @@ superset db upgrade
 Note that some configuration is mandatory for production instances of Superset. In particular, Superset will not start without a user-specified value of SECRET_KEY. Please see [Configuring Superset](https://superset.apache.org/docs/installation/configuring-superset).
 :::
 
+In order to make the configuration file available to `superset` you need to execute the following command:
+
+```
+export SUPERSET_CONFIG_PATH=/path/to/your/superset_config.py
+```
+
 Finish installing by running through the following commands:
 
 ```
 # Create an admin user in your metadata database (use `admin` as username to be able to load the examples)
-export FLASK_APP=superset
 superset fab create-admin
 
 # Load some data to play with


john-bodley:
@lgharibpmu would you mind expanding your PR description to explain why this change is needed?

ddxv:
> @lgharibpmu would you mind expanding your PR description to explain why this change is needed? Typically one doesn't need to explicitly set said environment variable.

@john-bodley I'm also new, following the superset from scratch for the first time and reached the same conclusions that @lgharibpmu did. Not sure what went wrong, so I made a clean environment to show you how I tried to follow the instructions:

https://superset.apache.org/docs/installation/installing-superset-from-scratch

image

1) Fresh Python3.11 environment, `pip install apache-superset`
2) Try `superset db upgrade`: This fails as it says FLASK_APP environment variable not set. So I tried `export FLASK_APP=superset` as I saw it was actually in the next section.
3) Try `superset db upgrade` This time fails because environment variable SECRET_KEY not set. Following the link in the configuration setup I saw SECRET_KEY could be set by adding SECRET_KEY in configure.py and setting it's location in `SUPERSET_CONFIG_PATH`.
4) Try `superset db upgrade` this time it works.

Again, not sure if I misinterpreted the instructions or if there is something missing. Hope this helps clarify.

lgharibpmu:
> @lgharibpmu would you mind expanding your PR description to explain why this change is needed? Typically one doesn't need to explicitly set said environment variable.

Hi @john-bodley thank you for your reply. Actually, as @ddxv explained, when installing Superset from scratch if a user doesn't set theses two environment variables prior to the rest of the steps, he faces theses errors:

- Error: Could not locate a Flask application. Use the 'flash --app' option, 'FLASK_APP' environment variable, ...
- Refusing to start due to insecure SECRET_KEY

john-bodley:
@ddxv and @lgharibpmu there's the [.flaskenv](https://github.com/apache/superset/blob/master/.flaskenv) file which sets both the `FLASK_APP` and `FLASK_DEBUG` variables (when the later is truthy one doesn't need to set the `SECRET_KEY`).

If you were to run `flask run` as opposed to `superset run` then the `.flaskenv` file would be read. 

@craig-rueda, @dpgaspar, @michael-s-molina, and @sfirke  I wonder if it's worth changing all the development documents, CI jobs, etc. to use `flask run` instead so we adhere to the DRY principle. Alternatively we could look into augmenting [this](https://github.com/apache/superset/blob/f6ba75a1859b89b196bb09970d0b6f8cafd6ee99/superset/cli/main.py#L34-L44) CLI command to use the `.flaskenv` file.

SLE-TheRealOne:
Actually for having recently installed superset in an AWS environment for production target purposes, the modification proposed by  @lgharibpmu is correct and other additions may be needed as i encountered multiple obastacles listed below :

* As Superset, in it's current latest 3.0.1 requires Python >=3.9, using pip3 rather than pip solves a lot of manual dependencies resolution problems (some of them below)
* Some downgrades (in a requirements.txt file for instance) are mandatory for installation to succeed
  * urllib3==1.26.17
  * idna==2.10
* In the "Installing and Initializing Superset" section non-mandatory and mandatory parts are mixed, without proper explanation on each step so that exprimentation is needed to determine which is which. For instance : 
  * `superset db upgrade` creates 86 required tables. I understand it cannot be embedded in pip package as underlying database may vary but more then "initialize the database" would be interesting. In addition, it does not upgrade the database (suggesting a working one will be upgraded) but initialize a working set from scratch. Maybe name it `superset db init` would be more appropriate
  * The `export FLASK_APP=something` export is mandatory but it's not explained why. In addition, the FLASK_APP also sets the base url of a working instance "/${FLASK_APP}/welcome" redirection when calling "/" for instance. This can be important when positioning a load balancer in front of multiple instances
  * A `superset init` command does exist but requires a working database and only writes in tables so maybe make it a subcommand of the superset db. Something like `superset db init-roles-permissions`
  * The javascript asset section is not-mandatory for a base working instance so no need to build any javascript assets
  * The result of `superset run -p 8088 --with-threads --reload --debugger` does not start on `hostname:port` but on `127.0.0.1:port`. It would with the `--host option`

Another suggestion in the "Configuring Superset" page for PostgeSQL package to install : 
installing psycog2 also requires multiple downgrades depending each on installed >=3.9 python version whereas psycog2-binary package does not and describe more accurate what is needed : a binary to connect to postgre

I would be glad to propose such changes. Should i enrich @lgharibpmu branch ?

Thanks guys

mmyrte:
For me, the confusion starts with the title of the page - doesn't "Installing From Scratch" usually mean installing from source?

1. Is it intended that by the end of the tutorial, I have a **minimally configured best-practice production install**? I cannot run any `superset cmd` without setting a secret key, which is being described as only necessary for production. https://github.com/apache/superset/blob/29633e7d8cad7b52669a9e6f5e064cd34da803a3/docs/docs/installation/installing-superset-from-scratch.mdx#L132-L134
2. Is it intended that I get an **installation for debugging**? I'm being told to use `superset run --debugger`. https://github.com/apache/superset/blob/29633e7d8cad7b52669a9e6f5e064cd34da803a3/docs/docs/installation/installing-superset-from-scratch.mdx#L155-L156
3. Is it intended that I **clone the repo** and build my own TS/JS/CSS assets, given that I'm being told to cd into a folder that doesn't exist after `pip install apache-superset`? https://github.com/apache/superset/blob/29633e7d8cad7b52669a9e6f5e064cd34da803a3/docs/docs/installation/installing-superset-from-scratch.mdx#L149-L153

From my experience, I expect pages titled `Install [XY]` to be linear tutorials with a minimum of branches, leaving me with a minimally configured installation according to best practices. In this case, I suppose that would mean something like the following structure:

- Install from PyPI
  - Disclaimer: Where production needs deviate from localhost/testing needs, the relevant docs will be linked to.
  - Disclaimer: We default to `pip` and `python`, but be aware that your OS might use `pip3` and `python3`
  - OS Dependencies
    - Check that these are actually needed! e.g. Postgres is listed for macOS but not actually relevant for a minimal sqlite based install.
  - Virtual Environment Setup
    - Create
    - Activate
    - `pip install apache-superset`
      - Mention that additional dependencies come into play for prod, like DBs, their drivers, WSGI servers, proxies...
  - Minimally Configuring Superset
    - Find `$PYTHONPATH`
      `python -c "import sys; print('\n'.join(sys.path))"`
    - Put `superset_config.py` on pythonpath
      - set `SECRET_KEY`
      - set listening address
      - set port
    - Mention <https://superset.apache.org/docs/installation/configuring-superset>
  - Initializing Superset
    - `superset db upgrade`
    - `superset fab create-admin`
    - `superset init`
  - Running as test instance
    - `superset run`
- ~Installing Superset with Helm in Kubernetes~ <- should be removed since there's now [Installing on Kubernetes](https://superset.apache.org/docs/installation/running-on-kubernetes)

Like @SLE-TheRealOne, I would be happy to support either @lgharibpmu or @john-bodley in writing or testing a revamped tutorial on a vanilla macOS or Linux installation. God knows I appreciate the work everyone's done on this project - but I have a hard time getting started, and that's just not ideal. While I am not a project member who could initiate a more high-level discussion about what-goes-where, à la <https://diataxis.fr/>, I am super happy to lend my outsider's eyes for feedback.

rusackas:
Hi there... just revisiting this. The title of the page actually did change to "Installing from PyPI" rather than "from scratch" so I guess you weren't alone in that thought :D 

To merge this PR, it would need a rebase since the name of that file changed as well. I assume the rest of your changes are still pertinent, but I'll leave that to those already discussing things on the thread :)

PR #25947, opened 2023-11-10
fix: update flask-caching to avoid breaking redis cache, solves #25339

### LINKED ISSUE #25339: Update flask-caching version to 2.0+ to avoid breaking redis cache
#### How to reproduce the bug

If we have flask-caching=~1.11.1 and cachelib>=9 (flask-caching did not pin down cahcelib dependency until 2.0, so it's very likely to happen...), we end up with this https://github.com/pallets-eco/flask-caching/issues/381

caused by this very line:
https://github.com/pallets-eco/flask-caching/blob/ac13e01baa11d65c68dcba64f99ccaafd9b5419c/src/flask_caching/backends/rediscache.py#L62

The fix might be as small as this (even if I'd like to test more in depth): https://github.com/ggbaro/superset-ggbaro/commit/185e59ea80901600d5dc88ddc1b67e1f191c9504
Do you suggest to make a PR?

### Expected results

App starts normally

### Actual results

flask-caching breaks with redis cache

### Environment

- browser type and version:
- superset version: `>=2`
- python version: `3.9 & 3.10`

### Checklist

  - see  https://github.com/pallets-eco/flask-caching/issues/381

ggbaro:
### SUMMARY
If we have `flask-caching=~1.11.1` and `cachelib>=9` (`flask-caching` did not place the necessary constraints on the `cahcelib` dependency until 2.0, so it's very likely to happen...), we end up with this https://github.com/pallets-eco/flask-caching/issues/381
caused by this very line:
https://github.com/pallets-eco/flask-caching/blob/ac13e01baa11d65c68dcba64f99ccaafd9b5419c/src/flask_caching/backends/rediscache.py#L62

We can fix updating dependencies

### ADDITIONAL INFORMATION

Fixes #25339

- [x] Has associated issue: closes #25339
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,7 @@ setup(
         "deprecation>=2.1.0, <2.2.0",
         "flask>=2.2.5, <3.0.0",
         "flask-appbuilder>=4.3.9, <5.0.0",
-        "flask-caching>=1.11.1, <2.0",
+        "flask-caching>=2.0, <3",
         "flask-compress>=1.13, <2.0",
         "flask-talisman>=1.0.0, <2.0",
         "flask-login>=0.6.0, < 1.0",
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,7 @@ setup(
         "deprecation>=2.1.0, <2.2.0",
         "flask>=2.2.5, <3.0.0",
         "flask-appbuilder>=4.3.9, <5.0.0",
-        "flask-caching>=2.0, <3",
+        "flask-caching>=2.1.0, <3",
         "flask-compress>=1.13, <2.0",
         "flask-talisman>=1.0.0, <2.0",
         "flask-login>=0.6.0, < 1.0",
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -31,7 +31,7 @@ bottleneck==1.3.7
     # via pandas
 brotli==1.0.9
     # via flask-compress
-cachelib==0.6.0
+cachelib==0.9.0
     # via
     #   flask-caching
     #   flask-session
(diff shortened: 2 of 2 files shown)


john-bodley:
Thanks @ggbaro for the fix. Regarding your comment:

> `flask-caching` did not pin down `cahcelib` dependency until 2.0

it seems a more accurate statement rather than "pin" would be "constrain", i.e., 

> `flask-caching` did not place the necessary constraints on the `cahcelib` dependency until 2.0

Would you mind updating the PR description to reflect this?

setup.py
```diff
@@ -84,7 +84,7 @@ def get_git_sha() -> str:
         "deprecation>=2.1.0, <2.2.0",
         "flask>=2.2.5, <3.0.0",
         "flask-appbuilder>=4.3.9, <5.0.0",
-        "flask-caching>=1.11.1, <2.0",
+        "flask-caching>=2.0, <3",
```
villebro, line 87: Minor comment: I think it's been customary to set the lower bound to the version we're locking to in the requirements. I would maybe consider doing the following, unless there's strong feelings about the need for supporting `>=2.0.0, <2.1`
```suggestion
        "flask-caching>=2.1.0, <3",
```

villebro (approved):
LGTM with an optional comment to restrict to `>=2.1`

villebro:
FYI @ggbaro I updated the description by prefixing the issue number with "closes: ", which will automatically close the referenced issue when this PR is merged.

setup.py
john-bodley: Plus one on @villebro's suggestion.

PR #25961, opened 2023-11-12
chore: harmonize and clean up list views

villebro:
### SUMMARY
Currently, there is a lot of discrepancy between how list views look/work:
- The "Search" filter is sometimes first or last. It is also sometimes called "Name".
- The ordering of the other filters is highly inconsistent, and not in line with the column ordering.
- The first column is sometimes "Name", other times the asset type, e.g. "Dashboard".
- The Owners column is missing on Charts page.
- The modified column is sometimes called "Modified", other times "Last modified"
- We sometimes have the "Created by" and "Created on" columns, but usually not.

In addition, with the introduction of Tags, the list views have become very cluttered, raising the need for removing less relevant fields.

This makes the layout consistent across the following list views:
- Annotation Layers
- Dashboards
- Charts
- Datasets
- Row Level Security
- Databases
- Tags
- Alerts & Reports
- CSS Templates
- Saved Queries

The following changes are made:
- The Search/Name filter is placed first. The filter name is set as "Name", as "Search" might indicate that it searches other fields, too.
- Owners is added to all list views where relevant. This was missing on the Charts view.
- The "Created by" and "Created at" fields are removed, as they're largely irrelevant.
- The "Modified by" and "Modified at" fields are merged, so that the modifier is shown when hovering.
- The "Created by" filter is replaced by "Modified by", to make it possible to search based on who last modified the asset.
- The Filters are reordered to be in line with the column orders. This is with the exception of less commonly needed filters, like "Favorite" and "Published".
- Single select filters are changed from plural to singular. This particularly affects the "Tags", "Owners" and "Dashboards" filters, which are now "Tag", "Owner" and "Dashboard" respectively. Note, that we may want to make these multi-select at some point, but that would require introducing an IN filter, either locally or in FAB.
- The facepile on "Tags" is removed, as that should IMO only be used for Owners.

#### Charts
image

#### Dashboards
image

#### Dataset
image

#### Saved Queries
image

#### Row Level Security
image

#### Database
image

#### CSS Templates
image

#### Tags
image

#### Alerts & Reports
image

#### Annotation Layers
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/AsyncAceEditor/index.tsx
```diff
@@ -25,6 +25,7 @@ import {
 } from 'brace';
 import AceEditor, { IAceEditorProps } from 'react-ace';
 import { acequire } from 'ace-builds/src-noconflict/ace';
+import 'ace-builds/webpack-resolver';
```
villebro, line 28: Without this we get a failed network request when opening up the edit modal in the CSS list view + no syntax highlighting (it's even uglier with `dev-server`):

image

After adding this, no more failed network request + CSS syntax highlighting works:

image

Adding this import was a recommendation in this thread, and seemed to be an ok fix: https://github.com/securingsincity/react-ace/issues/725

kasiazjc:
Thanks @villebro for cleaning it up 🙏 very much appreciated! 

I have one concern about removing "created by" column. I think there is no other place (besides metadata bar in explore, but we don't have it in dashboards?) where you could see who created a dashboard or a chart + owners are not required I think. So if there are some random dashboards/charts it could be hard to find the right person to talk to about it if there is no information like this easily available. Thoughts?

villebro:
> I have one concern about removing "created by" column. I think there is no other place (besides metadata bar in explore, but we don't have it in dashboards?) where you could see who created a dashboard or a chart + owners are not required I think. So if there are some random dashboards/charts it could be hard to find the right person to talk to about it if there is no information like this easily available. Thoughts?

Thanks for the review @kasiazjc ! ❤️ I agree that removal of "Created by" column may indeed be controversial. However, some reasoning behind this proposal:
- When a user creates a chart/dashboard/dataset, they always become the owner of it. It's only the sample assets that don't have an owner.
- We were already missing it in the Dataset list view.
- When a user creates a dashboard, they also automatically become the last modifier.
- I find that it's often of very limited utility to know who created a chart/dashboard - usually you really want to know who owns it, or who modified it last (the creator can often have left the team/org, hence may no longer be relevant, especially for long lived assets).

I'd be curious to hear thoughts from other users with big deployments - Ping @eschutho @sfirke @cccs-rc @john-bodley @michael-s-molina @giftig 

Something to consider:
- We could add the filter for the creator, even without adding it to the list
- We may even add the creator to the hover of "Last modified". However, that may be confusing, so maybe that's not a great idea..

john-bodley:
Ha! @villebro I was just speaking to @michael-s-molina about this and how inconsistent these list views are.  The columns and reorderings generally look great. A few comments:

1. I'm mixed about whether the creator (aka the first modifier) should remain and I think @villebro raised some good points in terms of why it should be removed. I sense the reason the creator was missing from a dataset is this was often misleading as for non-virtual datasets it merely is the person who registered the dataset.
2. You've renamed the dashboard `Title` column to `Name` (for consistency) and I was wondering whether this should be reflected throughout for consistency, i.e., in the modal when editing.

kasiazjc:
> > I have one concern about removing "created by" column. I think there is no other place (besides metadata bar in explore, but we don't have it in dashboards?) where you could see who created a dashboard or a chart + owners are not required I think. So if there are some random dashboards/charts it could be hard to find the right person to talk to about it if there is no information like this easily available. Thoughts?
> 
> Thanks for the review @kasiazjc ! ❤️ I agree that removal of "Created by" column may indeed be controversial. However, some reasoning behind this proposal:
> 
> * We have very limited screen real estate, and this is a major issue especially when the Tagging system is enabled.
> * When a user creates a chart/dashboard/dataset, they always become the owner of it. It's only the sample assets that don't have an owner.
> * We were already missing it in the Dataset list view.
> * When a user creates a dashboard, they also automatically become the last modifier.
> * I find that it's often of very limited utility to know who created a chart/dashboard - usually you really want to know who owns it, or who modified it last (the creator can often have left the team/org, hence may no longer be relevant, especially for long lived assets).
> 
> I'd be curious to hear thoughts from other users with big deployments - Ping @eschutho @sfirke @cccs-rc @john-bodley @michael-s-molina @giftig
> 
> Something to consider:
> 
> * We could add the filter for the creator, even without adding it to the list
> * We may even add the creator to the hover of "Last modified". However, that may be confusing, so maybe that's not a great idea..

That makes sense to me! Consider me convinced. Thank you 🩵

kasiazjc (approved):
design LGTM

giftig:
@villebro in my case I sometimes look for virtual datasets I've created by searching by creator. I don't think it's necessary for the list view (in fact I've only just noticed it's missing from there since "modified by" was generally the creator).

I think it'd be good to keep the ability to search by it. We encourage our users to always create virtual datasets when building charts, as we want them to use  a macro we've written which enforces a date filter and prevents accidentally querying too much data. That means we have a lot of virtual datasets (added to thousands of physical ones).

Searching by owner would usually work, but users working with their teams might be assigning ownership to other people and it might get muddied. I'm somewhat speculating here as we don't have enough usage information from our users. Searching by modified would probably be misleading if someone else has since touched your dataset.

So I think I'd lean towards continuing to allow searching by original creator, but agree that it's best to consistently exclude it from list views.

giftig:
I just looked again, and I'm wrong; even in 2.1 I can't search datasets by creator, I'm actually searching by owner. I think this is fine then; I never even noticed the distinction between owner and creator when searching. We're also looking forward to getting tagging when we upgrade to 3.x, which will make searching for and categorising charts and dashboards more powerful anyway.

superset-frontend/src/pages/ChartList/index.tsx
```diff
@@ -438,44 +434,30 @@ function ChartList(props: ChartListProps) {
...
-            original: { last_saved_at: lastSavedAt },
+            original: { last_saved_at: changedOn, last_saved_by: changedBy },
           },
         }: any) => (
-          <span className="no-wrap">
-            {lastSavedAt ? moment.utc(lastSavedAt).fromNow() : null}
-          </span>
+          <AuditInfo
+            type={AuditInfoType.Modified}
+            date={changedOn}
```
kgabryje, line 453: Shouldn't we use `changed_on_delta_humanized` here? I'm seeing an inconsistency between ChartList and for example DashboardList

image

image

kgabryje: Nevermind, it was noted by Michael already 🙂

villebro:
@michael-s-molina @kgabryje thanks for catching that non-humanized modified bug! I'll sort it out right away. Regarding Michael's comment:

> My suggestion would be to add a card to the [4.0 board](https://github.com/orgs/apache/projects/292) to remove the creator in the next major and for this PR, I would change the creator to be an avatar similar to the owners for consistency and to preserve real estate.

I think it would be confusing to use the avatars for both Owners and creators, and I believe that's the reason why creators/modifiers have been shown as text until now. If we do want to keep creators as a column, I propose we retain the username as-is, and not use the facepile component for them.

@kasiazjc how do you feel about showing creators as an avatar vs showing it as the full name as text (current behavior)?

sfirke:
These changes all look excellent to me!  We don't use the "created by" field and I'd be happy to see it gone.

To me it's right on the borderline of being a tiny breaking change or not, I don't feel strongly if that part is held off for 4.0 or done now.  If it's easy to pull that part out for 4.0, great, if it's significant extra work then I'd just say merge it.

villebro:
> These changes all look excellent to me! We don't use the "created by" field and I'd be happy to see it gone.
> 
> To me it's right on the borderline of being a tiny breaking change or not, I don't feel strongly if that part is held off for 4.0 or done now. If it's easy to pull that part out for 4.0, great, if it's significant extra work then I'd just say merge it.

Thanks for the review @sfirke . I also agree that this feels like an acceptable breaking change, and could be swept under the "UX debt cleanup" rug 😛 I defer to @michael-s-molina and @eschutho for semver compliance - if you feel strongly about retaining the existing functionality then I'm happy to "disagree and commit" and add it back, but otherwise I propose doing like Sam said above.

superset-frontend/src/pages/SavedQueryList/index.tsx
```diff
@@ -497,28 +489,22 @@ function SavedQueryList({
...
-  if (isFeatureEnabled(FeatureFlag.TAGGING_SYSTEM) && canReadTag) {
-    filters.push({
-      Header: t('Tags'),
-      id: 'tags',
-      key: 'tags',
-      input: 'search',
-      operator: FilterOperator.savedQueryTags,
-    });
-  }
-
```
villebro, line 521: This was done outside the useMemo, causing the tags filter to get appended multiple times to the filter array: 
image

villebro:
> * The filter is called `Owner` but the list header is `Owners` (plural).
> * Some filters use plural form and others use singular form in a inconsistent manner. I believe the nomenclature here should reflect the filter type. Singular for filters that allow only one value and plural for filters that allow multiple values.

I've updated these so that the single select filters are now singular (Tag, Owner).

> * `Last modified` generally has the form of `15 hours ago` but in the charts list it's `2023-11-30T13:53:56.840190`

Good catch - fixed.

> * The filter is called `Chart Type` in the charts list but in the datasets list it's called `Type`

Renamed to "Type" throughout.

> * The filter in the tags list is called `Search` instead of `Name`

Renamed to "Name".

> * In the database list the column is called `Name` but in the modal it's called `Display name`

Renamed "Display name" to "Name" in the modal.

john-bodley:
Though I believe @giftig's comment is now obsolete (per a self correction), regarding 

> @villebro in my case I sometimes look for virtual datasets I've created by searching by creator. I don't think it's necessary for the list view (in fact I've only just noticed it's missing from there since "modified by" was generally the creator).

I think if we have a filter for an attribute, said attribute should appear in the table. The reason being is trust (or lack there of), i.e., if we're filtering by datasets created by Jane Doe then seeing Jane Doe's name in the result set under the appropriately name column serves as positive affirmation that the filters are working correctly.

john-bodley:
@michael-s-molina you previously mentioned, 

> The filter is called `Chart Type` in the charts list but in the datasets list it's called `Type`

do you mean that the column is called `Visualization type`?

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -109,14 +109,12 @@ describe('Charts list', () => {
 
     it('should load rows in list mode', () => {
       cy.getBySel('listview-table').should('be.visible');
-      cy.getBySel('sort-header').eq(1).contains('Chart');
+      cy.getBySel('sort-header').eq(1).contains('Name');
       cy.getBySel('sort-header').eq(2).contains('Visualization type');
```
john-bodley, line 113:
```suggestion
      cy.getBySel('sort-header').eq(2).contains('Type');
```

villebro:
> > Thanks for the review comments @michael-s-molina - I've addressed all bugs/comments, but I'll still defer from reintroducing the creation columns until we decide on whether or not we want to keep it until 4.0.
> 
> @villebro I raised this question during the Town Hall and folks seem to be ok with removing the creator column in this PR.

Thanks for raising this @michael-s-molina ! To strike a compromise, I introduced a "Modified by" filter to make it possible to search based on the last modifier. I'll update the screens to reflect this and ping you for a final review once CI is green.

villebro:
/testenv up FEATURE_TAGGING_SYSTEM=True FEATURE_ALERT_REPORTS=True

michael-s-molina (approved):
LGTM. Thanks for all the consistency improvements @villebro! I left some non-blocking comments. Given the number of changed files, a second pair of 👀 would be good @kgabryje.

rusackas (approved):
Approving as cypress codeowner

superset-frontend/src/components/AuditInfo/AuditInfo.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { AuditInfo, AuditInfoProps, AuditInfoType } from '.';
+
+const TEST_DATE = '2023-11-20';
+const USER = {
+  id: 1,
+  first_name: 'Foo',
+  last_name: 'Bar',
+};
+
+const Wrapper: React.FC<AuditInfoProps> = ({ type, user, date }) => (
```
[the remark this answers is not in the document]
villebro: Thanks, I didn't know about this, very helpful and much cleaner!

superset-frontend/src/components/AuditInfo/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import Owner from 'src/types/Owner';
+import { Tooltip } from 'src/components/Tooltip';
+import getOwnerName from 'src/utils/getOwnerName';
+import { t } from '@superset-ui/core';
+
+export const enum AuditInfoType {
+  Modified = 'Modified',
+  Created = 'Created',
+}
```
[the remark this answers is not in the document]
villebro: Done

kgabryje (approved):

PR #25967, opened 2023-11-13
fix(typings): model_id is a multiple option

gnought:
### SUMMARY
`superset compute-thumbnails` accepts multiple `--model_id` options. The `model_id` should be a `List[int]`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/cli/thumbnails.py
+++ b/superset/cli/thumbnails.py
@@ -56,13 +56,13 @@ logger = logging.getLogger(__name__)
     default=False,
     help="Force refresh, even if previously cached",
 )
-@click.option("--model_id", "-i", "model_ids", multiple=True)
+@click.option("--model_id", "-i", multiple=True)
 def compute_thumbnails(
     asynchronous: bool,
     dashboards_only: bool,
     charts_only: bool,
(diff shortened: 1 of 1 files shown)


superset/cli/thumbnails.py
```diff
@@ -76,7 +76,7 @@ def compute_thumbnails(
     def compute_generic_thumbnail(
         friendly_type: str,
         model_cls: Union[type[Dashboard], type[Slice]],
-        model_id: int,
+        model_id: list[int],
```
villebro, line 79: Nit: For disambiguation and clarity, we may want to call the argument in the inner function `model_ids`.

villebro (commented):
LGTM, thanks for the cleanup! LMKWYT about my small improvement proposal.

villebro:
Let's not change the outer variable name, only the one in the inner function.

gnought:
@villebro Thanks for the advise. Amended.

john-bodley (approved):

PR #25975, opened 2023-11-13
feat(alerts):  Adding custom Header and StyledPanel components for use in alerts&reports modal redesign

fisjac:
### SUMMARY

Context: This change is part of a broader design proposal https://github.com/apache/superset/discussions/25729 to change the Alerts & Reports modal requested by the Preset team. 

Added two components to the alerts frontend directory in preparation for future changes to the alerts&reports modal: 
(i) StyledPanel component which adds custom styling to the Antd Collapse.Panel component.
(ii) A ValidatedPanelHeader component which takes the following props {
  title: string;
  subtitle: string;
  required: boolean;
  validateCheckStatus: boolean;
};

#### BEFORE

image

#### AFTER

image
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/components/StyledPanel.tsx
+++ b/superset-frontend/src/features/alerts/components/StyledPanel.tsx
@@ -1,0 +1,56 @@
+import React from 'react';
+import { css, SupersetTheme } from '@superset-ui/core';
+import { Collapse as AntdCollapse } from 'antd';
+import { CollapsePanelProps } from 'antd/lib/collapse';
+
+const anticonHeight = 12;
+const antdPanelStyles = (theme: SupersetTheme) => css`
+  .ant-collapse-header {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: flex-start;
+    padding: 0px ${theme.gridUnit * 4}px;
+
+    .anticon.anticon-right.ant-collapse-arrow {
+      padding: 0;
+      top: calc(50% - ${anticonHeight / 2}px);
+    }
+
+    .collapse-panel-title {
+      font-size: ${theme.gridUnit * 4}px;
+      font-weight: ${theme.typography.weights.bold};
+      line-height: 130%;
+    }
+
+    .collapse-panel-subtitle {
+      color: ${theme.colors.grayscale.base};
+      font-size: ${theme.typography.sizes.s}px;
+      font-weight: ${theme.typography.weights.normal};
+      line-height: 150%;
+      margin-bottom: 0;
+    }
+
+    .collapse-panel-asterisk {
+      color: var(--semantic-error-base, ${theme.colors.warning.dark1});
+    }
+    .validation-checkmark {
+      width: ${theme.gridUnit * 4}px;
+      height: ${theme.gridUnit * 4}px;
+      margin-left: ${theme.gridUnit}px;
+      color: ${theme.colors.success.base};
+    }
+  }
+`;
+
+export interface PanelProps extends CollapsePanelProps {
+  children?: React.ReactNode;
+}
+const StyledPanel = (props: PanelProps) => (
+  <AntdCollapse.Panel
+    css={(theme: SupersetTheme) => antdPanelStyles(theme)}
+    {...props}
+  />
+);
+
+export default StyledPanel;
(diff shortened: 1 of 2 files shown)


superset-frontend/src/components/Panel/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const header = (
+    <div className="collapse-panel-header">
+      <div className="collapse-panel-title">
+        {/* @ts-ignore */}
+        <span>{t(props.header.title)}</span>
+        <span className="collapse-panel-asterisk">{asterisk}</span>
+        <span className="validation-checkmark">{checkmark}</span>
+      </div>
+      <p className="collapse-panel-subtitle">
+        {/* @ts-ignore */}
```
hughhhh, line 90: why do we need the ts-ignore here?

superset-frontend/src/components/Panel/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+        {props.header.subtitle ? t(props.header.subtitle) : undefined}
+      </p>
+    </div>
+  );
+
+  // eslint-disable-next-line no-param-reassign
+  props.header = header;
```
hughhhh, line 97: why do we need to reassign the `props.header` here?

superset-frontend/src/components/Panel/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  children?: React.ReactNode;
+}
+
+const Panel = ({ ...props }: PanelProps): JSX.Element => {
+  if (!props.header.hasOwnProperty('title')) {
+    return <AntdPanel {...props} />;
+  }
+
+  let asterisk;
+  // @ts-ignore
```
hughhhh, line 70: why do we need the ts-ignore here?

superset-frontend/src/components/Panel/index.tsx
fisjac: Since props.header can possibly take type string, TS gets angry saying that there's no property "title" on type string. Due to the conditional check in lines 65-67, we know that if this logic block is reached, props.header will be an object containing the title property.

john-bodley:
@fisjac would you mind updating the PR description with a more high level overview/context of what the goal of this change is and how you intent to leverage the enriched metadata? It's not apparent what the current use case is.

michael-s-molina (changes requested):
Thanks for the PR @fisjac. I can understand the need to customize the header in your installation. I think the best approach here would be to actually receive the Panel component as the parameter instead of creating one that might be sufficient for your use case but not for others. By receiving a generic component, we allow all types of header customization.

michael-s-molina:
> @michael-s-molina I've refactored the code such that it is only adding new components to the alerts&reports components where it we be implemented. Changes to the Collapse component have been reverted. @hughhhh In the refactor, I've removed the need for override the ts checks, so all of the `@ts-ignore` lines have been removed.

@fisjac I'm trying to better understand the context for this PR. Are you planning to change the default implementation of Alerts and Reports of Superset or override the default implementation of Superset in your installation? The reason I'm asking is because the components you created are not referenced anywhere.

If you are trying to override the default implementation of Superset in your installation, then these components don't belong to Superset codebase but to your codebase. What the Superset Collapse panel component needs to provide is a property that allow you to override the header component.

fisjac:
> > @michael-s-molina I've refactored the code such that it is only adding new components to the alerts&reports components where it we be implemented. Changes to the Collapse component have been reverted. @hughhhh In the refactor, I've removed the need for override the ts checks, so all of the `@ts-ignore` lines have been removed.
> 
> @fisjac I'm trying to better understand the context for this PR. Are you planning to change the default implementation of Alerts and Reports of Superset or override the default implementation of Superset in your installation? The reason I'm asking is because the components you created are not referenced anywhere.
> 
> If you are trying to override the default implementation of Superset in your installation, then these components don't belong to Superset codebase but to your codebase. What the Superset Collapse panel component needs to provide is a property that allow you to override the header component.

The context for this PR is in response to the broader design proposal linked here: https://github.com/apache/superset/discussions/25729 While these components are currently orphaned, the idea is to implement them within the redesign of the Alerts&Reports Modal, which is where I've placed the components within the codebase.

michael-s-molina:
> The context for this PR is in response to the broader design proposal linked here: https://github.com/apache/superset/discussions/25729 While these components are currently orphaned, the idea is to implement them within the redesign of the Alerts&Reports Modal, which is where I've placed the components within the codebase.

Oh, now I understood. There's still a long way to go regarding that discussion, which will become a SIP (Superset Improvement Proposal) and be submitted for voting before implementation begins. This means that a lot can change regarding the design. I'll add the `hold` label to this PR and revisit after the SIP is approved.

geido:
The work was moved in a feature branch and merged. Closing this one

PR #23853, opened 2023-04-27
chore: Migrate warm up cache endpoint to api v1

jfrag1:
### SUMMARY

This PR deprecates the `superset/warm_up_cache` endpoint and creates the `api/v1/charts/warm_up_cache` and `api/v1/dataset/warm_up_cache` endpoints to replace it.

The new endpoint has the following key differences:
* Follows command pattern
* uses `PUT` instead of `GET`
* accepts a json body validated by marshmallow instead of query params
* added openapi spec

Also improves test coverage

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #10397, opened 2020-07-22
fix(charts): disable CSRF for chart data endpoint

dpgaspar:
### SUMMARY
CSRF exempts by default the chart data REST API endpoint. This is a POST endpoint that is used like a GET endpoint. We use POST because the amount of options is so big we need a payload.

Also did a small refactor on the way CSRF is initialised, now we can import it and use it's decorator if needed

Related to issue #10354, will help user's leverage the chart data API

### ADDITIONAL INFORMATION
- [ ] Has associated issue: #10354 
- [ ] Changes UI
- [ ] Requires DB Migration.
- [ ] Confirm DB Migration upgrade and downgrade tested.
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25978, opened 2023-11-14
chore: exempt cache-warmup endpoint from CSRF protection

mapledan:
The default configuration for WTF_CSRF_ENABLED is 'True,' and this was causing the cache-warmup task to fail. 
The API request for the celery worker only includes session cookie.
Therefore, It should include superset.charts.api.warm_up_cache in the WTF_CSRF_EXEMPT_LIST.

### TESTING INSTRUCTIONS
1. Configure the cache-warmup task.
2. The Celery task successfully executed the fetch_url action.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -245,6 +245,7 @@ WTF_CSRF_ENABLED = True
 WTF_CSRF_EXEMPT_LIST = [
     "superset.views.core.log",
     "superset.views.core.explore_json",
+    "superset.charts.api.warm_up_cache",
     "superset.charts.data.api.data",
 ]
 


john-bodley:
@dpgaspar do you know why we have CSRF exemption in the first place? @michael-s-molina and myself are somewhat perplexed as to why a handful of API endpoints are exempt from requiring the CSRF token.

dpgaspar:
> @mapledan Although I understand why you would want to disable this in the Celery context, you need to consider that anyone can invoke the endpoint. @dpgaspar do we need to revisit this configuration?

By default only non HTTP GET endpoints are protected by CSRF, we have considered `superset.views.core.explore_json` and `superset.charts.data.api.data` to work as HTTP GET since no state is changed by these endpoints. 

`charts.data` was exempted to make it easier for users to use this API #10397 
`explore_json` probably required by embedded, yet it does work as an HTTP GET endpoint also #17530 

no sure about the log endpoint

mapledan:
Yes, the warm_up_cache endpoint is called by cache-warmup task. #23853 
It's clear to me now that superset.charts.api.warm_up_cache has changed the cache resource, and it does not work like a GET request.
I submitted another PR (#26003) that allows machine_auth to generate a CSRF token.
Possibly, this is the more correct approach.

dpgaspar (commented):
probably unnecessary because of: https://github.com/apache/superset/pull/26003

PR #25993, opened 2023-11-15
fix: Make Select component fire onChange listener when a selection is pasted in

### LINKED ISSUE #25601: Pasting a value into filters doesn't select the value
#### How to reproduce the bug
1. Copy a value eg: Classic Cars
2. Create or Open a chart
3. Click on filter
4. Choose a column, paste the copied value and click anywhere on the Filter popup and then click Save

### Expected results
The pasted value is applied as a filter

### Actual results
Value looks like it is added but the filter doesn't save with the value 

#### Screenshots

https://github.com/apache/superset/assets/10627051/2a8ddab6-951e-45cc-9c83-d89f7b69c850

### Environment

- browser type and version: `Firefox`
- superset version: `master`
- node.js version: `node -v`
- any feature flags active:

### Checklist

jfrag1:
### SUMMARY

Currently, pasting a value into a `Select` does not trigger the `onChange` callback passed to the `Select`.  This resulted in a buggy experience when pasting values into the adhoc filter select component, as demonstrated below.  It's possible similar issues exist in other areas of the application which this change will also fix.

Fixes https://github.com/apache/superset/issues/25601

### Before:
adhoc_paste_before

### After:
adhoc_paste_after

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/25601
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Select/Select.test.tsx
+++ b/superset-frontend/src/components/Select/Select.test.tsx
@@ -985,6 +985,20 @@ test('fires onChange when clearing the selection in multiple mode', async () => 
   expect(onChange).toHaveBeenCalledTimes(1);
 });
 
+test('fires onChange when pasting a selection', async () => {
+  const onChange = jest.fn();
+  render(<Select {...defaultProps} onChange={onChange} />);
+  await open();
+  const input = getElementByClassName('.ant-select-selection-search-input');
+  const paste = createEvent.paste(input, {
+    clipboardData: {
+      getData: () => OPTIONS[0].label,
+    },
+  });
+  fireEvent(input, paste);
+  expect(onChange).toHaveBeenCalledTimes(1);
+});
+
 test('does not duplicate options when using numeric values', async () => {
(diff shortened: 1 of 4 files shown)


superset-frontend/src/components/Select/Select.test.tsx
```diff
@@ -985,6 +985,20 @@ test('fires onChange when clearing the selection in multiple mode', async () =>
   expect(onChange).toHaveBeenCalledTimes(1);
 });
 
+test('fires onChange when pasting a selection', async () => {
```
jfrag1, line 988: This test fails without the fix

michael-s-molina:
Thanks for the fix @jfrag1. Could you also add the fix/test to the async version?

jfrag1:
Added

michael-s-molina (approved):
LGTM. Thank you for the fix!

PR #25995, opened 2023-11-15
chore: Optimize fetching samples logic

john-bodley:
### SUMMARY

At Airbnb (as mentioned previously) we have an uber wide dataset (circa thousands of columns) which is can be somewhat problematic for certain workflows. One such workflow is fetching sample data—which is used in explore and drill-to-detail. 

The existing logic constructs and executes a query for fetching both the sample data and `COUNT(*)` where the former is rather complex—relying on a very expensive SQLAlchemy query to fetch all the columns—before processing the results.

This PR introduces an optimization which first constructs and executes the `COUNT(*)` query (which is both cheap to construct and execute) and only proceeds to the sample data if the first query succeeds. Given that (in our formulation) the `COUNT(*)` aggregate is not supported the endpoint returns in a few seconds as opposed from a few minutes. The TL;DR is we should fail early if possible.

#### BEFORE

Screenshot 2023-11-15 at 2 16 57 PM

#### AFTER

Screenshot 2023-11-15 at 2 17 15 PM

### TESTING INSTRUCTIONS

Existing unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/datasource/utils.py
+++ b/superset/views/datasource/utils.py
@@ -104,21 +104,18 @@ def get_samples(  # pylint: disable=too-many-arguments,too-many-locals
         result_type=ChartDataResultType.FULL,
         force=force,
     )
-    samples_results = samples_instance.get_payload()
-    count_star_results = count_star_instance.get_payload()
 
     try:
-        sample_data = samples_results["queries"][0]
-        count_star_data = count_star_results["queries"][0]
-        failed_status = (
-            sample_data.get("status") == QueryStatus.FAILED
-            or count_star_data.get("status") == QueryStatus.FAILED
-        )
-        error_msg = sample_data.get("error") or count_star_data.get("error")
-        if failed_status and error_msg:
-            cache_key = sample_data.get("cache_key")
-            QueryCacheManager.delete(cache_key, region=CacheRegion.DATA)
-            raise DatasetSamplesFailedError(error_msg)
+        count_star_data = count_star_instance.get_payload()["queries"][0]
+
+        if count_star_data.get("status") == QueryStatus.FAILED:
+            raise DatasetSamplesFailedError(count_star_data.get("error"))
+
+        sample_data = samples_instance.get_payload()["queries"][0]
+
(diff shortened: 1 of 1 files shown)


superset/views/datasource/utils.py
```diff
@@ -104,21 +104,18 @@ def get_samples(  # pylint: disable=too-many-arguments,too-many-locals
...
-            raise DatasetSamplesFailedError(error_msg)
+        count_star_data = count_star_instance.get_payload()["queries"][0]
+
+        if count_star_data.get("status") == QueryStatus.FAILED:
+            raise DatasetSamplesFailedError(count_star_data.get("error"))
+
+        sample_data = samples_instance.get_payload()["queries"][0]
+
+        if sample_data.get("status") == QueryStatus.FAILED:
+            QueryCacheManager.delete(sample_data.get("cache_key"), CacheRegion.DATA)
```
john-bodley, line 117: I'm not entirely sure why we only delete the sample data cache key, however I thought it was best to preserve the existing logic.

justinpark (approved):
LGTM

superset/views/datasource/utils.py
zhaoyongjie: The "count star" might be a heavy workload in some db, it trigger full table scan and aggregation, so just removed "sample_data" but keep "count star", it means that even though "sample_data" query is failed, count_star_data is cached still.

I think the new logic should remove cache key if the "count star" query is failed.

zhaoyongjie: and...another think is that we should customize count(*) in different db.

john-bodley: Thanks @zhaoyongjie for the comment. Thinking about this more, if a query fails then there’s shouldn’t be anything cached. Hence, in the new sequential formulation, if the `COUNT(*)` query succeeds but the sample data query fails we should actually be removing the cached result from the former.

PR #25996, opened 2023-11-15
chore(tags): Allow for lookup via ids vs. name in the API

hughhhh:
### SUMMARY

Previously we the API only supported queries using the names as way of querying tags, I've updated the API to now accept IDs (better pattern) to allow names to have commas. We discovered the issue in testing and took advantage the opportunity to refactor this piece of code, but also leaving `tags` param for backwards compatibility.

```
curl 'http://127.0.0.1:54647/api/v1/tag/get_objects/?tagIds=17,12,13' \
-X 'GET' \
-H 'Accept: application/json' \
# ...
```

### TESTING INSTRUCTIONS

1. goto all entities page (http://127.0.0.1:8080/superset/all_entities)
2. see page render with new network call

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/tags/tags.ts
+++ b/superset-frontend/src/features/tags/tags.ts
@@ -194,3 +194,20 @@ export function fetchObjects(
     .then(({ json }) => callback(json.result))
     .catch(response => error(response));
 }
+
+export function fetchObjectsByTagIds(
+  {
+    tagIds = [],
+    types,
+  }: { tagIds: number[] | undefined; types: string | null },
+  callback: (json: JsonObject) => void,
+  error: (response: Response) => void,
+) {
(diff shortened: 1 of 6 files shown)


lilykuang (approved):

superset-frontend/src/features/tags/tags.ts
```diff
@@ -194,3 +194,20 @@ export function fetchObjects(
     .then(({ json }) => callback(json.result))
     .catch(response => error(response));
 }
+
+export function fetchObjectsByTagIds(
```
rusackas, line 198: is `fetchObjects` above this just dead code now? I don't see it being used elsewhere.

sadpandajoe:
🏷️ preset:2023.47

PR #26003, opened 2023-11-16
chore: allow machine auth to generate CSRF token

mapledan:
The celery task get permissions through machine_auth.
The 'cache-warmup' task needs to call the '/api/v1/chart/warm_up_cache' endpoint .However, this endpoint is non GET requests and is therefore protected by CSRF.
Hence, when machine_auth retrieves the auth cookies, it also needs to generate a CSRF token.
This ensures that the Celery task can operate successfully.

### TESTING INSTRUCTIONS
1. Configure the cache-warmup task.
2. The Celery task successfully executed the fetch_url action.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/utils/machine_auth.py
```diff
@@ -143,6 +144,31 @@ def get_auth_cookies(user: User) -> dict[str, str]:
 
         return cookies
 
+    @staticmethod
+    def get_auth_cookie_and_csrf_token(user: User) -> tuple[dict[str, str], str]:
```
dpgaspar, line 148: We could check for `app.config["WTF_CSRF_ENABLED"]` here and DRY the code a bit more by removing `get_auth_cookies`

mapledan:
It's related to [#28706](https://github.com/apache/superset/pull/28706)

PR #24776, opened 2023-07-22
chore(sqlalchemy): Remove erroneous SQLAlchemy ORM session.merge operations

john-bodley:
### SUMMARY

This PR addresses a non-controversial observation from [[SIP-99A] Primer on managing SQLAlchemy sessions](https://github.com/apache/superset/issues/25107) where we have a tendency to unnecessarily re-merge objects into the SQLAlchemy session which are already present in the seession.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue: https://github.com/apache/superset/issues/25107
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

Issue #25108, opened 2023-08-28
[SIP-98B] Proposal for (re)defining a “unit of work”


john-bodley:
## [SIP-98B] Proposal for (re)defining a “unit of work”

This SIP is part of the [[SIP-98] Proposal for correctly handling business logic](https://github.com/apache/superset/issues/25048) series. Specifically, it proposes formalizing the “unit of work” construct. The topics outlined here are highly coupled with [[SIP-98A] Primer for managing SQLAlchemy sessions](https://github.com/apache/superset/issues/25107) as they both relate to managing database transactions.

### Unit of Work

The “unit of work” is used to group multiple service operations into a single atomic unit. It materializes as a database transaction which—with the support of a SQLAlchemy session—allows us to create a block of code within which database atomicity is guaranteed. If the block of code is successfully completed, the changes are committed to the database. If there is an exception, the changes are rolled back.

SQLAlchemy also supports this [feature](https://www.sqlalchemy.org/features.html):

> The Unit Of Work system, a central part of SQLAlchemy's Object Relational Mapper (ORM), organizes pending `INSERT`/`UPDATE`/`DELETE` operations into queues and flushes them all in one batch.

Historically, Superset’s “unit of work” has been ill-defined, mismanaged,  and/or misconstrued. It has resulted in us over (partial) committing*—a somewhat typical SQLAlchemy mistake (see [SIP-?A](...) for more details)—which violates the atomicity of the operation and leads to unnecessary complexity.

\* The atomic unit vagueness has led to code inconsistencies which have resulted in us often adopting the “when in doubt, commit” mentality.

[[SIP-35] Proposal for Improving Superset’s Python Code Organization](https://github.com/apache/superset/issues/9077) introduced the concept of a Command and Data Access Object (DAO) which (clearly and rightfully) stated,

> Commands perform a single cohesive business operation, have a very small public API, and either entirely succeed or entirely fail. In practice, this requires that database transaction behavior be controlled at the Command level. When possible, commands should be modeled such that they perform a single database transaction. In the case of failure, they raise an exception or return an error code, and it is the responsibility of the caller to translate that into a correct response for a user.

however, sadly the code examples (where the database logic is defined entirely within the DAO which—via the commit and rollback operations—ends the transaction) likely led people astray—violating the “unit of work” concept.

#### Examples

The following examples illustrate where the atomic business operation has been violated:

[...]

PR #26009, opened 2023-11-16
chore: Remove unnecessary autoflush from tagging and key/value workflows

john-bodley:
### SUMMARY

This PR addresses this [comment](https://github.com/apache/superset/pull/24776#discussion_r1388352188) associated with https://github.com/apache/superset/pull/24776. After speaking with @michael-s-molina and @villebro it seems like the `autoflush` logic was likely not necessary.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/tags/models.py
+++ b/superset/tags/models.py
@@ -20,9 +20,9 @@ import enum
 from typing import TYPE_CHECKING
 
 from flask_appbuilder import Model
-from sqlalchemy import Column, Enum, ForeignKey, Integer, String, Table, Text
+from sqlalchemy import Column, Enum, ForeignKey, Integer, orm, String, Table, Text
 from sqlalchemy.engine.base import Connection
-from sqlalchemy.orm import relationship, Session, sessionmaker
+from sqlalchemy.orm import relationship, sessionmaker
 from sqlalchemy.orm.mapper import Mapper
 
 from superset import security_manager
@@ -35,7 +35,7 @@ if TYPE_CHECKING:
     from superset.models.slice import Slice
     from superset.models.sql_lab import Query
 
-Session = sessionmaker(autoflush=False)
+Session = sessionmaker()
 
 user_favorite_tag_table = Table(
     "user_favorite_tag",
@@ -111,7 +111,7 @@ class TaggedObject(Model, AuditMixinNullable):
     tag = relationship("Tag", back_populates="objects", overlaps="tags")
 
 
-def get_tag(name: str, session: Session, type_: TagType) -> Tag:
+def get_tag(name: str, session: orm.Session, type_: TagType) -> Tag:
     tag_name = name.strip()
     tag = session.query(Tag).filter_by(name=tag_name, type=type_).one_or_none()
     if tag is None:
@@ -148,7 +148,7 @@ class ObjectUpdater:
     @classmethod
     def _add_owners(
         cls,
-        session: Session,
+        session: orm.Session,
         target: Dashboard | FavStar | Slice | Query | SqlaTable,
     ) -> None:
         for owner_id in cls.get_owners_ids(target):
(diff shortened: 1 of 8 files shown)


superset/tags/models.py
```diff
@@ -166,9 +166,7 @@ def after_insert(
         connection: Connection,
         target: Dashboard | FavStar | Slice | Query | SqlaTable,
     ) -> None:
-        session = Session(bind=connection)
-
-        try:
+        with Session(bind=connection) as session:
```
john-bodley, line 169: Opting for a context manager which closes the session on exit.

superset/tags/models.py
```diff
@@ -111,7 +111,7 @@ class TaggedObject(Model, AuditMixinNullable):
     tag = relationship("Tag", back_populates="objects", overlaps="tags")
 
 
-def get_tag(name: str, session: Session, type_: TagType) -> Tag:
+def get_tag(name: str, session: orm.Session, type_: TagType) -> Tag:
```
john-bodley, line 114: We redefine the `Session` class on line 38 so the previous type hint wasn't completely accurate.

villebro:
@john-bodley so those closed sessions in the event handlers were caused by those explicit `finally: session.close()` calls?

john-bodley:
@villebro I _think_ the error you might have been seeing, per what I was witnessing earlier when trying the `session = inspect(target).session` approach, was actually, 

> This transaction is closed

i.e., it was transaction as opposed to session related. The reason being (as mentioned in Slack) was that in the SQLAlchemy [event documentation](https://docs.sqlalchemy.org/en/20/orm/events.html#mapper-events) it states:

> Mapper-level flush events only allow **very limited operations**, on attributes local to the row being operated upon only, as well as allowing any SQL to be emitted on the given `Connection`.

So when trying to leverage the Flask-SQLAlchemy  session—via either `inspect(target).session` or `db.session`—the associated transaction was closed because said code was trying to perform operations outside of operations limited to the row (`target`) being operated on, i.e., the `after_insert` event listener is associated with a chart and we then try to create an associated tag. Creating a new session from the connection results in a new transaction for us to emit any SQL operation (per the SQLAlchemy statement).

villebro (approved):
Very nice 👍 So I assume one of the next steps is to start moving those event listener based mutations into the parent commands?

villebro:
> @villebro I _think_ the error you might have been seeing, per what I was witnessing earlier when trying the `session = inspect(target).session` approach, was actually,
> 
> > This transaction is closed
> 
> i.e., it was transaction as opposed to session related. The reason being (as mentioned in Slack) was that in the SQLAlchemy [event documentation](https://docs.sqlalchemy.org/en/20/orm/events.html#mapper-events) it states:
> 
> > Mapper-level flush events only allow **very limited operations**, on attributes local to the row being operated upon only, as well as allowing any SQL to be emitted on the given `Connection`.
> 
> So when trying to leverage the Flask-SQLAlchemy session—via either `inspect(target).session` or `db.session`—the associated transaction was closed because said code was trying to perform operations outside of operations limited to the row (`target`) being operated on, i.e., the `after_insert` event listener is associated with a chart and we then try to create an associated tag. Creating a new session from the connection results in a new transaction for us to emit any SQL operation (per the SQLAlchemy statement).
> 
> Note in [[SIP-99B] Proposal for (re)defining a "unit of work"](https://github.com/apache/superset/issues/25108) this type of operation will not be allowed because it violates the "unit of work".

@john-bodley got it, this is all finally starting to make sense. Thanks for taking this on - once finished, this will have a big positive impact on the backend code quality, performance, maintainability etc..

john-bodley:
@villebro in terms of next steps, today's discussion (in conjunction with this PR) helped to solidify what is viable with SQLAlchemy event handlers and thus I updated SIP-99B with regards to how this should be handled moving forward. The TL;DR is ideally we should move said logic either to the database (if said events previously were handling logic which could be captured via a `CASCADE` constraint) or Command.

Issue #25728, opened 2023-10-20
Translation: missing terms in french


cchristofr:
Some terms are not translated in the UI

#### How to reproduce the bug

1. Open a dashboard
2. Click on ADD/EDIT Filters
3. Add a Time range filter
4. Filter has default value  : Choose Last
5. Choose 'Last quarter'
6. Save

### Expected results

The value of the filter should be "Le trimestre dernier" 

### Actual results

But it is "Last quarter"

The same for each period: "Last day" to ... "Last year" are not translated as value for the filter.
#### Screenshots

![bug_french](https://github.com/apache/superset/assets/74195127/097f0b35-b667-483b-8172-897b5640e6cf)

### Environment

(please complete the following information):

- browser type and version:Firefox 118.0.2
- superset version: 2.1.0
- python version: 3.8.16
- node.js version: `node -v`
- any feature flags active:

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [ ] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [ ] I have reproduced the issue with at least the latest released version of superset.
- [ ] I have checked the issue tracker for the same issue and I haven't found one similar.

sfirke:
This was fixed in https://github.com/apache/superset/pull/20078 🚀  Looks like that PR made it into 3.0.0.  If you're able to upgrade to that version or 3.0.1 (the latest), you should see it.  Please reply if that is not the case.

aehanno:
HI @sfirke, I upgrade my version in 3.0.1 and when I choose French, I got all those string in English
Before my upgrade I got that
![image](https://github.com/apache/superset/assets/104373599/1c5cb503-0176-45d7-970d-d3886eec56a1)
and now in 3.0.1, I have that
![image](https://github.com/apache/superset/assets/104373599/22a4a548-05a7-49e9-89d1-0bf0fed0cfb9)

sfirke:
Hmmm. To confirm, you're showing the translation coverage got *worse* when you upgraded to 3.0.0/3.0.1 ?

aehanno:
I was in 2.0.1

aehanno:
@sfirke and I just saw that there is custom and last that are not translate
But the other filter are correct
![image](https://github.com/apache/superset/assets/104373599/5be7aad9-debe-4793-91d8-fb084a3b94ce)

sfirke:
[...]

PR #26010, opened 2023-11-16
fix: Remove annotation Fuzzy to get french translation

aehanno:
Fix this pb : https://github.com/apache/superset/issues/25728

### DIFF
--- a/superset/translations/fr/LC_MESSAGES/messages.json
+++ b/superset/translations/fr/LC_MESSAGES/messages.json
@@ -2959,7 +2959,7 @@
       "Select values in highlighted field(s) in the control panel. Then run the query by clicking on the %s button.": [
         ""
       ],
-      "Send as CSV": ["Envoyer comme CSV"],
+      "Send as CSV": ["Envoyer au format CSV"],
       "Send as PNG": ["Envoyer comme PNG"],
       "Send as text": ["Envoyer comme texte"],
       "Send range filter events to other charts": [""],
--- a/superset/translations/fr/LC_MESSAGES/messages.json
+++ b/superset/translations/fr/LC_MESSAGES/messages.json
@@ -226,6 +226,7 @@
       "Add calculated temporal columns to dataset in \"Edit datasource\" modal": [
         ""
       ],
+      "Add cross-filter": ["Ajouter un filtre"],
       "Add custom scoping": [""],
       "Add delivery method": ["Ajouter méthode de livraison"],
       "Add filter": ["Ajouter un filtre"],
@@ -556,6 +557,8 @@
         ""
       ],
       "Append": ["Ajouter"],
+      "Applied filters (%d)": ["Filtres appliqués (%d)"],
+      "Applied filters: %s": ["Filtres appliqué: %s"],
       "Applied rolling window did not return any data. Please make sure the source query satisfies the minimum periods defined in the rolling window.": [
         "La fenêtre glissante appliquée n'a pas retourné de données. Assurez-vous que la requête source satisfasse les périodes minimum définies dans la fenêtre glissante."
       ],
@@ -1089,7 +1092,7 @@
       ],
       "Creator": ["Créateur"],
       "Cross-filter will be applied to all of the charts that use this dataset.": [
-        ""
+        "Le filtre va être appliqué à tous les graphiques qui utilise cet ensemble de données"
       ],
       "Currently rendered: %s": [""],
       "Custom": ["Personnalisée"],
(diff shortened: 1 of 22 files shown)


sfirke:
It looks like all of the other languages have lower case ([example](https://github.com/apache/superset/blob/bece2ea3e4b9979f3f45f63aa490f499095c7078/superset/translations/sl/LC_MESSAGES/messages.json#L5415)), which matches the base set for translation.  French has uppercase and it's not matching.  Do you agree the French should be changed to lowercase?  I worry that this PR would fix French and break all of the other languages.

sfirke (changes requested):
See my comment in the PR conversation, I think the proper fix would be to make the French translation values lower case, not the base ones capitalized.

aehanno:
Hey @sfirke, yes I just saw that, I saw thaht the problem is the annatation #, fuzzy in po file
Do you know why we have that and is it's correct if I just remove this annotation ?

sfirke:
I'm unfamiliar with translations in general.  Sounds like the [fuzzy there means, attention needed](https://stackoverflow.com/questions/15653093/what-does-fuzzy-mean-at-the-top-of-a-xgettext-generated-pot-file).  What problem is the fuzzy comment causing?

sfirke:
Also thanks for updating the French strings to match the base set, it looks good now.

sfirke:
Ah okay.  I don't see "fuzzy" appearing in the codebase anywhere so I'd say go ahead and delete that.  If you can test locally that all the desired translations are appearing, that works for me.

superset/translations/fr/LC_MESSAGES/messages.json
```diff
@@ -4591,18 +4618,10 @@
       "y: values are normalized within each row": [""],
       "year": ["année"],
       "zoom area": [""],
-      "No matching records found": ["Aucun résultat trouvé"],
```
sfirke, line 4594: Did you replace this string?  I see it deleted here but no other mention of "no matching records found"

sfirke (approved):

superset/translations/fr/LC_MESSAGES/messages.json
aehanno: Just annotation fuzzy on this string, I remove it

sfirke (approved):
LGTM

sfirke:
This is my first time working with translations - anyone who has more experience want to review + merge?

aehanno:
@sfirke, Hey thanks for the review.
Do you know if it can be merge in the 3.0.2 release
It will be so great

superset/translations/fr/LC_MESSAGES/messages.json
```diff
@@ -1635,7 +1643,10 @@
       "Export": ["Exporter"],
       "Export dashboards?": ["Exporter les tableaux de bords ?"],
       "Export query": ["Exporter la requête"],
-      "Export to YAML": ["Exporter en YAML"],
+      "Export to .CSV": ["Exporter au format CSV"],
+      "Export to .JSON": ["Exporter au format YAML"],
```
sfirke, line 1647: Yikes I missed this one - it should say JSON

superset/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -7312,24 +7308,21 @@ msgstr "Exporter la requête"
...
-#, fuzzy
 msgid "Export to .CSV"
-msgstr "Exporter en YAML"
+msgstr "Exporter au format CSV"
 
 #: superset-frontend/src/explore/components/useExploreAdditionalActionsMenu/index.jsx:323
-#, fuzzy
 msgid "Export to .JSON"
-msgstr "Exporter en YAML"
+msgstr "Exporter au format YAML"
```
sfirke, line 7316: Should this be JSON?

superset/translations/fr/LC_MESSAGES/messages.json
sfirke: I don't follow, sorry.  Will the string "No matching records found" still be translated?

sfirke (changes requested):
Did you mean to translate "JSON" to "YAML"?

superset/translations/fr/LC_MESSAGES/messages.json
aehanno: Oh yes sure, my bad

aehanno:
for the quesiton : https://github.com/apache/superset/pull/26010#discussion_r1399541627

The translation is present line 2254

sfirke (approved):
Merci pour les améliorations! (Looks good to me 😜)

rusackas (approved):
LGTM!

PR #26022, opened 2023-11-19
feat: Time Filter shortcuts

zhaoyongjie:
### SUMMARY
The Time Filter currently supports DSL for defining arbitrary time expressions, such as `date_trunc`, `date_add`, `holidays`, and more. However, users without a technical background often find it challenging to use. This pull request proposes the introduction of a new feature in the Time Filter called 'Shortcut.' 

In this feature, we can easily define custom time ranges in `config.py`, and the arbitrary time expressions will dynamically appear in it.

The pre-defined shortcuts are as following, of course, we can change them if you want
```
TIME_FILTER_SHORTCUTS = [
    ["Current week", "DATETRUNC(DATETIME('today'), WEEK) : today"],
    ["Current month(MTD)", "DATETRUNC(DATETIME('today'), MONTH) : today"],
    ["Current year(YTD)", "DATETRUNC(DATETIME('today'), YEAR) : today"],
]
```

## Shortcuts in Frame
image

## Shortcuts

image

## Time filter in Native Filter
image

### TESTING INSTRUCTIONS

CI and manually

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


zhaoyongjie:
> Small nitpick: Before the "(YTD)" and "(MTD)" part, there should be a space.

Good catch! will update codes and screenshots.

john-bodley:
@zhaoyongjie weren't there some Cartel designs for more human-esque time ranges? Additionally I'm not entirely sure about the design/UX, i.e., it's not evident what represents a "shortcut" as opposed to any of the existing filters—many of which are shortcuts by themselves.

michael-s-molina:
Thank you for the PR @zhaoyongjie! Given that the shortcuts don't need extra input from the user, do you think we could display them as normal select options instead of a separate section?

Using the screenshot you provided as example, we could have:
```
Current week
Current month (MTD)
Current year (YTD)
Last
Previous
...
```
If some of these become really popular, we can even make them official in the future and nothing would change for the users.

zhaoyongjie:
> Thank you for the PR @zhaoyongjie! Given that the shortcuts don't need extra input from the user, do you think we could display them as normal select options instead of a separate section?
> 
> Using the screenshot you provided as example, we could have:
> 
> ```
> Current week
> Current month (MTD)
> Current year (YTD)
> Last
> Previous
> ...
> ```
> 
> If some of these become really popular, we can even make them official in the future and nothing would change for the users.

If so, should we extract all values in `Last` and `Previous`? 

```
Current week
Current month (MTD)
Current year (YTD)
Last day
Last week
Last month
...
Previous week
...
```

zhaoyongjie:
> @zhaoyongjie weren't there some Cartel designs for more human-esque time ranges? Additionally I'm not entirely sure about the design/UX, i.e., it's not evident what represents a "shortcut" as opposed to any of the existing filters—many of which are shortcuts by themselves.

The goal of this PR is providing a flexible approach that should define **any** time range from user config. it's not only for "Current" or "Cartel design".

michael-s-molina:
> If so, should we extract all values in Last and Previous?

I think this would be a good idea. If we can reduce the number of steps users need to find their desired option, it would be great. Especially because the Select allows searching so they could type "Last" and see official e shortcut options all at once. I would only present the extra panel when more input is needed from the users. 

@kasiazjc @rusackas might be interested in the discussion too. 

By the way, I think simplifying the user experience is really valuable so I like the intent of this PR @zhaoyongjie.

zhaoyongjie:
> > If so, should we extract all values in Last and Previous?
> 
> I think this would be a good idea. If we can reduce the number of steps users need to find their desired option, it would be great. Especially because the Select allows searching so they could type "Last" and see official e shortcut options all at once. I would only present the extra panel when more input is needed from the users.
> 
> @kasiazjc @rusackas might be interested in the discussion too.
> 
> By the way, I think simplifying the user experience is really valuable so I like the intent of this PR @zhaoyongjie.

I totally agree with you. The "Last", "Previous", "Current" and more customized item might be provide by `config.py` if user want to change order or to remove some item, it should be easy.

superset/config.py
```diff
@@ -153,6 +153,11 @@ def _try_json_readsha(filepath: str, length: int) -> str | None:
 # default time filter in explore
 # values may be "Last day", "Last week", "<ISO date> : now", etc.
 DEFAULT_TIME_FILTER = NO_TIME_RANGE
+TIME_FILTER_SHORTCUTS = [
+    ["Current week", "DATETRUNC(DATETIME('today'), WEEK) : today"],
+    ["Current month (MTD)", "DATETRUNC(DATETIME('today'), MONTH) : today"],
+    ["Current year (YTD)", "DATETRUNC(DATETIME('today'), YEAR) : today"],
+]
```
villebro, line 160: A minor thing, but would it make sense to define these as a map instead of an array? I will review/test in more detail later in the week.

zhaoyongjie: Thanks for the review Ville! The reseaon of using an array is to create an ordered list for the 'shortcut', e.g. week before month

rusackas:
My two cents here: If we are considering a redesign, then we should make the smallest possible UI change here to support the use case for the moment. 

I fear that pulling out all the day/month/etc ranges into the Select menu is... not ideal. I would just copy the pattern for "Last" and make a new option for "Current" with similar radio button options:

Screenshot_2023-11-20_at_1_34_33 PM

The customizable options via config.py are still interesting/useful, but this seems like a common use case that we ought to support as a mainstream feature. I'm also not sure what the maintenance/documentation requirements would be if people start using these custom config options frequently in installations around the world.

alesconti:
I have a Superset installation configured via Docker-compose. How can I activate this feature? @zhaoyongjie

zhaoyongjie:
The feature hasn't merged, you should cherry-pick this PR to your own codebase and build it by youself.

PR #26023, opened 2023-11-20
fix: Prevent cached bootstrap data from leaking between users w/ same first/last name

jfrag1:
### SUMMARY

The caching mechanism for bootstrap data was not working as intended in the case where two users share the same first/last name.  The cache key was dependent on the `User` object passed. When a non-primitive object is used to generate a cache key, `flask-caching` calls `repr()` on it to determine the key: https://flask-caching.readthedocs.io/en/latest/index.html#memoization.

The `User` class comes from flask appbuilder and has the following `__repr__`:
```python
def get_full_name(self):
    return "{0} {1}".format(self.first_name, self.last_name)

def __repr__(self):
    return self.get_full_name()
```
Therefore, when two users shared a first and last name, they would generate the same cache key and the cache could leak between the two users.  This was noticed because it would result in affected users seeing menu results they shouldn't see, though it potentially had other subtle effects as well, particularly for anyone adding a user-dependent `COMMON_BOOTSTRAP_OVERRIDES_FUNC`.

This PR fixes the issue by passing the actual user id to the memoized function so that it's included in the cache key.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/base.py
+++ b/superset/views/base.py
@@ -380,7 +380,9 @@ def menu_data(user: User) -> dict[str, Any]:
 
 
 @cache_manager.cache.memoize(timeout=60)
-def cached_common_bootstrap_data(user: User, locale: str) -> dict[str, Any]:
+def cached_common_bootstrap_data(
+    user: User, user_id: int | None, locale: str
+) -> dict[str, Any]:
     """Common data always sent to the client
 
     The function is memoized as the return value only changes when user permissions
@@ -424,8 +426,9 @@ def cached_common_bootstrap_data(user: User, locale: str) -> dict[str, Any]:
 
 
 def common_bootstrap_payload(user: User) -> dict[str, Any]:
+    user_id = user.id if hasattr(user, "id") else None
     return {
-        **cached_common_bootstrap_data(user, get_locale()),
+        **cached_common_bootstrap_data(user, user_id, get_locale()),
         "flash_messages": get_flashed_messages(with_categories=True),
     }
 
--- a/superset/views/base.py
+++ b/superset/views/base.py
@@ -380,7 +380,7 @@ def menu_data(user: User) -> dict[str, Any]:
 
 
 @cache_manager.cache.memoize(timeout=60)
-def cached_common_bootstrap_data(
+def cached_common_bootstrap_data(  # pylint: disable=unused-argument
     user: User, user_id: int | None, locale: str
 ) -> dict[str, Any]:
     """Common data always sent to the client
--- a/superset/views/base.py
+++ b/superset/views/base.py
@@ -296,7 +296,7 @@ class BaseSupersetView(BaseView):
     ) -> FlaskResponse:
         payload = {
             "user": bootstrap_user_data(g.user, include_perms=True),
-            "common": common_bootstrap_payload(g.user),
+            "common": common_bootstrap_payload(),
             **(extra_bootstrap_data or {}),
         }
         return self.render_template(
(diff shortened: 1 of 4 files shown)


superset/views/base.py
```diff
@@ -380,7 +380,9 @@ def menu_data(user: User) -> dict[str, Any]:
 
 
 @cache_manager.cache.memoize(timeout=60)
-def cached_common_bootstrap_data(user: User, locale: str) -> dict[str, Any]:
+def cached_common_bootstrap_data(  # pylint: disable=unused-argument
+    user: User, user_id: int | None, locale: str
```
john-bodley, line 384: Why do we need both `User` and `user_id`? Surely the former contains the ID which should then be used for the cache key given it's globally unique.

jfrag1: While yes, `User` "contains" the ID, it's `__repr__` method that flask-caching uses to generate the cache key does not (see PR description).

The 2 alternatives to this fix as-is would be to:
1. change the `__repr__` upstream on FAB to include the id. (cc @dpgaspar if this is a welcome change)
2. Subclass FAB's user class and add our own __repr__ then update to use that user class everywhere, which seems like a pretty large change for this small fix

john-bodley: @jfrag1 sorry I missed that detail when initially glancing over the PR description. Is a third alternative to just pass in `user_id` instead of both `user` and `user_id`?

superset/views/base.py
```diff
@@ -424,8 +426,9 @@ def cached_common_bootstrap_data(user: User, locale: str) -> dict[str, Any]:
 
 
 def common_bootstrap_payload(user: User) -> dict[str, Any]:
+    user_id = user.id if hasattr(user, "id") else None
```
john-bodley, line 429: It seems like the FAB [`User`](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/sqla/models.py#L94-L167) class will always have the `id` attribute defined and thus simply passing in `user.id` on line 431 _should_ be safe.

john-bodley: Grokking through the [code](https://github.com/search?q=repo%3Aapache%2Fsuperset%20common_bootstrap_payload&type=code) it seems like this is always `g.user`, thus rather than passing around the global, it seems clearer to remove the `user` variable from said method and thus said logic simply becomes, 

```python
 from superset.utils.core import get_user_id 

def common_bootstrap_payload() -> dict[str, Any]:
    return {
        **cached_common_bootstrap_data(get_user_id(), get_locale()),
        "flash_messages": get_flashed_messages(with_categories=True),
    }
```

jfrag1: Oh great point it is always `g.user`, I'll make that change.

> It seems like the FAB [`User`](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/sqla/models.py#L94-L167) class will always have the `id` attribute defined and thus simply passing in `user.id` on line 431 _should_ be safe.

The type hint was a bit of a lie since `g.user` could also be `AnonymousUserMixin`.  I'd update it but it's getting removed instead

john-bodley: @jfrag1 I think my comment is somewhat of a moot point given that it seems like we're going down the `get_user_id()` route.

john-bodley (approved):
LGTM

sadpandajoe:
🏷️ preset:2023.47

PR #26035, opened 2023-11-20
feat(dashboard): Hide chart header controls in reports

kgabryje:
### SUMMARY
If report is running, hide the slice header controls (filters badge, 3 dots options menu etc.).
To determine if report is running, we check if `standalone` query param is equal to `DashboardStandaloneMode.REPORT`

Before:
image

After:
image

### TESTING INSTRUCTIONS
1. Open a dashboard - chart header controls (filters badge, 3 dots for menu etc.) should be visible
2. Run a report -chart header controls should not be visible

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/SliceHeader/SliceHeader.test.tsx
+++ b/superset-frontend/src/dashboard/components/SliceHeader/SliceHeader.test.tsx
@@ -486,10 +486,10 @@ test('Add extension to SliceHeader', () => {
   expect(screen.getByText('This is an extension')).toBeInTheDocument();
 });
 
-test('Do not display slice header controls if user is a bot', () => {
-  Object.defineProperty(window.navigator, 'webdriver', {
+test('Do not display slice header controls in reports', () => {
+  Object.defineProperty(window, 'location', {
     get() {
-      return true;
+      return { search: 'standalone=3' };
     },
   });
   const props = createProps();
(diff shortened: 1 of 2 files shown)


john-bodley:
@kgabryje I think having the filter indicators et al. in the render report or alert is actually helpful to the reader as it indicates which charts are impacted by the filters. Without them the results could more easily* be misinterpreted.

\* Note even currently these can still be misinterpreted given it's not overly evident (without hovering) which filters are impacting said chart.

kgabryje:
> @kgabryje I think having the filter indicators et al. in the render report or alert is actually helpful to the reader as it indicates which charts are impacted by the filters. Without them the results could more easily* be misinterpreted.
> 
> * Note even currently these can still be misinterpreted given it's not overly evident (without hovering) which filters are impacting said chart.

Yeah that was our thinking that the filters badge is not very helpful as you can't even see the filters bar in the reports.
It's a fair point though that it might be useful to know if a chart is impacted by any filter at all.
@yousoph do you have an opinion on this?

yousoph:
Yeah, part of the thought process was also to not make the filters indicator look clickable in the screenshot where they aren't actually clickable. 

Perhaps separately we need a clearer way to communicate filters information to the users in reports, that could also better prompt to open the dashboard link to see full filters details?

john-bodley:
@kgabryje do you think—given that this feels like a UX regression—that we close this current PR and collectively try to formulate a solution (if possible) which better indicates which filters are being applied in the context of reporting?

PR #26038, opened 2023-11-20
docs(drivers): refresh guide on adding a db driver in docker

sfirke:
### SUMMARY
I went in to fix a couple of broken links reported in BugHerd.  While there, I refreshed the whole page, since it had been a while.  I didn't remove much, just extraneous text.  I did add a bit about debugging, including `docker exec -it`.

### ADDITIONAL INFORMATION
Fixes BugHerd 110, noting that the files linked to (requirements.txt and requirements-dev.txt) no longer exist.

### DIFF
--- a/docs/docs/databases/docker-add-drivers.mdx
+++ b/docs/docs/databases/docker-add-drivers.mdx
@@ -7,30 +7,21 @@ version: 1
 
 ## Adding New Database Drivers in Docker
 
-Superset requires a Python database driver to be installed for each additional type of database you
-want to connect to.
+Superset requires a Python database driver to be installed for each additional type of database you want to connect to.
 
-In this example, we'll walk through how to install the MySQL connector library. The connector
-library installation process is the same for all additional libraries.
+In this example, we'll walk through how to install the MySQL connector library. The connector library installation process is the same for all additional libraries.
 
 ### 1. Determine the driver you need
 
-Consult the [list of database drivers](/docs/databases/installing-database-drivers) and find the PyPI
-package needed to connect to your database. In this example, we're connecting to a MySQL database,
-so we'll need the `mysqlclient` connector library.
+Consult the [list of database drivers](/docs/databases/installing-database-drivers) and find the PyPI package needed to connect to your database. In this example, we're connecting to a MySQL database, so we'll need the `mysqlclient` connector library.
 
 ### 2. Install the driver in the container
 
-We need to get the `mysqlclient` library installed into the Superset docker container (it doesn't matter if it's
-installed on the host machine). We could enter the running container with `docker exec -it <container_name> bash`
-and run `pip install mysqlclient` there, but that wouldn't persist permanently.
+We need to get the `mysqlclient` library installed into the Superset docker container (it doesn't matter if it's installed on the host machine). We could enter the running container with `docker exec -it <container_name> bash` and run `pip install mysqlclient` there, but that wouldn't persist permanently.
 
-To address this, the Superset `docker compose` deployment uses the convention of a `requirements-local.txt` file. All packages
-listed in this file will be installed into the container from PyPI at runtime. This file will be ignored by Git for
-the purposes of local development.
+To address this, the Superset `docker compose` deployment uses the convention of a `requirements-local.txt` file. All packages listed in this file will be installed into the container from PyPI at runtime. This file will be ignored by Git for the purposes of local development.
 
-Create the file `requirements-local.txt` in a subdirectory called `docker` that exists in the directory with your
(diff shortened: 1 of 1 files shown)


docs/docs/databases/docker-add-drivers.mdx
```diff
@@ -8,86 +8,75 @@ version: 1
...
-[requirements.txt](https://github.com/apache/superset/blob/master/requirements.txt) and
-[requirements-dev.txt](https://github.com/apache/superset/blob/master/requirements-dev.txt)
-will be installed automatically.
+want to connect to.
 
-In this section, we'll walk through how to install the MySQL connector library. The connector
-library installation process is the same for all additional libraries and we'll end this section
-with the recommended connector library for each database.
+In this example, we'll walk through how to install the MySQL connector library. The connector
+library installation process is the same for all additional libraries.
```
john-bodley, line 14: @sfirke any reason we're force breaking lines in markdown? I'm seeing this throughout the docs and it seems somewhat atypically for us to be doing this given that markdown auto-wraps.

sfirke: Nope just ignorance on my part.  What is the guidance for breaking lines - should I manual newline after X (?) characters outside of markdown and then inside markdown, have indefinitely-long lines?

sfirke: And to be clear, is all of this documentation markdown - in which case do I do no manual line breaks in my IDE?

If there is a standard, maybe we can add it to the contributing guidelines (I will confess it could be there and I missed it).

sfirke: Forgot to do the @ mention @john-bodley  ^^

john-bodley: @sfirke I gather the logic should only be to include a manual newline when necessary from a rendering perspective,  new paragraph, bullet list, etc. In your editor enabling "word wrap" should ensure that you're not dealing with any horizontal scrolling.

I wonder if `prettier` or similar could somehow ensure that our Markdown was rendered correctly.

sfirke: Thanks @john-bodley, I enabled that in my VSCode and it's a big help.  I pushed changes that undid my manual newlines.

docs/docs/databases/docker-add-drivers.mdx
```diff
@@ -7,87 +7,57 @@ version: 1
...
-docker compose build --force-rm
-```
+You can check its presence by entering the running container with `docker exec -it <container_name> bash` and running `pip freeze`.  The PyPI package should be present in the printed list.
 
-After the rebuild of the Docker images is complete (which may take a few minutes) you can relaunch using the following command:
+**If you're running a customized docker image**, rebuild your local image with the new driver baked in:
 
 ```
-docker compose up
+docker compose build --force-rm
```
rusackas, line 44: Is this part necessary, or does `docker-compose up` do effectively the same thing?

rusackas (approved):
These look like great improvements, and I think the whole page makes more sens overall now.

docs/docs/databases/docker-add-drivers.mdx
sfirke: I'm not sure rebuilding is necessary if one is using requirements-local.txt 🤷  This command was already there and I was wary of deleting since I don't run a custom image.  I moved it to the "if you're running a custom image" section so hopefully most folks ignore it and those who would need it, know what they're doing.

docs/docs/databases/docker-add-drivers.mdx
mistercrunch: This came up here https://github.com/apache/superset/pull/26923/files where i was pointed to this comment. Wrote this over there:
> Personally I like line breaks and happily will get into a tabs-vs-space religious debate over it :) Main point is around "diffing". If you have a long paragraph and fix a typo, you're in the git blame for that whole line, and diffs will be much more verbose than needed. I vote for line breaks soft at 80-90, hard at 100, and across file most file types.

sfirke: I've added this item to the town hall agenda for the next meeting.

mistercrunch:
BTW while this is up, I noticed that our "lean" docker has stuff like Druid, Hive, and both Postgres and MySQL. Actually "lean" also has this development file reqs: https://github.com/apache/superset/blob/master/requirements/development.in

Arguably, I'd say we'd layer a "raw" layer without any drivers, a "lean" with Postgres only and the rest of drivers needed for tests/CI would be brought into the "ci" (top) layer

PR #26040, opened 2023-11-20
feat(helm): use env redis_driver, render additional deployment pod spec

qlimenoque:
### SUMMARY

- Allow to override redis connection driver with `REDIS_DRIVER` environment variable.

This change allows to use Redis Sentinel. So it's possible to use `sentinel://` or anything else instead of `redis://` in connection string

It's much easier to manage it through environment variable instead of rewriting the whole `broker_url` or `result_backend` in _configOverride_ value.

I leave `redis://`  as a default connection type in _values.yaml_.

- Allow to render additional pod spec to deployment/job template

In my case, I need to add `shareProcessNamespace: true` to the **init-db-job**
I've added this option to all possible workloads

- Add labels to init job

Allow to select helm hook pod with matchLabels in different scenarios (i.e. webhook)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


helm/superset/templates/_helpers.tpl
```diff
@@ -86,11 +86,11 @@ SQLALCHEMY_TRACK_MODIFICATIONS = True
 class CeleryConfig:
   imports  = ("superset.sql_lab", )
   {{- if .Values.supersetNode.connections.redis_password }}
-  broker_url = f"redis://:{env('REDIS_PASSWORD')}@{env('REDIS_HOST')}:{env('REDIS_PORT')}/0"
-  result_backend = f"redis://:{env('REDIS_PASSWORD')}@{env('REDIS_HOST')}:{env('REDIS_PORT')}/0"
+  broker_url = f"{env('REDIS_DRIVER')}://:{env('REDIS_PASSWORD')}@{env('REDIS_HOST')}:{env('REDIS_PORT')}/0"
+  result_backend = f"{env('REDIS_DRIVER')}://:{env('REDIS_PASSWORD')}@{env('REDIS_HOST')}:{env('REDIS_PORT')}/0"
```
dpgaspar, line 90: Can you include `CACHE_CONFIG` also?
do we need to add `REDIS_DRIVER` to `secrets-env.yaml` also so that the added new default `redis_driver` on values.yaml is set as an env var

qlimenoque: Thanks for the review! Yes, you are right I should add `REDIS_DRIVER` key and value to `secrets-env.yaml`
CACHE_CONFIG is configured in configOverride section by overriding `CACHE_CONFIG['CACHE_TYPE']` and other values.

For example:

```yaml
configOverrides:
  CACHE_TYPE: "CACHE_CONFIG['CACHE_TYPE'] = 'RedisSentinelCache'"
  CACHE_REDIS_SENTINELS: "CACHE_CONFIG['CACHE_REDIS_SENTINELS'] = [('rfs-redis', 26379)]"
  CACHE_REDIS_SENTINEL_MASTER: "CACHE_CONFIG['CACHE_REDIS_SENTINEL_MASTER'] = 'mymaster'"`
 ```
  At first, I added such changes for `CACHE_CONFIG`, but later I realized that it would be inconvenient and confusing to manage some changes through env and others through `configOverride`, so I deleted those changes. In this case, we need to convert the entire configuration to env variables, which will not be as flexible as using `configOverride`

qlimenoque: @dpgaspar
Hi! Any updates on this? :)

villebro:
@qlimenoque can you rebase this PR to resolve the conflict? Also, I'm wondering if it would make sense to separate the additional pod specs into a separate PR to have a more focused review of the requested changes?

qlimenoque:
> @qlimenoque can you rebase this PR to resolve the conflict? Also, I'm wondering if it would make sense to separate the additional pod specs into a separate PR to have a more focused review of the requested changes?

Sure, well pointed. I'll do it

rusackas:
This still needs a rebase. I'll mark it as draft for now, but please tell us when it's ready! We'd love to see this through, but it may be closed in time if it sees no further activity.

qlimenoque:
@rusackas done ✅

rusackas:
Hi @qlimenoque\! This PR has developed conflicts again since your last rebase. Could you update it when you get a chance? @villebro @dpgaspar - this one's been waiting for your review\!

rusackas (approved):
LGTM... I'll see about rebasing this since we've asked you to do it too often already.

rusackas:
Rebased this onto current `master` on the author's behalf (thanks @qlimenoque for the contribution! 🙏) and pushed a small follow-up:

- **Rebase**: resolved the only conflict (`Chart.yaml` version) and let `helm-docs` regenerate `README.md` for the new values.
- **Bug fix**: the `worker`/`beat`/`flower`/`ws` deployment templates guarded on their own component's `deploymentAdditionalPodSpec` but rendered `supersetNode`'s, so setting the value on a non-node component had no effect. Each now renders its own value (with a comment to prevent the copy-paste from recurring). `deployment.yaml` (node) and `init-job.yaml` were already correct.
- **Versioning**: bumped the chart to `0.16.0` (minor — adds the additional-pod-spec hooks and `redis_driver`).

Now conflict-free and ready for review.

rusackas:
@qlimenoque this is approved and ready on my end — it's just slipped back into conflict against `master`. I'll rebase it again since I'm here, but let me/us know your thoughts on the bot comments above.

qlimenoque:
@rusackas I made the branch up-to-date

rusackas (changes requested):
@qlimenoque circling back on my own approval here - @dpgaspar's question on `_helpers.tpl` still stands: `REDIS_DRIVER` needs wiring into `secret-env.yaml` (and ideally `CACHE_CONFIG`), and the bot's right that hardcoding `/0` drops `REDIS_CELERY_DB` and the SSL URL params. Could you square those away? I'd like to land it once that's sorted.

qlimenoque:
@rusackas rebased on latest master. `REDIS_DRIVER` is now wired into `secret-env.yaml` and the driver override applies to both the cache and Celery URLs

rusackas (approved):
Thanks @qlimenoque, this is looking great now. Folding the driver override into `REDIS_BASE_URL` is a clean way to do it, and since both the cache and Celery URLs build off that, @dpgaspar's `CACHE_CONFIG` question is covered in the same spot. LGTM!

One tiny optional thing the bot flagged: `redis_driver: ""` in `values.yaml` has no comment like its neighbors. Want to add a one-liner on what it's for (overriding `REDIS_PROTO` for Sentinel etc.)? Not a blocker either way.

I'll clear out the bot threads since they're all addressed by the rework. Thanks for sticking with this through all the rebases!

rusackas:
Calling it good enough... in it goes!

PR #26057, opened 2023-11-21
fix: radar chart sort by error

morpheus-lee:
fix(radar chart): 'sort by' in controlPanel correctly

### SUMMARY
There is a typo in radar chart buildQuery.ts. fix it by changed  `timeseries_limit_metric` to `orderby` and by formData.orderby in buildQuery.ts.

### TESTING INSTRUCTIONS
sort by will be normal function in radar chart.
It can be checked by result table

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
@@ -23,8 +23,7 @@ import {
 } from '@superset-ui/core';
 
 export default function buildQuery(formData: QueryFormData) {
-  const { series_limit_metric } = formData;
-  const sortByMetric = ensureIsArray(series_limit_metric)[0];
+  const sortByMetric = ensureIsArray(formData.orderby)[0];
 
   return buildQueryContext(formData, baseQueryObject => {
     let { metrics, orderby = [] } = baseQueryObject;
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
@@ -67,7 +67,7 @@ const config: ControlPanelConfig = {
       controlSetRows: [
         ['groupby'],
         ['metrics'],
-        ['timeseries_limit_metric'],
+        ['orderby'],
         ['adhoc_filters'],
         [
           {


john-bodley:
@i-love-thinking thanks for the PR. I've cc'ed a couple of frontend engineers who may have more context.

rusackas:
Re-pinging @kgabryje on this... 

Also, @yousoph do you know anyone using this chart? It'd be good to test on some various examples and check for regressions.

kgabryje:
@i-love-thinking Tested on Radar chart from examples, for some reason it's crashing when I add `Count` as sort by

image

i-love-thinking:
> @i-love-thinking Tested on Radar chart from examples, for some reason it's crashing when I add `Count` as sort by
> 
> image

@kgabryje Thanks your testing, I think maybe cause by `Count` is not a column. it's a metric and only one value, and maybe exclude by `sortby` control?

kgabryje:
> > @i-love-thinking Tested on Radar chart from examples, for some reason it's crashing when I add `Count` as sort by
> 
> @kgabryje Thanks your testing, I think maybe cause by `Count` is not a column. it's a metric and only one value, and maybe exclude by `sortby` control?

I think sort by works on metrics, not columns, and it seems to be working when I use `AVG(price_each)`. So maybe it's not handling a saved metric for some reason?

i-love-thinking:
@kgabryje Can you put on the same Radar chart, but without `sortby` setting.

superset-frontend/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
```diff
@@ -23,8 +23,7 @@ import {
 } from '@superset-ui/core';
 
 export default function buildQuery(formData: QueryFormData) {
-  const { series_limit_metric } = formData;
```
kgabryje, line 26: I double checked and I think that `timeseries_limit_metric` control is correct.
The problem is in this line - it should be `const { timeseries_limit_metric } = formData;` instead of `const { series_limit_metric } = formData;`.
After I change that name, everything seems to be working fine

kgabryje: And that's a good thing, because if we changed the control from `timeseries_limit_metric` to `orderby` we'd also need to add a migration for that control 😄 so fixing the bug here is less work

i-love-thinking:
@kgabryje The `sortby` function is for LEGEND. could it work fine?

PR #26059, opened 2023-11-21
fix: Flaky test_explore_json_async test

michael-s-molina:
### SUMMARY
This PR fixes the following flaky CI error:

```
FAILED tests/integration_tests/core_tests.py::TestCore::test_explore_json_async
- AssertionError: 200 != 202
```
The problem is that the response status will vary [if the results are cached](https://github.com/apache/superset/blob/97121465ddf772013604ffdb5d7378885bc6ee26/superset/views/core.py#L313) which depends on the tests execution order. Given that both `200` and `202` are valid response status for this endpoint, I updated the test to reflect that.

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/core_tests.py
+++ b/tests/integration_tests/core_tests.py
@@ -714,7 +714,7 @@ class TestCore(SupersetTestCase):
         keys = list(data.keys())
 
         # If chart is cached, it will return 200, otherwise 202
-        self.assertTrue(rv.status_code == 200 or rv.status_code == 202)
+        self.assertTrue(rv.status_code in {200, 202})
         self.assertCountEqual(
             keys, ["channel_id", "job_id", "user_id", "status", "errors", "result_url"]
         )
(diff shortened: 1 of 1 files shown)


tests/integration_tests/core_tests.py
```diff
@@ -713,7 +713,8 @@ def test_explore_json_async(self):
         data = json.loads(rv.data.decode("utf-8"))
         keys = list(data.keys())
 
-        self.assertEqual(rv.status_code, 202)
+        # If chart is cached, it will return 200, otherwise 202
+        self.assertTrue(rv.status_code == 200 or rv.status_code == 202)
```
john-bodley, line 717:
```suggestion
        self.assertTrue(rv.status_code in {200, 202})
```

john-bodley (approved):
One small comment, but otherwise LGTM.

PR #26060, opened 2023-11-21
fix: Optimize fetching samples logic

john-bodley:
### SUMMARY

Not a fix per se to https://github.com/apache/superset/pull/25995, but an alteration based on https://github.com/apache/superset/pull/25995#discussion_r1398499898. Per [this](https://github.com/apache/superset/blob/e7797b65d1dadc1c466d1852747657b0aade9690/superset/common/query_context_processor.py#L138-L169) code block it seems like my hypothesis was correct, i.e., though a cache key is known a priori it's only materialized when the query is successful and thus in our case if the `COUNT(*)` succeeds but the sample data query fails we should be clearing the cache for the former rather than the later.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/datasource/utils.py
+++ b/superset/views/datasource/utils.py
@@ -114,7 +114,7 @@ def get_samples(  # pylint: disable=too-many-arguments,too-many-locals
         sample_data = samples_instance.get_payload()["queries"][0]
 
         if sample_data.get("status") == QueryStatus.FAILED:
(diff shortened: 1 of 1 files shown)


michael-s-molina (approved):
LGTM

zhaoyongjie (approved):
LGTM

PR #26077, opened 2023-11-22
fix: Add all Trino dependencies for Kerberos enablement

john-bodley:
### SUMMARY

Per our [documentation](https://superset.apache.org/docs/databases/trino/#2-kerberos-authentication), Superset supports Kerberos authentication for Trino, however installing the `superset[trino]` dependencies are not suffice as the `trino` package additionally requires the [requests-kerberos](https://github.com/trinodb/trino-python-client/blob/f99ef245f28b4e1784a829b7935a31d0215ace4d/setup.py#L29-L34) package.

Rather than defining `trino[kerberos]` as a dependency I though it would be prudent to defining `trino[all]` as this contains additional requirements like SQLAlchemy.

### TESTING INSTRUCTIONS

Ran `pip-compile-multi --no-upgrade`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -143,9 +143,7 @@ geographiclib==1.52
 geopy==2.2.0
     # via apache-superset
 greenlet==2.0.2
-    # via
-    #   shillelagh
-    #   sqlalchemy
+    # via shillelagh
 gunicorn==21.2.0
     # via apache-superset
 hashids==1.3.1
@@ -371,6 +369,7 @@ werkzeug==2.3.3
     # via
     #   apache-superset
     #   flask
+    #   flask-appbuilder
     #   flask-jwt-extended
     #   flask-login
 wrapt==1.15.0
--- a/.github/workflows/bashlib.sh
+++ b/.github/workflows/bashlib.sh
@@ -40,8 +40,10 @@ default-setup-command() {
 apt-get-install() {
   say "::group::apt-get install dependencies"
   sudo apt-get update && sudo apt-get install --yes \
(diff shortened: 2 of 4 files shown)


requirements/base.txt
```diff
@@ -371,6 +369,7 @@ werkzeug==2.3.3
     # via
     #   apache-superset
     #   flask
+    #   flask-appbuilder
```
villebro, line 372: I think we need to add a CI linting rule to make sure the requirements files are always up to date (I assume this change didn't bring about this change). Any objections to me adding a step to the python lint CI workflow that does this in a separate PR?

villebro (approved):
LGTM

requirements/base.txt
john-bodley: @villebro I agree. We do have [this](https://github.com/apache/superset/blob/master/.pre-commit-config.yaml#L59-L62) check, however this just makes sure that the SHAs are correct. Maybe we need to run `pip-compile-multi --no-upgrade` and make sure there's no resulting diff?

villebro: > Maybe we need to run pip-compile-multi --no-upgrade and make sure there's no resulting diff?

@john-bodley yeah, I was thinking something along those lines

john-bodley:
Closing in favor of https://github.com/apache/superset/pull/26119 as we likely don't want to bloat our CI dependencies when there is an alternative workaround.

PR #26082, opened 2023-11-23
chore: lock the databend-sqlalchemy version

hantmac:
### SUMMARY

Lock the databend-sqlalchemy version

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -143,7 +143,7 @@ setup(
             "google-cloud-bigquery>=3.10.0",
         ],
         "clickhouse": ["clickhouse-connect>=0.5.14, <1.0"],
-        "databend": ["databend-sqlalchemy>=0.3.2, <0.5"],
+        "databend": ["databend-sqlalchemy>=0.3.2, <1.0"],
         "cockroachdb": ["cockroachdb>=0.3.5, <0.4"],
         "cors": ["flask-cors>=2.0.0"],
         "crate": ["crate[sqlalchemy]>=0.26.0, <0.27"],
--- a/setup.py
+++ b/setup.py
@@ -143,6 +143,7 @@ setup(
             "google-cloud-bigquery>=3.10.0",
         ],
         "clickhouse": ["clickhouse-connect>=0.5.14, <1.0"],
+        "databend": ["databend-sqlalchemy>=0.3.2, <0.5"],
         "cockroachdb": ["cockroachdb>=0.3.5, <0.4"],
         "cors": ["flask-cors>=2.0.0"],
         "crate": ["crate[sqlalchemy]>=0.26.0, <0.27"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -143,6 +143,7 @@ def get_git_sha() -> str:
             "google-cloud-bigquery>=3.10.0",
         ],
         "clickhouse": ["clickhouse-connect>=0.5.14, <1.0"],
+        "databend": ["databend-sqlalchemy>=0.3.2, <0.5"],
```
villebro, line 146: Is there some reason we want to support 0.3? Could we just update this to have the lower bound on the most recent verified version?

hantmac: The versions less than 0.3 have some compatibility issues. 
Of course, we can use this  lower bound. You mean we just `"databend": ["databend-sqlalchemy>=0.3.2"]`?

john-bodley:
Thanks @hantmac for the PR. Would you mind including a complete PR description per the template?

hantmac:
> Thanks @hantmac for the PR. Would you mind including a complete PR description per the template?

HI @john-bodley , what is the mean of `description per the template`? Or would u like to describe it more detail? Thank you very much!

john-bodley:
@hantmac when you create a PR it should have been auto populated with [this](https://github.com/apache/superset/blob/master/.github/PULL_REQUEST_TEMPLATE.md) template. You could copy-and-paste the markdown from said template and complete the necessary fields. This helps to provide reviewers et al. with the necessary information.

hantmac:
> @hantmac when you create a PR it should have been auto populated with [this](https://github.com/apache/superset/blob/master/.github/PULL_REQUEST_TEMPLATE.md) template. You could copy-and-paste the markdown from said template and complete the necessary fields. This helps to provide reviewers et al. with the necessary information.

Thanks! I will add the template with necessary information.

setup.py
john-bodley: @hantmac per [PyPI](https://pypi.org/project/databend-sqlalchemy/#history) the latest release of said package is `0.3.2` and thus the open upper limit of `0.5` seems somewhat arbitrary. Generally—assuming that the package is adhering to semver—placing an open upper limit of the major version (in this case `1.0`) is preferred as this _should_ prevent breaking changes.

hantmac: > @hantmac per [PyPI](https://pypi.org/project/databend-sqlalchemy/#history) the latest release of said package is `0.3.2` and thus the open upper limit of `0.5` seems somewhat arbitrary. Generally—assuming that the package is adhering to semver—placing an open upper limit of the major version (in this case `1.0`) is preferred as this _should_ prevent breaking changes.

You mean I should set `databend-sqlalchemy>=0.3.2, <1.0` ?

john-bodley (approved):

aliblade (approved):

aliblade (approved):

PR #26087, opened 2023-11-23
fix(plugin-chart-echarts): support numerical x-axis

### LINKED ISSUE #26034: [Regression] GENERIC_CHART_AXES: numeric x-axis treated as categorical values
When using a _numeric_ column as the generic x-axis, the x-axis values aren't linearly distributed along the x-axis but they are evenly distributed, just as if they were categorical values

#### How to reproduce the bug

0. Have the `GENERIC_CHART_AXES` feature flag enabled
1. Create a dataset or a SQLLab query with numerical columns, e.g.

```sql
select
  *
from (values
  (1,1)
  , (2,2)
  , (5,5)
  , (10,10)
) as v(x,y)
```

2. Visualize the query/dataset as bar chart (or line chart)
3. Use `x` as x-axis and, e.g., `max(y)` as y-axis
4. See that the x-axis values are shown with equal spacing instead of the expected linear distribution of the x values

### Expected results

This is what it should look like and what it did correctly in Superset 2.0.1:

image

### Actual results

X-axis values are seemingly treated as categorical values:

image

Any distribution/scatter/histogram plots based on this functionality are actually unusable due to this bug.

### Environment

- browser type and version:
- superset version: 3.0.2rc1
- node.js version: `node -v`
- any feature flags active:

### Checklist

villebro:
### SUMMARY
Currently using a numeric x-axis produces a chart with a categorical x-axis. This sets the x-axis to 

### AFTER
Now the x-axis is numerical:
image

### BEFORE
Previously the x-axis was set as categorical:
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: fixes #26034
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -20,9 +20,13 @@
 import { invert } from 'lodash';
 import {
   AnnotationLayer,
+  AxisType,
+  buildCustomFormatters,
   CategoricalColorNamespace,
+  CurrencyFormatter,
   ensureIsArray,
   GenericDataType,
+  getCustomFormatter,
   getMetricLabel,
   getNumberFormatter,
   getXAxisLabel,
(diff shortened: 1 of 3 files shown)


zhaoyongjie (approved):
LGTM

WojtekWaga (approved):

PR #26094, opened 2023-11-24
fix(migration): Support non-alphanumeric passwords in alembic.

### LINKED ISSUE #26029: URL encoded passwords are getting decoded in env.py
https://github.com/apache/superset/blob/890bf59ce4d820330fa545c33a6082e203442bc5/superset/migrations/env.py#L45

As suggested in https://github.com/psycopg/psycopg2/discussions/1546#discussion-4866365 we are using ```urllib.parse.quote(DB_PASSWORD)``` as the password contains special characters. But in the above line it gets decoded. Hence the host name is not getting resolved as part of the password is taken as host. Currently as a workaround line 45 is replaced with ```decoded_uri = DATABASE_URI.replace("%", "%%")```

ramki88:
This commit modifies alembic's `env.py`, so that it supports non-alphanumeric passwords.

### ADDITIONAL INFORMATION

- [ x] Has associated issue: Fixes #26029 
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/env.py
```diff
@@ -42,8 +42,8 @@
         "SQLite Database support for metadata databases will \
         be removed in a future version of Superset."
     )
-decoded_uri = urllib.parse.unquote(DATABASE_URI)
-config.set_main_option("sqlalchemy.url", decoded_uri)
+escaped_uri = DATABASE_URI.replace('%', '%%')
```
craig-rueda, line 45: This looks like it might break existing functionality. Should we instead do the replace on the "decoded_url" var?

ramki88: @craig-rueda You mean use the same var name? Or apply the replace on top of `urllib.parse.unquote(DATABASE_URI)`.

craig-rueda: Like this:

```
decoded_uri = urllib.parse.unquote(DATABASE_URI)
decoded_uri = decoded_uri.replace('%', '%%')
config.set_main_option("sqlalchemy.url", decoded_uri)
```

So basically just add the replace in there. 

Note that if your PWD has a `%` symbol in it, you may need to "enquote" it going in, i.e. `some_pass_with_%_in_it` would need to be url encoded as `some_pass_with_%25_in_it`

ramki88: Ahhh.. I got it now. Sorry I should have tried to explain in detail the need for this change. 

Example : Password with @ in them super2p@ss. As suggested in [SQLAlechemy Docs](https://docs.sqlalchemy.org/en/14/core/engines.html#escaping-special-characters-such-as-signs-in-passwords)  if the password is either escaped and passed to superset or in superset_config.py if the user escapes the special char using ```urllib.parse.quote_plus("super2p@ss")``` alembic's env.py is going to unquote it and we end up with "super2p@ss" which will not be parsed by psycopg2.

1. user passes  -> ```"super2p@ss"```
2. superset_config.py parsers -> ```"super2p%40ss"```
3. superset_config.py sets DATABASE_URI -> ```"postgresql+psycopg2://user:super2p%40ss@localhost:5432/superset"```
4. env.py decodes DATABASE_URI into decoded_uri -> ```"postgresql+psycopg2://user:super2p@ss@localhost:5432/superset"```

And superset app starts up, but the migration fails with the error "psycopg2.OperationalError: could not translate host name "ss@localhost" to address" as it splits at symbol "@".

Just by removing the unquote in env.py will not fix the problem as the password is being set as a config, which uses configparser hence the "%" has to be double escaped into %% for it to be valid.

so the flow becomes..
1. user passes  -> ```"super2p@ss"```
2. superset_config.py parsers -> ```"super2p%40ss"```
3. superset_config.py sets DATABASE_URI -> ```"postgresql+psycopg2://user:super2p%40ss@localhost:5432/superset"```
4. env.py doubles escapes DATABASE_URI into escaped_uri -> ```"postgresql+psycopg2://user:super2p%%40ss@localhost:5432/superset"```

References:
https://github.com/psycopg/psycopg2/discussions/1546
https://github.com/sqlalchemy/alembic/issues/700#issuecomment-636036888

craig-rueda: Ok, makes sense. From what it looks like here, we're just trying to address an errant `@` from finding its way into the connection string.

Hoping this doesn't break anyone 🙏 

CC - @john-bodley @dpgaspar  ??

rusackas:
Hmm... title looks fine... but other CI tasks are stuck too. Closing/reopening to kick this. It might need a rebase, and @craig-rueda might want to take a fresh look.

rusackas:
@ramki88 would you mind running the [pre-commit hooks](https://superset.apache.org/docs/contributing/hooks-and-linting/) on this so it can pass CI/linting?

ramki88:
> @ramki88 would you mind running the [pre-commit hooks](https://superset.apache.org/docs/contributing/hooks-and-linting/) on this so it can pass CI/linting?

Sure, I've ran it and rebased with master.
image

rusackas:
Running CI. Fingers crossed!

kahlua-kol:
Hey! Do you have any news on this topic? :) Would be really helpful to have a working fix for special characters

ramki88:
> I think the fact that some tests aren't running, and we have a weird (possibly unrelated) MySQL test error, indicates that some of these CI tasks have been fixed on `master` but aren't being reflected here. A rebase of this branch/PR (again, I know, sorry...) might resolve that. CI has been a bit turbulent lately.

Sure, I have rebased with master https://github.com/apache/superset/pull/26094/commits/8fb2cae8fc8cbc36aca02ab2caf72b755663da4c

rusackas:
Resolved the conflict that's been holding this up forever. Would love to get this merged. Maybe @dpgaspar has opinions on this and can provide a review.

villebro:
@rusackas this was already fixed by #30532, so we can close this one (after the conflict resolution this PR is only adding a single blank line 😆 ). I'm gonna close this.

PR #26106, opened 2023-11-27
fix: flaky test_explore_json_async test v2

villebro:
### SUMMARY
#26059 relaxed the assertion in `test_explore_json_async` to not fail if the result was already cached. However, since the response payload is different for the 202 response, the test still fails if a 200 is received. This PR changes the assertion from the legacy `self.assert` notation to the preferred `assert` one, and makes the `keys` assertion both conditional on a 202 response, and explicitly asserts the returned keys, instead of the key count.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/core_tests.py
+++ b/tests/integration_tests/core_tests.py
@@ -714,10 +714,16 @@ class TestCore(SupersetTestCase):
         keys = list(data.keys())
 
         # If chart is cached, it will return 200, otherwise 202
-        self.assertTrue(rv.status_code in {200, 202})
-        self.assertCountEqual(
-            keys, ["channel_id", "job_id", "user_id", "status", "errors", "result_url"]
-        )
+        assert rv.status_code in {200, 202}
+        if rv.status_code == 202:
+            assert keys == [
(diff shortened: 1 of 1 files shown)


michael-s-molina (approved):
LGTM. Thanks for the improvement!

tests/integration_tests/core_tests.py
```diff
@@ -714,10 +714,16 @@ def test_explore_json_async(self):
         keys = list(data.keys())
 
         # If chart is cached, it will return 200, otherwise 202
-        self.assertTrue(rv.status_code in {200, 202})
-        self.assertCountEqual(
-            keys, ["channel_id", "job_id", "user_id", "status", "errors", "result_url"]
-        )
+        assert rv.status_code in {200, 202}
```
john-bodley, line 717: @villebro I was wondering whether it would be easier/simpler to disable the caching for this test so the response status code is always 200. The only caveat is whether we should be testing both scenarios.

PR #26120, opened 2023-11-28
fix: alias column when fetching values

betodealmeida:
### SUMMARY

The endpoint that returns values for a given column, which powers filter dropdowns, is failing for some databases (Snowflake, Clickhouse) because the query generated has an unusual alias for the column. 

For example, for Clickhouse, when selecting the values of a column called `name` we have the following query:

```sql
SELECT DISTINCT name AS name_12345
FROM ...
```

Clickhouse adds the stochastic alias `name_#####`, and reading the values from the Pandas dataframe fails because the column `name` is not present.

For Snowflake we have a different issue:

```sql
SELECT DISTINCT NAME AS NAME
FROM ...
```

Here, the resulting dataframe will have `NAME` but not `name`, since for Snowflake these are identical.

This PR fixes the problem by adding an alias to the projection. This way we know the alias is present in the dataframe.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -1352,7 +1352,13 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
         tbl, cte = self.get_from_clause(tp)
 
         qry = (
-            sa.select([target_col.get_sqla_col(template_processor=tp)])
+            sa.select(
+                # The alias (label) here is important because some dialects will
+                # automatically add a random alias to the projection because of the
+                # call to DISTINCT; others will uppercase the column names. This
+                # gives us a deterministic column name in the dataframe.
+                [target_col.get_sqla_col(template_processor=tp).label("column_values")]
+            )
             .select_from(tbl)
             .distinct()
         )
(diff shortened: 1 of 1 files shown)


zephyring (approved):

lilykuang (approved):

sadpandajoe:
🏷️ preset:2023.47

superset/models/helpers.py
```diff
@@ -1352,7 +1352,13 @@ def values_for_column(self, column_name: str, limit: int = 10000) -> list[Any]:
         tbl, cte = self.get_from_clause(tp)
 
         qry = (
-            sa.select([target_col.get_sqla_col(template_processor=tp)])
+            sa.select(
+                # The alias (label) here is important because some dialects will
+                # automatically add a random alias to the projection because of the
+                # call to DISTINCT; others will uppercase the column names. This
+                # gives us a deterministic column name in the dataframe.
+                [target_col.get_sqla_col(template_processor=tp).label("column_values")]
```
michael-s-molina, line 1360: Thanks for adding a comment explaining the context! It's always helpful.

PR #26129, opened 2023-11-29
docs: add quickstart

artofcomputing:
### SUMMARY
As per Discussion #25976 and [DISCUSS] sent over the [Apache dev mailing list](https://lists.apache.org/thread/7w51kd3fktgmx2mnjgk33y554zcbypb1), this PR creates a quickstart page, simplifying the installation process, and providing:
- Simplifying the installation method.
- Allowing users to get up and running a Superset instance faster.

While allowing us to:
- Split installation and configuration topics.
- Provide a stable playground for users to explore (by making the user to install a stable release of Superset).

### DIFF
--- a/docs/docs/quickstart.mdx
+++ b/docs/docs/quickstart.mdx
@@ -4,19 +4,22 @@ hide_title: false
 sidebar_position: 2
 ---
 
-**Ready to give Apache Superset a try?** This quickstart will help you run Superset on your local machine in **5 simple steps**.
-It assumes that you have [Docker](https://www.docker.com) installed.
+**Ready to give Apache Superset a try?** This quickstart will help you run Superset on your local machine in
+**5 simple steps**. It assumes that you have [Docker](https://www.docker.com) installed.
 
 ### 1. Get Superset
-To get started, add the latest Superset version in the `SUPERSET_VERSION` environment variable, and pull the container from Docker Hub:
+To get started, set the `SUPERSET_VERSION` environment variable with the latest Superset version.
+[Click here](https://github.com/apache/superset/releases) to check the latest version.
+
 ```
-$ export SUPERSET_VERSION="";
-         docker pull apache/superset:$SUPERSET_VERSION
+$ export SUPERSET_VERSION=<latest_version>
 ```
-:::tip
-Not sure what the latest stable Superset version is? [Click here](https://github.com/apache/superset/releases) to check our release page 
-for up-to-date information.
-:::
+
+Pull the Superset image from Docker Hub:
+
+```
+$ docker pull apache/superset:$SUPERSET_VERSION
+```
 
 ### 2. Start Superset
 ```
@@ -49,7 +52,8 @@ $ docker exec -it superset superset db upgrade &&
 ```
 :::tip
 This step can take some time. While you wait, feel free to join the official Slack channel to check for new releases,
-ask questions, and engage with the community. [Click here to join.](https://apache-superset.slack.com/join/shared_invite/zt-26ol9ge4y-kzUnSo9inRepOay0ufBTsA#/shared-invite/email)
+ask questions, and engage with the community.
+[Click here to join.](https://apache-superset.slack.com/join/shared_invite/zt-26ol9ge4y-kzUnSo9inRepOay0ufBTsA#/shared-invite/email)
 :::
 
 ### 5. Start using Superset
--- a/docs/docs/quickstart.mdx
+++ b/docs/docs/quickstart.mdx
@@ -8,20 +8,26 @@ sidebar_position: 2
 It assumes that you have [Docker](https://www.docker.com) installed.
 
 ### 1. Get Superset
+To get started, add the latest Superset version in the `SUPERSET_VERSION` environment variable, and pull the container from Docker Hub:
 ```
-$ docker pull apache/superset:3.0.2
+$ export SUPERSET_VERSION="";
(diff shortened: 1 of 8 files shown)


docs/docs/quickstart.mdx
```diff
@@ -0,0 +1,76 @@
...
+hide_title: false
+sidebar_position: 2
+---
+
+**Ready to give Apache Superset a try?** This quickstart will help you run Superset on your local machine in **5 simple steps**.
+It assumes that you have [Docker](https://www.docker.com) installed.
+
+### 1. Get Superset
+```
+$ docker pull apache/superset:3.0.2
```
john-bodley, line 12: Is it possible to not pin the version, maybe something like:

```suggestion
$ docker pull apache/superset:<version>
```

artofcomputing: Right now it's possible but it would require versioning the documentation. The pinning is done currently to "stabilize" versions as not to let the latest tag be used, since that is currently bound to the master branch, instead of the latest release.

artofcomputing: Something that we could do (just an idea) would be to update the version in the docs with every release, and prioritize the latest upstream version (e.g 3.1 instead of 3.0.3)

@sfirke @michael-s-molina @rusackas would that be feasible?

john-bodley: @artofcomputing the challenge is, without a 🔮, we can't be certain what the upstream version is, i.e., it could be 3.1 or 4.0.

artofcomputing: True, there's no way to tell. IMO we could define a semantic rule to favor the latest major version on the page, and minor/patch versions if there's no new major updates. But yeah, that would require manual bumps, or some other way, like <version> as you mentioned and let user manually set it. Any thoughts?

artofcomputing: That's reasonable, I'll add the placeholder with information on where to instruct users.

artofcomputing: @john-bodley @michael-s-molina made the change in the step 1 and 2 with commit [328e871](https://github.com/apache/superset/pull/26129/commits/328e8715016a5a5871a412dc6c6a4c703b18c9d2), and added a tip on how users can find the latest release

docs/docs/quickstart.mdx
```diff
@@ -0,0 +1,82 @@
...
+### 1. Get Superset
+To get started, add the latest Superset version in the `SUPERSET_VERSION` environment variable, and pull the container from Docker Hub:
+```
+$ export SUPERSET_VERSION="";
+         docker pull apache/superset:$SUPERSET_VERSION
+```
+:::tip
+Not sure what the latest stable Superset version is? [Click here](https://github.com/apache/superset/releases) to check our release page 
+for up-to-date information.
+:::
```
michael-s-molina, line 19: Maybe this would be more intuitive?

To get started, set the `SUPERSET_VERSION` environment variable with the latest Superset version. [Click here](https://github.com/apache/superset/releases) to check the latest version.
```
$ export SUPERSET_VERSION=<latest_version>
```

Pull the Superset image from Docker Hub:
```
$ docker pull apache/superset:$SUPERSET_VERSION
```

docs/docs/quickstart.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+### 2. Start Superset
+```
+$ docker run -d -p 8080:8088 \
+             -e "SUPERSET_SECRET_KEY=$(openssl rand -base64 42)" \
+             -e "TALISMAN_ENABLED=False" \
+             --name superset apache/superset:$SUPERSET_VERSION
+```
+:::tip
```
michael-s-molina, line 28: Maybe this should be placed before the command so users set the config before running.

docs/docs/quickstart.mdx
artofcomputing: Yeah that's better, I just changed the first step to it
image

docs/docs/quickstart.mdx
artofcomputing: You mean the TALISMAN environment variable, or about the container?

michael-s-molina: I mean placing the tip before the command:

Note that some configuration is mandatory for Superset in order to start. In particular, Superset will not start without a user-specified value of `SECRET_KEY` in a Superset configuration file or `SUPERSET_SECRET_KEY` as an environment variable.Please see [Configuring Superset](https://superset.apache.org/docs/installation/configuring-superset/) for more details.

```
$ docker run -d -p 8080:8088 \
             -e "SUPERSET_SECRET_KEY=$(openssl rand -base64 42)" \
             -e "TALISMAN_ENABLED=False" \
             --name superset apache/superset:$SUPERSET_VERSION
```

artofcomputing: Ah, that can be done too, just made the change on commit https://github.com/apache/superset/pull/26129/commits/229bf700e80ab2d49d49b4eaecfeea8ee3415c20

michael-s-molina (approved):
LGTM. Thanks for the improvement @artofcomputing!

sfirke:
Sorry I didn't get my review in earlier - but it's okay because it looks great!  Fantastic work @artofcomputing !!  And thanks @michael-s-molina and @john-bodley for the feedback.  I like having the users check the latest and specify that version manually.

PR #25557, opened 2023-10-06
feat: Improves the Waterfall chart

michael-s-molina:
### SUMMARY
This PR improves the Waterfall chart developed in https://github.com/apache/superset/pull/17906 with the following changes:

1 - Improves the chart description and provide example thumbnails

Screenshot 2023-10-19 at 14 51 47

<br>2 - Fixes a bug with negative values where the bars are displayed above the 0 line

Screenshot 2023-10-18 at 16 03 20

<br>3 - Augments the tooltip with intermediate totals

Screenshot 2023-10-18 at 16 04 26

<br>4 - Adds support for generic chart axes

Screenshot 2023-10-18 at 16 05 08

<br>5 - Adds support for series color selection

Screenshot 2023-10-18 at 16 05 45

<br>6 - Adds support for currency formatting

Screenshot 2023-10-18 at 16 06 16

<br>7 - Adds support for time formatting

Screenshot 2023-10-18 at 16 06 46

<br>8 - Adds support for x-axis with mixed data types when using breakdowns

Screenshot 2023-10-18 at 16 08 36

<br>Hat tip to @rusackas for testing the plugin and documenting the issues [here](https://github.com/apache/superset/pull/17906#issuecomment-1749740928) and @stephenLYZ for creating the first version of the plugin.

### TESTING INSTRUCTIONS
1 - Create a Waterfall chart
2 - Test supported controls
3 - Test supported dashboard interactions

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26136, opened 2023-11-29
feat: Adds legacy time support for Waterfall chart

michael-s-molina:
### SUMMARY
Follow-up of https://github.com/apache/superset/pull/25557 to add support for legacy time controls in Waterfall charts.

Screenshot 2023-11-29 at 15 47 46

### TESTING INSTRUCTIONS
1 - Set `GENERIC_CHART_AXES` feature flag to False
2 - Test if you are able to create a Waterfall chart using the legacy time controls

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Waterfall/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Waterfall/controlPanel.tsx
@@ -17,25 +17,27 @@
  * under the License.
  */
 import React from 'react';
-import { t } from '@superset-ui/core';
+import { hasGenericChartAxes, t } from '@superset-ui/core';
 import {
   ControlPanelConfig,
   ControlSubSectionHeader,
   D3_TIME_FORMAT_DOCS,
   DEFAULT_TIME_FORMAT,
   formatSelectOptions,
+  sections,
   sharedControls,
 } from '@superset-ui/chart-controls';
 import { showValueControl } from '../controls';
 
 const config: ControlPanelConfig = {
   controlPanelSections: [
+    sections.genericTime,
     {
       label: t('Query'),
       expanded: true,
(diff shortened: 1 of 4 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Waterfall/controlPanel.tsx
```diff
@@ -17,25 +17,27 @@
...
 const config: ControlPanelConfig = {
   controlPanelSections: [
+    sections.genericTime,
     {
       label: t('Query'),
       expanded: true,
       controlSetRows: [
-        ['x_axis'],
-        ['time_grain_sqla'],
+        [hasGenericChartAxes ? 'x_axis' : null],
```
justinpark, line 39: Doesn't a row value of null result in an empty DOM placeholder?
How about conditionally appending generic x-axis control?
```suggestion
        ...(hasGenericChartAxes ? ['x_axis', 'time_grain_sqla'] : []),
```

justinpark (approved):
LGTM upon @villebro 's approval

michael-s-molina:
> @michael-s-molina a general observation about this change - as we've already enabled GENERIC_CHART_AXES by default, do we really want to support non-generic mode in new viz types? This adds maintenance burden, and makes the task of deprecating those legacy controls even bigger. As an alternative, maybe we could just disable this chart type if generic chart axes is disabled?

@villebro You're correct. I think there's no obligation to support non-generic mode in new viz types. I just did it for this chart because we want it at Airbnb but we didn't enable `GENERIC_CHART_AXES` yet. Luckily, 4.0 is coming (March 2024) to delete the legacy code and resolve our problems. We'll enable that feature flag by then.

PR #26138, opened 2023-11-29
feat(accessibility): add tabbing to chart menu in dashboard

eschutho:
### SUMMARY

This change adds the ability to tab through elements on a dashboard, the title, charts, and into the chart controls, and execute each of them with a keyboard. 

https://github.com/apache/superset/assets/5186919/4dd70521-131a-438d-b370-ab7557e1176b

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


yousoph:
When using keyboard navigation to tab through on a dashboard, the highlight on the chart title in dashboards doesn't seem to be appearing for me

kasiazjc:
> @eschutho Ephemeral environment spinning up at http://35.87.72.46:8080. Credentials are `admin`/`admin`. Please allow several minutes for bootstrapping and startup.

I think this looks good! Thanks @eschutho 🙏

superset-frontend/cypress-base/cypress/e2e/dashboard/drilltodetail.test.ts
```diff
@@ -249,9 +249,13 @@ describe('Drill to detail modal', () => {
       it('opens the modal with the correct filters', () => {
         interceptSamples();
 
+        // focus on table first to trigger browser scroll
+        cy.get("[data-test-viz-type='table']").contains('boy').rightclick();
+
+        cy.wait(500);
```
rusackas, line 255: I wonder if there's something we can do with timeouts instead of waiting, but... probably not worth the hassle.

superset-frontend/src/components/PageHeaderWithActions/index.tsx
```diff
@@ -43,6 +43,9 @@ export const menuTriggerStyles = (theme: SupersetTheme) => css`
   &:hover:not(:focus) > span.anticon {
     color: ${theme.colors.primary.light1};
   }
+  &:focus-visible {
+    outline: 2px solid ${theme.colors.primary.dark2};
```
rusackas, line 47: One of these years I'm going to add some borderWidth properties to the theme...

superset-frontend/src/dashboard/components/DashboardBuilder/DashboardContainer.tsx
```diff
@@ -233,6 +234,18 @@ const DashboardContainer: FC<DashboardContainerProps> = ({ topLevelTabs }) => {
...
             allowOverflow
+            onFocus={e => {
+              if (
+                // prevent scrolling when tabbing to the tab pane
+                e.target.classList.contains('ant-tabs-tabpane') &&
+                window.scrollY < TOP_OF_PAGE_RANGE
+              ) {
+                // prevent window from jumping down when tabbing
+                // if already at the top of the page
+                // to help with accessibility when using keyboard navigation
```
rusackas, line 245: OMG comments! ❤️

superset-frontend/src/dashboard/components/SliceHeaderControls/SliceHeaderControls.test.tsx
```diff
@@ -350,3 +353,127 @@ test('Should not show the "Edit chart" without proper permissions', () => {
...
+  );
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('should continue with system tab navigation if dropdown is closed and tab key is pressed', () => {
+    const event = { key: 'Tab', preventDefault: jest.fn() };
+    handleDropdownNavigation(
+      // @ts-ignore
```
rusackas, line 376:
```suggestion
      // @ts-ignore
```
Not blocking, but... how bad is it? 😰

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -305,6 +399,7 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
       case MENU_KEYS.EXPLORE_CHART:
         // eslint-disable-next-line no-unused-expressions
         props.logExploreChart?.(props.slice.slice_id);
+        window.open(props.exploreUrl, '_blank');
```
rusackas, line 402: Is popping open in  `_blank` the right default here, or should it open in the same window?

rusackas:
This changes the default behavior of the Edit Chart link, from opening in the same window to a new one. This might open an old can of worms — is it intentional?

superset-frontend/src/dashboard/styles.ts
```diff
@@ -97,3 +111,43 @@ export const chartContextMenuStyles = (theme: SupersetTheme) => css`
...
+    }
+    &:not(.superset-button, .ant-menu-item
+        a, .fave-unfave-icon, .ant-tabs-tabpane, .header-controls span) {
+      &:focus-visible {
+        padding: ${theme.gridUnit / 2}px;
+      }
+    }
+  }
+
+  @keyframes fadeOut {
```
rusackas, line 136: Oooooh, fancy! Just another "note to self" that we can add shadows to the theme as well, so they can be appropriate "glows" in dark mode, and be standardized across the UI.

superset-frontend/cypress-base/cypress/e2e/dashboard/drilltodetail.test.ts
eschutho: Yeah this was really tricky because the browser scrolls automatically on focus, and then the field goes under the header and the right click doesn't work. I adjusted for this behavior for real users in the onFocus below but for automations, the wait was the only way to get the browser to catch up. But, that said, there are a few waits earlier in this script that I think we can change to timeouts/assert.is.visible, etc.

superset-frontend/src/dashboard/components/SliceHeaderControls/SliceHeaderControls.test.tsx
eschutho: Oh, it didn't like the type of one of the arguments. I think it was the event that I mocked. I could try to redo it if you think it would be better.

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: I took from the same behavior as the click, which is a new window. But we could change it if we need to?

superset-frontend/cypress-base/cypress/e2e/dashboard/drilltodetail.test.ts
eschutho: let's see if the above changes make this a net neutral/positive.

eschutho:
> This changes the default behavior of the Edit Chart link, from opening in the same window to a new one. This might open an old can of worms — is it intentional?

No, not intentional. Thanks for flagging that.

superset-frontend/src/dashboard/components/SliceHeaderControls/SliceHeaderControls.test.tsx
rusackas: Well thanks for fixing... I'm not worried about ts-ignore in tests, so much as the actual application. It's another pile I want to track as tech debt when that dashboard is out in public hands :)

superset-frontend/src/dashboard/styles.ts
rusackas: If you wanna have some fun, you could still add a base box-shadow style to the component with no visible shadow, and add a `transition` property so that the shadow fades on and/or moves when the `focus-visible` pseudoclasss is applied.

eschutho: I think we're ok with the outline when someone tabs into an element, but it just didn't feel right when clicking on an element, especially when the click bubbles up to an outer element and you have no idea why the entire tab is highlighting.

superset-frontend/cypress-base/cypress/e2e/dashboard/editmode.test.ts
```diff
@@ -483,51 +483,6 @@ describe('Dashboard edit', () => {
...
-      // open main tab and nested tab
-      openTab(0, 0);
-      openTab(1, 1);
-
-      // label Anthony
-      cy.get(
-        '[data-test-chart-name="Top 10 California Names Timeseries"] .line .nv-legend-symbol',
-      )
-        .first()
-        .should('have.css', 'fill', 'rgb(244, 176, 42)');
```
eschutho, line 530: I'm just testing what's actually breaking in this file. Will revert

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
```diff
@@ -123,7 +123,7 @@ export default function ErrorAlert({
         {!isExpandable && !description && (
           <span
             role="button"
-            tabIndex={0}
+            tabIndex={-1}
```
kgabryje, line 126: Why are we setting tab index to -1 (here and other places)? Doesn't it have a negative effect on accessibility, since you can't navigate to it with tab anymore?

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -231,7 +310,7 @@ const ViewResultsModalTrigger = ({
               <Button
                 buttonStyle="primary"
                 buttonSize="small"
-                onClick={closeModal}
+                onClick={() => setShowModal(false)}
```
kgabryje, line 313: is that change necessary? I think it's the same as `closeModal`

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -261,10 +340,23 @@ const ViewResultsModalTrigger = ({
...
+  const [drillModalIsOpen, setDrillModalIsOpen] = useState(false);
+  const [selectedKeys, setSelectedKeys] = useState<string[]>([]);
   const [openScopingModal, scopingModal] = useCrossFiltersScopingModal(
     props.slice.slice_id,
   );
 
+  const queryMenuRef: RefObject<any> = useRef(null);
+  const menuRef: RefObject<any> = useRef(null);
+
+  const toggleDropdown = ({ close }: { close?: boolean } = {}) => {
```
kgabryje, line 354: Why an object in args? Won't `(close: boolean) => {...}` work the same?

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -309,6 +403,7 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
       case MENU_KEYS.EXPLORE_CHART:
         // eslint-disable-next-line no-unused-expressions
         props.logExploreChart?.(props.slice.slice_id);
+        window.open(props.exploreUrl);
```
kgabryje, line 406: Why do we remove a link and use `window.open` instead? The previous approach had all the benefits of using a normal link, like right click context menu

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -170,25 +182,91 @@ const dropdownIconsStyles = css`
...
 `;
 
+export const handleDropdownNavigation = (
+  e: React.KeyboardEvent<HTMLElement>,
+  dropdownIsOpen: boolean,
+  menu: React.ReactElement,
+  toggleDropdown: () => void,
+  setSelectedKeys: (keys: string[]) => void,
+) => {
+  if (e.key === NAV_KEYS.tab && !dropdownIsOpen) {
```
kgabryje, line 192: Should it be `||`?

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -573,15 +691,39 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
...
+          css={() => css`
             display: flex;
             align-items: center;
           `}
           id={`slice_${slice.slice_id}-controls`}
           role="button"
           aria-label="More Options"
+          tabIndex={0}
+          onClick={() => {
+            toggleDropdown();
```
kgabryje, line 706: I think `toggleDropdown` should go to `onVisibleChange` callback of `NoAnimationDropdown`

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -573,15 +691,39 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
...
+            // close unless the dropdown menu is clicked
+            const relatedTarget = e.relatedTarget as HTMLElement;
+            if (
+              dropdownIsOpen &&
+              menuRef?.current?.props.id !== relatedTarget?.id
+            ) {
+              toggleDropdown({ close: true });
+            }
+          }}
+          onKeyDown={e =>
```
kgabryje, line 718: I'm wondering if it wouldn't be better to bake all that logic in `NoAnimationDropdown`. We could add those new props (onKeyDown, onBlur) to `NoAnimationDropdown` child by calling `cloneElement`

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
eschutho: Even when navigating to the element, there were no onKeyDown handlers, so you wouldn't be able to do anything after you tabbed to the element, so I removed the ability to tab, because it was out of scope, but I just added the error modal tab functionality since I was here. :)

https://github.com/apache/superset/assets/5186919/6141f4da-c474-47ba-a64e-9bc652b072ec

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: Here's the expected behavior: if you are tabbing through the page and you are on the slice controls and hit tab when the dropdown is not open, you should jump to the next element in the page instead of interacting with the dropdown elements. So, I think `&&` is right?

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: thanks for catching! Yes, I removed the `closeModal` method in one of the iterations and forgot to put it back.

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: Yes, you're right, the functionality is the same. It's so that when you call `toggleDropdown({ close: true });` it's more explicit than `toggleDropdown(true);`

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: When tabbing, the user will land on the Menu item (the li) and then would need to tab again to hit the link in order to open it. This way the action for enter is on the Menu item, so only one tab is needed. It makes sense to have the other context menu benefits, etc, but it's actually really difficult to right click on that element because it gets picked up by the context menu in the chart below. If you have any suggestions about how to add it back though, lmk.

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: nice, thanks for that tip! That worked great.

superset-frontend/src/features/home/Menu.tsx
```diff
@@ -306,11 +306,15 @@ export function Menu({
...
+              <GenericLink
+                className="navbar-brand"
+                to={brand.path}
+                tabIndex={-1}
+              >
                 <img src={brand.icon} alt={brand.alt} />
               </GenericLink>
             ) : (
-              <a className="navbar-brand" href={brand.path}>
+              <a className="navbar-brand" href={brand.path} tabIndex={-1}>
```
eschutho, line 317: I'm removing these tabs because they were out of scope. They currently will work and open the homepage when someone tabs and hits enter, but the logo gets larger and distorted when focused and doesn't highlight very well, and the functionality (sending someone to the homepage) for someone who will be using keyboard navigation on this page is not really helpful. We can add it in later, or in this PR if someone feels it's an important step.

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
kgabryje: In that case, maybe we could set tabIndex=-1 on that particular menu item, so that tab would focus on the link?

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -170,25 +182,91 @@ const dropdownIconsStyles = css`
   }
 `;
 
+export const handleDropdownNavigation = (
```
kgabryje, line 185: I know that this PR is focused on dashboard interactions, but maybe (future iterations) we should add this functionality to the main Dropdown component, so that the keyboard navigation works in all dropdowns?

kgabryje:
When I'm navigating the dropdown with arrows, the whole page scrolls with every click 🙁 And it looks like I can't navigate to Submenus


https://github.com/apache/superset/assets/15073128/9ec868f4-46c5-40bb-acdd-dab953baaeab

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
eschutho: Yeah, good idea!

eschutho:
> When I'm navigating the dropdown with arrows, the whole page scrolls with every click 🙁 And it looks like I can't navigate to Submenus
> 
Yeah, I didn't add the submenus yet. That's for a separate PR.

superset-frontend/src/dashboard/types.ts
```diff
@@ -237,3 +237,32 @@ export type Slice = {
...
+  SaveModal = 'save_modal',
+  RefreshDashboard = 'refresh_dashboard',
+  AutorefreshModal = 'autorefresh_modal',
+  SetFilterMapping = 'set_filter_mapping',
+  EditProperties = 'edit_properties',
+  EditCss = 'edit_css',
+  ToggleFullscreen = 'toggle_fullscreen',
+  ManageEmbedded = 'manage_embedded',
+  ManageEmailReports = 'manage_email_reports',
+}
```
eschutho, line 268: Is this file just for typescript types or .. more?

geido: What do you mean for "more"?

geido (approved):

superset-frontend/src/dashboard/types.ts
eschutho: I thought maybe this file was only to be used for typescript types rather than runtime enums.

superset-frontend/cypress-base/cypress/e2e/dashboard/editmode.test.ts
```diff
@@ -771,7 +771,7 @@ describe('Dashboard edit', () => {
       visitEdit();
     });
 
-    it('should save', () => {
+    it.skip('should save', () => {
```
rusackas, line 774: @eschutho I'm not sure the fate of these tests... should we scrap them, or try to rekindle them?

PR #26140, opened 2023-11-29
fix(annotations): time grain column

betodealmeida:
### SUMMARY

A confusing bug in the annotation layers... it seems that in the form data the `granularity` attribute stores the time grain (eg, `P1D` for daily time grain):

https://github.com/apache/superset/blob/f1a6b2f852eb879e69c7a93803cf0dd2ed58ce76/superset-frontend/src/explore/controls.jsx#L248-L270

But in the request payload, `granularity` should point to the column name, not the grain:

https://github.com/apache/superset/blob/f1a6b2f852eb879e69c7a93803cf0dd2ed58ce76/superset/charts/schemas.py#L1146-L1149

When generating an annotation query we need to move things around to match the form data to the request schema.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Chart/chartAction.js
+++ b/superset-frontend/src/components/Chart/chartAction.js
@@ -269,9 +269,13 @@ export function runAnnotationQuery({
       return Promise.resolve();
     }
 
-    const granularity = fd.time_grain_sqla || fd.granularity;
-    fd.time_grain_sqla = granularity;
-    fd.granularity = granularity;
+    // remap from controls to formData naming; in the original formData the
+    // `granularity` attribute represents the time grain, but in the request
+    // payload it corresponds to the name of the column where the time grain
+    // should be applied...
+    fd.time_grain_sqla = fd.time_grain_sqla || fd.granularity;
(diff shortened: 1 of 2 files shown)


superset-frontend/src/components/Chart/chartAction.js
```diff
@@ -269,9 +269,13 @@ export function runAnnotationQuery({
       return Promise.resolve();
     }
 
-    const granularity = fd.time_grain_sqla || fd.granularity;
-    fd.time_grain_sqla = granularity;
-    fd.granularity = granularity;
+    // remap from controls to formData naming; in the original formData the
+    // `granularity` attribute represents the time grain, but in the request
+    // payload it corresponds to the name of the column where the time grain
+    // should be applied...
```
zephyring, line 275: it will be helpful to add sample data for time_grain and granularity as inline comment. .e.g
original `granularity` can be `P1D` but in request payload expect to be column name like `Date`

Antonio-RiveroMartnez (approved):

PR #25579, opened 2023-10-09
fix(Charts): Set max row limit + removed the option to use an empty row limit value

CorbinBullard:
### SUMMARY

- Can no longer clear/remove the row limit
- New validator that checks if a value is larger than a given input

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:
Row Limit 1
Row Limit 2
After:
Screenshot 2023-10-09 120420
Screenshot 2023-10-09 120447

PR #26151, opened 2023-11-30
fix(chart): Set max row limit + removed the option to use an empty row limit value

CorbinBullard:
### SUMMARY

- Can no longer clear/remove the row limit
- New validator that checks if a value is larger than a given input
- **Removed ValidateNonEmpty() validator that was causing charts before the update not to render**

Before:
Row Limit 1
Row Limit 2
After:
Screenshot 2023-10-09 120420
Screenshot 2023-10-09 120447


CorbinBullard:
Here is the original PR that was reverted [https://github.com/apache/superset/pull/25579](https://github.com/apache/superset/pull/25579)

michael-s-molina (changes requested):
The max value should come from [SQL_MAX_ROW](https://github.com/apache/superset/blob/56fa2b29d398142ab4171615737bb185ae82df94/superset/config.py#L898) and not be a fixed 100,000.

superset-frontend/packages/superset-ui-core/test/validator/validateMaxValue.test.ts
```diff
@@ -0,0 +1,38 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { validateMaxValue } from '@superset-ui/core';
+import './setup';
+
+describe('validateInteger()', () => {
```
michael-s-molina, line 23: Can you avoid `describe` to adhere to our [testing best practices](https://github.com/apache/superset/wiki/Testing-Guidelines-and-Best-Practices#avoid-nesting-when-youre-testing)? You can just use `test` instead.

superset-frontend/packages/superset-ui-chart-controls/src/utils/getBootstrapData.ts
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { BootstrapData } from '../types';
+import { DEFAULT_BOOTSTRAP_DATA } from '../constants';
+
+export default function getBootstrapData(): BootstrapData {
```
michael-s-molina, line 23: You don't need to replicate the bootstrap logic. You can define `row_limit` as:

``` typescript
const row_limit: SharedControlConfig<'SelectControl'> = {
   validators: [legacyValidateInteger, (v, state) => validateMaxValue(v, state?.maxValue)],
   mapStateToProps: state => ({ maxValue: state?.common.conf.SQL_MAX_ROW }),
   ...
}
```

You can remove all changes related to bootstrapping the configs.

michael-s-molina (changes requested):

superset-frontend/packages/superset-ui-chart-controls/src/types.ts
```diff
@@ -557,3 +557,10 @@ export type ControlFormItemSpec<T extends ControlType = ControlType> = {
       defaultValue?: Currency;
     }
   : {});
+
+interface CommonBootstrapData {
+  conf: JsonObject;
+}
+export interface BootstrapData {
+  common: CommonBootstrapData;
+}
```
michael-s-molina, line 566:
```suggestion
```

michael-s-molina (changes requested):

michael-s-molina (approved):
LGTM

Stefan781:
Using Superset 3.1, I am limited to 100 000 rows even if I set both SQL_MAX_ROW and ROW_LIMIT. Is that the expected behavior? How can I set a higher limit?

Stefan781:
I debugged further. For that, I had to use an already existing chart that had a limit of more than 100 000 prior to the upgrade to 3.1.
Changing the SQL_MAX_ROW to 5000 results in a proper _LIMIT 5000;_ in the query. Bringing it back up updates the value accordingly, up until the chart's original row limit. However, the check is still kept to 100 000 when creating the chart. This points to an issue with those changes. Could somebody confirm?

PR #26156, opened 2023-12-01
fix(sqllab): flaky json explore modal due to over-rendering

justinpark:
### SUMMARY
The queries state contains the status and results of running and completed queries, making it fragile. Since the SouthPane observes the entire queries state, it often causes over-rendering. This means that the current result panel will trigger rendering at each interval of checking pending queries.

This causes the visibility of the JSON modal to be flaky because the modal disappears when it is re-rendered.
This commit reduces the observation of queries properties to resolve the over-rendering issue. (As a result, the completed result panel component will not be disturbed by the status updates of other pending queries.)

In addition, this commit enhances the queriesLastUpdate timestamp by incrementally updating it only when its query result remains unchanged.

Before:

https://github.com/apache/superset/assets/1392866/8d81ba78-7142-438c-8c5c-b5049ef6ea94

Screenshot 2023-11-29 at 10 58 17 AM

After:

Issue is gone

https://github.com/apache/superset/assets/1392866/c2023683-47f1-4a45-97c0-c14e4d7b6469

Screenshot 2023-11-30 at 8 32 44 PM

### TESTING INSTRUCTIONS

Go to SQL Lab and run a query contains a json blob
Select a table in order to execute a table preview query
Click `Results` tab before the table preview query ends and then click the JSON column to open the JSON modal
Check the JSON modal persists while the async query status is updating

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/QueryAutoRefresh/index.tsx
+++ b/superset-frontend/src/SqlLab/components/QueryAutoRefresh/index.tsx
@@ -45,7 +45,7 @@ export interface QueryAutoRefreshProps {
 
 // returns true if the Query.state matches one of the specifc values indicating the query is still processing on server
 export const isQueryRunning = (q: Query): boolean =>
-  runningQueryStateList.includes(q?.state);
+  runningQueryStateList.includes(q?.state) && !q?.resultsKey;
 
 // returns true if at least one query is running and within the max age to poll timeframe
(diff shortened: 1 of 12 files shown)


michael-s-molina (approved):
LGTM

PR #26157, opened 2023-12-01
fix(init-job): Fix envFrom for init job in helm chart

sumagoudb:
It breaks when I use envFromSecrets in helm value as the quotes are missing for the init job.


dpgaspar (commented):
Thank you!
LGTM, you have to bump the chart version also

sumagoudb:
> Thank you! LGTM, you have to bump the chart version also

Done, can you merge?

helm/superset/Chart.yaml
```diff
@@ -29,7 +29,7 @@ maintainers:
   - name: craig-rueda
     email: craig@craigrueda.com
     url: https://github.com/craig-rueda
-version: 0.11.1
+version: 0.11.2
```
craig-rueda, line 32: The README will need to be updated with the latest version as well

sumagoudb: Done

dpgaspar (approved):

PR #22458, opened 2022-12-19
feat: update time comparison choices (again)

john-bodley:
### SUMMARY

Similar to [#11294](https://github.com/apache/superset/pull/17968), this PR adds Y4Y time comparison choices to make it faster to compare with pre-pandemic times.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #22150, opened 2022-11-17
chore: Localization of Superset pt. 1

artemonsh:
### SUMMARY
The localization of some components and especially charts is not complete yet. This PR is the effort to achieve the full localization of Superset. Additionally, it is a real pain to upgrade Superset as one should solve tons of merge conflicts, if they wrapped everything in `t()` and `__()` manually, while it is not reflected in the official repo :/

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26173, opened 2023-12-05
feat: update time comparison choices (again)

john-bodley:
### SUMMARY

Similar to https://github.com/apache/superset/pull/22458, this PR adds Y5Y time comparison choices to make it faster to compare with pre-pandemic times.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/sections/advancedAnalytics.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/sections/advancedAnalytics.tsx
@@ -121,6 +121,10 @@ export const advancedAnalyticsControls: ControlPanelSectionConfig = {
             ['2 years ago', t('2 years ago')],
             ['156 weeks ago', t('156 weeks ago')],
             ['3 years ago', t('3 years ago')],
+            ['208 weeks ago', t('208 weeks ago')],
+            ['4 years ago', t('4 years ago')],
+            ['261 weeks ago', t('261 weeks ago')],
+            ['5 years ago', t('5 years ago')],
           ],
           description: t(
             'Overlay one or more timeseries from a ' +
--- a/superset-frontend/plugins/legacy-plugin-chart-partition/src/controlPanel.tsx
+++ b/superset-frontend/plugins/legacy-plugin-chart-partition/src/controlPanel.tsx
@@ -322,6 +322,10 @@ const config: ControlPanelConfig = {
                 ['2 years', t('2 years')],
                 ['156 weeks', t('156 weeks')],
                 ['3 years', t('3 years')],
+                ['208 weeks ago', t('208 weeks ago')],
+                ['4 years ago', t('4 years ago')],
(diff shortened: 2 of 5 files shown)


superset-frontend/packages/superset-ui-chart-controls/src/sections/advancedAnalytics.tsx
```diff
@@ -121,6 +121,10 @@ export const advancedAnalyticsControls: ControlPanelSectionConfig = {
             ['2 years ago', t('2 years ago')],
             ['156 weeks ago', t('156 weeks ago')],
             ['3 years ago', t('3 years ago')],
+            ['208 weeks ago', t('208 weeks ago')],
```
john-bodley, line 124: It seems like the Y4Y comparison was removed in https://github.com/apache/superset/pull/22150, though it's not evident—per the PR description and/or comments—as to why, hence I re-added them.

superset-frontend/plugins/legacy-plugin-chart-partition/src/controlPanel.tsx
```diff
@@ -322,6 +322,10 @@ const config: ControlPanelConfig = {
                 ['2 years', t('2 years')],
                 ['156 weeks', t('156 weeks')],
                 ['3 years', t('3 years')],
+                ['208 weeks ago', t('208 weeks ago')],
```
michael-s-molina, line 325: @john-bodley Is there are reason to keep the time period in weeks when they match the year? Why not use weeks only when the number of weeks is not a multiple of 52?

justinpark: As @michael-s-molina mentioned, the custom time shift option can vary in different scenarios. To ensure consistency, it is advisable to override the custom option by utilizing the chartControlRegistry on the vendor side.

justinpark: > john-bodley Is there are reason to keep the time period in weeks when they match the year? Why not use weeks only when the number of weeks is not a multiple of 52?

In order to effectively analyze daily and weekly data and account for seasonal trends, it is essential to incorporate the weekly offset.

FYI:
Certain industries or businesses may experience specific patterns or trends on a weekly basis. For example, in the hospitality industry, weekends may have higher demand compared to weekdays. By using weeks as the time period, analysts can effectively capture these patterns and make more accurate predictions or recommendations.

michael-s-molina: Thanks @justinpark for the additional context. @john-bodley I wonder if we should make these configurable somehow.

john-bodley:
Closing in favor of implementing an internal custom override. It seems like Y4Y, Y5Y, etc. comparisons are very much organization dependent.

Issue #25107, opened 2023-08-28
[SIP-98A] Primer on managing SQLAlchemy sessions


john-bodley:
## [SIP-98A] Primer on managing SQLAlchemy sessions

This SIP is part of the [[SIP-98] Proposal for correctly handling business logic](https://github.com/apache/superset/issues/25048) series. Specifically, it serves as a primer—reiterating best practices—for working with SQLAlchemy sessions as these were typically mismanaged. The topics outlined here are highly coupled with [[SIP-98B] Proposal for (re)defining a “unit of work”](...) as they both relate to managing database transactions.

### Session Management 

The [SQLAlchemy session](https://docs.sqlalchemy.org/en/20/orm/session_basics.html)—which establishes all conversations with the database—is managed by [Flask-SQLAlchemy](https://github.com/pallets-eco/flask-sqlalchemy) which also handles cleaning up connections and sessions after each [request](https://github.com/pallets-eco/flask-sqlalchemy/blob/564a96f2ee07819aac2dcca7d9eb58dcfeb1765f/src/flask_sqlalchemy/extension.py#L379-L386). 

The preconfigured scoped session (called `db.session`) should be used for the lifetime of a request to ensure there is only one “virtual” transaction—guaranteeing that the “unit of work” is indeed atomic ([SIP-98B](...)).

The SQLAlchemy [Managing Transactions](https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#managing-transactions) documentation states:

> The Session tracks the state of a single “virtual" transaction at a time. This “virtual" transaction is created automatically when needed, or can alternatively be started using the `Session.begin()` method. `Session.commit()` ends the transaction.

The transaction is typically framed as follows, 

```python
try:
    db.session.add(Dashboard(slug="foo"))
    db.session.commit()
except SQLAlchemyError:
    db.session.rollback()
    raise
```

though, via leveraging a context manager, the `Session.begin()` method serves as a demarcated transaction and provides the same sequence of operations:

```python
with db.session.begin():
    db.session.add(Dashboard(slug="foo"))
```

#### Session States

There are [five states](https://docs.sqlalchemy.org/en/20/orm/session_state_management.html) which an object can have within a session:

- Transient
- Pending
- Persisted
- Deleted
- Detached

The session tracks objects that are either new, dirty, or deleted.

[...]

Issue #10530, opened 2020-08-06
celery SQLAlchemy errors in celery tasks that use sqlalchemy pools


JasonD28:
While running multiple `alert.run_query` celery tasks, some tasks fail with a `sqlalchemy.exc.NoSuchColumnError` error here https://github.com/apache/incubator-superset/blob/ece91928a9339190163c0bc72b96e51217a90d1e/superset/tasks/schedules.py#L544 on the first few iterations of the celery beat schedule after its initialization. These errors eventually stop occurring if the celery worker and the beat schedule are left to run for a few iterations without stopping.  Tests have shown that the errors stop occurring on average after around 2 iterations of the beat schedule.

### Expected results

Alerting celery tasks run without error

### Actual results

```
Traceback (most recent call last):
  File "/srv/yaps/mounts/superset/superset_build/bbe899d629f9d61e5d2a8ffdccbcc57724a557ea-e230e418e5960bc45d735cb21d806661/celery.runfiles/__main__/thirdparty/superset_python3/pip/celery/celery-cpython-38/lib/celery/app/trace.py", line 412, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/srv/yaps/mounts/superset/superset_build/bbe899d629f9d61e5d2a8ffdccbcc57724a557ea-e230e418e5960bc45d735cb21d806661/celery.runfiles/__main__/thirdparty/superset_python3/superset-cpython-38/lib/superset/app.py", line 115, in __call__
    return task_base.__call__(self, *args, **kwargs)
  File "/srv/yaps/mounts/superset/superset_build/bbe899d629f9d61e5d2a8ffdccbcc57724a557ea-e230e418e5960bc45d735cb21d806661/celery.runfiles/__main__/thirdparty/superset_python3/pip/celery/celery-cpython-38/lib/celery/app/trace.py", line 704, in __protected_call__
    return self.run(*args, **kwargs)
  File "/srv/yaps/mounts/superset/superset_build/bbe899d629f9d61e5d2a8ffdccbcc57724a557ea-e230e418e5960bc45d735cb21d806661/celery.runfiles/__main__/thirdparty/superset_python3/superset-cpython-38/lib/superset/tasks/schedules.py", line 562, in schedule_alert_query
    if run_alert_query(
[...]

PR #26186, opened 2023-12-05
refactor: Ensure Celery leverages the Flask-SQLAlchemy session

john-bodley:
### SUMMARY

As part of SIP-99 (specifically [SIP-99A](https://github.com/apache/superset/issues/25107) and [SIP-99B](https://github.com/apache/superset/issues/25108)) in order to ensure a consistent "unit of work"—via a single atomic unit—all operations should be associated with the same Flask-SQLAlchemy session.

The Flask-SQLAlchemy extension provides a scoped session (on a per request basis) with the necessary oversight, i.e., the session is closed after the request is complete which aids with connection pool management. 

Historically Celery tasks have defined their own scoped session (with the option to use connection pooling) which needed to managed independently which added unnecessary code bloat and complexity and likely violated the "unit of work" construct if operations were leveraging both the Flask-SQLAlchemy and Celery sessions. Per [this](https://gist.github.com/twolfson/a1b329e9353f9b575131) post it seems like Celery can piggyback off of the Flask-SQLAlchemy session so long as setup/teardown is handled correctly.  There is actually an example of this in the official Flask documentation per the [Celery with Flask](https://flask.palletsprojects.com/en/2.3.x/patterns/celery/) document where it references the use of the Flask-SQLAlchemy session (`db.session`).

This PR removes the need for a Celery specific session which is a step towards the goal of having all database operations (outside of the Alembic migrations) handled by the global `db.session` which is a necessary requirement in order for us to achieve the goal of an atomic unit of work.

The one wrinkle with this approach is https://github.com/apache/superset/pull/10819 which explicitly leveraged a `NullPool` (as opposed to `QueuePool`—if defined via the `SQLALCHEMY_ENGINE_OPTIONS ` configuration) to avoid using same session across multiple celery workers, however since we're tearing down the session a per task basis where, per [here](https://stackoverflow.com/questions/12044776/how-to-use-flask-sqlalchemy-in-a-celery-task/14146403#14146403), it states,

> Flask SQLAlchemy will automatically create new sessions for you from scoped session factory, given that we are maintaining the same app context, this ensures tasks have a fresh session (e.g. session errors won't propagate across tasks).

i.e., we should no longer be experiencing the connection bleeding issue and thus can leverage the efficiency of connection pooling which standardizing the code.

### TESTING INSTRUCTIONS

CI. Additionally I wasn't able to repo the issue mentioned in https://github.com/apache/superset/issues/10530.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/25107
- [ ] Required feature flags:
- [ ] Changes UI


superset/sql_lab.py
```diff
@@ -176,30 +170,27 @@ def get_sql_results(  # pylint: disable=too-many-arguments
...
-                    start_time=start_time,
-                    expand_data=expand_data,
-                    log_params=log_params,
-                )
-            except Exception as ex:  # pylint: disable=broad-except
-                logger.debug("Query %d: %s", query_id, ex)
-                stats_logger.incr("error_sqllab_unhandled")
-                query = get_query(query_id, session)
-                return handle_query_error(ex, query, session)
+    with override_user(security_manager.find_user(username)):
```
john-bodley, line 173: Pretty much the same code as previously without the outer `with` block.

superset/tasks/cache.py
```diff
@@ -130,28 +123,24 @@ def __init__(self, top_n: int = 5, since: str = "7 days ago") -> None:
         self.since = parse_human_datetime(since) if since else None
 
     def get_payloads(self) -> list[dict[str, int]]:
-        payloads = []
-        session = db.create_scoped_session()
+        records = (
```
john-bodley, line 126: Pretty much the same code as previously without the `try` block.

superset/tasks/cache.py
```diff
@@ -178,48 +167,44 @@ def __init__(self, tags: Optional[list[str]] = None) -> None:
...
-
-            # add dashboards that are tagged
-            tagged_objects = (
-                session.query(TaggedObject)
-                .filter(
-                    and_(
-                        TaggedObject.object_type == "dashboard",
-                        TaggedObject.tag_id.in_(tag_ids),
-                    )
+        tags = db.session.query(Tag).filter(Tag.name.in_(self.tags)).all()
```
john-bodley, line 170: Pretty much the same code as previously without the `try` block.

superset/tasks/scheduler.py
```diff
@@ -46,35 +45,32 @@ def scheduler() -> None:
...
-        triggered_at = (
-            datetime.fromisoformat(scheduler.request.expires)
-            - app.config["CELERY_BEAT_SCHEDULER_EXPIRES"]
-            if scheduler.request.expires
-            else datetime.utcnow()
-        )
-        for active_schedule in active_schedules:
-            for schedule in cron_schedule_window(
-                triggered_at, active_schedule.crontab, active_schedule.timezone
+    active_schedules = ReportScheduleDAO.find_active()
```
john-bodley, line 48: Pretty much the same code as previously without the outer `with` block.

john-bodley:
ping @michael-s-molina @villebro

michael-s-molina:
@john-bodley I think it would be great to label this PR with v4.0 and merge it during the breaking window to reuse the test/stabilization efforts that will occur during that period.

john-bodley:
Per,

> @john-bodley I think it would be great to label this PR with v4.0 and merge it during the breaking window to reuse the test/stabilization efforts that will occur during that period.

as discussed with @michael-s-molina, though this is technically a non-breaking change, it seems prudent (from a safety perspective) to hold off merging this until the v4.0 breaking window.

superset/tasks/celery_app.py
```diff
@@ -43,3 +43,27 @@ def reset_db_connection_pool(**kwargs: Any) -> None:  # pylint: disable=unused-a
...
+@task_postrun.connect
+def teardown(  # pylint: disable=unused-argument
+    retval: Any,
+    *args: Any,
+    **kwargs: Any,
+) -> None:
+    """
+    After each Celery task teardown the Flask-SQLAlchemy session.
+
+    Note for non eagar requests Flask-SQLAlchemy will perform the teardown.
```
michael-s-molina, line 57:
```suggestion
    Note for non eager requests Flask-SQLAlchemy will perform the teardown.
```

michael-s-molina (approved):
LGTM. I'm assuming that removing unnecessary `commit` operations is not in the scope of this PR.

Issue #26153, opened 2023-11-30
Pyarrow version 12.0.0, having a critical vulnerability, upgrade to new pyarrow version is causing a Error in superset


nagarajmmu:
#### How to reproduce the bug

1. Superset 3.0.2,  is having pyarrow version 12.0.0.
2. But  pyarrow version 12.0.0 is having a critical vulnerability.
3. Upgrade the pyarrow to 14.0.1 to resolve the vulnerability is successful in supserset.
4. Once pyarrow is upgraded, start the superset.
5. Superset failed to start with below error.

ERROR:

                  
    [notice] A new release of pip is available: 23.0.1 -> 23.3.1
    [notice] To update, run: pip install --upgrade pip
    Upgrading DB schema...
    Traceback (most recent call last):
      File "/usr/local/lib/python3.9/site-packages/pkg_resources/__init__.py", line 568, in _build_master
        ws.require(__requires__)
      File "/usr/local/lib/python3.9/site-packages/pkg_resources/__init__.py", line 886, in require
        needed = self.resolve(parse_requirements(requirements))
      File "/usr/local/lib/python3.9/site-packages/pkg_resources/__init__.py", line 777, in resolve
        raise VersionConflict(dist, req).with_context(dependent_req)
    pkg_resources.ContextualVersionConflict: (pyarrow 14.0.1 (/usr/local/lib/python3.9/site-packages), Requirement.parse('pyarrow<13,>=12.0.0'), {'apache-superset'})
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/usr/local/bin/superset", line 33, in <module>
        sys.exit(load_entry_point('apache-superset', 'console_scripts', 'superset')())
      File "/usr/local/bin/superset", line 25, in importlib_load_entry_point
        return next(matches).load()
      File "/usr/local/lib/python3.9/importlib/metadata.py", line 86, in load
        module = import_module(match.group('module'))
      File "/usr/local/lib/python3.9/importlib/__init__.py", line 127, in import_module
        return _bootstrap._gcd_import(name[level:], package, level)
      File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
      File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
      File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
[...]

PR #26187, opened 2023-12-05
fix: bump pyarrow constraints (CVE-2023-47248)

cwegener:
PyArrow < 14.0.1 is vulnerable to RCE when using IPC, Flight or Parquet
from untrusted sources.

Superset SQLLab does so.

So we need to care about this vulnerability.


dpgaspar (approved):

cwegener:
> Fixes #26153

Only partially. :wink: as explained in my comment in #26153

PR #26188, opened 2023-12-05
chore: Lower giveup log level for retried functions to warning

jfrag1:
### SUMMARY

The goal of this PR is to eliminate unnecessary error logs.  The `retry_call` function uses the `backoff` lib to decorate functions we want to retry.  After the retry limit is reached, the `backoff` decorator generates a log saying that it's giving up.  This log defaults to the error level, which isn't necessary here since the exception raised by the wrapped function gets bubbled up and logged at the appropriate level already.

This PR solves the issue by lowering the log from the `backoff` decorator to a warning

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/retries.py
+++ b/superset/utils/retries.py
@@ -16,7 +16,6 @@
 # under the License.
 
 import logging
-
 from collections.abc import Generator
 from typing import Any, Callable, Optional
 
--- a/superset/utils/retries.py
+++ b/superset/utils/retries.py
@@ -15,8 +15,9 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from collections.abc import Generator
 import logging
+
+from collections.abc import Generator
 from typing import Any, Callable, Optional
 
 import backoff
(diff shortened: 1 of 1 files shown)


superset/utils/retries.py
```diff
@@ -26,13 +27,15 @@ def retry_call(
     *args: Any,
     strategy: Callable[..., Generator[int, None, None]] = backoff.constant,
     exception: type[Exception] = Exception,
+    giveup_log_level: int = logging.WARNING,
```
eschutho, line 30: maybe type this as optional?

eschutho (approved):
Approved, with a suggested small nit on typing.

superset/utils/retries.py
jfrag1: I think just `int` is more precise here, since the default `logging.WARNING` is an int, and it wouldn't make sense to pass `giveup_log_level=None`

PR #26189, opened 2023-12-05
fix(dashboard): title formatting

nytai:
### SUMMARY

Titles with special characters are not rendered correctly. This PR ensures that special characters are handled correctly and titles match the user input 

### TESTING INSTRUCTIONS

- Dashboards load and titles are rendered correctly

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/DynamicEditableTitle/index.tsx
+++ b/superset-frontend/src/components/DynamicEditableTitle/index.tsx
@@ -113,10 +113,7 @@ export const DynamicEditableTitle = ({
   // then we can measure the width of that span to resize the input element
   useLayoutEffect(() => {
     if (sizerRef?.current) {
-      sizerRef.current.innerHTML = (currentTitle || placeholder).replace(
-        /\s/g,
-        '&nbsp;',
-      );
+      sizerRef.current.textContent = currentTitle || placeholder;
     }
   }, [currentTitle, placeholder, sizerRef]);
 


superset-frontend/src/components/DynamicEditableTitle/index.tsx
```diff
@@ -113,10 +113,12 @@ export const DynamicEditableTitle = ({
   // then we can measure the width of that span to resize the input element
   useLayoutEffect(() => {
     if (sizerRef?.current) {
-      sizerRef.current.innerHTML = (currentTitle || placeholder).replace(
-        /\s/g,
-        '&nbsp;',
-      );
+      sizerRef.current.innerHTML = (currentTitle || placeholder)
```
john-bodley, line 116: @nytai are there libraries we can leverage for sanitizing the HTML?

nytai: do we want to sanitize (ie, remove) html or just escape it? I think the purpose of this code is to render the title to the DOM so we can measure how much space it takes up, so we probably want to escape it. 

Looking at an escape-html lib, it's pretty basic. Eg https://github.com/component/escape-html/blob/master/index.js 

Happy to update the PR to use this lib if we think it's best, but... you know the joke about node modules 

image

john-bodley: Sorry I meant escape rather than sanitize. It [seems](https://github.com/search?q=repo%3Aapache%2Fsuperset%20escape-html&type=code) like we're already installing the `escape-html` library and thus maybe we should just use that directly.

nytai: That's the websocket sidecar app (for async queries), but sure we can use that.

superset-frontend/src/components/DynamicEditableTitle/index.tsx
```diff
@@ -113,10 +114,7 @@ export const DynamicEditableTitle = ({
   // then we can measure the width of that span to resize the input element
   useLayoutEffect(() => {
     if (sizerRef?.current) {
-      sizerRef.current.innerHTML = (currentTitle || placeholder).replace(
-        /\s/g,
-        '&nbsp;',
-      );
+      sizerRef.current.innerHTML = escapeHTML(currentTitle || placeholder);
```
dpgaspar, line 117: This is a good improvement, yet it's a breaking change, minimal we need a line on `UPDATING.md`

rusackas (approved):
Well now it's nice and simple! :D

Issue #26198, opened 2023-12-06
Chart filter options are not populated for datasets with normalize column enabled


Vitor-Avila:
For engines that support column de-normalization, it's possible to opt-out by enabling **column normalization**. This is enabled by default to all datasets created prior to the implementation of column denormalization. For more context, check below PRs:
* Column de-normalization implementation: https://github.com/apache/superset/pull/24471
* Opt-in control: https://github.com/apache/superset/pull/24982

Chart filters are not displaying options for datasets that have **column normalization** enabled.

#### How to reproduce the bug

1. Create a dataset powered by an engine that supports column de-normalization (such as Snowflake). Note that:
 a. All columns are **uppercase**.
 b. Column normalization is disabled (under the **SETTINGS** tab).
2. Modify the dataset, and enable column normalization.
3. Save changes.
4. Modify the dataset again, and sync columns. Note that all columns are now **lowercase**.
5. Save changes.
6. Create a new chart using this dataset, and drop any column in the **FILTERS** section.

### Expected results
Filter values should be populated in the dropdown. 

### Actual results
The dropdown is empty. The network request to fetch the column values return below `400` error:
```
Column name {column_name} does not exist
```

### Environment

* browser type and version: Chrome version `119.0.6045.159`.
* superset version: `0.0.0-dev` (running `master`).
* python version: `3.9.0`
* node.js version: `v16.14.2`
* any feature flags active: not relevant to reproduce the issue

### Checklist
- [x] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [x] I have reproduced the issue with at least the latest released version of superset.
- [x] I have checked the issue tracker for the same issue and I haven't found one similar.

PR #26199, opened 2023-12-06
fix(chart-filter): Avoid column denormalization if not enabled

Vitor-Avila:
### SUMMARY
Avoid de-normalizing column names in case the engine supports it **but column normalization is enabled in the dataset level**. This is enabled by default to datasets created prior to this feature, to make sure that syncing columns wouldn't break existing charts/etc.

**Before**

https://github.com/apache/superset/assets/96086495/f30e8688-81c8-4019-b10d-4e4aaef932f6

**After**

https://github.com/apache/superset/assets/96086495/31909ae1-91b7-4710-9c9e-3f531a344821

### TESTING INSTRUCTIONS
1. Create a dataset powered by an engine that supports column de-normalization (such as Snowflake). Note that:
 a. All columns are **uppercase**.
 b. Column normalization is disabled (under the **SETTINGS** tab).
2. Modify the dataset, and enable column normalization.
3. Save changes.
4. Modify the dataset again, and sync columns. Note that all columns are now **lowercase**.
5. Save changes.
6. Create a new chart using this dataset, and drop any column in the FILTERS section.
7. Validate the filter is showing available options in the dropdown.

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes https://github.com/apache/superset/issues/26198
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -1340,14 +1340,18 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
             )
         return and_(*l)
 
-    def values_for_column(self, column_name: str, limit: int = 10000) -> list[Any]:
-        # always denormalize column name before querying for values
+    def values_for_column(
+        self, column_name: str, limit: int = 10000, denormalize_column: bool = False
+    ) -> list[Any]:
+        # denormalize column name before querying for values unless disabled in the dataset
         db_dialect = self.database.get_dialect()
-        denormalized_col_name = self.database.db_engine_spec.denormalize_name(
-            db_dialect, column_name
+        column_name_ = (
+            self.database.db_engine_spec.denormalize_name(db_dialect, column_name)
+            if denormalize_column
+            else column_name
         )
         cols = {col.column_name: col for col in self.columns}
-        target_col = cols[denormalized_col_name]
+        target_col = cols[column_name_]
         tp = self.get_template_processor()
         tbl, cte = self.get_from_clause(tp)
 
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -792,7 +792,7 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
         self,
         template_processor: Optional[  # pylint: disable=unused-argument
             BaseTemplateProcessor
-        ] = None,  # pylint: disable=unused-argument
+        ] = None,
     ) -> TextClause:
         return self.fetch_values_predicate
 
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
This makes sense. The original naming could be better and definitely needs a cleanup, seems like we're using "denormalize" for "don't normalize". Ideally we'd have a boolean feature called something like `needs_normalization`, and we'd do normalization/denormalization as needed, otherwise we wouldn't touch column names.

Vitor-Avila:
thanks @betodealmeida. @hughhhh @villebro I know you've worked similarly with this feature. Do you have any concerns with this PR? thank you!

superset/models/helpers.py
```diff
@@ -1340,14 +1340,19 @@ def get_time_filter(  # pylint: disable=too-many-arguments
             )
         return and_(*l)
 
-    def values_for_column(self, column_name: str, limit: int = 10000) -> list[Any]:
-        # always denormalize column name before querying for values
+    def values_for_column(
```
sadpandajoe, line 1343: Do our unit tests already have cases where it covers if `denormalize_column` returns the correct values for both true and false already? If not, I would add tests to confirm that the correct things are returned for when the flag is true since we're defaulting to false.

villebro (approved):
LGTM, thanks for the fix. My apologies for contributing to the confusing naming here, it seems I hadn't thought it fully through (is it really in fact denormalizing, or not normalizing etc)..

michael-s-molina:
@Vitor-Avila Tip: If you include the text "Fixes: https://github.com/apache/superset/issues/26198" in the PR description, when the PR is merged, the issue is automatically closed. This only works for the description, not comments.

michael-s-molina:
I don't know why but when it's part of a checkbox (`Has associated issue` in the template) it does not work. To see if the link worked, you can check the original issue for a message saying that the issue will be closed by the PR.

Vitor-Avila:
@michael-s-molina thanks for the tip! I remember a previous PR I created did automatically closed the bug, but I never understood why. I'll make sure to include that out of the checkbox next time 🙌

superset/models/helpers.py
Vitor-Avila: @sadpandajoe @villebro I just added some basic tests to https://github.com/apache/superset/pull/26220 (since this one got merged). It doesn't test the logic implemented in the DB engine level to denormalize a column (I believe this is DB-specific and would require a more complex setup), but should be at least testing the business logic.

PR #26200, opened 2023-12-06
refactor: Ensure Flask framework leverages the Flask-SQLAlchemy session (Phase I)

john-bodley:
### SUMMARY

Somewhat related to https://github.com/apache/superset/pull/26186 (though for non-Celery related workflows) this PR ensures that the remainder of the codebase (excluding tests and Alembic migrations) leverage the Flask-SQLAlchemy session (`db.session`).

This refactor helps to reduce the code complexity and ensures we're not generating any rouge sessions which are often not explicitly closed and can lead to connection pooling issues.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/25107
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/scripts/benchmark_migration.py
+++ b/scripts/benchmark_migration.py
@@ -142,8 +142,6 @@ def main(
     filepath: str, limit: int = 1000, force: bool = False, no_auto_cleanup: bool = False
 ) -> None:
     auto_cleanup = not no_auto_cleanup
-    session = db.session()
-
     print(f"Importing migration script: {filepath}")
     module = import_migration_script(Path(filepath))
 
@@ -174,10 +172,9 @@ def main(
     models = find_models(module)
     model_rows: dict[type[Model], int] = {}
     for model in models:
-        rows = session.query(model).count()
+        rows = db.session.query(model).count()
         print(f"- {model.__name__} ({rows} rows in table {model.__tablename__})")
         model_rows[model] = rows
-    session.close()
 
     print("Benchmarking migration")
     results: dict[str, float] = {}
@@ -199,16 +196,16 @@ def main(
                 print(f"- Adding {missing} entities to the {model.__name__} model")
                 bar = ChargingBar("Processing", max=missing)
                 try:
-                    for entity in add_sample_rows(session, model, missing):
+                    for entity in add_sample_rows(model, missing):
                         entities.append(entity)
                         bar.next()
                 except Exception:
-                    session.rollback()
+                    db.session.rollback()
                     raise
                 bar.finish()
                 model_rows[model] = min_entities
-                session.add_all(entities)
-                session.commit()
+                db.session.add_all(entities)
+                db.session.commit()
 
                 if auto_cleanup:
                     new_models[model].extend(entities)
@@ -227,10 +224,10 @@ def main(
         print("Cleaning up DB")
         # delete in reverse order of creation to handle relationships
         for model, entities in list(new_models.items())[::-1]:
-            session.query(model).filter(
+            db.session.query(model).filter(
                 model.id.in_(entity.id for entity in entities)
             ).delete(synchronize_session=False)
-        session.commit()
+        db.session.commit()
 
     if current_revision != revision and not force:
         click.confirm(f"\nRevert DB to {revision}?", abort=True)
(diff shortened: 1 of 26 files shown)


superset/utils/log.py
```diff
@@ -152,8 +153,7 @@ def log_with_context(  # pylint: disable=too-many-locals
         # need to add them back before logging to capture user_id
         if user_id is None:
             try:
-                session = current_app.appbuilder.get_session
-                session.add(g.user)
+                db.session.add(g.user)
```
john-bodley, line 156: Same session as `current_app.appbuilder.get_session`. Consistency is key here.

tests/unit_tests/datasource/dao_tests.py
```diff
@@ -106,7 +106,6 @@ def test_get_datasource_sqlatable(session_with_data: Session) -> None:
     result = DatasourceDAO.get_datasource(
         datasource_type=DatasourceType.TABLE,
         datasource_id=1,
-        session=session_with_data,
```
john-bodley, line 109: The `db.session` is mocked [here](https://github.com/apache/superset/blob/master/tests/unit_tests/conftest.py#L65) and is included as a fixture to the test.

superset/daos/datasource.py
```diff
@@ -45,15 +44,14 @@ class DatasourceDAO(BaseDAO[Datasource]):
     @classmethod
     def get_datasource(
         cls,
-        session: Session,
```
john-bodley, line 48: It is not apparent to me why the `session` was included here. There are unit tests which use a session other than `db.session` but, per [here](https://github.com/apache/superset/blob/master/tests/unit_tests/conftest.py#L65), the global `db.session` is mocked to use said session.

superset/models/dashboard.py
```diff
@@ -71,38 +71,33 @@
...
-            position_json=template.position_json,
-            description=template.description,
-            css=template.css,
-            json_metadata=template.json_metadata,
-            slices=template.slices,
-            owners=[new_user],
-        )
-        session.add(dashboard)
+    session = sqla.inspect(target).session
+    new_user = session.query(User).filter_by(id=target.id).first()
```
john-bodley, line 80: Same logic without the `try` block.

john-bodley:
ping @michael-s-molina @villebro

michael-s-molina:
@john-bodley I think it would be great to label this PR with v4.0 and merge it during the breaking window to reuse the test/stabilization efforts that will occur during that period.

scripts/benchmark_migration.py
```diff
@@ -174,10 +172,9 @@ def main(
     models = find_models(module)
     model_rows: dict[type[Model], int] = {}
     for model in models:
-        rows = session.query(model).count()
+        rows = db.session.query(model).count()
         print(f"- {model.__name__} ({rows} rows in table {model.__tablename__})")
         model_rows[model] = rows
-    session.close()
```
john-bodley, line 180: There is no need to explicitly close the session as this is handled by Flask-SQLAlchemy when the session is [torn down](https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/extension.py#L441-L448)—be that at the end of request or when a script/shell terminates.

john-bodley:
Regarding, 

> @john-bodley I think it would be great to label this PR with v4.0 and merge it during the breaking window to reuse the test/stabilization efforts that will occur during that period.

as discussed with @michael-s-molina, though this is technically a non-breaking change, it seems prudent (from a safety perspective) to hold off merging this until the v4.0 breaking window.

john-bodley:
@michael-s-molina would you mind reviewing this PR?

michael-s-molina (approved):
LGTM

mazeem-arbisoft:
Hi @john-bodley, we have been facing an issue where overwriting a dashboard doesn't go well sometimes. I looked into import code. A bit curious about the following code in import dashboard's import dataset part:
image

- Why don't we overwrite datasets here? 
- Apparently from source code it looks like this line is the issue. I think we should pass the overwrite flag received as a part of dashboard import request. Apparently, the import code never goes into the sync flow in `import_dataset` method if dataset already exists.

sfirke:
@mazeem-arbisoft I think John Bodley has stopped working on Superset. You might be more likely to get attention if you open a new Issue or Discussion on GitHub and link back to this one for context.

PR #26214, opened 2023-12-07
chore: Reduce and restrict innerHTML use

rusackas:
### SUMMARY

innerHTML is gross... this PR adds linting rules to restrict use of this method, and replaces or annotates existing usage.

### TESTING INSTRUCTIONS

There are a few plugins that were touched that need testing:
* Chord
* Heatmap
* NVD3

Dynamic plugins have their web URL bundle sanitized... this should work, but if not, very few people will be affected since this is an Alpha feature.

Icon file paths are also being sanitized - we should check for these in general, though it _seems_ things should work.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/legacy-plugin-chart-heatmap/src/Heatmap.js
```diff
@@ -211,8 +211,7 @@ function Heatmap(element, props) {
     return d3.scale.ordinal().domain(domain).range(d3.range(domain.length));
   }
 
-  // eslint-disable-next-line no-param-reassign
-  element.innerHTML = '';
+  element.replaceChildren();
```
rusackas, line 214: The old way is fine, but this is less icky. I'm NOT sure why this wasn't called out by eslint in any of these plugins. But, this should work fine (let's test to make sure)

rusackas:
This'll need tests for superset-ui-core utils added if we want to bother with that approach.

nytai:
👍 this approach looks good and should help with any future issues in this area. Not sure why the frontend build is failing. Would be good to bring up an ephemeral env to test this more thoroughly

PR #26215, opened 2023-12-08
fix(plugin-chart-echarts): support truncated numeric x-axis

### LINKED ISSUE #26139: Bar chart with numeric x-axis always starts x-axis at zero regardless of data
Trying out 3.1.0rc1 I noticed that one of my bar charts looked wrong.

In version 3.0.0:

image

In version 3.1.0rc1:
image

I tried making a new chart with another dataset where I have year as a numeric and I experienced the same thing, though the x-axis terminated around 2500 instead of 7000 in the screenshot above.

### Environment
GENERIC_CHART_AXES is enabled

### Checklist

villebro:
### SUMMARY
A recent PR that added support for numerical x-axes caused unintuitive results when the x-axis contained a highly constrained interval, e.g. the years 1980-2017. In this case, the x-axis was bounded between [0, 7000], causing a highly spiked appearance. This is in contrast with other visualization tools, that default automatically truncating the x-axis.

This PR adds two new x-axis truncation controls, similar to the existing ones for the y-axis. With this change, the x-axis will be truncated by default. If the lower and/or upper bounds are left undefined, they will default to the min and max values from the dataset (calculated automatically by ECharts). This ensures that the data is bounded correctly.

### AFTER
Now the x-axis is automatically truncated (see the new "TRUNCATE X AXIS" control, which defaults to being checked):
image

By defining the lower and/or upper bound, the x-axis is updated accordingly:
image

### BEFORE
When disabling x-axis truncation, the current "dirac delta"-like behavior is reproduced on the Video Game example chart:
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes #26139
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/constants.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/constants.ts
@@ -26,6 +26,7 @@ export const DEFAULT_FORM_DATA: Partial<EchartsBubbleFormData> = {
   logYAxis: false,
   xAxisTitleMargin: 30,
   yAxisTitleMargin: 30,
+  truncateXAxis: false,
   truncateYAxis: false,
   yAxisBounds: [null, null],
   xAxisLabelRotation: 0,
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/controlPanel.tsx
@@ -26,7 +26,7 @@ import {
 } from '@superset-ui/chart-controls';
 
 import { DEFAULT_FORM_DATA } from './constants';
-import { legendSection } from '../controls';
+import { legendSection, truncateXAxis, xAxisBounds } from '../controls';
 
 const { logAxis, truncateYAxis, yAxisBounds, xAxisLabelRotation, opacity } =
   DEFAULT_FORM_DATA;
@@ -247,6 +247,8 @@ const config: ControlPanelConfig = {
             },
           },
         ],
+        [truncateXAxis],
+        [xAxisBounds],
         [
           {
             name: 'truncateYAxis',
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/transformProps.ts
@@ -158,6 +158,7 @@ export default function transformProps(chartProps: EchartsBubbleChartProps) {
     convertInteger(xAxisTitleMargin),
   );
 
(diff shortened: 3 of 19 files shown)


villebro:
@michael-s-molina @sfirke I've created a chart that demonstrates how this feature works: [Truncated numerical x-axis](http://35.93.27.117:8080/explore/?slice_id=97) I may add a new integration test to ensure the e2e flow works, but in the meantime I think the PR should be ready for review.

superset-frontend/plugins/plugin-chart-echarts/src/Bubble/constants.ts
```diff
@@ -26,6 +26,7 @@ export const DEFAULT_FORM_DATA: Partial<EchartsBubbleFormData> = {
   logYAxis: false,
   xAxisTitleMargin: 30,
   yAxisTitleMargin: 30,
+  truncateXAxis: false,
```
villebro, line 29: Note that for bubble chart, I'm assuming truncation is disabled, as it's more likely people want to see it starting from origo in this case (having a truncated x axis but a non-truncated y axis doesn't make sense for a scatter plot IMO).

michael-s-molina:
Thanks for the PR @villebro. While testing, I found a weird behavior when applying the bounds to temporal and categorial axes. I pushed a change to ignore the bounds for non-numerical axis and also improved the tooltip description to make it more clear that this control is only for numerical types.

michael-s-molina (approved):
LGTM. Thanks for the fix @villebro.

sfirke:
This is a nice enhancement, thank you @villebro 🙌 !

john-bodley:
Out of interest @villebro where was the original [1900, 2017) range defined?

john-bodley:
@villebro sorry by bad I was looking at the wrong screenshot. I meant the [0, 2,500) interval which resulted in the original "dirac delta"-like behavior.

superset-frontend/plugins/plugin-chart-echarts/src/Bubble/constants.ts
michael-s-molina: No objection. We can cherry-pick the change for the next patch.

villebro:
> @villebro sorry by bad I was looking at the wrong screenshot. I meant the [0, 2,500) interval which resulted in the original "dirac delta"-like behavior.

@john-bodley oh, that range wasn't explicitly defined anywhere, that's just what ECharts defaults to unless the bounds are either overridden or set to `dataMin` and `dataMax`.

sadpandajoe:
@villebro I think this pr might have a weird affect on current echart bubble charts. I've documented the issue here: https://github.com/apache/superset/issues/26242.

PR #26218, opened 2023-12-08
chore(linting): Trying to make eslint quieter

rusackas:
### SUMMARY

Disabling the rule about exhaustive hook dependencies, and just tweaking a bunch of other things to lessen the noise. 

Reduces the warnings from 275 down to 81.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina (commented):
@rusackas I think we shouldn't disable this rule as it's really important. Maybe make it as an error so it's not ignored anymore for new files. For current files, we could ignore the error for now and slowly fix them. It will require registering all current files that contain this error 😅

rusackas:
> @rusackas I think we shouldn't disable this rule as it's really important. Maybe make it as an error so it's not ignored anymore for new files. For current files, we could ignore the error for now and slowly fix them. It will require registering all current files that contain this error 😅

Agreed. Sounds like the perfect thing to do over a beer some evening. It'll be a long, but simple PR. There are some other bits in this PR that might be worth looking at before I muddy the waters ;)

superset-frontend/packages/superset-ui-chart-controls/src/operators/sortOperator.ts
```diff
@@ -28,42 +28,40 @@ import {
...
-    sortableLabels.includes(formData.x_axis_sort) &&
-    // the sort operator doesn't support sort-by multiple series.
-    isEmpty(formData.groupby)
-  ) {
-    if (formData.x_axis_sort === getXAxisLabel(formData)) {
       return {
         operation: 'sort',
         options: {
-          is_sort_index: true,
+          by: formData.x_axis_sort,
```
rusackas, line 61: Ummm.... I'm not sure how this got changed... it's certainly not a linting rule.

superset-frontend/packages/superset-ui-chart-controls/src/operators/sortOperator.ts
```diff
@@ -28,42 +28,40 @@ import {
 import { PostProcessingFactory } from './types';
 import { extractExtraMetrics } from './utils';
 
-export const sortOperator: PostProcessingFactory<PostProcessingSort> = (
-  formData,
-  queryObject,
```
rusackas, line 33: I think this was unused, leading to all the reformatting here.

rusackas:
Closing for now... I'll reconsider/rebuild this after we merge the eslint metrics and do some other things like whittling and bumping packages.

PR #26220, opened 2023-12-08
chore(tests): Add tests to the column denormalization flow

Vitor-Avila:
### SUMMARY
Following up on https://github.com/apache/superset/pull/26199, this PR adds some tests around the column denormalization logic.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/datasource/api_tests.py
+++ b/tests/integration_tests/datasource/api_tests.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 import json
-from unittest.mock import Mock, patch
+from unittest.mock import ANY, Mock, patch
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


villebro (approved):
Great tests, thanks for adding these @Vitor-Avila 👍 As there seems to be general consensus that the inverse is the better name here (`denormalize_columns` as opposed to `norrmalize_columns`), do you think it would be helpful to update the UI, table column, and the method in `BaseEngineSpec`? This would provide a more intuitive devex for future devs who have to maintain this logic. If needed I can probably take on this task, as I feel guilty for having introduced the original name..

Vitor-Avila:
> Great tests, thanks for adding these @Vitor-Avila 👍 As there seems to be general consensus that the inverse is the better name here (`denormalize_columns` as opposed to `norrmalize_columns`), do you think it would be helpful to update the UI, table column, and the method in `BaseEngineSpec`? This would provide a more intuitive devex for future devs who have to maintain this logic. If needed I can probably take on this task, as I feel guilty for having introduced the original name..

hey @villebro! I'm not 100% sure on the most accurate name to represent this feature, so I'd leave this decision up for you. I think the confusing part to me was that we were introducing a new feature to handle the column casing, but then we can "enable the disablement" via the UI. I think it would make more sense to me that the checkbox in the UI would be used to enable the feature, which would then be disabled for older datasets, but enabled by default for new ones. I understand the decision process here (the checkbox was introduced after the feature, so it made sense to keep it disabled when we were doing it since it was already live). 

Let me know your thoughts -- happy to help as much as I can.

PR #26224, opened 2023-12-08
fix: Use page.locator in Playwright reports

kgabryje:
### SUMMARY
Per [Playwright's docs](https://playwright.dev/python/docs/api/class-page#page-wait-for-selector), `page.wait_for_selector` method is deprecated. This PR replaces usages of that method with `page.locator` suggested by the docs.

### TESTING INSTRUCTIONS
1. Enable `PLAYWRIGHT_REPORTS_AND_THUMBNAILS` feature flag
2. Run some reports and verify that it works like before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/webdriver.py
+++ b/superset/utils/webdriver.py
@@ -48,8 +48,8 @@ if TYPE_CHECKING:
 if feature_flag_manager.is_feature_enabled("PLAYWRIGHT_REPORTS_AND_THUMBNAILS"):
     from playwright.sync_api import (
         BrowserContext,
-        ElementHandle,
         Error as PlaywrightError,
+        Locator,
         Page,
         sync_playwright,
         TimeoutError as PlaywrightTimeout,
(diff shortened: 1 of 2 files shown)


superset/utils/webdriver.py
```diff
@@ -166,34 +152,40 @@ def get_screenshot(self, url: str, element_name: str, user: User) -> bytes | Non
...
                     logger.exception("Timed out requesting url %s", url)
                     raise ex
 
                 try:
                     # chart containers didn't render
                     logger.debug("Wait for chart containers to draw at url: %s", url)
-                    page.wait_for_selector(
-                        ".slice_container", timeout=self._screenshot_locate_wait * 1000
-                    )
+                    for slice_container_elem in page.locator(".slice_container").all():
```
geido, line 187: I assume all of these .all() methods are safe for the loop and will just return empty if none are present

geido (approved):

PR #26011, opened 2023-11-17
feat(telemetry): Adding Scarf based telemetry to Superset

rusackas:
### SUMMARY

This PR does a handful of things:
* Adds the Scarf pixel to Superset (as approved by SIP) via a new `TelemetryPixel` component
* Adds the `ENABLE_TELEMETRY` feature flag 
* Adds basic tests for the above
* Bumps react types to support the `no-referrer-when-downgrade` 
* Bumps the npm version... not that I intended to, but it keeps coming along for the ride
* Fixes the shim for frontend tests, which had a weird incompatibility with certain node 18.x versions, throwing a `TypeError: Cannot redefine property: window` error

TODO:
* Get the pixel to load via Scarf gateway so that url params (version/sha/build) are reliably captured
* Feature flag documentation
* Turn off _more_ Scarf stuff with this flag?

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N/A

### TESTING INSTRUCTIONS

RTL tests for the new component with FF on and off should pass. Run it, and have any PMC member check the Scarf data for proper capture. 

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/25639
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [x] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26251, opened 2023-12-12
chore: improve CSP add base uri restriction

dpgaspar:
### SUMMARY

Improves our current default CSP policy by adding `base-uri` to `self`. base-uri directive restricts the URLs which can be used in a document's [<base>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) element.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -1425,6 +1425,7 @@ TALISMAN_ENABLED = utils.cast_to_boolean(os.environ.get("TALISMAN_ENABLED", True
 # If you want Talisman, how do you want it configured??
 TALISMAN_CONFIG = {
     "content_security_policy": {
+        "base-uri": ["'self'"],
         "default-src": ["'self'"],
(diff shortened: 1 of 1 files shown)


kgabryje (approved):
LGTM

nytai (approved):

nytai:
@rusackas do we have any front-end logging enabled by default? This might affect any telemetry we might have recently added

rusackas:
> @rusackas do we have any front-end logging enabled by default? This might affect any telemetry we might have recently added

I _think_ we'll be OK with the telemetry. [The PR](https://github.com/apache/superset/pull/26011) is still open until I sort out the tests, but the only change to the CSP change there just pokes a couple holes in the `img-src` setting.

PR #26269, opened 2023-12-13
fix(plugin-chart-echarts): use scale for truncating x-axis

### LINKED ISSUE #26254: Bar chart spilling out past x-axis on both ends
Superset 3.1.0rc2:

image

How it looks previously in 3.0.0:

image

Here's the chart config in 3.1.0rc2: 

image

Again there's a numeric x-axis, maybe broken by #26215 ?

villebro:
### SUMMARY
Following up on #26215 by playing whack-a-mole on the numerical x-axis truncation problem. Move from using `dataMin`/`dataMax` to `scale: true` when truncating x-axis for bar charts. This applies automatic truncation by ECharts, which is able to handle truncation more gracefully. For other than bar charts, we'll retain the `dataMin`/`dataMax` logic without the `scale` parameter.

### AFTER
Now there is a proper margin on the x-axis to not overlap y-axis.
image

Now also mixed series chart has the same functionality:
image

### BEFORE
Previously the smallest and largest bars would 
image

### ADDITIONAL INFORMATION

- [c] Has associated issue: fixes #26254
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -35,6 +35,8 @@ import { EchartsTimeseriesSeriesType } from '../Timeseries/types';
 import {
   legendSection,
   richTooltipSection,
+  truncateXAxis,
+  xAxisBounds,
   xAxisLabelRotation,
 } from '../controls';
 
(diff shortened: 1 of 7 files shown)


michael-s-molina (changes requested):
You need to pass `seriesType` from `transformProps`.

villebro:
> You need to pass `seriesType` from `transformProps`.

Thanks for catching, too many iterations on how to implement this caused the final commit to miss this change..

michael-s-molina:
You also need to adjust the mixed timeseries chart 🙂

villebro:
> You also need to adjust the mixed timeseries chart 🙂

😬

michael-s-molina (approved):
LGTM

PR #26281, opened 2023-12-14
fix(sql lab): Use quote_schema instead of quote method to format schema name

guenp:
### SUMMARY
DuckDB supports multiple databases. This means schemas can be specified as `<db name>.<schema name>`, e.g., `"my db".main`. However, when such a schema name is specified, it gets quoted as `"my db.main"` instead of `"my db"."main"`. This PR fixes that by using the `quote_schema` method of the `identifier_preparer` instead of `quote`. This method defaults to `identifier_preparer.quote` is overridden in the DuckDB sqlalchemy driver to parse the string and add quotes in the right places.

N/A

### ADDITIONAL INFORMATION

- [x] Has associated PR: https://github.com/Mause/duckdb_engine/pull/848

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -1433,8 +1433,9 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         if show_cols:
             fields = cls._get_fields(cols)
         quote = engine.dialect.identifier_preparer.quote
+        quote_schema = engine.dialect.identifier_preparer.quote_schema
         if schema:
-            full_table_name = quote(schema) + "." + quote(table_name)
+            full_table_name = quote_schema(schema) + "." + quote(table_name)
         else:
             full_table_name = quote(table_name)
 


superset/db_engine_specs/base.py
```diff
@@ -1433,8 +1433,9 @@ def select_star(  # pylint: disable=too-many-arguments,too-many-locals
         if show_cols:
             fields = cls._get_fields(cols)
         quote = engine.dialect.identifier_preparer.quote
+        quote_schema = engine.dialect.identifier_preparer.quote_schema
```
villebro, line 1436: Daydreaming: How I wish Python had this syntax like JS/TS does..
```ts
{ quote, quote_schema } = engine.dialect.identifier_preparer
```

villebro (approved):
Nice! A quick glance at my `venv` revealed that at least the default MSSQL and Snowflake dialects are overriding this with their own logic, with the default falling back to `quote` as stated here. LGTM

superset/db_engine_specs/base.py
guenp: Wherever a schema is the input argument, yes. For table names and the like, `quote` is fine. As far as I could tell this is the only place where a schema name is formatted

guenp:
Please merge this PR. We need it to address this issue: https://github.com/apache/superset/issues/26286

PR #26282, opened 2023-12-15
feat(async_queries): allow dynamic feature flag to take effect after initial app bootstrapping

zephyring:
### SUMMARY
Before this change, we only initialize _async_query_manager instance on app init.
If `GLOBAL_ASYNC_QUERIES` feature flag changes afterwards, we will not able to an instance of get async_query_manager and thus all requests fail without restarting the app.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/async_events/async_query_manager_factory.py
+++ b/superset/async_events/async_query_manager_factory.py
@@ -32,4 +32,11 @@ class AsyncQueryManagerFactory:
         self._async_query_manager.init_app(app)
 
     def instance(self) -> AsyncQueryManager:
+        # check instance init again to allow dynamic feature flag change to apply
+        # after initial app bootstrapping
+        if not self._async_query_manager:
+            from superset import is_feature_enabled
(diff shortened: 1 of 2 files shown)


superset/async_events/async_query_manager_factory.py
```diff
@@ -32,4 +32,11 @@ def init_app(self, app: Flask) -> None:
         self._async_query_manager.init_app(app)
 
     def instance(self) -> AsyncQueryManager:
+        # check instance init again to allow dynamic feature flag change to apply
+        # after initial app bootstrapping
+        if not self._async_query_manager:
+            from superset import is_feature_enabled
```
john-bodley, line 38: I believe these imports could be global.

zephyring: It throws `cannot import name 'is_feature_enabled' from partially initialized module 'superset' `

PR #26290, opened 2023-12-16
feat(echarts-funnel): Implement % calculation type

kgabryje:
### SUMMARY
Add "% Calculation" field to Funnel chart's control panel. Currently, the % of each level of funnel chart is calculated as % of total. This PR allows user to also display % of each level as % of the first step or % of previous step.

Example of calculations:

| Value | Percent of total | Calculate from first step | Calculate from previous step |
| --- | --- | --- | --- | 
| 100 | 57% | 100% | 100% | 
| 50 | 28% | 50% | 50% | 
| 25 | 14% | 25% | 50% |

The default value for new charts is "Calculate from first step". For existing charts, a migration was added to keep "Percent of total" as selected option.

image

image

image

### TESTING INSTRUCTIONS
1. Create a funnel chart
2. Verify that the calculated percents of each step match expected values for each option

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
Current: 0.21 s
10+: 0.21 s
100+: 0.12 s
1000+: 0.13 s

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
@@ -19,12 +19,12 @@
 import {
   CategoricalColorNamespace,
   DataRecord,
+  getColumnLabel,
   getMetricLabel,
   getNumberFormatter,
+  getValueFormatter,
   NumberFormats,
   ValueFormatter,
-  getColumnLabel,
-  getValueFormatter,
 } from '@superset-ui/core';
 import { CallbackDataParams } from 'echarts/types/src/util/types';
 import { EChartsCoreOption, FunnelSeriesOption } from 'echarts';
@@ -34,6 +34,7 @@ import {
   EchartsFunnelFormData,
   EchartsFunnelLabelTypeType,
   FunnelChartTransformedProps,
+  PercentCalcType,
 } from './types';
 import {
   extractGroupbyLabel,
@@ -43,7 +44,7 @@ import {
   sanitizeHtml,
 } from '../utils/series';
 import { defaultGrid } from '../defaults';
-import { OpacityEnum, DEFAULT_LEGEND_FORM_DATA } from '../constants';
+import { DEFAULT_LEGEND_FORM_DATA, OpacityEnum } from '../constants';
 import { getDefaultTooltip } from '../utils/tooltip';
 import { Refs } from '../types';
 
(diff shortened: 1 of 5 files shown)


eschutho (approved):
Looks great! Thanks for including the migration timings, too.

superset-frontend/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
```diff
@@ -154,23 +171,33 @@ export default function transformProps(
...
       datum,
       groupby: groupbyLabels,
       coltypeMapping: {},
     });
+    const value = datum[metricLabel] as number;
     const isFiltered =
       filterState.selectedValues && !filterState.selectedValues.includes(name);
+    const firstStepPercent = value / (data[0][metricLabel] as number);
+    const prevStepPercent =
+      index === 0 ? 1 : value / (data[index - 1][metricLabel] as number);
```
villebro, line 189: nit: maybe this slightly shortens it, although it's maybe less explicit..
```suggestion
      index ? value / (data[index - 1][metricLabel] as number) : 1;
```

superset/migrations/versions/2023-12-15_17-58_06dd9ff00fe8_add_percent_calculation_type_funnel_.py
```diff
@@ -0,0 +1,71 @@
...
+    id = Column(Integer, primary_key=True)
+    viz_type = Column(String(250))
+    params = Column(Text)
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = db.Session(bind=bind)
+
+    for slc in session.query(Slice).filter(Slice.viz_type == "funnel"):
```
villebro, line 50: Could we use `paginated_update` here? I'm not expecting people to have 100k funnel charts, but as a principle, I think it's a good idea to always paginate updates when possible.

superset/migrations/versions/2023-12-15_17-58_06dd9ff00fe8_add_percent_calculation_type_funnel_.py
```diff
@@ -0,0 +1,71 @@
...
+            slc.params = json.dumps(params)
+            session.commit()
+    session.close()
+
+
+def downgrade():
+    bind = op.get_bind()
+    session = db.Session(bind=bind)
+
+    for slc in session.query(Slice).filter(Slice.viz_type == "funnel"):
```
villebro, line 64: Same here

villebro (approved):
Great improvement! Small theoretical scalability improvement for the migration, other than that LGTM

superset/migrations/versions/2023-12-15_17-58_06dd9ff00fe8_add_percent_calculation_type_funnel_.py
michael-s-molina: +1

kgabryje: Good point!

superset-frontend/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
kgabryje: TBH in case such as this, I think it's better to be explicit that we perform an operation only for the first element of an array, rather than when "index is not falsy"

villebro: Yes, very good point. This was mostly a knee-jerk reaction that I often get when I get when I see something that can be shortened with truthy/falsy checking. Carry on 👍

superset/migrations/versions/2023-12-15_17-58_06dd9ff00fe8_add_percent_calculation_type_funnel_.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for slc in paginated_update(
+        session.query(Slice).filter(Slice.viz_type == "funnel")
+    ):
+        params = json.loads(slc.params)
+        percent_calculation = params.get("percent_calculation_type")
+        if not percent_calculation:
+            params["percent_calculation_type"] = "total"
+            slc.params = json.dumps(params)
+            session.commit()
```
villebro, line 59: I don't think we should be committing on the changes (`paginated_update` should do that for us)
```suggestion
```

superset/migrations/versions/2023-12-15_17-58_06dd9ff00fe8_add_percent_calculation_type_funnel_.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for slc in paginated_update(
+        session.query(Slice).filter(Slice.viz_type == "funnel")
+    ):
+        params = json.loads(slc.params)
+        percent_calculation = params.get("percent_calculation_type")
+        if percent_calculation:
+            del params["percent_calculation_type"]
+            slc.params = json.dumps(params)
+            session.commit()
```
villebro, line 75: same here
```suggestion
```

villebro (changes requested):
We should remove the explicit commit, as that's already happening in the paginator

kgabryje:
> We should remove the explicit commit, as that's already happening in the paginator

TIL 🙂 done

villebro (approved):

PR #26310, opened 2023-12-19
chore: Disables minor ticks by default

michael-s-molina:
### SUMMARY
This PR follows the same standard defined by ECharts and disables [minor ticks](https://echarts.apache.org/en/option.html#xAxis.minorTick.show) by default for ECharts charts. From a data visualization point of view, they only add visual noise/don’t facilitate reading the chart better for the majority of cases. For the specific cases where they might help with data interpretation, I added a new control that allows users to enable them. This will also help with consistency given that legacy charts also have minor ticks disabled by default. 

Screenshot 2023-12-19 at 12 00 57
Screenshot 2023-12-19 at 12 00 35

### TESTING INSTRUCTIONS
1 - Create a timeseries chart
2 - Check the new control behavior
3 - Check that charts are rendered without minor ticks by default

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -34,6 +34,7 @@ import { DEFAULT_FORM_DATA } from './types';
 import { EchartsTimeseriesSeriesType } from '../Timeseries/types';
 import {
   legendSection,
+  minorTicks,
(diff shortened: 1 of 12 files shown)


villebro (approved):
LGTM. One thing to keep in mind is minor ticks will now disappear from existing charts. However, I think this is ok, as long as we're in agreement that disabled minor ticks are a more sensible default.

PR #26314, opened 2023-12-20
fix(logging): Filter out undefined columns

john-bodley:
### SUMMARY

The `LOG_ACTIONS_CHANGE_DASHBOARD_FILTER` action logs the chart ID and columns which a filter impacts. The issue is if one were to clear a filter the key still persists in the `newSelectedValues` object, i.e., `{'country_name': null}`, resulting in a misleading log entry.

This PR simply filters out any key in which the associated value is `null`.

### TESTING INSTRUCTIONS

Verified that when clearing the `COUNTRY_NAME` filter on the "World Bank's Data" dashboard the `country_name` column was not logged, i.e., the following SQL statement,

```sql
SELECT
    json_extract(json, '$.columns')
FROM 
    logs
WHERE 
    json_extract(json, '$.event_name') = 'change_dashboard_filter' 
ORDER BY
    dttm DESC
```

returned:

```
[]
["country_name"]
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/gridComponents/Chart.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/Chart.jsx
@@ -274,7 +274,9 @@ class Chart extends React.Component {
   changeFilter(newSelectedValues = {}) {
     this.props.logEvent(LOG_ACTIONS_CHANGE_DASHBOARD_FILTER, {
       id: this.props.chart.id,
-      columns: Object.keys(newSelectedValues),
+      columns: Object.keys(newSelectedValues).filter(
+        key => newSelectedValues[key] !== null,
+      ),
     });
     this.props.changeFilter(this.props.chart.id, newSelectedValues);
   }


john-bodley:
@michael-s-molina I gather is might be obsolete due to your 4.0 efforts to delete filter-boxes.

michael-s-molina:
Given that the name of the action is `LOG_ACTIONS_CHANGE_DASHBOARD_FILTER`, when a filter is cleared, we need to indicate that the filter changed. Do you expect that consumers of this action will keep the previous state to compare?

john-bodley:
@michael-s-molina it's unclear. I believe this logic was added by Airbnb and after an internal code search there's no references where the payload is introspected, i.e., for performance tracking we simply track the event name state change. 

Additionally, per https://github.com/apache/superset/pull/26314, one can see that currently the logging is misleading when filters are cleared.

rusackas (approved):
Looks straightforward to me!

PR #26317, opened 2023-12-20
chore: Adds a tooltip for the alert's SQL input

### LINKED ISSUE #26298: Alerts aren't firing with error 'Alert query returned a non-number value.' when returning a date column
#### How to reproduce the bug

1. Create an alert
2. Use the examples database
3. Use this query: `SELECT year FROM main.wb_health_population LIMIT 1`
4. Set the condition as `!=`
5. Set the value a `0`
6. Finish filling out the rest of the alert form

### Expected results

The alert is sent because the condition is met

### Actual results

Send error with `Alert query returned a non-number value.`

### Environment

- browser type and version:
- superset version: `superset version`
- node.js version: `node -v`
- any feature flags active:

### Checklist

### Additional context
Looks like the first failure was with this SHA `2ac28927a326fee6431b5e01f7930e650c636c17`

michael-s-molina:
### SUMMARY
It's not very clear that the SQL used to create an alert should return a numeric value. When non-numeric values are returned, the error `Alert query returned a non-number value` is thrown when firing the alert. This PR adds a tooltip for the alert's SQL input to instruct users in advance. 

Closes https://github.com/apache/superset/issues/26298

Screenshot 2023-12-20 at 09 03 28

Screenshot 2023-12-20 at 09 04 04

### TESTING INSTRUCTIONS
1 - Open the alert creation modal
2 - Check the displayed tooltip

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -401,11 +401,12 @@ export const TRANSLATIONS = {
   ALERT_CONDITION_TEXT: t('Alert condition'),
   DATABASE_TEXT: t('Database'),
   SQL_QUERY_TEXT: t('SQL Query'),
-  SQL_QUERY_TOOLTIP: t('The result of this query should be a numeric value'),
+  SQL_QUERY_TOOLTIP: t(
+    'The result of this query should be a numeric-esque value',
+  ),
   TRIGGER_ALERT_IF_TEXT: t('Trigger Alert If...'),
   CONDITION_TEXT: t('Condition'),
   VALUE_TEXT: t('Value'),
-  VALUE_TOOLTIP: t('Threshold value should be double precision number'),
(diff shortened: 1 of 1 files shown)


john-bodley:
@michael-s-molina technically it doesn't need to be numeric, i.e., it could be a string, as we use the Python `float()` method. Maybe something like:

> ... a numeric-esque value

While we're at it, should we also simply remove tooltip for the threshold value,

Screenshot 2023-12-21 at 11 15 57 AM

given that it technically doesn't need to be a double precision number, e.g., an integer will work. It seems like the input box rejects non-numerical values (excluding the period) and thus the input in sanitized.

michael-s-molina:
@john-bodley Applied your suggestions.

john-bodley (approved):

Issue #25945, opened 2023-11-10
[SIP] Proposal for customize email subject for reports and alerts


puridach-w:
### Motivation

The primary motivation for this feature comes from our need to send parameters along with the subject of the alert or report email. This feature is designed to make the email subject customizable, enhancing users' ability to make targeted and specific email subject lines.

### Proposed Change

I propose a feature that allows users to customize the email subject of a report or alert. In the report submission creation interface, I plan to add a text area for entering the email subject. If a user inputs a value in this field, the email's subject will be this value. If not, the system will assign a default email subject line.

**Before:**

<img width="1171" alt="Screenshot 2566-11-10 at 15 15 52" src="https://github.com/apache/superset/assets/76474644/ec28fb0c-bd8e-4f5d-a577-26df3f7a0e27">

<img width="1173" alt="Screenshot 2566-11-10 at 14 28 09" src="https://github.com/apache/superset/assets/76474644/edb3d66f-2704-48bd-b4ab-21b7a282c8d4">

**After:** 

<img width="1177" alt="Screenshot 2566-11-10 at 14 08 56" src="https://github.com/apache/superset/assets/76474644/4b2e2329-e9cd-49ad-af79-c384a6f45469">

<img width="1172" alt="Screenshot 2566-11-10 at 14 09 34" src="https://github.com/apache/superset/assets/76474644/1fb4a563-522e-42bb-81a8-eccaaf77755d">

### New or Changed Public Interfaces

I will make a small change to the UI in the 'create alert/report' modal section. If the user selects the email sending method, an additional text area will be available for the user to populate the email subject line.

### New dependencies

No new packages or libraries will be required to implement this feature.

### Migration Plan and Compatibility

To incorporate this feature, we will need to initiate a database migration to add columns dedicated to storing email subject lines. We have tested this feature with Superset version 3.0.1 and confirmed that it operates correctly.

### Rejected Alternatives

None

michael-s-molina:
Hi @puridach-w. Thanks for the proposal. I think it would be a good idea to incorporate your changes into this [discussion](https://github.com/apache/superset/discussions/25729) which will become a SIP in the future.

rusackas:
CC @yousoph and @eschutho on the above. I think we can roll this into the discussion/SIP on the other thread, but I'm also wondering if this change even requires a SIP. If the default remains the same, but this just adds a dimension of customization, it feels harmless.

puridach-w:
[...]

Issue #26183, opened 2023-12-05
[SIP] Proposal for Alert & Report Modal with Dashboard Filters and Tabs


lilykuang:
### Motivation

We've identified a recurring request for enhanced capabilities in alert and report configurations specifically, the inclusion of more granular filters and tab details. To address this, we are considering updates to the user interface for creating alerts and reports. These updates are aimed at incorporating additional filter and tab options, and reorganizing field groupings to heighten modal clarity and simplify the setup process for users.

### Proposed Change

For an overview of the proposed enhancements to Alert and Report Modal, please refer to the design mock-ups in [this Superset discussion](https://github.com/apache/superset/discussions/25729). Additionally, the proof of concept (POC) can be viewed in [this pull request](https://github.com/apache/superset/pull/26141).

1. **Refined User Interface:** We're addressing concerns of clutter in the current modal by introducing a design that emphasizes simplicity and order. The new layout will dispense with superfluous elements to reduce user confusion and streamline interaction.
2. **Intuitive Configuration Sequence:** The setup process will be restructured into a clear, step-by-step sequence. By guiding users through each phase methodically, we aim to simplify the creation of alerts and reports, thereby minimizing potential errors.
3. **Enhancements and New Functionalities:**
    - **Dashboard-Imported Filters/Tabs for Customization:** The latest system update introduces a variety of filters and tabs, all imported from the dashboard. This new functionality allows users to effectively customize their alerts and reports to meet specific requirements, enhancing the adaptability and precision of data handling and reporting.
    - **Provision for Test Report Deployment:** This new feature enables the dispatch of test reports, a crucial tool for validating report configurations. It ensures reliable and accurate delivery of these reports, operating independently of the standard report scheduling process.

### New or Changed Public Interfaces

- **Dashboard State Management**: Upon user interaction with the dashboard dropdown, the **`useDashboard`** hook will be utilized to access the selected dashboard. New methods will be implemented to update the **`dataMask`** and **`nativeFilters`** states in the Redux store. This enhancement is aimed at facilitating the integration of dashboard-specific native filters and tab selections using established hooks like **`useNativeFiltersDataMask`** and **`useFilters`**. This strategy is designed to ensure a seamless and efficient integration within the existing system's framework.
[...]

PR #26327, opened 2023-12-21
feat: Customizable email subject name

puridach-w:
### SUMMARY
This feature allows users to customize the email subject when creating a report or an alert within Superset. This is driven by a need to include specific parameters within the email subject line, enhancing the specificity and resonance of alert or report notifications. 

Changes:
- Implementation of a new text input field within the 'create alert/report' interface, through which users can specify their desired email subject line.
- If the text field is left blank, a default email subject line will be generated by the system.
- The user interface will show the new text field solely when the email sending method is selected by a user.

This feature has been tested with Superset master branch.
This proposed feature has already been discussed in the Superset Improvement Proposal (see https://github.com/apache/superset/issues/25945)

### BEFORE SCREENSHOTS
image

image

### AFTER SCREENSHOTS
image

image

### TESTING INSTRUCTIONS
To verify the changes made, please follow these steps: (Assume that there TAG for alert and report and open)
- Go to the Alert and report creation page
- Create one of them
- Fill all the information that required
- Select Notification method as "Email"
- You will see input box for "email subject name"
- Fill the desired subject name and save it
- Wait for email send and check the email subject name

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


puridach-w:
Run: superset db downgrade

Screenshot 2566-12-21 at 12 53 24

Run: superset db upgrade

Screenshot 2566-12-21 at 12 53 35

superset/reports/notifications/email.py
```diff
@@ -177,8 +177,7 @@ def _get_content(self) -> EmailContent:
 
     def _get_subject(self) -> str:
         return __(
-            "%(prefix)s %(title)s",
-            prefix=app.config["EMAIL_REPORTS_SUBJECT_PREFIX"],
```
giftig, line 181: Feels like this could still be useful even with per-schedule customisation, as you might want to prefix all your emails with a standard prefix and then choose individual subject lines after the prefix. I'd continue supporting it, though we could probably improve the behaviour where prefix is `None` or empty by removing any leading space etc.

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -420,6 +420,7 @@ export const TRANSLATIONS = {
   CUSTOM_SCREENSHOT_WIDTH_TEXT: t('Screenshot width'),
   CUSTOM_SCREENSHOT_WIDTH_PLACEHOLDER_TEXT: t('Input custom width in pixels'),
   NOTIFICATION_METHOD_TEXT: t('Notification method'),
+  SUBJECT_TEXT: t('Subject name'),
```
giftig, line 423: "name" doesn't seem right here. Subject line, or subject text, or just subject, makes more sense I think.

puridach-w: I already resolve this one it is unused code ad I already remove it thanks

superset/reports/notifications/email.py
puridach-w: I think if the user wants to have prefix then they can always add it themselves and we should allow user 100% control over their email subject.

For the default subject line it still have prefix I just modify code to this

```
if self._report_schedule.chart:
                name = (
                    f"{prefix} "
                    f"{self._report_schedule.name}: "
                    f"{self._report_schedule.chart.slice_name}"
                )
            else:
                name = (
                    f"{prefix} "
                    f"{self._report_schedule.name}: "
                    f"{self._report_schedule.dashboard.dashboard_title}"
                )
```

puridach-w:
> ## [Codecov](https://app.codecov.io/gh/apache/superset/pull/26327?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache) Report
> Attention: `6 lines` in your changes are missing coverage. Please review.
> 
> > Comparison is base [(`8c32c6d`)](https://app.codecov.io/gh/apache/superset/commit/8c32c6da169afec312923e516850d90a69e78f46?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache) 69.18% compared to head [(`f7c7a9a`)](https://app.codecov.io/gh/apache/superset/pull/26327?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache) 59.04%.
> > Report is 5 commits behind head on master.
> 
> [Files](https://app.codecov.io/gh/apache/superset/pull/26327?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache)	Patch %	Lines
> [superset/commands/report/execute.py](https://app.codecov.io/gh/apache/superset/pull/26327?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache#diff-c3VwZXJzZXQvY29tbWFuZHMvcmVwb3J0L2V4ZWN1dGUucHk=)	0.00%	[6 Missing ⚠️ ](https://app.codecov.io/gh/apache/superset/pull/26327?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache)
> Additional details and impacted files
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/apache/superset/pull/26327?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=apache).

What should I do with this? Can you suggest please.

john-bodley:
ping @eschutho and @lilykuang who are working on [SIP-110](https://github.com/apache/superset/issues/26183).

puridach-w:
@giftig @john-bodley Could you please help me review this PR again?

giftig:
@puridach-w the code all looks ok to me but I'm not very familiar with the subject of reports. I think it'd be a good idea to get a review from @eschutho or @lilykuang since they're working on the related SIP John mentioned above; would be good to confirm what's done here aligns ok with that work.

superset/migrations/versions/2024-01-03_16-57_61304d31c995_add_subject_column_to_report_schedule.py
```diff
@@ -0,0 +1,40 @@
...
+revision = '61304d31c995'
+down_revision = '06dd9ff00fe8'
+
+import sqlalchemy as sa
+from alembic import op
+
+
+def upgrade():
+    op.add_column(
+        "report_schedule", sa.Column("email_subject", sa.Text(), nullable=True)
```
eschutho, line 35: I think a `String(length=255)` type should be all you need here, rather than a text string?

superset/reports/models.py
```diff
@@ -170,6 +170,8 @@ class ReportSchedule(AuditMixinNullable, ExtraJSONMixin, Model):
 
     extra: ReportScheduleExtra  # type: ignore
 
+    email_subject = Column(Text)
```
eschutho, line 173: Same here `String(255)`

superset/reports/notifications/email.py
eschutho: To keep backward compatibility, can we continue to include the prefix if `app.config["EMAIL_REPORTS_SUBJECT_PREFIX"]` exists or is not blank/empty/nullish? This way each application can decide for themselves if they would like to include a prefix or not.

superset/reports/schemas.py
```diff
@@ -146,6 +146,12 @@ class ReportSchedulePostSchema(Schema):
         allow_none=True,
         required=False,
     )
+    email_subject = fields.String(
+        description=description_description,
+        allow_none=True,
+        required=False,
+        example="[Report]  Report name: Dashboard or chart name",
```
eschutho, line 153: I think you want to move `example` and description into a metadata dictionary. You can also define a new description variable for the email_subject.

superset/reports/schemas.py
```diff
@@ -272,6 +278,12 @@ class ReportSchedulePutSchema(Schema):
...
```
eschutho, line 285: same as above

superset/commands/report/execute.py
```diff
@@ -377,16 +378,21 @@ def _get_notification_content(self) -> NotificationContent:
...
+                    f"{prefix} "
+                    f"{self._report_schedule.name}: "
+                    f"{self._report_schedule.dashboard.dashboard_title}"
+                )
         else:
-            name = (
-                f"{self._report_schedule.name}: "
-                f"{self._report_schedule.dashboard.dashboard_title}"
-            )
+            name = self._report_schedule.email_subject
```
eschutho, line 395: This name variable is really just for internal logs. I'm not sure if the subject line would help here?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -888,6 +890,10 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     const parsedValue = type === 'number' ? parseInt(value, 10) || null : value;
 
     updateAlertState(name, parsedValue);
+
+    if (name === 'name') {
+      updateEmailSubject();
+    }
```
eschutho, line 896: instead of using `onInputChange` here, you can just create a new event handler like this for example:
```
  const onOwnersChange = (value: Array<SelectValue>) => {
    updateAlertState('owners', value || []);
  };
  ```

File: superset-frontend/src/features/alerts/AlertReportModal.tsx [hunk 9]
```diff
@@ -922,12 +928,14 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
   const onDashboardChange = (dashboard: SelectValue) => {
     updateAlertState('dashboard', dashboard || undefined);
     updateAlertState('chart', null);
+    updateEmailSubject();
```
eschutho, line 931: Instead of updating the subject line here, you can also add the formatting when saving.

superset-frontend/src/features/alerts/components/NotificationMethod.tsx
```diff
@@ -110,6 +123,18 @@ export const NotificationMethod: FunctionComponent<NotificationMethodProps> = ({
     }
   };
 
+  const onSubjectChange = (
+    event: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>,
+  ) => {
+    const { target } = event;
+
+    setSubjectValue(target.value);
```
eschutho, line 131: since you're passing the email subject down as a prop, you can use the prop value from the parent instead of storing it here.

eschutho (dismissed):
I left a few comments. Feel free to ping me if you'd like any clarification on anything or have questions.

superset-frontend/src/features/alerts/components/NotificationMethod.tsx
puridach-w: Modified

superset/migrations/versions/2024-01-03_16-57_61304d31c995_add_subject_column_to_report_schedule.py
puridach-w: You're right I already modified it

superset/reports/models.py
puridach-w: Modified

superset/reports/schemas.py
puridach-w: Modified

superset/reports/schemas.py
puridach-w: Modified

superset-frontend/src/features/alerts/AlertReportModal.tsx
puridach-w: But it also needs to update the 'report name' value in the state before updating the email subject. That is the reason why I added this code. It is used to show the default subject line immediately when changing the report name. If a separate function is used, code must also be added to set the new value into the state, which will cause redundant code.
```
if (name === 'name') {
      updateEmailSubject();
    }
```

superset/commands/report/execute.py
puridach-w: If i didn't misunderstand the code that use for send alert is smtp
`send_email_smtp(
                to,
                subject,
                content.body,
                app.config,
                files=[],
                data=content.data,
                images=content.images,
                bcc="",
                mime_subtype="related",
                dryrun=False,
                header_data=content.header_data,
            )`
   
 and subject it come from this 
 `def _get_subject(self) -> str:
        return __(
            "%(title)s",
            title=self._content.name,
        )`

superset-frontend/src/features/alerts/AlertReportModal.tsx
puridach-w: Modified

puridach-w:
Ping @eschutho I have already modified the code based on your comment.

puridach-w:
Ping @eschutho Could you please help me review this PR again?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -1160,6 +1167,18 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+  const updateEmailSubject = () => {
+    if (contentType === 'chart') {
+      setEmailSubject(
+        `${currentAlert?.name}: ${currentAlert?.chart?.label || ''}`,
+      );
+    } else {
+      setEmailSubject(
+        `${currentAlert?.name}: ${currentAlert?.dashboard?.label || ''}`,
+      );
+    }
```
eschutho, line 1179: since the default is handled in execute.py (server side) I don't think you need to also define it here in the client. It can just be left blank.

superset/reports/notifications/email.py
eschutho: I agree with @giftig on the premise that the app.config is set by the application admin, whereas the title will be set by individual users. If an application admin set an app.config prefix, it's likely something important that they don't want removed from the emails, such as "Company Confidential", etc.

superset/commands/report/execute.py
eschutho: Thanks for pointing that out. You're correct. Maybe just a small nit for readability, I would switch the logic to 
````
if self._report_schedule.email_subject:
    name = self._report_schedule.email_subject
else:
...
```
I also left a separate comment about persisting the app.config prefix.

superset-frontend/src/features/alerts/AlertReportModal.tsx
puridach-w: As you can see in the image the default value will be shown if user doesn't input any text and this code is a part of that function.

image

superset/commands/report/execute.py
puridach-w: Modified

superset/reports/notifications/email.py
puridach-w: Ok, I got it. I already modify to be as previous.

superset-frontend/src/features/alerts/AlertReportModal.tsx
puridach-w: First video: https://github.com/apache/superset/assets/76474644/47df84ce-e7c1-4db9-b008-3b43b6c39592

From the first video, it can be observed that the Report or Dashboard name changes whenever the _default value that's shown as a placeholder changes. The need for two variables arises from the requirement to display the default Subject name, and another one to store the customized name.

Second Video: https://github.com/apache/superset/assets/76474644/bb024b5e-96e6-4323-ada7-cef981752341

As shown in the second video, initially when I customize the "Email subject name", it changes again when I edit the report name. This is caused by using only the currentAlert.subject_name as the placeholder and value for the "Email subject name" field. Therefore, when the report name is changed, the "Email subject name" field gets updated.

Do you think it's important to display the default email subject line?

puridach-w:
Ping @eschutho Could you please help me review this PR again?

rusackas:
A rebase of this branch should fix the CI issue you're seeing. 🤞

superset-frontend/src/features/alerts/AlertReportModal.tsx
puridach-w: @eschutho would you kindly provide your opinion on this matter?

eschutho: Thanks for the videos. I'll defer to a product or design person on the functionality. @yousoph do you think the default subject line should appear in the email subject field?

eschutho: Sorry for the delayed response here, but the rest is looking good to me. Thanks for all the contributions here @puridach-w!

yousoph:
Hi @puridach-w ! Thanks for this new feature, it's looking good 

About the subject line UI, I think we can make it a required field in the form, but filled out with the default report name when the form is opened and it can update as the user changes the inputs. If the user clears it they will see that the field is required. There should be patterns on the modal that you can look to for behavior of required fields (red asterisk for example), but let me know if there's anything you'd like help defining! 

Thanks!

geido:
Hello @puridach-w, feel free to ping me when you are ready for another round of review. Thanks!

puridach-w:
> http://18.246.236.152:8080



> /testenv up FEATURE_ALERT_REPORTS=true

FYI, I still haven't edited the code from yousoph comment. But I appreciate your care and attention. Thank you very much.

geido:
Np @puridach-w. There are a few conflicts as I just merged a PR that changes the styles of the A&R modal. I hope that does not become too much of a hassle. Let me know!

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: Perhaps the placeholder could show the default if empty? This would also allow for users to revert back to default if they've saved a report with a name, but want to return to the Superset default subject line.

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -973,7 +981,8 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
 
     notificationSettings.forEach(setting => {
       if (!!setting.method && setting.recipients?.length) {
-        hasInfo = true;
+        hasInfo =
+          setting.method === 'Email' ? !!setting.email_subject?.length : true;
```
rtexelm, line 985: I think this is causing [the valid report unit test](https://github.com/apache/superset/blob/d6dbb95cc45360ce24778811ab5a6d5f9a4198e6/superset-frontend/src/features/alerts/AlertReportModal.test.tsx#L255) to fail. Might be causing your CI error. Do you need to push `email_subject` to `settings` in `onNotificationAdd`?

superset-frontend/src/features/alerts/AlertReportModal.tsx
puridach-w: Ping @eschutho @yousoph, What are your thoughts on reverting to displaying the default value in the placeholder? This way, users can easily revert to the default if they accidentally delete the email_subject. For example, if we make the "email_subject" field a required field and the user deletes it, it could be harder for them to restore the default value.

geido: I think it makes sense

yousoph: Sounds good to me

puridach-w:
Ping @eschutho @geido I request another review of my PR? Thank you!

geido:
Hey @puridach-w doing some manual testing. It seems that the spacing is off here

Screenshot 2024-03-11 at 16 41 35

puridach-w:
> Hey @puridach-w if you look at the testenv above, you will see that there is a` :` in the placeholder. I think something might be broken in the placeholder logics.
> 
> Screenshot 2024-03-19 at 19 44 18

Hi @geido , I've updated the placeholder logic, but now there are some failing checks in the [test-postgres-hive]. I haven't made changes to that area could you help me with the problem?

puridach-w:
Ping @geido Could you please help me review this PR again?

geido:
@puridach-w not sure if this is related to your PR or not but I cannot add a report. It thinks recipients are not entered even though I did. Can you verify that, please?

puridach-w:
> @puridach-w not sure if this is related to your PR or not but I cannot add a report. It thinks recipients are not entered even though I did. Can you verify that, please? 

Sure let me check

puridach-w:
Hi @geido , as I checked, I found that the issue come from the updateNotificationSetting function in the AlertReportModal file. The problem arises because the function does not update the state with the new setting value, resulting in the recipient variable doesn't have a value. I've fixed this by adding an else statement to update the setting value.

geido:
Hey @puridach-w this is looking much better now. However, I noticed that you can add an empty space as a valid value for the subject. Should we probably strip it or allow for a minimum amount of chars? Wdyt?

puridach-w:
> Hey @puridach-w this is looking much better now. However, I noticed that you can add an empty space as a valid value for the subject. Should we probably strip it or allow for a minimum amount of chars? Wdyt?

Hi @geido. I agree and have added the condition to trap that case. Could you please review it again? There's a failed job under 'Build & Publish Docker Images / docker-build', but I noticed there is an open PR aimed at resolving the issue.

geido (approved):
Thanks for all the hard work on this!

PR #26328, opened 2023-12-21
refactor: Removes the Filter Box code

michael-s-molina:
### SUMMARY
As part of the [4.0 approved initiatives](https://github.com/orgs/apache/projects/292/views/1), this PR removes the Filter Box code and it's associated dependencies `react-select` and `array-move`. It also removes the `DeprecatedSelect` and `AsyncSelect` components that were exclusively used by filter boxes. Existing filter boxes will be automatically migrated to native filters.

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


john-bodley:
@michael-s-molina thanks for tackling this. I think there's more dead code which could be removed, i.e., the [changeFilter](https://github.com/search?q=repo%3Aapache%2Fsuperset%20changeFilter&type=code) logic seems to be filter-box related. I also suspect that the entirety of [superset-frontend/src/dashboard/actions/dashboardFilters.js](https://github.com/apache/superset/blob/5e85f5c81f66b2da6fa11d9112216004f35170c8/superset-frontend/src/dashboard/actions/dashboardFilters.js) et al. could be removed.

UPDATING.md
```diff
@@ -24,6 +24,8 @@ assists people when migrating to a new version.
 
 ## Next
 
+- [26328](https://github.com/apache/superset/issues/26328): Removes the deprecated Filter Box code. To be able to upgrade, you need to first migrate your deprecated filter boxes using the `superset native-filters upgrade` command available in previous versions of Superset.
```
john-bodley, line 27:
```suggestion
- [26328](https://github.com/apache/superset/issues/26328): Removes the deprecated filter-box visualization type code. To be able to upgrade, you need to first migrate your deprecated filter-box charts using the `superset native-filters upgrade` command available in previous versions of Superset.
```

michael-s-molina: The `available in previous versions of Superset` is important because 4.0 will not contain the script to migrate filter boxes.

michael-s-molina:
> @michael-s-molina thanks for tackling this. I think there's more dead code which could be removed, i.e., the [changeFilter](https://github.com/search?q=repo%3Aapache%2Fsuperset%20changeFilter&type=code) logic seems to be filter-box related. I also suspect that the entirety of [dashboardFilters.js](https://github.com/apache/superset/blob/5e85f5c81f66b2da6fa11d9112216004f35170c8/superset-frontend/src/dashboard/actions/dashboardFilters.js) et al. could be removed.

@john-bodley I also thought this was related but in fact it's used by the feature that shows which filters are applied to a chart.

john-bodley:
@michael-s-molina I also thought that, however the `changeFilter` method doesn’t fire (confirmed by the lack of the logged event) and thus I suspect there’s some additional dead code there.

michael-s-molina:
> @michael-s-molina I also thought that, however the changeFilter method doesn’t fire (confirmed by the lack of the logged event) and thus I suspect there’s some additional dead code there.

You're correct. The `changeFilter` is not being called.

> I also suspect that the entirety of [dashboardFilters.js](https://github.com/apache/superset/blob/5e85f5c81f66b2da6fa11d9112216004f35170c8/superset-frontend/src/dashboard/actions/dashboardFilters.js) et al. could be removed.

The `addFilter` and `removeFilter` are being called when you add/remove charts to/from the dashboard.

michael-s-molina:
@john-bodley `changeFilter` is called from some plugins.

https://github.com/apache/superset/assets/70410625/b7953d94-da6f-4ca4-bc06-02e77c4badbd

justinpark (approved):
LGTM

john-bodley:
@michael-s-molina both this PR and https://github.com/apache/superset/pull/26329 handle aspects of removing filter-box charts in favor of the native dashboard filters. These PRs seem highly coupled and per my comments in the later I wonder if there's any incentive to combine them and/or (for review purposes) make one PR are precursor (base branch) of the other?

Additionally, though necessary, I don't believe these two PRs are sufficient for the removal of the legacy filter box charts, i.e., a migration is necessary. The [native filter](https://github.com/apache/superset/blob/master/superset/cli/native_filters.py) CLI commands should provide the necessary logic which (thankfully) has no dependencies on any of the changes outlined in these two PRs.

I'm not sure where best to add this, be it in a separate PR, or including with the proposed logic if combined into a single PR, but it seems there should be an Alembic migration added where the upgrade is akin to:

```python
from superset.cli import native_filters


def upgrade() -> None:
    native_filters.upgrade(["--all"], standalone_mode=False)
    native_filters.cleanup(["--all"], standalone_mode=False)


def downgrade() -> None:
    pass
```

michael-s-molina:
> I'm not sure where best to add this, be it in a separate PR, or including with the proposed logic if combined into a single PR, but it seems there should be an Alembic migration added where the upgrade is akin to:

I added a migration script to forcefully migrate legacy filter boxes. I copied the logic from the cli command into the migration file and made the adjustments such as binding the session and using pagination. This allows us to encapsulate filter box dependencies inside a migration script and remove the cli command.

john-bodley:
@michael-s-molina regarding, 

> I added a migration script to forcefully migrate legacy filter boxes. I copied the logic from the CLI command into the migration file and made the adjustments such as binding the session and using pagination. This allows us to encapsulate filter box dependencies inside a migration script and remove the CLI command.

There's still the need to cleanup the dashboard metadata (the temporary fields allows one to downgrade) as well as removal of the filter box charts. 

I'm not sure if it makes more sense to make this migration either:

- Irreversible (akin to https://github.com/apache/superset/pull/26328#issuecomment-1881931298 where I included the cleanup logic in the upgrade and made the downgrade a no-op—this was under the assumption that keeping around dangling filter-box charts would be problematic without the supporting code). 
- Reversible (akin to the logic you currently have) with a future migration in 4.1 or 5.0 which executes the cleanup logic.

If we opt for the later, the issue is that the cleanup logic no longer exists. My sense is we will need to preserve the `cleanup` portion of the CLI command—though delete the `upgrade`/`downgrade` logic.

superset-frontend/package.json
```diff
@@ -192,7 +191,6 @@
     "react-reverse-portal": "^2.1.1",
     "react-router-dom": "^5.3.4",
     "react-search-input": "^0.11.3",
-    "react-select": "^3.2.0",
```
rusackas, line 195: YAAAAAAAAAAAAYYYYY!!!!

rusackas (approved):
Approving cypress changes as codeowner. I ♥️ this PR!

john-bodley:
@eschutho, @michael-s-molina, and @villebro with regards to https://github.com/apache/superset/pull/26328#issuecomment-1885303583 and the decision to make the migration irreversible I pushed https://github.com/apache/superset/pull/26328/commits/6e513c1e94215afbe411205bf0890151ede1d11e which:

1. Updates the migration logic to remove the need to cleanup at a later time, i.e., the dashboard metadata will no longer be polluted with temporary fields and the obsolete filter-box charts will be deleted.
2. Ensures that imported dashboard are upgraded. Any filter-box charts imported alongside the dashboards will be deleted afterwards—they are initially needed a priori in order to upgrade the dashboard*.
3. Ensures that imported filter-box charts (outside the context of importing a dashboard) are ignored*.

\* @eschutho I'm not overly familiar with the import/export logic as this is something that we don't support at Airbnb. I _hope_ the logic I've implemented is correct, but any additional 👀 would be greatly appreciated.

eschutho:
This is great. I'll ping @betodealmeida for a look at the import script, too.

michael-s-molina:
Thanks @john-bodley for the migration improvements!

superset/commands/dashboard/importers/v1/utils.py
```diff
@@ -182,8 +182,6 @@ def import_dashboard(
                 logger.info("Unable to encode `%s` field: %s", key, value)
 
     dashboard = Dashboard.import_from_dict(session, config, recursive=False)
-    if dashboard.id is None:
-        session.flush()
```
betodealmeida, line 186: Why is this being removed? We currently need it for `ImportAssetsCommand` and `ImportExamplesCommand`.

betodealmeida (changes requested):
Just a concern about removing a `flush`, I'm worried it could break examples or importing assets.

superset/commands/dashboard/importers/v0.py
```diff
@@ -252,20 +255,30 @@ def alter_native_filters(dashboard: Dashboard) -> None:
...
+    )
 
-    dashboard_to_import.slices = new_slices
-    session.add(dashboard_to_import)
+    # Migrate any filter-box charts to native dashboard filters.
+    migrate_dashboard(dashboard)
     session.flush()
-    return dashboard_to_import.id  # type: ignore
+
+    # Remove all obsolete filter-box charts.
```
john-bodley, line 274: I originally updated `import_dashboard` and `import_chart` to return `Dashboard` and `Slice` objects respectively (as opposed to the associated ID), however this became a messy yak shaving experience. For the interest of minimizes the footprint of the change, we simply keep track of the IDs associated with the charts added and then refetch them from the session.

john-bodley:
@michael-s-molina remote CI is failing currently (it works locally) with the following error, 

```
FAILED tests/integration_tests/import_export_tests.py::TestImportExport::test_import_override_dashboard_slice_reset_ownership - AssertionError: admin user != gamma user
```

which was the reason for my initial yak shaving foray which consumed a number of cycles. [This](https://github.com/apache/superset/pull/26328/files#diff-ffdf55d23aa995fd31e11532c4a3f0a13279ac3ea6f8f15960aa2cd4042b7722R272) is the "culprit" line, though the the `migrate_dashboard` only modifies the `json_metadata` and `position_json` attributes and doesn't invoke any session level operations.

My "hunch" is (akin to our discussion previously) there's some quirk/gremlin with SQLAlchemy/Flask-AppBuilder. I've already confirmed that the re-imported dashboard references an existing dashboard, i.e., exists in the session, so ownership should not be reset . The reason why it works locally but not remotely still has me puzzled, but it _may_ have something to do with individual tests not being idempotent. Additionally if I add `print()` statements the test passes—go figure.

Give that this test for the legacy v0 dashboard importer I was wondering whether we just skip this test for now? I've likely already committed numerous hours to this PR and (due to other work requirements) likely couldn't revisit it until towards the end of next week.

superset/commands/dashboard/importers/v0.py
```diff
@@ -252,20 +255,30 @@ def alter_native_filters(dashboard: Dashboard) -> None:
...
+        db.session.query(Slice)
+        .filter(Slice.id.in_(old_to_new_slc_id_dict.values()))
+        .all()
+    )
 
-    dashboard_to_import.slices = new_slices
-    session.add(dashboard_to_import)
+    # Migrate any filter-box charts to native dashboard filters.
+    migrate_dashboard(dashboard)
     session.flush()
```
john-bodley, line 272: In actuality only the new (non-existing) dashboard strictly needs to be flushed (after being added to the Flask-SQLAlchemy session) to the database in order to have the dashboard ID assigned. I'm merely preserving the previous behavior by always flushing prior to returning.

michael-s-molina:
> Give that this test for the legacy v0 dashboard importer I was wondering whether we just skip this test for now? I've likely already committed numerous hours to this PR and (due to other work requirements) likely couldn't revisit it until towards the end of next week.

@john-bodley Given that this test is named `test_import_override_dashboard_slice_reset_ownership` and the previous one is `test_import_new_dashboard_slice_reset_ownership`, I just [copied the logic](https://github.com/apache/superset/pull/26328/commits/8e07c9e63862e69b900ecc424130c99f700ddf3a) from the previous test given that they should be similar tests with the only difference being a overridden dashboard instead of a new one.

tests/integration_tests/import_export_tests.py
```diff
@@ -542,9 +542,11 @@ def test_import_override_dashboard_slice_reset_ownership(self):
         self.assertEqual(imported_slc.owners, [gamma_user])
 
         # re-import with another user shouldn't change the permissions
-        g.user = admin_user
```
john-bodley, line 541: @michael-s-molina I don't believe your logic is correct, i.e., the original logic was actually correct. The intention of the test (as far as I can tell) is to import a dashboard into Superset (as the `gamma` user) and then reimport the same dashboard (as the `admin` user). Given that the dashboard existed the creator and owner should not change.

I _presume_ the `changed_by` doesn't change because none of the dashboard attributes (metadata) changed, but that's simply a hunch.

michael-s-molina: Let me know how you want to proceed here. I assumed the tests were similar because they have the exact names with the only difference being new vs override. Given that you wanted to skip the test, I would keep it as is.

john-bodley: @michael-s-molina I wonder if we should leave the test as-is and just skip it. It's the legacy version of the importer as well and thus it's _likely_ unused. Maybe @betodealmeida would have some thoughts regarding this?

betodealmeida: Yeah, it's a common pattern, where admins will manage dashboards under source control and periodically sync them to Superset. In that case you want to preserve the original ownership.

But like John said, if this is for the v0 importer (I'm on my phone, hard to see) then I wouldn't worry about it too much, since a lot of the behavior there is not strictly defined.

betodealmeida (approved):

tests/integration_tests/import_export_tests.py
michael-s-molina: @john-bodley I reverted and skipped the test.

superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
```diff
@@ -76,10 +76,7 @@ export interface ChartPropsConfig {
   annotationData?: AnnotationData;
   /** Datasource metadata */
   datasource?: SnakeCaseDatasource;
-  /**
-   * Formerly called "filters", which was misleading because it is actually
-   * initial values of the filter_box and table vis
-   */
+  /** Initial values */
```
kgabryje, line 79: Nit: maybe we should just remove this comment?

superset-frontend/src/components/TelemetryPixel/index.tsx
```diff
@@ -47,7 +47,7 @@ const TelemetryPixel = ({
   const pixelPath = `https://apachesuperset.gateway.scarf.sh/pixel/${PIXEL_ID}/${version}/${sha}/${build}`;
   return process.env.SCARF_ANALYTICS === 'false' ? null : (
     <img
-      referrerPolicy="no-referrer-when-downgrade"
+      referrerPolicy="no-referrer"
```
kgabryje, line 50: Is this change related to the pr or just a rebase mistake?

superset-frontend/src/dashboard/actions/dashboardState.js
```diff
@@ -581,7 +581,7 @@ export function addSliceToDashboard(id, component) {
     ]).then(() => {
       dispatch(addSlice(selectedSlice));
 
-      if (selectedSlice && selectedSlice.viz_type === 'filter_box') {
+      if (selectedSlice) {
```
kgabryje, line 584: If the code in the `if` block was meant to run only for filter box, shouldn’t we remove it?

superset-frontend/src/dashboard/actions/dashboardState.js
```diff
@@ -591,7 +591,7 @@ export function addSliceToDashboard(id, component) {
 export function removeSliceFromDashboard(id) {
   return (dispatch, getState) => {
     const sliceEntity = getState().sliceEntities.slices[id];
-    if (sliceEntity && sliceEntity.viz_type === 'filter_box') {
+    if (sliceEntity) {
```
kgabryje, line 594: Same as above

superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
michael-s-molina: Removed!

superset-frontend/src/components/TelemetryPixel/index.tsx
michael-s-molina: It was a rebase mistake! Reverted!

superset-frontend/src/dashboard/actions/dashboardState.js
michael-s-molina: Great call! It by removing this I was able to remove even more code!

superset-frontend/src/dashboard/actions/dashboardState.js
michael-s-molina: Removed!

kgabryje (approved):
Amazing work! LGTM

PR #26332, opened 2023-12-21
fix(embedded+async queries): support async queries to work with embedded guest user

zephyring:
### SUMMARY

This PR is aimed to solve problem where guest user cannot view an embedded dashboard when async queries feature flag is turned on.
Async queries first cache query_context into cache. Then later on the query_context is re-constructed from cache which contains original access context in form_data and queries to fetch actual data for chart.
Before this PR, the form_data is never set into cache so that we lost the original access context(which embedded dashboard original request is accessing), thus security_manager will always raise for access.
The second part of this PR set guest_token temporarily into job_metadata so that when job_metadata passes to celery task, the task can use it to set the current user context so as to load rls properly, which fix the consistency issue between sync and async invoke of the cache key(cache key is calculated with rls and rls is based on current guest user context)

Screenshot 2024-01-03 at 2 15 46 PM
Screenshot 2024-01-03 at 2 18 48 PM
Screenshot 2024-01-03 at 2 16 05 PM
Screenshot 2024-01-03 at 2 16 28 PM
Screenshot 2024-01-03 at 2 17 13 PM

### TESTING INSTRUCTIONS

1. turn on `GLOBAL_ASYNC_QUERIES` feature flag.
2. go to a dashboard and make it embedded.
3. view the embedded dashboard as a guest user(to logged in as guest user, you will need to call guest token API first and include the jwt in following API calls)
4. notice that API calls to the dashboard e.g. `/api/v1/chart/data`, `/superset/explore_json`, `/api/v1/async_event/`(if polling mode), `/api/v1/chart/data/{key}` will all succeed and charts get rendered.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/async_events/async_query_manager.py
+++ b/superset/async_events/async_query_manager.py
@@ -198,7 +198,7 @@ class AsyncQueryManager:
         if guest_user := security_manager.get_current_guest_user_if_guest():
             job_metadata["guest_token"] = guest_user.guest_token
         self._load_explore_json_into_cache_job.delay(
-            job_metadata,
+            job_metadata.copy(),
             form_data,
             response_type,
             force,
@@ -209,7 +209,10 @@ class AsyncQueryManager:
         return job_metadata
 
     def submit_chart_data_job(
-        self, channel_id: str, form_data: dict[str, Any], user_id: Optional[int]
+        self,
+        channel_id: str,
+        form_data: dict[str, Any],
+        user_id: Optional[int] = None,
     ) -> dict[str, Any]:
         # pylint: disable=import-outside-toplevel
         from superset import security_manager
@@ -221,7 +224,7 @@ class AsyncQueryManager:
         job_metadata = self.init_job(channel_id, user_id)
         if guest_user := security_manager.get_current_guest_user_if_guest():
             job_metadata["guest_token"] = guest_user.guest_token
-        self._load_chart_data_into_cache_job.delay(job_metadata, form_data)
+        self._load_chart_data_into_cache_job.delay(job_metadata.copy(), form_data)
         # clean up guest token so that it doesn't get exposed upper level
         if "guest_token" in job_metadata:
             del job_metadata["guest_token"]
--- a/superset/async_events/async_query_manager.py
+++ b/superset/async_events/async_query_manager.py
@@ -191,13 +191,21 @@ class AsyncQueryManager:
         force: Optional[bool] = False,
         user_id: Optional[int] = None,
     ) -> dict[str, Any]:
+        # pylint: disable=import-outside-toplevel
+        from superset import security_manager
+
         job_metadata = self.init_job(channel_id, user_id)
+        if guest_user := security_manager.get_current_guest_user_if_guest():
+            job_metadata["guest_token"] = guest_user.guest_token
         self._load_explore_json_into_cache_job.delay(
             job_metadata,
             form_data,
             response_type,
             force,
         )
+        # clean up guest token so that it doesn't get exposed upper level
+        if "guest_token" in job_metadata:
+            del job_metadata["guest_token"]
         return job_metadata
 
     def submit_chart_data_job(
@@ -214,6 +222,7 @@ class AsyncQueryManager:
         if guest_user := security_manager.get_current_guest_user_if_guest():
             job_metadata["guest_token"] = guest_user.guest_token
         self._load_chart_data_into_cache_job.delay(job_metadata, form_data)
+        # clean up guest token so that it doesn't get exposed upper level
         if "guest_token" in job_metadata:
             del job_metadata["guest_token"]
         return job_metadata
--- a/superset/async_events/async_query_manager.py
+++ b/superset/async_events/async_query_manager.py
@@ -203,8 +203,19 @@ class AsyncQueryManager:
     def submit_chart_data_job(
         self, channel_id: str, form_data: dict[str, Any], user_id: Optional[int]
     ) -> dict[str, Any]:
+        # pylint: disable=import-outside-toplevel
+        from superset import security_manager
+
+        # if it's guest user, we want to pass the guest token to the celery task
+        # chart data cache key is calculated based on the current user
+        # this way we can keep the cache key consistent between sync and async command
+        # so that it can be looked up consistently
         job_metadata = self.init_job(channel_id, user_id)
+        if guest_user := security_manager.get_current_guest_user_if_guest():
+            job_metadata["guest_token"] = guest_user.guest_token
         self._load_chart_data_into_cache_job.delay(job_metadata, form_data)
+        if "guest_token" in job_metadata:
+            del job_metadata["guest_token"]
         return job_metadata
 
     def read_events(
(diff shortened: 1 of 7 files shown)


superset/common/query_context_processor.py
```diff
@@ -600,7 +600,15 @@ def get_payload(
...
                 cache_manager.cache,
                 cache_key,
-                {"data": self._query_context.cache_values},
+                {
+                    "data": {
+                        # setting form_data into query context cache value as well
+                        # so that it can be used to reconstruct form_data field
+                        # for query context object when reading from cache
+                        "form_data": self._query_context.form_data,
+                        **self._query_context.cache_values,
```
zephyring, line 609: to be used by this later https://github.com/apache/superset/blob/master/superset/security/manager.py#L1898C1-L1898C1

superset/tasks/async_queries.py
```diff
@@ -106,10 +112,10 @@ def load_explore_json_into_cache(  # pylint: disable=too-many-locals
 ) -> None:
     cache_key_prefix = "ejr-"  # ejr: explore_json request
 
-    user = (
-        security_manager.get_user_by_id(job_metadata.get("user_id"))
```
zephyring, line 110: if user_id is None, get_user_by_id() will crash the celery task

schmiman:
Looking forward to this PR being merged

superset/tasks/async_queries.py
```diff
@@ -140,7 +152,13 @@ def load_explore_json_into_cache(  # pylint: disable=too-many-locals
                 "response_type": response_type,
             }
             cache_key = generate_cache_key(cache_value, cache_key_prefix)
-            set_and_log_cache(cache_manager.cache, cache_key, cache_value)
+            cache_instance = cache_manager.cache
+            cache_timeout = (
+                cache_instance.cache.default_timeout if cache_instance.cache else None
```
zephyring, line 157: using cache_instance default_timeout if found. Otherwise it will fallback to config["CACHE_DEFAULT_TIMEOUT"]

superset/async_events/async_query_manager.py
```diff
@@ -191,20 +191,43 @@ def submit_explore_json_job(
...
     ) -> dict[str, Any]:
+        # pylint: disable=import-outside-toplevel
+        from superset import security_manager
+
         job_metadata = self.init_job(channel_id, user_id)
+        if guest_user := security_manager.get_current_guest_user_if_guest():
+            job_metadata["guest_token"] = guest_user.guest_token
         self._load_explore_json_into_cache_job.delay(
-            job_metadata,
+            job_metadata.copy(),
```
eschutho, line 201: @zephyring if you like this pattern, it's a bit more succinct:
```
self._load_explore_json_into_cache_job.delay(
    {**job_metadata, "guest_token": guest_user.guest_token} 
    if (guest_user := security_manager.get_current_guest_user_if_guest()) else job_metadata.copy(),
    form_data,
    response_type,
    force,
)
```

superset/common/query_context_processor.py
eschutho: @villebro do you have any thoughts on this?

eschutho: @zephyring from where does the raise_for_access get called when it is passing in the query context from cache?

superset/tasks/async_queries.py
```diff
@@ -106,10 +112,16 @@ def load_explore_json_into_cache(  # pylint: disable=too-many-locals
...
+    if user_id := job_metadata.get("user_id"):
+        # logged in user
+        user = security_manager.get_user_by_id(user_id)
+    elif guest_token := job_metadata.get("guest_token"):
+        # embedded guest user
+        user = security_manager.get_guest_user_from_token(guest_token)
+        del job_metadata["guest_token"]
+    else:
+        # default to anonymous user if no user is found
+        user = security_manager.get_anonymous_user()
```
eschutho, line 124: it looks like we can dry up this logic to assign a user. I know there's been some discussions in the past about this, maybe @jfrag1 or @john-bodley may have some opinions, but I think we could even move this to the security manager. I see a lot of different ways that we can fetch users like this across the app, but none that combine them like this. It's possible that this user definition could be used in other places.

superset/common/query_context_processor.py
zephyring: The first call to `/chart/data` endpoint will set the query context into cache and return async job metadata.
Then later once job is done frontend will receive a job ready event with the result_url something like (`/api/v1/chart/data/qc-2772209e0a4b1ca513375118855b27af`).
Then it uses that to call `chart/data/<cache>` endpoint which invokes the [ChartCommand](https://github.com/apache/superset/blob/master/superset/charts/data/api.py#L307) and the command validation will call the raise_for_access

superset/tasks/async_queries.py
zephyring: I feel like setting the user context here is still async queries domain knowledge so a bit hesitate to generalize it to security_manager

eschutho: Ok, this makes sense.

superset/async_events/async_query_manager.py
```diff
@@ -191,20 +191,40 @@ def submit_explore_json_job(
...
     ) -> dict[str, Any]:
+        # pylint: disable=import-outside-toplevel
+        from superset import security_manager
+
         job_metadata = self.init_job(channel_id, user_id)
         self._load_explore_json_into_cache_job.delay(
-            job_metadata,
+            {**job_metadata, "guest_token": guest_user.guest_token}
+            if (guest_user := security_manager.get_current_guest_user_if_guest())
+            else job_metadata,
```
villebro, line 201: As we have the same logic in two places, could we DRY this out into a simple local util `_get_job_metadata(job_metadata)` to avoid divergence down the line?

superset/common/query_context_processor.py
```diff
@@ -600,7 +600,15 @@ def get_payload(
             set_and_log_cache(
                 cache_manager.cache,
                 cache_key,
-                {"data": self._query_context.cache_values},
+                {
+                    "data": {
+                        # setting form_data into query context cache value as well
+                        # so that it can be used to reconstruct form_data field
+                        # for query context object when reading from cache
+                        "form_data": self._query_context.form_data,
```
villebro, line 608: I have concerns here.. We've discussed this at length previously, and I feel we should always consider `form_data` to be chart specific internal state. IIRC, the original reason that form data is being passed along to the celery task is because the legacy charts require it to function. However, the v1 charts shouldn't need it. I know many fields within `form_data` are sort of standardized, but the whole point of `QueryContext`/`QueryObject` was to standardize the chart data request contract, thus decoupling the viz plugins from the backend. Therefore, I feel any field that is required from `form_data` should ideally be allocated a dedicated property on either the query context or query object.

villebro (commented):
First pass comments. It may be a good idea to schedule a 30 min discussion to find consensus on the form data issue, as it may well be I'm missing something obvious here.

villebro (approved):
LGTM - we discussed this today with @zephyring , and concluded that refactoring `form_data` out of the GAQ query context is outside the scope of this PR.

PR #26333, opened 2023-12-21
fix(logging): Add logging of change_dashboard_filter event for native dashboard filters

john-bodley:
### SUMMARY

For filter-box charts whenever the chart filters are changed we [log an event](https://github.com/apache/superset/blob/5e85f5c81f66b2da6fa11d9112216004f35170c8/superset-frontend/src/dashboard/components/gridComponents/Chart.jsx#L275-L278) per the `LOG_ACTIONS_CHANGE_DASHBOARD_FILTER` action—containing the filter-box chart ID and the associated filtered dataset columns—which is useful for performance logging.

The native dashboard filters log no such event and thus this PR simply logs the same action whenever anyone clicks the `APPLY FILTERS` button. I omitted both the chart ID and columns from the payload. The former is obsolete and the later is somewhat irrelevant as (unlike a filter-box chart) there are multiple datasets/columns impacted. Simply logging the event is suffice for tracking when the filters were updated resulting in a re-render of the dashboard.

### TESTING INSTRUCTIONS

Verified the action was logged to the `logs` table when the `APPLY FILTERS` button was clicked:

```
sqlite> SELECT json FROM logs WHERE json_extract(json, '$.event_name') = 'change_dashboard_filter' ORDER BY dttm;
{"source": "dashboard", "source_id": 2, "impression_id": "z8yL1F6XI", "version": "v2", "ts": 1703189280582, "event_name": "change_dashboard_filter", "event_type": "user", "event_id": "UXk2HJFqs", "visibility": "visible"}
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
@@ -44,6 +44,8 @@ import { getInitialDataMask } from 'src/dataMask/reducer';
 import { URL_PARAMS } from 'src/constants';
 import { getUrlParam } from 'src/utils/urlUtils';
 import { useTabId } from 'src/hooks/useTabId';
+import { logEvent } from 'src/logger/actions';
+import { LOG_ACTIONS_CHANGE_DASHBOARD_FILTER } from 'src/logger/LogUtils';
 import { FilterBarOrientation, RootState } from 'src/dashboard/types';
 import { checkIsApplyDisabled } from './utils';
 import { FiltersBarProps } from './types';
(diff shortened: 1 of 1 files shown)


superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
```diff
@@ -223,6 +225,7 @@ const FilterBar: React.FC<FiltersBarProps> = ({
   }, [dashboardId, dataMaskAppliedText, history, updateKey, tabId]);
 
   const handleApply = useCallback(() => {
+    dispatch(logEvent(LOG_ACTIONS_CHANGE_DASHBOARD_FILTER, {}));
```
michael-s-molina, line 228: Maybe `LOG_ACTIONS_CHANGE_DASHBOARD_FILTERS` (plural)?

michael-s-molina (approved):
LGTM. Left a non-blocking suggestion.

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
john-bodley: @michael-s-molina I though there was merit in keeping the action name the same for logging consistency.

Note that previously a single filter-box chart could actually have multiple filters so the singular term was also likely misleading then as well.

PR #26344, opened 2023-12-22
refactor: Removes the deprecated ENABLE_EXPLORE_JSON_CSRF_PROTECTION feature flag

michael-s-molina:
### SUMMARY
As part of the [4.0 approved initiatives](https://github.com/orgs/apache/projects/292/views/1), this PR removes the `ENABLE_EXPLORE_JSON_CSRF_PROTECTION` feature flag.

The previous value of the feature flag was `False` and now the feature is permanently removed.

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/views/core.py
```diff
@@ -239,8 +239,6 @@ def explore_json_data(self, cache_key: str) -> FlaskResponse:
             return json_error_response(utils.error_msg_from_exception(ex), 400)
 
     EXPLORE_JSON_METHODS = ["POST"]
-    if not is_feature_enabled("ENABLE_EXPLORE_JSON_CSRF_PROTECTION"):
```
geido, line 242: As this was `False` by default, shouldn't we keep the `EXPLORE_JSON_METHODS.append("GET")` ?

michael-s-molina: Nice catch! @dpgaspar can you confirm the expected behavior?

michael-s-molina: @geido Confirmed and added the `GET` method by default as you suggested.

dpgaspar: To my understanding this feature flag would allow `explore_json` to accept HTTP GETs but still imposing CSRF protection. So when `False` this endpoint would only accept HTTP POSTs and so would be automatically protected to CSRF. Shouldn't we keep the same behaviour and just allow for POST and be always protected?

michael-s-molina: > To my understanding this feature flag would allow explore_json to accept HTTP GETs but still imposing CSRF protection.

This feature flag was `False` by default, allowing `GET`. There's no other control than excluding `GET` when this feature flag is `True`.

dpgaspar: missed the `not` on the feature flag check. Makes sense

dpgaspar (approved):

geido (approved):

PR #26351, opened 2023-12-23
fix: stringify scarf pixel value

eschutho:
### SUMMARY

In order to match the expression here: https://github.com/apache/superset/blob/master/superset-frontend/src/components/TelemetryPixel/index.tsx#L48
we need to cast the env var for SCARF_ANALYTICS to a string in the webpack config.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sfirke (approved):

rusackas:
I fear this might cause as much ambiguity as it solves, but I'll do some testing to confirm...

rusackas (approved):

PR #26353, opened 2023-12-25
fix(SelectControl): select zero value

rekilina:
### SUMMARY
If SelectControl has options with numeric values, then the value 0 can be selected.
In this case, if defaultValue is not set or it is set to undefined, then currentValue will be set to undefined instead of 0.
This fix allows correctly save SelectControl with selected option, which value equals to 0, when the default value is undefined.

### TESTING INSTRUCTIONS
1. Create SelectControl instance in the controlPanel and set choices with numeric value:
[[0, "label_1"],  [1, "label_2"]]
Don't specify default value or set default value to undefined.
2. Select option with value 0 and save chart.
3. Reload page and check, whether the SelectControl has selected value. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/SelectControl.jsx
+++ b/superset-frontend/src/explore/components/controls/SelectControl.jsx
@@ -215,7 +215,7 @@ export default class SelectControl extends React.PureComponent {
 
     const getValue = () => {
       const currentValue =
-        value ||
+        value ??
         (this.props.default !== undefined ? this.props.default : undefined);
 
       // safety check - the value is intended to be undefined but null was used


villebro (approved):
Great catch! I would have asked if you could add a test, but I see this is one of those old non-TS pieces of code that likely doesn't have good tests to begin with. So LGTM.

villebro:
@rekilina I didn't have my laptop handy yesterday, but I noticed now that there's `SelectControl.test.jsx` where you should be able to add a test for this. Would you be able to take a stab at that?

villebro:
@rekilina I'll go ahead and merge this one and follow up with a test PR, as the test suite is arguably pretty messy in its current state. But if you want to add the test, let me know by tagging me here so I can avoid opening a new PR for it.

PR #26355, opened 2023-12-25
fix: Trino - handle table not found in SQLLab

Khrol:
### SUMMARY

SQLAlchemy doesn't strictly define what `get_indexes` method should do when table is not found.
Some dialects return the empty list while some of them raise `NoSuchTableError`.

When `NoSuchTableError` error is thrown, API returns 500 when non-existing table details are requested.
This PR fixes this issue and returns 404 in such cases.

### TESTING INSTRUCTIONS

Paste some non-existing table name in table name field and hit Enter:

Screenshot 2023-12-25 at 18 33 10

Expected: no 500 errors from the backend.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


john-bodley:
Thanks @Khrol for the PR. It seems—as in [this](https://github.com/apache/superset/blob/9033e72679fad7efdabc0f5b28607d974a7b5e1f/superset/db_engine_specs/bigquery.py#L322-L342) example—that the expected behavior for the `extra_table_metadata()` method is to return an empty dictionary if no indexes exist (assuming the underlying SQLAlchemy dialect returns an empty list as opposed to throwing a `NoSuchTableError` error.

From a consistency standpoint—given that the [Trino Python client](https://github.com/trinodb/trino-python-client/blob/2a60ac97852ed567a65070e4fc823f673dadeaf8/trino/sqlalchemy/dialect.py#L296-L313) throws a  `NoSuchTableError` error whereas [PyHive](https://github.com/dropbox/PyHive/blob/d4ae481675ac5588ba9101596fa26f22ef0e77c4/pyhive/sqlalchemy_presto.py#L184-L205) likely returns an empty list—it seems like the preferred fix would be to augment the [`TrinoEngineSpec.extra_table_metadata()`](https://github.com/apache/superset/blob/9033e72679fad7efdabc0f5b28607d974a7b5e1f/superset/db_engine_specs/trino.py#L54C1-L101) method to handle the `NoSuchTableError` error.

superset/db_engine_specs/trino.py
```diff
@@ -58,7 +58,10 @@ def extra_table_metadata(
         table_name: str,
         schema_name: str | None,
     ) -> dict[str, Any]:
-        metadata = {}
+        metadata: dict[str, Any] = {}
+
+        if not database.has_table_by_name(table_name):
```
john-bodley, line 63: @Khrol the issue with this method is it only checks physical tables (as opposed to tables and views).

Is there any reason not to implement my previous suggestion and override the `TrinoEngineSpec.get_indexes()` method? In addition to working both for tables and views  itwould actually be more efficient as well as it requires one fewer checks, i.e., asking for forgiveness (via a `try`/`except` block) instead of for permission (via the `has_table_by_name()` method).

tests/integration_tests/databases/api_tests.py
```diff
@@ -1340,6 +1340,22 @@ def test_get_invalid_table_table_extra_metadata(self):
         self.assertEqual(rv.status_code, 200)
         self.assertEqual(data, {})
 
+    def test_get_invalid_table_table_extra_metadata_trino(self):
```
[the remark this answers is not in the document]
john-bodley: If I were you, I would add it to `tests/integration_tests/db_engine_specs/test_trino.py`. Assuming you augment the `get_indexes()` method—per my previous comment—you can see other test cases where the `get_indexes()` method is being tested. You'll have to mock SQLAlchemy's `get_indexes()` method in a similar fashion to [this](https://github.com/apache/superset/blob/ea4e9a6403086d4e1de84e8197c08e8114dceeed/tests/integration_tests/db_engine_specs/base_engine_spec_tests.py#L530-L550) test.

superset/db_engine_specs/trino.py
Khrol: > Is there any reason not to implement my previous suggestion and override the TrinoEngineSpec.get_indexes() method?

I was a bit lost between `TrinoEngineSpec`, `Database` and `Inspector`. 🤦‍♂️  Some methods are static, others are not... 

I understand what you mean now. I'll update PR.

superset/db_engine_specs/trino.py
```diff
@@ -395,3 +396,26 @@ def get_columns(
...
+    def get_indexes(
+        cls,
+        database: Database,
+        inspector: Inspector,
+        table_name: str,
+        schema: str | None,
+    ) -> list[dict[str, Any]]:
+        """
+        Get the indexes associated with the specified schema/table.
+        Trino dialect raises NoSuchTableError in get_indexes if table is empty.
```
john-bodley, line 410:
```suggestion

        Trino dialect raises NoSuchTableError in get_indexes if table is empty.
```

john-bodley (commented):
LGTM

john-bodley (approved):

PR #26368, opened 2023-12-27
chore(docs): point to correct StackOverflow page

sfirke:
### SUMMARY
A BugHerd comment (No. 116) pointed out that one of the links to StackOverflow is wrong.  This fixes that, along with another not-optimal link to SO.  It also adds a link to GitHub discussions, encouraging people to look there to see if their question has been answered before.

No code modified.


michael-s-molina (approved):
LGTM

PR #26369, opened 2023-12-27
refactor: Removes the filters set feature

michael-s-molina:
### SUMMARY
As part of the [4.0 approved initiatives](https://github.com/orgs/apache/projects/292/views/1), this PR removes the Filters Set feature including the deprecated `DASHBOARD_NATIVE_FILTERS_SET` feature flag and all related API endpoints. The feature is permanently removed as it was not being actively maintained, it was not widely used, and it was full of bugs. We also considered that if we were to provide a similar feature, it would be better to re-implement it from scratch given the amount of technical debt that the current implementation has.

The previous value of the feature flag was `False` and now the feature is permanently removed.

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2023-12-27_13-14_59a1450b3c10_drop_filter_sets_table.py
```diff
@@ -14,17 +14,30 @@
...
+    "superset.migrations.versions.2021-03-29_11-15_3ebe0993c770_filterset_table"
+)
+
+
+def upgrade():
+    module.downgrade()
+
+
+def downgrade():
+    module.upgrade()
```
villebro, line 43: Nice, it never occurred to me we could do something like this 👍

villebro (approved):
LGTM. Slightly sad to see this go, but maybe we can reuse our learnings from this POC if this feature request comes up again.

superset/migrations/versions/2023-12-27_13-14_59a1450b3c10_drop_filter_sets_table.py
john-bodley: I ❤️ DRY code.

PR #26371, opened 2023-12-27
docs: fix config webdriver snippet in install on K8s

### LINKED ISSUE #26370: docs: Install a supported webdriver in the Celery worker
The instructions in the [Install a supported webdriver in the Celery worker](https://github.com/apache/superset/blob/863f8bbbcdd078814973d444368c12e06ad0c0c0/docs/docs/installation/running-on-kubernetes.mdx#L301) section in running-on-kubernetes.mdx cause the `chromedriver` installation to fail when configuring superset workers.

#### How to reproduce the bug

1. Configure helm `values.yaml` as per the guide by adding the following snippet:
```
  supersetWorker:
    # Install chromedriver used to send Alerts and Reports
    command:
      - /bin/sh
      - -c
      - |
        # Install chrome webdriver
        # See https://github.com/apache/superset/blob/4fa3b6c7185629b87c27fc2c0e5435d458f7b73d/docs/src/pages/docs/installation/email_reports.mdx
        apt update
        wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
        apt install -y --no-install-recommends ./google-chrome-stable_current_amd64.deb
        wget https://chromedriver.storage.googleapis.com/88.0.4324.96/chromedriver_linux64.zip
        unzip chromedriver_linux64.zip
        chmod +x chromedriver
        mv chromedriver /usr/bin
        apt autoremove -yqq --purge
        apt clean
        rm -f google-chrome-stable_current_amd64.deb chromedriver_linux64.zip
  
        # Run
        . {{ .Values.configMountPath }}/superset_bootstrap.sh; celery --app=superset.tasks.celery_app:app worker

```
2. Deploy Superset on K8s
4. Check celery worker logs: `kubectl logs -f superset-worker-<xxx>`
5. See multiple errors and warnings:
```
22 packages can be upgraded. Run 'apt list --upgradable' to see them.
/bin/sh: 4: wget: not found

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
E: Unsupported file ./google-chrome-stable_current_amd64.deb given on commandline
/bin/sh: 6: wget: not found
/bin/sh: 7: unzip: not found
chmod: cannot access 'chromedriver': No such file or directory
mv: cannot stat 'chromedriver': No such file or directory

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

0 upgraded, 0 newly installed, 0 to remove and 22 not upgraded.

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

### Expected results

Clean logs and `chromodriver successfully installed as per below logs:
```
Archive:  chromedriver_linux64.zip
  inflating: chromedriver  
```

### Actual results

`chromedriver` installation fails

### Environment

- superset version: `3.0.1`

### Checklist

### Additional context

The errors are being caused by `wget` and `unzip` not being present in the container image and the warnings are due to the use of `apt` rather than `apt-get`.

dbaltor:
### SUMMARY
The current example of installing `webdriver` on worker nodes when installing Superset on Kubernetes causes the configuration to fail. This PR is fix the issue.

### TESTING INSTRUCTIONS
1. Configure the Helm `values.yaml` file as per the new snippet included in this PR
2. Deploy Superset on K8s
3. Check celery worker logs: `kubectl logs -f superset-worker-<xxx>`

### ADDITIONAL INFORMATION
Fixes #26370 

- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/installation/running-on-kubernetes.mdx
+++ b/docs/docs/installation/running-on-kubernetes.mdx
@@ -310,15 +310,17 @@ supersetWorker:
     - |
       # Install chrome webdriver
       # See https://github.com/apache/superset/blob/4fa3b6c7185629b87c27fc2c0e5435d458f7b73d/docs/src/pages/docs/installation/email_reports.mdx
-      apt update
+      apt-get update
+      apt-get install -y wget
       wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
-      apt install -y --no-install-recommends ./google-chrome-stable_current_amd64.deb
+      apt-get install -y --no-install-recommends ./google-chrome-stable_current_amd64.deb
       wget https://chromedriver.storage.googleapis.com/88.0.4324.96/chromedriver_linux64.zip
+      apt-get install -y zip
       unzip chromedriver_linux64.zip
       chmod +x chromedriver
       mv chromedriver /usr/bin
-      apt autoremove -yqq --purge
-      apt clean
+      apt-get autoremove -yqq --purge
+      apt-get clean
       rm -f google-chrome-stable_current_amd64.deb chromedriver_linux64.zip
 
       # Run


sfirke:
Thanks for the issue & fix!  This is not my most familiar area so hopefully someone else will review.

From [reading a comparison ](https://aws.amazon.com/compare/the-difference-between-apt-and-apt-get/) it sounds like `apt` is the modern successor to `apt-get` -- this PR goes in the other direction which feels odd given that.  Could it be made to work with `apt` / is there something about your environment that is making these `apt` calls fail?  This PR could use more context re: why `apt-get` is better here for all users.

dbaltor:
Hi @sfirke. Thanks for taking time to review this PR. 
The script runs within the Superset worker pod which is using the original container image picked by the project ( I haven't touched it). The errors are caused because the two required tools, `wget` and `zip`, don't come installed by default so I updated the example to install them.

Regarding the use of `apt-get` vs `apt`, the [official recommendation](https://unix.stackexchange.com/a/590703) is to use the former within scripts due to its strict backward-compatibility policy whilst `apt`, being more user-friendly, is supposed to be used by humans. This is confirmed by the tool itself which prints in the logs the message below when used in scripts:
```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

sfirke (approved):
LGTM

sfirke:
Thank you for the explanation!  And for the fix.

Superset's k8s deployment is in a gray area, it has many users but is not "officially supported" yet due to lack of developer maintenance and oversight from the community.  So your contribution is especially appreciated!

dbaltor:
My pleasure. Will try to help as much as I can. Btw, happy holidays! :tada:

sfirke:
You too! 😁

PR #23865, opened 2023-04-28
fix: Native time range filter in legacy charts

kgabryje:
### SUMMARY
Some legacy charts (e.g. Heatmap and Bar Chart) were not getting dashboard time range filters applied with GENERIC_CHART_AXES enabled. This PR manually converts the time range into an adhoc filter, similarly how it's done for v1 charts.

Additionally, this PR brings back the time section for Calendar Heatmap chart, which was broken without it.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
Before:

image

After:

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26372, opened 2023-12-27
refactor: Removes the deprecated GENERIC_CHART_AXES feature flag

michael-s-molina:
### SUMMARY
As part of the [4.0 approved initiatives](https://github.com/orgs/apache/projects/292/views/1), this PR removes the deprecated `GENERIC_CHART_AXES` feature flag.

The previous value of the feature flag was `True` and now the feature is permanently enabled.

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


yousoph:
We have some small improvements/fixes that we are planning to work on to help unblock some of our users from being able to enable the GENERIC_CHART_AXES flag. It would be appreciated if we could hold on merging this  PR near the end of the breaking change window for Superset 4.0 so that we can get our changes completed first :) Thanks!!

rusackas:
Cypress Tests LGTM.... I can hit the code owner approval button, but might like to have a more holistic review/approval first.

Vitor-Avila:
hey @michael-s-molina one concern with the `GENERIC_CHART_AXES` removal, is that currently with the FF set to `True` the **TIME GRAIN** control for legacy charts is no longer visible:

image

Reported in https://github.com/apache/superset/issues/26474.

Existing charts will still work, but users are no longer able to change the time grain configuration if needed. Also, new legacy charts can't be properly created without this configuration. 

Given we're not yet automatically migrating all legacy charts to ECharts, the full removal of the FF would introduce this issues to Organizations still using legacy charts. 

I suspect we can fix this by setting the `time_grain_sqla`'s visibility to return `true` in case we don't have an `x_axis` set in `controls` (in `superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/sharedControls.tsx`). Something like this: https://github.com/apache/superset/compare/master...Vitor-Avila:superset:time-grain-legacy-charts?expand=1

Unsure if we would want to fix the bug first, and then make sure it's working properly before we move forward with this removal PR, or if you would actually prefer to include the fix here. I'm also further testing that branch to confirm if additional changes would be needed.

Let me know your thoughts! Happy to submit a PR from that branch.

michael-s-molina:
Hi @Vitor-Avila. You're correct, we should solve this in a specific PR and merge it before merging this PR. This problem was reported before and fixed for each chart that does not correctly handle both feature flag states. Check https://github.com/apache/superset/pull/26439 and https://github.com/apache/superset/pull/23865 for more context.

villebro:
I suggest we introduce two new shared time sections for legacy charts:
- one that works with temporal charts (=has time grain)
- one that works with non-temporal charts

Essentially these would be the same as `genericTime` and `legacyRegularTime`, but without the check for `hasGenericChartAxes`. These could then be used on legacy charts that don't want to support generic x-axis. These could then be removed once the last legacy chart has been removed.

michael-s-molina:
@villebro @Vitor-Avila Can any of you open a PR to fix this problem? Maybe we can even ask help from @sivasathyaseeelan. I have many things on my plate right now 😢

Vitor-Avila:
I feel like @villebro's suggestion seems more granular and scalable, but I'm still ramping up so I might not have the knowledge to implement it. 

I thought this was mostly related with how we're checking if `time_grain_sqla` should be visible in the chart controls, but seems like it's also related with each visualization type configuration.

villebro:
@Vitor-Avila this should be a fairly straight forward change, and I can support you if you wish. But if you prefer I can open a PR and tag you for a review.

sivasathyaseeelan:
> @villebro @Vitor-Avila Can any of you open a PR to fix this problem? Maybe we can even ask help from @sivasathyaseeelan. I have many things on my plate right now 😢

@michael-s-molina I can help you in this if you give me permission

Vitor-Avila:
hey @villebro I was checking both the line and bar charts from nvd3 (the charts I reproduced the missing time grain issue), and I believe they both use the `legacyTimeseriesTime`:

``` ts
export const legacyTimeseriesTime: ControlPanelSectionConfig = {
  ...baseTimeSection,
  controlSetRows: [
    ['granularity'],
    ['granularity_sqla'],
    ['time_grain_sqla'],
    ['time_range'],
  ],
};
```

I suspect the issue here is that the visibility settings for `time_grain_sqla` actually doesn't account for the scenario in which `GENERIC_CHART_AXES` is enabled but the chart doesn't have an X Axis (a legacy temporal chart):

``` ts
const time_grain_sqla: SharedControlConfig<'SelectControl'> = {
  type: 'SelectControl',
  label: TIME_FILTER_LABELS.time_grain_sqla,

  ...

  visibility: ({ controls }) => {
    if (!hasGenericChartAxes) {
      return true;
    }

    const xAxis = controls?.x_axis;
    const xAxisValue = xAxis?.value;
    if (isAdhocColumn(xAxisValue)) {
      return true;
    }
    if (isPhysicalColumn(xAxisValue)) {
      return !!(xAxis?.options ?? []).find(
        (col: ColumnMeta) => col?.column_name === xAxisValue,
      )?.is_dttm;
    }
    return false;
  },
};

```

Is your suggestion to create these two new shared time sections so that legacy nvd3 charts also support X-Axis (categorical) axes? We can also sync tomorrow via Slack.

superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -191,35 +188,41 @@ const config: ControlPanelConfig = {
...
-                      ensureIsArray(controls?.groupby?.options).map(option => [
-                        option.column_name,
-                        option.is_dttm,
-                      ]),
-                    );
+          {
+            name: 'time_grain_sqla',
+            config: {
+              ...sharedControls.time_grain_sqla,
+              visibility: ({ controls }) => {
```
geido, line 195: Not related to this PR. These are bunch of repeated logics that we should probably centralize at some point

superset-frontend/src/explore/components/controls/DateFilterControl/DateFilterLabel.tsx
```diff
@@ -401,7 +401,7 @@ export default function DateFilterLabel(props: DateFilterControlProps) {
           ref={labelRef}
         />
       </Tooltip>
-      {/* the zIndex value is from trying so that the Modal doesn't overlay the AdhocFilter when GENERIC_CHART_AXES is enabled */}
+      {/* the zIndex value is from trying so that the Modal doesn't overlay the AdhocFilter */}
```
geido, line 404: Unrelated: Not sure I get the meaning of the sentence in this comment

geido (approved):

superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
michael-s-molina: True. I also noticed many repeated code between ECharts plugins. I'm delaying removing this repetition until after the migrations because I think we'll have a better overview of the common patterns.

superset-frontend/src/explore/components/controls/DateFilterControl/DateFilterLabel.tsx
zhaoyongjie: @geido The z-index value comes from _trying_, I'm not sure what's value is an appreciate value when I made this function.

michael-s-molina: I think the developer meant that the z-index value was set, after multiple tentative values, to make sure the modal doesn't overlay the filter.

Agree it's confusing 🤷🏼‍♂️

michael-s-molina: The problem here is that we don't have a z-index registry which would automatically resolve this type of issue. Maybe we'll have one some day.

michael-s-molina:
@mistercrunch It seems the CI step called Label Migration PR is [failing](https://github.com/apache/superset/actions/runs/7673342918/job/20915775011?pr=26372). Could you take a look?

rusackas:
Woohoo!!!!

PR #19078, opened 2022-03-09
feat: add permalink to dashboard and explore

villebro:
### SUMMARY

This PR adds new permanent link functionality that makes it possible to persist Dashboard and Explore state in the metadata database. The state is persisted in a new metadata table called `key_value`, which is able to store pickleable binary objects (can be used in the future for other resources, too). The new permalinks look as follows:
- Dashboard: `http://host/superset/dashboard/p/<key>/`
- Explore: `http://host/superset/explore/p/<key>/`

The dashboard state is a `dict` with the following fields:
- `filterState`: native filter state
- `hash` (optional): the anchor (used for direct links to charts/tabs)
- `urlParams` (optional): any URL params that are active, excluding the reserved URL params like `native_filter_key` etc. This is also where filter box state is persisted. Note that these are stored as a list of key-value tuples to support having multiple same keys (not currently supported by the chart data API but implemented like this for future proofing)

The Explore state is also a `dict` with the following fields:
- `formData` the chart form data
- `urlParams` (optional): any URL params that are active, excluding the reserved URL params like `form_data_key`, `slice_id` etc. (same format as in the dashboard state)

The benefits compared to the deprecated `shorturl` functionality:
- Leverages existing resource security mechanism: if you don't have access to a chart/dashboard, you are also not able to view permalink state for those charts/dashboards
- Uses UUID-based keys by default to make permalinks unguessable. However, where needed, numeric serial keys (similar to the legacy shorturl keys) can be used by flipping a config flag. Do note, however, that changing this setting will cause old links to stop working, i.e. it's not possible to retrieve permalink state based by its UUID key if the feature has been configured to work with the serial id and vice versa.
- Is not restricted by the URL length limitations that affects shorturls (however, Filter Box state is still subject to URL overflows).

Notice that old shorturls will still work like before - but links created after this will be based on the new permalink feature.

Detailed list of changes:
- Rename current `key_value` package to `temporary_cache` to better reflect the ephemeral nature of the state (not guaranteed to be persisted)
- Add new `key_value` table to metadata database along with relevant commands for adding, updating, getting and deleting key-value pairs
- Add `permalink` endpoints (POST and GET) to both Explore and Dashboard that leverage the new `key_value` store
- Add `PERMALINK_KEY_TYPE` to `config.py` which makes it possible to switch between serial and UUID based keys (UUID = default)
- fully support persisting filter box state and url parameters in permalink state (this was currently mostly broken)
- remove old shorturl creation functionality from frontend and backend
[...]

PR #26377, opened 2023-12-28
refactor: Removes the deprecated redirect endpoint

michael-s-molina:
### SUMMARY
As part of the [4.0 approved initiatives](https://github.com/orgs/apache/projects/292/views/1), this PR removes the deprecated Redirect API that supported short URLs (`/r`) and the `url` metadata table used to store them that was used before the permalink feature. Users lost the ability to generate R links ~1.5 years ago which seems sufficient time to remove the endpoint. It looks like we forgot to officially deprecate the endpoint in `config.py` when working on 3.0 but https://github.com/apache/superset/pull/19078 has the following message in `UPDATING.md` which is more relevant from a communication standpoint as opposed to the `# deprecated` comment in `config.py`:

> - [19078](https://github.com/apache/superset/pull/19078): Creation of old shorturl links has been deprecated in favor of a new permalink feature that solves the long url problem (old shorturls will still work, though!). By default, new permalinks use UUID4 as the key. However, to use serial ids similar to the old shorturls, add the following to your `superset_config.py`: `PERMALINK_KEY_TYPE = "id"`.

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


tests/integration_tests/tags/commands_tests.py
```diff
@@ -112,6 +112,7 @@ def test_delete_tags_command(self):
                 TaggedObject.object_id == example_dashboard.id,
                 Tag.type == TagType.custom,
             )
+            .order_by(Tag.name)
```
john-bodley, line 115: What does this change have to do with the redirect logic?

john-bodley (commented):
@michael-s-molina although the redirect endpoint is deprecated and the functionality for creating `/r` links no longer exists, I wonder if we should _try_ to preserve these URLs (for some period) by:

1. Migrate the records from the `url` to the `key_value` table where the `resource` column is `redirect` and the `value` column is a JSON key/value pair mapping `url.id` to `url.url`. These records could have an expiration of 1–2 years given the somewhat transient nature of URLs.
2. Keep the `/r` link endpoint which—albeit inefficiently—scans the `key_value` table (by leveraging JSON SQL UDFs) where the `resource` column is `redirect` and the `value` column contains the `/r` ID as they key.
3. Remove the deprecated RESTful HTTP endpoints in a far future version.

Alternatively, if code bloat is less of a concern, we could maybe punt this PR (as is) until v6.0 (or later) given the increased likelihood that the legacy redirects would be unused.

Note both options suffer from the same achilles heel in the sense that there’s no guarantee that an institution upgrade cadence will reflect that of the release cycle, hence users may feel short changed when the redirects no longer exist.

tests/integration_tests/tags/commands_tests.py
michael-s-molina: Nothing. It was something I caught while running the tests. It's a flaky test given that order is not guaranteed.

michael-s-molina:
> @michael-s-molina although the redirect endpoint is deprecated and the functionality for creating `/r` links no longer exists, I wonder if we should _try_ to preserve these URLs (for some period) by:
> 
> 1. Migrate the records from the `url` to the `key_value` table where the `resource` column is `redirect` and the `value` column is a JSON key/value pair mapping `url.id` to `url.url`. These records could have an expiration of 1–2 years given the somewhat transient nature of URLs.
> 2. Keep the `/r` link endpoint which—albeit inefficiently—scans the `key_value` table (by leveraging JSON UDFs) where the `resource` column is `redirect` and the `value` column contains the `/r` ID as they key.
> 
> Alternatively, if code bloat is less of a concern, we could maybe punt this PR (as is) until v6.0 (or later) given the increased likelihood that the legacy redirects would be unused.

If we're going to keep the `/r` endpoint for some period, I think is easier to keep the values in the `url` table, avoid the migrations, and punt this proposal to 5.0.

eschutho:
I don't have a strong opinion on whether to keep the urls, but I did notice that it doesn't look like the endpoint was ever marked as deprecated. Not a blocker but something to keep in mind for future.

michael-s-molina:
> I don't have a strong opinion on whether to keep the urls, but I did notice that it doesn't look like the endpoint was ever marked as deprecated. Not a blocker but something to keep in mind for future.

@eschutho I updated the PR description with more context. Let me know if that's sufficient to resolve your concern.

john-bodley (approved):

yousoph:
@michael-s-molina do you have examples handy of what a full short chart URL and what a short dashboard URL looked like before? 

Thanks!

superset/migrations/versions/2024-01-19_10-03_e863403c0c50_drop_url_table.py
```diff
@@ -0,0 +1,45 @@
...
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+module = import_module("superset.migrations.versions.2016-01-13_20-24_8e80a26a31db_")
+
+
+def upgrade():
+    module.downgrade()
```
dpgaspar, line 39: interesting approach

dpgaspar (approved):

Issue #26390, opened 2024-01-02
[SIP-114] Proposal for Google Sheets Export


jessie-ross:
### Motivation

One-click export to Google Sheets is an important usability improvement that's desired at my company, as well as generally by Superset users:
- https://github.com/apache/superset/issues/10332
- https://github.com/apache-superset/superset-roadmap/issues/99
- https://github.com/apache/superset/issues/8840
- https://github.com/apache/superset/issues/7904
- https://github.com/apache/superset/issues/6796

Uploading the data directly to Google Sheets removes a barrier for less technical users, allowing them to work with data in a more familiar way, and is a major workflow improvement for users that have to frequently upload data to Google Sheets to share.

### Proposed Change

I propose adding Google Sheets as an export option for SqlLab results and Reports, as well as providing a bit more of a framework for future export extensions, both for the UI and code. 

[See the working implementation in this PR](https://github.com/apache/superset/pull/26391).

**New UI:**

![Screenshot 2024-01-02 at 4 25 34 pm](https://github.com/apache/superset/assets/971886/a9f85f6b-5ae9-42a7-b30a-d0938b40fe4a)

https://github.com/apache/superset/assets/971886/c495dcb4-84ee-4d5c-9353-decefd71cbf0

![Screenshot 2024-01-12 at 3 35 18 pm](https://github.com/apache/superset/assets/971886/b3661bb4-53c2-4f92-a587-a49211e7a1da)

### New or Changed Public Interfaces

- **UI:** Replacing "Download to CSV" with "Export" dropdown menu in the results toolbar.
- **UI:** New `/export/:clientId/google-sheets/` page that performs the export and navigates to the Google Sheet.
- **UI:** Add "Send as Google Sheets link" option for Chart Reports.
- **API Endpoint**: New `/api/v1/sqllab/export/<string:client_id>/google-sheets/`
- **API Endpoint**: For existing CSV endpoint:
   - Added matching `/api/v1/sqllab/export/<string:client_id>/csv/`
   - Keeping existing `/api/v1/sqllab/export/<string:client_id>/` (I would like to deprecate this somehow, I've changed all internal references.)
- **Feature flag:** `GOOGLE_SHEETS_EXPORT`.
- **Configuration:** `GOOGLE_SHEETS_EXPORT_SERVICE_ACCOUNT_JSON_PATH`.
- **Configuration:** `GOOGLE_SHEETS_EXPORT_SHARE_PERMISSIONS`.
[...]

PR #26391, opened 2024-01-02
feat: [SIP-114] Google Sheets Export (part 1)

jessie-ross:
### SUMMARY

**This PR implements https://github.com/apache/superset/issues/26390, providing one-click Google Sheets Export.**

I've marked this as a WIP but it's a fully working implementation, missing is:
- [x] Testing.
- [x] Need to look into the documentation system.

I've split the SIP implementation into two PRs:
- (This PR) Export route, lib code, SqlLab export implementation.
- Adding Google Sheets to Alerts and Reports.

**Before:**
Screenshot 2024-01-02 at 5 08 30 pm

**After**
Screenshot 2024-01-02 at 5 08 43 pm

**Video Demo**

https://github.com/apache/superset/assets/971886/78504a9f-727e-4a9e-a278-bb58a547f025

### TESTING INSTRUCTIONS

See the added docs:
1. Follow setup instructions in the docs.
2. Run a query in SqlLab.
3. Click the "Google Sheets" button under the new "Export" dropdown.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/26390
- [x] Required feature flags:
   - `GOOGLE_SHEETS_EXPORT`
- [x] Changes UI


superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -283,10 +320,15 @@ const ResultSet = ({
...
             )}
-            {csv && (
-              <Button buttonSize="small" href={getExportCsvUrl(query.id)}>
-                <i className="fa fa-file-text-o" /> {t('Download to CSV')}
-              </Button>
+            {hasExports && (
+              <AntdDropdown overlay={ExportMenu}>
+                <Button>
+                  <Space>
+                    Export
```
justinpark, line 327:
```suggestion
                    {t('Export')}
```

superset-frontend/src/pages/ExportGoogleSheets/index.tsx
```diff
@@ -0,0 +1,76 @@
+/**
```
justinpark, line 1: This blank page appears to be superfluous. It might be more efficient to incorporate this logic within the SQL Lab page and only display the page for the Google Sheets link if the operation is successful.

superset-frontend/src/pages/ExportGoogleSheets/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+    dispatch(
+      addInfoToast(t('Exporting results to Google Sheets'), { duration: 0 }),
+    );
+    SupersetClient.get({
+      endpoint: `/api/v1/sqllab/export/${clientId}/google-sheets/`,
+    })
+      .then(res => setState({ data: res.json, isLoading: false, error: null }))
+      .catch(e => setState({ data: null, isLoading: false, error: e }));
+  }, [dispatch, clientId]);
```
justinpark, line 49: You can simplify this logic by using [rdk-query](https://redux-toolkit.js.org/tutorials/rtk-query).

```suggestion
const gSheetApi = api.injectEndpoints({
  endpoints: builder => ({
    export: builder.query({
      query: ({ clientId }) => ({
        endpoint: `/api/v1/sqllab/export/${clientId}/google-sheets/`,
        transformResponse: ({ json }) => json,
      })
    }),
  }),
});
export const { useExportQuery } = gSheetApi;

```

superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -264,6 +272,35 @@ const ResultSet = ({
...
+        onClick: () =>
+          window.open(getExportCsvUrl(query.id), '_blank')?.focus(),
+      });
+      if (isFeatureEnabled(FeatureFlag.GOOGLE_SHEETS_EXPORT)) {
+        exportMenuItems.push({
+          label: t('Google Sheets'),
+          key: 'google-sheets',
+          icon: <Icons.GoogleOutlined />,
+          onClick: () =>
+            window.open(getExportGoogleSheetsUrl(query.id), '_blank')?.focus(),
```
justinpark, line 290: This onClick handler should have the logic for gSheetApi request.
```suggestion
     const [trigger, { data, isLoading, isError }] = useLazyExportQuery();
     ---
     onClick: () => trigger({ clientId: getExportGoogleSheetsUrl(query.id) }),
```

superset-frontend/src/views/routes.tsx
```diff
@@ -127,6 +127,13 @@ const Profile = lazy(
   () => import(/* webpackChunkName: "Profile" */ 'src/pages/Profile'),
 );
 
+const ExportGoogleSheets = lazy(
+  () =>
+    import(
+      /* webpackChunkName: "ExportGoogleSheets" */ 'src/pages/ExportGoogleSheets'
+    ),
+);
+
```
justinpark, line 136:
```suggestion
```

superset-frontend/src/views/routes.tsx
```diff
@@ -245,6 +252,12 @@ if (isFeatureEnabled(FeatureFlag.TAGGING_SYSTEM)) {
     Component: Tags,
   });
 }
+if (isFeatureEnabled(FeatureFlag.GOOGLE_SHEETS_EXPORT)) {
+  routes.push({
+    path: '/export/:clientId/google-sheets/',
+    Component: ExportGoogleSheets,
+  });
+}
```
justinpark, line 260:
```suggestion
```

justinpark (changes requested):
Thank you, @jessie-ross, for contributing to the new feature. I have provided some feedback in relation to the frontend components, specifically concerning the new user interface for the Google Sheets API loading page.

superset-frontend/src/pages/ExportGoogleSheets/index.tsx
jessie-ross: My reasoning is that using an separate page provides a better user experience, and helps decouple the code.

It prevents popup blocker issues, but importantly, this process can take some time, especially as the backend might need to rerun a query for export:
https://github.com/apache/superset/blob/9033e72679fad7efdabc0f5b28607d974a7b5e1f/superset/commands/sql_lab/export.py?plain=1#L112-L126

Our databases (SparkSQL mainly) are particularly slow, and then adding on top of that expensive queries and large amounts of data that need to be moved, we're looking at a 1m average wait time, with the upper bounds being around 15m.

By having this process occur on a separate page it allows the export to happen asynchronously, and for the user to continue their use of Superset while the export occurs.
Notably the CSV export already behaves like this - the download starts asynchronously to the UI.

Simple alternatives would be a modal, or hover popup, suffering from the long loading time issue.

A more complex alternative would be to perform the export on the backend, and provide a notification to the user when it's done. I'm not sure if Superset has this kind of thing set up.

jessie-ross: > A more complex alternative would be to perform the export on the backend, and provide a notification to the user when it's done. I'm not sure if Superset has this kind of thing set up.

I guess I could do this in the front-end with toasts actually, that would be a viable alternative to me.

But I still prefer the UX and simplicity of the separate page.

jessie-ross: > I guess I could do this in the front-end with toasts actually, that would be a viable alternative to me.

After experimenting with this I've found that Superset navigation can unintuitively trigger a full page reload, e.g. navigating from "SQL Lab" to "Datasets", meaning that the front-end export process would be lost.

This could be mitigated by a prevent-navigate-away alert, but I don't think that would be preferable UX, and we're heading further away from simplicity.

jessie-ross:
Hi @justinpark, just wondering if there was any feedback on this?

superset/sqllab/api.py
```diff
@@ -236,6 +240,7 @@ def format_sql(self) -> FlaskResponse:
             return self.response_400(message=error.messages)
 
     @expose("/export/<string:client_id>/")
```
eschutho, line 242: @jessie-ross Thanks for the great feature proposal. I was reading through the SIP, and thought I might suggest another option here which would be to use the existing api "/export/<string:client_id>/" and pass the export type with a query string `?format=csv` for example. That could potentially simplify this. It seems like the request/response schemas should be fairly similar, with just a switch in the command that is run.

jessie-ross: I'm happy to go either way on the url format, the two are equivalent:
```
/export/<client-id>/?format=csv
/export/<client-id>/?format=google-sheets
/export/<client-id>/   (csv, deprecated)
```

```
/export/<client-id>/csv/
/export/<client-id>/google-sheets/
/export/<client-id>/   (csv, deprecated)
```

eschutho: Great. Maybe another option is to make csv the default format if none is passed. Then we wouldn't need to deprecate the existing api.

jessie-ross: I'm happy to do that too, it is really an overall API design point, and I'm not sure how Superset does things.

In my mind deprecated doesn't mean it has to ever be removed, so it never has to be a breaking change.

superset-frontend/src/pages/ExportGoogleSheets/index.tsx
jessie-ross: @justinpark Since there has been no feedback on this, I'm going to go with the loading page as it seems to be the best option.

superset/sqllab/api.py
jessie-ross: Since csv is default if none is passed (meaning no breaking changes,) and the SIP includes the current format, I'm going to go with that.

jessie-ross:
It seems like @justinpark's code suggestions are relevant to approach we discussed [here](https://github.com/apache/superset/pull/26391#discussion_r1439663162), unless there is more to discuss it seems like we won't be going with that approach, so I have marked all the code suggestions as resolved.

aavanzyl:
Hi @jessie-ross, super eager for this feature. What is the possibility of also allowing this functionality to export to Google Sheets as well. Our use case is to daily update a sheet with the exported data for one of our departments.

jessie-ross:
> Hi @jessie-ross, super eager for this feature. What is the possibility of also allowing this functionality to export to Google Sheets as well. Our use case is to daily update a sheet with the exported data for one of our departments.

This SIP (Superset Improvement Proposal) covers on-demand export to a new sheet from SqlLab, and export to a new sheet from reports/alerts (e.g. you could do a daily report/alert that emails the new Google Sheet.)

Exporting to an existing sheet is a bit more complex, as we have to have to store the sheet we want to export to somewhere, and find a way to add that into the UI. It would likely require a new SIP.

sivankumar86:
Could you add support in report creation as well ? export the report to google sheet from schedule

sivankumar86:
https://github.com/apache/superset/discussions/29660

superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -316,6 +326,37 @@ const ResultSet = ({
         user?.roles,
       );
```
jessie-ross, line 327: I'm not sure if we missed an upgrade step in our installation but the view was named `SqlLab` instead of `SQLLab` in our database and we had to run `flask fab security-converge`.

jessie-ross:
> Could you add support in report creation as well

@sivankumar86 Something like this is in the SIP, I plan to add it in after this PR.

jessie-ross:
It's taken me a while to get back round to this but I've successfully tested this change in our work environment, I've fixed bugs around serialization of data, and I've improved the handling of permissions in the front-end to match other changes to the codebase.

This PR is now ready for review :)

superset-frontend/src/pages/ExportGoogleSheets/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+      return;
+    }
+    dispatch(
+      addInfoToast(t('Exporting results to Google Sheets'), { duration: 0 }),
+    );
+    SupersetClient.get({
+      endpoint: `/api/v1/sqllab/export/${clientId}/google-sheets/`,
+    })
+      .then(res => setState({ data: res.json, isLoading: false, error: null }))
+      .catch(e => setState({ data: null, isLoading: false, error: e }));
```
justinpark, line 48: The below process of parsing the error message can be standarized through getClientErrorObject.

```suggestion
      .catch(response => {
        getClientErrorObject(response).then(
          ({ error, message }: { error: any; message: string }) => {
            const errorMessage = error
              ? error.error || error.statusText || error
              : message;
            dispatch(addDangerToast(errorMessage));
          },
        );
      });
```

superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -310,12 +320,50 @@ const ResultSet = ({
...
+          },
+        });
+      }
+      if (isFeatureEnabled(FeatureFlag.GoogleSheetsExport) && canExportGoogleSheets) {
+        exportMenuItems.push({
+          label: t('Google Sheets'),
+          key: 'google-sheets',
+          icon: <Icons.GoogleOutlined />,
+          onClick: () =>
+            window.open(getExportGoogleSheetsUrl(query.id), '_blank')?.focus(),
```
justinpark, line 353: I hope to add the action logging like the above export csv.

```
```
```suggestion
            logAction(LOG_ACTIONS_SQLLAB_DOWNLOAD_GSHEET, {});
            window.open(getExportGoogleSheetsUrl(query.id), '_blank')?.focus(),
```

superset-frontend/src/pages/ExportGoogleSheets/index.tsx
jessie-ross: Addressed in https://github.com/apache/superset/pull/26391/commits/b537fcbc175ee18ff75284a373e9c93ad24dae16

jessie-ross:
@justinpark @rusackas Hi, this is still waiting on review.

justinpark (approved):
LGTM. Thanks for all hard-works!

PR #19841, opened 2022-04-25
fix: Table chart column config issue

codemaster08240328:
### SUMMARY

This PR resolved column config issue which happened when we update the chart.

### How To Check
1. Create a table chart with custom Metric.
2. Change the alignment in customize tab.
3. Update the custom Metric name. 
4. Update the chart.

Previously changed alignment should be kept in the updated chart.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

[[DEV] Add new chart - 25 April 2022 - Watch Video](https://www.loom.com/share/ceb1977b5ad841918dc0c4f192ab6dc9)

### TESTING INSTRUCTIONS

[[DEV] Add new chart - 25 April 2022 - Watch Video](https://www.loom.com/share/af61215d8e404199b3b8d3492f589c05)

### ADDITIONAL INFORMATION

- [x] Fix Table Chart Column Config Bug

PR #26393, opened 2024-01-02
fix(chart): Resolve incorrect column customization when switching metrics in table chart

soniagtm:
### SUMMARY

I have found that when attempting to swap the positions of metrics in the table chart with applied column configuration, the column configuration is switched to another column, which is related to the addition of #19841 for maintaining configurations when metrics are renamed. To address this issue, I have added a condition to first check if the column name is found in the old metrics. If it is found, it will avoid updating the column configuration. Reassignment will only occur if the new column name is not found among the existing ones to maintain configurations when metrics are renamed.

**Before:** Currency prefix is switched to SUM(quantity_ordered)

https://github.com/apache/superset/assets/96282005/8f83aa5f-2794-4f9a-adfb-2085293dc0f2

**After:** Currency prefix is applied to SUM(sales) as expected

https://github.com/apache/superset/assets/96282005/e17f0ed8-5dff-482e-9797-4edb0bcdf8d7

### TESTING INSTRUCTIONS

1. Create a table chart with dimensions and two or more metrics. 
2. Apply number formatting to any of the metrics. 
3. Attempt to swap the positions of the metrics.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/actions/exploreActions.test.js
+++ b/superset-frontend/src/explore/actions/exploreActions.test.js
@@ -21,6 +21,63 @@ import { defaultState } from 'src/explore/store';
 import exploreReducer from 'src/explore/reducers/exploreReducer';
 import * as actions from 'src/explore/actions/exploreActions';
 
+const METRICS = [
+  {
+    expressionType: 'SIMPLE',
+    column: {
+      advanced_data_type: null,
+      certification_details: null,
+      certified_by: null,
+      column_name: 'a',
+      description: null,
+      expression: null,
+      filterable: true,
+      groupby: true,
+      id: 1,
+      is_certified: false,
+      is_dttm: false,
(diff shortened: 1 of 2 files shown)


john-bodley:
Thanks @soniagtm for the fix. Would you mind adding a unit test to ensure we don't regress in the future?

soniagtm:
Hi, I've added test cases for switching metric columns and changing their names. I'm not sure if this is the correct way to do so. Please let me know if any corrections or further additions are needed. 

Thank you in advance!

john-bodley:
@justinpark would you mind reviewing this?

justinpark (approved):
LGTM. Thanks for adding test specs

PR #26397, opened 2024-01-03
feat: add custom sort by x-axis for mixed charts

cmedley:
### SUMMARY

Enables sorting mixed data chart by x-axis using different columns

Before:
image

After: 
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
@@ -47,7 +47,7 @@ export default function buildQuery(formData: QueryFormData) {
     ...formData,
   };
 
-  if (!!baseFormData.orderby) {
+  if (baseFormData.orderby) {
     baseFormData.x_axis_sort = formData.orderby.label;
     baseFormData.x_axis_sort_asc = !formData.order_desc_x_axis;
   }
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
@@ -35,6 +35,7 @@ import {
   rollingWindowOperator,
   timeCompareOperator,
   resampleOperator,
+  sortOperator,
 } from '@superset-ui/chart-controls';
 import {
   retainFormDataSuffix,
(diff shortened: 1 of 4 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
```diff
@@ -46,6 +47,11 @@ export default function buildQuery(formData: QueryFormData) {
     ...formData,
   };
 
+  if (!!baseFormData.orderby) {
+    baseFormData.x_axis_sort = formData.orderby.label;
```
villebro, line 51: I think you may need to do this here to make sure we can handle all types of metrics (please also add the import):
```suggestion
    baseFormData.x_axis_sort = getMetricLabel(formData.orderby);
```

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
```diff
@@ -46,6 +47,11 @@ export default function buildQuery(formData: QueryFormData) {
     ...formData,
   };
 
+  if (!!baseFormData.orderby) {
```
villebro, line 50: I'm curious, do we really need the double bang here?
```suggestion
  if (baseFormData.orderby) {
```

villebro (commented):
Thanks for adding this to Mixed chart - I know I often forget to implement new features from the regular chart to the mixed chart, so I'm sure this will be well received 👍 A few minor comments, LMKWYT. Also, tests are always very welcome, so if you feel up to either adding a new test or changing an existing one to incorporate the new functionality, that would be very much appreciated. Take a look at `plugins/plugin-chart-echarts/test/MixedTimeseries/buildQuery.test.ts` if you feel up for making sure this functionality doesn't break in the future 😉

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
```diff
@@ -75,6 +82,7 @@ export default function buildQuery(formData: QueryFormData) {
         time_offsets: isTimeComparison(fd, queryObject) ? fd.time_compare : [],
         post_processing: [
           pivotOperatorInRuntime,
+          sortOperator(fd, queryObject),
```
zhaoyongjie, line 85: The sortOperator should move before the flattenOperator because the other operators(rolling, resample, and so on) also changed original data.

rusackas:
@cmedley I'm just finding that this slipped between the cracks somehow. Sorry about that. If you're willing to rebase PR to resolve conflicts, we can probably get a fresh review on it, and get it merged (finally). Thanks!

rusackas:
Looks like this still hasn't gotten a rebase, and now the GENERIC_CHART_AXES flag used in the PR has been removed. I'll convert this to draft for now, in hopes that it can still be brought up to speed and marked ready for review again. If not, it may eventually be closed.

muralikrishnaIrakam:
if i try to sort with two or more Metrics then sort is not working , only one Metrics then it will work , can i get if you have any possible solution ..?

rusackas:
Closing due to contributor inactivity. Feel free to reopen if/when warranted, and thank you for the contribution!

Issue #26286, opened 2023-12-15
404 on GET and DELETE on tableschemaview for Admin role user with all permissions on TableSchemaView for 3.0.0


rajivpatki:
A clear and concise description of what the bug is:

I am the admin for Superset 3.0.0. When in SQL Lab I click on a table (from DuckDB), I get an error message at the bottom right corner that says "An error occurred while fetching table metadata" and on selecting two tables "An error occurred while removing the table schema. Please contact your administrator."

The logs show:

404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
superset              | 192.168.176.1 - - [15/Dec/2023:18:59:38 +0000] "GET /tableschemaview/dtvBbokbR HTTP/1.1" 404 0 "-" 
^ Same 404 error for DELETE on tableschemaview

#### How to reproduce the bug
This bug did not exist earlier for me and I could see a preview of the schema of the table I was requesting. 
1. Go to 'SQL Lab, clikc on database (select a DuckDB database that has create table, view, DML, exploration' enabled
2. Click on 'schema duck.main'
3. Click on any table > "An error occurred while fetching table metadata". Remove table from the selection > "An error occurred while removing the table schema. Please contact your administrator." The logs show 404 errors on tableschemaview for GET and DELETE

### Expected results
The table schema preview loads in the left pane of the SQL Lab

### Actual results
Error box in bottom right corner

### Environment

- browser type and version: MS Edge Version 120.0.2210.77 (Official build) (64-bit)
- superset version: `3.0.0`
- python version: `from docker image superset:3.0.0`
- any feature flags active: I have tried removing all caching and features from superset_config.py but the problem remains
### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [ .] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [ .] I have reproduced the issue with at least the latest released version of superset.
- [ .] I have checked the issue tracker for the same issue and I haven't found one similar.

### Additional context

- Superset is running with built in google OAuth behind a nginx reverse proxy 
- Queries, dashboard, charts, SQL Lab is working fine. This error is thrown only in SQL Lab

michael-s-molina:
Hi @rajivpatki. Does this error happen with the latest stable release 3.0.2?

rajivpatki:
> Hi @rajivpatki. Does this error happen with the latest stable release 3.0.2?

Yes, I still get both errors along with an additional 404:

[...]

PR #26405, opened 2024-01-04
fix(sqllab): Bump duckdb-engine version to 0.9.5

guenp:
### SUMMARY
Currently, when a user selects a DuckDB database, the schema dropdown shows multiple instances of `main`. This was an issue in the SQLAlchemy driver and has been fixed on the latest version of `duckdb_engine`.

### BEFORE/AFTER SCREENSHOTS

Before:
Screenshot 2024-01-03 at 4 32 23 PM

After:
Screenshot 2024-01-03 at 4 34 02 PM

### TESTING INSTRUCTIONS
- Add a DuckDB database to Superset
- Open SQLLab
- Select the duckdb database
- Click on the "schema" dropdown
- Schema dropdown should contain unique `db_name.schema_name` items instead of many `main` schema duplicates

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/26286
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -155,7 +155,7 @@ setup(
         "dremio": ["sqlalchemy-dremio>=1.1.5, <1.3"],
         "drill": ["sqlalchemy-drill==0.1.dev"],
         "druid": ["pydruid>=0.6.5,<0.7"],
-        "duckdb": ["duckdb-engine==0.9.2"],
+        "duckdb": ["duckdb-engine==0.9.5"],
         "dynamodb": ["pydynamodb>=0.4.2"],
         "solr": ["sqlalchemy-solr >= 0.2.0"],
         "elasticsearch": ["elasticsearch-dbapi>=0.2.9, <0.3.0"],
--- a/setup.py
+++ b/setup.py
@@ -155,7 +155,7 @@ setup(
         "dremio": ["sqlalchemy-dremio>=1.1.5, <1.3"],
         "drill": ["sqlalchemy-drill==0.1.dev"],
         "druid": ["pydruid>=0.6.5,<0.7"],
-        "duckdb": ["duckdb-engine>=0.9.5, <1.0"],
+        "duckdb": ["duckdb-engine>=0.9.5, <0.10"],
         "dynamodb": ["pydynamodb>=0.4.2"],
         "solr": ["sqlalchemy-solr >= 0.2.0"],
         "elasticsearch": ["elasticsearch-dbapi>=0.2.9, <0.3.0"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -155,7 +155,7 @@ def get_git_sha() -> str:
         "dremio": ["sqlalchemy-dremio>=1.1.5, <1.3"],
         "drill": ["sqlalchemy-drill==0.1.dev"],
         "druid": ["pydruid>=0.6.5,<0.7"],
-        "duckdb": ["duckdb-engine==0.9.2"],
+        "duckdb": ["duckdb-engine==0.9.5"],
```
villebro, line 158: As this is a 0.x version, should we restrict it to `<0.10` as per semver breaking change conventions?

villebro (commented):
LGTM, but wondering if we should add an upper bound? I know you maintain the driver, so you're the best person to say if this is a valid concern or not..

setup.py
john-bodley: @villebro for semver shouldn't the upper bound be `<1.0`?

Additionally @guenp providing a bound (as opposed to pinning to an exact version) is preferred as it improves the viable dependency set. It's not atypical for one to have an infeasible dependency set when one explicitly pins a dependency.

john-bodley, line 158:
```suggestion
        "duckdb": ["duckdb-engine>=0.9.5, <1.0"],
```

villebro: @john-bodley IIRC, in strict semver, the minor versions of the 0 major are considered breaking. In other words, 0.10 should not be assumed to be backwards compatible with 0.9. @michael-s-molina @eschutho I believe we've discussed this at length at some point, any pointers here would be valuable.

setup.py
```diff
@@ -155,7 +155,7 @@ def get_git_sha() -> str:
         "dremio": ["sqlalchemy-dremio>=1.1.5, <1.3"],
         "drill": ["sqlalchemy-drill==0.1.dev"],
         "druid": ["pydruid>=0.6.5,<0.7"],
-        "duckdb": ["duckdb-engine==0.9.2"],
+        "duckdb": ["duckdb-engine>=0.9.5, <1.0"],
```
john-bodley, line 158:
```suggestion
        "duckdb": ["duckdb-engine>=0.9.5, <0.10"],
```

setup.py
michael-s-molina: Here's the [semver definition](https://semver.org/#spec-item-4) about 0.x:

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

guenp: @john-bodley, I agree with you in general, but `"duckdb-engine==0.9.5"` and `"duckdb-engine>=0.9.5, <0.10"` are effectively identical in this case.

villebro (approved):
LGTM

PR #26429, opened 2024-01-09
fix(post-processing): handle missing values in cumulative operator

### LINKED ISSUE #21093: Cumsum Creating Discontinuous Lines
#### How to reproduce the bug

1. Create a cumsum with multiple series where some series do not have all of the dates
2. Plot Smooth Line

### Expected results

I would expect that when there are gaps in the date data, that the superset cumsum function creates a continuous line filling with the last value rather than producing NAs.

### Actual results

Lines break due to NAs in table when one series has a data point and the others don't.

#### Screenshots

image

### Environment

- Chrome latest version
- superset version: 1.5.0

villebro:
### SUMMARY
If your data contains nulls, using the cumsum operator will cause gaps in the chart. This is because null values appear as nulls in the series instead of the cumulative value. To avoid this, we replace nulls with zeros before applying the operator, ensuring that all values are present in the series.

### AFTER
Now gaps in the data won't cause the bar to be truncated:
image

### BEFORE
Previously, missing values would cause the cumulative series to be missing in the chart, causing gaps in the graph:
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes #21093
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/pandas_postprocessing/cum.py
+++ b/superset/utils/pandas_postprocessing/cum.py
@@ -46,6 +46,7 @@ def cum(
     """
     columns = columns or {}
     df_cum = df.loc[:, columns.keys()]
+    df_cum = df_cum.fillna(0)
     operation = "cum" + operator
     if operation not in ALLOWLIST_CUMULATIVE_FUNCTIONS or not hasattr(
         df_cum, operation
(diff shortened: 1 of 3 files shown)


kgabryje (approved):
LGTM

JohnDietrich-Pepper:
Ah this fix really broke our charting. We wanted to show discontinuous lines when the data is null because the data is null and NOT zero particularly when the time period hasn't occurred yet. Now Superset is assuming all null data is zero which is not the case.
@villebro @rusackas

PR #26443, opened 2024-01-10
feat: add chart id and dataset id to global logs

eschutho:
### SUMMARY

part two of the [context logs](https://github.com/apache/superset/pull/26418), this adds the chart and dataset ids to global context when the /data endpoints are called for use in the sql mutator.

An example of a sql_mutator override to log these values would be:

```python
def mutator(sql: str, **kwargs: Any) -> str:
    key = "6dcd92a04feb50f14bbcf07c661680ba"
    output = (
        f"-- {key}\n"
        + f"-- slice_id: {g.logs_context.get('slice_id') if hasattr(g,'logs_context') else ''}\n"
        + f"-- dataset_id:{g.logs_context.get('dataset_id') if hasattr(g,'logs_context') else ''}\n"
    )
    return output + f"{sql}\n-- {key}"
```

After with the mutator above:
USA_Births_Names

### TESTING INSTRUCTIONS

Update the sql mutator override, load a dashboard and click "view query" on a chart.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -427,15 +427,19 @@ class ChartDataRestApi(ChartRestApi):
     def _load_query_context_form_from_cache(self, cache_key: str) -> dict[str, Any]:
         return QueryContextCacheLoader.load(cache_key)
 
-    @logs_context(
-        context_func=lambda self, form_data: {
+    def _map_form_data_datasource_to_dataset_id(
+        self, form_data: dict[str, Any]
+    ) -> dict[str, Any]:
+        return {
             "dataset_id": form_data.get("datasource", {}).get("id")
             if isinstance(form_data.get("datasource"), dict)
-            and form_data.get("datasource").get("type") == DatasourceType.TABLE.value
+            and form_data.get("datasource", {}).get("type")
+            == DatasourceType.TABLE.value
             else None,
             "slice_id": form_data.get("form_data", {}).get("slice_id"),
         }
-    )
+
+    @logs_context(context_func=_map_form_data_datasource_to_dataset_id)
     def _create_query_context_from_form(
         self, form_data: dict[str, Any]
     ) -> QueryContext:
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -47,7 +47,13 @@ from superset.daos.exceptions import DatasourceNotFound
 from superset.exceptions import QueryObjectValidationError
 from superset.extensions import event_logger
 from superset.models.sql_lab import Query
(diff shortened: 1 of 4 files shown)


superset/utils/decorators.py
```diff
@@ -61,6 +65,82 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
...
+    """
+
+    def decorate(f: Callable[..., Any]) -> Callable[..., Any]:
+        def wrapped(*args: Any, **kwargs: Any) -> Any:
+            if not hasattr(g, "logs_context"):
+                g.logs_context = {}
+
+            # limit data that can be saved to logs_context
+            # in order to prevent antipatterns
+            available_logs_context_keys = [
```
hwmarkcheng, line 84: Would it make sense for this to be a parameter? Right now this is hard coded for query related logging, but it might be useful for other logging purposes. 

If we add the `available_logs_context_keys` as a parameter when the decorator is called we can still keep the functionality of limiting data but for that specific function.

eschutho: Yeah, that's a good question. In the SIP review we talked about trying to limit antipatterns of people putting random data into global context so the intention here is to use this list to limit what data can be passed in with the decorator, but there are definitely a lot of other use cases. I have three others on my todo list. I would suggest that if you think there's more data that should be considered for logging, then add it to the list in a PR and we'll see if it fits. That way we're still limiting the data that the decorator can accept, but we're not completely limited to just the set that is here now. How does that sound?

hwmarkcheng: Hey Elizabeth, that sounds great! As for now there's no other data we need aside from dataset_id/slice_id so this is good on our end. Was just pointing this out as a future suggestion.

superset/charts/data/api.py
```diff
@@ -421,6 +427,15 @@ def _get_data_response(
     def _load_query_context_form_from_cache(self, cache_key: str) -> dict[str, Any]:
         return QueryContextCacheLoader.load(cache_key)
 
+    @logs_context(
+        context_func=lambda self, form_data: {
+            "dataset_id": form_data.get("datasource", {}).get("id")
+            if isinstance(form_data.get("datasource"), dict)
+            and form_data.get("datasource").get("type") == DatasourceType.TABLE.value
+            else None,
+            "slice_id": form_data.get("form_data", {}).get("slice_id"),
```
betodealmeida, line 436: Can we move this to a separate function to increase readability?

geido (approved):

PR #17196, opened 2021-10-22
fix(bigquery): calculated column cannot orderby in BigQuery

zhaoyongjie:
### SUMMARY

closed: https://github.com/apache/superset/issues/17178

Calculated column cannot use in native filter when using Google BigQuery.  This is due to the column in `ORDER BY` must be in `SELECT` clause or `GROUP BY` clause. 

added a new class property `allows_hidden_cc_in_orderby` in `BaseEngineSpec` to fix this error.

#### a calculated column in orderby
image

### TESTING INSTRUCTIONS

Test this PR follow this issue: https://github.com/apache/superset/issues/17178

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/17178
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26461, opened 2024-01-11
fix(BigQuery): Support special characters in column/metric names used in ORDER BY

### LINKED ISSUE #26459: Chart does not work in case metric/dimension name used in order by has special characters for BQ connections
The `SORT BY` clause for BigQuery connections is not adding quotes to the column/metric name used, so in case it has special characters it throws an error.

#### How to reproduce the bug

1. Make sure you have a dataset powered by BQ.
2. Use it to create a table chart.
3. Add any dimension to it.
4. Create an ad-hoc metric using `COUNT(*)` and name it as `métric`.

### Expected results
The chart works properly.

### Actual results
An error is displayed:

https://github.com/apache/superset/assets/96086495/f84569ac-313a-40df-be09-f1f070eea436

Log: 
```
File "/Users/vitoravila/.pyenv/versions/3.9.0/envs/superset-mods/lib/python3.9/site-packages/google/api_core/retry.py", line 207, in retry_target
    result = target()
  File "/Users/vitoravila/.pyenv/versions/3.9.0/envs/superset-mods/lib/python3.9/site-packages/google/cloud/bigquery/job/query.py", line 1584, in do_get_result
    super(QueryJob, self).result(retry=retry, timeout=timeout)
  File "/Users/vitoravila/.pyenv/versions/3.9.0/envs/superset-mods/lib/python3.9/site-packages/google/cloud/bigquery/job/base.py", line 952, in result
    return super(_AsyncJob, self).result(timeout=timeout, **kwargs)
  File "/Users/vitoravila/.pyenv/versions/3.9.0/envs/superset-mods/lib/python3.9/site-packages/google/api_core/future/polling.py", line 261, in result
    raise self._exception
google.api_core.exceptions.BadRequest: 400 Syntax error: Illegal input character "\303" at [5:11]
```

### Environment
- browser type and version: Chrome latest version.
- superset version: Running from `master`.

### Checklist

Vitor-Avila:
### SUMMARY
The `BigQueryEngineSpec` has `allows_hidden_cc_in_orderby` set to `True`, so the `ORDER BY` statement uses the actual metric/dimension alias, rather than its SQL expression (implementation PR: https://github.com/apache/superset/pull/17196).

In case the metric alias has a special characters (like accents or Chinese characters for example), the query will throw an error.

Fixes https://github.com/apache/superset/issues/26459

**Before:**

https://github.com/apache/superset/assets/96086495/3db44c50-1fbb-4809-a123-cd4ea23f65ca

**After:**

https://github.com/apache/superset/assets/96086495/36a0745a-0f67-4130-934a-2abc77b0ff70

### TESTING INSTRUCTIONS
1. Create a dataset from a BQ connection.
2. Use it to create a **Table** chart.
3. Add any column as its dimension, and create a metric named `métric` using `COUNT(*)` as its SQL expression.

### ADDITIONAL INFORMATION
- [X] Has associated issue: https://github.com/apache/superset/issues/26459
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -1974,9 +1974,9 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
                 and db_engine_spec.allows_hidden_cc_in_orderby
                 and col.name in [select_col.name for select_col in select_exprs]
             ):
-                engine = self.database._get_sqla_engine()
-                quote = engine.dialect.identifier_preparer.quote
-                col = literal_column(quote(col.name))
+                with self.database.get_sqla_engine_with_context() as engine:
+                    quote = engine.dialect.identifier_preparer.quote
+                    col = literal_column(quote(col.name))
             direction = sa.asc if ascending else sa.desc
             qry = qry.order_by(direction(col))
 
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -1974,7 +1974,9 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
                 and db_engine_spec.allows_hidden_cc_in_orderby
                 and col.name in [select_col.name for select_col in select_exprs]
             ):
-                col = literal_column(col.name)
+                engine = self.database._get_sqla_engine()
+                quote = engine.dialect.identifier_preparer.quote
+                col = literal_column(quote(col.name))
             direction = sa.asc if ascending else sa.desc
             qry = qry.order_by(direction(col))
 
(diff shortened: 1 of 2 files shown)


superset/models/helpers.py
```diff
@@ -1974,7 +1974,9 @@ def get_sqla_query(  # pylint: disable=too-many-arguments,too-many-locals,too-ma
                 and db_engine_spec.allows_hidden_cc_in_orderby
                 and col.name in [select_col.name for select_col in select_exprs]
             ):
-                col = literal_column(col.name)
+                engine = self.database._get_sqla_engine()
+                quote = engine.dialect.identifier_preparer.quote
+                col = literal_column(quote(col.name))
```
john-bodley, line 1979: Interesting. Typically the SQLAlchemy dialect should handle the quoting. Additionally it's unclear whether forceably quoting the column would be problematic when combined with an alias.

@villebro do you have any insights on this? I see previously you registered https://github.com/sqlalchemy/sqlalchemy/issues/4730 related to SQLAlchemy not quoting.

betodealmeida (approved):
I was also confused by the behavior, but I don't think this does any harm and fixes the problem.

PR #26469, opened 2024-01-12
fix(database): allow filtering by UUID

betodealmeida:
### SUMMARY

I found a regression where we can longer filter databases by the UUID. While this functionality is not used by Superset, the [Preset CLI](https://github.com/preset-io/backend-sdk) relies on it (and is used not just with Preset workspaces, but with standalone Superset deployments for multiple reasons).

This PR brings back the functionality, and adds a unit test to prevent future breakage.

N/A

### TESTING INSTRUCTIONS

Run something like:

http://localhost:8089/api/v1/database/?q=(filters:!((col:uuid,opr:eq,value:%276ae566ae36f64a19a210621253414d03%27),(col:changed_by,opr:rel_o_m,value:1)),order_column:changed_on_delta_humanized,order_direction:desc,page:0,page_size:25)

And verify that it works.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -205,6 +205,7 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
         "changed_by",
         "database_name",
         "expose_in_sqllab",
+        "uuid",
     ]
     search_filters = {"allow_file_upload": [DatabaseUploadEnabledFilter]}
     allowed_rel_fields = {"changed_by", "created_by"}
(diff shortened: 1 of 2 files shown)


craig-rueda (approved):

villebro:
Tagging the PR that caused the regression: #25961 For future reference, it's probably a good idea to always explicitly specify all searchable columns when creating new API classes to avoid regressions like this.

PR #26473, opened 2024-01-12
fix: docker ephemeral environment, push only on testenv comment

dpgaspar:
### SUMMARY

Currently docker ephemeral env (https://github.com/apache/superset/blob/master/.github/workflows/docker-ephemeral-env.yml) triggers when docker workflow finishes, this changes moves this step to when an actual committer requests for a docker env to be created. Both workflows triggers have `secrets` access but `issue_comment` uses current default branch commit SHA so it's no subject to PR changes and validates that only privilege users can execute this step.

I was not able to properly test this change since it run with current master git SHA, so I expect that further changes may be needed.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


betodealmeida (approved):
Looks good!

geido (approved):

.github/workflows/ephemeral-env.yml
```diff
@@ -79,6 +79,80 @@ jobs:
...
+
+    steps:
+      - name: "Download artifact"
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            const artifacts = await github.actions.listWorkflowRunArtifacts({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                run_id: ${{ github.event.workflow_run.id }},
```
mistercrunch, line 96: hey @dpgaspar this change broke the ephemeral envs because the context is different. Before the trigger was a workflow_run, and now being comment created this property is not available in this context.

PR #21535, opened 2022-09-21
fix(parsedquery): resolve conflict when aliases and tables have the same name

victorarbuesmallada:
### SUMMARY
We discovered that in queries that have aliases with the same name as tables, that removes said tables from the list returned in the `ParsedQuery` `tables` method. This affects the way we prevent users to access to certain tables, since they can circumvent the restricted access by retrieving data through subqueries that have the same alias as the name of the involved table.
In essence, this won't work:
`select * from forbidden_table`
This will work:
`select * from (select * from forbidden_table)  forbidden_table`

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
A normal query shows that the table can't be accessed:

blocked

A query with a forbidden table on a subquery:

allowed

### TESTING INSTRUCTIONS
Create a user that has only certain tables permission and try to perform a query with a forbidden table inside a subquery with the alias the same as the table name.

### ADDITIONAL INFORMATION
- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #24032, opened 2023-05-12
feat: robust(er) adhoc query validation

betodealmeida:
### SUMMARY

The function `has_table_query` is used to detect if a given query is selecting from one or more tables, in order to prevent malicious ad-hoc expressions that bypass RLS. Because `sqlparse` is non-validating the function needs to keep track of a lot of state, and often fails on edge cases.

This PR rewrites the function to use `sqloxide`. The advantage is that the function is now more robust, supporting custom dialects and handling more edge cases. There are a couple disadvantages, though:

- `sqloxide` was an optional dependency, with this PR it becomes mandatory.
- The function `has_table_query` now requires a full SQL query, so for validating expressions they need to be wrapped in `f"SELECT {expression}"`.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N/A

### TESTING INSTRUCTIONS

The unit tests pass mostly unmodified, the only exception is that `table` is usually a reserved keyword and had to be quoted in order to `sqloxide` to parse it (`sqlparse` is much more lenient). I added a new test to cover another edge case that was found could bypass RLS.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26476, opened 2024-01-12
fix(sqlparse): improve table parsing

betodealmeida:
### SUMMARY

This PR improves the `ParsedQuery.tables` property, so it works with more edge cases. It fixes https://github.com/apache/superset/pull/21535, covering some security issues. It makes use of `sqlglot`. I've tried a [different approach](https://github.com/apache/superset/pull/24032) in the past using `sqloxide` instead (since we already had it as an optional dependency), but `sqlglot` is a better dependency since it's pure Python, and people have had problems in the past to install `sqloxide` on arm systems.

In the future I think it would be nice to consider replacing our usage of `sqlparse` with `sqlglot`. (We might want to keep `sqloxide` as optional, like it currently is, for places where performance is crucial.) In this PR modified `ParsedQuery` to receive an optional `engine` parameter, so that it can leverage dialect-specific parsers. I think this offers a nice abstraction, since all the parser-dependent features live inside `ParsedQuery` now.

N/A

### TESTING INSTRUCTIONS

Unit tests pass. I added a couple more cases that were security issues not working correctly.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -16,15 +16,15 @@
 # under the License.
 import logging
 import re
-from collections.abc import Iterator
+import urllib.parse
+from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from typing import Any, cast, Optional
-from urllib import parse
 
 import sqlparse
 from sqlalchemy import and_
-from sqlglot import exp, parse_one
-from sqlglot.optimizer.scope import traverse_scope
+from sqlglot import exp, parse, parse_one
+from sqlglot.optimizer.scope import Scope, ScopeType, traverse_scope
 from sqlparse import keywords
 from sqlparse.lexer import Lexer
 from sqlparse.sql import (
@@ -181,7 +181,7 @@ class Table:
         """
 
         return ".".join(
-            parse.quote(part, safe="").replace(".", "%2E")
+            urllib.parse.quote(part, safe="").replace(".", "%2E")
             for part in [self.catalog, self.schema, self.table]
             if part
         )
@@ -191,11 +191,17 @@ class Table:
 
 
 class ParsedQuery:
-    def __init__(self, sql_statement: str, strip_comments: bool = False):
+    def __init__(
+        self,
+        sql_statement: str,
+        strip_comments: bool = False,
+        dialect: Optional[str] = None,
+    ):
         if strip_comments:
             sql_statement = sqlparse.format(sql_statement, strip_comments=True)
 
         self.sql: str = sql_statement
+        self.dialect = dialect
         self._tables: set[Table] = set()
         self._alias_names: set[str] = set()
         self._limit: Optional[int] = None
@@ -208,14 +214,90 @@ class ParsedQuery:
     @property
     def tables(self) -> set[Table]:
         if not self._tables:
-            self._tables = {
-                Table(source.name, source.db if source.db != "" else None)
-                for scope in traverse_scope(parse_one(self.sql))
-                for source in scope.sources.values()
-                if isinstance(source, exp.Table)
-            }
-
+            self._tables = self._extract_tables_from_sql()
         return self._tables
+
+    def _extract_tables_from_sql(self) -> set[Table]:
+        """
+        Extract all table references in a query.
+
+        Note: this uses sqlglot, since it's better at catching more edge cases.
+        """
+        try:
+            statements = parse(self.sql, dialect=self.dialect)
+        except Exception:  # pylint: disable=broad-exception-caught
+            logger.warning("Unable to parse SQL (%s): %s", self.dialect, self.sql)
+            return set()
+
+        return {
+            table
+            for statement in statements
+            for table in self._extract_tables_from_statement(statement)
+            if statement
+        }
+
+    def _extract_tables_from_statement(self, statement: exp.Expression) -> set[Table]:
+        """
+        Extract all table references in a single statement.
+
+        Please not that this is not trivial; consider the following queries:
+
+            DESCRIBE some_table;
+            SHOW PARTITIONS FROM some_table;
+            WITH masked_name AS (SELECT * FROM some_table) SELECT * FROM masked_name;
+
+        See the unit tests for other tricky cases.
+        """
+        sources: Iterable[exp.Table]
+
+        if isinstance(statement, exp.Describe):
+            # A `DESCRIBE` query has no sources in sqlglot, so we need to explicitly
+            # query for all tables.
+            sources = statement.find_all(exp.Table)
+        elif isinstance(statement, exp.Command):
+            # Commands, like `SHOW COLUMNS FROM foo`, have to be converted into a
+            # `SELECT` statetement in order to extract tables.
+            literal = statement.find(exp.Literal)
+            if not literal:
+                return set()
+
+            pseudo_query = parse_one(f"SELECT {literal.this}", dialect=self.dialect)
+            sources = pseudo_query.find_all(exp.Table)
+        elif statement:
+            sources = []
+            for scope in traverse_scope(statement):
+                for source in scope.sources.values():
+                    if not isinstance(source, exp.Table):
+                        continue
+
+                    # CTEs in the parent scope look like tables (and are represented by
+                    # exp.Table objects), but should not be considered as such;
+                    # otherwise a user with access to table `foo` could access any table
+                    # with a query like this:
+                    #
+                    #   WITH foo AS (SELECT * FROM bar) SELECT * FROM foo
+                    #
+                    parent_sources = scope.parent.sources if scope.parent else {}
+                    ctes_in_scope = {
+                        name
+                        for name, parent_scope in parent_sources.items()
+                        if isinstance(parent_scope, Scope)
+                        and parent_scope.scope_type == ScopeType.CTE
+                    }
+                    if source.name not in ctes_in_scope:
+                        sources.append(source)
+
+        else:
+            return set()
+
+        return {
+            Table(
+                source.name,
+                source.db if source.db != "" else None,
+                source.catalog if source.catalog != "" else None,
+            )
+            for source in sources
+        }
 
     @property
     def limit(self) -> Optional[int]:
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -14,6 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+# pylint: disable=too-many-lines
+
 import logging
 import re
 import urllib.parse
@@ -24,6 +27,8 @@ from typing import Any, cast, Optional
 import sqlparse
 from sqlalchemy import and_
 from sqlglot import exp, parse, parse_one
+from sqlglot.dialects import Dialects
+from sqlglot.errors import ParseError
 from sqlglot.optimizer.scope import Scope, ScopeType, traverse_scope
 from sqlparse import keywords
 from sqlparse.lexer import Lexer
@@ -55,7 +60,7 @@ from superset.utils.backports import StrEnum
 
 try:
     from sqloxide import parse_sql as sqloxide_parse
-except:  # pylint: disable=bare-except
+except (ImportError, ModuleNotFoundError):
     sqloxide_parse = None
 
 RESULT_OPERATIONS = {"UNION", "INTERSECT", "EXCEPT", "SELECT"}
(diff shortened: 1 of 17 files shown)


tests/unit_tests/sql_parse_tests.py
```diff
@@ -311,7 +318,7 @@ def test_extract_tables_where_subquery() -> None:
             """
 SELECT name
 FROM t1
-WHERE regionkey EXISTS (SELECT regionkey FROM t2)
```
betodealmeida, line 314: This SQL is invalid.

superset/sql_parse.py
```diff
@@ -206,14 +275,91 @@ def __init__(self, sql_statement: str, strip_comments: bool = False):
...
 
+    def _extract_tables_from_sql(self) -> set[Table]:
+        """
+        Extract all table references in a query.
+
+        Note: this uses sqlglot, since it's better at catching more edge cases.
+        """
+        try:
+            statements = parse(self.sql, dialect=self._dialect)
+        except Exception:  # pylint: disable=broad-exception-caught
```
geido, line 289: We can catch `ParseError` from `sqlglot.errors`. Also, I am thinking we should capture an exception here rather than a warning?

betodealmeida: Good point, a more specific exception is better.

I'm also not a fan of the warning, but the current behavior today is to return an empty set if the parsing fails. I wanted to keep the existing behavior because changing the parsing is already a big change.

superset/sql_parse.py
```diff
@@ -206,14 +276,91 @@ def __init__(self, sql_statement: str, strip_comments: bool = False):
...
+    def _extract_tables_from_sql(self) -> set[Table]:
+        """
+        Extract all table references in a query.
+
+        Note: this uses sqlglot, since it's better at catching more edge cases.
+        """
+        try:
+            statements = parse(self.sql, dialect=self._dialect)
+        except ParseError:
+            logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
```
giftig, line 291: I saw you mentioned this was the existing behaviour in cases of failure; this implies it's going to find that the SQL doesn't access any tables and therefore allow the access, if it couldn't figure out which tables were queried (unless I'm missing some surrounding context as I haven't scanned through all this logic again recently).

Superset errors can be very difficult for users to understand sometimes, so I'd avoid squashing the exception here and provide as much context as possible in the log, and at least the original exception text + a warning that the likely result of this is the assumption that the user should have access. If you use `logger.exception` instead, you'll be providing the full stack trace of the original exception in the log. It'll be error level instead of warning, but I'd argue this is an error we're having to recover from imperfectly rather than a warning which users might be able to ignore, so I think that's an improvement too.

superset/sql_parse.py
```diff
@@ -206,14 +276,91 @@ def __init__(self, sql_statement: str, strip_comments: bool = False):
...
+                    if not isinstance(source, exp.Table):
+                        continue
+
+                    # CTEs in the parent scope look like tables (and are represented by
+                    # exp.Table objects), but should not be considered as such;
+                    # otherwise a user with access to table `foo` could access any table
+                    # with a query like this:
+                    #
+                    #   WITH foo AS (SELECT * FROM bar) SELECT * FROM foo
+                    #
```
giftig, line 341: Maybe a good candidate for an accessory method to make this a little more readable and a little less nested, especially since you basically have a nice docstring here.

superset/sql_parse.py
```diff
@@ -206,14 +276,91 @@ def __init__(self, sql_statement: str, strip_comments: bool = False):
...
+                    ctes_in_scope = {
+                        name
+                        for name, parent_scope in parent_sources.items()
+                        if isinstance(parent_scope, Scope)
+                        and parent_scope.scope_type == ScopeType.CTE
+                    }
+                    if source.name not in ctes_in_scope:
+                        sources.append(source)
+
+        else:
```
giftig, line 352: Does this scenario just represent incorrect use of the method? Or is there some scenario in which an `exp.Expression` could be falsy, like an empty query?

I see the codecov plugin has complained it's not covered by your tests, so I guess trying to write a test for this scenario may answer whether it needs special handling.

tests/unit_tests/sql_parse_tests.py
```diff
@@ -1306,6 +1327,14 @@ def test_sqlparse_issue_652():
             "(SELECT table_name FROM /**/ information_schema.tables WHERE table_name LIKE '%user%' LIMIT 1)",
             True,
         ),
+        (
+            "SELECT FROM (SELECT FROM forbidden_table) AS forbidden_table;",
+            True,
+        ),
+        (
+            "SELECT * FROM (SELECT * FROM forbidden_table) forbidden_table",
```
giftig, line 1335: Nice :+1:

giftig:
Thanks @betodealmeida, this looks like a big security improvement if the number of dialects in that mapping for sqlglot are to be believed. A couple of minor comments but LGTM otherwise.

superset/sql_parse.py
betodealmeida: Ah, good point!

superset/sql_parse.py
betodealmeida: Ah, this is leftover code. In sqlglot the result of parsing SQL is `Optional[Expression]`, so I added this logic. But we now take care of false value sin the caller.

superset/sql_parse.py
betodealmeida: > this implies it's going to find that the SQL doesn't access any tables and therefore allow the access, if it couldn't figure out which tables were queried

Right, but it works under the assumption that if it can't parse the query then the query is invalid, so it will never run.

I agree that this is a very risky assumption that delegates security responsibilities to a parser, and we should fix it. I will do it in a next PR, though.

giftig: Ah ok, that makes sense. I do recall it seemed like superset tends to play it safe if it couldn't parse the query rather than allowing it, so I must have been missing some surrounding context here.

giftig (approved):

mistercrunch:
Can we add pytests around the cases specified in https://github.com/apache/superset/pull/21535 ? I thought we had some, but it'd be good to prove previous issues that are now addressed, and prevent regressions on those.

mistercrunch:
Also wondering whether hoping of deprecating `sqlparse` dep is reasonable. Are there things that sqlparse can do that sqlglot cannot?

john-bodley:
@betodealmeida there seems to be a regression with the introduction of SQLGlot in this PR, i.e., it cannot correctly parse statements of the form

```sql
SELECT * FROM foo WHERE ds = '{{ presto.latest_partition('foo') }}'
```

whereas `sqlparse` could.

tobymao:
this is not valid sql syntax. ansi sql needs to escape single quotes with another single quote.

```sql 
select '{{ presto(''foo'') }}'
```

spark uses backticks, sqlglot will handle the appropriate escape character given the dialect.

betodealmeida:
@john-bodley is that supposed to be valid SQL? Or are we using a SQL parser to try to parse a Jinja template?

john-bodley:
@betodealmeida see the example I linked previously from the official documentation

Users have likely been using this pattern for years—irrespective of whether it’s valid SQL or not—thus we should preserve said functionality. I’m really not sure how feasible it would be to run a migration to replace the violating instances of `’`with either `’’` or `”`.

For context this pattern is heavily used at Airbnb and this regression will likely result in a SEV given that Superset is likely providing access to all cached charts utilizing this macro—irrespective of whether the user has access to the underlying dataset or not.

Personally, given the severity of the regression (from a security perspective), I think we need to either revert this PR or find a feasible fix/workaround ASAP.

betodealmeida:
This PR fixes a public CVE. Please don't revert it. I'm sure we can find a solution.

betodealmeida:
It's trivial to convert the Jinja template to valid SQL before parsing, it shouldn't be hard to solve this.

tobymao:
I see what's going on.

@john-bodley 

```sql
SELECT * FROM foo WHERE ds = '{{ presto.latest_partition('foo') }}'
```

Is not SQL. 

This is a jinja template and needs to be rendered as such before sql parsing. You're expecting {{presto.latest_partition('foo') }} to evaluate first in a jinja preprocessor, the output of which will not contain quotes.

a workaround for this could be for superset to evaluate jinja first with a place holder like empty string.

betodealmeida:
Exactly this. ^

betodealmeida:
@john-bodley can you share where the problem is happening? If you create a ticket with a repro I can fix it.

john-bodley:
@betodealmeida the problem occurs [here](https://github.com/apache/superset/blob/fde93dcf08122d8b41ca296213e09616d1b71782/superset/security/manager.py#L1910-L1916), i.e., where the security manager tries to determine the tables associated with a query—be that from SQL Lab or a virtual dataset. 

Given that the SQLGlot parsing fails, the [`_extract_tables_from_sql`](https://github.com/apache/superset/blob/fde93dcf08122d8b41ca296213e09616d1b71782/superset/sql_parse.py#L282-L299) method returns an empty set meaning that—from the security manager's standpoint—there are no tables which need to be checked from a permissions standpoint.

betodealmeida:
> Given that the SQLGlot parsing fails, the [`_extract_tables_from_sql`](https://github.com/apache/superset/blob/fde93dcf08122d8b41ca296213e09616d1b71782/superset/sql_parse.py#L282-L299) method returns an empty set meaning that—from the security manager's standpoint—there are no tables which need to be checked, implying all users have access to the query.

Yeah, this seems to be the biggest culprit. The original behavior was to fail silently, and I kept that behavior initially. I have a PR open https://github.com/apache/superset/pull/26748 where I change that behavior, but I'll also update the security manager so that we apply the Jinja template before trying to parse it with a SQL parser.

betodealmeida:
For context, this seems to fix it:

```diff
diff --git a/superset/commands/dataset/create.py b/superset/commands/dataset/create.py
index 16b87a567a..0c4a61487f 100644
--- a/superset/commands/dataset/create.py
+++ b/superset/commands/dataset/create.py
@@ -21,6 +21,7 @@ from flask_appbuilder.models.sqla import Model
 from marshmallow import ValidationError
 from sqlalchemy.exc import SQLAlchemyError
 
+from superset import jinja_context
 from superset.commands.base import BaseCommand, CreateMixin
 from superset.commands.dataset.exceptions import (
     DatabaseNotFoundValidationError,
@@ -85,10 +86,12 @@ class CreateDatasetCommand(CreateMixin, BaseCommand):
             exceptions.append(TableNotFoundValidationError(table_name))
 
         if sql:
+            processor = jinja_context.get_template_processor(database=database)
+            rendered_sql = processor.process_template(sql)
             try:
                 security_manager.raise_for_access(
                     database=database,
-                    sql=sql,
+                    sql=rendered_sql,
                     schema=schema,
                 )
             except SupersetSecurityException as ex:
```

I'll add some tests and push the fix for review.

betodealmeida:
The two main problems seem to be:

1. We are raising no exceptions when we try to extract tables from SQL (https://github.com/apache/superset/pull/26748)
2. We are passing something that's no SQL to the security manager, which expects SQL

john-bodley:
@betodealmeida as far as repro cases go here's the most simplest (unit test worthy) example:

```python
>>> from superset.sql_parse import ParsedQuery
>>>
>>> ParsedQuery("SELECT '{{ presto.latest_partition('foo') }}'").tables
set()
```

This functionally is used in a number of places, though the most concerning—from a security perspective—is likely [here](https://github.com/apache/superset/blob/b1adede1ee670ad800c07aeb76f2f73799228960/superset/security/manager.py#L1910-L1915). If one doesn't detect any tables associated with the query then no access checks are performed.

Note from a testing perspective a query of the form (which _weirdly_ references both `foo` and `bar` tables), 

```sql
SELECT * FROM foo WHERE ds =  '{{ presto.latest_partition('bar') }}'
```

as opposed to a query of the form (which references only the `foo` table):

```sql
SELECT * FROM foo WHERE ds =  '{{ presto.latest_partition('foo') }}'
```

would be super useful, per https://github.com/apache/superset/pull/27464#issuecomment-1989433700, to test the multi phase approach, especially given that in 3.0.3 `sqlparse` would fail to identify the `bar` table.

betodealmeida:
Thanks, @john-bodley, but that repro is not very useful because we shouldn't be passing non-valid SQL to a class that represents a SQL query, to start. 😊 The proper fix, IMHO, is to find all the places where we're passing a Jinja template where SQL is expected and process the template to generate SQL, which is why I asked where this is happening for you.

And I understand that in the past `ParsedQuery` accepted strings that were not valid SQL, but with [SIP-117](https://github.com/apache/superset/issues/26786) `ParsedQuery` is going away, and the new class `SQLStatement` should represent an actual SQL statement, with an associated dialect. I'd rather keep the class responsible only for parsing and manipulating SQL, which is why I think the template processing should happen elsewhere.

For database-dependent macros like `presto.latest_partition` we could add an option to the template processor that would be used by the macros to return immediately. For example (pseudocode):

```python
class BaseTemplateProcessor:
    def __init__(self, mode: TemplateProcessorMode, ...):
        self.mode = mode
        ....


class PrestoTemplateProcessor(JinjaTemplateProcessor):
    def latest_partitions(self, table_name: str) -> Optional[list[str]]:
        if self.mode == TemplateProcessorMode.parsing:
            return table_name

        ...
```

tobymao:
Just another note, is that jinja can expand to anything. 

```select * from foo where ds = '{{ partition() }}'```

can expand into even more tables, since it's string manipulation. it's an even bigger security hazard.

```select * from foo where ds = '2022-01-01' join payments``` 

is certainly a legal jinja transformation. so you can kind of see why it's impossible to parse.

john-bodley:
@betodealmeida whenever either `sql` or `query` is passed to the `raise_for_access` method the `ParsedQuery(...).tables` statement will be invoked. AFAIK this can happen from:

- [SQL Lab](https://github.com/apache/superset/blob/2a5b5fe0498bfa9359b8761a0df13a0ce2589645/superset/models/sql_lab.py#L247)
- Explore when you're referencing a virtual dataset
- Explore when you're using ad-hoc filters (be that via a `WHERE` or `HAVING` clause) using custom SQL

john-bodley:
@betodealmeida and @michael-s-molina I've found another wrinkle with the SQLGlot integration. The `/savedqueryview/list/` which displays a table of your saved queries includes a column named "Tables" which uses the `ParsedQuery` class to extract the tables names from every saved query.

These saved queries may contain Jinja templating—including the `latest[_sub]_partition]` macros resulting in SQL statements which SQLGlot is unable to parse—especially if the is no proper escaping. The resulting UX is users see [this](https://github.com/apache/superset/blob/ad7bd09819d3fd6ac8f669c6f884551cbd66bcd6/superset/sql_parse.py#L479-L485) error and no saved queries are listed.

The long-term fix (as discussed previously) of having a multi-phase approach  to extracting the table names without having to render aspects of the Jinja templates should resolve this, however I'm not entirely sure what the short-term solution should be. 

Some possible solutions could be:

1. Rendering the entire query—irrespective of how inefficient it may be and/or the unnecessary load/burden on the analytical engines. Note this likely may ot be viable from a scalability/performance perspective and depends on the type/quantity of saved query a user has.
2. Simply removing the "Tables" column, though this would be deemed a breaking change? I'm not entirely sure how valuable this is to users.
3. Reverting back to using `sqlparse` and only introduce SQLGlot when i) the various workflows are hardened, and ii) [[SIP-117] Improve SQL parsing](https://github.com/apache/superset/issues/26786) is fully implemented. For context we've had about four reported regressions with the SQLGlot integration and thus we might not quite be ready for its widespread release.

PR #26540, opened 2024-01-12
chore(API): Include changed_by.id in Get Charts and Get Datasets API responses

Vitor-Avila:
### SUMMARY
The API responses for fetching all charts (`/api/v1/chart/`) and all datasets (`/api/v1/dataset/`) don't include the `changed_by.id` attribute. This is useful when implementing automations to maintain content (that involve reaching out to the user that last modified it). 

This PR includes this field in the two API responses. It is already available when fetching all dashboards (`/api/v1/dashboard/`).

### TESTING INSTRUCTIONS
1. Make sure you have at least 1 chart and 1 dashboard created.
2. Overwrite/modify both of them.
3. Send a `GET` request to `/api/v1/chart/` and confirm the `changed_by.id` is visible there. 
4. Send a `GET` request to `/api/v1/dataset/` and confirm the `changed_by.id` is visible there. 

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -159,6 +159,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
         "cache_timeout",
         "changed_by.first_name",
         "changed_by.last_name",
+        "changed_by.id",
         "changed_by_name",
         "changed_on_delta_humanized",
         "changed_on_dttm",
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -108,6 +108,7 @@ class DatasetRestApi(BaseSupersetModelRestApi):
         "changed_by_name",
         "changed_by.first_name",
         "changed_by.last_name",
+        "changed_by.id",
         "changed_on_utc",
         "changed_on_delta_humanized",
         "default_endpoint",


Vitor-Avila:
hey @eschutho @betodealmeida @john-bodley while working on this PR, I noticed that the response for getting all charts (`/api/v1/chart/`) includes more fields than when getting a single chart (`/api/v1/chart/<pk>`). 

Is this by design? I would expect the same amount of fields, or the other way around (fetching a specific chart returns more information about it than fetching all charts). 

What are your thoughts on making sure that:
* All asset endpoints (charts, datasets, dashboards, etc) return the same fields (when available)
* Fetching a single resource returns the same information than fetching all resources

I would imagine this wouldn't require a SIP (given we're not removing fields from the API or adding complexity) but would love to get your opinion.

john-bodley (approved):

john-bodley:
@Vitor-Avila regarding your comment:

> Is this by design? I would expect the same amount of fields, or the other way around (fetching a specific chart returns more information about it than fetching all charts).

I sense this is simply one (of many) inconsistencies which currently exist in our APIs—likely because they were probably written to handle specific use cases (both from a request and response perspective), i.e., historically the APIs (which evolved from Flask views) were built in a vertical rather than horizontal manner. There are pros/cons with both approaches where one could argue that the current approach is optimized for the UX, i.e., payload reduction where possible.

There likely needs to be a concerted effort to ensure that our APIs are consistent.

Vitor-Avila:
Thanks for the additional context, @john-bodley!

dpgaspar (approved):

sadpandajoe:
🏷️ preset:2024.3

Issue #26190, opened 2023-12-05
Accessibility: Elements that are not an "a" or "button" are being used for for clickable UI


yousoph:
#### How to reproduce the bug

1. Create a dashboard 
2. In the Inspect console, look at the different button elements 

### Expected results
Clickable elements should have `<a>` or `<button>` 
https://www.w3.org/WAI/WCAG21/Understanding/keyboard
https://www.w3.org/WAI/WCAG21/Techniques/html/H91 

### Actual results

Elements such as the tab headers in dashboards, the icon to copy a link, etc don't have `<a>` or `<button>`.
If an element is clickable, then it should semantically show as a clickable item by using the proper html semantics. 
If neither an `<a>` or `<button>` can be used, then a `role=button` attribute can be applied to the element

### Environment

(please complete the following information):

- browser type and version:
- superset version: `master`
- python version: `python --version`
- node.js version: `node -v`
- any feature flags active:

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [ ] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [x] I have reproduced the issue with at least the latest released version of superset.
- [x] I have checked the issue tracker for the same issue and I haven't found one similar.

### Additional context

We may also be able to check if adding role=button works

eulloa10:
@yousoph @eschutho I can handle this one if it hasn't been claimed. Thanks for the additional notes in the project-accessibility channel.

eulloa10:
@eschutho I submitted #26602 for this issue. Please let me know if you have any questions, comments, or suggestions.

rusackas:
I think this was supposed to be closed by #26602, so I'll just do that here. Holler if this needs to be reopened.

PR #12920, opened 2021-02-03
refactor: Bootstrap to AntD - Collapse

michael-s-molina:
### SUMMARY

- Migrates `Collapse` component from Bootstrap to AntD.
- Unifies all `Collapse` stories into a single one with controls support.
- Moves `Collapse` component to own folder.
- Fixes `padding` and `border` in storybook gallery examples.
- Increases clickable area in SQL Lab collapsable tables.

See: #10254

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
Screen Shot 2021-02-02 at 4 23 50 PM
Screen Shot 2021-02-02 at 4 21 34 PM
Screen Shot 2021-02-02 at 4 24 15 PM
Screen Shot 2021-02-02 at 4 21 01 PM
Screen Shot 2021-02-02 at 4 24 53 PM
Screen Shot 2021-02-02 at 4 19 04 PM
Screen Shot 2021-02-02 at 4 25 29 PM
Screen Shot 2021-02-02 at 4 20 28 PM
Screen Shot 2021-02-02 at 4 26 02 PM
Screen Shot 2021-02-02 at 4 17 25 PM

@rusackas @junlincc @geido 

### TEST PLAN
1 - Open any screen that contains a Collapse (explore, welcome, dashboards, etc.)
2 - All collapse components should have the same theme and behavior
3 - Additional testing can be done using Storybook

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [x] Changes UI
- [ ] Requires DB Migration.
- [ ] Confirm DB Migration upgrade and downgrade tested.
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26602, opened 2024-01-13
refactor: add "button" role to clickable UI elements for improved accessibility

eulloa10:
### SUMMARY

Clickable UI elements lacking `<a>` or `<button>` tags, or the `role="button"` attribute, weren't semantically clear as clickable elements when using a screen reader. 

To improve accessibility, I reviewed the React codebase for `onClick` handlers. When encountering HTML elements with `onClick` handlers AND without `<a>` or `<button>` tags, I added the `role="button"` attribute to identify them as clickable. 

This approach targets the most immediate and straightforward cases, ensuring that elements triggered by onClick handlers are now clearly identified as clickable for users relying on screen readers. 

### TESTING INSTRUCTIONS

Testing options include:
- Using a screen reader: identify and navigate through the clickable components and ensure they are announced correctly 
- Using browser devtools: inspect clickable elements and verify that elements without `<a>` or `<button>` tags have the `role="button"` attribute.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #26190 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/legacy-preset-chart-deckgl/src/components/Legend.tsx
+++ b/superset-frontend/plugins/legacy-preset-chart-deckgl/src/components/Legend.tsx
@@ -106,6 +106,7 @@ const Legend = ({
       <li key={k}>
         <a
           href="#"
+          role="button"
           onClick={() => toggleCategory(k)}
           onDoubleClick={() => showSingleCategory(k)}
         >
--- a/superset-frontend/src/SqlLab/components/TableElement/index.tsx
+++ b/superset-frontend/src/SqlLab/components/TableElement/index.tsx
@@ -345,6 +345,7 @@ const TableElement = ({ table, ...props }: TableElementProps) => {
             <Fade
               data-test="fade"
               hovered={hovered}
+              role="button"
               onClick={e => e.stopPropagation()}
             >
               {renderControls()}
--- a/superset-frontend/src/SqlLab/components/TableElement/index.tsx
+++ b/superset-frontend/src/SqlLab/components/TableElement/index.tsx
@@ -345,7 +345,6 @@ const TableElement = ({ table, ...props }: TableElementProps) => {
             <Fade
               data-test="fade"
               hovered={hovered}
-              role="button"
               onClick={e => e.stopPropagation()}
             >
               {renderControls()}
--- a/superset-frontend/src/components/CachedLabel/index.tsx
+++ b/superset-frontend/src/components/CachedLabel/index.tsx
@@ -44,6 +44,7 @@ const CacheLabel: React.FC<CacheLabelProps> = ({
       <Label
         className={`${className}`}
         type={labelType}
+        role="button"
         onClick={onClick}
         onMouseOver={() => setHovered(true)}
         onMouseOut={() => setHovered(false)}
--- a/superset-frontend/src/components/CachedLabel/index.tsx
+++ b/superset-frontend/src/components/CachedLabel/index.tsx
@@ -44,7 +44,6 @@ const CacheLabel: React.FC<CacheLabelProps> = ({
       <Label
         className={`${className}`}
         type={labelType}
-        role="button"
         onClick={onClick}
         onMouseOver={() => setHovered(true)}
         onMouseOut={() => setHovered(false)}
--- a/superset-frontend/src/components/Datasource/CollectionTable.tsx
+++ b/superset-frontend/src/components/Datasource/CollectionTable.tsx
@@ -311,13 +311,13 @@ export default class CRUDCollection extends React.PureComponent<
   }
 
   renderSortIcon(col: string) {
-    if (this.state.sortColumn === col && this.state.sort === SortOrder.Asc) {
-      return <Icons.SortAsc onClick={this.sortColumn(col, 2)} />;
+    if (this.state.sortColumn === col && this.state.sort === SortOrder.asc) {
+      return <Icons.SortAsc role="button" onClick={this.sortColumn(col, 2)} />;
     }
-    if (this.state.sortColumn === col && this.state.sort === SortOrder.Desc) {
-      return <Icons.SortDesc onClick={this.sortColumn(col, 0)} />;
+    if (this.state.sortColumn === col && this.state.sort === SortOrder.desc) {
+      return <Icons.SortDesc role="button" onClick={this.sortColumn(col, 0)} />;
     }
-    return <Icons.Sort onClick={this.sortColumn(col, 1)} />;
+    return <Icons.Sort role="button" onClick={this.sortColumn(col, 1)} />;
   }
 
   renderTH(col: string, sortColumns: Array<string>) {
--- a/superset-frontend/src/components/Datasource/CollectionTable.tsx
+++ b/superset-frontend/src/components/Datasource/CollectionTable.tsx
@@ -311,13 +311,13 @@ export default class CRUDCollection extends React.PureComponent<
   }
 
   renderSortIcon(col: string) {
-    if (this.state.sortColumn === col && this.state.sort === SortOrder.asc) {
-      return <Icons.SortAsc role="button" onClick={this.sortColumn(col, 2)} />;
+    if (this.state.sortColumn === col && this.state.sort === SortOrder.Asc) {
+      return <Icons.SortAsc onClick={this.sortColumn(col, 2)} />;
     }
-    if (this.state.sortColumn === col && this.state.sort === SortOrder.desc) {
-      return <Icons.SortDesc role="button" onClick={this.sortColumn(col, 0)} />;
+    if (this.state.sortColumn === col && this.state.sort === SortOrder.Desc) {
+      return <Icons.SortDesc onClick={this.sortColumn(col, 0)} />;
     }
-    return <Icons.Sort role="button" onClick={this.sortColumn(col, 1)} />;
+    return <Icons.Sort onClick={this.sortColumn(col, 1)} />;
   }
 
   renderTH(col: string, sortColumns: Array<string>) {
--- a/superset-frontend/src/components/Icons/AntdEnhanced.tsx
+++ b/superset-frontend/src/components/Icons/AntdEnhanced.tsx
@@ -26,7 +26,11 @@ const AntdEnhancedIcons = Object.keys(AntdIcons)
   .filter(k => !k.includes('TwoTone'))
   .map(k => ({
     [k]: (props: IconType) => (
-      <StyledIcon component={AntdIcons[k]} {...props} />
+      <StyledIcon
+        component={AntdIcons[k]}
+        role={props.onClick ? 'button' : undefined}
+        {...props}
+      />
     ),
   }))
   .reduce((l, r) => ({ ...l, ...r }));
--- a/superset-frontend/src/components/Icons/AntdEnhanced.tsx
+++ b/superset-frontend/src/components/Icons/AntdEnhanced.tsx
@@ -26,11 +26,7 @@ const AntdEnhancedIcons = Object.keys(AntdIcons)
   .filter(k => !k.includes('TwoTone'))
   .map(k => ({
     [k]: (props: IconType) => (
-      <StyledIcon
-        component={AntdIcons[k]}
-        role={props.onClick ? 'button' : undefined}
-        {...props}
-      />
+      <StyledIcon component={AntdIcons[k]} {...props} />
     ),
   }))
   .reduce((l, r) => ({ ...l, ...r }));
--- a/superset-frontend/src/components/Icons/AntdEnhanced.tsx
+++ b/superset-frontend/src/components/Icons/AntdEnhanced.tsx
@@ -26,7 +26,7 @@ const AntdEnhancedIcons = Object.keys(AntdIcons)
   .filter(k => !k.includes('TwoTone'))
   .map(k => ({
     [k]: (props: IconType) => {
-      const whatRole = props.onClick ? 'button' : props?.role || 'img';
+      const whatRole = props?.onClick ? 'button' : 'img';
       return <StyledIcon component={AntdIcons[k]} role={whatRole} {...props} />;
     },
   }))
--- a/superset-frontend/src/components/Icons/AntdEnhanced.tsx
+++ b/superset-frontend/src/components/Icons/AntdEnhanced.tsx
@@ -25,9 +25,10 @@ import IconType from './IconType';
 const AntdEnhancedIcons = Object.keys(AntdIcons)
   .filter(k => !k.includes('TwoTone'))
   .map(k => ({
-    [k]: (props: IconType) => (
-      <StyledIcon component={AntdIcons[k]} {...props} />
-    ),
+    [k]: (props: IconType) => {
+      const whatRole = props.onClick ? 'button' : props?.role || 'img';
+      return <StyledIcon component={AntdIcons[k]} role={whatRole} {...props} />;
+    },
   }))
   .reduce((l, r) => ({ ...l, ...r }));
 
--- a/superset-frontend/src/components/Icons/Icon.tsx
+++ b/superset-frontend/src/components/Icons/Icon.tsx
@@ -47,7 +47,7 @@ export interface IconProps extends IconType {
 }
 
 export const Icon = (props: IconProps) => {
-  const { fileName, onClick, ...iconProps } = props;
+  const { fileName, ...iconProps } = props;
   const [, setLoaded] = useState(false);
   const ImportedSVG = useRef<React.FC<React.SVGProps<SVGSVGElement>>>();
   const name = fileName.replace('_', '-');
@@ -70,7 +70,6 @@ export const Icon = (props: IconProps) => {
 
   return (
     <StyledIcon
-      role={onClick ? 'button' : undefined}
       component={ImportedSVG.current || TransparentIcon}
       aria-label={name}
       {...iconProps}
--- a/superset-frontend/src/components/Icons/Icon.tsx
+++ b/superset-frontend/src/components/Icons/Icon.tsx
@@ -47,7 +47,7 @@ export interface IconProps extends IconType {
 }
 
 export const Icon = (props: IconProps) => {
-  const { fileName, ...iconProps } = props;
+  const { fileName, onClick, ...iconProps } = props;
   const [, setLoaded] = useState(false);
   const ImportedSVG = useRef<React.FC<React.SVGProps<SVGSVGElement>>>();
   const name = fileName.replace('_', '-');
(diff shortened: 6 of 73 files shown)


john-bodley:
Thank @eulloa10 for the change. I've kicked off CI, though it seems you have a couple of conflicts you need to resolve.

superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -563,6 +563,7 @@ class Header extends React.PureComponent {
                               onClick={undoLength && onUndo}
                               data-test="undo-action"
                               iconSize="xl"
+                              role="button"
```
justinpark, line 566: Thanks @eulloa10 for all your works.

I propose moving the `onClick` event handler to the `StyledUndoRedoButton` component itself, rather than applying a button role to the icon within it.

superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -583,6 +584,7 @@ class Header extends React.PureComponent {
                               onClick={redoLength && onRedo}
                               data-test="redo-action"
                               iconSize="xl"
+                              role="button"
```
justinpark, line 587: same [issue](https://github.com/apache/superset/pull/26602/files#r1456511910) here

superset-frontend/src/dashboard/components/filterscope/FilterScopeSelector.jsx
```diff
@@ -687,6 +687,7 @@ export default class FilterScopeSelector extends React.PureComponent {
         onClick={this.onChangeFilterField}
         onCheck={this.onCheckFilterField}
         onExpand={this.onExpandFilterField}
+        role="button"
```
justinpark, line 690: This prop is redundant since `FilterFieldTree` doesn't handle it.

superset-frontend/src/explore/components/controls/CheckboxControl.jsx
```diff
@@ -65,6 +65,7 @@ export default class CheckboxControl extends React.Component {
           <ControlHeader
             {...this.props}
             leftNode={this.renderCheckbox()}
+            role="button"
```
justinpark, line 68: This prop is redundant and https://github.com/apache/superset/pull/26602/files#diff-1786e2046531b3067d8ed91311079f60a31e6503a457a043c92c17f29ec31adaR114 already covered

superset-frontend/src/explore/components/controls/FixedOrMetricControl/index.jsx
```diff
@@ -127,7 +127,7 @@ export default class FixedOrMetricControl extends React.Component {
           <Collapse.Panel
             showArrow={false}
             header={
-              <Label onClick={() => undefined}>
+              <Label role="button" onClick={() => undefined}>
```
justinpark, line 130: Since onClick handler does nothing, both role and onClick are redundant.

```suggestion
              <Label>
```

superset-frontend/src/explore/components/controls/VizTypeControl/VizTypeGallery.tsx
```diff
@@ -478,6 +478,7 @@ const Selector: React.FC<{
       key={selector}
       name={selector}
       className={cx(className, isSelected && 'selected')}
+      role="button"
```
justinpark, line 481: Since SelectorLabel is a button element, role is not necessary (implicitly defined) but type is necessary for a11y POV. 

```suggestion
      type="button"
```

eulloa10:
Thank you for reviewing @justinpark. I made the requested changes and they are ready to be reviewed.

eulloa10:
I am looking into the failing test cases.

geido:
Thanks for this @eulloa10. Let me know if I can help with the failing test cases

eulloa10:
Thanks @geido. I looked through the failing unit tests (17 tests) to start, but I am confused on how the changes I made are contributing to the failing tests. What is the best way to approach debugging these?

geido:
@eulloa10 both Cypress tests and unit tests are failing as you changed some of the roles associated to elements. Please check the errors individually and update the tests according to the new roles. If you need more help you can find me in [Apache Superset Slack](https://join.slack.com/t/apache-superset/shared_invite/zt-2be0drwz8-bxPfkdz28ozzk1Iox29ufg)

mistercrunch:
Is there an opportunity as part of this PR to improve accessibility by reusing (or introducing if they don't exist already) low-level reusable components in `src/components` and make those accessible?

The original idea with `src/components` was [amongst other things] to offer a subset of `antd` (and native React like `<a>` or `button`) components that are more opinionated on properties (offer only a subset of properties) and used/reused consistently across Superset. 

Say if we always use `src.components.Button` and that one always has a `role`, then no need to sprinkle `role` in a hundred places (?) Similar with `antd.MenuItem`, if we have a nice little abstraction where you can't get it wrong, then no need to go low level across the codebase, simply in `src/components/`

superset-frontend/src/components/CachedLabel/index.tsx
```diff
@@ -44,6 +44,7 @@ const CacheLabel: React.FC<CacheLabelProps> = ({
       <Label
         className={`${className}`}
         type={labelType}
+        role="button"
```
mistercrunch, line 47: Following up on my higher level comment about fixing in low-level components, it looks like in this particular context, `Label` is pointing to `superset-frontend/src/components/Label/index.tsx`.

Could we simply for `role="button"` when `onClick` is defined? This would tackle all labels at once and prevent having to sprinkle the role prop across the code base, and forgetting to specify it in the future.

superset-frontend/src/dashboard/components/PublishedStatus/index.jsx
```diff
@@ -65,6 +65,7 @@ export default class PublishedStatus extends React.Component {
             title={draftButtonTooltip}
           >
             <Label
+              role="button"
```
mistercrunch, line 68: the fix I mentioned above should tackle this too

mistercrunch:
Similarly, looking at all the many `<Menu />` and `<MenuItem />` entries here, you could really just force role="button" if/when `onClick` is defined, and not have to sprinkle `role="button"` everywhere. We should have a hard rule to ALWAYS use antd component through the wrappers in `src/components/` meaning we always do this lower level stuff from that central point.

eulloa10:
@mistercrunch thank you, appreciate the additional context here and feedback. I am going to use your approach and also split the changes into multiple PRs.

mistercrunch:
The beauty is that a single PR handling both `Menu` and `MenuItem` might be fairly simple and tackle a lot of the changes here in a very simple way.

superset-frontend/src/dashboard/components/menu/DownloadMenuItems/DownloadAsPdf.tsx
```diff
@@ -29,7 +29,7 @@ export default function DownloadAsPdf({
 
   return (
     <Menu.Item key="download-pdf" {...rest}>
-      <div onClick={onDownloadPdf} role="button" tabIndex={0}>
+      <div role="button" onClick={onDownloadPdf} tabIndex={0}>
```
mistercrunch, line 32: Wondering if we can just bring the onClick and tabIndex to the parent (Menu.Item) and get rid of that child `<div>` altogether here

superset-frontend/plugins/legacy-preset-chart-deckgl/src/components/Legend.tsx
```diff
@@ -106,6 +106,7 @@ const Legend = ({
       <li key={k}>
         <a
```
geido, line 107: Let's create a base component for links. When onClick is present we might assign the role button. Let's also account for the possibility to override the prop

superset-frontend/src/components/CachedLabel/index.tsx
geido: Agreed

superset-frontend/src/components/Datasource/CollectionTable.tsx
```diff
@@ -312,12 +312,12 @@ export default class CRUDCollection extends React.PureComponent<
...
     if (this.state.sortColumn === col && this.state.sort === SortOrder.asc) {
-      return <Icons.SortAsc onClick={this.sortColumn(col, 2)} />;
+      return <Icons.SortAsc role="button" onClick={this.sortColumn(col, 2)} />;
     }
     if (this.state.sortColumn === col && this.state.sort === SortOrder.desc) {
-      return <Icons.SortDesc onClick={this.sortColumn(col, 0)} />;
+      return <Icons.SortDesc role="button" onClick={this.sortColumn(col, 0)} />;
     }
-    return <Icons.Sort onClick={this.sortColumn(col, 1)} />;
+    return <Icons.Sort role="button" onClick={this.sortColumn(col, 1)} />;
```
geido, line 320: Wondering if we can modify the base `Icons` component to support the conditional button role when `onClick` is used

superset-frontend/src/components/Label/index.tsx
```diff
@@ -92,6 +92,7 @@ export default function Label(props: LabelProps) {
 
   return (
     <Tag
+      role="button"
```
geido, line 95: We could apply the same logic of checking for an onClick prop on the Tag component too

superset-frontend/src/components/PopoverDropdown/index.tsx
```diff
@@ -93,7 +93,10 @@ const PopoverDropdown = (props: PopoverDropdownProps) => {
       trigger={['click']}
       overlayStyle={{ zIndex: theme.zIndex.max }}
       overlay={
-        <Menu onClick={({ key }: HandleSelectProps) => onChange(key)}>
+        <Menu
```
geido, line 96: I believe the MenuItem should get the button role but not the Menu itself

superset-frontend/src/components/Table/cell-renderers/ActionCell/index.tsx
```diff
@@ -103,7 +103,7 @@ function ActionMenu(props: ActionMenuProps) {
   };
 
   return (
-    <StyledMenu onClick={handleClick}>
+    <StyledMenu role="button" onClick={handleClick}>
```
geido, line 106: Same here. I think each item should have a button role instead of the Menu itself?

superset-frontend/src/dashboard/components/CssEditor/index.jsx
```diff
@@ -82,7 +82,7 @@ class CssEditor extends React.PureComponent {
   renderTemplateSelector() {
     if (this.props.templates) {
       const menu = (
-        <Menu onClick={this.changeCssTemplate}>
+        <Menu role="button" onClick={this.changeCssTemplate}>
```
geido, line 85: Same as other comments about the Menu getting the button role instead of its items

superset-frontend/src/dashboard/components/FiltersBadge/FilterIndicator/index.tsx
```diff
@@ -40,6 +40,7 @@ const FilterIndicator: FC<IndicatorProps> = ({
   const resultValue = getFilterValueForDisplay(value);
   return (
     <FilterItem
+      role="button"
```
geido, line 43: Similar to other comments, the button role could probably be added conditionally

superset-frontend/src/SqlLab/components/TableElement/index.tsx
```diff
@@ -345,6 +345,7 @@ const TableElement = ({ table, ...props }: TableElementProps) => {
             <Fade
               data-test="fade"
               hovered={hovered}
+              role="button"
```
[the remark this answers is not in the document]
eulloa10: This appears to be the only component where the Fade component is used.

rusackas:
Agreed with all the comments... I appreciate this PR and all moves toward accessibility. We just need to fix it as close to the root as we can.

mistercrunch:
nip

eulloa10:
@geido I pushed my latest changes, but there are still more to come. Additionally, some of the tests are failing, so I am going to look into those and update.

superset-frontend/plugins/legacy-preset-chart-deckgl/src/components/Legend.tsx
eulloa10: Per the note below, I believe we settled on leaving as is for now because this is a plugin 
https://github.com/apache/superset/pull/26602/files#r1504672422

geido: I agree otherwise we might have to face issues with circular dependencies here

superset-frontend/src/components/Icons/AntdEnhanced.tsx
```diff
@@ -25,9 +25,10 @@ import IconType from './IconType';
 const AntdEnhancedIcons = Object.keys(AntdIcons)
   .filter(k => !k.includes('TwoTone'))
   .map(k => ({
-    [k]: (props: IconType) => (
-      <StyledIcon component={AntdIcons[k]} {...props} />
-    ),
+    [k]: (props: IconType) => {
+      const whatRole = props.onClick ? 'button' : props?.role || 'img';
```
geido, line 29:
```suggestion
      const whatRole = props?.onClick ? 'button' : 'img';
```

superset-frontend/src/components/Icons/Icon.tsx
```diff
@@ -68,10 +68,13 @@ export const Icon = (props: IconProps) => {
     };
   }, [fileName, ImportedSVG]);
 
+  const whatRole = props?.onClick ? 'button' : iconProps?.role || 'img';
```
geido, line 71:
```suggestion
  const whatRole = props?.onClick ? 'button' : 'img';
```

superset-frontend/src/components/Label/index.tsx
```diff
@@ -93,6 +93,7 @@ export default function Label(props: LabelProps) {
   return (
     <Tag
       onClick={onClick}
+      role={onClick ? 'button' : undefined}
```
geido, line 96: We should move this logic inside the Tag component

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/CrossFilters/CrossFilterTag.tsx
```diff
@@ -81,6 +84,7 @@ const CrossFilterTag = (props: {
       `}
       closable
       onClose={() => removeCrossFilter(filter.emitterId)}
+      closeIcon={customCloseIcon}
```
geido, line 87: All the time you have a closable Tag, the x should get a role button, so I believe it makes sense to move this within the `Tag` component

superset-frontend/src/explore/components/controls/FixedOrMetricControl/index.jsx
```diff
@@ -127,7 +127,7 @@ export default class FixedOrMetricControl extends React.Component {
           <Collapse.Panel
             showArrow={false}
             header={
-              <Label onClick={() => undefined}>
```
geido, line 130: Can we git blame this and see why it was done in the first place?

eulloa10: This was introduced as part of a migration from Bootstrap to AntD components in this [PR](https://github.com/apache/superset/pull/12920/files#diff-a7be32a17ea0cae5c0409a1c8ef7abd669132a56052f97a24aac78570f074cef). However, there doesn't seem to be additional context as to why this `onClick` is returning `undefined`.

superset-frontend/src/components/Label/index.tsx
geido: Ok I see this `Tag` is pulled directly from Antdesign rather than our custom `Tag` component, so it makes sense to keep it here

superset-frontend/src/components/MetadataBar/MetadataBar.test.tsx
```diff
@@ -265,3 +265,27 @@ test('correctly renders the tags tooltip', async () => {
...
+    { ...ITEMS[0], onClick },
+    { ...ITEMS[1], onClick },
+  ];
+  render(<MetadataBar items={items} />);
+
+  const styledItems = screen.getAllByRole('button');
+
+  styledItems.forEach(styledItem => {
+    expect(styledItem).toHaveAttribute('role', 'button');
+  });
```
geido, line 281: This seems redundant. Maybe we can check whether these elements exist by checking the length?

superset-frontend/src/components/MetadataBar/MetadataBar.test.tsx
```diff
@@ -265,3 +265,27 @@ test('correctly renders the tags tooltip', async () => {
...
+});
+
+test('renders StyledItem with role=undefined when onClick is not defined', () => {
+  const items = [ITEMS[0], ITEMS[1]];
+  render(<MetadataBar items={items} />);
+  const styledItem = screen
+    .getByText(DASHBOARD_TITLE)
+    .closest('.metadata-text');
+  expect(styledItem).not.toHaveAttribute('role');
```
geido, line 290: Maybe we can do the opposite here and verify that the length of elements with `role=button` is zero?

superset-frontend/src/components/Tags/Tag.tsx
```diff
@@ -47,6 +50,15 @@ const Tag = ({
...
   const handleClose = () => (index ? onDelete?.(index) : null);
 
+  let whatRole;
+  if (onClick) {
+    if (!id) {
+      whatRole = 'button';
+    } else {
+      whatRole = 'link';
+    }
+  }
```
geido, line 60:
```suggestion
  const whatRole = onClick ? (!id ? 'button' : 'link') : undefined;
```

superset-frontend/src/explore/components/controls/FixedOrMetricControl/index.jsx
geido: @michael-s-molina we are keeping this here unless you give us the green light to remove it

geido (approved):
I think this is looking good. Thanks for all the work on this. Just a few nits and this should be good to go!

superset-frontend/src/components/MetadataBar/MetadataBar.test.tsx
eulloa10: Yes, this will be changed to: 
```
  const styledItems = screen.getAllByRole('button');

  expect(styledItems.length).toBe(2);
```

superset-frontend/src/components/MetadataBar/MetadataBar.test.tsx
eulloa10: This will be changed to: 
```
  const styledItems = screen.queryAllByRole('button');

  expect(styledItems.length).toBe(0);
```

geido (approved):

PR #26074, opened 2023-11-22
fix: Translations related to the date range filter

Ralkion:
### SUMMARY
This PR *partially* addresses discussion #25994 & issue #26061. It fixes the issue with the date range filter not showing translated values.

The issue was that the front-end client was sending requests that asked for strings like "Last year" and displaying the `timeRange` property in the response. As it stands now, the backend translates the human-readable string into actual dates, along with the requested string in the `timeRange` property (see [api.py#L98](https://github.com/apache/superset/blob/master/superset/views/api.py#L98). The front-end needed to translate the property coming back, but also needed to be sure that it expected the same value it sent (see changes in DateRangeFilter.tsx).

One thing I will say is that it was not clear at first how to go about fixing translation issues. I initially did the same as #26010, thinking that updating the .po and .json files was the correct path.
It wasn't until later that I realized that a 'proper' fix would follow the instructions [in the docs](https://superset.apache.org/docs/contributing/translations/). Most of this PR is actually the result of trying to resolve the issues that existed in the translation pipeline.

I'm going to outline it here. I know I was confused for a while, and it seems like others were also confused as per the discussion in #26010.
The translation is a multi-step process:
1. One should do a full extraction [/scripts/babel_update.sh](https://github.com/apache/superset/blob/master/scripts/babel_update.sh)
  - This will generate the .pot file based on the strings in the code
  - Then it will merge the .pot file with the existing .po files to the best of its ability
    - This will sometimes cause the tool to make 'guesses' at what the value should be if it doesn't already exist. Those guesses are annotated as 'fuzzy'.
2. The next step is to review/edit/fix the values in the .po file
3. Next generate the .json file(s) with the [/scripts/po2json.sh](https://github.com/apache/superset/blob/master/scripts/po2json.sh) file.
  - The .json file is used by the frontend code for translations
  - 'fuzzy' values in the .po are not translated to the .json files.
4. Finally generate the .mo file(s) with the command `pybabel compile -d superset/translations`
  - The .mo file is used by the backend code for translations
  - The .mo file is *not* checked into the git codebase. I'm unclear as to why the .json file is if the .mo file is not.
  - It is unclear looking at the UI which strings are generated by the frontend and which are generated by the backend

[...]

PR #26626, opened 2024-01-15
fix(translations): Date range filter

fatiiates:
Merges #26074. Extended to test whether pipeline bugs in #26074 persist.


fatiiates:
Hey @rusackas, there was a development in #26074 and it was stuck because of the CI/CD problems. I merged it into updated superset repository and run the workflow again. Now it works, I thought you could help us about the PR and approval process.

superset/translations/de/LC_MESSAGES/messages.json
```diff
@@ -284,6 +283,9 @@
       "A valid color scheme is required": [
         "Ein gültiges Farbschema ist erforderlich"
       ],
+      "A waterfall chart is a form of data visualization that helps in understanding\n          the cumulative effect of sequentially introduced positive or negative values.\n          These intermediate values can either be time based or category based.": [
```
john-bodley, line 286: Why do we need newlines and spaces here?

john-bodley (commented):
Thanks @fatiiates for the change. 

Out of interest any reason you didn't just push a commit to the original PR which was failing CI? Additionally it seems like your PR title is somewhat misleading at the change contains more than simply addressing various translations for the date range filter.

fatiiates:
@john-bodley thank you for your help.

I wast just opened this to try the workflow and to solve if there is an error with containerization. LGTM now, we can close this and continue with the #26074

superset/translations/de/LC_MESSAGES/messages.json
fatiiates: @john-bodley could you please move this question to #26074? @Ralkion can help you with this.

Ralkion: The short answer is that this is how it's built in the code. The first entry of each array is the 'key' that is used for lookup to get the 'value'.

In this case, you can trace the key via the messages.pot file (superset/translations/messages.pot), which is a generated file. The pot file has a comment indicating where the string originated - which for this message is (superset-frontend/plugins/plugin-chart-echarts/src/Waterfall/index.ts:53)

PR #26635, opened 2024-01-16
chore: Deprecates the ENABLE_JAVASCRIPT_CONTROLS feature flag

michael-s-molina:
### SUMMARY
As part of the [4.0 approved initiatives](https://github.com/orgs/apache/projects/292/views/1), this PR deprecates the `ENABLE_JAVASCRIPT_CONTROLS` feature flag which is currently `False` by default and used to enable programmers to customize certain charts (like the geospatial ones) by inputting javascript in controls. This exposes an XSS security vulnerability.

### TESTING INSTRUCTIONS
CI.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


john-bodley (approved):

JZ6:
will there be a new feature to replace this deprecation?

rusackas:
> will there be a new feature to replace this deprecation?

We're open to proposals! As it stands, there are no contenders, but this feature needs to be removed as a known security vector. In my mind, its replacement would be something templating-based (markdown or jinja) that is handed the full data object, and can render tooltip content. It should be something that can be added to any plugin (not just DeckGL). It should not allow arbitrary JavaScript (that's a security problem). It would definitely be subject to a SIP, and we're willing to consider a SIP from anyone who wants to contribute such a feature.

PR #26638, opened 2024-01-16
fix: Avoid 500 if end users write bad SQL

Khrol:
### SUMMARY

While end users deal with Superset+Trino we are observing quite some 500 errors on backend.

For example, a user can create/edit a virtual dataset and make mistakes in SQL. Trino client raises the exception and it goes uncaught to 500.

If SupersetDBAPIProgrammingError is raised it's better to return 400 to the end user because nothing bad happens with Superset instance.

I don't know what to do with `SupersetDBAPIError` in general. If underlying Database has problems, should it lead to Superset's 500 or it's ok to stay with 400 because Superset itself works fine.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/exceptions.py
+++ b/superset/db_engine_specs/exceptions.py
@@ -38,4 +38,4 @@ class SupersetDBAPIOperationalError(SupersetDBAPIError):
 
 
 class SupersetDBAPIProgrammingError(SupersetDBAPIError):
-    pass
+    status = 400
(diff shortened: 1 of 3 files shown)


villebro (approved):
We have lots of these unnecessary 500s, which makes logs noisy, thanks for cleaning these up 👍 While the new Trino exception mappings are probably very uncontroversial, I'd like a few extra eyes on changing the status type of `SupersetDBAPIProgrammingError`. However, I feel 400 is probably a better default than 500 for that one.

Khrol:
> While the new Trino exception mappings are probably very uncontroversial

I've initially submitted Draft PR to see at least some feedback. The situation there is a bit more complex and I've pushed more code.

> If underlying Database has problems, should it lead to Superset's 500 or it's ok to stay with 400 because Superset itself works fine?

I don't know how to interpret 500 errors with connection to underlying source Databases... 🤔

PR #26639, opened 2024-01-16
feat(components): Add static class name with button style

mskelton:
### SUMMARY

Adds a static class name to buttons based on their style (e.g. primary, secondary, etc.) to more easily allow custom CSS styling of buttons.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Button/index.tsx
+++ b/superset-frontend/src/components/Button/index.tsx
@@ -148,7 +148,14 @@ export default function Button(props: ButtonProps) {
     <AntdButton
       href={disabled ? undefined : href}
       disabled={disabled}
-      className={cx(className, 'superset-button', { cta: !!cta })}
+      className={cx(
+        className,
+        'superset-button',
+        // A static class name containing the button style is available to
+        // support customizing button styles in embedded dashboards.
+        `superset-button-${buttonStyle}`,
+        { cta: !!cta },
+      )}
       css={{
         display: 'inline-flex',
         alignItems: 'center',


shabadsobti (approved):

john-bodley:
Thanks @mskelton for the change. Given that we're changing the button class is there expected to be any UI changes? If so would you mind including before/after screenshots.

mskelton:
@john-bodley This PR adds a new class, it doesn't change or remove any existing classes, so there are no expected UI changes.

mistercrunch:
Seems superset-frontend/package-lock.json should not be affected by this PR, mind pulling it out?

rusackas:
Can you explain the use case for the class? We're a little nervous about adding code that serves no purpose to most folks.

If this is to support CSS templates at the dashboard level (where you can add custom CSS) that's one thing... we would just want to annotate the lines of code serving this purpose.

If this is to customize the button for an entire Superset installation, the correct way would be to edit the Emotion based styling in the Button component, rather than adding hooks to layer on additional CSS files. CSS and LESS have been deprecated in favor of Emotion, and we'd rather see changes made on those grounds, or even steer toward more proper Emotion-based [theming](https://preset.io/blog/theming-superset-progress-update/).

mskelton:
@rusackas The purpose is to support custom CSS for embedded dashboards.  The use case I have is to customize the controls in the dashboard (e.g. buttons, checkbox, dropdowns, etc. to align more closely with the app we are embedding superset in. Most controls I'm able to use class names from the Ant components (e.g. `.ant-menu-item`), but buttons are styled differently based on the style, however there isn't an associated class to allow me to control that.

Happy to add a comment to the line explaining the purpose of the class name.

mskelton:
@mistercrunch The package-lock.json was committed by mistake, I've reverted it.

mistercrunch (approved):
LGTM

PR #26652, opened 2024-01-17
fix(import): only import FORMULA annotations

mistercrunch:
### SUMMARY
export/import currently attempts to bring in annotations that are attached to charts, but does not handle the dependencies properly, resulting in some referential integrity-type issues. This happens when the annotation depends on either an "annotation layer", or on another chart that doesn't exist in the current environment.

The solution here is lazy, and refrains from importing any annotation that may depend on other objects that are not handled by the current import/export logic.

Longer term solution would including handling of dependencies, and allowing the users to package dependencies on export and/or ask/warn about what to do here. 

### TESTING
On top of a simple unit test on the new function I introduce, I conducted the following manual test
* exported a dashboard that contains a chart with multiple annotations of different types
* deleting the chart objects in the environment
* importing using the CLI
* verifying that only FORMULA annotation survived the export/import

### DIFF
--- a/superset/cli/importexport.py
+++ b/superset/cli/importexport.py
@@ -130,12 +130,13 @@ def export_datasources(datasource_file: Optional[str] = None) -> None:
 @click.option(
     "--path",
     "-p",
+    required=True,
     help="Path to a single ZIP file",
 )
 @click.option(
     "--username",
     "-u",
-    default=None,
+    required=True,
     help="Specify the user name to assign dashboards to",
 )
 def import_dashboards(path: str, username: Optional[str]) -> None:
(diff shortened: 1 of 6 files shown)


superset/cli/importexport.py
```diff
@@ -133,12 +133,14 @@ def export_datasources(datasource_file: Optional[str] = None) -> None:
     @click.option(
         "--path",
         "-p",
+        required=True,
```
mistercrunch, line 136: Changes to this file are not directly related to this PR, but seemed like an easy improvement while I ran manual testing.

betodealmeida (approved):
Awesome!

Yeah, I think the ideal solution would be exporting all annotation data by default, since it's what would cause least surprise IMO.

PR #26669, opened 2024-01-18
feat: prevent adhoc metrics from guest user

betodealmeida:
### SUMMARY

Ensure that guest users (from embedded dashboards) can only use predefined metrics in chart queries.

N/A

### TESTING INSTRUCTIONS

Added unit tests covering both `EMBEDDED_SUPERSET` enabled and disabled.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/common/query_object.py
+++ b/superset/common/query_object.py
@@ -27,12 +27,14 @@ from flask import g
 from flask_babel import gettext as _
 from pandas import DataFrame
 
-from superset import feature_flag_manager
+from superset import feature_flag_manager, security_manager
 from superset.common.chart_data import ChartDataResultType
+from superset.errors import ErrorLevel, SupersetError, SupersetErrorType
 from superset.exceptions import (
     InvalidPostProcessingError,
     QueryClauseValidationException,
     QueryObjectValidationError,
+    SupersetSecurityException,
 )
 from superset.sql_parse import sanitize_clause
 from superset.superset_typing import Column, Metric, OrderBy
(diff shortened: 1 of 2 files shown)


michael-s-molina:
Should this be a `fix` instead of `feat`?

superset/common/query_object.py
```diff
@@ -189,6 +191,19 @@ def _set_metrics(self, metrics: list[Metric] | None = None) -> None:
...
 
+        # When connecting with a guest token users should only be allowed to use
+        # predefined metrics.
+        if security_manager.is_guest_user() and any(
+            is_adhoc_metric(metric) for metric in metrics or []
+        ):
+            raise SupersetSecurityException(
+                SupersetError(
+                    error_type=SupersetErrorType.DASHBOARD_SECURITY_ACCESS_ERROR,
+                    message="Guest users can only use predefined metrics",
```
michael-s-molina, line 202:
```suggestion
                    message=_("Guest users can only use predefined metrics"),
```

tests/unit_tests/queries/query_object_test.py
```diff
@@ -343,3 +347,68 @@ def test_cache_key_cache_impersonation_on_with_different_user_and_db_impersonati
             ),
         ]
     )
+
+
+def test_query_object_metrics():
```
michael-s-molina, line 352: Nice!

mistercrunch:
Does this mean that if I create a dashboard today with a chart in it that uses a SQL expression metric, embed it and share it with guests it would fail? If that's the case it's likely a breaking change. 

Or does this apply only in some other context, say some sort of "explore-embedded" experience?

betodealmeida:
> Does this mean that if I create a dashboard today with a chart in it that uses a SQL expression metric, embed it and share it with guests it would fail? If that's the case it's likely a breaking change.

Yeah, this is a breaking change for security reasons. We need to think of a better way of preventing guest users from modifying the SQL.

mistercrunch:
Doing some thinking here, the user should be able to run what the dashboard designer built for them, in other words, the query that's served on the server side should be ok to run. What we don't want is for the user to be able to alter the chart definition and make up new metrics.

Even with this predefined metrics restriction, the user could in theory point to another pre-defined metric that's out of the scope of the dashboard.

Would an approach of sending a unique identifier of the chart, along with dashboard level filter would work as a better approach here. Meaning the form data sent by the client is simply a reference to a chart + the context from the dashboard, and we can trust that the query executed is the one that the dashboard designer and dashboard "embedder" cleared?

betodealmeida:
Closing in favor of a different approach, see https://github.com/apache/superset/pull/26749.

PR #26670, opened 2024-01-19
feat(translations): Script to automatically fill in translations using GPT.

rusackas:
### SUMMARY

WIP, but it does just what it says. This PR does not add any translations, it just sets up the script. All translation changes here (.po files) are just from reading and overwriting the files with no changes using the `polib` library, so they're formatted consistently. 

POC translations will be opened as a PR against this one, so we can see the result more clearly.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/scripts/complete_translations.py
+++ b/scripts/complete_translations.py
@@ -1,45 +1,49 @@
+import csv
 import glob
+import io
+
 import polib
 from langchain_community.callbacks import get_openai_callback
 from langchain_openai import OpenAI
-import json
-import csv
-import io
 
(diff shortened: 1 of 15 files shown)


geido:
This looks cool! In terms of the prompt, do you have consistent outputs from the LLM?

The LLM might train to explain what it is doing. If you end up in that issue, the magic instruction is usually `Do not explain`.

rusackas:
Closing since @mistercrunch basically upleveled this approach...

PR #26674, opened 2024-01-19
fix: helm chart comment on SECRET_KEY

dpgaspar:
### SUMMARY

Fix comment to avoid users from using the "proposed" value for the SECRET_KEY that would be a valid setting for Superset.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


craig-rueda (approved):

helm/superset/values.yaml
```diff
@@ -166,7 +166,7 @@ configOverrides: {}
   #   AUTH_USER_REGISTRATION_ROLE = "Admin"
   # secret: |
   #   # Generate your own secret key for encryption. Use openssl rand -base64 42 to generate a good key
```
villebro, line 168:
```suggestion
  #   # Generate your own secret key for encryption. Use `openssl rand -base64 42` to generate a good key
```
Could we also change this to make it easier to spot the command?

villebro (approved):
It seems our helm chart is always lagging the rest of the codebase.. :( Note to myself to try to always remember to update the chart when doing changes that affect it.

THATOMADIRA5 (approved):

PR #26679, opened 2024-01-19
fix: Revert "buld(deps): bump swagger-ui-react from 4.1.3 to 5.11.0 in docs (#26552)

michael-s-molina:
### SUMMARY
https://github.com/apache/superset/pull/26552 bumped `swagger-ui-react` from 4.1.3 to 5.11.0. As a side effect, the `react-redux` dependency in `yarn.lock` was also upgraded from 7.1.20 to 9.0.4 which requires React 18. Given that the `docs` project uses React 17, it produced the following error:

```
Attempted import error: 'useSyncExternalStore' is not exported from 'react' (imported as 'React2').
```

Another side effect of the upgrade was:

```
Module not found: Error: Can't resolve 'url' in '/home/runner/work/superset/superset/docs/
node_modules/@saucelabs/theme-github-codeblock/build/theme/ReferenceCodeBlock'
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
Error:  Client bundle compiled with errors therefore further build is impossible.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "url": require.resolve("url/") }'
	- install 'url'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "url": false }
```

This PR reverts the change.

### TESTING INSTRUCTIONS
CI.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -41,7 +41,7 @@
     "react-dom": "^17.0.1",
     "react-github-btn": "^1.2.0",
     "stream": "^0.0.2",
-    "swagger-ui-react": "^5.11.0",
+    "swagger-ui-react": "^4.1.3",
     "url-loader": "^4.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


rusackas (approved):

rusackas:
Now it seems we can't even rever this thing because there was a less-than-ideal version of `prismjs` - maybe we can bump that.

PR #26683, opened 2024-01-19
feat: auto-label PRs that contain db migrations

mistercrunch:


.github/workflows/label-miration-pr.yml
```diff
@@ -0,0 +1,25 @@
+name: Label Migration PR
```
john-bodley, line 1: There's a typo in the name, `s/miration/migration`.

john-bodley:
Thanks @mistercrunch for the PR. Would you please complete the PR description template?

rusackas:
The PR won't run new actions from a fork, so this one, so you might have to YOLO merge the action, and then open a test PR with the stub file. 

Does this also run on changed migration files, or deleted files? Looks like it would.

mistercrunch:
This is a bit of an experiment, should have been DRAFT, I was going to fill in the template if/when I can make it work.

john-bodley:
BTW @mistercrunch previously CI didn't run in `DRAFT` mode, though now it does.

PR #26693, opened 2024-01-19
chore(hail mary): Update package-lock.json via npm-audit-fix

rusackas:
### SUMMARY

Just thought this was worth a shot to see if it can pass CI. This removes our dependency warnings from:
~~(37 moderate, 33 high, 14 critical)
to
(31 moderate, 28 high, 1 critical)~~
Storybook PR got a bunch of the critical ones (bringing it down to 4) but this now brings it down to 32 (18 moderate, 13 high, 1 critical)

### TESTING INSTRUCTIONS

Good luck, CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
It seems this is mostly blocked by typescript problems... this might be fixable, which would be fantastic from a security perspective.

rusackas:
Rekindling this effort, now that storybook is in. This will probably break a whole slew of things, but I'll try to patch what I spot...

Here's the good news, if we can pull this off:
image

rusackas:
@mistercrunch @michael-s-molina this finally passed CI! With this, `vm2` is the only critical dependency, but removing it is a breaking change (killing JS tooltips) so we'll have to punt that to 5.0.

This will need testing, so I'm spinning up a test environment. I'm most worried about things that use Airbnb dependencies (data-ui, visx, vx) since those are not well maintained, and their use is not well tested. Legacy charts that use them may be quietly broken by this PR. Let me know if you want to join in on the testing effort, and we can coordinate a bit.

mistercrunch (approved):
giphy (2)

mistercrunch:
Now I'm curious to pull this branch and run `npm audit`, this is great progress! Now if we can fix https://github.com/apache/superset/pull/26944 on the Python side we'll be golden

rusackas:
Hmm... I thought ephemerals were working again... 😞

michael-s-molina:
> If we're willing to kill off some charts that have no replacement due to package vulnerabilities, we can take that route, but it sure seems like it'd need official consensus, and that ship has sailed for 4.0.

Yep, that would have been a great 4.0 proposal 😢 I do think we should do that in 5.0 though given the security implications.

rusackas:
Impacted things to test:
* Sparkline cell (Timeseries Table)
* Histogram
* _...and more?_

Results:

Histogram seems to work (even if it _is_ giving me the finger)
image

Sparkline seems to work, too!
image

rusackas:
I think the data-ui problem I was thinking of was part of the React 17 upgrade effort, and this PR doesn't break it. I'll merge this if/when CI passes.

PR #26695, opened 2024-01-19
chore(actions): chill out on docker builds

rusackas:
### SUMMARY

This needs discussion! 

Does anyone see a need for docker builds on every PR? The only purpose seems to be as a canary for docker builds on master. Now that this takes about 75 minutes of build time for every PR, I think we should just stop it, and bring back a more optimized/parallelized version if someone wants to invest in that. People keep an eye on builds on master pretty regularly, so let's just revert anything that breaks this if/when we need to.

Hopefully, this will save a lot of coal out there somewhere in the world.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,9 +3,7 @@ name: Docker
 on:
   push:
     branches:
-      - 'master'
-  pull_request:
-    types: [synchronize, opened, reopened, ready_for_review]
+      - "master"
 
 jobs:
   config:


villebro:
@rusackas I think the ephemeral environments need these docker builds. If we disable this for PRs, then I fear we won't be able to use eph envs anymore. Maybe introduce a new tag that tells CI to build docker images, or something similar?

craig-rueda:
Also, removing them from PRs will make it impossible to test any docker changes before merging.

rusackas:
Fair enough... killing this idea 🗡️

rusackas:
@craig-rueda @villebro if you can, please take a look at [this](https://github.com/apache/superset/pull/26697) then, so hopefully we can get these slow builds to be less of a blocker.

PR #25564, opened 2023-10-08
build: Parallelize the CI image builds

sebastianliebscher:
### SUMMARY

This PR replaces the custom script `docker_build_push.sh` that runs the docker builds sequentially with GitHub action jobs running in parallel.

- set tags with `docker/metadata-action` GitHub Action
- build images with `docker/build-push-action` GitHub Action
- on release tag docker image with `docker/metadata-action` `tags: type=pep440,pattern={{version}}`

No functionality changed, except:

- removes docker buildx local cache usage as every build runs on its own job hence on different machines where local cache does not apply
- docker buildx cache will be a follow-up PR to use GitHub's container registry to speed up builds

----

### TESTING INSTRUCTIONS
**Use case: PR from fork** (can be verified by the successful jobs from this PR)
- detect that no secrets are present
- images build without failures
- the subsequent steps to push the images are skipped as intended (PR from fork)
- tags: (no tags)

**Use case: PR from Superset repo**
- detect that secrets are present
- images build without failures
- images push to registry
- an example run https://github.com/sebastianliebscher/fluentd-exchangerates/actions/runs/6447721935
- tags: 
image

**Use case: commit to master**
- detect that secrets are present
- images build without failures
- images push to registry
- an example run https://github.com/sebastianliebscher/fluentd-exchangerates/actions/runs/6447623640
- tags: 
image

**Use case: release**
- detect that secrets are present
- images build without failures
- images push to registry
- an example run https://github.com/sebastianliebscher/fluentd-exchangerates/actions/runs/6447658262
- tags: 
image

----

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #25322, opened 2023-09-16
chore: add latest docker tag

eschutho:
### SUMMARY
This adds a new tag for docker with "latest" for all releases where the release tag is the most recent.
The next step would be to update docs and docker-compose files for new installs to point to the latest official release instead of master. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26698, opened 2024-01-19
build: Parallelize the CI image builds (continued)

mistercrunch:
# new approach
originally started from https://github.com/apache/superset/pull/25564, but got tangled in intricate changes around release tag management that conflicted with the approach used originally, so I started over by

1. refactoring `scripts/docker_build_push.sh`, that originally generated a bunch of docker build commands to be executed sequentially, here I move to this bash script running a single command given the parameters, so it can be reused to run multiple build commands in parallel while orchestrating in GitHub actions. I tried to DRY what was in there cautiously
1. changed GitHub actions to generate a matrix of commands for each "target" (dev, lean, lean39, lean310, ...) and "platform" (linux/amd64, linux/arm64), similar to https://github.com/apache/superset/pull/25564

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -54,6 +54,12 @@ jobs:
         run: |
           ./scripts/docker_build_push.sh "" ${{ matrix.target }} ${{ matrix.platform }}
 
+  ephemeral-docker-build:
+    needs: config
+    if: needs.config.outputs.has-secrets
+    name: docker-build
+    runs-on: ubuntu-latest
+    steps:
       - name: Build ephemeral env image
         if: github.event_name == 'pull_request'
         run: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -36,15 +36,15 @@ jobs:
       fail-fast: false
     steps:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           persist-credentials: false
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@v3
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
 
       - name: Build Docker Image
         shell: bash
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -60,6 +60,17 @@ jobs:
     name: docker-build
     runs-on: ubuntu-latest
     steps:
+      - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Build ephemeral env image
         if: github.event_name == 'pull_request'
         run: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Upload build artifacts
         if: github.event_name == 'pull_request'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: build
           path: build/
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,27 +24,35 @@ jobs:
             echo "has-secrets=0" >> "$GITHUB_OUTPUT"
             echo "no secrets!"
           fi
-
   docker-build:
     needs: config
     if: needs.config.outputs.has-secrets
     name: docker-build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: ["dev", "lean", "lean310", "lean39", "websocket", "dockerize"]
+        platform: ["linux/amd64", "linux/arm64"]
+      fail-fast: false
     steps:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/docker.yml
```diff
@@ -21,33 +24,249 @@ jobs:
...
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          target: lean
+          build-args: |
+            PY_VER=${{ matrix.image.version }}
+      - name: Login to Docker Hub
+        if: needs.config.outputs.has-secrets
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
dpgaspar, line 80: this is `secrets.DOCKERHUB_USER`

.github/workflows/docker.yml
```diff
@@ -1,12 +1,15 @@
...
+name: Build
 
 on:
+  release:
+    types: [ published ]
   push:
     branches:
       - 'master'
   pull_request:
     types: [synchronize, opened, reopened, ready_for_review]
```
dpgaspar, line 10: Just a comment: `pull_request` trigger will use the PR SHA and run any changes made on any PR. PRs from forks don't contain secrets

dpgaspar (commented):
Thanks for recovering this!
just a comment

.github/workflows/docker.yml
```diff
@@ -21,33 +24,249 @@ jobs:
...
+        with:
+          images: ${{ github.repository }}
+          flavor: |
+            latest=false
+            suffix=${{ matrix.image.suffix }}
+          tags: |
+            type=sha,prefix=,format=long
+            type=ref,event=pr
+            type=raw,value=master,enable={{is_default_branch}}
+            type=raw,value=latest,enable={{is_default_branch}}
```
eschutho, line 58: I updated the docker build at the same time that @sebastianliebscher was writing his initial PR so he didn't have a chance to capture the changes from [this PR](https://github.com/apache/superset/pull/25322), but @sfirke and I had updated the docker image `latest` tag to point to the latest official release image only and not the latest production master image so that subsequent docker deployments would run from the latest official release instead of master. There's a [script](https://github.com/apache/superset/pull/25322/files#diff-bcd99032795eb96dd42d38ceeeec262f6d4ab3952e3dbfb75905763c116cd0d7R43) that you can use from the PR to get the value for latest, or feel free to update the way that runs, too.

mistercrunch:
Ok, starting over and try a different approach

mistercrunch:
Request for comments here, given this docker build matrix:
```yaml
    strategy:
      matrix:
        target: ["dev", "lean", "lean310", "lean39", "websocket", "dockerize"]
        platform: ["linux/amd64", "linux/arm64"]
      fail-fast: false
 ```
Do we need every combination for each PR and each SHA in `master`? Seems a bit crazy. I get that people are on Apple hardware and that's why we need `arm` as a base platform, but 2 versions of python too?  @craig-rueda @rusackas @dpgaspar

File: Dockerfile [hunk 4]
```diff
@@ -85,14 +85,15 @@ RUN --mount=type=bind,target=./requirements/local.txt,src=./requirements/local.t
...
 
-COPY --chown=superset:superset --from=superset-node /app/superset/static/assets superset/static/assets
-## Lastly, let's install superset itself
-COPY --chown=superset:superset superset superset
 RUN --mount=type=cache,target=/root/.cache/pip \
     pip install -e . \
     && flask fab babel-compile --target superset/translations \
     && chown -R superset:superset superset/translations
 
+COPY --chown=superset:superset --from=superset-node /app/superset/static/assets superset/static/assets
```
mistercrunch, line 93: I think moving this AFTER the `pip install` command is going to be a huge win. Currently if you touch any JS file you'll have to re-pip install everything, where now it shouldn't be the case anymore. Frontend PRs/SHAs won't have to import and rebuild the backend.

eschutho:
> Request for comments here, given this docker build matrix:
> 
> ```yaml
>     strategy:
>       matrix:
>         target: ["dev", "lean", "lean310", "lean39", "websocket", "dockerize"]
>         platform: ["linux/amd64", "linux/arm64"]
>       fail-fast: false
> ```
> 
> Do we need every combination for each PR and each SHA in `master`? Seems a bit crazy. I get that people are on Apple hardware and that's why we need `arm` as a base platform, but 2 versions of python too? @craig-rueda @rusackas @dpgaspar

We can probably remove lean39. I added that version as a request from someone in the community who later found a workaround, and it's probably not needed. We can maybe check with @rusackas on data around who is using it, but I don't think we really need to have docker images based on multiple versions of python.

eschutho (approved):
LGTM with or without the lean39 image. 👍

dpgaspar:
> > Request for comments here, given this docker build matrix:
> > ```yaml
> >     strategy:
> >       matrix:
> >         target: ["dev", "lean", "lean310", "lean39", "websocket", "dockerize"]
> >         platform: ["linux/amd64", "linux/arm64"]
> >       fail-fast: false
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do we need every combination for each PR and each SHA in `master`? Seems a bit crazy. I get that people are on Apple hardware and that's why we need `arm` as a base platform, but 2 versions of python too? @craig-rueda @rusackas @dpgaspar
> 
> We can probably remove lean39. I added that version as a request from someone in the community who later found a workaround, and it's probably not needed. We can maybe check with @rusackas on data around who is using it, but I don't think we really need to have docker images based on multiple versions of python.

I may be missing some requirements, but totally agree, I actually don't see the point on building every flavour on every PR SHA, also on a PR from a fork, images are just built and not pushed so we are just testing them. My take would be to just leave  `lean, linux/amd64` and `websocket, linux/amd64` on every PR SHA, so be enough for testing purposes. Build push all of them when merging to master

dpgaspar (approved):

craig-rueda (approved):
LGTM

geido (approved):

michael-s-molina (approved):
LGTM

PR #26313, opened 2023-12-19
fix(dashboard): narrow empty drop area

justinpark:
### SUMMARY
We've noticed that users are having difficulty adding an element into the tab content area because of its restricted drop zone. This commit expands the drop area for empty content to resolve the problem.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:

https://github.com/apache/superset/assets/1392866/1927a803-a18f-4336-9105-13b0dbef61f7

After:

https://github.com/apache/superset/assets/1392866/bffae35e-6b9d-449e-b912-28912f6a8ecb

### TESTING INSTRUCTIONS
Edit Dashboard and drag and drop to the new tab content

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26699, opened 2024-01-19
fix(dashboard): drag and drop indicator UX

justinpark:
### SUMMARY
Following up #26313
This commit addresses and resolves the issue of an unclear drop zone indicator that was negatively impacting the user experience in the dashboard editor. 

When a component was being dragged towards the edge of the tab container or the row/column containers, multiple drop indicators were often displayed. This created confusion about the exact insertion point of the element. [fig. 1]

|fig. 1|
|--|
|drop-indicator-bug|

The root of the problem was that these dashboard components were wrapped by both draggable and droppable interfaces, which led to overlapping and conflicting drop zones. This commit modifies this by making the dashboard components draggable only, and builds a distinct, non-conflicting area for the drop zone. Moreover, it also highlights the drop zone during the dragging process to clearly indicate where the element will be placed. [fig. 2]

|fig. 2|
|--|
|drop-indicator-after|

Before:

https://github.com/apache/superset/assets/1392866/315309e9-a8ad-45c4-a292-190a318c8d67

After:

https://github.com/apache/superset/assets/1392866/2342eee1-418d-4c35-9a21-f5772136c0ec

### TESTING INSTRUCTIONS
Go to dashboard and edit mode
Drag and drag multiple components and then save
Verify the dashboard shown as designed

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/DashboardBuilder/DashboardBuilder.tsx
+++ b/superset-frontend/src/dashboard/components/DashboardBuilder/DashboardBuilder.tsx
@@ -44,7 +44,7 @@ import BuilderComponentPane from 'src/dashboard/components/BuilderComponentPane'
 import DashboardHeader from 'src/dashboard/containers/DashboardHeader';
 import Icons from 'src/components/Icons';
 import IconButton from 'src/dashboard/components/IconButton';
-import DragDroppable from 'src/dashboard/components/dnd/DragDroppable';
+import { Droppable } from 'src/dashboard/components/dnd/DragDroppable';
 import DashboardComponent from 'src/dashboard/containers/DashboardComponent';
 import WithPopoverMenu from 'src/dashboard/components/menu/WithPopoverMenu';
 import getDirectPathToTabIndex from 'src/dashboard/util/getDirectPathToTabIndex';
@@ -81,6 +81,7 @@ import {
   MAIN_HEADER_HEIGHT,
   OPEN_FILTER_BAR_MAX_WIDTH,
   OPEN_FILTER_BAR_WIDTH,
+  EMPTY_CONTAINER_Z_INDEX,
 } from 'src/dashboard/constants';
 import { getRootLevelTabsComponent, shouldFocusTabs } from './utils';
 import DashboardContainer from './DashboardContainer';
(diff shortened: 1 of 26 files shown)


michael-s-molina:
@sadpandajoe

yousoph:
@justinpark this looks and feels so much better! definitely a much needed improvement :) 

A few things I noticed: 

1. When adding tabs to a dashboard, the drop area extends a bit too far, into the right panel
image
2. When adding a chart element, the dashboard header row is also highlighted as a possible drop area 
3. I managed to get an error when dragging a chart in between the header and the dashboard area somewhere, but I can't figure out how to repro it :( 
image

kasiazjc:
> @justinpark this looks and feels so much better! definitely a much needed improvement :)
> 
> A few things I noticed:
> 
> 1. When adding tabs to a dashboard, the drop area extends a bit too far, into the right panel
>    image
> 2. When adding a chart element, the dashboard header row is also highlighted as a possible drop area
> 3. I managed to get an error when dragging a chart in between the header and the dashboard area somewhere, but I can't figure out how to repro it :(
> 
> image

agree with Sophie on all of the above! 

Definitely an improvement to the dnd, thank you @justinpark ☺️ 

one comment about the header thing Sophie mentioned: when you try to hover over it it shows a red box (error), which gave me an idea. I think maybe it could be used in places where there is not enough space to place the item, for example if the row is full. What do you think? 

image

justinpark:
@yousoph Sorry for the late response. Here are my answers.

> 1. When adding tabs to a dashboard, the drop area extends a bit too far, into the right panel

Given that the tab (in the header) will occupy the whole row, including the right panel as shown in the following screenshot, it seems more logical for this highlight area to indicate its transition into a global header layout.

Screenshot_2024-02-07_at_4_05_40 PM

> 2. When adding a chart element, the dashboard header row is also highlighted as a possible drop area

As discussed during our plans for further D&D improvements, I've earmarked this particular enhancement for @rtexelm's project. In the interim, this version will highlight in red (instead of the primary color) when you hover over the drop area, as shown in the following screenshot.

chart_disabled_area

> 3. I managed to get an error when dragging a chart in between the header and the dashboard area somewhere, but I can't figure out how to repro it :(

I tried to reproduce the same problem but cannot get the issue yet. please let me know if you find the problem again.

justinpark:
> one comment about the header thing Sophie mentioned: when you try to hover over it it shows a red box (error), which gave me an idea. I think maybe it could be used in places where there is not enough space to place the item, for example if the row is full.

I had a similar thought as well, but there are complexities when it comes to dropping an existing component into a space that's just enough. 
Specifically, it's challenging to programmatically differentiate this from the prohibited drop zones.

(For instance, the highlight box should not turn red when item [1] is dropped into a zone without space, as it's possible to reorder among the same row items. However, it should turn red when item [4] is dropped into a zone without space.)
|[ 1 ]|[ 2 ]|[ 3 ][no space drop zone]|
--------
|[ 4 ]|[             ]|

Therefore, I've decided to retain the post notification as it is currently displayed.

justinpark:
@kasiazjc @yousoph, barring any further objections from your end, I intend to proceed to the next step.

michael-s-molina (approved):
LGTM

PR #26324, opened 2023-12-20
fix(tagging): adding tags containing a “:” to dashboards

lilykuang:
### SUMMARY

- Issue with Tagging for Dashboards: There is a problem with saving tags that have the “:” character for dashboards. These tags cannot be assigned to a dashboard, although they can be successfully assigned to charts.
- removed the check on ":" or ","

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26701, opened 2024-01-20
fix(tags): Filter system tags from the tags list

Vitor-Avila:
### SUMMARY
System tags used to be filtered from the Tags list (by filtering tags with `:` on their name). As a consequence, any user-created tag that included `:` were not listed as well. https://github.com/apache/superset/pull/26324 fixed the issue by showing PRs with `:`, which re-introduced system tags to the table. 

This PR adds a `custom_tag` boolean filter operator to the Tags API so that system tags are properly filtered out from the API response. Filtering these tags in the Client/Frontend side is not effective since your first page might be only system tags, so you would end up with 0 results in the table (and no option to paginate) + you would also end up with inconsistent counts per page.

**Before**

image

**After**

image

### TESTING INSTRUCTIONS
1. Make sure that `TAGGING_SYSTEM` is enabled. 
2. Add any account as an owner to any asset. 
3. Navigate to **Settings > Tags**.
4. Validate that the system tag (`owner:{{user_id}}`) is not listed.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [x] Required feature flags: `TAGGING_SYSTEM`
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/pages/Tags/index.tsx
+++ b/superset-frontend/src/pages/Tags/index.tsx
@@ -59,6 +59,17 @@ function TagList(props: TagListProps) {
   const { addDangerToast, addSuccessToast, user } = props;
   const { userId } = user;
 
+  const initialFilters = useMemo(
+    () => [
+      {
+        id: 'type',
+        operator: 'custom_tag',
+        value: true,
+      },
+    ],
+    [],
+  );
+
   const {
     state: {
       loading,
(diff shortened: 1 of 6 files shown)


superset-frontend/src/pages/Tags/index.tsx
```diff
@@ -59,6 +59,17 @@ function TagList(props: TagListProps) {
   const { addDangerToast, addSuccessToast, user } = props;
   const { userId } = user;
 
+  const initialFilters = useMemo(
+    () => [
+      {
+        id: 'type',
+        operator: 'custom_tag',
```
eschutho, line 66: Is there ever a case where the client would need to use this api to return system-generated tags? I'm wondering if this filter should just be applied by default server-side.

Vitor-Avila: @eschutho I thought about that as well (wasn't sure if we ever wanted to expose those system-tags). I think the biggest advantage with it being a filter is the flexibility. 

Could be useful for example in case an admin wants to retrieve all assets favorited by user X, or owned by user Y.

eschutho (approved):
LGTM. Thanks @Vitor-Avila!

PR #26702, opened 2024-01-20
chore: fix GitHub 'Unchanged files with check annotations' reports in PR

mistercrunch:
### SUMMARY
Addressing the GitHub-reported lint warnings that show up in every PR. It's unclear to me why those warnings are surfaced by not others, but let's silence those clean up future PR from this. Note that there are 255 other warnings that are not surfaced in PRs, wondering if they do some sort of top N (?). We'll know soon enough.

Screenshot 2024-01-20 at 1 40 52 PM

Related issue filed with GitHub about this annoying beta feature -> https://github.com/actions/toolkit/issues/457

### DIFF
--- a/.github/workflows/superset-websocket.yml
+++ b/.github/workflows/superset-websocket.yml
@@ -23,7 +23,7 @@ jobs:
         run: npm ci
       - name: eslint
         working-directory: ./superset-websocket
-        run: npm run eslint -- .
+        run: npm run eslint -- . --quiet
       - name: typescript checks
         working-directory: ./superset-websocket
         run: npm run type
--- a/superset-frontend/cypress-base/cypress/e2e/dashboard/drillby.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/dashboard/drillby.test.ts
@@ -77,6 +77,7 @@ const drillBy = (targetDrillByColumn: string, isLegacy = false) => {
 
 const verifyExpectedFormData = (
   interceptedRequest: Interception,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   expectedFormData: Record<string, any>,
 ) => {
   const actualFormData = interceptedRequest.request.body?.form_data;
--- a/superset-frontend/cypress-base/cypress/e2e/dashboard/editmode.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/dashboard/editmode.test.ts
@@ -88,6 +88,7 @@ function visitEdit(sampleDashboard = SAMPLE_DASHBOARD_1) {
 }
 
 function resetTabbedDashboard(go = false) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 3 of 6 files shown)


superset-frontend/package.json
```diff
@@ -37,7 +37,7 @@
     "src/setup/*"
   ],
   "scripts": {
-    "_lint": "eslint --ignore-path=.eslintignore --ext .js,.jsx,.ts,tsx .",
+    "_lint": "eslint --ignore-path=.eslintignore --ext .js,.jsx,.ts,tsx",
```
mistercrunch, line 40: removing the `.` at the end so this private base command can be reused more generically

superset-frontend/package.json
```diff
@@ -51,8 +51,9 @@
     "dev": "webpack --mode=development --color --watch",
     "dev-server": "cross-env NODE_ENV=development BABEL_ENV=development node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --mode=development",
     "format": "npm run _prettier -- --write",
-    "lint": "npm run _lint && npm run type",
-    "lint-fix": "npm run _lint -- --fix && npm run type",
+    "lint": "npm run _lint -- .",
```
mistercrunch, line 54: using base command, but maintaining original behavior, even though it's a `_` private command. Also removing `tsc` command to run here. It's a different linter and type of checks altogether, let's break this apart.

superset-frontend/package.json
```diff
@@ -51,8 +51,9 @@
...
+    "lint-fix": "npm run _lint . -- --fix",
```
mistercrunch, line 55: breaking it apart from tsc here too

mistercrunch:
Oh it looks like the "Unchanged files with check annotations" is reporting some top N of `eslint` warnings, not this particular subset I tackled in this PR.

geido (approved):

mistercrunch:
I'm going to lookup a way to shut off the stupid unrelated lint warnings as I really don't think they help. We know we have lint warning, and they really don't help in the context of an unrelated PR.

mistercrunch:
Commented [here](https://github.com/actions/toolkit/issues/457), going to look and see if there's a way to run `eslint` with no-warnings in CI

mistercrunch:
I think prior PR fixed https://github.com/apache/superset/pull/26704 the reporting of these, with `--quiet` now properly taking effect as it's passed to `eslint`, not `tsc` like it was before. Let's enjoy some squeaky clean PRs with no unrelated smell warnings.

rusackas (approved):
Meant to review this sooner... I thought some of these `estlint-disable` bits might be in _actual UI files_ files. I don't think we should be sweeping `any` under the rug in those cases but I don't think it's a problem in these test files.

mistercrunch:
Yeah in the end what fixed the annoying messages in PRs was to pass a `--quiet` flag to eslint in the context of CI. The warnings are still very much there, just not in CI / PRs.

PR #26704, opened 2024-01-20
chore: improve/decouple eslint and tsc 'npm run' commands

mistercrunch:
### SUMMARY
I was having issues running just `tsc` or `eslint` using our `npm run` scripts in `package.json`, because of the way they were set up. Comment line-per-line in the PR here.

1. main goal of the PR: improve `npm run` commands
2. secondary mission: break down some linting CI steps into sub-step (one for eslint, and one for tsc), allowing better time accounting and error/warning attribution

### DIFF
--- a/.github/workflows/superset-frontend.yml
+++ b/.github/workflows/superset-frontend.yml
@@ -40,7 +40,7 @@ jobs:
         if: steps.check.outcome == 'failure'
         working-directory: ./superset-frontend
         run: |
-          npm run lint -- --quiet
+          npm run eslint -- . --quiet
           npm run type
           npm run prettier-check
       - name: Build plugins packages
--- a/.github/workflows/superset-frontend.yml
+++ b/.github/workflows/superset-frontend.yml
@@ -41,6 +41,7 @@ jobs:
         working-directory: ./superset-frontend
         run: |
           npm run lint -- --quiet
+          npm run type
           npm run prettier-check
       - name: Build plugins packages
         if: steps.check.outcome == 'failure'
--- a/.github/workflows/superset-frontend.yml
+++ b/.github/workflows/superset-frontend.yml
@@ -36,12 +36,20 @@ jobs:
         uses: ./.github/actions/cached-dependencies
         with:
           run: npm-install
-      - name: lint
+      - name: eslint
         if: steps.check.outcome == 'failure'
         working-directory: ./superset-frontend
         run: |
           npm run eslint -- . --quiet
+      - name: tsc
+        if: steps.check.outcome == 'failure'
+        working-directory: ./superset-frontend
+        run: |
           npm run type
+      - name: prettier
+        if: steps.check.outcome == 'failure'
+        working-directory: ./superset-frontend
+        run: |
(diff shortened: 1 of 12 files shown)


superset-frontend/package.json
```diff
@@ -37,7 +37,7 @@
     "src/setup/*"
   ],
   "scripts": {
-    "_lint": "eslint --ignore-path=.eslintignore --ext .js,.jsx,.ts,tsx .",
+    "eslint": "eslint --ignore-path=.eslintignore --ext .js,.jsx,.ts,tsx",
```
mistercrunch, line 40: killing `_lint` with a simple `eslint` command with the parameters that should always be used, but **importantly** removing that `.` at the end that prevents using this construct to run a specific folder of file

superset-frontend/package.json
```diff
@@ -51,8 +51,8 @@
     "dev": "webpack --mode=development --color --watch",
     "dev-server": "cross-env NODE_ENV=development BABEL_ENV=development node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --mode=development",
     "format": "npm run _prettier -- --write",
-    "lint": "npm run _lint && npm run type",
-    "lint-fix": "npm run _lint -- --fix && npm run type",
+    "lint": "npm run eslint -- . && npm run type",
```
mistercrunch, line 54: maintaining backward compatibility, though I think we should deprecate eventually as it's unclear that you're running two commands here, and arguably `tsc` isn't really linting.

superset-frontend/package.json
```diff
@@ -51,8 +51,8 @@
...
+    "lint-fix": "npm run eslint -- . --fix",
```
mistercrunch, line 55: same here, maintaining backward compat

.github/workflows/superset-frontend.yml
```diff
@@ -36,11 +36,20 @@ jobs:
         uses: ./.github/actions/cached-dependencies
         with:
           run: npm-install
-      - name: lint
+      - name: eslint
```
mistercrunch, line 39: breaking down the steps in CI makes it more clear where the output is coming from, and gets timing info on the different steps. Helping understand the time it takes for each step

superset-frontend/js_build.sh
```diff
@@ -20,6 +20,7 @@ cd "$(dirname "$0")"
 npm --version
 node --version
 time npm ci
-time npm run lint
+time npm run eslint -- .
```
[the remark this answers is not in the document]
mistercrunch: `--` from my understanding is a `npm run` feature where it allows you to pass arguments to the underlying command. npm will basically concatenate what's on the right of `--` with the script / string you define. 

In this case, I designed the `npm run eslint` to not include the `.` like it did before, so that a user that knows these conventions could run `npm run eslint -- myfile.js` or `npm run eslint -- --fix .`. Somehow `npm run type` script is `tsc --noEmit`, it doesn't require a file argument like eslint

geido (approved):
Thanks for this!

PR #26708, opened 2024-01-22
chore(dependencies): loosen constraints on dependency checker

rusackas:
### SUMMARY

Right now, the dependency checker action blocks PRs if they introduce a package with a "high" severity CVE. This is becoming counterproductive in several PRs, particularly as we mess around with cleaning up packages. For example, you might fix one or more `critical` issues by upgrading to packages that include a `high` issue in their dependency tree. Let's just block on `critical` for now, and tigthen the screws after 4.0, perhaps.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -4,7 +4,7 @@
 #
 # Source repository: https://github.com/actions/dependency-review-action
 # Public documentation: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement
-name: 'Dependency Review'
+name: "Dependency Review"
 on: [pull_request]
 
 permissions:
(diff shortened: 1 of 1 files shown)


mistercrunch:
Is there any way to make it warn on `high` but not block. I mean in theory it's good to know if there's a known issue with a new package being brought in, but clearly it misses the mark as vulnerabilities are often discovered after the package is introduced.

rusackas:
No, the action either blocks or it doesn't.... there's no warning system (or at least not one people will pay attention to). The moderate/high/critical warnings people are mostly likely to see are the ones that show up when they do `npm install` -where you'll currently see 14 critical warnings. In some cases, it's impossible to fix the `critical` vulnerabilities without introducing `high` vulnerabilities, thus the need to (temporarily) lower this threshold.

mistercrunch (approved):

PR #26710, opened 2024-01-22
fix(dependencies): stopping (and preventing) full lodash library import... now using only method level imports.

rusackas:
### SUMMARY

There were a few spots where the entire lodash library was being loaded, bloating our bundles for no good reason. Now there's an eslint rule to prevent this, and the code has been adjusted to use/allow only method level imports.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


kgabryje:
According to [lodash docs](https://lodash.com/per-method-packages), [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) (which we are using) makes imports like `import { x } from "lodash"` tree-shakable just like `import x from lodash/x`. Do you know which module appears to import the whole lodash without tree shaking?

kgabryje:
I see that some files are using `import _ from 'lodash';` which is definitely not treeshakable. Maybe it'd be enough to fix just those few cases?

michael-s-molina:
+1 to @kgabryje's comments.

rusackas:
It's always a slippery slope...

Indeed, there were a few places importing `_` which loads the whole darn thing. I fixed those, and this was a nice, small PR.

Then, I thought "gee, we should prevent this from happening, is there a lint rule?" and there is! So installed it, and found a catch... while you can block the full import, it forces consistency between `module` and `method` imports... it doesn't let you mix and match. 

I chose method because (a) it was a simpler, smaller migration, and (b) this produces a reliably small bundle, even in an environment where we might not trust our tree-shaking fully (and I'm not sure if I do).

michael-s-molina:
@rusackas Is it possible to replace the occurrences of `_` with a named import `{ map }` and use `member` instead of `method`?

michael-s-molina:
I think the main reason for using named imports/exports is because is the default syntax in Superset.

kgabryje:
@rusackas Is there any chance that that eslint plugin could flag only default imports from `"lodash"`, but not default imports from `"lodash/feature_name"` and named imports from `"lodash"`? I think that would be perfect for us

rusackas:
Nope, it doesn't support that. I can revise this PR to do module imports if this pattern bothers people, or close this one and open a new PR... whichever is easier. It's not a priority, so that would happen on some other rainy day (I opened this on a Sunday night on a whim)

michael-s-molina:
>  I can revise this PR to do module imports if this pattern bothers people

Do you mean `member` imports? The reason I'm asking is because we don't need the whole module.

kgabryje:
> Nope, it doesn't support that. I can revise this PR to do module imports if this pattern bothers people, or close this one and open a new PR... whichever is easier. It's not a priority, so that would happen on some other rainy day (I opened this on a Sunday night on a whim)

Oh that's a shame... In such case, I think it's more important to prevent default imports from lodash than to stick to named imports style - especially since if we let just 1 bad import slip past code reviews, it will include the whole library in our bundle.

rusackas:
OK, everything is now converted to `member` rather than `method` :) Let's hope it all builds correctly now, and our bundles get smaller!

villebro (approved):
NICE! 👏  I had no idea we had these still lingering around. LGTM! 👍

michael-s-molina (approved):
LGTM. Thanks for working on reducing bundle size!

superset-frontend/packages/generator-superset/generators/plugin-chart/index.js
```diff
@@ -20,7 +20,14 @@
 /* eslint-disable sort-keys */
 
 const Generator = require('yeoman-generator');
-const _ = require('lodash');
```
kgabryje, line 23: Can't we do `const { kebabCase, startCase, camelCase, upperFirst } = require('lodash');` instead?

geido (approved):

superset-frontend/packages/generator-superset/generators/plugin-chart/index.js
rusackas: Just saw this after merging..... but... yes I suppose we can. I'll open a smaller/less contentious PR on some rainy day

PR #26736, opened 2024-01-22
feat(legacy-plugin-chart-map-box): Add fixed initial viewport settings option

### LINKED ISSUE #17269: Cannot set default latitude/longitude in MapBox chart
#### How to reproduce the bug

1. Create a new Chart
2. Select MapBox
3. Select latitude and longitude
4. Click on Customize
5. Zoom and pan on the map

### Expected results

Change viewport default parameter.

### Actual results

Latitude and longitude do not change, while the zoom parameter is glitchy.

### Environment

- browser type and version: Firefox 93
- superset version: `superset version` the latest Superset from Docker

### Checklist

- [X] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [X] I have reproduced the issue with at least the latest released version of superset.
- [X] I have checked the issue tracker for the same issue and I haven't found one similar.

asyncjason:
### SUMMARY

It's currently not possible to fix the initial map box view. The current settings are not used when setting up the initial view. 
The initial is auto calculated from bounds, and in some cases doesn't seem to place all the points on the screen.
This PR adds a drop down to enable a "fixed" initial view option which will use the user entered settings as the initial view instead of the calculation from bounds.

image

### TESTING INSTRUCTIONS
By default, the MapBox chart will still auto calculate the initial view from the bounds.
To test the fixed option, set the Initial Viewport settings dropdown to fixed and update the default lat/long/zoom

### ADDITIONAL INFORMATION

- [x ] Has associated issue:
Fixes https://github.com/apache/superset/issues/17269


jseparovic:
The last commit fixes an issue where the plots do not load until the map is dragged.
The first time the component loads after login it works fine, but after that, the plots do not load until dragged.
Putting the clusters in state fixes the issue.

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -69,87 +95,82 @@ class MapBox extends React.Component {
...
+      const offsetVertical = (height * 0.5) / 100;
+      const bbox = [
+        bounds[0][0] - offsetHorizontal,
+        bounds[0][1] - offsetVertical,
+        bounds[1][0] + offsetHorizontal,
+        bounds[1][1] + offsetVertical,
+      ];
+      setClusters(clusterer.getClusters(bbox, Math.round(viewport.zoom)));
+    }
+  }, [clusterer, viewport]);
```
justinpark, line 129: Given that `clusters` is assigned solely based on the `viewport` value, utilizing `useMemo` would be a more efficient approach compared to a combination of `useState` and `useEffect`.

```suggestion
  useEffect(() => {
    if (viewport) {
      // Compute the clusters based on the original bounds and current zoom level. Note when zoom/pan
      // to an area outside of the original bounds, no additional queries are made to the backend to
      // retrieve additional data.
      // add this variable to widen the visible area
      const offsetHorizontal = (width * 0.5) / 100;
      const offsetVertical = (height * 0.5) / 100;
      const bbox = [
        bounds[0][0] - offsetHorizontal,
        bounds[0][1] - offsetVertical,
        bounds[1][0] + offsetHorizontal,
        bounds[1][1] + offsetVertical,
      ];
      setClusters(clusterer.getClusters(bbox, Math.round(viewport.zoom)));
    }
  }, [clusters, viewport]);
  const clusters = useMemo(() => {
    if (viewport) {
      // Compute the clusters based on the original bounds and current zoom level. Note when zoom/pan
      // to an area outside of the original bounds, no additional queries are made to the backend to
      // retrieve additional data.
      // add this variable to widen the visible area
      const offsetHorizontal = (width * 0.5) / 100;
      const offsetVertical = (height * 0.5) / 100;
      const bbox = [
        bounds[0][0] - offsetHorizontal,
        bounds[0][1] - offsetVertical,
        bounds[1][0] + offsetHorizontal,
        bounds[1][1] + offsetVertical,
      ];
      return clusterer.getClusters(bbox, Math.round(viewport.zoom));
    }
    return undefined;
  }, [viewport]);
```

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -69,87 +95,82 @@ class MapBox extends React.Component {
...
-      bounds[0][0] - offsetHorizontal,
-      bounds[0][1] - offsetVertical,
-      bounds[1][0] + offsetHorizontal,
-      bounds[1][1] + offsetVertical,
-    ];
-    const clusters = clusterer.getClusters(bbox, Math.round(viewport.zoom));
-
-    return (
-      <MapGL
+  const isDragging =
```
justinpark, line 135: It's equivalent to `Boolean()`
```suggestion
  const isDragging = Boolean(viewport.isDragging);
```

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -69,87 +95,82 @@ class MapBox extends React.Component {
...
+    viewport.isDragging === undefined ? false : viewport.isDragging;
+
+  const handleViewportChange = newViewport => {
+    setViewport(newViewport);
+    const { onViewportChange } = props;
```
justinpark, line 140: It would be more efficient to consolidate all prop spreading in a single location, such as line 83.

justinpark (commented):
@jseparovic, thank you for your contribution to the mapbox options. I've left some comments on the PR that might be of assistance. Your efforts are greatly appreciated!

jseparovic:
@justinpark Thanks for the review! I updated the PR to use useMemo as per your suggestion, but unfortunately it broke the fix for when the plots do not load until the map is dragged. The latest commit I pushed shows the working code where i added `// ++` comments for the added lines and `// --` for the removed lines so you can see what the code looked like with useMemo.
Any ideas on how to fix this bug with useMemo? Essentially the issue is that when the map loaded for the first time after logging in, it works ok, but then if you go back to that dashboard, or refresh the page, the plots do not load until you drag the map. The useEffect+useState fixes this, but the useMemo does not.

rusackas (approved):
I think this looks OK to me, but might like a second pair of eyes from @michael-s-molina who is uniquely aware of the risks of upgrading/migrating form data schemas from version to version.

rusackas:
@jseparovic I think this just needs the [precommit hooks] run - one of the linters (prettier, specifically) is not happy with the changelog file.

rusackas:
Marking this as draft until it can be made mergeable via pre-commit fixes and a rebase. Please feel free to mark it as ready for review when it's lookin' good, otherwise it may be closed in time if it doesn't get any love.

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -18,7 +18,7 @@
  */
 /* eslint-disable react/jsx-sort-default-props, react/sort-prop-types */
 /* eslint-disable react/forbid-prop-types, react/require-default-props */
-import { Component } from 'react';
+import { Component, useEffect, useMemo } from 'react';
```
rusackas:
```suggestion
import { useEffect } from 'react';
```

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -55,11 +59,35 @@ const defaultProps = {
...
+    mapboxApiKey,
+    pointRadius,
+    pointRadiusUnit,
+    renderWhileDragging,
+    rgb,
+    hasCustomMetric,
+    onViewportChange,
+  } = props;
+
+  const [viewport, setViewport] = React.useState();
```
rusackas:
```suggestion
  const [viewport, setViewport] = useState();
```

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -55,11 +59,35 @@ const defaultProps = {
...
+  // ++
+  const [clusters, setClusters] = React.useState();
```
rusackas:
```suggestion
  const [clusters, setClusters] = useState();
```

superset-frontend/plugins/legacy-plugin-chart-map-box/src/MapBox.jsx
```diff
@@ -18,7 +18,7 @@
  */
 /* eslint-disable react/jsx-sort-default-props, react/sort-prop-types */
 /* eslint-disable react/forbid-prop-types, react/require-default-props */
-import { Component } from 'react';
+import { useEffect } from 'react';
```
rusackas:
```suggestion
import { useEffect, useState } from 'react';
```

rusackas:

PR #26744, opened 2024-01-22
fix(db2): Improving support for ibm db2 connections

Vitor-Avila:
### SUMMARY
This PR fixes some issues with the IBM Db2 engine spec:

#### Connection issues
When trying to establish a connection to an IBM Db2 Cloud Instance, below error is displayed in case `ibm_db_sa==0.3.8` is installed:

image

To fix this, I updated the version limits in `setup.py`.

#### Unable to create a physical dataset
Trying to create a virtual dataset would return an error, because `get_table_comment()` returns a `Tuple()` for this engine:

image

stack trace:
```
2024-01-22 14:04:37,371:ERROR:flask_appbuilder.api:Object of type LegacyRow is not JSON serializable
  File "/Users/vitoravila/.pyenv/versions/3.10.12/envs/superset-mods/lib/python3.10/site-packages/flask_appbuilder/api/__init__.py", line 110, in wraps
    return f(self, *args, **kwargs)
  File "/Users/vitoravila/code/superset-mods/superset/superset/views/base_api.py", line 127, in wraps
    raise ex
  File "/Users/vitoravila/code/superset-mods/superset/superset/views/base_api.py", line 121, in wraps
    duration, response = time_function(f, self, *args, **kwargs)
  File "/Users/vitoravila/code/superset-mods/superset/superset/utils/core.py", line 1454, in time_function
    response = func(*args, **kwargs)
  File "/Users/vitoravila/code/superset-mods/superset/superset/utils/log.py", line 255, in wrapper
    value = f(*args, **kwargs)
  File "/Users/vitoravila/code/superset-mods/superset/superset/databases/api.py", line 749, in table_metadata
    return self.response(200, **table_info)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/envs/superset-mods/lib/python3.10/site-packages/flask_appbuilder/api/__init__.py", line 772, in response
    _ret_json = jsonify(kwargs)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/envs/superset-mods/lib/python3.10/site-packages/flask/json/__init__.py", line 342, in jsonify
    return current_app.json.response(*args, **kwargs)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/envs/superset-mods/lib/python3.10/site-packages/flask/json/provider.py", line 309, in response
    f"{self.dumps(obj, **dump_args)}\n", mimetype=mimetype
  File "/Users/vitoravila/.pyenv/versions/3.10.12/envs/superset-mods/lib/python3.10/site-packages/flask/json/provider.py", line 230, in dumps
    return json.dumps(obj, **kwargs)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/lib/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/lib/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/Users/vitoravila/.pyenv/versions/3.10.12/lib/python3.10/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/Users/vitoravila/.pyenv/versions/3.10.12/envs/superset-mods/lib/python3.10/site-packages/flask/json/provider.py", line 122, in _default
    raise TypeError(f"Object of type {type(o).__name__} is not JSON serializable")
```

To fix this, I created a db2-specific `get_table_comment()` method in the `Db2EngineSpec`.

#### Unable to execute un-classified queries in SQL Lab
It was not possible to run a query without specifying the schema in SQL Lab (even if the schema was selected in the dropdown):

image

To fix this, I created a db2-specific `get_prequeries()` method in the `Db2EngineSpec`, and also set `supports_dynamic_schema` to `True`.

Screenshots for **Before** added in Summary. After this PR:

image

image

image

### TESTING INSTRUCTIONS
Unit tests added. To manually test:
1. Make sure you're able to connect to an IBM Db2 instance.
2. Validate you're able to create physical datasets.
3. Check you're able to run a SQL query without specifying the schema in the SQL statement as long as it's selected in the dropdown. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/db2.py
+++ b/superset/db_engine_specs/db2.py
@@ -14,8 +14,15 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import logging
+from typing import Optional, Union
+
+from sqlalchemy.engine.reflection import Inspector
+
 from superset.constants import TimeGrain
 from superset.db_engine_specs.base import BaseEngineSpec, LimitMethod
+
+logger = logging.getLogger(__name__)
 
 
 class Db2EngineSpec(BaseEngineSpec):
@@ -25,6 +32,8 @@ class Db2EngineSpec(BaseEngineSpec):
     limit_method = LimitMethod.WRAP_SQL
     force_column_alias_quotes = True
     max_column_name_length = 30
+
+    supports_dynamic_schema = True
 
     _time_grain_expressions = {
         None: "{col}",
@@ -52,3 +61,49 @@ class Db2EngineSpec(BaseEngineSpec):
     @classmethod
     def epoch_to_dttm(cls) -> str:
         return "(TIMESTAMP('1970-01-01', '00:00:00') + {col} SECONDS)"
+
+    @classmethod
+    def get_table_comment(
+        cls, inspector: Inspector, table_name: str, schema: Union[str, None]
+    ) -> Optional[str]:
+        """
+        Get comment of table from a given schema
+
+        Ibm Db2 return comments as tuples, so we need to get the first element
+
+        :param inspector: SqlAlchemy Inspector instance
+        :param table_name: Table name
+        :param schema: Schema name. If omitted, uses default schema for database
+        :return: comment of table
+        """
+        comment = None
+        try:
+            table_comment = inspector.get_table_comment(table_name, schema)
+            comment = table_comment.get("text")
+            return comment[0]
+        except IndexError:
+            return comment
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.error("Unexpected error while fetching table comment", exc_info=True)
+            logger.exception(ex)
+            return comment
+
+    @classmethod
+    def get_prequeries(
+        cls,
+        catalog: Union[str, None] = None,
+        schema: Union[str, None] = None,
+    ) -> list[str]:
+        """
+        Set the search path to the specified schema.
+
+        This is important for two reasons: in SQL Lab it will allow queries to run in
+        the schema selected in the dropdown, resolving unqualified table names to the
+        expected schema.
+
+        But more importantly, in SQL Lab this is used to check if the user has access to
+        any tables with unqualified names. If the schema is not set by SQL Lab it could
+        be anything, and we would have to block users from running any queries
+        referencing tables without an explicit schema.
+        """
+        return [f'set current_schema "{schema}"'] if schema else []
--- a/setup.py
+++ b/setup.py
@@ -151,7 +151,7 @@ setup(
             "databricks-sql-connector>=2.0.2, <3",
             "sqlalchemy-databricks>=0.2.0",
         ],
-        "db2": ["ibm-db-sa>=0.3.5, <0.4"],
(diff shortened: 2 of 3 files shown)


superset/db_engine_specs/db2.py
```diff
@@ -52,3 +61,49 @@ class Db2EngineSpec(BaseEngineSpec):
...
+        :param inspector: SqlAlchemy Inspector instance
+        :param table_name: Table name
+        :param schema: Schema name. If omitted, uses default schema for database
+        :return: comment of table
+        """
+        comment = None
+        try:
+            table_comment = inspector.get_table_comment(table_name, schema)
+            comment = table_comment.get("text")
+            return comment[0]
```
betodealmeida, line 83: I'm a fan of doing:

```python
comment, other_thing = table_comment.get("text")  # pylint: disable=unused-variable
```

Even if you don't use `other_thing`, because it gives more information:

1. How many elements `table_comment.get()` returns.
2. A hint of what they are.

superset/db_engine_specs/db2.py
```diff
@@ -52,3 +61,49 @@ class Db2EngineSpec(BaseEngineSpec):
...
+        except IndexError:
```
betodealmeida, line 84: Ohhh, so the behavior depends on the library version?

betodealmeida (approved):
This is great, love the summary!

superset/db_engine_specs/db2.py
Vitor-Avila: In my tests it always returned a `Tuple()`, just added this as a precaution to avoid an issue in case we end up getting a `string`. I could remove this if we want to be more deterministic.

superset/db_engine_specs/db2.py
Vitor-Avila: @betodealmeida I think this would raise a `ValueError` because `text` is supposed to contain a tuple with a single element. Something like this:
``` python
{
  "text": ("This is my table comment",)
}
```

I tested adding multiple comments to the same table, and it always get replaced (never appended). Also seems accurate with [their documentation](https://www.ibm.com/docs/en/db2oc?topic=statements-comment). 

I just needed to override the `BaseEngineSpec` because it's [supposed to return a string and it doesn't account for the engine returning a tuple/array](https://github.com/apache/superset/blob/master/superset/db_engine_specs/base.py#L1309-L1319).

Vitor-Avila: that's really cool I didn't know that haha 😍

sadpandajoe:
🏷️ preset:2024.5

PR #26748, opened 2024-01-22
fix: `improve _extract_tables_from_sql`

betodealmeida:
### SUMMARY

Improve `_extract_tables_from_sql` from https://github.com/apache/superset/pull/26476 so that a proper exception is returned if the SQL fails to parse when extracting tables mentioned in a query. The previous behavior was to return an empty set of tables in the query.

N/A

### TESTING INSTRUCTIONS

Updated unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -28,6 +28,7 @@ from typing import Any, cast
 
 import sqlglot
 import sqlparse
+from flask_babel import gettext as __
 from sqlalchemy import and_
 from sqlglot import exp, parse, parse_one
 from sqlglot.dialects.dialect import Dialect, Dialects
@@ -58,7 +59,12 @@ from sqlparse.tokens import (
 )
 from sqlparse.utils import imt
 
-from superset.exceptions import QueryClauseValidationException, SupersetParseError
+from superset.errors import ErrorLevel, SupersetError, SupersetErrorType
+from superset.exceptions import (
+    QueryClauseValidationException,
+    SupersetParseError,
+    SupersetSecurityException,
+)
 from superset.utils.backports import StrEnum
 
 try:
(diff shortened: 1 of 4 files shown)


superset/sql_parse.py
```diff
@@ -287,9 +292,15 @@ def _extract_tables_from_sql(self) -> set[Table]:
...
-        except ParseError:
-            logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
-            return set()
+        except ParseError as ex:
+            dialect = self._dialect or "generic"
+            raise SupersetSecurityException(
+                SupersetError(
+                    error_type=SupersetErrorType.QUERY_SECURITY_ACCESS_ERROR,
+                    message=__(f"Unable to parse SQL ({dialect}): {self.sql}"),
+                    level=ErrorLevel.ERROR,
```
eschutho, line 301: @betodealmeida is this a user error or system error?

betodealmeida: It's hard to tell. It could be invalid SQL (user error), or it could be because limitations of the parser (system error).

eschutho: Ok, that makes sense. I'm thinking of a scenario where someone is making a lot invalid sql calls and it just keeps loading up a system monitoring queue with what looks to be an outage of a sort.

betodealmeida: Yeah, I'm worried about that too, and I think that's why originally the behavior was to not to fail. But the alternative is allowing people to run queries on tables they don't have access because we fail to extract the reference from the SQL.

john-bodley: @betodealmeida I _think_ having Superset being less lenient/blasé is likely a good thing, i.e., previously it blindly would could have* given a user access whenever it wasn't able to parse the SQL, irrespective of whether the user had access to the underly data.

The major challenge I see is if SQLGlot wrongfully isn't able to parse a SQL statement and thus we're wrongfully blocking a query. How do we help unblock users here?

\* This is dependent on how one configured their security manager.

john-bodley:
The TL;DR is  (per learnings during my internal postmortem writeup) `sqlparse` is a non-validating SQL parser whereas SQLGlot is a pseudo-validating* SQL parser. I'm not saying one is right or wrong, but sqlparse is definitely more lenient which could be perceived as beneficial for out specific use case, i.e., trying to identify the tables that a SQL query (be it valid or not) is referencing.

\* SQLGlot doesn't claim to validate SQL (per its [README.md](https://github.com/tobymao/sqlglot/blob/main/README.md)), 

> SQLGlot does not aim to be a SQL validator - it is designed to be very forgiving

it does seem to be somewhat opinionated (understandably so, i.e., if you're unable to successfully parse token(s) anything downstream in the AST is likely deemed questionable).

john-bodley (approved):

PR #26749, opened 2024-01-22
fix: prevent guest user from modifying metrics

betodealmeida:
### SUMMARY

Improve the security manager so that guest users can't change the metrics being requested from a chart; only the metrics stored with the chart can be requested.

N/A

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -1818,7 +1818,7 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
         return []
 
     def raise_for_access(
-        # pylint: disable=too-many-arguments,too-many-branches,too-many-locals
+        # pylint: disable=too-many-arguments,too-many-branches,too-many-locals,too-many-statements
         self,
         dashboard: Optional["Dashboard"] = None,
         database: Optional["Database"] = None,
@@ -1906,6 +1906,30 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             if denied:
                 raise SupersetSecurityException(
                     self.get_table_access_error_object(denied)
+                )
+
+        if self.is_guest_user() and query_context:
+            # Guest users MUST not modify the payload so it's requesting a different
+            # chart or different ad-hoc metrics from what's saved.
+            form_data = query_context.form_data
+            stored_chart = query_context.slice_
+
+            if (
+                form_data is None
+                or stored_chart is None
+                or form_data.get("slice_id") != stored_chart.id
+                or form_data.get("metrics", []) != stored_chart.params_dict["metrics"]
+                or any(
+                    query.metrics != stored_chart.params_dict["metrics"]
+                    for query in query_context.queries
+                )
+            ):
+                raise SupersetSecurityException(
+                    SupersetError(
+                        error_type=SupersetErrorType.DASHBOARD_SECURITY_ACCESS_ERROR,
+                        message=_("Guest user cannot modify chart payload"),
+                        level=ErrorLevel.ERROR,
+                    )
                 )
 
         if datasource or query_context or viz:
--- a/tests/unit_tests/security/manager_test.py
+++ b/tests/unit_tests/security/manager_test.py
@@ -18,6 +18,7 @@
 import pytest
 from pytest_mock import MockFixture
 
+from superset.common.query_object import QueryObject
 from superset.exceptions import SupersetSecurityException
 from superset.extensions import appbuilder
 from superset.security.manager import SupersetSecurityManager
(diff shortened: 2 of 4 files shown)


superset/security/manager.py
```diff
@@ -1908,6 +1908,30 @@ def raise_for_access(
...
+                or any(
+                    query.metrics != stored_chart.params_dict["metrics"]
+                    for query in query_context.queries
+                )
+            ):
+                raise SupersetSecurityException(
+                    SupersetError(
+                        error_type=SupersetErrorType.DASHBOARD_SECURITY_ACCESS_ERROR,
+                        message=_("Guest user cannot modify chart payload"),
+                        level=ErrorLevel.ERROR,
```
eschutho, line 1931: Similar question as here: https://github.com/apache/superset/pull/26748/files#diff-30f4c6ffdcb1f78a9e1ebbb60e1f297b379c181534d5a185a4cd37b1b16ac6f8R301
Does this raise a 5xx error? I think these are generally user-generated, right? We would want to know about them, but be able to distinguish them from a system error.

tests/unit_tests/security/manager_test.py
```diff
@@ -31,6 +32,81 @@ def test_security_manager(app_context: None) -> None:
...
+    # tampered requests
+    query_context.form_data = {
+        "slice_id": 43,
+        "metrics": stored_metrics,
+    }
+    query_context.queries = [QueryObject(metrics=stored_metrics)]  # type: ignore
+    with pytest.raises(SupersetSecurityException):
+        sm.raise_for_access(query_context=query_context)
+
+    tampered_metrics = [
```
michael-s-molina, line 81: @betodealmeida Not a blocker, but I think it's a good practice to separate different test cases in different tests. That way, it's easier to identify which part of a feature is failing. In this case, we could have a test for a legit request and another for a tampered request.

rusackas (approved):
Looks good to the best of my understanding. Ran the diff (did you know about [adding .diff to the URL](https://github.com/apache/superset/pull/26749.diff)? Handy!) through GPT conversations for review, too, and things sound all good!

superset/security/manager.py
Vitor-Avila: Could this logic impact queries to fetch filter values? Also, are we concerned with the dimensions as well (or un-aggregated columns for table charts)

tests/unit_tests/security/manager_test.py
betodealmeida: > @betodealmeida Not a blocker, but I think it's a good practice to separate different test cases in different tests. That way, it's easier to identify which part of a feature is failing. In this case, we could have a test for a legit request and another for a tampered request.

Yeah, good point.

superset/security/manager.py
betodealmeida: Ah, good point. Let me verify!

lindenh:
Hey, super glad this got prioritized. I've just pulled this change in on top of a fork of 3.0.2, so if there have been any changes in particular that would change the following please let me know, but we have seen two issues:
1. For big number charts, we primarily have been using saved metrics. When using them, we are seeing this error out:

February 06, 2024 at 10:53 (UTC-6:00) | File "/app/superset/security/manager.py", line 1927, in raise_for_access | 94b81be5bee24fc985dd3ce245f9cc46 | superset_app
-- | -- | -- | --
February 06, 2024 at 10:53 (UTC-6:00) | if form_data.get("metrics", []) != stored_chart.params_dict["metrics"]: | 94b81be5bee24fc985dd3ce245f9cc46 | superset_app
February 06, 2024 at 10:53 (UTC-6:00) | KeyError: 'metrics' | 94b81be5bee24fc985dd3ce245f9cc46 | superset_app
February 06, 2024 at 10:53 (UTC-6:00) | 2024-02-06 16:53:41,586:ERROR:superset.views.base:'metrics' | 94b81be5bee24fc985dd3ce245f9cc46 | superset_app
February 06, 2024 at 10:53 (UTC-6:00) | Traceback (most recent call last): | 94b81be5bee24fc985dd3ce245f9cc46 | superset_app
February 06, 2024 at 10:53 (UTC-6:00) | File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1823, in full_dispatch_request | 94b81be5bee24fc985dd3ce245f9cc46 | superset_app

Which tells me that we might want this line:
```
if form_data.get("metrics", []) != stored_chart.params_dict["metrics"]
```
to be 
```
if form_data.get("metrics", []) != stored_chart.params_dict.get(["metrics"], [])
```

We don't see this error when not using saved metrics. Changing to the .get() call also appears to fix the issue.

2. For some charts (it seems primarily/only mixed charts), this line is getting hit and denying access to charts even when the guest user hasn't actually changed anything:

```any(
                query.metrics != stored_chart.params_dict["metrics"]
                for query in query_context.queries
                )```
                
For some reason, it seems the metrics from stored_chart will only contain one of the two metrics from the mixed chart.

I'd like to test this locally and may if I get more time, but I've been bumping into an incredible amount of weird configuration errors. If I'm able to test this in other ways I'll update here.

PR #26751, opened 2024-01-23
chore: parallelize pytest

mistercrunch:
DRAFT! 

just trying to see if parallelizing pytest is going to shave some significant time. 

Locally at 4 threads I get:

```
real    0m31.041s
user    1m43.388s
sys     0m11.605s
```
at 8 I get
```
real    0m24.856s
user    2m27.023s
sys     0m20.348s
```
16 I get
```
real    0m37.087s
user    4m24.692s
sys     0m57.609s
```

That's around 30 seconds, as opposed to this baselien without parallel testing ->
```
real    1m5.498s
user    1m3.328s
sys     0m4.222s
```


mistercrunch:
doesn't seem worth it just yet

PR #26765, opened 2024-01-23
perf(export): export generates unnecessary files content

Always-prog:
### SUMMARY
Currently, exporting models functionality is slower than it can be.

Model export is slow because models export their related models, and since related models cannot check if they have already been exported to the final archive, they are exported anyway. This leads to generating unnecessary export model files.
For example, exporting a dashboard with 2 charts will generate 4 files - a dashboard, two charts, a dataset, and a database, but ““behind the scenes” there will actually be 7 files generated, since each chart will re-export the dataset and database, and at the end, the highest-level export model (dashboard export) will remove these duplicates.

To solve this problem, it is proposed to start generating export files not immediately, but only when they are written to a file, by convert `file_content` to function that generates content when we call it.

pay attention to speed up of generating an dashboard export archive!

## Before

https://github.com/apache/superset/assets/66589759/76551088-5d87-4c4d-b89d-fcc4c66f1b27

## After 

https://github.com/apache/superset/assets/66589759/4489b671-6c17-4f9a-a955-e5b285f5e5c4

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


geido:
Hello @Always-prog thanks for the PR! Would mind taking a look at the failing CI steps?

superset/commands/database/export.py
```diff
@@ -100,9 +100,18 @@ def _export(
...
-        yield file_path, file_content
+        return file_content
+
+    @staticmethod
+    def _export(
+        model: Database, export_related: bool = True
+    ) -> Iterator[tuple[str, Callable[[], str]]]:
+        yield ExportDatabasesCommand._file_name(
+            model
+        ), lambda: ExportDatabasesCommand._file_content(model)
```
betodealmeida, line 111: Maybe use `functools.partial(ExportDatabasesCommand._file_content, model)` here for consistency?

superset/commands/dataset/export.py
```diff
@@ -110,4 +122,4 @@ def _export(
             payload["version"] = EXPORT_VERSION
 
             file_content = yaml.safe_dump(payload, sort_keys=False)
-            yield file_path, file_content
+            yield file_path, lambda: file_content
```
betodealmeida, line 125: Why not:

```python
yield file_path, lambda: yaml.safe_dump(payload, sort_keys=False)
```

?

superset/commands/database/export.py
Always-prog: It is not required, because here we are making lambda function not in the loop.
In places where I use functools, I use it because of https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/cell-var-from-loop.html.

superset/commands/dataset/export.py
Always-prog: Really. Fixed :+1:

Always-prog:
@betodealmeida Hi! Thanks for review! I have committed fixes by your comments.

But the only thing that pre-commit fails due to file .github/workflows/update-monorepo-lockfiles.yml, which I have not fixed.

geido:
Hi @Always-prog it seems you have a conflicting file. Can you fix it, please? I am adding myself as a reviewer and checking back on this asap. Thank you!

geido:
👀

Always-prog:
@geido Hi! Rebased!

michael-s-molina (approved):
I tested the PR and was able to successfully export/import dashboards.

Thank you for the work @Always-prog, the performance improvements are impressive!

Always-prog:
@michael-s-molina Thank you for testing my PR. Can I get merge?

PR #26766, opened 2024-01-23
chore: prevent CI double runs on push + pull_request

mistercrunch:
### SUMMARY
I noticed recently that we get many "double-triggers" of CI actions on some PRs, most notable on PRs opened from the main apache repo, triggering against the master branch (see screenshot bellow). This is typically one for the `push` action and one for the `pull_request` action, running effectively the exact same CI script twice. Note that I think the double triggers also on forks (I'd have to double check), but just doesn't show under the PR list of required CI steps, and is probably just associated with the fork of the user.

Anyhow, so I went through all the githbub workflows and applied more restrictions. In most cases we apply the main pattern:
```yaml
on:
  push:
    branches:
      - 'master'
  pull_request:
```
This will trigger for each change on the pull request, and for every commit as it merges on the `master` branch.

In some other cases, I added a `paths` statement where I knew we should only run a workflow given a certain obvious folder, like frontend tests should only run if we touch `superset-frontend/**` and such.

Screenshot 2024-01-23 at 10 32 13 AM

### DIFF
--- a/.github/workflows/prefer-typescript.yml
+++ b/.github/workflows/prefer-typescript.yml
@@ -2,9 +2,10 @@ name: Prefer TypeScript
 
 on:
   push:
-    branches-ignore:
-      - "dependabot/**"
+    branches:
+      - 'master'
   pull_request:
+    types: [synchronize, opened, reopened, ready_for_review]
 
 jobs:
   prefer_typescript:
--- a/.github/workflows/prefer-typescript.yml
+++ b/.github/workflows/prefer-typescript.yml
@@ -10,7 +10,6 @@ on:
     types: [synchronize, opened, reopened, ready_for_review]
     paths:
       - "superset-frontend/src/**"
-
 jobs:
   prefer_typescript:
     if: github.ref == 'ref/heads/master' && github.event_name == 'pull_request'
--- a/.github/workflows/prefer-typescript.yml
+++ b/.github/workflows/prefer-typescript.yml
@@ -4,8 +4,12 @@ on:
   push:
     branches:
       - 'master'
+    paths:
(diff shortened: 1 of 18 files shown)


.github/workflows/prefer-typescript.yml
```diff
@@ -2,9 +2,10 @@ name: Prefer TypeScript
 
 on:
   push:
-    branches-ignore:
-      - "dependabot/**"
+    branches:
+      - 'master'
   pull_request:
+    types: [synchronize, opened, reopened, ready_for_review]
```
betodealmeida, line 8: Should we add a path here, eg:

```suggestion
    paths:
      - "superset-frontend/src/**"
    types: [synchronize, opened, reopened, ready_for_review]
```

betodealmeida (approved):
sagan-awesome

mistercrunch:
I ended up squeezing in more cache optimizations for docker in there as I was testing all sorts of things to speed up the builds:
- I got the docker buildx `--cache-from` and `--cache-to` to actually work with remote registry, no exactly sure why it didn't work before, something about using a different repo helped here
- disabled some slow builds in the matrix for py39/arm combo, where there's some wheels to build from scratch and QEMU (the arm emulator used here) is super slow, we can bring them back if needed, now that we cache things it should be fast unless a python dependency changes
- I did disable some tests, so that say frontend unit tests only run if the frontend changes, or some python checks (tests, lint, ..) only runs if the python package changes, ... but those tests were marked as required, which GitHub is a bit stupid about. If you try to skip a required test, the PR hangs forever. So I found a workaround that I documented around no-op.yml workflow for those edge cases. There were other workarounds, but this one seemed best

PR #26767, opened 2024-01-23
chore: improve SQL parsing

betodealmeida:
### SUMMARY

This is the first PR in a series to improve our SQL parsing, in effort to clean up the code base and make it more secure.

Currently, we have a class called `superset.sql_parse.ParsedQuery` that is used as an interface every time we need to parse or manipulate SQL. There are a few problems with the status quo:

1. Even though the class is called `ParsedQuery`, it's designed to accept a single SQL **statement**, not necessarily a query. For example, it has a `set_or_update_query_limit` method that works only on the first statement, if there are multiple.
2. Even though it's designed to accept a single statement, it doesn't raise an error when instantiated with a query that has multiple statements, so in several places the class is used with a full query.
3. There are multiple places that import `sqlparse` directly to manipulate SQL. For example, when inserting RLS into a query, or when modifying a query to insert a `LIMIT` or `TOP`.
4. In many places we manipulate SQL inefficiently, serializing and deserializing the SQL multiple times.

Ideally we'd have a single abstraction that handles SQL parsing. Superset code should never parse code directly by importing a 3rd party library like `sqlparse`, `sqloxide`, or `sqlglot`; instead, it should only call the abstraction. Similarly, the abstraction should wrap library specific exceptions (like `sqlglot.errors.ParseError`) with its own.

This PR introduces two new classes that aim to be those abstractions: `SQLScript` and `SQLStatement`. The `SQLStatement` strictly represents a single SQL statement. These classes are instantiated with SQL and an optional engine (a `BaseEngineSpec.engine` attribute), and should provide all the methods needed for manipulating and introspecting SQL queries and statements. The initial implementation uses `sqlglot`, but the clean interface allows the parser to be changed in the future if we want.

My plan for improving the SQL parsing is first to get rid of code outside of `sql_parse.py` that using `sqlparse` directly:

- [x] Introduce `SQLScript` and `SQLStatement`, and replace any simple calls to `sqlparse` with calls to the new classes (this PR).
- [ ] Implement a custom `SQLStatement` class for `Kusto`, since [it doesn't use conventional SQL](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/).
- [ ] Implement `insert_rls_in_predicate` inside `SQLStatement`.
- [ ] Implement `get_cte_query` inside `SQLStatement`.
- [ ] Implement `apply_top_to_sql` in `SQLStatement`.

At this point we should have `sqlparse` being used only inside `sql_parse.py` by `ParsedQuery`. The last steps would be:

- [ ] Replace all calls to `ParsedQuery` with calls to the new classes.
- [ ] Get rid of the `sqlparse` dependency.

Note that despite the big changes, no public interfaces will change. Also, we've been gradually changing the parser due to security issues, having introduced `sqlglot` as a dependency in https://github.com/apache/superset/pull/26476. Because of this, I thought it wouldn't be necessary to write a SIP. I'm happy to write one if people think it's necessary.

N/A

### TESTING INSTRUCTIONS

Updated unit tests, with mostly cosmetic changes because the SQL pretty-printing has changed.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/SqlLab/actions/sqlLab.js
```diff
@@ -912,6 +912,7 @@ export function formatQuery(queryEditor) {
     const { sql } = getUpToDateQuery(getState(), queryEditor);
     return SupersetClient.post({
       endpoint: `/api/v1/sqllab/format_sql/`,
+      // TODO (betodealmeida): pass engine as a parameter for better formatting
```
betodealmeida, line 915: This is a nice benefit of this change: we can now have DB-specific formatting.

superset/connectors/sqla/models.py
```diff
@@ -1414,26 +1415,6 @@ def mutate_query_from_config(self, sql: str) -> str:
     def get_template_processor(self, **kwargs: Any) -> BaseTemplateProcessor:
         return get_template_processor(table=self, database=self.database, **kwargs)
 
-    def get_query_str_extended(
```
betodealmeida, line 1417: This method is **identical** to the base class.

superset/connectors/sqla/models.py
```diff
@@ -1475,33 +1456,6 @@ def get_from_clause(
 
         return from_clause, cte
 
-    def get_rendered_sql(
```
betodealmeida, line 1478: This method is **identical** to the base class.

superset/db_engine_specs/postgres.py
```diff
@@ -281,8 +281,9 @@ def get_default_schema_for_query(
         This method simply uses the parent method after checking that there are no
         malicious path setting in the query.
         """
-        sql = sqlparse.format(query.sql, strip_comments=True)
-        if re.search(r"set\s+search_path\s*=", sql, re.IGNORECASE):
+        statement = SQLStatement(query.sql)
+        settings = statement.get_settings()
+        if "search_path" in settings:
```
betodealmeida, line 286: Look, ma! No regexes! 😄

superset/sql_parse.py
```diff
@@ -252,6 +253,163 @@ def __eq__(self, __o: object) -> bool:
         return str(self) == str(__o)
 
 
+def extract_tables_from_statement(
```
betodealmeida, line 256: This function was originally a method in `ParsedQuery`.

superset/sql_parse.py
```diff
@@ -252,6 +253,163 @@ def __eq__(self, __o: object) -> bool:
...
+        Table(
+            source.name,
+            source.db if source.db != "" else None,
+            source.catalog if source.catalog != "" else None,
+        )
+        for source in sources
+    }
+
+
+def is_cte(source: exp.Table, scope: Scope) -> bool:
```
betodealmeida, line 304: This function was originally a method in `ParsedQuery`.

mistercrunch (commented):
Few comments:
- Let's remove `sqlparse` from setup.py and requirements. I noticed `pip-compile-multi` was SUPER slow last time I tried it
- about the description and now improved support for multi-statements, are all the cases covered by new tests in `sql_parse_tests.py`?

villebro:
I think this is a welcome change and a move in the right direction, but I feel a big change like this should probably be discussed in the form of a SIP first. I anticipate the final implementation to end up looking similar to this PR, but it may be beneficial to discuss the change first before jumping directly into the code.

betodealmeida:
> I think this is a welcome change and a move in the right direction, but I feel a big change like this should probably be discussed in the form of a SIP first. I anticipate the final implementation to end up looking similar to this PR, but it may be beneficial to discuss the change first before jumping directly into the code.

SIP here: https://github.com/apache/superset/issues/26786

villebro:
Thanks @betodealmeida !

betodealmeida:
https://github.com/apache/superset/issues/26786 approved!

superset/sql_parse.py
```diff
@@ -252,6 +253,182 @@ def __eq__(self, __o: object) -> bool:
...
+
+        """
+        settings: dict[str, str] = {}
+        for statement in self.statements:
+            settings.update(statement.get_settings())
+
+        return settings
+
+
+class SQLStatement:
```
mistercrunch, line 366: Looking at the class design in this module, this looks like it really extract the first statement from a set of statement. It may not be for this PR, but ideally the classes become a better representation of the real word. Maybe there's a notion of a `SqlScript` or `SqlStatementCollection`, that's an array of other SQL objects like `SqlQuery`, `SqlDMLCommand`, ...

Maybe there's benefit in having inheritance defined there where a SqlQuery and DmlCommand are both derivative or `SqlStatement`

tests/integration_tests/core_tests.py
```diff
@@ -537,7 +537,7 @@ def test_mssql_engine_spec_pymssql(self):
         )
 
     def test_comments_in_sqlatable_query(self):
-        clean_query = "SELECT '/* val 1 */' as c1, '-- val 2' as c2 FROM tbl"
+        clean_query = "SELECT\n  '/* val 1 */' AS c1,\n  '-- val 2' AS c2\nFROM tbl"
```
mistercrunch, line 540: [suggesting] it could be good to have some sort of reusable `compare_sql(strict=False, case_sensitive=False, disregard_schema_prefix=True)` function that could be reused in unit tests. Maybe it's a method of `ParseQuery` (`is_identical` or `is_similar`)

mistercrunch (commented):
Left some comments. At a high level the main thing for me would be to clarify the object model and inheritance scheme for the SQL-related utilities. At a high level thinking about this, it sits close to db_engine_spec or at least relate to it in some ways.

superset/sql_parse.py
betodealmeida: Not just the first. The hierarchy is that a `SQLQuery` has 0+ `SQLStatement`s. When you pass a SQL query to `SQLQuery` it will split the query into statements, and instantiate one `SQLStatement` for each one. The `SQLStatement._parse_statement` validates that the SQL passed corresponds to exactly one statement, which is why it looks like it's extracting the first one.

This is done because today in the codebase there's an implicit assumption that a `ParsedQuery` represents a single statement, but because this is not enforced there are many places in the code that pass a full query to it when using methods that really don't care.

superset/sql_parse.py
```diff
@@ -22,12 +22,13 @@
 import urllib.parse
 from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
-from typing import Any, cast, Optional
+from typing import Any, cast, Optional, Union
```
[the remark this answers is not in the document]
betodealmeida: Ah, I like this!

@john-bodley, you did some work on reorganizing the codebase, what are your thought here?

superset/sql_parse.py
```diff
@@ -252,6 +253,182 @@ def __eq__(self, __o: object) -> bool:
...
+        return statements[0]
+
+    def format(self, comments: bool = True) -> str:
+        """
+        Pretty-format the SQL statement.
+        """
+        write = Dialect.get_or_raise(self._dialect)
+        return write.generate(self._parsed, copy=False, comments=comments, pretty=True)
+
+    def get_settings(self) -> dict[str, str]:
```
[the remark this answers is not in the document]
betodealmeida: This method is meant for statements that set parameters using the `SET foo = 'bar'` syntax, which seems to be more universal than `USE ...`. We have a few DB engine specs that need to know if `search_path` was set to something different for security reasons, which is why I added this method. And while today it's used just to check for the `search_path,` I used a more generic name because in the future it could be leveraged in different ways.

I agree that there's functionality here that is DB-specific, but I'm hoping to address that by always requiring the dialect to be specified together with the SQL when instantiating the `SQLStatement`, so we can leave the heavy work to `sqlglot`. If we need custom functionality we can always subclass the `SQLStatement` for specific dialects, though hopefully we'd be able to avoid that by implementing the dialect upstream in `sqlglot` directly.

The main reason for having DB-specific handling here and not in the DB engine specs is that I'd like for all the SQL parsing to be done only by these new classes, so we can have a cleaner interface. So I actually want to remove the custom DB engine spec methods that are related to SQL parsing that exist today, since the DB engine specs already do too much.

tests/integration_tests/core_tests.py
betodealmeida: Yeah, good point, I'll add that!

superset/sql_parse.py
eschutho: I had a similar question, reading `SQLQuery` as solely a data retrieval query or a type of statement, rather than a collection of statements. `SQLInstruction` as another option?

superset/sql_parse.py
```diff
@@ -252,6 +253,182 @@ def __eq__(self, __o: object) -> bool:
...
+    ) -> exp.Expression:
+        """
+        Parse a single SQL statement.
+        """
+        statements = [
+            statement
+            for statement in sqlglot.parse(sql_statement, dialect=dialect)
+            if statement
+        ]
+        if len(statements) != 1:
```
mistercrunch, line 404: Super minor, but here you could fail faster (assuming that sqlglot.parse is a generator), meaning you could:
```python
has_statement = False
for statement in sqlglot.parse(sql_statement, dialect=dialect):
    if has_statement and statement:
        raise{...}
    else:
        has_statement = True
 ```

File: superset/sql_parse.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (mistercrunch) [reply superset/sql_parse.py:416]: Oh sorry I got confused on my previous comment, I read the `SET` as a `USE` and commented based on that. I guess they both have something in comment as they are context-related statements.

File: superset/sql_parse.py (hunk 11 above)
Comment (betodealmeida) [reply superset/sql_parse.py:404]: Ah, good idea. Unfortunately I tested and it doesn't seem to be the case:

```python
import time

import sqlglot

padding = "a" * 100000000
start = time.time()
for stmt in sqlglot.parse(f"SELECT 'small'; SELECT 'la{padding}rge';"):
    print(time.time() - start)
```

```bash
$ python test-sqlglot.py
22.041950941085815
22.04202389717102
$
```

rafehi:
Where are we at with this work? There's some additional changes that I'd be willing to take on (https://github.com/apache/superset/pull/19572) but looks like this is a dependency. 

We've made and validated significant performance improvements in our Superset fork, but makes sense to consolidate those into SQLScript and SQLStatement rather than directly replace sqlparse calls with sqlglot.

mistercrunch (approved):
LGTM

eschutho (approved):
I reviewed just the cypress test portion as a code-owner requirement. That LGTM. 👍

mistercrunch:
Hey @betodealmeida , this PR might have broken a Hive-specific test -> https://github.com/apache/superset/actions/runs/8272398041/job/22634142747#step:11:608

It appears to fail after this PR was merged (I noticed since it was failing on some other unrelated PR) and trace it back here. Note that this particular GHA test for Hive runs only when `superset/**` is altered by a PR, that explains why other builds succeeded on `master` since then.

Hive seems to be complaining about a `GROUP BY 1` but from memory it totally supports that. Unclear what's up with the test but I thought you might know more given this PR.

PR #26787, opened 2024-01-24
fix(docker): improve docker tags to be cleared and avoid conflicts

mistercrunch:
Improving the docker tag naming scheme to:
- we now suffix most tags with `-arm` for platofrm = `linux-arm64`, and leave the amd builds without suffix since they're used already, before this fix we would have conflicts between tags / images. More on this here https://github.com/apache/superset/issues/26788
- we now point the main `dev` tag (ie `apache/supserset:dev`, and `apache/supserset:dev-arm`) to top of the `master` branch. This tag only gets generated then. 
- on release, this should create the right top-level docker tags like `apache/supserset:lean` and `apache/supserset:py310`, ...
- on release, the very top level for the whole repo is the lean/amd image

# A question remains
Where should we point top level builds like `apache/supserset:lean` and `apache/supserset:lean-arm` or even the top-most level `apache/supserset`? Seems it should be pointed to the latest official ASF release version/build, unclear if it should be done in CI, or a simple `docker tag` command as part of the ASF release process, pointing to the image of the right SHA (that tag should be computed and exist, so all we need is to point to the right place)


eschutho:
> unclear if it should be done in CI, or a simple `docker tag` command as part of the ASF release process

@mistercrunch for the ASF release, the docker image creation is still handled through the CI [when the sha is marked for release](https://github.com/apache/superset/blob/master/.github/workflows/docker-release.yml#L5), so there's currently no separate or manual process that's done.

sfirke:
Note that when we talk about the `dev` image, its main difference from `lean` seems to be that it bundles the headless browser and driver needed for Alerts & Reports to work.  I would like eventually to see those bundled with the main images if possible, it confuses people that they have to run an image called "dev" to get that functionality.

I'm only bringing that up here to ask that we retain functionality such that when a version is released tagged like `3.1.0`, a corresponding version `3.1.0-dev` is also built that contains the tools necessary for Alerts & Reports.

craig-rueda (approved):
LGTM

PR #26789, opened 2024-01-24
chore(RESOURCES): fix markdown for table formatting

qleroy:
Just fixing markdown to restore the preview !


john-bodley:
Thanks @qleroy. I believe the table is auto-formatted via IDEs, which typically makes markdown tables more readable. What do you mean by preview?

qleroy:
I just removed the last bit `|---|` on line 2
which was added by mistake some time ago.
It breaks the formatting and you cannot preview it as a table, for example on GitHub.

john-bodley:
@qleroy would you mind running `tox -e pre-commit` locally? I _believe_ that `prettier` will reformat the table so that the columns are aligned. 

cc @michael-s-molina

qleroy:
Will try.
Sorry for the confusion but the problem is not so the formatting but the additional column `|---|` only on line 2 that breaks the whole table

qleroy:
Cc @sfirke

sfirke (approved):
LGTM!  I like how the "before" is all messed up and many screens long and then the "after" is a table again 😁

sfirke:
@john-bodley here is how this .md page is rendering currently:
image

thanks @qleroy for the fix!

sfirke:
Just confirmed that it's rendering correctly again:

image

michael-s-molina:
@sfirke We should commit using the same output generated by pre-commit. If it's generating this weird preview, we need to fix the formatter. The problem of committing a manually generated format is that every time this file is touched and formatted by pre-commit, we'll get a diff.

john-bodley:
@sfirke my comment was related to how the table presented pre-rendering. There's auto-formatters which add whitespace to ensure that the columns align which help when reading the table in an IDE.

I realize that the fix addressed an issue with the markdown structure, but it augmented the formatting and thus whitespace is no longer preserved.

sfirke:
Oops, sorry this wasn't the proper structural fix. I don't completely follow the correct way to address this but would like to learn, maybe by looking at a PR that does it?

I am unfamiliar with maintaining a table like this, I had just noticed it was broken and put out a call in Slack asking for a fix from the community.

qleroy:
I just checked that the pre-commit hooks pass and pre-commit outputs the same file.
I my fix-md-formatting branch, the first commit that removes `|---|` is necessary, the second way removing whitespace is not necessary and pre-commit ignores the difference as well
Screenshot 2024-01-25 at 21 38 10


```
tox -e pre-commit
pre-commit: commands[0]> pre-commit run --all-files
auto-walrus..............................................................Passed
pyupgrade................................................................Passed
pycln....................................................................Passed
isort....................................................................Passed
mypy.....................................................................Passed
pip-compile-multi verify.................................................Passed
check docstring is first.................................................Passed
check for added large files..............................................Passed
check yaml...............................................................Passed
debug statements (python)................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
black....................................................................Passed
prettier.................................................................Passed
Blacklist................................................................Passed
```

PR #26790, opened 2024-01-24
chore: refactor the github actions has-secrets logic

mistercrunch:
While working on https://github.com/apache/superset/pull/26772, I realized that the has-secret logic was broken for unclear reasons. Now after doing this fix, I looked and realized that there's similar logic across about a dozen other gh actions, and thought it'd be a good thing to refactor/fix this with a reusable action.

Now many of these workflows are set to trigger on push-on-master only meaning it's less of an issue since a false positive on has-secret dpesn't matter in that context since there's always a secret.

I still thought I should refactor this to something we can trust and build upon.

The solution introduces this new simple reusable action.

One minor note is in cases where we need multiple secrets, as in say DOCKERHUB_TOKEN and DOCKERHUB_USER, we simply look at one and assume that's clear-enough of an indicator.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -64,6 +64,7 @@ github:
           - cypress-matrix (1, chrome)
           - cypress-matrix (2, chrome)
           - cypress-matrix (3, chrome)
+          - docker-build
           - frontend-build
           - pre-commit (3.9)
           - python-lint (3.9)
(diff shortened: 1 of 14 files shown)


mistercrunch:
closing/reopening from a branch on the main fork to test further

Issue #1918, opened 2017-01-06
Field "Created by" is html and it's improperly rendered


0x0ece:
### Superset version
0.15.1 and master

### Expected results
values in "Changed by" should be links

### Actual results
values in "Changed by" are texts e.g. '<a href="/superset/profile/emanuele/">Emanuele Cesena</a>'

### Steps to reproduce
Got to the url /druiddatasourcemodelview/list/
(note: this works fine with the Creator, e.g. in the list of dashboards)

xrmx:
This looks like an easy patch, care to open a pull request?

0x0ece:
I'd love to, but I spent quite some time and didn't really figured out the whole path from the header definition to the react rendering.
 
Maybe @ascott can chime in and give me a hint?

ascott:
looks like a Flask App Builder template issue. i think the fix may be to add the `safe` filter here where the column value is rendered: 
https://github.com/ascott/superset/blob/master/superset/templates/superset/fab_overrides/list_with_checkboxes.html#L69

http://stackoverflow.com/questions/12672469/how-to-render-html-content-with-jinja-using-flask#12673984

PR #1858, opened 2016-12-16
[sql lab] only show single run query button

ascott:
- show single run button rather than `Run Query` and `Run Async Query`
- if async is possible, run async, otherwise run sync
- move buttons to their own component

fixes this issue: https://github.com/airbnb/superset/issues/1819

**before**
screenshot 2016-12-16 13 40 52
screenshot 2016-12-16 13 41 03

**after**
screenshot 2016-12-16 13 28 25
screenshot 2016-12-16 13 27 54
run-selected

plz review @mistercrunch @vera-liu @bkyryliuk

PR #26794, opened 2024-01-24
chore(dependencies): bumping jinja2

rusackas:
### SUMMARY

Just another opportunistic bump. Fixes include:

Fix compiler error when checking if required blocks in parent templates are empty. [#1858](https://github.com/pallets/jinja/pull/1858)

xmlattr filter does not allow keys with spaces. GHSA-h5c8-rqwp-cp95

Make error messages stemming from invalid nesting of {% trans %} blocks more helpful. [#1918](https://github.com/pallets/jinja/pull/1918)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -90,8 +90,6 @@ dnspython==2.1.0
     # via email-validator
 email-validator==1.1.3
     # via flask-appbuilder
-exceptiongroup==1.2.0
-    # via cattrs
 flask==2.2.5
     # via
     #   apache-superset
@@ -161,10 +159,7 @@ idna==3.2
     #   email-validator
     #   requests
 importlib-metadata==6.6.0
-    # via
-    #   apache-superset
-    #   flask
-    #   shillelagh
+    # via apache-superset
 importlib-resources==5.12.0
     # via limits
 isodate==0.6.0
@@ -355,9 +350,7 @@ tabulate==0.8.9
 typing-extensions==4.4.0
     # via
     #   apache-superset
-    #   cattrs
     #   flask-limiter
-    #   kombu
     #   limits
     #   shillelagh
 tzdata==2023.3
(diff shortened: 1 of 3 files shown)


mistercrunch (approved):

mistercrunch (changes requested):
Oh, I approved earlier but changed my mind after seeing now that you skip the what appears to be dysfunctional `pip-compile-multi` step (super slow, maybe never-ending?), wondering if that could lead to a future regression when re-generating. As far as I'm guessing, pip-compile does not look into the .txt file for input. I'd make sure to add the right rule in the corresponding `.in` file to prevent regression. Also that tool does do some hash checking and may fail for the next person...

villebro (approved):

rusackas:
@mistercrunch @villebro am I doing this right?

Ran pip-compile-multi --no-upgrade, and committed everything except where it replaced `.` with my local path to Superset.

michael-s-molina:
I wonder if we should run `pip-compile-multi --no-upgrade` on CI when dependency files change, and fail if there are differences between the output and what's being committed.

rusackas:
> I wonder if we should run `pip-compile-multi --no-upgrade` on CI when dependency files change, and fail if there are differences between the output and what's being committed.

Yep... we should probably have one that does this for all the "this shouldn't change" jobs — translations, npm, pip, maps, etc, all alongside the pre-commit check.

mistercrunch (approved):

mistercrunch:
LGTM apart from the conflicts. Am I the only for which this takes forever? Any `--verbose` flag or other flags to accelerate this. It's good to know about the `-P` thought, hopefully it scopes down the work and make execution faster.

PR #26795, opened 2024-01-25
chore: refactor the github actions has-secrets logic

mistercrunch:
[reopening] While working on https://github.com/apache/superset/pull/26772, I realized that the has-secret logic was broken for unclear reasons. Now after doing this fix, I looked and realized that there's similar logic across about a dozen other gh actions, and thought it'd be a good thing to refactor/fix this with a reusable action.

Now many of these workflows are set to trigger on push-on-master only meaning it's less of an issue since a false positive on has-secret dpesn't matter in that context since there's always a secret.

I still thought I should refactor this to something we can trust and build upon.

The solution introduces this new simple reusable action.

One minor note is in cases where we need multiple secrets, as in say DOCKERHUB_TOKEN and DOCKERHUB_USER, we simply look at one and assume that's clear-enough of an indicator.

### DIFF
--- a/.github/workflows/check-secrets.yml
+++ b/.github/workflows/check-secrets.yml
@@ -9,6 +9,7 @@ on:
     outputs:
       has-secrets:
         type: string
+        value: ''
 
 jobs:
   check-secrets:
--- a/.github/workflows/check-secrets.yml
+++ b/.github/workflows/check-secrets.yml
@@ -8,7 +8,6 @@ on:
         type: string
     outputs:
       has-secrets:
-        type: string
         value: ''
 
 jobs:
--- a/.github/workflows/check-secrets.yml
+++ b/.github/workflows/check-secrets.yml
@@ -17,6 +17,8 @@ jobs:
         run: |
           if [[ -n "${{ secrets[inputs.secret_name] }}" ]]; then
             echo "has-secrets=true" >> "$GITHUB_ENV"
+            echo "DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}"
           else
             echo "has-secrets=false" >> "$GITHUB_ENV"
+            echo "DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}"
           fi
--- a/.github/workflows/check-secrets.yml
+++ b/.github/workflows/check-secrets.yml
@@ -6,6 +6,9 @@ on:
       secret_name:
         required: true
         type: string
+    outputs:
+      has-secrets:
+        type: string
 
 jobs:
   check-secrets:
--- a/.github/workflows/check-secrets.yml
+++ b/.github/workflows/check-secrets.yml
@@ -8,19 +8,19 @@ on:
         type: string
     outputs:
       has-secrets:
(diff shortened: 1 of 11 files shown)


rusackas:
> One minor note is in cases where we need multiple secrets, as in say DOCKERHUB_TOKEN and DOCKERHUB_USER, we simply look at one and assume that's clear-enough of an indicator.

This part worries me a bit. Each key is set separately by Apache Infra, and sometimes they are (or soon will be) secrets for disparate systems (e.g. Google, Github, OpenAI, etc). I don't think it's safe to assume that if one secret exists that the others are all good... it would be best to pass an array of secrets, and have them all validated by the shared action.

rusackas:
Just to make sure I'm understanding, is this fixing the issue wherein the `docker-build` is currently being skipped (seemingly due to https://github.com/apache/superset/pull/26772)? If so, I'm down to revert that one if this one needs more TLC.

.github/workflows/docker.yml
```diff
@@ -64,8 +50,8 @@ jobs:
           ./scripts/docker_build_push.sh "" ${{ matrix.target }} ${{ matrix.platform }}
 
   ephemeral-docker-build:
-    needs: config
-    if: needs.config.outputs.has-secrets
+    needs: check-secrets
+    if: needs.check-secrets.outputs.has-secrets == 'true'
```
dpgaspar, line 54: Not totally related with this PR, but I think it only makes sense to build and push to a local file on the workspace workflow while the PR is open, so this means we are supposed to not have secrets here. the `ephemeral-env` workflow is expecting the build dir

.github/workflows/docker.yml
```diff
@@ -8,27 +8,13 @@ on:
...
-          fi
+  check-secrets:
+    uses: ./.github/workflows/check-secrets.yml
+    with:
+      secret_name: 'DOCKERHUB_USER'
   docker-build:
-    needs: config
-    if: needs.config.outputs.has-secrets == 'true'
+    needs: check-secrets
+    if: needs.check-secrets.outputs.has-secrets == 'true'
```
dpgaspar, line 17: should we always build images on PRs? like we were doing before? `docker_build_push.sh` will only push if `DOCKERHUB_USER` is present

mistercrunch: Yes that's right, more work to do here

mistercrunch: About both your comments, I think I figured it out in https://github.com/apache/superset/pull/26801, where we always run docker, but in forks it runs without authentication, which means it builds an DOESNT push, just making sure that docker build works.

Now I was also able to improve caching there, where in forks, it's use the registry cache as input, but it cannot push the cache and won't. Unclear what the cache hit rate will be, but should be pretty good unless the PR touches dependencies. Also if you don't touch the frontend there's a lot savings on docker build time. 

As docker cache can be really good if done well, it could be tempting to eventually run tests in-docker, and saving a lot on build times.

mistercrunch:
Frustrated head-butting with GitHub reusable actions not working as expected. Tried a million things and couldn't make the output value work.  Closing for now

PR #26798, opened 2024-01-25
chore: Add permission to view and drill on Dashboard context

geido:
### SUMMARY
Introducing the `can_view_and_drill` permission to enable drilling and read-only related chart interactions, such as "View as table" in Dashboard.

This is particularly useful in contexts like embedded dashboards where one would not want to give the `can_explore` permission but still allow certain charts' interactions within the dashboard context.

N.A.

### TESTING INSTRUCTIONS
All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


dpgaspar (commented):
Don't we need to create a permission on the backend?

superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
```diff
@@ -108,12 +111,12 @@ const ChartContextMenu = (
...
   const showDrillToDetail =
     isFeatureEnabled(FeatureFlag.DRILL_TO_DETAIL) &&
-    canExplore &&
+    (canExplore || canViewDrill) &&
     isDisplayed(ContextMenuItem.DrillToDetail);
 
   const showDrillBy =
     isFeatureEnabled(FeatureFlag.DRILL_BY) &&
-    canExplore &&
+    (canExplore || canViewDrill) &&
```
kgabryje, line 119: In the drill by modal, there's an "Edit chart" button which redirects to Explore. I believe it should be disabled if `canViewDrill` is true but `canExplore` isn't (with some tooltip like "insufficient permissions")

geido: I think in general if can_explore is not found in the permissions, this button should be disabled independent of `can_view_drill_dashboard`

kgabryje: Yeah that's what I mean 😄 It wasn't there before because user couldn't open that modal anyways without `can_explore` permission

geido: Right makes sense. Thanks for spotting that!

geido:
It appears some reformatting of changelog files happened inadvertently. I'd keep

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -465,7 +470,7 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
       )}
 
       {isFeatureEnabled(FeatureFlag.DRILL_TO_DETAIL) &&
-        props.supersetCanExplore && (
+        (props.supersetCanExplore || canViewDrill) && (
```
kgabryje, line 473: Could we DRY it? We’re using the same `or` in 3 places

kgabryje (approved):
1 small nit, otherwise lgtm!

rohitpawar2811:
@mistercrunch @rusackas 
Hii
I'd like to point out that these MR changes are not available in the 4.0.0 release. I have checked both the git repo and inside the docker-image as well. However, according to the changelog, they should be included in 4.0.0. It is intentionally left because of some reason ?

eschutho:
Hi @rohitpawar2811 if you are looking specifically for the `can_view_and_drill` permission, it was removed in a [later commit](https://github.com/apache/superset/pull/27029).

agunoglu:
hi I have all roles. And can not see drill down menu in superset graphs. Why?

geido:
> hi I have all roles. And can not see drill down menu in superset graphs. Why?

Hello @agunoglu what roles do you currently have and what version are you on?

PR #26802, opened 2024-01-25
fix: docker should always run, even in forks (SIBLING/TEST/DRAFT PR)

mistercrunch:
sibling DRAFT PR for https://github.com/apache/superset/pull/26801, to test the same logic against a fork, make sure it behaves as expected.


dpgaspar:
@mistercrunch we cannot write to a centralized cache but just locally

PR #26814, opened 2024-01-26
fix(time-series table): Can't compare from the beginning of the time range

michael-s-molina:
### SUMMARY
When using the Time-series table chart, many times users want to compare the current value of a metric with the beginning of the time range. The problem is that the time lag is a fixed number of periods calculated from the last data point, and as time advances, users lose the initial reference, making it impossible to compare a value agains the beginning of the time range. This PR fixes this problem by allowing negative time lag values which will calculate the time period starting from the beginning of the time range, increasing the chart's flexibility. 

https://github.com/apache/superset/assets/70410625/dd5daaab-7573-4c5e-a94c-f8a330837d25

### TESTING INSTRUCTIONS
Check the video for instructions.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


geido (approved):
Nice!

PR #26847, opened 2024-01-29
feat: Custom server page length

aobapirak:
### SUMMARY

This contribution delivers usability enhancements to the Custom Server Page Length feature in Superset. This upgrade now enables custom input values to persist in the drop-down list. The previous version of this feature posed an inconvenience as any new custom value entered by the user would disappear from the drop-down list when the user moved to a different choice.

The new implementation, however, ensures any custom value entered by the user remains in the drop-down list, thereby enriching user accessibility for frequent inputs. Moreover, the system now auto-arranges the drop-down list every time a new value is included, facilitating easy and efficient navigation.

Additional modifications include introducing a cap on the maximum custom page length value. The cap will be equivalent to either the total rows of data results or a fixed 10k rows in scenarios where the total rows of data exceed 10,000. 

### BEFORE SCREENSHOTS OR ANIMATED GIF

https://github.com/apache/superset/assets/87610632/d6462d1f-1a74-4505-89f2-bf8f459147c5

### AFTER SCREENSHOTS OR ANIMATED GIF

https://github.com/apache/superset/assets/87610632/dbb1d509-5886-4452-a559-4bd465bb6e30

### TESTING INSTRUCTIONS

1. Navigate to the Charts page
2. Choose any dataset
3. Choose table chart
4. Check the Server Pagination checkbox
5. Enter a new value in the selection drop-down list
6. Verify that the new input is sticky in the drop-down list

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


aobapirak:
> Thanks @aobapirak for the PR. Would you mind adding some tests?

Yes, sure! I will let you know once I'm done.

aobapirak:
Ping @john-bodley, could you please review? I have added some tests.

john-bodley:
Thanks @aobapirak. Sorry I was out last week and thus wasn't able to respond. I've added a couple of frontend reviewers to your PR.

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -226,36 +226,61 @@ const transformProps = (
...
+  if (Number(formData.server_page_length) > 10000) {
+    formData.server_page_length = 10000;
+  }
+
+  if (
+    formData.server_pagination &&
+    !Number.isNaN(formData.server_page_length) &&
+    (formData.server_page_length || 0) <= data.length &&
+    !formData.server_page_length_options?.some(
+      (option: [number, string]) => option[0] == formData.server_page_length,
```
rusackas, line 253:
```suggestion
      (option: [number, string]) => option[0] === formData.server_page_length,
```

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -226,36 +226,61 @@ const transformProps = (
...
+  if (formData.server_pagination) {
+    [baseQuery, countQuery, totalQuery] = queriesData;
+    rowCount = (countQuery?.data?.[0]?.rowcount as number) ?? 0;
+  } else {
+    [baseQuery, totalQuery] = queriesData;
+    rowCount = baseQuery?.rowcount ?? 0;
+  }
+  const data = processDataRecords(baseQuery?.data, columns);
+
+  if (Number(formData.server_page_length) > 10000) {
```
geido, line 244: Can we move this to a constant for better visibility?

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -226,36 +226,61 @@ const transformProps = (
...
+    !Number.isNaN(formData.server_page_length) &&
+    (formData.server_page_length || 0) <= data.length &&
+    !formData.server_page_length_options?.some(
+      (option: [number, string]) => option[0] === formData.server_page_length,
+    )
+  ) {
+    formData.server_page_length_options?.push([
+      Number(formData.server_page_length),
+      formData.server_page_length === 0
+        ? 'All'
```
geido, line 259: I think this should be localized

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -226,36 +226,61 @@ const transformProps = (
...
+        : String(formData.server_page_length),
+    ]);
+    formData.server_page_length_options?.sort(
+      (a: any[], b: any[]) => a[0] - b[0],
```
geido, line 263: Is there a better type representation than `any`?

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -226,36 +226,61 @@ const transformProps = (
...
+  }
+
+  if (
+    formData.server_pagination &&
+    !Number.isNaN(formData.server_page_length) &&
+    (formData.server_page_length || 0) <= data.length &&
+    !formData.server_page_length_options?.some(
+      (option: [number, string]) => option[0] === formData.server_page_length,
+    )
+  ) {
```
geido, line 255: I think it would be better to improve readability of this logic by assigning some of these to constants

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
aobapirak: Solved, I created MAX_SERVER_PAGE_LENGTH in consts.ts

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
aobapirak: I'm not sure what do you mean but I'm using the t function provided by @superset-ui/core as `t('All')` correct me if I'm wrong

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
aobapirak: Modified

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
aobapirak: Modified

geido:
@aobapirak it seems you might need to update your branch with master in order for me to spin up a test environment for this PR. Would you mind doing that, please?

superset-frontend/src/explore/components/controls/SelectControl.jsx
```diff
@@ -124,6 +124,17 @@ export default class SelectControl extends React.PureComponent {
...
+    if (!this.state.options.some(option => option.value === onChangeVal)) {
+      this.setState(previousState => {
+        const newOptions = [
+          ...previousState.options,
+          { value: onChangeVal, label: onChangeVal },
+        ];
+        newOptions.sort((a, b) => a.value - b.value);
+        return { options: newOptions };
+      });
+    }
```
geido, line 137: @aobapirak can you confirm the need for this? To my understanding as long as the `allowNewOptions` for the Select is true, the component will allow for the creation of new options without having to rely on custom state?

aobapirak: The new option doesn't stick to the options, i.e. whenever I enter a new option it will disappear when I add a new option, except for refreshing the page it will show all new options from this code:
`mapStateToProps: ({ controls }: ControlPanelState) => ({ choices: controls?.server_page_length_options?.value, })`

So the code above I created to allow SelectControl can be sticked new option to the options and sort them.

geido: @aobapirak you are right that the options are not persisted unless selected. I had a different memory of this piece of the Select component. There is a problem of introducing this behaviour in the SelectControl wrapper as there is a risk of introducing inconsistency. Would you be open to have a wider discussion to make this a default behaviour of the Select component instead?

PR #26854, opened 2024-01-29
chore(CI): get docs building on ALL branches.

rusackas:
### SUMMARY

This PR does a few things:

* Splits the github action for docs in half
  * One now only runs on pushes to master... it builds and deploys docs after checking for secrets
  * One now only runs on PRs, and doesn't need secrets. It does typechecking (new step) and builds docs, and should run even on forks (e.g. dependabot bumps)
 * Fixes a couple of typescript checks that were noticed by the typechecking script.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/superset-docs-deploy.yml
+++ b/.github/workflows/superset-docs-deploy.yml
@@ -1,12 +1,29 @@
-name: Docs Testing
+name: Docs Deployment
 
 on:
-  pull_request:
+  push:
     paths:
       - "docs/**"
-    types: [synchronize, opened, reopened, ready_for_review]
+    branches:
+      - "master"
+
 jobs:
+  config:
+    runs-on: "ubuntu-latest"
+    outputs:
+      has-secrets: ${{ steps.check.outputs.has-secrets }}
+    steps:
+      - name: "Check for secrets"
+        id: check
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/superset-docs-deploy.yml
```diff
@@ -0,0 +1,29 @@
...
+    defaults:
+      run:
+        working-directory: docs
+    steps:
+      - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          submodules: recursive
+      - name: yarn install
```
mistercrunch, line 21: for these yarn commands, shouldn't there be a `cd docs` before?

.github/workflows/superset-docs-verify.yml
```diff
@@ -1,15 +1,11 @@
-name: Docs
+name: Docs Deployment
```
mistercrunch, line 1: I think the filenames are reversed from what they should be

mistercrunch (changes requested):

.github/workflows/superset-docs-deploy.yml
rusackas: I believe the `working-directory` parameter does that.

mistercrunch (approved):

PR #26868, opened 2024-01-29
chore(actions): run docs actions on Node 16 to conform with the project

rusackas:
### SUMMARY

Some dependabot PRs were passing CI because they were running on node 18 instead of 16. These installs ought to fail on node 16, but weren't. This should make things fail when they ought to. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
I pushed a dummy commit touching the docs/ folder to trigger that CI step

mistercrunch (approved):
Pre-approving pending CI running including the docs/ build

PR #26872, opened 2024-01-29
chore(dependabot): auto-update lockfiles for monorepo package bumps

rusackas:
### SUMMARY

When dependabot bumps packages for monorepo packages/plugins, it doesn't work out. It bumps package.json, but it's not smart enough to know that it needs to run `npm install` from the root of the `superset-frontend` folder. This action will keep an eye out for these PRs from dependabot on those paths, and run the installation command, and push the package lock file.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/update-monorepo-lockfiles.yml
+++ b/.github/workflows/update-monorepo-lockfiles.yml
@@ -1,0 +1,41 @@
+name: Update Lockfiles for Dependabot Monorepo PRs
+
+on:
+  pull_request:
+    branches: 
+      - main
+    paths:
+      - 'superset-frontend/packages/**/package.json'
+      - 'superset-frontend/plugins/**/package.json'
+    # Trigger this workflow when Dependabot creates a pull request
+    types: [opened, synchronize, reopened]
+
+jobs:
+  update-lock-file:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]' # Ensure it only runs for Dependabot PRs
+    defaults:
+      run:
+        working-directory: superset-frontend
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }} # Checkout the branch that made the PR
+
+      - name: Set up Node.js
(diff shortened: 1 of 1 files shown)


.github/workflows/update-monorepo-lockfiles.yml
```diff
@@ -0,0 +1,39 @@
+name: Update Lockfiles for Dependabot Monorepo PRs
+
+on:
+  pull_request:
+    paths:
+      - 'superset-frontend/packages/**/package.json'
+      - 'superset-frontend/plugins/**/package.json'
+    # Trigger this workflow when Dependabot creates a pull request
+    types: [opened, synchronize, reopened]
```
mistercrunch, line 9: I'd say this should only trigger on `opened`, not the others. We don't want it to trigger itself, even though I think `git push` should be a no-op when the files does not change on the second run.

.github/workflows/update-monorepo-lockfiles.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          npm install
+
+      - name: Commit and Push Changes
+        run: |
+          git config user.name "GitHub-Actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add package-lock.json
+          git commit -m "Update lock file for Dependabot PR" -a # Commit the changes
+          git push # Push the changes back to the branch
```
mistercrunch, line 39: Assuming this is a no-op on the second run.

mistercrunch (changes requested):
This will lead to double triggering CI in cases we're trying to fix here. Meaning all `pull_request` events workflows trigger once, including this one, and once the push happens again, CI does a full re-trigger maybe a few minutes later. We could have all other CI tasks depend on this one, but that would delay things. I'd let's make sure this only triggers on `opened` and be ok with the double-trigger stuff

mistercrunch (approved):

mistercrunch:
I approved the change I submitted 😄

PR #26877, opened 2024-01-30
feat(ci): add a check to make sure there's no hold label on the PR

mistercrunch:
noticed on another PR (https://github.com/apache/superset/pull/26836) that we use `hold.*` labels, but they don't really hold the merge. After almost merging and realizing the hold after, I got GPT to generate the action here and tested it on that PR.

Note that this will hold all PRs with any label containing the string `hold`, which is what we need since we have multiple hold-related labels.

Screenshot 2024-01-29 at 5 18 56 PM

### DIFF
--- a/.github/workflows/no-hold-label.yml
+++ b/.github/workflows/no-hold-label.yml
@@ -1,0 +1,20 @@
+name: Hold Label Check
+
+on:
+  pull_request:
+    types: [labeled, unlabeled, opened, reopened, synchronize]
+
+jobs:
+  check-hold-label:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check for 'hold' label
+      uses: actions/github-script@v3
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/no-hold-label.yml
```diff
@@ -0,0 +1,20 @@
...
+  check-hold-label:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check for 'hold' label
+      uses: actions/github-script@v3
+      with:
+        github-token: ${{secrets.GITHUB_TOKEN}}
+        script: |
+          const payload = context.payload.pull_request
+          const label = !!payload.labels.find(label => label.name.includes('hold'))
```
dpgaspar, line 17: nit: `const holdLabelPresent = payload.labels.some(label => label.name.startsWith('hold'));`

dpgaspar (approved):

michael-s-molina:
Nice!

PR #26878, opened 2024-01-30
chore(npm): npm audit fix

mistercrunch:
Just running the old `npm audit fix` to see if it passes CI


rusackas:
I've got one [here](https://github.com/apache/superset/pull/26693) that has a bunch of TS fixes that are close to making the audit actually work. We talked a bit about this, and decided to rekindle that PR (or one like it) after we get through this wave of Dependabot bumps, toward the end of the 4.0 breaking change window.

rusackas:
Closing this for now... let's get through some more dependabot stuff and try again toward the end of the 4.0 window before the stabilization period.

PR #26880, opened 2024-01-30
feat: configuring an extensible PR auto-labeler

mistercrunch:
Configuring this nice automated labeler https://github.com/actions/labeler.

Setting a few rules to get started, but it should be easy to grow and add more conditions as we go.


rusackas:
Looks awesome! Cant wait to pile on more configs for localization, tests, database support, etc etc.

rusackas (approved):

dpgaspar (approved):

PR #26906, opened 2024-01-30
fix(pinot): typo in the name for epoch_ms_to_dttm

ege-st:
### SUMMARY
A typo in the method name for `epoch_ms_to_dttm` (it was called `epoch_ms_to_dttm_`) in the Pinot DB Engine Spec caused `epoch_ms_to_dttm` to not be correctly overridden. This meant that the base method was called when processing Epoch MS timestamps and results of datetime conversions were incorrect.

### TESTING INSTRUCTIONS
1. Start a Pinot server and create a table with a datetime column that is storing data in Milliseconds from Epoch
2. Add that table as a data source to Superset
3. Choose any chart and make the time column the X-Axis and pick any function as the metric
4. Without this fix, the timestamps for data should incorrectly be shown as 1970.
5. With this fix, the timestamp will be correct.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/pinot.py
+++ b/superset/db_engine_specs/pinot.py
@@ -62,7 +62,7 @@ class PinotEngineSpec(BaseEngineSpec):
         )
 
     @classmethod
-    def epoch_ms_to_dttm_(cls) -> str:
+    def epoch_ms_to_dttm(cls) -> str:
         return (
             "DATETIMECONVERT({col}, '1:MILLISECONDS:EPOCH', "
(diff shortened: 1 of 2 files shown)


villebro (approved):
LGTM, thanks for following up and adding a test for this 👍

villebro:
Restarted the flaky test (will follow up with a fix to unflake it)

rusackas:
re-running the last CI check 🤞

ege-st:
Thanks!

PR #26911, opened 2024-01-31
fix: handle CRLF endings causing sqlglot failure

mapledan:
In SQLLab history page, the `query` table is used to store SQL queries. However, the `sql` column may contain CRLF 
 characters due to user operations on Windows systems. This can lead to errors in SQL Lab history display, as it interferes with the operation of sqlglot(#26767). 
The solution involves using the `strip` method when parsing the input string with `sqlgolt.parse(...)` to remove leading and trailing whitespace, including CRLF characters. This ensures that the SQL queries are properly formatted and displayed without errors in the SQLLab history.

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
I'm surprised that `sqlglot.parse` chokes on `\n\t\r`, we may want to wrap that method so that we don't hit that bug again, as in:
```
def safe_sqlglot_parse(sql):
    return parse(sql.strip("\t\r\n ")
```

dpgaspar (commented):
pre-commit is failling with:

```
trim trailing whitespace.................................................Failed
```

make sure to run `pre-commit`

mapledan:
The failure pre-commit [https://github.com/apache/superset/actions/runs/7724816073/job/21057720031#step:6:78](job) is not related to the changes made in this pull request.

mistercrunch (approved):

mistercrunch:
Did some prevention here -> https://github.com/apache/superset/pull/26942 :)

mistercrunch:
Mind rebasing? Otherwise LGTM

mistercrunch:
giphy

PR #26920, opened 2024-01-31
fix(ci): lint issue on update-monorepo-lockfiles.yml

dpgaspar:
### SUMMARY
Fixes lint issue on `update-monorepo-lockfiles.yml`, impacts open PRs

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26942, opened 2024-01-31
chore(ci): run pre-commit across the repo

mistercrunch:
https://github.com/apache/superset/pull/26920 caught an issue that should have been picked up by pre-commit (by the user), but then re-caught by CI, preventing the merge.

The reason is was not caught was that the CI check is part of a group of python misc checks that trigger only where the python package's folder has changed. In reality, the python utility pre-commit we use here is used for much more than just Python files, so we should run it across the repo.

Here I simply factored out the check into its own GitHub Action that triggers for every single CI run. The code block in this PR is identical/unchanged

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,54 @@
+name: pre-commit checks
+
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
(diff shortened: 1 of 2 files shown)


rusackas (approved):
Thanks! Here's to a lint-free future.

john-bodley (approved):

PR #25243, opened 2023-09-09
build: use GitHub Actions concurrency groups to cancel workflows

sebastianliebscher:
### SUMMARY
This PR removes custom scripts to cancel running GitHub Actions workflows and replaces them with the built-in functionality [concurrency groups](https://docs.github.com/en/actions/using-jobs/using-concurrency#overview).

This will cancel all running workflow jobs for PRs only. In case it is not a PR, this won't cancel running jobs as `run_id` is unique.
```
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
  cancel-in-progress: true
```

Partially reverts https://github.com/apache/superset/pull/12090

---

In case we want to cancel running jobs for commits to master or release tags also, it can be modified to:
```
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

### TESTING INSTRUCTIONS
You can see the cancelled workflows for the first commit:
image

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26945, opened 2024-02-01
feat(ci): kill duplicate CI jobs on PRs

mistercrunch:
Taking over https://github.com/apache/superset/pull/25243, where @sebastianliebscher did great work and submitted this idea. I'm re-opening under a new PR to accelerate things a bit. I've been triggering lots and lots of CI jobs while working on docker and feeling guilty about my carbon footprint.

Here I grepped for workflows that have pull_request events and copy pasted the exact same snippet everywhere except for one place and that's the comment-triggered ephemeral builds, where I reference the issue number instead of the PR number. This should allow for someone to oops and retrigger the ephemeral build, and the lastest one should always land first, fixing potential race conditions

### DIFF
--- a/.github/workflows/ephemeral-env.yml
+++ b/.github/workflows/ephemeral-env.yml
@@ -4,6 +4,11 @@ on:
   issue_comment:
     types: [created]
 
+# cancel previous workflow jobs for PRs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.issue.number || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   config:
     runs-on: "ubuntu-latest"
--- a/.github/workflows/check_db_migration_confict.yml
+++ b/.github/workflows/check_db_migration_confict.yml
@@ -10,6 +10,11 @@ on:
       - "superset/migrations/**"
     types: [synchronize, opened, reopened, ready_for_review]
 
+# cancel previous workflow jobs for PRs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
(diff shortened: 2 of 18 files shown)


.github/workflows/ephemeral-env.yml
```diff
@@ -4,6 +4,11 @@ on:
   issue_comment:
     types: [created]
 
+# cancel previous workflow jobs for PRs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.issue.number || github.run_id }}
```
mistercrunch, line 9: this is the only edit that's different from the other ones, pointing to `github.event.issue.number` instead of `github.event.pull_request.number`

rusackas (approved):
So excited about this... thanks @sebastianliebscher for kicking this off. This project will have a much smaller carbon footprint, and hopefully we'll see a lot less competition for resources among Apache projects :)

rusackas:
Actually... if this _does_ solve our problem with double-runs (like pushing those package lock files) then we can solve another DevEx problem in a similar way... we can run `pre-commit` and `lint-fix` on PRs and push the changes, which would probably un-block PRs where contributors didn't know they need to do that.

dpgaspar (approved):

mistercrunch:
> we'll see a lot less competition for resources among Apache projects
there's an incentive for us to go and do a similar PR in Apache projects that are using a significant amount of resources :) 

@rusackas do we know of a way to monitor resources we compete for at the org (apache) level? Wouldn't be crazy to submit a similar PR against say `apache/kafka` if we knew they're hogging a significant part of the resource pool...

PR #26946, opened 2024-02-01
fix: column values with NaN

betodealmeida:
### SUMMARY

Simple fix so that `values_for_column` doesn't return `NaN`; instead, they are replaced by `None`s, which can be serialized correctly to `null` in JS. The `NaN` values show up when a column has a `NULL` for some databases — we had a report with Athena, and I was able to reproduce with Postgres.

### TESTING INSTRUCTIONS

I added an integration test covering the bug.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/helpers.py
+++ b/superset/models/helpers.py
@@ -1340,7 +1340,10 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
         return and_(*l)
 
     def values_for_column(
-        self, column_name: str, limit: int = 10000, denormalize_column: bool = False
+        self,
+        column_name: str,
+        limit: int = 10000,
+        denormalize_column: bool = False,
     ) -> list[Any]:
         # denormalize column name before querying for values
         # unless disabled in the dataset configuration
@@ -1378,6 +1381,8 @@ class ExploreMixin:  # pylint: disable=too-many-public-methods
             sql = self.mutate_query_from_config(sql)
 
             df = pd.read_sql_query(sql=sql, con=engine)
+            # replace NaN with None to ensure it can be serialized to JSON
+            df = df.replace({np.nan: None})
             return df["column_values"].to_list()
 
     def get_timestamp_expression(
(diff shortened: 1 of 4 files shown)


superset/models/helpers.py
```diff
@@ -1378,6 +1381,8 @@ def values_for_column(
             sql = self.mutate_query_from_config(sql)
 
             df = pd.read_sql_query(sql=sql, con=engine)
+            # replace NaN with None to ensure it can be serialized to JSON
+            df = df.replace({np.nan: None})
```
villebro, line 1385: I'm curious, do we ever _not_ want to do this?

villebro (commented):
First pass existential question

image

superset/models/helpers.py
john-bodley: What I'm somewhat perplexed about is why historically this hasn't been an issue? I'm also not entirely sure when  the `values_for_column` function is invoked.

Note it seems like Trino et al. support [NaNs](https://trino.io/docs/current/functions/math.html), though interesting in SQL Lab if I run `SELECT nan()` the result set renders `NULL`.

betodealmeida: @john-bodley this only happens in numeric columns with `NULL` values. The function is called when populating values for a filter dropdown. My assumption is that we haven't noticed this because numeric columns are not commonly used for filters; most of them are strings.

The biggest issue here, which my PR does not address, is that our backend JSON encoder is happily encoding the `NaN`s like this:

```json
[1, NaN, 3]
```

This is not valid JSON as far as I know (the spec is unclear on `null` and `NaN`, which are Javascript objects but technically not valid JSON).

john-bodley: Thanks @betodealmeida for clarifying. Who doesn't love the whole encoding/decoding of these more atypical types?

villebro: TBH I had forgotten JS actually has a `NaN` 😁

geido (approved):

sadpandajoe:
🏷️ preset:2024.5

PR #26965, opened 2024-02-01
fix(tags): Improve support for tags with colons

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/26701, to rely on the API filter so that only tags created by users are returned in modals/etc, avoiding the need to filter for tags with `:`. This improves the support for user tags created with `:`. 

I checked all the methods updated by this PR and where they're used and these changes shouldn't affect other functionalities. Also confirmed that system tags are still being added as usual. 

**Before**
* Adding a tag with `:` to an asset wouldn't persist from the asset list. 
* Tagging the asset via the Tags list would work, but then the tag wasn't included when accessing the asset properties.

https://github.com/apache/superset/assets/96086495/dec372bb-c48d-4d28-b837-cd6b97ca8bc8

**After**
* Tags with `:` can now be assigned from the assets list and are also returned when accessing the properties modal

https://github.com/apache/superset/assets/96086495/09e0e4e4-9428-458c-9877-024805b224ac

### TESTING INSTRUCTIONS
1. Create a tag named `test:tag`. 
2. Navigate to **Dashboards**.
3. Click on the pencil icon under the **Actions** column to modify any dashboard.
4. Add the `test:tag` to the asset and save changes.
5. Refresh the page, and validate that the tag is successfully associated with the asset.
6. Click on the pencil icon again to access the properties, and validate the tag persist to the dropdown. 

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [x] Required feature flags: `TAGGING_SYSTEM`
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Tags/utils.tsx
+++ b/superset-frontend/src/components/Tags/utils.tsx
@@ -56,7 +56,10 @@ export const loadTags = async (
 ) => {
   const searchColumn = 'name';
   const query = rison.encode({
-    filters: [{ col: searchColumn, opr: 'ct', value: search }],
+    filters: [
+      { col: searchColumn, opr: 'ct', value: search },
+      { col: 'type', opr: 'custom_tag', value: true },
+    ],
     page,
     page_size: pageSize,
     order_column: searchColumn,
@@ -78,9 +81,7 @@ export const loadTags = async (
       const data: {
         label: string;
         value: string | number;
-      }[] = response.json.result
-        .filter((item: Tag & { table_name: string }) => item.type === 1)
-        .map(tagToSelectOption);
+      }[] = response.json.result.map(tagToSelectOption);
       return {
         data,
         totalCount: response.json.count,
--- a/superset-frontend/src/features/tags/tags.ts
+++ b/superset-frontend/src/features/tags/tags.ts
@@ -94,11 +94,7 @@ export function fetchTags(
     endpoint: `/api/v1/${objectType}/${objectId}`,
   })
     .then(({ json }) =>
-      callback(
-        json.result.tags.filter(
-          (tag: Tag) => tag.name.indexOf(':') === -1 || includeTypes,
-        ),
-      ),
+      callback(json.result.tags.filter((tag: Tag) => tag.type === 1)),
     )
     .catch(response => error(response));
 }
@@ -167,9 +163,6 @@ export function addTag(
 ) {
   if (objectType === undefined || objectId === undefined) {
     throw new Error('Need to specify objectType and objectId');
-  }
-  if (tag.indexOf(':') !== -1 && !includeTypes) {
-    return;
   }
   const objectTypeId = map_object_type_to_id(objectType);
   SupersetClient.post({
--- a/superset-frontend/src/features/tags/tags.ts
+++ b/superset-frontend/src/features/tags/tags.ts
@@ -47,10 +47,15 @@ const map_object_type_to_id = (objectType: string) => {
 };
 
 export function fetchAllTags(
+  // fetch all tags (excluding system tags)
   callback: (json: JsonObject) => void,
   error: (response: Response) => void,
 ) {
-  SupersetClient.get({ endpoint: `/api/v1/tag` })
+  SupersetClient.get({
+    endpoint: `/api/v1/tag/?q=${rison.encode({
+      filters: [{ col: 'type', opr: 'custom_tag', value: true }],
+    })}`,
+  })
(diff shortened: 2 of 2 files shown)


superset-frontend/src/features/tags/tags.ts
```diff
@@ -47,10 +47,15 @@ const map_object_type_to_id = (objectType: string) => {
...
 
 export function fetchAllTags(
+  // fetch all tags (excluding system tags)
   callback: (json: JsonObject) => void,
   error: (response: Response) => void,
 ) {
-  SupersetClient.get({ endpoint: `/api/v1/tag` })
+  SupersetClient.get({
+    endpoint: `/api/v1/tag/?q=${rison.encode({
+      filters: [{ col: 'type', opr: 'custom_tag', value: true }],
```
john-bodley, line 56: @Vitor-Avila you mention in your PR that you're excluding system tags yet here you're including only custom tags. Could these differ?

Vitor-Avila: @john-bodley It should be the same thing. I implemented the `custom_tag` filter operator in [this PR](https://github.com/apache/superset/pull/26701/files#diff-8db349e723a165326c5a22dd71dd76e16dafa97288e38318b16dd0c50e867eb2R34-R36) which when set to `True` returns only tags with `type == 'custom'` (created by users):
``` python
class UserCreatedTagTypeFilter(BaseFilter):  # pylint: disable=too-few-public-methods
    """
    Filter for tag type.
    When set to True, only user-created tags are returned.
    When set to False, only system tags are returned.
    """

    name = _("Is custom tag")
    arg_name = "custom_tag"

    def apply(self, query: Query, value: bool) -> Query:
        if value:
            return query.filter(Tag.type == TagType.custom)
        if value is False:
            return query.filter(Tag.type != TagType.custom)
        return query
```

[Reference](https://github.com/apache/superset/blob/master/superset/tags/models.py#L60-L76):
``` python
class TagType(enum.Enum):
    """
    Types for tags.

    Objects (queries, charts, dashboards, and datasets) will have with implicit tags based
    on metadata: types, owners and who favorited them. This way, user "alice"
    can find all their objects by querying for the tag `owner:alice`.
    """

    # pylint: disable=invalid-name
    # explicit tags, added manually by the owner
    custom = 1

    # implicit tags, generated automatically
    type = 2
    owner = 3
    favorited_by = 4
```

superset-frontend/src/components/Tags/utils.tsx
```diff
@@ -78,9 +81,7 @@ export const loadTags = async (
       const data: {
         label: string;
         value: string | number;
-      }[] = response.json.result
-        .filter((item: Tag & { table_name: string }) => item.type === 1)
-        .map(tagToSelectOption);
+      }[] = response.json.result.map(tagToSelectOption);
```
Vitor-Avila, line 84: Added the filter to the API call, so we don't have to post-filter results.

superset-frontend/src/features/tags/tags.ts
```diff
@@ -89,11 +94,7 @@ export function fetchTags(
     endpoint: `/api/v1/${objectType}/${objectId}`,
   })
     .then(({ json }) =>
-      callback(
-        json.result.tags.filter(
-          (tag: Tag) => tag.name.indexOf(':') === -1 || includeTypes,
-        ),
-      ),
+      callback(json.result.tags.filter((tag: Tag) => tag.type === 1)),
```
Vitor-Avila, line 97: Using the `type` to filter so that tags with `:` can be used accordingly.

superset-frontend/src/features/tags/tags.ts
```diff
@@ -163,9 +164,6 @@ export function addTag(
   if (objectType === undefined || objectId === undefined) {
     throw new Error('Need to specify objectType and objectId');
   }
-  if (tag.indexOf(':') !== -1 && !includeTypes) {
-    return;
-  }
```
Vitor-Avila, line 168: This was preventing tags with `:` to be successfully added to assets. Given that system tags are not included in the Tags list (in the Properties modal) we should be good to remove this check.

betodealmeida (approved):

superset-frontend/src/features/tags/tags.ts
eschutho: @Vitor-Avila can you point to the enum object instead of the value? Something like TagType.custom instead of `===1`?

PR #24112, opened 2023-05-18
chore: Bump to Python3.10

EugeneTorap:
### SUMMARY
Running CI integration and unit tests for Py3.10 and Py3.11.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #26988, opened 2024-02-02
chore: Comment 3.9 PY required CI checks

EugeneTorap:
### SUMMARY

Before remove 3.9 PY support and bump to 3.10 PY version in #24112 we need to comment 3.9 PY CI tests in github config file.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


.asf.yaml
```diff
@@ -67,16 +67,16 @@ github:
           - cypress-matrix (2, chrome)
           - cypress-matrix (3, chrome)
           - frontend-build
-          - pre-commit (3.9)
-          - python-lint (3.9)
-          - test-mysql (3.9)
-          - test-postgres (3.9)
-          - test-postgres (3.10)
-          - test-sqlite (3.9)
+#          - pre-commit (3.9)
```
john-bodley, line 70: @EugeneTorap why not simply remove as opposed to comment out. Additionally why is line #74 commented out?

mistercrunch:
We do want to keep these checks, I think what's unclear/confusing on your other PR is that you need to rebase. Messing around with GitHub actions and checks recently, I learned that:

- required checks are dictated by what's on the `master` (target) branch
- "executed" checks are defined by the GitHub actions in branch of the PR
- if new checks are now required and your branch doesn't have the GH actions related to it, it will hold, in other words, you need to rebase
- sometimes, say if you want to remove/delete, you'll need a no-op check, or you have to do it in two phases (remove the required check first, and once that's merged, then remove the check), I added one in `.github/workflows/no-op.yml` to work around that

PR #26989, opened 2024-02-02
chore: Remove database ID dependency for SSH Tunnel creation

geido:
### SUMMARY
This PR removes the requirement to pass a database ID to the SSHTunnel command. It also removes the premature call of SQLAlchemy ORM listeners, such as `after_insert`, caused by the use of `flush` and `begin_nested`.

N.A.

### TESTING INSTRUCTIONS
- [ ] Create an SSH server.
- [ ] Make sure that this server is accessible from the Preset IPs.
- [ ] Create a DB connection.
- [ ] Restrict access to the DB so that only traffic from the SSH server is accepted.
- [ ] Try setting up the connection via SSH on a Preset Workspace.
- [ ] All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/create.py
+++ b/superset/commands/database/create.py
@@ -99,7 +99,11 @@ class CreateDatabaseCommand(BaseCommand):
                     "schema_access", security_manager.get_schema_perm(database, schema)
                 )
 
-        except (SSHTunnelInvalidError, SSHTunnelCreateFailedError, SSHTunnelingNotEnabledError) as ex:
+        except (
+            SSHTunnelInvalidError,
+            SSHTunnelCreateFailedError,
+            SSHTunnelingNotEnabledError,
+        ) as ex:
             db.session.rollback()
             event_logger.log_with_context(
                 action=f"db_creation_failed.{ex.__class__.__name__}.ssh_tunnel",
--- a/superset/commands/database/create.py
+++ b/superset/commands/database/create.py
@@ -114,6 +114,7 @@ class CreateDatabaseCommand(BaseCommand):
         except (
             DAOCreateFailedError,
             DatabaseInvalidError,
+            Exception,
         ) as ex:
             db.session.rollback()
             event_logger.log_with_context(
@@ -152,4 +153,6 @@ class CreateDatabaseCommand(BaseCommand):
             raise exception
 
     def _create_database(self) -> Database:
-        return DatabaseDAO.create(attributes=self._properties, commit=False)
+        database = DatabaseDAO.create(attributes=self._properties, commit=False)
+        database.set_sqlalchemy_uri(database.sqlalchemy_uri)
+        return database
--- a/superset/commands/database/create.py
+++ b/superset/commands/database/create.py
@@ -84,14 +84,13 @@ class CreateDatabaseCommand(BaseCommand):
                 if not is_feature_enabled("SSH_TUNNELING"):
                     raise SSHTunnelingNotEnabledError()
 
-                database = self._do_create_database()
+                database = self._create_database()
                 ssh_tunnel = CreateSSHTunnelCommand(
                     database, ssh_tunnel_properties
                 ).run()
             else:
-                database = self._do_create_database()
+                database = self._create_database()
 
-            db.session.add(database)
             db.session.commit()
 
             # adding a new database we always want to force refresh schema list
@@ -153,5 +152,5 @@ class CreateDatabaseCommand(BaseCommand):
             )
             raise exception
 
-    def _do_create_database(self) -> Database:
-        return DatabaseDAO.create(attributes=self._properties, commit=False, add=False)
+    def _create_database(self) -> Database:
+        return DatabaseDAO.create(attributes=self._properties, commit=False)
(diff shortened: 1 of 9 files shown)


superset/commands/database/ssh_tunnel/create.py
```diff
@@ -28,48 +28,39 @@
...
     def run(self) -> Model:
         try:
-            # Start nested transaction since we are always creating the tunnel
-            # through a DB command (Create or Update). Without this, we cannot
-            # safely rollback changes to databases if any, i.e, things like
-            # test_do_not_create_database_if_ssh_tunnel_creation_fails test will fail
-            db.session.begin_nested()
             self.validate()
-            return SSHTunnelDAO.create(attributes=self._properties, commit=False)
+            ssh_tunnel = SSHTunnelDAO.create(attributes=self._properties, commit=False)
```
Antonio-RiveroMartnez, line 45: Do we need to change something at the Model level with a migration?

geido: I don't think so as Database was already a relationship in the model

superset/commands/database/create.py
```diff
@@ -76,34 +77,22 @@ def run(self) -> Model:
...
-                    # So we can show the original message
-                    raise ex
-                except Exception as ex:
-                    event_logger.log_with_context(
-                        action=f"db_creation_failed.{ex.__class__.__name__}.ssh_tunnel",
-                        engine=self._properties.get("sqlalchemy_uri", "").split(":")[0],
-                    )
-                    raise DatabaseCreateFailedError() from ex
+
+                database = self._do_create_database()
```
eschutho, line 87: @geido why wouldn't we want to add the database to the session here?

eschutho (commented):
I think this looks really good. Thanks @geido. My only question is around whether we need the add param on the db dao.

superset/commands/database/create.py
geido: @eschutho the rollback will look for the last item added to the session. I want the database to be last, so if the creation of the ssh tunnel fails, the database will be rolled back. There are other ways to define what to rollback, using checkpoints for instance, however from my experience those will fire flush events, which will eventually cause the premature fire of the event listeners.

eschutho: Wouldn't it roll back everything in the session? Including both the db and ssh_tunnel if either of them raised?

geido: I could not get it to behave like that unfortunately. Let me see if I've missed anything

superset/daos/base.py
```diff
@@ -133,6 +133,7 @@ def create(
         item: T | None = None,
         attributes: dict[str, Any] | None = None,
         commit: bool = True,
+        add: bool = True,
```
betodealmeida, line 136: I'm concerned about this, because from my understanding this shouldn't be the issue, and I'm worried that we're introducing a pattern that would cause more subtle bugs.

Can we implement this only in `DatabaseDAO.create`, and only for the SSH tunnel case? (ie, don't make it a parameter)

geido: You are right that wasn't the issue. This should not be required. I removed it. Thanks!

superset/commands/database/create.py
```diff
@@ -76,34 +77,21 @@ def run(self) -> Model:
...
-                    # So we can show the original message
-                    raise ex
-                except Exception as ex:
-                    event_logger.log_with_context(
-                        action=f"db_creation_failed.{ex.__class__.__name__}.ssh_tunnel",
-                        engine=self._properties.get("sqlalchemy_uri", "").split(":")[0],
-                    )
-                    raise DatabaseCreateFailedError() from ex
+
+                database = self._create_database()
```
betodealmeida, line 87: I think we can move this out of the `try/if` block to remove a few lines of code; maybe put it after line 80?

tests/integration_tests/databases/api_tests.py
```diff
@@ -538,26 +538,28 @@ def test_cascade_delete_ssh_tunnel(
...
+        Database API: Test rollback is called if SSH Tunnel creation fails
         """
         mock_create_is_feature_enabled.return_value = True
         self.login(username="admin")
-        example_db = get_example_database()
-        if example_db.backend == "sqlite":
-            return
+        # example_db = get_example_database()
+        # if example_db.backend == "sqlite":
+        # return
```
betodealmeida, line 556: Remove these lines?

tests/integration_tests/databases/api_tests.py
```diff
@@ -566,21 +568,17 @@ def test_do_not_create_database_if_ssh_tunnel_creation_fails(
...
-        model = (
-            db.session.query(Database)
-            .filter(Database.database_name == "test-db-failure-ssh-tunnel")
-            .one_or_none()
-        )
-        # the DB should not be created
-        assert model is None
+
+        # Check that rollback was called
+        mock_rollback.assert_called()
```
betodealmeida, line 581: Nice!

betodealmeida (approved):
Looks good, I just left a few comments.

PR #27013, opened 2024-02-05
fix(plugins): Fix dashboard filter in Period Over Period KPI plugin

Antonio-RiveroMartnez:
### SUMMARY

When the PoP KPI chart is rendered in a dashboard and the dashboard has a time range filter, this filter is making the chart useless because it's overriding both queries thus no comparison is happening.

Also, previously if the bigNumber or the prev Number were zero, the % change would be zero when it should have been 100%.

Error:

https://github.com/apache/superset/assets/38889534/92ddc188-2aff-43b1-9f96-4af3ae4f06fa

Fix:

https://github.com/apache/superset/assets/38889534/4cf6162f-eba5-45f3-8d68-53834382f540

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-period-over-period-kpi/src/plugin/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-period-over-period-kpi/src/plugin/buildQuery.ts
@@ -217,7 +217,11 @@ function calculatePrev(
 }
 
 export default function buildQuery(formData: QueryFormData) {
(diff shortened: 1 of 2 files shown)


geido (approved):

sadpandajoe:
🏷️ preset:2024.5

PR #23586, opened 2023-04-05
fix(dashboard-rbac): use normal rbac when no roles chosen

villebro:
### SUMMARY
Currently it's impossible to use the normal RBAC model when the Dashboard RBAC  feature flag is enabled. This is contrary to the description, which on one hand states that when the FF is enabled and no roles are chosen, ALL datasets are accessible, and on the other hand states that regular dataset perms apply.

With these changes, the logic is as follows:
- If no roles are defined for a dashboard, regular RBAC always applies.
- If a role/roles are defined for a dashboard, those take precedent, i.e. regular RBAC no longer applies. This is unchanged
- Datasource permission checks now check if the user has Dashboard RBAC access to a dashboard that references the dataset - if it does, the datasource is assumed to be accessible. If not, regular RBAC applies.

In addition, tests are added, and the documentation is aligned with how the feature actually works.

In addition, when trying to access a dashboard the user didn't have access to, they would either be greeted by one of the following:

Regular RBAC:
image

Dashboard RBAC:
image

Now they get redirected to the main page with the following toast:
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: Closes #18634
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27015, opened 2024-02-05
fix(security manager): Users should not have access to all draft dashboards

Vitor-Avila:
### SUMMARY
The `DASHBOARD_RBAC` FF allows Organizations to manage dashboard access through roles (instead of managing via dataset access). According to the documentation and [recent PR](https://github.com/apache/superset/pull/23586), in case the dashboard has no roles associated with it, the regular dataset validation should happen.

This PR improves the security logic so that draft dashboard metadata isn't leaked. This was the previous logic:

``` python
if is_feature_enabled("DASHBOARD_RBAC") and dashboard.roles:
    if dashboard.published and {role.id for role in dashboard.roles} & {
        role.id for role in self.get_user_roles()
    }:
        return
elif (
    not dashboard.published
    or not dashboard.datasources
    or any(
        self.can_access_datasource(datasource)
        for datasource in dashboard.datasources
    )
):
    return
```

The `not dashboard.published` condition was leaking metadata for any draft dashboard to all users (regardless of the FF state). The actual chart data was not exposed due to dataset validation.

No UI changes.

### TESTING INSTRUCTIONS
1. Create a dashboard in draft state.
2. Access Superset with a user from a role that doesn't have access to the dashboard (either via `DASHBOARD_RBAC` or regular RBAC).
3. Access the dashboard URL.

The dashboard shouldn't load.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -2018,26 +2018,30 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             if self.is_admin() or self.is_owner(dashboard):
                 return
 
-            # If not guest user (embedded), owner or admin,
-            # user should only access published dashboards.
-            if dashboard.published:
-                # DASHBOARD_RBAC logic - Manage dashboard access through roles.
-                # Only applicable in case the dashboard has roles set.
-                if is_feature_enabled("DASHBOARD_RBAC") and dashboard.roles:
-                    if {role.id for role in dashboard.roles} & {
-                        role.id for role in self.get_user_roles()
-                    }:
-                        return
+            # TODO: Once a better sharing flow is in place, we should move the
+            # dashboard.published check here so that it's applied to both
+            # regular RBAC and DASHBOARD_RBAC
 
-                # REGULAR RBAC logic
-                # User can only acess the dashboard in case:
-                #    It doesn't have any datasets; OR
-                #    They have access to at least one dataset used.
-                if not dashboard.datasources or any(
-                    self.can_access_datasource(datasource)
-                    for datasource in dashboard.datasources
-                ):
+            # DASHBOARD_RBAC logic - Manage dashboard access through roles.
+            # Only applicable in case the dashboard has roles set.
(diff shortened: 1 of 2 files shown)


Vitor-Avila:
@john-bodley @villebro @michael-s-molina @betodealmeida @eschutho @dpgaspar would really appreciate your thoughts here! I'm not sure if this should be considered a breaking change, given the user wouldn't be able to load the charts either way because they don't have access to the datasets. Also it feels to me that exposing the draft metadata should be considered a bug.

Thank you!

betodealmeida:
> @john-bodley @villebro @michael-s-molina @betodealmeida @eschutho @dpgaspar would really appreciate your thoughts here! I'm not sure if this should be considered a breaking change, given the user wouldn't be able to load the charts either way because they don't have access to the datasets. Also it feels to me that exposing the draft metadata should be considered a bug.
> 
> Thank you!

AFAIK, we kept this behavior when we introduced Dashboard RBAC for backwards compatibility, but it disagrees with what the documentation says ("if a dashboard has no roles then regular dataset roles are applied", or something like that).

@eschutho for our customers, they should still be able to share draft dashboards, as long as the people seeing them have access to the underlying datasets. Would that work for their use case? My understanding is that they would need dataset access anyway to be able to see any charts in the dataset.

Vitor-Avila:
thanks for the update and additional context, @betodealmeida! I understand now that this "gap" is actually used as a way to get feedback from users before publishing the dashboard.  I would have to update this PR to reflect the scenario discussed here (user is capable of accessing a draft in case they have access to at least one dataset) but I can do that in a new commit (also have to fix some tests anyways).

betodealmeida (approved):
LGTM, makes sense, and matches the documentation.

Vitor-Avila:
thank you @betodealmeida! @eschutho @dpgaspar @john-bodley @michael-s-molina @villebro let me know if you have any concerns with this change. 🙏

dpgaspar (approved):

PR #23730, opened 2023-04-18
docs: add npm publish steps to release/readme

lilykuang:
### SUMMARY

- update release/readme for npm publish steps

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #13593, opened 2021-03-12
chore(explore): bump superset-ui 0.17.19

zhaoyongjie:
### SUMMARY

For enable Metric drag and drop, bumping the latest superset-ui
- PR in main repo metric DND: https://github.com/apache/superset/pull/13575
- PR in superset-ui repo: https://github.com/apache-superset/superset-ui/pull/1004

#### HOW TO USE
setting `ENABLE_EXPLORE_DRAG_AND_DROP` to True

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Changes UI
- [ ] Requires DB Migration.
- [ ] Confirm DB Migration upgrade and downgrade tested.
- [x] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27017, opened 2024-02-05
fix(embedded-sdk): add accessible title to iframe

### LINKED ISSUE #27016: Embedded dashboard iframe missing title (accessibility)
## Screenshot

image

## Description

Screen reader users may rely on the (missing) title attribute to know that the iframe is where the dashboard content is.

## Design input

Propose that the title be "Embedded Dashboard".

bhaugeea:
### SUMMARY

Screen reader users may rely on the (missing) title attribute to know that the iframe is where the dashboard content is. This PR adds a title to the iframe with the generic value "Embedded Dashboard".

image

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #27016 
- [ ] Required feature flags:
- [ ] Changes UI


superset-embedded-sdk/src/index.ts
```diff
@@ -154,6 +154,7 @@ export async function embedDashboard({
       });
 
       iframe.src = `${supersetDomain}/embedded/${id}${dashboardConfig}${filterConfigUrlParams}`;
+      iframe.title = 'Embedded Dashboard';
```
rusackas, line 157: This works in a pinch, but does the dashboard config (or something else) happen to provide the actual title of the dashboard? That would be the ideal thing to put here.

mistercrunch: It seems the sdk would expose the option to set this `iframeTitle`, and fall back on some default like this one "Embedded Dashboard".  Note that for extra points you could i18n that string as well.

bhaugeea: Moved it to the config. Didn't provide a fallback for now because I'd probably butcher the i18n. Would a fallback be required for this, or okay without?

mistercrunch: Ok without AFAIC, the i18n is easy though:

```js
import { t } from '@superset-ui/core';

translatableString = t("Embedded Dashboard");
```

What happens then is the i18n framework scans through the code base and inventorize all those strings on the backend and frontend, where other people can then translate then in various languages.

rusackas:
Thanks for following up... running CI 🤞

bhaugeea:
I believe the whole ui/core package would need to be added as a dependency here. I haven't tried to open the can of worms of actually running this repo so not confident in being able to get that working.

superset-embedded-sdk/src/index.ts
```diff
@@ -77,6 +80,7 @@ export async function embedDashboard({
   id,
   supersetDomain,
   mountPoint,
+  iframeTitle = t("Embedded Dashboard"),
```
mistercrunch, line 83: Hey this look like a public interface, so this new [optional] param should be the last one in the function call for backwards compatibility, otherwise we'l have off-by-one param matching on the function call!

superset-embedded-sdk/src/index.ts
```diff
@@ -154,6 +158,9 @@ export async function embedDashboard({
       });
 
       iframe.src = `${supersetDomain}/embedded/${id}${dashboardConfig}${filterConfigUrlParams}`;
+      if(iframeTitle) {
```
mistercrunch, line 161: I think given the default value set in the function header, this will always have a value, so no `if` required. Actually I think if someone actually wanted an empty string and pass `iframeTitle = ''` or `iframeTitle = null` in the function call, it would prevent them from overriding the `Embedded Dashboard` default :/

mistercrunch (changes requested):

mistercrunch:
> I believe the whole ui/core package would need to be added as a dependency here. I haven't tried to open the can of worms of actually running this repo so not confident in being able to get that working.
Oh you're right about that! Let's actually not do this then. Not worth adding the dependency for the sake of a i18n that one string!

rusackas:
@mistercrunch @bhaugeea Curious if you see a way forward with this, so we can close out the PR and Issue. If there's no intent to rebase this in the near future and get it across the finish line, I'll convert this to a Draft, and the issue may closed as stale in time.

mistercrunch (approved):

mistercrunch:
@rusackas I just made the changes I suggested and removed `superset-ui/core` dependency that was reference, but wasn't added as a dependancy of the sdk since it could have implications.

nikola-arcadis:
I don't see this change is published on npmjs registry with version 0.1.0-alpha.11. How can we get the new package with this change?

rusackas:
Sorry... we need to update a bunch of npm packages. We'll try to get that effort underway soon.

mistercrunch:
@rusackas is the process to do that documented someplace? Oh nevermind found an example PR -> https://github.com/apache/superset/pull/13593/files

rusackas:
That brings in the newly published ones, but here are the instructions (the most current I'm aware of) to actually publish the modules to NPM:

https://github.com/apache/superset/pull/23730/files

This should be made part of the release process, but for now, I have it on my task list to go back and try it on some of the voted-in release branches.

PR #27022, opened 2024-02-05
fix(Chart Annotation modal): Table and Superset annotation options will paginate, exceeding previous max limit 100

### LINKED ISSUE #25421: Annotation Layer modal doesn't list all table charts when there are more than 100 tables
When creating a chart, users are able to add Annotation Layers that can be set to be powered by an existing table chart. However, the dropdown selector to specify a chart isn't returning all table charts available in the Workspace.

#### How to reproduce the bug

1. Make sure you have over 100 charts (duplicate dashboards to get to 100 more quickly if needed) 
2. Create a new Table Chart -- make sure the name starts with Z to see the bug more easily 
3. Try to use this table chart as an annotation layer source on another chart, like a Mixed chart or a time series chart 

### Expected results

When searching for a chart when adding annotations, the new chart appears 

### Actual results

image
The new chart doesn't appear in the chart list 

### Environment

- browser type and version: 
- superset version: `master`
- node.js version: `node -v`
- any feature flags active:

### Checklist

### Additional context

It seems that the endpoint to fetch the charts is only returning 100 results - there doesn't seem to be pagination and it also doesn't seem to be filtering by viz_type = table, which might also reduce the number of charts that would be returned.

rtexelm:
### SUMMARY

Replace the SelectControl component with an AsyncSelect component when choosing the source of the annotation. This addresses the response limiting table options to 100 max. There are relevant changes to the process of fetching and where the fetch is being made in the file.

BEFORE:
limited-charts
AFTER:
all-charts

### TESTING INSTRUCTIONS

- Create 100+ annotations or Tables. Saving dashboards as duplicates and selecting the duplicate charts box will speed this process up.
- Charts must be **created by you** (annotation options are filtered by `'chart_owned_created_favored_by_me'`)
- Create a new chart
- Click the "Add Annotation" button
- Choose an Annotation type requiring a query, e.g. "Event"
- Scroll through the options to verify display of all 100+ charts

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #25421 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
+++ b/superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
@@ -205,15 +205,21 @@ class AnnotationLayer extends React.PureComponent {
     this.handleSelectValue = this.handleSelectValue.bind(this);
     this.handleTextValue = this.handleTextValue.bind(this);
     this.isValidForm = this.isValidForm.bind(this);
-    this.asyncFetch = this.asyncFetch.bind(this);
+    this.fetchOptions = this.fetchOptions.bind(this);
+    this.fetchCharts = this.fetchCharts.bind(this);
+    this.fetchNativeAnnotations = this.fetchNativeAnnotations.bind(this);
+    this.fetchAppliedAnnotation = this.fetchAppliedAnnotation.bind(this);
     this.fetchAppliedChart = this.fetchAppliedChart.bind(this);
-    this.shouldFetchAppliedChart = this.shouldFetchAppliedChart.bind(this);
+    this.fetchAppliedNativeAnnotation =
+      this.fetchAppliedNativeAnnotation.bind(this);
+    this.shouldfetchAppliedAnnotation =
+      this.shouldfetchAppliedAnnotation.bind(this);
   }
 
   componentDidMount() {
-    if (this.shouldFetchAppliedChart()) {
+    if (this.shouldfetchAppliedAnnotation()) {
       const { value } = this.state;
-      this.fetchAppliedChart(value);
+      this.fetchAppliedAnnotation(value);
     }
   }
 
@@ -235,7 +241,7 @@ class AnnotationLayer extends React.PureComponent {
     return sources;
   }
 
-  shouldFetchAppliedChart() {
+  shouldfetchAppliedAnnotation() {
     // If the annotation is not new and the source is a chart
     const { value, valueOptions, sourceType } = this.state;
     return value && !valueOptions[value] && requiresQuery(sourceType);
@@ -312,94 +318,112 @@ class AnnotationLayer extends React.PureComponent {
     });
   }
 
-  asyncFetch = async (_, page, pageSize) => {
-    const { annotationType, sourceType } = this.state;
+  fetchNativeAnnotations = async (search, page, pageSize) => {
+    const queryParams = rison.encode({
+      filters: [
+        {
+          col: 'name',
+          opr: 'ct',
+          value: search,
+        },
+      ],
+      page,
+      page_size: pageSize,
+    });
 
-    if (sourceType === ANNOTATION_SOURCE_TYPES.NATIVE) {
-      const queryParams = rison.encode({
-        page,
-        page_size: pageSize,
-      });
+    const { json } = await SupersetClient.get({
+      endpoint: `/api/v1/annotation_layer/?q=${queryParams}`,
+    });
 
-      const { json } = await SupersetClient.get({
-        endpoint: `/api/v1/annotation_layer/?q=${queryParams}`,
-      });
+    const { result, count } = json;
 
-      const { result, count } = json;
+    const layersObj = result.reduce((acc, layer) => {
+      acc[layer.id] = {
+        value: layer.id,
+        label: layer.name,
+      };
+      return acc;
+    }, {});
 
-      const layersObj = result.reduce((acc, layer) => {
-        acc[layer.id] = {
-          value: layer.id,
-          label: layer.name,
+    const layersArray = Object.values(layersObj);
+
+    this.setState(prevState => ({
+      valueOptions: { ...prevState.valueOptions, ...layersObj },
+    }));
+
+    return {
+      data: layersArray,
+      totalCount: count,
+    };
+  };
+
+  fetchCharts = async (search, page, pageSize) => {
+    const { annotationType } = this.state;
+
+    const queryParams = rison.encode({
+      filters: [
+        { col: 'slice_name', opr: 'chart_all_text', value: search },
+        {
+          col: 'id',
+          opr: 'chart_owned_created_favored_by_me',
+          value: true,
+        },
+      ],
+      order_column: 'slice_name',
+      order_direction: 'asc',
+      page,
+      page_size: pageSize,
+    });
+    const { json } = await SupersetClient.get({
+      endpoint: `/api/v1/chart/?q=${queryParams}`,
+    });
+
+    const { result, count } = json;
+    const registry = getChartMetadataRegistry();
+
+    const chartsObj = result
+      .filter(chart => {
+        const metadata = registry.get(chart.viz_type);
+        return metadata && metadata.canBeAnnotationType(annotationType);
+      })
+      .reduce((acc, chart) => {
+        acc[chart.id] = {
+          value: chart.id,
+          label: chart.slice_name,
+          slice: {
+            ...chart,
+            data: {
+              ...chart.form_data,
+              groupby: chart.form_data.groupby?.map(column =>
+                getColumnLabel(column),
+              ),
+            },
+          },
         };
         return acc;
       }, {});
 
-      const layersArray = Object.values(layersObj);
+    const chartsArray = Object.values(chartsObj);
 
-      this.setState(prevState => ({
-        valueOptions: { ...prevState.valueOptions, ...layersObj },
-      }));
+    this.setState(prevState => ({
+      valueOptions: { ...prevState.valueOptions, ...chartsObj },
+    }));
 
-      return {
-        data: layersArray,
-        totalCount: count,
-      };
+    return {
+      data: chartsArray,
+      totalCount: count,
+    };
+  };
+
+  fetchOptions = (search, page, pageSize) => {
+    const { sourceType } = this.state;
+
+    if (sourceType === ANNOTATION_SOURCE_TYPES.NATIVE) {
+      return this.fetchNativeAnnotations(search, page, pageSize);
     }
 
     if (requiresQuery(sourceType)) {
-      const queryParams = rison.encode({
-        filters: [
-          {
-            col: 'id',
-            opr: 'chart_owned_created_favored_by_me',
-            value: true,
-          },
-        ],
-        order_column: 'slice_name',
-        order_direction: 'asc',
-        page,
-        page_size: pageSize,
-      });
-      const { json } = await SupersetClient.get({
-        endpoint: `/api/v1/chart/?q=${queryParams}`,
-      });
-
-      const { result, count } = json;
-      const registry = getChartMetadataRegistry();
-
-      const chartsObj = result
-        .filter(chart => {
-          const metadata = registry.get(chart.viz_type);
-          return metadata && metadata.canBeAnnotationType(annotationType);
-        })
-        .reduce((acc, chart) => {
-          acc[chart.id] = {
-            value: chart.id,
-            label: chart.slice_name,
-            slice: {
-              ...chart,
-              data: {
-                ...chart.form_data,
-                groupby: chart.form_data.groupby?.map(column =>
-                  getColumnLabel(column),
-                ),
-              },
-            },
-          };
-          return acc;
-        }, {});
-
-      const chartsArray = Object.values(chartsObj);
-
-      this.setState(prevState => ({
-        valueOptions: { ...prevState.valueOptions, ...chartsObj },
-      }));
-
-      return {
-        data: chartsArray,
-        totalCount: count,
-      };
+      return this.fetchCharts(search, page, pageSize);
     }
 
     return {
@@ -408,7 +432,8 @@ class AnnotationLayer extends React.PureComponent {
     };
   };
 
-  fetchAppliedChart(id) {
+  fetchAppliedAnnotation(id) {
+    const { sourceType } = this.state;
     const queryParams = rison.encode({
       filters: [
         {
@@ -418,6 +443,13 @@ class AnnotationLayer extends React.PureComponent {
         },
       ],
     });
+    if (sourceType === ANNOTATION_SOURCE_TYPES.NATIVE) {
+      return this.fetchAppliedNativeAnnotation(id);
+    }
+    return this.fetchAppliedChart(queryParams);
+  }
+
+  fetchAppliedChart(queryParams) {
     SupersetClient.get({
       endpoint: `/api/v1/chart/?q=${queryParams}`,
     }).then(({ json }) => {
@@ -438,6 +470,24 @@ class AnnotationLayer extends React.PureComponent {
                 ),
               },
             },
+          },
+        },
+      }));
+    });
+  }
+
+  fetchAppliedNativeAnnotation(id) {
+    SupersetClient.get({
+      endpoint: `/api/v1/annotation_layer/${id}`,
+    }).then(({ json }) => {
+      const { result } = json;
+      const layer = result;
+      this.setState(prevState => ({
+        valueOptions: {
+          ...prevState.valueOptions,
+          [layer.id]: {
+            value: layer.id,
+            label: layer.name,
           },
         },
       }));
@@ -534,7 +584,7 @@ class AnnotationLayer extends React.PureComponent {
           ariaLabel={t('Annotation layer value')}
           name="annotation-layer-value"
           header={this.renderChartHeader(label, description, value)}
-          options={this.asyncFetch}
+          options={this.fetchOptions}
           value={valueOptions[value] || null}
           onChange={this.handleSelectValue}
           notFoundContent={<NotFoundContent />}
--- a/superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
+++ b/superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
@@ -33,6 +33,7 @@ import {
   withTheme,
 } from '@superset-ui/core';
 import SelectControl from 'src/explore/components/controls/SelectControl';
+import { AsyncSelect } from 'src/components';
 import TextControl from 'src/explore/components/controls/TextControl';
 import CheckboxControl from 'src/explore/components/controls/CheckboxControl';
 import PopoverSection from 'src/components/PopoverSection';
@@ -60,7 +61,8 @@ const propTypes = {
   width: PropTypes.number,
   showMarkers: PropTypes.bool,
   hideLine: PropTypes.bool,
-  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  value: PropTypes.object,
+  currentSlice: PropTypes.object,
   overrides: PropTypes.object,
   show: PropTypes.bool,
   showLabel: PropTypes.bool,
@@ -109,6 +111,21 @@ const NotFoundContentWrapper = styled.div`
   }
 `;
 
+// TODO remove if we don't use it
+const StyledSelectContainer = styled.div`
+  ${({ theme }) => `
+  .type-label {
+    margin-right: ${theme.gridUnit * 2}px;
+  }
+  .Select__multi-value__label > span,
+  .Select__option > span,
+  .Select__single-value > span {
+    display: flex;
+    align-items: center;
+  }
+`}
+`;
+
 const NotFoundContent = () => (
   <NotFoundContentWrapper>
     <EmptyStateSmall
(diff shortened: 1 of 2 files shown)


superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -194,29 +210,26 @@ class AnnotationLayer extends React.PureComponent {
       hideLine,
       // refData
       isNew: !name,
-      isLoadingOptions: true,
-      valueOptions: [],
+      optionsFetch: this.asyncFetch.bind(this),
```
Antonio-RiveroMartnez, line 213: We should not be setting a function in the state nor we should be binding here

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -194,29 +210,26 @@ class AnnotationLayer extends React.PureComponent {
...
-    this.fetchOptions(annotationType, sourceType, isLoadingOptions);
+    this.fetchAppliedChart();
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    if (prevState.sourceType !== this.state.sourceType) {
-      this.fetchOptions(this.state.annotationType, this.state.sourceType, true);
-    }
+  componentDidUpdate() {
+    this.fetchAppliedChart();
```
Antonio-RiveroMartnez, line 232: Do we want/need to fetch this on every update of the component?

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -279,17 +292,26 @@ class AnnotationLayer extends React.PureComponent {
...
   handleAnnotationSourceType(sourceType) {
     const { sourceType: prevSourceType } = this.state;
 
     if (prevSourceType !== sourceType) {
-      this.setState({ sourceType, value: null, isLoadingOptions: true });
+      this.setState({
+        sourceType,
+        value: null,
+        valueOptions: {},
+        optionsFetch: this.asyncFetch.bind(this),
```
Antonio-RiveroMartnez, line 307: Same as above

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,73 +320,149 @@ class AnnotationLayer extends React.PureComponent {
...
-        const queryParams = rison.encode({
-          filters: [
-            {
-              col: 'id',
-              opr: 'chart_owned_created_favored_by_me',
-              value: true,
+  asyncFetch = async (search, page, pageSize) => {
+    const { annotationType, sourceType } = this.state;
+
+    // Bypassing eslint no-unused-vars error
```
Antonio-RiveroMartnez, line 326: Curious about why was this disabled?

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,73 +320,149 @@ class AnnotationLayer extends React.PureComponent {
...
+    return {
+      data: [],
+      totalCount: 0,
+    };
+  };
+
+  fetchAppliedChart() {
+    const { value, valueOptions, sourceType } = this.state;
+    const id = value;
+    if (id && !valueOptions[id] && requiresQuery(sourceType)) {
```
Antonio-RiveroMartnez, line 425: You could move this conditional to the `didUpdate` so you don't call this function is not met?

Antonio-RiveroMartnez (commented):
Thanks for taking care of this 💪 !

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -109,6 +109,22 @@ const NotFoundContentWrapper = styled.div`
   }
 `;
 
+/* TODO remove if we don't use it
```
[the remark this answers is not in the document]
rtexelm: Not totally leftover, this was style I retrieved from the SelectControl component but it looks like the AsyncSelect might not need any additional styling, I just want to make doubly sure before I remove it

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: This and the setting in handleAnnotationSourceType were a way to update the asyncFetch function so that the options don't persist between sourceType changes. This was the only reliable way I could find so far to trigger the re-render. I'll look into a more correct way but any suggestions would help

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: eslint was causing an error because search was never being used anywhere but it's part of the necessary args for AsyncSelect's fetchOptions function

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: Found a solution using the key prop on AsyncSelect

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,73 +320,149 @@ class AnnotationLayer extends React.PureComponent {
...
+
+      const { result, count } = json;
+
+      const layersObj = {};
+      result.forEach(layer => {
+        layersObj[layer.id] = {
+          value: layer.id,
+          label: layer.name,
+        };
+      });
```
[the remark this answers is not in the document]
rtexelm: Kind of a pain to use the map if I'm trying to create an object directly, I'll change it to use reduce which is consistent with what I did to the charts

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -279,17 +292,26 @@ class AnnotationLayer extends React.PureComponent {
...
+        valueOptions: {},
+        optionsFetch: this.asyncFetch.bind(this),
+      });
     }
   }
 
   handleValue(value) {
     this.setState({
-      value,
+      value: value.value,
```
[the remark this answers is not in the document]
rtexelm: Putting the function to fetch in state was a solution to having it re-render when changing the `sourcetype`, otherwise the options persisted. I'm changing it to use the key prop on `AsyncSelect` instead, removing the function from state.

Also I realized using `value.value` is major problem when setting a formula as the value is then just a string

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -462,20 +528,16 @@ class AnnotationLayer extends React.PureComponent {
...
+          /* key to force re-render on sourceType change */
+          key={sourceType}
           ariaLabel={t('Annotation layer value')}
           name="annotation-layer-value"
-          showHeader
-          hovered
-          description={description}
-          label={label}
+          header={this.renderChartHeader(label, description, value)}
           placeholder=""
```
michael-s-molina, line 537: Any reason to disable the default placeholder?

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,74 +306,142 @@ class AnnotationLayer extends React.PureComponent {
...
+        acc[layer.id] = {
+          value: layer.id,
+          label: layer.name,
+        };
+        return acc;
+      }, {});
+
+      const layersArray = Object.values(layersObj);
+
+      this.setState(prevState => ({
```
michael-s-molina, line 340: You shouldn't have a `setState` inside the async function. To fix this, instead of caching all the options loaded by the select and augmenting this state with chart data, you can have a specific state entry to store that information. You can use the annotation layer id as the key and chart data as the value.

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,74 +306,142 @@ class AnnotationLayer extends React.PureComponent {
...
-              col: 'id',
-              opr: 'chart_owned_created_favored_by_me',
-              value: true,
+  handleTextValue(inputValue) {
+    this.setState({
+      value: inputValue,
+    });
+  }
+
+  asyncFetch = async (_, page, pageSize) => {
```
michael-s-molina, line 315: The first parameter is the search input and it's currently being ignored. Any reason for disabling search?

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,74 +306,142 @@ class AnnotationLayer extends React.PureComponent {
...
+        valueOptions: { ...prevState.valueOptions, ...layersObj },
+      }));
+
+      return {
+        data: layersArray,
+        totalCount: count,
+      };
+    }
+
+    if (requiresQuery(sourceType)) {
```
michael-s-molina, line 350: It would be a good idea to break this function into 2: one for `sourceType === ANNOTATION_SOURCE_TYPES.NATIVE` and another for `requiresQuery(sourceType)`

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -298,74 +306,142 @@ class AnnotationLayer extends React.PureComponent {
...
+        },
+      ],
+    });
+    SupersetClient.get({
+      endpoint: `/api/v1/chart/?q=${queryParams}`,
+    }).then(({ json }) => {
+      const { result } = json;
+      const chart = result[0];
+      this.setState(prevState => ({
+        valueOptions: {
```
michael-s-molina, line 427: See my previous comment. This should be stored in its own state entry.

michael-s-molina (changes requested):
Thank you for the fix @rtexelm and for adding tests! I left some first-pass comments.

michael-s-molina:
In order to fix the failing Docker CI you just need to rebase your PR 😉

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: No this was just holdover from the SelectControl component I neglected to review, I'll take it out

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
michael-s-molina: The problem is that this shouldn't be an unused var. You should have something like this:

```
const queryParams = rison.encode({
  filters: [{ col: 'name', opr: 'ct', value: search }],
  page,
  page_size: pageSize,
});
```

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: Could you elaborate on this one a little more? Do you mean get rid of the `valueOptions` state and save only the selected single chart in a state entry?

michael-s-molina: Yes. You don't need to store the options in the state, they are already maintained by the Select component. You need the chart data information associated with a particular option. You can have a state entry specifically for that like a map of option id and chart data to support multiple charts considering that users might change selected options while interacting with the component.

michael-s-molina: @rtexelm I'll try to explain my comment in more depth:

If you check the `asyncFetch` function, you'll see that there's a call to the chart endpoint:

```
const { json } = await SupersetClient.get({
   endpoint: `/api/v1/chart/?q=${queryParams}`,
});
```

If you check what's actually being used from the result you'll see `chart.id`, `chart.slice_name`, `chart.form_data` and `chart.viz_type`. So the first step is to change `queryParams` to only select these columns, using the `columns` key, to avoid fetching more data than needed.

You need the slice information when selecting an option to use in `renderSliceConfiguration`. Here you have 2 options:

1 - Instead of loading `chart.form_data` for all charts in `asyncFetch`, you could load only the `id`, `slice_name` and `viz_type`, and fire another query to load the `form_data` when a user selects a chart, keeping the previous loaded charts in case a user changes the selection. That's what I meant by:

> You can have a state entry specifically for that like a map of option id and chart data to support multiple charts considering that users might change selected options while interacting with the component.

2 - If you want to fire only one query, loading all `form_data` at once, even if probably only one will be used, you don't need a separate state to store that. The `onChange` event of the Select component sends the `value` as the first parameter and the `option` as the second parameter. Your `option` already contains the slice:

```
{
   value: chart.id,
   label: chart.slice_name,
   slice: {...}
}
```

rtexelm: @michael-s-molina I'm not sure if this newest revision is close to what you were thinking but I can always continue to refine it

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: hopefully I addressed this correctly

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -276,96 +306,224 @@ class AnnotationLayer extends React.PureComponent {
...
+
+    if (sourceType === ANNOTATION_SOURCE_TYPES.NATIVE) {
+      return this.fetchNativeAnnotations(search, page, pageSize);
     }
+
+    if (requiresQuery(sourceType)) {
+      return this.fetchCharts(search, page, pageSize);
+    }
+
+    return {
```
michael-s-molina, line 432: It's weird that `fetchOptions` could return:
```
return {
  data: [],
  totalCount: 0,
};

```
I think it would be better to not render the select depending on `sourceType`.

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -276,96 +306,224 @@ class AnnotationLayer extends React.PureComponent {
...
+        {
+          col: 'id',
+          opr: 'eq',
+          value: id,
+        },
+      ],
+      columns: ['form_data'],
+    });
+    SupersetClient.get({
+      endpoint: `/api/v1/chart/?q=${queryParams}`,
```
michael-s-molina, line 450: Did you use the endpoint `/api/v1/chart/ `instead of `/api/v1/chart/<id>` because you can't only query for the `form_data`?

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -194,28 +194,44 @@ class AnnotationLayer extends React.PureComponent {
       hideLine,
       // refData
       isNew: !name,
-      isLoadingOptions: true,
-      valueOptions: [],
+      valueOptions: {},
```
michael-s-molina, line 197: `valueOptions` is not being set anymore. You can remove it.

michael-s-molina (changes requested):
Thanks for addressing the comments @rtexelm.

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: Yes, the properties contained in `form_data` were nested and spread in the response to `/api/v1/chart/<id>` in such a way that `?q=queryParams` seemed like a more efficient method. In fact, I'm not sure how to properly query for this information according to superset guidelines, I followed the patterns already established in this component. The only doc I could find on the subject [is here](https://github.com/apache/superset/blob/master/CONTRIBUTING.md#query) after you mentioned the `columns` keyword.

michael-s-molina: @rtexelm You can use [Superset's Swagger](https://superset.apache.org/docs/api/) to check the API. It looks like `/api/v1/chart/<id>` supports `columns` too so it would be good to call that endpoint 😉

rtexelm: @michael-s-molina wow, this is exactly what I was looking for! Who would have guessed the information is precisely where it should be? haha 🫠  thank you sincerely, I was just scouring the codebase this whole time

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: Getting errors when this isn't initialized in state upon editing an older Annotation

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: spoke too soon, looks like `form_data` is nested in the `query_context`. not sure how to isolate that using the `/${id}` endpoint if it's possible

rtexelm: @michael-s-molina I was able to fix this, thought there was an issue with the result object but it was just my mistake. Curious if it this looks good, when you have a moment.

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
rtexelm: Forgot to remove `prevState` param from `setState`, should be good now

superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -415,29 +557,20 @@ class AnnotationLayer extends React.PureComponent {
...
   renderValueConfiguration() {
-    const {
-      annotationType,
-      sourceType,
-      value,
-      valueOptions,
-      isLoadingOptions,
-    } = this.state;
+    const { annotationType, sourceType, value, valueOptions } = this.state;
+    const validValue = valueOptions?.[value] ?? null;
```
michael-s-molina, line 573: `valueOptions` is just being read here to get the actual `value`. Your state already have a `value` property so I think you can remove all references to `valueOptions`.

rtexelm: I'm using 'valueOptions' to store the object passed to `handleSelectValue` which contains the id and label. This addresses a bug in which the dropdown field appears blank after first selecting/mounting a saved annotation when feeding it the id alone (the `value` in state).

michael-s-molina: Why don't you change the `value` to be an object with `id` and `label`?

rtexelm: `value` has to be kept strictly as the id to suit the way the annotation is being applied to the chart.

michael-s-molina: I'm not sure if I understood. In places where you need the id, you can use `value.id` and in places where you need the id/label pair, you can use `value`. Is there something I'm missing?

rtexelm: I'm thinking of the `applyAnnotation` function where the `value` is being taken from state and can only be the id of the native annotation or chart in order for the annotation to be applied in `AnnotationLayerControl`. If it contains an object the annotation errors. I ran into this when trying to get rid of valueOptions much earlier on and thought best to leave `value` as is to avoid increasing the scope of the changes. Looking at it now, I could change `newAnnotation.value` to be `value.id` if you think that's a better idea?

michael-s-molina: I think you can set

```
newAnnotation.id = value.id;
```

The same way you manually set `color`. You just need to remove `value` from `annotationFields`. What I'm trying to achieve is to simplify the state where the value is set in just one place.

rtexelm: Pushed a commit to address this, let me know what you think. Added a comment to explain the change in type of `value`

michael-s-molina (commented):
Thanks for addressing the comments @rtexelm!

michael-s-molina (approved):
Code LGTM. It would be nice to spin up a test env and get other eyes too @geido @kgabryje

PR #27038, opened 2024-02-07
chore(docs): bump version number in docs example

sfirke:
### SUMMARY
There are example version numbers in the Alerts & Reports docs listed as 2.0.1. At least one new user took this as a sign that this page was stale. It is not. I am bumping it to 3.1.0 to make that clearer.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/installation/alerts-reports.mdx
+++ b/docs/docs/installation/alerts-reports.mdx
@@ -38,10 +38,10 @@ Screenshots will be taken but no messages actually sent as long as `ALERT_REPORT
 Note: All the components required (Firefox headless browser, Redis, Postgres db, celery worker and celery beat) are present in the *dev* docker image if you are following [Installing Superset Locally](/docs/installation/installing-superset-using-docker-compose/).
 All you need to do is add the required config variables described in this guide (See `Detailed Config`).
 
-If you are running a non-dev docker image, e.g., a stable release like `apache/superset:2.0.1`, that image does not include a headless browser.  Only the `superset_worker` container needs this headless browser to browse to the target chart or dashboard.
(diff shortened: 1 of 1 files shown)


rusackas (approved):
Thanks!

mistercrunch:
Is there an easy way in docausurus "templatize" and use a variable like `{{ latest_asf_version }}` so that we could just bump in one place and add a note to bump this variable as part of the release process?

mistercrunch (changes requested):
Setting to "Request changes" as otherwise it'll get stale or force us to grep + multi-update at every release... This also sets a new pattern to keep the docs more dynamic, worth doing if we can figure this out quickly.

rusackas:
I'm fine with letting this through and following up on a fancier version. For getting fancy, we need to differentiate between latest _stable_ (i.e. not a major x.0.0) and _latest_, so doing it as part of the release process might not work the way we want.

We _could_ make a little script as a "prebuild" step for docusaurus, so that whenever it builds, it scrapes GitHub (or Apache, potentially) for all the latest releases, and saves a little JSON object like :

```
{
  latest: 'x.x.x',
  latest_stable: 'x.x.x',
  latest_major: 'x.0.0',
  latest_minor: 'x.x.0',
  latest_patch: 'x.x.x',
  lts: 'x.x.x'
}
```

This could then be directly included in any MDX page to spit out whatever we want.

mistercrunch (approved):

rusackas:
We have a task board for Documentation. I'll put this idea on there for a rainy day :)

PR #27041, opened 2024-02-07
fix(plugins): missing currency on small number format in table chart

justinpark:
### SUMMARY
When the small number format is set, it appears that the currency format is not present in the table chart.
This commit addresses that issue by restoring the missing currency formatter in the small number formatting.

Before:

Screenshot 2024-02-05 at 5 00 13 PM

After:

Screenshot 2024-02-05 at 5 02 03 PM

### TESTING INSTRUCTIONS
Create a table chart and select a column contains a number value less than 1.
Set small number format for the column and currency format as well

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/utils/formatValue.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/utils/formatValue.ts
@@ -17,6 +17,7 @@
  * under the License.
  */
 import {
+  CurrencyFormatter,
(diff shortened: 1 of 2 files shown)


rusackas (approved):
LGTM. Thanks for adding the test coverage!

PR #27047, opened 2024-02-08
feat(actions): auto-label PRs that close issues.

### LINKED ISSUE #5904: dist_bar row_limit doesn't work well with series and breakdown
The distribution bar chart uses row limit that applies a limit to the end of a single query grouping by series and breakdown. Ideally there should be a limit on a subquery for the series groupby.

Here's an example using contribution, the results are limited to 10 rows so it looks like some states have all boys and 0 girls, showing incorrect data because it's getting truncated by the limit.
image

When the fix is in, add a cypress test for the dist bar chart with series, breakdown, and limit 10.

- [x] I have checked the superset logs for python stacktraces and included it here as text if any
- [x] I have reproduced the issue with at least the latest released version of superset
- [x] I have checked the issue tracker for the same issue and I haven't found one similar

### Superset version
master

### Expected results
The limit applies to the number of series results that show up, and the breakdown of that series shows up and does not get limited.

### Actual results
The limit applies to all rows returned series and breakdown

### Steps to reproduce
Create a dist bar with a series and breakdown and add a low row limit

rusackas:
### SUMMARY

Just a theory, but this should look at any given PR and check the description body to see if it auto-closes any issues. If so, label the thing! This should help us prioritize PRs that will also help whittle the Issues backlog.

This also looks at more triggers (including "edited") to run the action on PR description edits. The jobs in the action can be triggered by subsets of these actions (e.g. the main labeler doesn't run on PR description edits).

### TESTING INSTRUCTIONS

Go ahead! Edit this PR description, and look for the addition/removal of the label!

### ADDITIONAL INFORMATION

- [ ] Has associated issue:  
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -40,7 +40,7 @@ jobs:
       with:
         github-token: ${{secrets.GITHUB_TOKEN}}
         script: |
-          const closingKeywords = /\b(close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)\s*:? ?#([0-9]+)/gi;
+          const closingKeywords = /\b(close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)\s*:? ?(?:#([0-9]+)|https:\/\/github\.com\/apache\/superset\/pull\/([0-9]+))/gi;
           const prBody = context.payload.pull_request.body;
           const issueNumbers = [...prBody.matchAll(closingKeywords)].map(match => match[0]);
           const labelName = 'closes-issue';
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -58,7 +58,7 @@ jobs:
           const prBody = context.payload.pull_request.body;
           const issueNumbers = prBody.match(closingKeywords);
           if (issueNumbers) {
-            github.issues.addLabels({
+            github.rest.issues.addLabels({
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -6,7 +6,7 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
-  
+
 
 jobs:
   labeler:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -19,3 +26,44 @@ jobs:
...
+  label-auto-closing:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Log PR Edit
+      run: echo "PR was edited"
+    - name: Auto-label PRs
+      uses: actions/github-script@v7
+      with:
+        github-token: ${{secrets.GITHUB_TOKEN}}
```
dpgaspar, line 41: don't think we have access to `secrets` when triggering by `pull_request`. it works on PRs that don't come from forks

rusackas: I had read somewhere that the repo does provide a generated GITHUB_TOKEN with limited authority (tagging, commenting, etc) on forks, so this _should_ work. 

I can add a secrets check to skip the job if the secret is not available, but I think that secret always is.

rusackas: I kinda want to merge it and test the theory. I can add the secret check to skip the action if it doesn't work.

dpgaspar: I think your right, but default access is readonly: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

rusackas: Hmm... and I see the labeller action above the one I'm adding here has

```
   permissions:
      contents: read
      pull-requests: write
```

... but I'm (a) not sure if _that_ works on forks, or (b) if we can elevate the default permissions like that. 

I'm not sure yet if there's a better approach here, or if there's a way to skip this action for forks if we can't make it work.

geido: Curious to know if we are moving forward with this as I am planning to consider it for a long-term proposal to improve the management of the backlog

rusackas: Well, I sure WANT to merge it, but I'm not sure how to proceed. Maybe @mistercrunch knows how to unblock this thing :P

mistercrunch: Maybe using `pull_request_target`?
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

mistercrunch: In that case, what happens is it will trigger the action that's defined on the main target repo's branch, and say if you do a checkout it would be the target's ref not the PR one. Does that work here? Actually do you need to checkout?

.github/workflows/labeler.yml
```diff
@@ -19,3 +26,44 @@ jobs:
     #   run: |
     #     echo "Running translation scripts"
     #     # Generate .pot -> .po -> .json files
+
+  # This'll look at the PR body and add a label if it contains an issue-closing keyword
+  # it runs on every PR event, including "edited" since people tend to add "closes #123" in the PR body after opening it
+  label-auto-closing:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
mistercrunch, line 35: Are you sure you need to checkout here?

PR #27052, opened 2024-02-08
feat(plugins): Adding colors to BigNumber with Time Comparison chart

Antonio-RiveroMartnez:
### SUMMARY
Currently our BigNumber with Time Comparison chart has no visual indicators to help user easily tell when the differences are positive or negatives. This PR adds a customize control to enable colors in the chart.

Also this PR is updating the thumbnail to reflect an up to date version of the chart with these colors.

Screenshot 2024-02-08 at 15 21 08

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/plugin-chart-period-over-period-kpi/src/PopKPI.tsx
```diff
@@ -30,16 +34,51 @@ const ComparisonValue = styled.div<PopKPIComparisonValueStyleProps>`
...
+        : isPositiveChange
+        ? theme.colors.success.base
+        : theme.colors.error.base;
+
+    return `
+      background-color: ${
+        comparisonColorEnabled ? backgroundColor : defaultBackgroundColor
+      };
+      color: ${comparisonColorEnabled ? textColor : defaultTextColor};
+      padding: 2px 5px;
```
kgabryje, line 60: could we use full grid units? Or does it look weird?

superset-frontend/plugins/plugin-chart-period-over-period-kpi/src/PopKPI.tsx
```diff
@@ -30,16 +34,51 @@ const ComparisonValue = styled.div<PopKPIComparisonValueStyleProps>`
   `}
 `;
 
+const SymbolWrapper = styled.div<PopKPIComparisonSymbolStyleProps>`
+  ${({ theme, percentDifferenceNumber, comparisonColorEnabled }) => {
+    const defaultBackgroundColor = theme.colors.grayscale.light4;
```
kgabryje, line 39: Interesting approach with putting js inside of styled component like that! But I wonder if deducing colours from more "high-level" data belongs to a styled component. Maybe we should move that logic to `PopKPI` and just pass `backgroundColor` and `textColor` to `SymbolWrapper`?
But if you think this solution is better then I won't argue, I don't have a strong opinion here

rusackas:
I'm trying to upgrade Storybook, but ironing out a few last details. When that's done, we should add this plugin to Storybook and make sure it has controls for these options

superset-frontend/plugins/plugin-chart-period-over-period-kpi/src/PopKPI.tsx
Antonio-RiveroMartnez: hehe yeah, I had it like that before and moved to this, it doesn't feel quite "well" but now that you noticed too, let's move it there instead. Thanks

superset-frontend/plugins/plugin-chart-period-over-period-kpi/src/PopKPI.tsx
```diff
@@ -77,18 +135,26 @@ export default function PopKPI(props: PopKPIProps) {
...
+                }
+                comparisonColorEnabled={comparisonColorEnabled}
+              >
+                {symbol}
+              </SymbolWrapper>
+              {index === 0
+                ? prevNumber
+                : index === 1
+                ? valueDifference
+                : percentDifferenceFormattedString}
```
kgabryje, line 155: Hmm I'm not sure about coupling the values with indexes like this. I think it would be cleaner if we had some key-value mapping between symbols and values, or an array like [['#', prevNumber], ['△', valueDifference], ['%', percentDifferenceFormattedString]]

mistercrunch:
Hey, quick note that I'm trying to fix ephemeral envs here -> https://github.com/apache/superset/pull/27057, hoping to merge quickly as soon as CI passes, and to confirm the fix on this PR right after...

kgabryje:
Nit for eventual follow up PR - add advanced analytics tag? cc @yousoph 

image

kgabryje:
I think the selected number formatting doesn't apply to %

image

kgabryje:
The styling is not responsive - when I added the chart to a dashboard in a small container, I can't see the values on the right (the container is also not scrollable horizontally so they're just hidden)

image

kgabryje:
1. Can we add a description tooltip?
2. Does week/month/year mean last week/last month/last year? If so, maybe we could add "last" prefix to the names?

image

kgabryje:
1. Since "Filters (custom)" control works only if Range For Comparison is "Custom", can we make the visibility of that control dynamic, so that it's actually only visible when it's useful? I believe we use such pattern in some other charts.
2. Can we remove the collapsible and maybe change the name of the control to sth like "Custom comparison range"?

image

Antonio-RiveroMartnez:
> 1. Can we add a description tooltip?
> 2. Does week/month/year mean last week/last month/last year? If so, maybe we could add "last" prefix to the names?
> 
> image

There's is work in the making that will enhance this select control with more info to help users. But will be worked in a separated PR.

Antonio-RiveroMartnez:
> 1. Since "Filters (custom)" control works only if Range For Comparison is "Custom", can we make the visibility of that control dynamic, so that it's actually only visible when it's useful? I believe we use such pattern in some other charts.
> 2. Can we remove the collapsible and maybe change the name of the control to sth like "Custom comparison range"?
> 
> image

Yes, this PR ha snot been updated with latest master but a PR got merged addressing this already https://github.com/apache/superset/pull/27054

kgabryje:
That's a TOTAL nitpick but... the name "huge" sounds a bit overdramatic since the text isn't really that big 😆 

image

kgabryje (approved):
LGTM, the bugs I listed can be worked on in the follow up PRs

PR #27053, opened 2024-02-08
chore: prevent prophet from logging non-errors as errors

betodealmeida:
### SUMMARY

Prophet will log errors when optional dependencies are not installed:

```python
# from prophet/plot.py
try:
    import plotly.graph_objs as go
    from plotly.subplots import make_subplots
except ImportError:
    logger.error('Importing plotly failed. Interactive plots will not work.')
```

(It does something similar for `matplotlib`.)

These calls to `logger.error` will pollute and trigger unnecessary alerts.

This PR introduces a decorator `suppress_logging` that can be used in cases like this to suppress the logging:

```python
with suppress_logging("prophet.plot"):
    from prophet import Prophet
```

### TESTING INSTRUCTIONS

```python
>>> import prophet
Importing plotly failed. Interactive plots will not work.
>>>
```

```python
>>> from superset.utils.decorators import suppress_logging
>>> with suppress_logging("prophet.plot"):
...     import prophet
...
>>>
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/decorators.py
+++ b/superset/utils/decorators.py
@@ -191,3 +191,22 @@ def debounce(duration: float | int = 0.1) -> Callable[..., Any]:
 
 def on_security_exception(self: Any, ex: Exception) -> Response:
     return self.response(403, **{"message": utils.error_msg_from_exception(ex)})
+
+
+@contextmanager
+def suppress_logging(
+    logger_name: str | None = None,
+    new_level: int = logging.CRITICAL,
+) -> Iterator[None]:
+    """
+    Context manager to suppress logging during the execution of code block.
+
+    Use with caution and make sure you have the least amount of code inside it.
+    """
+    target_logger = logging.getLogger(logger_name)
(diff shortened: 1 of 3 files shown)


superset/utils/decorators.py
```diff
@@ -191,3 +191,22 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
 
 def on_security_exception(self: Any, ex: Exception) -> Response:
     return self.response(403, **{"message": utils.error_msg_from_exception(ex)})
+
+
+@contextmanager
+def suppress_logging(
```
Antonio-RiveroMartnez, line 197: Not blocking: could we add some tests here? https://github.com/apache/superset/blob/master/tests/unit_tests/utils/test_decorators.py

Antonio-RiveroMartnez (approved):

PR #27056, opened 2024-02-08
fix(ephemeral): fix ephemeral builds in PR

mistercrunch:
fixes the issue "Error response from daemon: No such image"
as observed here
https://github.com/apache/superset/actions/runs/7836111298/job/21383010014


rusackas (approved):
...and there was much rejoicing

mistercrunch:
It's triggering the action defined on the default branch `master`, makes these events super hard to test/fix. I think moving forward the best practice it to add and handle a `workflow_dispatch` event, and wire the right parameters in the right place for testing. Then you can run commands like this to trigger the workflow with any param:
```
 gh workflow run .github/workflows/anyworkflow.yml  -r {any_branch} -f comment_body="@supersetbot unlabel noop" -f issue_number="27046"
```

mistercrunch:
I'm going to have to merge this to fix blindly, #yolo !

PR #27072, opened 2024-02-09
docs(installation): document multi-platform support in Docker builds

mistercrunch:
see title

### DIFF
--- a/docs/docs/installation/docker.mdx
+++ b/docs/docs/installation/docker.mdx
@@ -6,14 +6,16 @@ schemes to help users navigate our offerings.
 
 Images are built and pushed to the [Superset Docker Hub repository](
 https://hub.docker.com/r/apache/superset). Different sets of images are created for:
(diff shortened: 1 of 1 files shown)


docs/docs/installation/docker.mdx
```diff
@@ -61,7 +64,20 @@ build times, larger images, lower layer cache hit rate, ...).
 For production use cases, we recommend that you derive our `lean` image(s) and
 add database support for the database you need.
 
-## On ARM builds and working with Apple silicon
+## On supporting arm64 AND amd64
+
+Only the release builds are multi-platform, supporting `linux/arm64`
```
rusackas, line 69: Based on all the github noise, I'm just wondering if there are _currently_ versions with this support, or if we're lacking this support until the _next_ version. I'm guessing the former, and not the latter 🤞

rusackas (approved):

PR #27073, opened 2024-02-09
fix(drill): no rows returned

### LINKED ISSUE #27028: Error when running a SQL query on an Apache Drill DB that returns an empty result (error - Issue 1002)
### Bug description

Hello,

First of all, Superset is working. I can use it normally and I already created SQL queries working within the SQL Lab and saved them as dataset, then used it for Charts.

**ISSUE**

My issue is when using the SQL Lab to create datasets when the result of a query is an empty result.
For instance here is a simple query fetching data from a postegreSQL database's table:

SELECT * from base_migration.test;

Case 1 : If the table "test" has got at least one row in it, the query will work normally.
superset - query with empty result issue - drill issue 1002 - picture 01

Case 2 : If that table doesn't have any row in it, I get the following error:
superset - query with empty result issue - drill issue 1002 - picture 02
Note : I deleted the only row from the table "test" before running the same query again.

This behavior from Superset is the same on my local computer (I'm using Docker Desktop 4.26.1 for Drill and Superset), and on our integration server.

But if I run the same query directly in Apache Drill, it works and display an empty result.
superset - query with empty result issue - drill issue 1002 - picture 03

Can you help in this issue please ?

----------

**CONTEXT**

I'm creating a dashboard to display the result of a data migration between csv files into a postgreSQL database. I want to use the Table type charts. For each Table chart, it would list rows that were in a given csv file, but not in the target table the data were migrated into. To say it differently, the goal is to display rejects that occured during the data loading. Sometime a table chart would display rows, meaning that there was a reject. But most of the time, there will be no result, meaning that the data loading from this csv file to that table succeed in full.

A first data migration already take place. We have two cases: tables with all rows from their source csv file, and tables missing some rows from their source csv file (aka: rejects). These Table charts would be used for several upcoming data migration rehearsals. This means that the goal is to have no reject on any table, so no row to display in all Table charts.

### How to reproduce the bug

1. Within a database management tool (e.g. pgAdmin 4), create a table "test" in a database you have the Storage Plugin setup for, with these columns: 
- id (integer)
- name (text)

2. Insert a row:
INSERT INTO database.test(
	id, name)
	VALUES ('1', 'abc');

3. In Superset, SQL Lab, use the following query:
SELECT * from database.test;
You should have that one row as the result of this query.

4. Within the database management tool, delete the row from that table:
DELETE FROM database.test
	WHERE id = '1';

5. In Superset, SQL Lab, use the query again:
SELECT * from database.test;
The error message comes.

[Optional] 6. In Superset, SQL Lab, use the query:
SELECT * from database.test WHERE id = 2;
The error message comes.

### Screenshots/recordings

_No response_

### Superset version

3.1.0

### Python version

3.12

### Node version

20.10.0

### Browser

Microsoft Edge

### Additional context

_No response_

### Checklist

betodealmeida:
### SUMMARY

The Drill engine spec currently shows an error when no rows are returned in SQL Lab:

Screenshot 2024-02-09 at 15-53-10 Superset

This PR introduces a custom `fetch_data` method to the Drill engine spec, so that the correct response is returned:

Screenshot 2024-02-09 at 15-53-24 Superset

Fixes https://github.com/apache/superset/issues/27028.

See above.

### TESTING INSTRUCTIONS

1. Run Drill on docker:

```bash
docker run -it --name drill -p 8047:8047 -p 31010:31010 apache/drill
```

2. Go to http://localhost:8047/storage and enable the RDBMS storage.
3. Update the storage and point it to the `examples` database in Postgres:

```json
{
  "type": "jdbc",
  "driver": "org.postgresql.Driver",
  "url": "jdbc:postgresql://host.docker.internal:5432/examples",
  "username": "beto",
  "sourceParameters": {
    "keepaliveTime": 0,
    "minimumIdle": 0,
    "idleTimeout": 3600000,
    "maximumPoolSize": 10,
    "maxLifetime": 21600000
  },
  "authMode": "SHARED_USER",
  "writerBatchSize": 10000,
  "enabled": true
}
```

4. Install the Postgres JDBC driver inside the container:

```bash
docker exec -it -u root drill bash
cd /opt/drill/jars/3rdparty/
wget https://jdbc.postgresql.org/download/postgresql-42.7.1.jar
exit
docker restart drill
```

5. Connect Superset to Drill using: `drill+sadrill://localhost:8047/cp?use_ssl=False`
6. Query the `bart_lines` from Superset (this assumes that the Postgres storage is called `rdbms`):

```sql
SELECT * FROM rdbms.bart_lines
```

7. Delete all rows from the `bart_lines` table — you can do this in Postgres directly, and you should be able to do it from Superset if you've enabled DML in the Drill database.
8. Run the query again. It should show:

Screenshot 2024-02-09 at 15-59-35 Superset

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/drill.py
+++ b/superset/db_engine_specs/drill.py
@@ -14,8 +14,11 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+from __future__ import annotations
+
 from datetime import datetime
-from typing import Any, Optional
+from typing import Any
 from urllib import parse
 
 from sqlalchemy import types
@@ -60,8 +63,8 @@ class DrillEngineSpec(BaseEngineSpec):
 
     @classmethod
     def convert_dttm(
-        cls, target_type: str, dttm: datetime, db_extra: Optional[dict[str, Any]] = None
-    ) -> Optional[str]:
+        cls, target_type: str, dttm: datetime, db_extra: dict[str, Any] | None = None
+    ) -> str | None:
         sqla_type = cls.get_sqla_column_type(target_type)
 
         if isinstance(sqla_type, types.Date):
@@ -76,8 +79,8 @@ class DrillEngineSpec(BaseEngineSpec):
         cls,
         uri: URL,
         connect_args: dict[str, Any],
-        catalog: Optional[str] = None,
-        schema: Optional[str] = None,
+        catalog: str | None = None,
+        schema: str | None = None,
     ) -> tuple[URL, dict[str, Any]]:
         if schema:
             uri = uri.set(database=parse.quote(schema.replace(".", "/"), safe=""))
@@ -89,7 +92,7 @@ class DrillEngineSpec(BaseEngineSpec):
         cls,
         sqlalchemy_uri: URL,
         connect_args: dict[str, Any],
-    ) -> Optional[str]:
+    ) -> str | None:
         """
         Return the configured schema.
         """
@@ -97,7 +100,7 @@ class DrillEngineSpec(BaseEngineSpec):
 
     @classmethod
     def get_url_for_impersonation(
-        cls, url: URL, impersonate_user: bool, username: Optional[str]
+        cls, url: URL, impersonate_user: bool, username: str | None
     ) -> URL:
         """
         Return a modified URL with the username set.
@@ -117,3 +120,23 @@ class DrillEngineSpec(BaseEngineSpec):
                 )
 
         return url
+
+    @classmethod
+    def fetch_data(
+        cls,
+        cursor: Any,
+        limit: int | None = None,
+    ) -> list[tuple[Any, ...]]:
+        """
+        Custom `fetch_data` for Drill.
+
+        When no rows are returned, Drill raises a `RuntimeError` with the message
+        "generator raised StopIteration". This method catches the exception and
+        returns an empty list instead.
+        """
+        try:
+            return super().fetch_data(cursor, limit)
+        except RuntimeError as ex:
+            if str(ex) == "generator raised StopIteration":
+                return []
+            raise
--- a/setup.py
+++ b/setup.py
@@ -154,7 +154,7 @@ setup(
         ],
         "db2": ["ibm-db-sa>0.3.8, <=0.4.0"],
         "dremio": ["sqlalchemy-dremio>=1.1.5, <1.3"],
-        "drill": ["sqlalchemy-drill==0.1.dev"],
+        "drill": ["sqlalchemy-drill>=1.1.4, <2"],
         "druid": ["pydruid>=0.6.5,<0.7"],
         "duckdb": ["duckdb-engine>=0.9.5, <0.10"],
         "dynamodb": ["pydynamodb>=0.4.2"],


eschutho (approved):
LGTM. not sure if you were planning to add any tests? :)

betodealmeida:
> LGTM. not sure if you were planning to add any tests? :)

Not really... since the solution is Drill specific (and we don't have integration tests using Drill) and the implementation so small, any unit test would basically consist of mocks. 🙃

superset/db_engine_specs/drill.py
```diff
@@ -117,3 +120,23 @@ def get_url_for_impersonation(
...
+    @classmethod
+    def fetch_data(
+        cls,
+        cursor: Any,
+        limit: int | None = None,
+    ) -> list[tuple[Any, ...]]:
+        """
+        Custom `fetch_data` for Drill.
+
+        When no rows are returned, Drill raises a `RuntimeError` with the message
```
john-bodley, line 133: @betodealmeida is the Drill SQLAlchemy dialect or DB-API not conforming to the standard? Ideally it would be great to adopt the "shift left" approach and ensure this is handled correctly upstream as opposed for us having to write custom logic.

villebro:
Tagging @cgivre if you're interested in improving the Drill DB-API to be more standards compliant

superset/db_engine_specs/drill.py
cgivre: @john-bodley I'm one of the authors of the Drill SQLAlchemy dialect.  It's been a few years however, so I don't remember much about how it works.  If something should be fixed there, I'm happy to do it.  However, could someone please explain what the exact issue is?

john-bodley: @cgivre looking at [PEP-249](https://peps.python.org/pep-0249/) it seems like a `StopIteration` should be raised if `Cursor.next()` is called and the result set is exhausted. In the base engine spec the [`fetch_data()`](https://github.com/apache/superset/blob/a912fafb6d592703c5b66f59be180220329ab9c4/superset/db_engine_specs/base.py#L749-L787) method calls `Cursor.fetchall()` or `Cursor.fetchmany()` rather than `Cursor.next()`. It would be prudent to double check the Drill DB-API and SQLAlchemy dialects are behaving in accordance with the standards.

I guess the TL;DR is I'm not sure why we need this custom logic for Drill.

Zakyrel:
Hello. 

Since I declared this issue (id 27028), I upgraded my Superset to 3.1.1 to be able to test the fix.

Let's say I want to do this query to save it as a dataset for a table chart :

SELECT company.taxnumber, company.name, company.status, company.termination_date
FROM company
WHERE company.status = 'In activity'
AND company.termination_date IS NOT NULL
;

And I know that right now, within the data I loaded into my table, I have no row matching these two WHERE clauses.

I would like Superset to allow me to display in my dashboard a table chart listing these four fields, with no result found. And later, if I have in this table rows matching WHERE clauses, my table chart would list them.

It looks like that, with this 3.1.1 update, I can't do that.
- The query's result is only a message saying that "The query returned no result".
- When I save this query as a dataset and try to use it to create a table chart, I have no field available to select as columns.

What do you think about this ?

Best regards,

rusackas:
@Zakyrel it's an unusual, but valid use case. We'd probably be open to a proposal/PR for such a fix. 

In the meantime, I wonder if you can do a workaround by adjusting the query to spit out a row of NULLs when there's no REAL data. Something like:

```
SELECT company.taxnumber, company.name, company.status, company.termination_date
FROM company
WHERE company.status = 'In activity'
AND company.termination_date IS NOT NULL

UNION ALL

SELECT NULL as taxnumber, NULL as name, NULL as status, NULL as termination_date
WHERE NOT EXISTS (
    SELECT 1
    FROM company
    WHERE company.status = 'In activity'
    AND company.termination_date IS NOT NULL
);
```

cgivre:
> Tagging @cgivre if you're interested in improving the Drill DB-API to be more standards compliant

Interested yes, but unfortunately I don't have the time right now.  I do think that's probably the "right" place for this fix however.

One thing to note is that Drill is a little tricky about how it handles schemas.  Since Drill doesn't always know the schema in advance, it can be difficult when you run a query and get zero results.

Zakyrel:
> @Zakyrel it's an unusual, but valid use case. We'd probably be open to a proposal/PR for such a fix.
> 
> In the meantime, I wonder if you can do a workaround by adjusting the query to spit out a row of NULLs when there's no REAL data. Something like:
> 
> ```
> SELECT company.taxnumber, company.name, company.status, company.termination_date
> FROM company
> WHERE company.status = 'In activity'
> AND company.termination_date IS NOT NULL
> 
> UNION ALL
> 
> SELECT NULL as taxnumber, NULL as name, NULL as status, NULL as termination_date
> WHERE NOT EXISTS (
>     SELECT 1
>     FROM company
>     WHERE company.status = 'In activity'
>     AND company.termination_date IS NOT NULL
> );
> ```

Hello,

I finally take some time to go back here and answer about this issue.

When I encoutnered this issue, I used a workaround like this :

<my query>
UNION ALL
SELECT NULL, NULL, NULL, NULL

The number of "NULL" I'm using in the last line must be equal tu the number of field fetched with the SELECT of my query.

It forces Drill to send to Superset all rows from my query, but with one row of NULL values. It works, but not the normal way since I have 1 row on each table graph I have (meaning that no data matches the query).

It would be a good improvment to fix this issue, because sometime I have the following issue: New fields are added in a table. And for queries when i do a SELECT *, sudenly my SELECT NULL, NULL, NULL, NULL doesn't have enough "NULL" anymore, ending in an error for the related table graph.

null

Regards,

Zakyrel:
Hello,

I checked the 3.1.3 changelog and didn't find any update for this issue.

Did I miss something ?

Best regards,

michael-s-molina:
> Hello,
> 
> I checked the 3.1.3 changelog and didn't find any update for this issue.
> 
> Did I miss something ?
> 
> Best regards,

This was included in 3.1.1.

Screenshot 2024-07-17 at 08 53 34

Zakyrel:
In my comment on Feb 28, I wrote that the 3.1.1 fix wasn't enough to solve the issue.

Seeing this 3.1.3 tag, I thought maybe something new has been done for that release.

cgivre:
@Zakyrel We did push some recent updates to the Sqlalchemy/Drill dialect.  Have you tried with that?

Zakyrel:
Hello. I tested with our Superset in 3.1.3: Same issue.

3 1 3 same issue 01

The only result we get is this message you emplemented in 3.1.1.

Instead of the usual empty columns result table we would get from other sources than Drill.

Example : 
3 1 3 same issue 02

cgivre:
@jnturton Tagging for visibility.

jnturton:
@Zakyrel did you use sqlalchemy-drill=1.1.5 for your most recent test?

superset/db_engine_specs/drill.py
jnturton: @john-bodley we believe this is now fixed sqlalchemy-drill. That's certainly where it _should_ be fixed.

Zakyrel:
I updated my local version of Superset to make sur I have it on 3.1.3 with sqlalchemy-drill 1.1.5.

image

image

The issue is worst than before. I no longet have the message "The query returned no data", intead I have this :

image

Here is my drill-override.conf with the drill.exec.http.rest.errors.verbose in it.

image

jnturton:
@Zakyrel would you mind coming to the #user channel in [the Drill Slack workspace](https://join.slack.com/t/apache-drill/shared_invite/enQtNTQ4MjM1MDA3MzQ2LTJlYmUxMTRkMmUwYmQ2NTllYmFmMjU4MDk0NjYwZjBmYjg0MDZmOTE2ZDg0ZjBlYmI3Yjc4Y2I2NTQyNGVlZTc) where we can get to the bottom of this? I don't believe the issues you're seeing are originating from Superset.

Zakyrel:
I agree this is not an issue from Apache Superset itself.

I encouter this while using Apache Superset, but with the source of the data being Apache Drill.

I remember starting talking about this issue on Slack  -> Apache Drill. I don't remember how it shifted to Github -> Apache Superset though.

Starting tomorrow I will get back to Slack to discuss the issue.

PR #27088, opened 2024-02-12
chore: POC Upgrade to AntD v5 one component at a time.

rusackas:
### SUMMARY

Imports Antd 5 as an alias. This would allow us to:
* Migrate a single component from `antd` (v4) to `antd-next` (v5)
* Touch up styles/javascript/etc. using Superset and React Storybook as a guide (Appitools would help here too)
* Once that component looks good, move on to the next.
* When all components have been migrated, remove v4, remove the npm alias, and do a simple find/replace to change `antd-next` to `antd`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/package.json
+++ b/superset-frontend/package.json
@@ -124,6 +124,7 @@
     "abortcontroller-polyfill": "^1.1.9",
     "ace-builds": "^1.4.14",
     "antd": "4.10.3",
+    "antd-next": "npm:antd@5.4.0",
     "babel-plugin-typescript-to-proptypes": "^2.0.0",
     "bootstrap": "^3.4.1",
     "brace": "^0.11.1",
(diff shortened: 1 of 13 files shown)


michael-s-molina:
Ant Design [does not recommend multiple versions to coexist](https://ant.design/docs/react/migration-v5#multiple-versions-coexist) but if that's the best route for us then don't forget to adjust the `ConfigProvider`.

rusackas:
I should also say, this POC isn't done yet... still on my "spare time" to do list:
• Stripping back the style overrides for the affected components, and rebuilding until they match the older stories.
• Auditing Storybook to make sure the relevant stories work as expected (controls, etc)
• Prove out the AntD v5 theming capabilities as planned for SIP-82
• Get Applitools back up and running (which might have a licensing issue to contend with)

rusackas:
I also reserve the right to back away from Select... that's probably a terrible choice to prove things out on. I should start with Label or Button or something more trivial.

PR #27098, opened 2024-02-13
fix: try to fix cypress with magic

mistercrunch:
Cypress is crashing our builds on `master`
Screenshot 2024-02-12 at 6 19 44 PM

I'm trying this because it might have worked for someone on the internets
https://github.com/cypress-io/cypress/issues/7204


mistercrunch:
For the record, first try passed, trying another CI build see if we get lucky again

geido (approved):

PR #27108, opened 2024-02-14
chore: Ensure local (tox) Python testing is somewhat akin to remote (GitHub actions)

john-bodley:
### SUMMARY

It _really pains_ me that local and remote CI testing is quite different which results in numerous wasted cycles. A case in point is for major refactors I often run tests locally (which are invoked via `tox`), however when running the following (as part of another PR), 

```
 tox -e py39 -- tests/unit_tests/security/manager_test.py
```

the test would fail—due to a UNIQUE constraint violation—as the unit tests expect that the database schema/records are not defined, though the `tox` `testenv` environment runs the following commands prior to invoking `pytest`,

```
superset db upgrade
superset init
```

and thus one cannot [redefine](https://github.com/apache/superset/blob/master/tests/unit_tests/security/manager_test.py#L247-L263) the `Alpha` role.  Personally—per a line item I've added to the monthly Superset Meetup—I think we shouldn't differentiate between unit and integration tests and ensure that any setup works for any and all tests.

This PR simply ensures that all Python tests (be that local or remote) are invoked via the `python_tests.sh` script. This should work under the condition that you're running either the unit tests or integration tests, but not both.

Finally I've moved the loading of test users to leverage the `load-test-users` CLI command. 

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/scripts/python_tests.sh
+++ b/scripts/python_tests.sh
@@ -26,10 +26,13 @@ export LD_PRELOAD=/lib/x86_64-linux-gnu/libstdc++.so.6
 export SUPERSET_CONFIG=${SUPERSET_CONFIG:-tests.integration_tests.superset_test_config}
 export SUPERSET_TESTENV=true
 echo "Superset config module: $SUPERSET_CONFIG"
-
-superset db upgrade
-superset init
-
 echo "Running tests"
 
-pytest --durations-min=2 --maxfail=1 --cov-report= --cov=superset ./tests/integration_tests "$@"
+if [[ $@ =~ "integration_tests" ]]; then
+    superset db upgrade
+    superset init
+    superset load-test-users
+    pytest --durations-min=2 --maxfail=1 --cov-report= --cov=superset "$@"
+else
+    pytest --durations-min=0.5 --cov-report= --cov=superset "$@"
+fi
--- a/tox.ini
+++ b/tox.ini
@@ -21,11 +21,7 @@
 basepython = python3.9
 ignore_basepython_conflict = true
 commands =
-    superset db upgrade
-    superset init
-    # use -s to be able to use break pointers.
-    # no args or tests/* can be passed as an argument to run all tests
-    pytest -s {posargs}
+    {toxinidir}/scripts/python_tests.sh {posargs}
 deps =
     -rrequirements/testing.txt
 setenv =
@@ -51,6 +47,7 @@ usedevelop = true
 allowlist_externals =
     npm
     pkill
+    {toxinidir}/scripts/python_tests.sh
 
 [testenv:cypress]
 setenv =
--- a/.github/workflows/superset-python-integrationtest.yml
+++ b/.github/workflows/superset-python-integrationtest.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Python integration tests (MySQL)
         if: steps.check.outcome == 'failure'
         run: |
-          ./scripts/python_tests.sh
+          ./scripts/python_tests.sh tests/integration_tests
       - name: Upload code coverage
         if: steps.check.outcome == 'failure'
         run: |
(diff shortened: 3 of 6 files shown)


tox.ini
```diff
@@ -23,6 +23,7 @@ ignore_basepython_conflict = true
 commands =
     superset db upgrade
     superset init
+    superset load-test-users
```
michael-s-molina, line 26: This order is different than the one specified by [Integration Testing](https://github.com/apache/superset/blob/master/CONTRIBUTING.md#integration-testing) in `CONTRIBUTING.md`. 

In the same file we have:

```
# Create default roles and permissions
superset init
```
Could you check if only the roles and permissions are created by `superset init` but not their associations with users? If the associations are also handled by `superset init` we would need to load the test users before that step.

john-bodley: @michael-s-molina my original intent was to follow the order you outlined in the `CONTRIBUTING.md` (which is referenced in a few places), however CI fails with the following, 

```
self = <tests.integration_tests.sqllab_tests.TestSqlLab testMethod=test_query_admin_can_access_all_queries>

    def test_query_admin_can_access_all_queries(self) -> None:
        """
        Test query api with all_query_access perm added to
        Admin and make sure only Admin queries show up. This is the default
        """
        # Test search_queries for Admin user
        self.run_some_queries()
        self.login("admin")
    
        url = "/api/v1/query/"
        data = self.get_json_resp(url)
>       self.assertEqual(3, len(data["result"]))
E       AssertionError: 3 != 2
```

I looked into this previously, but couldn't find a resolution. I suspect this is just one of a magnitude of gremlins we have in the code base due to us not adhering to best practices. Previously the test users used to be loaded as part of the integration tests, i.e., after the `superset init` and I suspect there's likely some weird race condition.

john-bodley: @michael-s-molina my original intent was to follow the order you outlined in the `CONTRIBUTING.md` (which is referenced in a few places), however CI fails with the following, 

```
self = <tests.integration_tests.sqllab_tests.TestSqlLab testMethod=test_query_admin_can_access_all_queries>

    def test_query_admin_can_access_all_queries(self) -> None:
        """
        Test query api with all_query_access perm added to
        Admin and make sure only Admin queries show up. This is the default
        """
        # Test search_queries for Admin user
        self.run_some_queries()
        self.login("admin")
    
        url = "/api/v1/query/"
        data = self.get_json_resp(url)
>       self.assertEqual(3, len(data["result"]))
E       AssertionError: 3 != 2
```

I looked into this previously, but couldn't find a resolution. I suspect this is just one of a magnitude of gremlins we have in the code base due to us not adhering to best practices. Previously the test users used to be loaded as part of the integration tests, i.e., after the `superset init` and I suspect there's likely some weird race condition.

scripts/python_tests.sh
```diff
@@ -26,10 +26,13 @@ export LD_PRELOAD=/lib/x86_64-linux-gnu/libstdc++.so.6
...
 echo "Running tests"
 
-pytest --durations-min=2 --maxfail=1 --cov-report= --cov=superset ./tests/integration_tests "$@"
+if [[ $@ =~ "integration_tests" ]]; then
+    superset db upgrade
+    superset init
+    superset load-test-users
+    pytest --durations-min=2 --maxfail=1 --cov-report= --cov=superset "$@"
+else
+    pytest --durations-min=0.5 --cov-report= --cov=superset "$@"
```
michael-s-molina, line 37: Previously, unit tests also included `./tests/common` for coverage reports.

`pytest --durations-min=0.5 --cov-report= --cov=superset ./tests/common ./tests/unit_tests --cache-clear`

michael-s-molina: There was also a `--cache-clear` which I don't know if it was relevant.

michael-s-molina:
I also noticed that CI didn't run the `Python-Unit / unit-tests` on this PR. Could you confirm if they were skipped correctly or if they should have run?

john-bodley:
@michael-s-molina and @mistercrunch this can probably be closed in favor of https://github.com/apache/superset/pull/29382 if we opt to use `act` rather than `tox` for testing locally.

mistercrunch:
making great progress and loving `act` so far

PR #27112, opened 2024-02-14
fix: gevent upgrade to 23.9.1

dpgaspar:
### SUMMARY

Important fixes for gevent:

https://www.gevent.org/changelog.html#id4 23.9.0
https://www.gevent.org/changelog.html#id2 23.9.1

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


requirements/docker.txt
```diff
@@ -10,7 +10,7 @@
     # via
     #   -r requirements/base.in
     #   -r requirements/docker.in
-gevent==22.10.2
+gevent==23.9.1
```
john-bodley, line 13: If we're bumping versions the bounds should be updated in `setup.py` as this ensures that the correct versions of `gevent` (or similar) will be installed irrespective of how people build Superset.

dpgaspar: This is only included on docker, we have no references of `gevent` on `setup.py`

villebro: Maybe we should add "gevent" as an optional `extras_require`?

dpgaspar: ok, I don't think we interact or will colide with `gevent` in any way since it could be used with `gunicorn`. `gunicorn` does not set any upper boundaries so I'm doing the same

geido (approved):

john-bodley:
@dpgaspar if `gevent` really isn’t a strict dependency then it seems like the constraint should have been added in `docker.in` instead.

PR #24911, opened 2023-08-08
fix(logs): increase json field for logs table

cnabro:
### SUMMARY
- Increase json field for logs table. (Text to MediumText)
- When insert logs while imports a specific dashboard, An error occured.
```
(MySQLdb._exceptions.DataError) (1406, "Data too long for column 'json' at row 1")
[SQL: INSERT INTO logs (`action`, user_id, dashboard_id, slice_id, json, dttm, duration_ms, referrer) VALUES (%s, %s, %s, %s, %s, %s, %s, %s)]
[parameters: ('copy_dash', '1', 9, 0, '{"path": "/superset/copy_dash/9/", "data": "{\\"certified_by\\":\\"\\",\\"certification_details\\":\\"\\",\\"css\\":\\"\\",\\"dashboard_title\\":\\"G ... (171666 characters truncated) ... BS\\"}},\\"filter_scopes\\":\\"{}\\"}", "url_rule": "/superset/copy_dash/<int:dashboard_id>/", "object_ref": "Superset.copy_dash", "dashboard_id": 9}', datetime.datetime(2023, 8, 7, 6, 50, 6, 897685), 6474, '[http://__/superset/dashboard/9/?native_filters_key=A27QtFLKT95Qdd0X-h4KkTLZTBhOslCg0Jshcisx4A8FM71hI_Ae6pBdTc4ac94f')](http://__/superset/dashboard/9/?native_filters_key=A27QtFLKT95Qdd0X-h4KkTLZTBhOslCg0Jshcisx4A8FM71hI_Ae6pBdTc4ac94f%27))]
```

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
```
0769ef90fddd -> 2e826adca42c (head), Fix schema for log
ee179a490af9 -> 0769ef90fddd, Fix schema perm for datasets
e0f6f91c2055 -> ee179a490af9, deckgl-path-width-units
bf646a0c1501 -> e0f6f91c2055, create_user_favorite_table
a23c6f8b1280 -> bf646a0c1501, json_metadata
```

### TESTING INSTRUCTIONS
```sql
ALTER TABLE logs MODIFY json MEDIUMTEXT;
Query OK, 556821 rows affected (5.10 sec)
```
- Estimate 100000 rows updated per 1 sec

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
[...]

PR #24510, opened 2023-06-26
fix(dashboard): increase json_metadata field

cnabro:
### SUMMARY
- Data too long for column `json_metadata`
- Default column size is 65535, we need to increase column size for saving dashboard.
  - ⚒️ Migration DDL required
- position_json already changed to `utils.MediumText`
```
(MySQLdb._exceptions.DataError) (1406, "Data too long for column 'json_metadata' at row 1")
[SQL: UPDATE dashboards SET changed_on=%s, json_metadata=%s, changed_by_fk=%s WHERE dashboards.id = %s]
```

### TESTING INSTRUCTIONS
```sql
ALTER TABLE dashboards MODIFY json_metadata MEDIUMTEXT;
```
- After modifying the column, it works normally.

### ADDITIONAL INFORMATION
- [x] Has associated issue: https://github.com/apache/superset/issues/5532
- [ ] Required feature flags:
- [ ] Changes UI
- [x] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [x] Migration is atomic, supports rollback & is backwards-compatible
  - [x] Confirm DB migration upgrade and downgrade tested
  - [x] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #20936, opened 2022-08-01
chore(db): ensure query_context is MediumText before viz migration

ktmud:
### SUMMARY

Since #20779 will just be an no-op if the columns are already MediumText, let's just do it again before the viz migration. This way, administrators don't have to remember to manually migrate the columns before running `superset db migration`.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N/A

### TESTING INSTRUCTIONS

CI and tested locally via `superset db upgrade`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #5618, opened 2018-08-13
[dashboards] Increasing position_json to MEDIUMTEXT for MySQL

john-bodley:
Most SQLA engines are unbounded for the `TEXT` type however for MySQL it's restricted to ~ 64 kB which can cause overflow for the `dashboards.position_json` column for the V2 layout.

This PR ensures that for MySQL engines the `TEXT` column is changed from `TEXT`(~ 64 kB) to `MEDIUMTEXT` (~ 16 MB). All other engines remained unchanged.

to: @graceguo-supercat @michellethomas @mistercrunch @williaster

PR #27119, opened 2024-02-14
refactor: Updates some database columns to MediumText

michael-s-molina:
### SUMMARY
Similarly to https://github.com/apache/superset/pull/24911, https://github.com/apache/superset/pull/24510, https://github.com/apache/superset/pull/20936, https://github.com/apache/superset/pull/20779, https://github.com/apache/superset/pull/5618, this PR updates the type of some database columns to `MediumText` to accommodate bigger text contents such as JSON, CSS or SQL. We have found some cases where SQL queries shared by users were truncated due to insufficient storage space. You can check the referenced PRs for more context on why the change is important.

### TESTING INSTRUCTIONS
1 - Execute `superset db upgrade` and check the affected tables
2 - Make sure the nullable state is preserved
3 - Execute `superset db downgrade` and check the affected tables

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2024-02-14_14-43_17fcea065655_change_text_to_mediumtext.py
```diff
@@ -0,0 +1,84 @@
...
+    "tab_state.sql",
+    "table_schema.extra_json",
+]
+
+NOT_NULL_COLUMNS = ["keyvalue.value", "row_level_security_filters.clause"]
+
+
+def upgrade():
+    for column in TABLE_COLUMNS:
+        with op.batch_alter_table(column.split(".")[0]) as batch_op:
```
zhaoyongjie, line 67: @michael-s-molina, I encountered a database deadlock error when running the database migration script a few days ago. We are using PostgreSQL as Superset metadata. This change only affects MySQL, so could you please add a condition to detect if MySQL is used as metadata? If not, pass the upgrade and downgrade functions.

john-bodley: @zhaoyongjie are these migrations a no-op if PostgreSQL doesn't differentiate between `TEXT` and `MEDIUMTEXT`?

zhaoyongjie: @john-bodley PG only has TEXT type for string-like column, I'm not sure what should be done under the SQLAlchemy core.

superset/migrations/versions/2024-02-14_14-43_17fcea065655_change_text_to_mediumtext.py
```diff
@@ -0,0 +1,84 @@
...
+"""
+
+# revision identifiers, used by Alembic.
+revision = "17fcea065655"
+down_revision = "87d38ad83218"
+
+import sqlalchemy as sa
+from alembic import op
+
+from superset.utils.core import MediumText
```
mistercrunch, line 32: Had to look this up, but MediumText here is:
```python
def MediumText() -> Variant:  # pylint:disable=invalid-name
    return Text().with_variant(MEDIUMTEXT(), "mysql")
```

mistercrunch (commented):
Can we make sure we only run this DDL for MySQL since MEDIUMTEXT is a MySQL-only construct? Something like `if engine == 'mysql': ` within both upgrade and downgrade (?)

Unless SQLAlchemy is smart enough to understand this is a no-op for Postgres and others(?)

mistercrunch:
Ooops just realized after my review that similar comments had been made :) 

Side note - if that's helpful - we now have a utils module that can expose methods to be reused across migrations (somehow it was tricky to introduce this) -> https://github.com/apache/superset/blob/master/superset/migrations/migration_utils.py

Meaning you can create more complex or engine-specific logic here that we'll be able to re-use in the future.

superset/migrations/versions/2024-02-14_14-43_17fcea065655_change_text_to_mediumtext.py
michael-s-molina: @john-bodley According to ChatGPT:

> No, the alter_column function in SQLAlchemy is not a no-op if the existing type is the same as the new type. When you use the alter_column function in SQLAlchemy to alter a column, it will generate an ALTER TABLE statement that modifies the column definition in the database schema. Even if the existing type is the same as the new type, SQLAlchemy will still execute the ALTER TABLE statement to ensure that the column definition matches the intended type.
This behavior ensures consistency and helps to handle any other modifications or changes associated with the column, such as constraints or defaults.

michael-s-molina:
> so could you please add a condition to detect if MySQL is used as metadata?

> Can we make sure we only run this DDL for MySQL since MEDIUMTEXT is a MySQL-only construct?

@zhaoyongjie @mistercrunch Added the MySQL condition 😉 

> Side note - if that's helpful - we now have a utils module that can expose methods to be reused across migrations

Nice! Thanks for sharing.

zhaoyongjie (approved):
LGTM, Thanks for the upgrading.

mistercrunch (approved):

ghost (approved):

PR #27122, opened 2024-02-15
feat: faster docker builds

mistercrunch:
### SUMMARY

While working on the docker build matrix, I noticed there were opportunities to improve cache hit rate on docker builds, here are some of the ideas, explained inline with the PR

One thing that's getting improved here is around the fact that the python dependencies (`pip install`) are getting re-triggered when `package.json` changes, simply because we use it as the source of truth for both npm and python (setup.py used to read the package.json to figure out which version it is). Here I extract the version in a file instead of COPY package.json

Other notable changes:
- removed `requirements/development.*` from the `lean` docker target, clearly lean shouldn't have a bunch development-related libraries. This may be a breaking change for some. mostly [postgres] being in there, and maybe druid and hive. We *may* want to promote postgres to lean, but unclear... maybe [cors] ?
- removed `requirements/local.*` that had become a pointer to `requirements/development.*`


Dockerfile
```diff
@@ -61,9 +74,7 @@ ENV LANG=C.UTF-8 \
     SUPERSET_HOME="/app/superset_home" \
     SUPERSET_PORT=8088
 
-RUN mkdir -p ${PYTHONPATH} superset/static superset-frontend apache_superset.egg-info requirements \
-    && useradd --user-group -d ${SUPERSET_HOME} -m --no-log-init --shell /bin/bash superset \
```
mistercrunch, line 65: Here I'm thinking some of the funky commands may trigger the apt-get stuff to not use the cache at times. I see a lot of builds re-running `apt-get` when nothing here should have changed.

Dockerfile
```diff
@@ -47,11 +47,24 @@ COPY ./superset-frontend ./
 # This seems to be the most expensive step
 RUN npm run ${BUILD_CMD}
 
+######################################################################
+# Python pre-lean image
```
mistercrunch, line 51: Other options:
- have the release manager bump both `setup.py` and `package.json`, have a CI / build-step that checks that they have been bumped
- have some other mechanism to store the version that get pushed into package.json?
- use a `package.js` insetad of a package.json, that dynamically read some sort of version.txt

I figure this target (pre_lean) could be useful as a place to prepare/cache more python-related stuff eventually. Install more stuff here and COPY into lean

PR #27113, opened 2024-02-14
fix: upgrade cryptography to major 42

dpgaspar:
### SUMMARY

Important fixes for cryptography:

https://cryptography.io/en/latest/changelog/#v42-0-0

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27129, opened 2024-02-15
chore: lower cryptography min version to 41.0.2

sadpandajoe:
### SUMMARY

At Preset we use [`snowflake-connector-python`](https://pypi.org/project/snowflake-connector-python/) and the highest version doesn't support `cryptography==42.0.0` which we bumped in this [PR](https://github.com/apache/superset/pull/27113) . Because of that, we'd need the minimum decreased back to `41.0.2` until it is supported in `snowflake-connector-python`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(
         "colorama",
         "croniter>=0.3.28",
         "cron-descriptor",
+        # snowflake-connector-python as of 3.7.0 doesn't support >=42.* therefore lowering the min to 41.0.2
         "cryptography>=41.0.2, <43.0.0",
         "deprecation>=2.1.0, <2.2.0",
         "flask>=2.2.5, <3.0.0",
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
         "colorama",
         "croniter>=0.3.28",
         "cron-descriptor",
-        "cryptography>=42.0.0, <43.0.0",
+        "cryptography>=41.0.2, <43.0.0",
         "deprecation>=2.1.0, <2.2.0",
         "flask>=2.2.5, <3.0.0",
         "flask-appbuilder>=4.3.11, <5.0.0",


john-bodley:
Note whenever we change the constraints ideally one would recompile the set of frozen dependencies used for CI, etc. in this case we’re actually relaxing the eligible versions and thus this change _should_ be ok.

mistercrunch (changes requested):
I looked up the frozen dependency here -> https://github.com/apache/superset/blob/master/requirements/base.txt#L81, things seem fine overall. 

Being less strict with the dependency ranges means more heterogenous deployments, and more confusion in the community. Clearly someone else out there is using [snowflake-connector-python](https://pypi.org/project/snowflake-connector-python/) with Superset and will walk into a similar trap at some point.

Also note that `pip-compile-multi` is still broken and the fix still held up by oddly failing tests here -> https://github.com/apache/superset/pull/26944

All that being said, I think it's reasonable to patch this in this  way, though when we to open ranges in `setup.py` or `requirements/*.in`, I think it's good to add a note about what's holding us back. @sadpandajoe can you add a comment above that line: 
```
# snowflake-connector-python as of 3.7.0 doesn't support >=42.*
```

mistercrunch (approved):

john-bodley:
Note in https://github.com/apache/superset/pull/27113 @dpgaspar bumped `cryptography` and thus relaxing the lower bound could be problematic.

sadpandajoe:
Looks like snowflake-connector-python will support cryptography 42.x starting with their 3.7.1 release which is still TBD according to this: https://github.com/snowflakedb/snowflake-connector-python/blob/1e50704713ffd05b04543fb8832d00ab1a6bb1a8/DESCRIPTION.md

dpgaspar (approved):
Unfortunately the most recent version of snowflake connector is still under cryptography 42. I agree with @mistercrunch point, the issue is patched on the requirements so our next release image will be patched also. It should be ok to lax setup.py and keep the deps patched on the requirements

PR #27130, opened 2024-02-15
fix: ID param for DELETE ssh_tunnel endpoint

geido:
### SUMMARY
The DELETE `database/{id}/ssh_tunnel `endpoint was incorrectly expecting an SSH Tunnel ID rather than the Database ID. This PR fixes it.

It deprecates the endpoint as the ssh_tunnel should be a property of the database and should be updated according to the database update endpoint.

N.A.

### TESTING INSTRUCTIONS
- [ ] CI should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/databases/api.py
```diff
@@ -1447,6 +1447,7 @@ def validate_parameters(self) -> FlaskResponse:
     @expose("/<int:pk>/ssh_tunnel/", methods=("DELETE",))
     @protect()
     @statsd_metrics
+    @deprecated(deprecated_in="5.0")
```
michael-s-molina, line 1450:
```suggestion
    @deprecated(deprecated_in="4.0")
```

michael-s-molina (approved):
LGTM

superset/databases/api.py
```diff
@@ -1447,6 +1447,7 @@ def validate_parameters(self) -> FlaskResponse:
     @expose("/<int:pk>/ssh_tunnel/", methods=("DELETE",))
     @protect()
     @statsd_metrics
+    @deprecated(deprecated_in="4.0")
```
eschutho, line 1450: @geido doesn't this mean that we'll be removing this endpoint in 4.0?

michael-s-molina: No. It means that the endpoint was deprecated in 4.0 to be removed in a future version. Possibly 5.0.

geido: Yeah the idea is indeed to remove it in a future version

PR #27136, opened 2024-02-16
chore(tests): Remove unnecessary explicit Flask-SQLAlchemy session expunges

john-bodley:
### SUMMARY

There's a few instances within our tests where we explicitly expunge either specific or all objects from the session, however this is unnecessary as the Flask-SQLAlchemy session is torn down after each test (verified via logging) in which the [current session is removed/closed](https://github.com/pallets-eco/flask-sqlalchemy/blob/42a36a3cb604fd39d81d00b54ab3988bbd0ad184/src/flask_sqlalchemy/extension.py#L448) and thus there's no need to explicitly expunge the records.

Note this change is a precursor to defining a "unit of work" as part of [SIP-99B](https://github.com/apache/superset/issues/25108). I've been struggling to get the various unit/integration tests to pass in my PoC—in part due to atypical testing patterns.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/base_tests.py
+++ b/tests/integration_tests/base_tests.py
@@ -199,11 +199,8 @@ class SupersetTestCase(TestCase):
     def login(self, username="admin", password="general"):
         return login(self.client, username, password)
 
-    def get_slice(self, slice_name: str, expunge_from_session: bool = True) -> Slice:
-        slc = db.session.query(Slice).filter_by(slice_name=slice_name).one()
-        if expunge_from_session:
-            db.session.expunge_all()
-        return slc
+    def get_slice(self, slice_name: str) -> Slice:
+        return db.session.query(Slice).filter_by(slice_name=slice_name).one()
 
     @staticmethod
     def get_table(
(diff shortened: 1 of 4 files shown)


tests/integration_tests/base_tests.py
```diff
@@ -199,11 +199,8 @@ def get_or_create(self, cls, criteria, **kwargs):
     def login(self, username="admin", password="general"):
         return login(self.client, username, password)
 
-    def get_slice(self, slice_name: str, expunge_from_session: bool = True) -> Slice:
-        slc = db.session.query(Slice).filter_by(slice_name=slice_name).one()
-        if expunge_from_session:
-            db.session.expunge_all()
```
john-bodley, line 205: It seems weird that we were expunging all objects from the session and not just the chart in question.

michael-s-molina (approved):
LGTM

PR #27142, opened 2024-02-16
fix(chart): apply number format in Box Plot tooltip only where necessary

### LINKED ISSUE #25327: Data Formatting is applied everywhere in box plot tooltips where it should not
data number format is applied every where in box plot tooltips where it should not:
   number of sample and number of outlier should have a specific adaptative formatting related to their meaning 

#### How to reproduce the bug
create box plot chart with data of type duration.
Set the number format to duration 
Display the chart and tool tips 
Number of sample and number of outlier are displayed with the wrong unit ( and are potentially converted in something unreadable) 

### Expected results
Have the number of sample and number of outlier using an adaptative 

### Actual results

number of sample and number of outlier are displayed as duration

#### Screenshots

image

### Environment

browser type and version: chrome version 117.0.5938.63
superset version: 2.1,3.0RC4
python version: from docker images
node.js version: from docker images
any feature flags active: {"ALERT_REPORTS": True,"EMBEDDED_SUPERSET": True,"DASHBOARD_CROSS_FILTERS":True,"DRILL_BY":True,"DRILL_TO_DETAIL":True, "DASHBOARD_RBAC" : True,"ENABLE_TEMPLATE_PROCESSING" : True}

goto-loop:
### SUMMARY

Fixes #25327.

Currently, the selected number format gets applied to all lines of the box plot tooltip. For example, if you are displaying durations (number format "Duration in ms") and you have 3544 observations, the tooltip will say "# Observations: 3.5s".

This PR removes the formatting for values which are by definition without unit, i.e. "# Observations" and "# Outliers".

I am aware that it would be better to skip the formatting based on the selected format, because formats like ",d" and ".1s" can still be applied. I would argue though that no formatting still beats the current wrong formatting.

image

image

### TESTING INSTRUCTIONS

Open one of the Box plots that get loaded with the example database and change the number format to either duration or currency. Then hover over one of the boxes to display the tooltip.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #25327
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Sorry this slid under the radar! Adding more reviewers in hopes that we can get it across the finish line and close the linked issue.

rusackas:
closing/reopening to kickstart CI.

rusackas (approved):
I think the change looks reasonable... the number of outliers/observations seems fine as an integer (unformatted) since that's seemingly always the case.

PR #27144, opened 2024-02-16
fix: Forced refresh async queries are executed synchronously

### LINKED ISSUE #24889: Forced refresh async queries are executed synchronously
When chart data is forced to refresh in chart view, query is executed in synchronous manner, even when `GLOBAL_ASYNC_QUERIES=True`.

#### How to reproduce the bug

1. Go to 'Chart'
2. Click on 'Cached' to force refresh 

### Expected results

Superset should:
- request chart data (`POST /api/v1/chart/data?form_data=<fomr_data>`)
- poll for results (`GET /api/v1/async_event/?last_id=<last_id>`)
- get results  (`GET /api/v1/chart/data/<cache_key>`)

### Actual results

Superset sends `POST /api/v1/chart/data` and waits for that request to return results.

### Environment

- browser type and version: Chromium
- superset version: `2.1.0`
- python version: `3.8.13`
- node.js version: `N/A`
- any feature flags active: `GLOBAL_ASYNC_QUERIES=True`

### Checklist

### Additional context

The issue seems to be in `QueryCacheManager`. When query is forced it returns [query_cache object early](https://github.com/apache/superset/blob/554ef07eeda094f5bcc081e7a265e0b7408e0868/superset/common/utils/query_cache_manager.py#L149) and ignores [forced_cached flag check](https://github.com/apache/superset/blob/554ef07eeda094f5bcc081e7a265e0b7408e0868/superset/common/utils/query_cache_manager.py#L188). In result, it ends getting [query results immediately](https://github.com/apache/superset/blob/554ef07eeda094f5bcc081e7a265e0b7408e0868/superset/common/query_context_processor.py#L157) rather than throwing exception and getting results [asynchronously](https://github.com/apache/superset/blob/554ef07eeda094f5bcc081e7a265e0b7408e0868/superset/charts/data/api.py#L345).

I also suspect that when query is forced, there's not need to check cache at all.

GAlexeyV:
### SUMMARY

Deleted a cache check for async queries when cache is forced.
Attached screenshots (please, see below) to demonstrate logs that show expected behaviour after the change.

Fixes https://github.com/apache/superset/issues/24889

Context:

When chart data is forced to refresh in chart view, query is executed in synchronous manner, even when GLOBAL_ASYNC_QUERIES=True.

Expected results
Superset should:
1. Request chart data (POST /api/v1/chart/data?form_data=<fomr_data>)
2. Poll for results (GET /api/v1/async_event/?last_id=<last_id>)
3. Get results (GET /api/v1/chart/data/<cache_key>)

Actual results:
Superset sends POST /api/v1/chart/data and waits for that request to return results.

Screenshot 2024-02-23 at 13 26 05

Screenshot 2024-02-23 at 13 28 51

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


john-bodley:
@zhaoyongjie would you mind reviewing this? It seems, per [here](https://github.com/apache/superset/commit/32d2aa0c404f47abaa270bf1c572fdd26feb16ed), that you may have authored this logic.

zhaoyongjie:
> @zhaoyongjie would you mind reviewing this? It seems, per [here](https://github.com/apache/superset/commit/32d2aa0c404f47abaa270bf1c572fdd26feb16ed), that you may have authored this logic.

Let me catch up the change....thanks for the mention.

superset/common/utils/query_cache_manager.py
```diff
@@ -145,7 +145,7 @@ def get(
         Initialize QueryCacheManager by query-cache key
         """
         query_cache = cls()
-        if not key or not _cache[region] or force_query:
+        if not key or not _cache[region] or not force_query:
```
zhaoyongjie, line 148: This change means that if **force_query is false**, then generate a new query_cache object.....after this change, we can't use cached data any more.

GAlexeyV: @zhaoyongjie hi,
Thank you for having a look.
Likely I am missing something, to clarify: we will use cached data in all cases which satisfy above "if statement". The only change made that we will not use cached data when a user forces a query (both for synch and async queries). Is not it what we want? 

E.g. when a user forces a query, data should not be extracted from a cache, but from the source (e.g. a database) and a new cache object created (based on forced query to a source). I understand that it will increase load on a source but otherwise we are loosing the sense of using "force query" and this functionality unlikely should be abused so load on source should not change significantly.

zhaoyongjie: @GAlexeyV, the default value of force_query is False(the frontend side also emit **false** as force_query), and the **not force_query** condition in **or** expression means that all of default queries(except true as force_query) emit to Superset should create a new "query_cache" object --- please don't use cached data. It is an opposite behavior to we want, what is the expected behavior before? 1. if the data cached 2. and the force_query is False 3. then the query should been responded by the Cache.

You could set a stop debugger at line 149 to verify this situation.

GAlexeyV: @zhaoyongjie, thank you for explaining. Struggling to understand the codebase. Submitted a different change: removed a check for cached data for async queries when cache is forced.

GAlexeyV: @zhaoyongjie hi,
Any chance you can have a look?

GAlexeyV: @john-bodley hi,
Maybe you can have a look?
Will update tests if you ok with the change

zhaoyongjie: @GAlexeyV I'll check out this part of code this week as well, but for now the CI seems broken.

GAlexeyV: @zhaoyongjie , got it, thank you
Yes, I have not updated tests as firstly wanted to agree on change

zhaoyongjie: @GAlexeyV Sorry for the waiting, I'll review this tonight(CET timeone)

GAlexeyV: Cool, thanks

superset/charts/data/api.py
```diff
@@ -327,12 +327,7 @@ def _run_async(
         """
         Execute command as an async query.
         """
-        # First, look for the chart query results in the cache.
-        with contextlib.suppress(ChartDataCacheLoadError):
-            result = command.run(force_cached=True)
```
zhaoyongjie, line 332: I believe that this line should refine and the bug lies in this line. This part of codes want to imminently response a *cached results* so do you mind to find a way to check if the cache key in the cache? We have [has](https://github.com/apache/superset/blob/a4c771e013957e95d80d252dcdedad7046348964/superset/common/utils/query_cache_manager.py#L214) method in QueryCacheManager that might help to this function.

superset/common/utils/query_cache_manager.py
zhaoyongjie: @GAlexeyV I put some thoughts in your codes, thanks a lot for the fix. ping me if you need supports.

rusackas:
Thanks for the review @zhaoyongjie - would love to get this merged so we can close https://github.com/apache/superset/issues/24889, which has been open for a long time.

superset/charts/data/api.py
GAlexeyV: @zhaoyongjie ,
not sure understand what you meant

Here an example:
`    def _run_async(
        self, form_data: dict[str, Any], command: ChartDataCommand
    ) -> Response:
        """
        Execute command as an async query.
        """
        # First, look for the chart query results in the cache.
        with contextlib.suppress(ChartDataCacheLoadError):
            # Check if the cache key exists in the cache
            if QueryCacheManager.has(form_data.get("cache_key"), region=CacheRegion.DEFAULT):
                result = command.run(force_cached=True)
                if result is not None:
                    return self._send_chart_response(result)

        # Otherwise, kick off a background job to run the chart query.`
        
       Did not get where from I am supposed to get cache_key and region parameters to pass them to QueryCacheManager

GAlexeyV: Did not get where from I am supposed to get cache_key and region parameters to pass them to QueryCacheManager

superset/common/utils/query_cache_manager.py
GAlexeyV: @zhaoyongjie , replied to a comment

superset/charts/data/api.py
zhaoyongjie: @GAlexeyV will check the details codes and reply you tomorrow.

GAlexeyV: @zhaoyongjie , thank you
I will finish it then when return back from holidays (in 2 weeks)

zhaoyongjie: Enjoy your vacation! I'll check the codes and then reply you.

GAlexeyV: @zhaoyongjie hi, 
Did you have a chance to look into?

swaresh: I'm also looking ahead for this change to get merged.

GAlexeyV: @john-bodley hi, maybe you can have a look?

harshit2283:
Till we figure out a better solve
Moved the cache lookup behind an if statement checking the value of `force`

rusackas:
This has been stuck for quite some time on failing CI. I'll convert it to draft, but would love to see it brought up to passing so we can revisit/merge, and close out the issue :)

rusackas:
Closing/reopening to kick-start CI, and see what it'll take to get this across the line.

rusackas:
Closing this very old draft PR (nearly 2 years old). There was extensive discussion with @zhaoyongjie but the PR has been stale for several months now.

@GAlexeyV if you're still interested in fixing this async queries issue, please feel free to open a fresh PR based on current `master`. The codebase has changed significantly since this was opened.

Thanks for your contribution!

PR #27145, opened 2024-02-16
refactor(plugins): Time Comparison Utils

Antonio-RiveroMartnez:
### SUMMARY
Previously we used to build the time range for time comparison in the frontend, this logic was duplicating what the backend already handles for us, so we are moving away from origin implementation and instead leverage the existing APIs that build queryObjects to handle time_shift.

Also, In order to DRY time comparison logic when implementing the logic into other viz plugins, we are extracting out the code that was before part of BigNumbers with time comparison plugin.

The new BigNumber with Time Comparison viz had no tests, so this PR is also including some for the new utils and changes made.

### TESTING INSTRUCTIONS
The BigNumber with Time Comparison chart should behave exactly the same as it is currently.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


zhaoyongjie:
@Antonio-RiveroMartnez Thanks for introduce the time comparison in Frontend codebase. Unfortunately, I don't think the time comparison should move into frontend, because there is a time comparison in Advanced Analytics that should build a separated query not only calculate the data in frontend. I remember that the BigNumber w/ trend line was supported Rolling and Resample but didn't support time shift, IMO, we should support time shift rather than move the time comparison into frontend.

image

Antonio-RiveroMartnez:
> @Antonio-RiveroMartnez Thanks for introduce the time comparison in Frontend codebase. Unfortunately, I don't think the time comparison should move into frontend, because there is a time comparison in Advanced Analytics that should build a separated query not only calculate the data in frontend. I remember that the BigNumber w/ trend line was supported Rolling and Resample but didn't support time shift, IMO, we should support time shift rather than move the time comparison into frontend.
> 
> ## Advanced Analytics in Line chart
> image
> ## Advanced Analytics in BigNumber w/ trend line
> image

Hey @zhaoyongjie ! Thanks for looking into this and pointing me to the `Advanced Analytics` section.  Just want to mention that this code is experimental and behind a `Feature Flag` and the main goal is to introduce some more tools on different visualizations down the road.

Now, regarding the time comparison, the frontend code is building the query that gets send when the appropriate time range is selected, however, I see your point on leveraging current `Time Shift` option since that's pretty much doing that same work.

What might be different and one of the main drivers of putting this "range calculation" on the frontend are:

1. The fact that the control for the existing `Time Shift` and the proposed one are different in how the operate, since the new one has an inherit option that would be complicate to accomplish using just the input text from the current implementation
1. We want to show the range to the user in different places in the UI, if we want to rely only on the backend for retrieving such info, we would have to create some API or similar and all the other components to support that since just the query is not very helpful to users, so it was done to keep simplicity of the solution.

All that being said, I am going to try approaching this using the `Time shift` you mentioned at least for the query construction, and will explore ways on getting the info to the UI without "translating" the range calculation in the frontend.

Again, appreciated your input and any other is more than welcome! Thanks so much.

Antonio-RiveroMartnez:
Hey @zhaoyongjie !  I refactored the solution based on your suggestion so we're no longer computing `Time comparison ranges` etc in the frontend. Instead we're now leveraging the existing backend and the `time_shift`. Also, introduced a minor fixed time_shift option for the `inherited` ranges.

Regarding using the `Time Shift` control from `Advanced Analytics`,  I opted not to do so and keep using our new experimental control because the main goal is to generate automatic time comparison based on the control's value and not manually constructing the comparison queries with time_offsets. Also, as I mentioned before, the new control has some options that are not available in the existing `Time Shift` so we are trying to keep this experimental feature isolated and behind the feature flag as much as possible.

Let me know if you want to discuss anything else further, thanks again for the great input 🎉

zhaoyongjie:
> Hey @zhaoyongjie ! I refactored the solution based on your suggestion so we're no longer computing `Time comparison ranges` etc in the frontend. Instead we're now leveraging the existing backend and the `time_shift`. Also, introduced a minor fixed time_shift option for the `inherited` ranges.
> 
> Regarding using the `Time Shift` control from `Advanced Analytics`, I opted not to do so and keep using our new experimental control because the main goal is to generate automatic time comparison based on the control's value and not manually constructing the comparison queries with time_offsets. Also, as I mentioned before, the new control has some options that are not available in the existing `Time Shift` so we are trying to keep this experimental feature isolated and behind the feature flag as much as possible.
> 
> Let me know if you want to discuss anything else further, thanks again for the great input 🎉

I'll review this PR tonight and thanks for your follow up changes.

zhaoyongjie:
> Hey @zhaoyongjie ! I refactored the solution based on your suggestion so we're no longer computing `Time comparison ranges` etc in the frontend. Instead we're now leveraging the existing backend and the `time_shift`. Also, introduced a minor fixed time_shift option for the `inherited` ranges.

We don't need to modify the `time_shift` parameters in the `query_object`. In fact, the `time_shift` within the `query_object` or `form_data` should be considered as a legacy configuration from the old time picker(about two years ago). Currently, the DATEADD function within the `time_range` field is used for time delta calculation.

image



> 
> Regarding using the `Time Shift` control from `Advanced Analytics`, I opted not to do so and keep using our new experimental control because the main goal is to generate automatic time comparison based on the control's value and not manually constructing the comparison queries with time_offsets. Also, as I mentioned before, the new control has some options that are not available in the existing `Time Shift` so we are trying to keep this experimental feature isolated and behind the feature flag as much as possible.
> 
> Let me know if you want to discuss anything else further, thanks again for the great input 🎉

IMO, we should ensure what the exact requirement is for the concept of time shift in a diagram. For example, if we have a query generated by the Explorer based on the time range 2024-02-01 <= date < 2024-02-20, with a time grain of Days, I want to compare it with "1 year ago," "1 month ago," and "1 week ago." We should send three different time range queries, merge the results together, and then return them to the frontend.

Antonio-RiveroMartnez:
> > Hey @zhaoyongjie ! I refactored the solution based on your suggestion so we're no longer computing `Time comparison ranges` etc in the frontend. Instead we're now leveraging the existing backend and the `time_shift`. Also, introduced a minor fixed time_shift option for the `inherited` ranges.
> 
> We don't need to modify the `time_shift` parameters in the `query_object`. In fact, the `time_shift` within the `query_object` or `form_data` should be considered as a legacy configuration from the old time picker(about two years ago). Currently, the DATEADD function within the `time_range` field is used for time delta calculation.
> 
> image
> > Regarding using the `Time Shift` control from `Advanced Analytics`, I opted not to do so and keep using our new experimental control because the main goal is to generate automatic time comparison based on the control's value and not manually constructing the comparison queries with time_offsets. Also, as I mentioned before, the new control has some options that are not available in the existing `Time Shift` so we are trying to keep this experimental feature isolated and behind the feature flag as much as possible.
> > Let me know if you want to discuss anything else further, thanks again for the great input 🎉
> 
> IMO, we should ensure what the exact requirement is for the concept of time shift in a diagram. For example, if we have a query generated by the Explorer based on the time range 2024-02-01 <= date < 2024-02-20, with a time grain of Days, I want to compare it with "1 year ago," "1 month ago," and "1 week ago." We should send three different time range queries, merge the results together, and then return them to the frontend.
> 
> BTW, thank you very much for tackling this part of the code. I completely understand that there are many technical debates in this area, and I don't want to offend you.

No offense taken, all input is welcome. Now, the use of `time_shift` is for the prefixed options in the control, users can still use a `Custom` range and use `DATEADD`, this control is just giving more `automatic` and `prefixed` ranges to compare without the need to worrying about extra queries setup.

Screenshot 2024-02-20 at 19 56 07


About supporting multiple time ranges as you mentioned, it's something we might consider for future iterations, right now it's not intended to be supported, meaning there's going to be just one time comparison period since the idea is once we make some base utils and experiment with the new control in other visualizations first.

zhaoyongjie:
> No offense taken, all input is welcome. Now, the use of `time_shift` is for the prefixed options in the control, users can still use a `Custom` range and use `DATEADD`, this control is just giving more `automatic` and `prefixed` ranges to compare without the need to worrying about extra queries setup.

The `time_shift` looks like redundant, it's too much specific use case.

> About supporting multiple time ranges as you mentioned, it's something we might consider for future iterations, right now it's not intended to be supported, meaning there's going to be just one time comparison period since the idea is once we make some base utils and experiment with the new control in other visualizations first.


No only for multiple time range, but also concern time unit and time calculate. Let me give you an example.

- time range "2023-02-01 : 2023-02-29", notice that 2023 is a leap year
- time grain: week
- TIME SHIFT in AA: 1 years ago

Currently, Superset will generate following SQLs




### The main query 

```SQL
SELECT DATE_TRUNC('WEEK', "DATE") AS "DATE",
       COUNT(*) AS "count"
FROM cinema_analytics_v2.cinema_page_view_model
WHERE "DATE" >= CAST('2024-02-01T00:00:00.000000' AS DATETIME)
  AND "DATE" < CAST('2024-02-29T00:00:00.000000' AS DATETIME)
GROUP BY DATE_TRUNC('WEEK', "DATE")
ORDER BY "count" DESC
LIMIT 10000
;
```

image


### 1 year ago query
```SQL
SELECT DATE_TRUNC('WEEK', "DATE") AS "DATE",
       COUNT(*) AS "count"
FROM cinema_analytics_v2.cinema_page_view_model
WHERE "DATE" >= CAST('2023-02-01T00:00:00.000000' AS DATETIME)
  AND "DATE" < CAST('2023-02-28T00:00:00.000000' AS DATETIME)
GROUP BY DATE_TRUNC('WEEK', "DATE")
ORDER BY "count" DESC
LIMIT 10000;
```

image



### The merged data frame in Explore page

image


We have to align "2024-1-29" wit "2023-1-30", and etc. Currently, this part of code was computed in Pandas in backend.

zhaoyongjie:
btw, I also believe that we should establish a consistent design for this system and "retire" the legacy field in the `query_object` as soon as possible. We are currently resorting to hacks to resolve some common use cases, which is not ideal.

Antonio-RiveroMartnez:
> > No offense taken, all input is welcome. Now, the use of `time_shift` is for the prefixed options in the control, users can still use a `Custom` range and use `DATEADD`, this control is just giving more `automatic` and `prefixed` ranges to compare without the need to worrying about extra queries setup.
> 
> The `time_shift` looks like redundant, it's too much specific use case.
> 

I could try using `dateadd`, however, for the `inherited` time ranges it fell short, isn't it? But let me see what I can do in the backend to compute them.

Antonio-RiveroMartnez:
> 
> The time_offsets is designed for this use case in query_object.
> 

Yup, just that `time_offsets` for what I can see is attached to other logic like specifying an `X Axis` column, `Time Grain` etc which again, seems something we could introduce later down the road when we allow more powerful control over the time comparison being built vs this "predefined" options/limited time comparison (Mental note we probably should name the components `<something>TimeComparison`)

So, will give it the `DATEADD` a go vs `time_shift` to compute the ranges in the backend (Probably would need something for the `inherited`) and hopefully that would ease the path to further iterations when a more powerful version of the control make it out of experimental.

Will ping you when the PR is updated to discuss further and if you see anything else that we might consider please fell free to comment, truly appreciated your help with this 🌮

zhaoyongjie:
> 
> So, will give it the `DATEADD` a go vs `time_shift` to compute the ranges in the backend (Probably would need something for the `inherited`) and hopefully that would ease the path to further iterations when a more powerful version of the control make it out of experimental.

So, what is "inherited"? Could you explain the design proposal?

Antonio-RiveroMartnez:
> > So, will give it the `DATEADD` a go vs `time_shift` to compute the ranges in the backend (Probably would need something for the `inherited`) and hopefully that would ease the path to further iterations when a more powerful version of the control make it out of experimental.
> 
> So, what is "inherited"? Could you explain the design proposal?

# New Time Comparison control



The main goal of this new control is to offer instant time comparison in some visualizations.

## What do we mean by instant time comparison and why offering it?

Right now when using the existing time comparison function, both the period that the user wants to review, as well as the period prior, have to be selected as part of the date range and care needs to be taken to ensure that the correct date ranges and time grains have been selected. Additionally, it also leads to issues with charts that don’t use the time comparison function, as they are now totalling both the period to be reviewed as well as the comparison period.

Due to this, we're introducing the instant time comparison as in the user doesn't need to build an specific query for the time comparison and deal with time ranges and grains.

We're supporting this in two ways:

1. Using new viz plugins entirely, like the **BigNumber with Time Comparison** one
2. Injecting a new control in the control panel that existing visualizations can leverage

Both ways would make use of the utils methods implemented in this Pull Request and both ways should be made available only behind the experimental feature flag.

## What options does this instant time comparison have and what do they do?

**Inherited time range:** The time comparison range is driven by the timedelta in days within the time range of the selected filter.
**Year:** Same period last Year
**Month:** Same period last Month
**Week:** Same period last week
**Custom:** User enters a time range to use as comparison using the TimeFilter modal

## Conclusion

We are not aiming to replace current time comparison function, we're just offering more usable and static time comparisons to users and presenting the compared data in a more explicit way to users. Advanced users would still make use of the current time comparison function.

Antonio-RiveroMartnez:
> > > No offense taken, all input is welcome. Now, the use of `time_shift` is for the prefixed options in the control, users can still use a `Custom` range and use `DATEADD`, this control is just giving more `automatic` and `prefixed` ranges to compare without the need to worrying about extra queries setup.
> > 
> > 
> > The `time_shift` looks like redundant, it's too much specific use case.
> 
> I could try using `dateadd`, however, for the `inherited` time ranges it fell short, isn't it? But let me see what I can do in the backend to compute them.

Using `DATEADD` when sending the queries doesn't work bacause it leads to miscalculations of the datetime. For example:

**Time filter comparator**: `last year : today`
**Time offset**: 1 year
in order to send the query with `DATEADD` you would need something like `DATEADD(DATETIME('last year'), -1, year)` and `DATEADD(DATETIME('ltoday'), -1, year)`

However, the moment you use `DATETIME('last year')` it miscalculates the datetime which gives you a different result from the original `last year`.

And trying to parse the possibles ways to still send the comparator with `DATEADD` might lead to fall under the same mistake form before of duplicating logic in the frontend.

This is driving the decision to instead compute the `DATEADD` function in the backend using the value from the new control as reference and sending this as a new optional field in the QueryObject.

With this approach we:
1. don't duplicate logic in frontend and we're in total control over when the `DATEADD` is applied
2. Do not make use of legacy `time_shift`
3. We don't make use of `time_offsets` which again is tied to time grains, x axis columns etc

zhaoyongjie:
> 
> With this approach we:
> 
> 1. don't duplicate logic in frontend and we're in total control over when the `DATEADD` is applied
> 2. Do not make use of legacy `time_shift`
> 3. We don't make use of `time_offsets` which again is tied to time grains, x axis columns etc

Sorry,..... I didn't quite catch that since lack of full context, but I think this is an experimental plugin and hidden in the FF, all the frontend changes should be safe.

If we simply avoid using legacy config and avoid from introducing too much complexity on the backend, this PR should be fine. 

Thanks again for the explanation.

zhaoyongjie:
> in order to send the query with `DATEADD` you would need something like `DATEADD(DATETIME('last year'), -1, year)` and `DATEADD(DATETIME('today'), -1, year)`
> 
> However, the moment you use `DATETIME('last year')` it miscalculates the datetime which gives you a different result from the original `last year`.

A hint:  `DATETIME('last year')` means that _relative to today in **the last year**_, and then `DATEADD(DATETIME('last year'), -1, year)` means _relative to today in **the last year of the last year****_

Antonio-RiveroMartnez:
PR updated. As mentioned before:

1. Not using legacy field
2. Computing shift in the backend
3. Computing shift with `Dateadd`
4. Putting everything behind the FF

Thanks.

kgabryje (approved):
Tested, looks great! Code also looks good to me, assuming Yongjie's concerns were addressed

superset/utils/date_parser.py
```diff
@@ -263,6 +264,47 @@ def get_since_until(  # pylint: disable=too-many-arguments,too-many-locals,too-m
...
+            time_unit = ""
+            delta_in_days = None
+            if instant_time_comparison_range == InstantTimeComparison.YEAR:
+                time_unit = "YEAR"
+            elif instant_time_comparison_range == InstantTimeComparison.MONTH:
+                time_unit = "MONTH"
+            elif instant_time_comparison_range == InstantTimeComparison.WEEK:
+                time_unit = "WEEK"
+            elif instant_time_comparison_range == InstantTimeComparison.INHERITED:
+                delta_in_days = (_until - _since).days if _since and _until else None
```
zhaoyongjie, line 284: I just understood what is "inherited" meaning, which is date time difference between since and until. So, for more general use we might create a new function in our custom date time expression e.g. `DATEDIFF([since], [until], [unit])`. 

Then we could generate a expression like below:

`DATEADD(DATETIME(...), -DATEDIFF(..., ..., DAY), DAY) : DATEADD(DATETIME(...), -DATEDIFF(..., ..., DAY), DAY)`

After that, we can move all of the backend hard-code into frontend.

zhaoyongjie: BTW, the PR is totally work for me, we could improve in the follow-up PR>

Antonio-RiveroMartnez:
Hey! I'm merging the PR as is and working on a follow up with `DATEDIFF` when using this in other viz 🙏 thanks both for the review and specially @zhaoyongjie for the amazing input.

superset/utils/date_parser.py
Antonio-RiveroMartnez: About doing the expression in the frontend, I still consider it's best to do it in the backend because the backend processes the right `datetime` no matter what value the filter has.

If the frontend wants to calculate the right `datetime` it will fail because as stated before, an expression that for example has the filter with value: `last year` will get an invalid `datetime` if you wrap it with `DATETIME` and making the frontend calculate the `"right" datetime` would lead to duplicating logic that already works well in the backend.

You could for example do a quick test of setting a filter's value to: `last year` vs `DATETIME('last year')` and see what I'm saying about the difference in both ways.

TLDR; Yes, we could introduce a `DATEDIFF` but IMO we should keep using the backend for building these ranges.

zhaoyongjie: There are some legacy logics and concepts from the initial Superset; let me explain them before, and then you might get more context of the time picker and the original design of the time_range.

First of all, the time_range has two forms:

1. plain text, e.g. `time_range = "last year"`
2. plain text with colon and spaces. e.g. `time_range = " 2021-03-01 : dateadd(datetime("tomorrow"), 1, day)"`

The plain text form is applied in "Last" and "Previous" Frame in the Time Picker, these frame will generate phrases like 'last year', 'last month', 'previous calendar month', using natural English languages to define a time period; The plain text with colon and spaces is more straightforward, it directly define a `since : until` time period.

Notice that these two forms of time_range will convert to the uniform "since : until" in the backend,
- if `Last / Next` in the beginning of the [time_range](https://github.com/apache/superset/blob/54037972f2f56ec86a613e6684e8bbe20cafcf50/superset/utils/date_parser.py#L185-L189), the compiled the expression will be [here](https://github.com/apache/superset/blob/54037972f2f56ec86a613e6684e8bbe20cafcf50/superset/utils/date_parser.py#L210-L219)
- if `Previous` in the beginning of the time_range, the compiled the expression will be [here](https://github.com/apache/superset/blob/54037972f2f56ec86a613e6684e8bbe20cafcf50/superset/utils/date_parser.py#L191-L208).

From the [codes](https://github.com/apache/superset/blob/54037972f2f56ec86a613e6684e8bbe20cafcf50/superset/utils/date_parser.py#L214), 
- If you choose`Last year` in the Last frame in Time Picker, the compiled expression should be: `DATEADD(DATETIME('today'), -1, year) : `, means that relative today in the last year.
- If you input `datetime('last year')` in the custom Frame, means that the first day of the last year, it's not equal to `Last year` choose in the Last frame.

So if you want to generate a time shift in the frontend, you need to 
1. implement DATEDIFF function in the [datetime_parser](https://github.com/apache/superset/blob/54037972f2f56ec86a613e6684e8bbe20cafcf50/superset/utils/date_parser.py#L440), in order to calculate time delta.
2. use `DATEADD(DATEADD(DATETIME('today'), -1, year), -1, year) : ` if you want to generate a relative date of `Last year` time_range.

In summary, we don't need to hard code in get_since_until function, we just need to find out what tools is existing.

Why I said so much in this PR since I found this project we've introduced lots of inconsistent design and hard code anywhere for some temporally request.

sadpandajoe:
🏷️ preset:2024.8

Issue #26997, opened 2024-02-02
ModuleNotFoundError: No module named 'sqlglot' error across many containers


andrekef:
### Bug description

Working on a 2023 Macbook Air M2

I followed the steps per wiki here https://superset.apache.org/docs/installation/installing-superset-using-docker-compose/ (which I believe needs updating) and I am unable to have stable build for many containers under the main superset container.

- Cannot reach `http://localhost:8088/`. Page is blank. 
- Error I am getting below.

```
Skipping local overrides
Starting web app (using development server)...
Skipping local overrides
Starting web app (using development server)...
Skipping local overrides
Starting web app (using development server)...
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
Usage: flask run [OPTIONS]
Try 'flask run --help' for help.

Error: While importing 'superset.app', an ImportError was raised:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/flask/cli.py", line 218, in locate_app
    __import__(module_name)
  File "/app/superset/__init__.py", line 21, in <module>
    from superset.app import create_app
  File "/app/superset/app.py", line 24, in <module>
    from superset.initialization import SupersetAppInitializer
  File "/app/superset/initialization/__init__.py", line 35, in <module>
    from superset.extensions import (
  File "/app/superset/extensions/__init__.py", line 30, in <module>
    from superset.async_events.async_query_manager import AsyncQueryManager
  File "/app/superset/async_events/async_query_manager.py", line 26, in <module>
    from superset.utils.core import get_user_id
  File "/app/superset/utils/core.py", line 90, in <module>
    from superset.sql_parse import sanitize_clause
  File "/app/superset/sql_parse.py", line 29, in <module>
    from sqlglot import exp, parse, parse_one
ModuleNotFoundError: No module named 'sqlglot'
```

This error above is looping endlessly 

### How to reproduce the bug

1. git clone https://github.com/apache/superset.git
2. cd superset
3. `open docker-compose.yml` 
4. paste `platform: linux/x86_64/v8` under each superset container - Please include in your wiki for arm64 users, as this will save folks a lot of time.

Eg

```
superset:
    platform: linux/x86_64/v8
    ...
  superset-websocket:
    platform: linux/amd64
    ...
  superset-init:
    platform: linux/x86_64/v8
    ...
   superset-node:
    platform: linux/x86_64/v8
    ...
[...]

PR #27146, opened 2024-02-16
feat(ci): no more docker builds on PR-related events

mistercrunch:
### SUMMARY
Speaking with the devexp committee, and in the recent community hall
meeting, it's pretty clear we don't need to build the whole docker
matrix on every PR event. It's somewhat expensive and isn't useful.

One of the reason we did it, from what I've gathered, is we want to
catch if some change breaks the docker build for whatever reason. Now
we may not need to actively prevent this pre-merge as this is unlikely,
yet we can still catch it on `master` as I'm keeping the push events on
master.

Note that:
- I'm also moving the master push build to be multi-platform (amd+arm)
  as we started doing this for the release builds, which is nice as
  the same image can be used on different archs
- The PR events were not "pushed" or cached in our shared, remote
  dockerhub-hosted repo anyways, they were just kind of dry-runs

Oh one more thing, notice the no-op.yml changes? That's because the
required checks related to branch protection are always based on what's
defined on .asf.yml ON `master`, meaning i need to add dummy/no-op
checks to get this merged, and those can be cleaned up later on in a
follow up PR. Let me know if you know a better way to deal with this.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -73,14 +73,6 @@ github:
           - test-postgres (3.9)
           - test-postgres (3.10)
           - test-sqlite (3.9)
-          - docker-build (dev, linux/amd64)
-          - docker-build (lean, linux/amd64)
-          - docker-build (py310, linux/arm64)
-          - docker-build (py310, linux/amd64)
-          - docker-build (websocket, linux/arm64)
-          - docker-build (websocket, linux/amd64)
-          - docker-build (dockerize, linux/arm64)
-          - docker-build (dockerize, linux/amd64)
 
       required_pull_request_reviews:
         dismiss_stale_reviews: false
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - 'master'
+  pull_request:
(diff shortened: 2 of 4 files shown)


craig-rueda:
How will users be able to test Dockerfile changes without running anything at all? I agree that the whole matrix probably doesn't need to be built, but what's stopping us from just allowing one path at a min to run? We could also remove it from the "required" checks as to not be a blocker

mistercrunch:
I agree @craig-rueda and adapted to run the "ci" build preset on pull_request, this will only `--load` since we don't have the creds in that context, but it should validate the build on both amd a arm, which we'll do systematically now across the board as the build matrix and resulting set of tags was getting confusing.

craig-rueda (approved):
LGTM

dpgaspar (approved):
This will probably help on https://github.com/apache/superset/issues/26997. I think that `docker-compose.yaml` -> `x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest-dev}` should point to `master-dev` now

proposing https://github.com/apache/superset/pull/27179 (if it makes sense)

PR #27149, opened 2024-02-17
chore(tests): Remove ineffectual login

john-bodley:
### SUMMARY

Whilst working on [[SIP-99B] Proposal for (re)defining a "unit of work"](https://github.com/apache/superset/issues/25108) I was running into an issue where (I thought) tests were failing due to an issue with the persistence of a logged in user whereas in actuality it was because I was missing the `Public` role.

The red herring was due to the slew of tests which don't interface with the test client "logging" in a user in the hope of persisting the logged in user to `g.user`. This is ineffectual outside of the client as the Flask globals only persist for the lifetime of the request, i.e., 

```python
>>> from flask import g

>>> self.login(username="admin")
>>> print(g.user)
<flask_login.mixins.AnonymousUserMixin object at 0x160c48be0>
```

Flask-Login automatically restores `g.user` from that cookie if it is not in the session when dealing with requests from the client. The TL;DR is authenticating (logging in/logging out) the user is only effectual in the context of web requests via the client.

This PR removes any erroneous logins (and logouts)—determined by first eliminating all references and re-adding them for API related tests. It also cleans up any inconsistencies in how logging in occurs. I updated the `login()` method to remove the default `admin` username so now it's a required argument which aids with readability.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/access_tests.py
+++ b/tests/integration_tests/access_tests.py
@@ -43,7 +43,7 @@ from superset.models import core as models
 from superset.utils.core import get_user_id, get_username, override_user
 from superset.utils.database import get_example_database
 
-from .base_tests import SupersetTestCase
+from tests.integration_tests.base_tests import SupersetTestCase
 
 ROLE_TABLES_PERM_DATA = {
     "role_name": "override_me",
@@ -103,15 +103,12 @@ class TestRequestAccess(SupersetTestCase):
             db.session.delete(security_manager.find_role(SCHEMA_ACCESS_ROLE))
             db.session.commit()
 
-    def setUp(self):
-        self.login("admin")
-
     def tearDown(self):
-        self.logout()
         override_me = security_manager.find_role("override_me")
         override_me.permissions = []
         db.session.commit()
         db.session.close()
+        super().tearDown()
 
 
 @pytest.mark.parametrize(
--- a/tests/integration_tests/base_tests.py
+++ b/tests/integration_tests/base_tests.py
@@ -106,6 +106,9 @@ class SupersetTestCase(TestCase):
 
     maxDiff = -1
 
+    def tearDown(self):
+        self.logout()
+
     def create_app(self):
         return app
 
@@ -196,7 +199,7 @@ class SupersetTestCase(TestCase):
         db.session.commit()
         return obj
 
-    def login(self, username="admin", password="general"):
+    def login(self, username, password="general"):
         return login(self.client, username, password)
 
     def get_slice(self, slice_name: str) -> Slice:
@@ -311,7 +314,7 @@ class SupersetTestCase(TestCase):
     ):
         if username:
             self.logout()
-            self.login(username=username)
+            self.login(username)
         dbid = SupersetTestCase.get_database_by_name(database_name).id
         json_payload = {
             "database_id": dbid,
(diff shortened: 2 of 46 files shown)


tests/integration_tests/access_tests.py
```diff
@@ -103,11 +103,7 @@ def tearDownClass(cls):
             db.session.delete(security_manager.find_role(SCHEMA_ACCESS_ROLE))
             db.session.commit()
 
-    def setUp(self):
```
john-bodley, line 106: My hope was to use a new fixture `auth_as_admin` (which would leverage the `test_client` fixture), however mixing some facets of `pytest` and `unittest` don't work. As a follow up we should remove all `unittest` logic and purely use `pytest` constructs.

tests/integration_tests/access_tests.py
```diff
@@ -103,11 +103,7 @@ def tearDownClass(cls):
...
-        self.login("admin")
-
     def tearDown(self):
-        self.logout()
```
john-bodley, line 110: The logout is unnecessary as the client only persists for the lifetime of the test.

michael-s-molina (approved):
LGTM. Thank you for the improvements @john-bodley.

My only suggestion would be to use constants for the commonly used usernames such as `admin`, `alpha`, `gamma`:

```
from tests.consts.usernames import ADMIN

self.login(username=ADMIN)
```

or something similar.

tests/integration_tests/base_tests.py
```diff
@@ -106,6 +106,9 @@ class SupersetTestCase(TestCase):
 
     maxDiff = -1
 
+    def tearDown(self):
```
john-bodley, line 109: The whole login/logout logic is convoluted given how our tests are defined. Ideally there would be an `auth_user` context manager and an associated `pytest` fixture which would be used to authenticate a user. Sadly this isn't overly viable with the `pytest`/`unittest` mix we have and so, although overkill—given that not every tests uses the API—this seemed like the best mechanism (given that all `unittest` classes are derived from `SupersetTestCase`) for ensuring that the tests are idempotent.

tests/integration_tests/sql_lab/api_tests.py
```diff
@@ -51,7 +56,11 @@ class TestSqlLabApi(SupersetTestCase):
         clear=True,
     )
     def test_get_from_empty_bootsrap_data(self):
-        self.login(username="gamma_sqllab_no_data")
+        if utils.backend() == "postgres":
```
john-bodley, line 59: @michael-s-molina I've spent/wasted hours trying to get a few failing PostgreSQL tests to pass. The tests succeed when testing locally and/or remotely when I run either i) the specific test, or ii) the specific file, however they fail when run as part of the entire test suite.

My hypothesis is that there are a slew of non-idempotent tests and this change merely (re)highlighted how fragile aspects of our tests are.

michael-s-molina: Yeah, it's kind of frustrating that we have many non-idempotent tests. Given that we'll need a specific effort for fixing this problem globally, I'm ok with skipping the test. We do the same for many sqlite tests.

michael-s-molina (approved):

tests/integration_tests/sql_lab/api_tests.py
john-bodley: Thanks @michael-s-molina. Also it's worth mentioning it's not like the test is being skipped globally, i.e., we have coverage on the said logic for MySQL and SQLite.

PR #27154, opened 2024-02-19
fix(import-datasources): Use "admin" user as default for importing datasources

### LINKED ISSUE #17049: Import of datasources (`superset import-datasources`) fails
The import of datasources (e.g. the connection to a PostgreSQL database) fails when using `superset import-datasources`. 

#### How to reproduce the bug

First, I performed an export of a database connection:

1. Spin up a new docker container (from my own image which is basically `superset/latest` + my config file)
1. Log into the application and navigate to `databaseview/list`
1. Create a new database by clicking on `+ DATABASE` (for reproduction, I created a new PostgreSQL connection named `db-for-debugging-import`)
1. Export the database by Clicking the Export-Icon in the database list (resulting in this ZIP file: [database_export_20211009T125106.zip](https://github.com/apache/superset/files/7315592/database_export_20211009T125106.zip))

Subsequently, I attempted to import the file: 

1. Unzip the exported datasource, replace the `xxxxxxx...` with the actual password (in this case a throw-away password), make a new zip archive (resulting in this ZIP file: [database_export_20211009T125106.zip](https://github.com/apache/superset/files/7315599/database_export_20211009T125106.zip))
1. Spin up a new docker container
1. Import the ZIP file by executing `superset import-datasources -p ...` in the container's shell

### Expected results

I expected the database to show up under `databaseview/list` and e.g. being able to use it when creating a new dataset (selecting the database, a schema, and a table). 

### Actual results

When importing via `superset import-datasources -p`, the logs contained the following:
````
superset@kolabri-dashboard-deployment-dd8dd7754-p5sq5:/app$ superset import_datasources -p /kolabri-assets/database_export_20211009T125106.zip
Loaded your LOCAL configuration at [/kolabri-assets/superset_config.py]
logging was configured successfully
2021-10-09 13:27:15,936:INFO:superset.utils.logging_configurator:logging was configured successfully
2021-10-09 13:27:15,946:INFO:root:Configured event logger of type <class 'superset.utils.log.DBEventLogger'>
/usr/local/lib/python3.8/site-packages/flask_caching/__init__.py:201: UserWarning: Flask-Caching: CACHE_TYPE is set to null, caching is effectively disabled.
  warnings.warn(
Command failed validation
2021-10-09 13:27:17,396:INFO:superset.datasets.commands.importers.dispatcher:Command failed validation
There was an error when importing the dataset(s), please check the exception traceback in the log
  File "/app/superset/cli.py", line 384, in import_datasources
    ImportDatasetsCommand(contents, overwrite=True).run()
  File "/app/superset/datasets/commands/importers/dispatcher.py", line 64, in run
    raise exc
  File "/app/superset/datasets/commands/importers/dispatcher.py", line 57, in run
    command.run()
  File "/app/superset/commands/importers/v1/__init__.py", line 63, in run
    self.validate()
  File "/app/superset/commands/importers/v1/__init__.py", line 146, in validate
    raise exception
superset.commands.exceptions.CommandInvalidError: Error importing dataset
2021-10-09 13:27:17,396:ERROR:superset.cli:There was an error when importing the dataset(s), please check the exception traceback in the log
  File "/app/superset/cli.py", line 384, in import_datasources
    ImportDatasetsCommand(contents, overwrite=True).run()
  File "/app/superset/datasets/commands/importers/dispatcher.py", line 64, in run
    raise exc
  File "/app/superset/datasets/commands/importers/dispatcher.py", line 57, in run
    command.run()
  File "/app/superset/commands/importers/v1/__init__.py", line 63, in run
    self.validate()
  File "/app/superset/commands/importers/v1/__init__.py", line 146, in validate
    raise exception
superset.commands.exceptions.CommandInvalidError: Error importing dataset
````

### Environment

- browser type and version: `Version 94.0.4606.71 (x86_64)`
- superset version: `0.0.0dev` (last pull of the base image was a on October 4th)
- python version: `Python 3.8.12`
- node.js version: No node.js installed
- any feature flags active:
  - `'VERSIONED_EXPORT' : True`

### Checklist

Regarding the second checkbox: I am happy to reproduce this with another "proper" version if this helps, but from [this](https://hub.docker.com/r/apache/superset) I am not 100% sure this is possible yet using the docker images :-)

### Additional context

- When importing the ZIP-file via the UI, the import succeeds. Though, an error message appears when trying to edit the database connection: `ERROR: Please re-enter the password`. I am not sure if this is the expected behaviour or part of the issue.
- It appears to me (from my outsider's perspective) that this bug is not identical to #16858 and #16848

I am happy to help if any additional info / debugging is useful and thanks in advance for the help :)

### PR DESCRIPTION
**Pull request by ddxv**
### SUMMARY
This is a potential fix for #17049 which prevents the use of CLI: `superset import-datasources` which appears to have been broken for for some time. Attempting to run `import-datasources` will result in a `NameError: user`.

Looking at the history of `import-datasources` there used to be a --user which could be applied, but was removed. I saw though that other CLI functions use the following `g.user = security_manager.find_user(username="admin")` to default to the admin user before the `security_manager` runs.

Please note, I'm new to Superset, so please help me check on whether using this is acceptable or an insecure workaround.

### TESTING INSTRUCTIONS
1) from CLI run: `superset import-datasources -p mydatasources.zip`

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #17049
- [ ] Required feature flags:
- [ ] Changes UI

### CODE REVIEW

**Comment by ddxv** [comment]
Mypy check was failing due to unrelated mypy errors from the master which have since been fixed. I merged in main again and mypy checks are passing now in my local pre-commit.

File: superset/cli/importexport.py [hunk 1]
```diff
@@ -176,6 +176,8 @@ def import_datasources(path: str) -> None:
     from superset.commands.dataset.importers.dispatcher import ImportDatasetsCommand
     from superset.commands.importers.v1.utils import get_contents_from_bundle
 
+    g.user = security_manager.find_user(username="admin")
```
john-bodley, line 179: Please use the `override_user` context manager.

john-bodley:
@ddxv you mentioned, 

> Looking at the history of import-datasources there used to be a `--user` ...

Would you mind sharing where you found this logic?

ddxv:
> @ddxv you mentioned,
> 
> > Looking at the history of import-datasources there used to be a `--user` ...
> 
> Would you mind sharing where you found this logic?

If that doesn't sound correct to you, I must have mixed up `import_dashboards` which has `--username`

```py
@click.option(
    "--username",
    "-u",
    default=None,
    help="Specify the user name to assign dashboards to",
)
 ```

**Comment by john-bodley** [comment]
I wonder if `import-datasources` should also have a user akin to `import-dashboards` as both entities are owned. Maybe @betodealmeida or @villebro would have thoughts on this.

File: superset/cli/importexport.py (hunk 1 above)
Comment (ddxv) [reply superset/cli/importexport.py]: @john-bodley I changed this line to `with override_user(user=security_manager.find_user(username="admin")):` as also used in `superset/sql_lab.py`

**Comment by ddxv** [comment]
I see some of MySQL tests are failing due to this PR being at the latest changes (ie they are due to other unrelated issues) or if it is directly caused by using the `security_manager`.  

If it is that I'm incorrectly using the master here, please let me know which branch or tag is better. If it's the `security_manager` I'll look into that.

File: superset/cli/importexport.py [hunk 2]
```diff
@@ -176,20 +177,21 @@ def import_datasources(path: str) -> None:
...
-            contents = {path: file.read()}
-    try:
-        ImportDatasetsCommand(contents, overwrite=True).run()
-    except Exception:  # pylint: disable=broad-except
-        logger.exception(
-            "There was an error when importing the dataset(s), please check the "
-            "exception traceback in the log"
-        )
-        sys.exit(1)
+    with override_user(user=security_manager.find_user(username="admin")):
```
dpgaspar, line 180: it seems it would probably make sense to add a `username` option here and leave `admin` as the default. @betodealmeida what to you think?

ddxv: From a consistency point of view, this makes sense to me. Let me know if you'd like me to go ahead and try with `--username` as an optional option.

geido: Agreed with Daniel's feedback there

geido: Hey @ddxv wondering if you had a chance to look at this last change. So tempted to merge it :)

ddxv: @geido Are you referring to Daniel's suggestion? I'm all for that. Do we need to wait for any confirmation from @betodealmeida? Just to be sure, I didn't see any recent "change" to code though, so let me know if I am missing something here.

ddxv: @geido I went ahead and pushed the change suggested by Daniel for an optional field with `default="admin"`. My only thought on that is that if you want to match `import_dashboards` it would be a required field without a default. This might be better just to have their API be the same for documentation and use, but let me know what you think.

geido: @ddxv it would be great if you want to open a PR to match `import_dashboards` unless there is any opposition. For the time being I am approving this PR and moving forward.

geido (approved):

PR #27163, opened 2024-02-19
feat(helm): optionally set pod disruption budgets

pradasouvanlasy:
### SUMMARY
This PRs optionally adds pod disruption budgets against all Superset deployments. By default, they are not enabled to keep the current behavior but we're now able to set PDBs on any deployment if relevant.

This also comes with a slight refactoring on the selector labels for each deployment: they are now defined as named templates in `_helpers.tpl` to mitigate selectors duplication between deployments and their respective related pod disruption budgets.

### TESTING INSTRUCTIONS
One can set any `supersetXxx.podDisruptionBudget.enabled` as `true` and add either `minAvailable` or `maxUnavailable` variables and test whether the generated manifests look as expected. 
PDB manifests are not generated when the relevant `podDisruptionBudget.enabled` is `false` and generation fails if any of the enabled PDB defines both `minAvailable` and `maxUnavailable` at the same time.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


villebro:
@pradasouvanlasy please bump the patch version number

helm/superset/templates/_helpers.tpl
```diff
@@ -136,3 +136,28 @@ RESULTS_BACKEND = RedisCache(
...
+
+{{- define "supersetWebsockets.selectorLabels" -}}
+app: {{ include "superset.name" . }}-ws
+release: {{ .Release.Name }}
+{{- end }}
+
+{{- define "supersetWorker.selectorLabels" -}}
+app: {{ include "superset.name" . }}-worker
+release: {{ .Release.Name }}
+{{- end }}
```
villebro, line 163: Maybe also mention this refactor in the description

helm/superset/values.yaml
```diff
@@ -248,6 +248,8 @@ supersetNode:
     maxReplicas: 100
     targetCPUUtilizationPercentage: 80
     # targetMemoryUtilizationPercentage: 80
+  # -- Sets the [pod disruption budget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) for supersetNode pods
```
craig-rueda, line 251: Can you add an `enabled` field here, along with a set of sensible defaults for each of these blocks? That way, we can add the various settings to the schema and can document them appropriately. Using an `enabled` also makes the config a bit more explicit

e.g.
```yaml
podDisruptionBudget:
  minAvailable: 1
  maxUnavailable: 1
```

craig-rueda (commented):
Thanks for your contribution! Left a comment below, otherwise LGTM

helm/superset/values.yaml
pradasouvanlasy: Thanks, I will take this into account. However both variables are mutually exclusive so we would have to comment these in `values.yaml` and document somehow that PDB only accepts either `minAvailable` or `maxUnavailable` at a time - https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget.

craig-rueda: Yep, that makes sense. For extra credit, you could fail if both are set:

```helm
{{- if .Values.podDisruptionBudget.minAvailable and .Values.podDisruptionBudget.maxAvailable }}
{{ required "Only one of minAvailable or maxAvailable should be set" .Values.podDisruptionBudget.minAvailable }}
{{- end}}
```

pradasouvanlasy: I've made the changes and slightly refactored the PDB templates I'm contributing there to make better use of variable scoping - reducing a bit the boilerplate and avoiding to repeat `.Values.supersetXXX.podDisruptionBudget` everywhere.

Please tell me if the updated documentation looks clear enough as well. Thanks!

craig-rueda (approved):
LGTM

PR #27179, opened 2024-02-20
fix: docker-compose point to master tag

dpgaspar:
### SUMMARY
We now use `master` and `master-dev` for latest master tags. This proposes a small change that will probably help people here: https://github.com/apache/superset/issues/26997 combined with https://github.com/apache/superset/pull/27146

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docker-compose-non-dev.yml
+++ b/docker-compose-non-dev.yml
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-master}
+x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest}
 x-superset-depends-on: &superset-depends-on
   - db
   - redis
--- a/docker-compose-non-dev.yml
+++ b/docker-compose-non-dev.yml
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest}
+x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-master}
 x-superset-depends-on: &superset-depends-on
   - db
   - redis
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest-dev}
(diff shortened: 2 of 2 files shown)


docker-compose-non-dev.yml
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest}
+x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-master}
```
eschutho, line 17: I think it makes sense to have devs work from master, but for production, wouldn't it be better for them to work from an official release tag?

sfirke: Agreed with @eschutho - I don't understand all of #26997 but it seems like these people are all trying to run latest builds off master instead of an official release?  Or using a mismatch of master repo clone + an official release.  Which underscores for me the value of directing new users to an official release instead of master, so that they have a positive experience right out of the box.

sfirke: See also my comment here: https://github.com/apache/superset/issues/26997#issuecomment-1956816278

sfirke (approved):

mistercrunch:
Awesome. I just merged https://github.com/apache/superset/pull/27146 which should make the `master` builds to also be multi-platform, meaning this should work and it will support Apple silicon.

But note that there's something loose here, where it looks like the assumption with docker-compose is that you can mount the code on your local `master` with the latest `master` image build and things should just work. That's not super solid/guaranteed. You could be on an older commit that say assumes a python req that's been removed on the latest master. Or you could be ahead of master, introduce a new python req, and for it now be reflected on the image.

Knowing this, should we set docker-compose to "docker build" from the Dockerfile and simply leverage the caching?

dpgaspar:
> Awesome. I just merged #27146 which should make the `master` builds to also be multi-platform, meaning this should work and it will support Apple silicon.
> 
> But note that there's something loose here, where it looks like the assumption with docker-compose is that you can mount the code on your local `master` with the latest `master` image build and things should just work. That's not super solid/guaranteed. You could be on an older commit that say assumes a python req that's been removed on the latest master. Or you could be ahead of master, introduce a new python req, and for it now be reflected on the image.
> 
> Knowing this, should we set docker-compose to "docker build" from the Dockerfile and simply leverage the caching?

that's a valid point, I guess it depends on how much time it would normally take to build vs the frequency of master requirements being off.

How can we leverage caching when `docker-compose build` points to Dockerfile?
There must be a way

mistercrunch:
> How can we leverage caching when docker-compose build points to Dockerfile?

I think it should leverage cached layers if pointed to the right cache that supports the `master` builds (`apache/superset-cache`). In theory a build that doesn't touch the dependencies should be really fast. I proposed something on https://github.com/apache/superset/issues/26997#issuecomment-1960343654

PR #27186, opened 2024-02-21
fix: SSH Tunnel configuration settings 

geido:
### SUMMARY
This PR fixes several issues with the SSH Tunnel option in the Database Modal which are related to edit the SSH Tunnel config, enable/disable the option, and misc fixes/enhancements. In particular:

- Removes a duplicate `SSHTunnelSwitch` component from dynamic forms and merges the logics of both dynamic and SQLAlchemy forms into one overridable component
- Fixes an issue that made removing the SSH tunnel config impossible when editing a database connection
- Delegates the deletion of a ssh tunnel to the `UpdateDatabaseCommand`
- Fixes an issue with the SSH tunnel config still being included in the request even when toggled off
- Fixes an issue for which the switch was showing also with `SSH_TUNNELING` OFF.  Saving would cause a "SSH Tunneling is not enabled" error
- Surfaces to the user a missing port error in the SQLAlchemy uri when enabling SSH Tunnel (the port is required for the tunnel to work)
- Clean the SSH Tunnel config from unnecessary credentials when switching from password to private key
- Improves the UX so that toggling ON and OFF the SSH Tunnel config won't clear the form until saved
- Misc fixes

### BEFORE

https://github.com/apache/superset/assets/60598000/c38052d6-53b0-47d8-a87a-391a647085d9

### AFTER

https://github.com/apache/superset/assets/60598000/f65ad7da-962f-4500-82f2-eb42bd2139ba

### TESTING INSTRUCTIONS
1. Create a new database connection using a SQLAlchemy uri and dynamic forms, make sure SSH Tunnel can be toggled ON and OFF and that configs can be saved successfully
2. Edit a database connection that has a SSH Tunnel config, make sure the SSH Tunnel can be toggled ON and OFF and that it keeps the config available to the user until toggled OFF and saved
3. Edit a database connection that has a SSH Tunnel config, make sure any change to the SSH tunnel can be saved successfully
4. Re-open the database config and make sure the SSH Tunnel has been deleted successfully
5. Create or edit a database connection using a SQLAlchemy uri without specifying the port, make sure an error is surfaced to the UI informing that specifying a port is necessary

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/features/databases/DatabaseModal/index.tsx
```diff
@@ -705,7 +721,7 @@ const DatabaseModal: FunctionComponent<DatabaseModalProps> = ({
     setSSHTunnelPrivateKeys({});
     setSSHTunnelPrivateKeyPasswords({});
     setConfirmedOverwrite(false);
-    setUseSSHTunneling(false);
+    setUseSSHTunneling(undefined);
```
eschutho, line 724: why undefined instead of false?

eschutho: Oh I see, undefined means it was never checked and `false` means it was unchecked?

superset-frontend/src/features/databases/DatabaseModal/index.tsx
```diff
@@ -1282,10 +1299,10 @@ const DatabaseModal: FunctionComponent<DatabaseModalProps> = ({
   }, [sshPrivateKeyPasswordNeeded]);
 
   useEffect(() => {
-    if (db && isSSHTunneling) {
-      setUseSSHTunneling(!isEmpty(db?.ssh_tunnel));
+    if (db?.parameters?.ssh !== undefined) {
```
eschutho, line 1302: similar thing expected here? `undefined` is different than `false`?

geido: Yes just differentiating between 3 possible states, never checked, checked, unchecked

kgabryje (approved):
Added a few nonblocking comments but looks great!

superset/commands/database/ssh_tunnel/update.py
```diff
@@ -43,6 +45,15 @@ def run(self) -> Model:
         self.validate()
         try:
             if self._model is not None:  # So we dont get incompatible types error
```
dpgaspar, line 47: nit: we could avoid nesting by:
```
if self._model is None:
    return
...
```

superset/commands/database/update.py
```diff
@@ -47,15 +51,19 @@
 
 
 class UpdateDatabaseCommand(BaseCommand):
+    _model: Optional[Database]
+
     def __init__(self, model_id: int, data: dict[str, Any]):
         self._properties = data.copy()
         self._model_id = model_id
-        self._model: Optional[Database] = None
+        self._model = DatabaseDAO.find_by_id(self._model_id)
```
dpgaspar, line 59: I think our current pattern is still to make all queries/validation on the `run`

superset/commands/database/update.py
```diff
@@ -70,36 +78,59 @@ def run(self) -> Model:
             database = DatabaseDAO.update(self._model, self._properties, commit=False)
             database.set_sqlalchemy_uri(database.sqlalchemy_uri)
 
-            if ssh_tunnel_properties := self._properties.get("ssh_tunnel"):
+            ssh_tunnel = DatabaseDAO.get_ssh_tunnel(database.id)
+
+            if "ssh_tunnel" in self._properties:
```
dpgaspar, line 83: nit: invert this if and return early to reduce nesting

dpgaspar (approved):
a couple of non blocking comments

PR #27198, opened 2024-02-22
chore(node): bumping Superset to Node 18

rusackas:
### SUMMARY

Not sure all the actions bumps are needed, but I'll just aim high and hope we can get everything updated.

This does NOT bump the websocket app or docs to Node 18.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch (approved):
LGTM after hold is removed

rusackas:
@michael-s-molina any objection to merging this now? We have these warnings cropping up, which makes me almost tempted to call this a "fix" rather than a "chore". It's non-blocking for 4.0 and doesn't _need_ to be cherried, necessarily. But I also don't want to rock the boat on your release :) 

image

michael-s-molina:
My only concern is this causing bugs on master which will be reported as issues and we'll need to check if they happen on 4.0. If you think the probability of new bugs is low then go ahead and merge it.

rusackas:
All conflicts resolved... hopefully we can merge this soon :D

.github/workflows/tech-debt.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "16"
-
+          node-version: '18'
+
```
rusackas, line 36:
```suggestion
```suggestion
          node-version: '18'

```
```

mistercrunch:
cleaning

eschutho:
@rusackas It looks like npm 9+ use lockfile version 3, so I also updated that check here: 
https://github.com/apache/superset/pull/29723/files

PR #27203, opened 2024-02-22
fix(plugin-chart-period-over-period-kpi): Blank chart when switching from BigNumberTotal

kgabryje:
### SUMMARY
When user creates and saves a Big Number chart and then switches to Big Number Period over Period, the chart appears blank. The reason for that is that Period over Period chart expects the values coming from `header_font_size` control to be pixels. However, the old Big Number chart uses proportions (0.2-0.6) in that control, so when the new chart inherits those values, the size in pixels is <1, so it seems blank.
This PR implements a mapping between proportion values used in Big Number and actual pixel size values expected by Period over Period chart.

Additionally, this PR moves the KPI chart to a common directory with other BigNumber plugins so that they can share to font size controls.

1. Create a new Big number chart using the Vehicle Sales chart.
2. Save the chart
3. Switch to Big Number with Time Period Comparison" chart
4. Finally, update the chart.
5. Verify that the chart renders correctly

### TESTING INSTRUCTIONS

Before:

https://github.com/apache/superset/assets/15073128/2ad7c87a-1344-467d-bca4-67a5319b1958

After:

https://github.com/apache/superset/assets/15073128/f4b3e61c-bdd9-4b0c-9e61-fb15ce232c1a

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
```diff
@@ -142,16 +143,15 @@ export default function transformProps(chartProps: ChartProps) {
...
     metric,
     metricName,
     bigNumber,
     prevNumber,
     valueDifference,
     percentDifferenceFormattedString: percentDifference,
     boldText,
-    headerFontSize,
-    subheaderFontSize,
+    headerFontSize: getHeaderFontSize(headerFontSize),
```
Antonio-RiveroMartnez, line 153: Should we make use of these utils in the other `BigNumber` viz too?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/index.ts
```diff
@@ -19,3 +19,4 @@
 
 export { default as BigNumberChartPlugin } from './BigNumberWithTrendline';
 export { default as BigNumberTotalChartPlugin } from './BigNumberTotal';
+export { default as PopKPIPlugin } from './BigNumberPeriodOverPeriod';
```
Antonio-RiveroMartnez, line 22: Totally doable later on in a follow up PR, but could we align the component's name with it's directory name? i.e, from `PopKPIPlugin` to `BigNumberPeriodOverPeriod `

Antonio-RiveroMartnez (approved):
❤️ thanks for making these changes, I left a couple comments but nothing blocking.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
kgabryje: No - in those charts the actual font size is calculated dynamically based on chart's container width and height

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/index.ts
kgabryje: Oh right, I just copied this one from previous file

kgabryje: Done

sadpandajoe:
🏷️ preset:2024.8

PR #27209, opened 2024-02-22
fix: Allow only dttm columns in comparison filter in Period over Period chart

kgabryje:
### SUMMARY
When user selected Custom range for comparison on Big Number With Time Period Comparison chart, the filter control would accept any column. That worked but didn't make much sense, so this PR restricts the control to accept only dttm columns for filtering.
Please note that user can still write custom sql which will filter anything, so that's just a UX improvement.

https://github.com/apache/superset/assets/15073128/b4a63051-72f4-4a3c-900d-67aabcc79ba1

### TESTING INSTRUCTIONS
1. Create Big Number With Time Period Comparison chart
2. Select Custom range for comparison
3. Verify that you can drag only temporal columns
4. Verify that in the columns select in the popover you can only select temporal columns

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/controlPanel.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/controlPanel.ts
@@ -17,7 +17,9 @@
  * under the License.
  */
 import {
+  AdhocFilter,
   ComparisonTimeRangeType,
+  SimpleAdhocFilter,
   t,
   validateTimeComparisonRangeValues,
 } from '@superset-ui/core';
@@ -84,7 +86,12 @@ const config: ControlPanelConfig = {
                     controlState,
                   ) || {};
                 const columns = originalMapStateToPropsRes.columns.filter(
-                  (col: ColumnMeta) => col.is_dttm,
+                  (col: ColumnMeta) =>
+                    col.is_dttm &&
+                    (state.controls.adhoc_filters.value as AdhocFilter[]).some(
+                      (val: SimpleAdhocFilter) =>
+                        val.subject === col.column_name,
+                    ),
                 );
                 return {
                   ...originalMapStateToPropsRes,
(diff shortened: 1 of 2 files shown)


superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.tsx
```diff
@@ -382,7 +382,18 @@ const DndFilterSelect = (props: DndFilterSelectProps) => {
     return new AdhocFilter(config);
   }, [droppedItem]);
 
-  const canDrop = useCallback(() => true, []);
+  const canDrop = useCallback(
```
Antonio-RiveroMartnez, line 385: non blocking: does any testable behavior changed with this block? If so, could add include it in the  `DndMetricSelect.test.tsx` file?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/controlPanel.ts
```diff
@@ -76,16 +77,24 @@ const config: ControlPanelConfig = {
...
-                ...(sharedControls.adhoc_filters.mapStateToProps?.(
-                  state,
-                  controlState,
-                ) || {}),
-                externalValidationErrors: validateTimeComparisonRangeValues(
-                  state.controls?.time_comparison?.value,
-                  controlState.value,
-                ),
-              }),
+              ) => {
```
Antonio-RiveroMartnez, line 80: ❤️

Antonio-RiveroMartnez (approved):
:fire:

superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.tsx
kgabryje: No. The drag and drop behavior tested in `DndMetricSelect.test.tsx` is related to reordering the metric values. Testing drag and drop from the datasource panel to the control field is more complicated - I suggest we add those tests in a separate PR, both for filters and metrics

PR #27217, opened 2024-02-22
fix(sqlglot): Address regressions introduced in #26476

john-bodley:
### SUMMARY

This PR addresses some regressions introducted in https://github.com/apache/superset/pull/26476 when `sqlparse` was replaced with `sqlglot` for SQL parsing. These issues were reported by our users who were unable to access their saved queries after we deployed Apache Superset 3.1.1.

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -28,7 +28,7 @@ import sqlparse
 from sqlalchemy import and_
 from sqlglot import exp, parse, parse_one
 from sqlglot.dialects import Dialects
-from sqlglot.errors import ParseError
+from sqlglot.errors import SqlglotError
 from sqlglot.optimizer.scope import Scope, ScopeType, traverse_scope
 from sqlparse import keywords
 from sqlparse.lexer import Lexer
@@ -287,7 +287,7 @@ class ParsedQuery:
         """
         try:
             statements = parse(self.stripped(), dialect=self._dialect)
-        except ParseError:
+        except SqlglotError:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
             return set()
 
@@ -319,12 +319,17 @@ class ParsedQuery:
         elif isinstance(statement, exp.Command):
             # Commands, like `SHOW COLUMNS FROM foo`, have to be converted into a
             # `SELECT` statetement in order to extract tables.
-            literal = statement.find(exp.Literal)
-            if not literal:
+            if not (literal := statement.find(exp.Literal)):
                 return set()
 
-            pseudo_query = parse_one(f"SELECT {literal.this}", dialect=self._dialect)
-            sources = pseudo_query.find_all(exp.Table)
+            try:
+                pseudo_query = parse_one(
+                    f"SELECT {literal.this}",
+                    dialect=self._dialect,
+                )
+                sources = pseudo_query.find_all(exp.Table)
+            except SqlglotError:
+                return set()
         else:
             sources = [
                 source
--- a/tests/unit_tests/sql_parse_tests.py
+++ b/tests/unit_tests/sql_parse_tests.py
@@ -271,6 +271,7 @@ def test_extract_tables_illdefined() -> None:
     assert extract_tables("SELECT * FROM catalogname..tbname") == {
         Table(table="tbname", schema=None, catalog="catalogname")
     }
+    assert extract_tables('SELECT * FROM "tbname') == set()
 
 
 def test_extract_tables_show_tables_from() -> None:
@@ -558,6 +559,10 @@ def test_extract_tables_multistatement() -> None:
         Table("t1"),
         Table("t2"),
     }
+    assert extract_tables(
+        "ADD JAR file:///hive.jar; SELECT * FROM t1;",
+        engine="hive",
+    ) == {Table("t1")}
 
 
 def test_extract_tables_complex() -> None:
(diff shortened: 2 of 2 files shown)


superset/sql_parse.py
```diff
@@ -285,9 +285,10 @@ def _extract_tables_from_sql(self) -> set[Table]:
 
         Note: this uses sqlglot, since it's better at catching more edge cases.
         """
+
         try:
             statements = parse(self.stripped(), dialect=self._dialect)
-        except ParseError:
+        except SqlglotError:
```
john-bodley, line 291: Parsing (for right or wrong) can throw either a `ParseError` or `TokenError`. It seems like both a derived from the the [`SqlglotError`](https://github.com/tobymao/sqlglot/blob/main/sqlglot/errors.py#L23C7-L23C19) error type and thus it seemed prudent to include a more broader except.

superset/sql_parse.py
```diff
@@ -319,12 +320,17 @@ def _extract_tables_from_statement(self, statement: exp.Expression) -> set[Table
...
             # Commands, like `SHOW COLUMNS FROM foo`, have to be converted into a
             # `SELECT` statetement in order to extract tables.
-            literal = statement.find(exp.Literal)
-            if not literal:
+            if not (literal := statement.find(exp.Literal)):
                 return set()
 
-            pseudo_query = parse_one(f"SELECT {literal.this}", dialect=self._dialect)
-            sources = pseudo_query.find_all(exp.Table)
+            try:
```
john-bodley, line 326: This issue surfaced when trying to parse an `ADD JAR ...` statement. which `sqlglot` believes contains literal expressions. The TL;DR is `parse_one` can throw and we should catch these errors (similar to when using the `sqlglot.parse` method).

tests/unit_tests/sql_parse_tests.py
```diff
@@ -1815,10 +1820,7 @@ def test_extract_table_references(mocker: MockerFixture) -> None:
     # test falling back to sqlparse
     logger = mocker.patch("superset.sql_parse.logger")
     sql = "SELECT * FROM table UNION ALL SELECT * FROM other_table"
-    assert extract_table_references(
-        sql,
-        "trino",
-    ) == {
+    assert extract_table_references(sql, "trino") == {
```
john-bodley, line 1823: This was picked up by Black as part of the `pre-commit` check.

tests/unit_tests/sql_parse_tests.py
```diff
@@ -271,6 +271,7 @@ def test_extract_tables_illdefined() -> None:
     assert extract_tables("SELECT * FROM catalogname..tbname") == {
         Table(table="tbname", schema=None, catalog="catalogname")
     }
+    assert extract_tables('SELECT * FROM "tbname') == set()
```
john-bodley, line 274: The ill-formed SQL statement throws a `TokenError` as opposed to a `ParseError` in `sqlglot`.

tests/unit_tests/sql_parse_tests.py
```diff
@@ -558,6 +559,10 @@ def test_extract_tables_multistatement() -> None:
         Table("t1"),
         Table("t2"),
     }
+    assert extract_tables(
```
john-bodley, line 562: This example throws as `sqlglot` thinks the `ADD JAR` is a valid literal.

superset/sql_parse.py
tobymao: i have no idea what your use case is, but if you want more lenient sql parsing, you can try error_level=IGNORE

john-bodley: Ha! I was wondering whether I should have raised this with you.

tobymao: always free to chat. you can hit me up on slack (i'm in your slack or you can come to mine)

tests/unit_tests/sql_parse_tests.py
john-bodley: @tobymao ideally should illformed SQL throw a `TokenError` or `ParseError`?

tobymao: it depends on if it's a token error or a parser error!

when you have unbalanced quotes, it's basically impossible to finish tokenization, so that's why we throw a token error.

most everything else is a parser error

superset/sql_parse.py
giftig: Aside, but I had a similar problem with the old `sqlparse` solution with giving CTEs names which are keywords in other engines; I'm curious if swapping to sqlglot has fixed that actually, since it's more engine-aware. Specifically I made the mistake of using the name `ref` for the CTE in a trino query, and `REF` is apparently a keyword in some engines (not trino as far as I'm aware). I'll give it a go with the new sqlglot approach sometime.

giftig (approved):
Won't merge just yet in case you want to try the alternative lenient parsing solution proposed by @tobymao but LGTM as it is.

michael-s-molina (approved):
LGTM. Thanks for the fix @john-bodley!

john-bodley:
@giftig regarding your comment, 

> Won't merge just yet in case you want to try the alternative lenient parsing solution proposed by @tobymao but LGTM as it is.

I agree we could include a more lenient parser—likely as a follow up—but I think it's prudent to include a `try`/`except` block around the parsing of all our SQL statements given that there maybe either a tokenization or parsing error for any number of reasons.

In the case of the `parse_one` call, even though all the statements have already been successfully parsed, we're creating a new statement of the form `SELECT {literal.this}` which may result in an invalid SQL statement.

PR #27221, opened 2024-02-23
feat(CLI command): Apache Superset "Factory Reset" CLI command #27207

mknadh:
feat(CLI command): Apache Superset "Factory Reset" CLI command #27207

### SUMMARY
Over time, Apache Superset instances can accumulate large amounts of data, including charts, dashboards, saved queries, and other artifacts. There might be scenarios where users want to start fresh with a clean slate, removing all existing data.

Testing and Development: Developers and administrators often require a quick and efficient way to reset Apache Superset instances to a default state for testing purposes or when setting up development environments.

Data Privacy and Security: In some cases, there might be sensitive or confidential data stored within Apache Superset that needs to be wiped out completely to ensure data privacy and security compliance.

What is the proposal?

Implement a Factory CLI command: Develop CLI command that allows users to trigger a factory reset of Apache Superset. This CLI should be designed to delete all existing data, including charts, dashboards, saved queries, databases, and other related artifacts.

Authorization and Confirmation: Ensure that the CLI requires appropriate authorization to prevent unauthorized access. Additionally, consider implementing a confirmation mechanism to prevent accidental data loss.

Documentation and Best Practices: Provide comprehensive documentation on how to use the Factory Reset CLI, along with best practices for when and how to perform a reset. Include warnings about data loss and the irreversible nature of the operation.
Error Handling and Logging: Implement robust error handling mechanisms within the CLI to handle any unexpected errors gracefully. Additionally, log all reset operations for auditing purposes.

Integration with Configuration Management: Optionally, provide integration with configuration management tools or scripts to automate the process of resetting Apache Superset instances in a controlled and reproducible manner.

### ADDITIONAL INFORMATION

- [x ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/cli/reset.py
+++ b/superset/cli/reset.py
@@ -1,0 +1,74 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import sys
+
+import click
+from flask.cli import with_appcontext
+from werkzeug.security import check_password_hash
+
+from superset.cli.lib import feature_flags
+
+if feature_flags.get("ENABLE_FACTORY_RESET_COMMAND"):
+
+    @click.command()
+    @with_appcontext
+    @click.option("--username", prompt="Admin Username", help="Admin Username")
+    @click.option(
+        "--silent",
+        is_flag=True,
+        prompt=(
+            "Are you sure you want to reset Superset? "
+            "This action cannot be undone. Continue?"
+        ),
+        help="Confirmation flag",
+    )
+    @click.option(
+        "--exclude-users",
+        default=None,
+        help="Comma separated list of users to exclude from reset",
+    )
+    @click.option(
+        "--exclude-roles",
+        default=None,
+        help="Comma separated list of roles to exclude from reset",
+    )
+    def factory_reset(
+        username: str, silent: bool, exclude_users: str, exclude_roles: str
+    ) -> None:
+        """Factory Reset Apache Superset"""
+
+        # pylint: disable=import-outside-toplevel
+        from superset import security_manager
+        from superset.commands.security.reset import ResetSupersetCommand
+
+        # Validate the user
+        password = click.prompt("Admin Password", hide_input=True)
+        user = security_manager.find_user(username)
+        if not user or not check_password_hash(user.password, password):
+            click.secho("Invalid credentials", fg="red")
+            sys.exit(1)
+        if not any(role.name == "Admin" for role in user.roles):
+            click.secho("Permission Denied", fg="red")
+            sys.exit(1)
+
+        try:
+            ResetSupersetCommand(silent, user, exclude_users, exclude_roles).run()
+            click.secho("Factory reset complete", fg="green")
+        except Exception as ex:  # pylint: disable=broad-except
+            click.secho(f"Factory reset failed: {ex}", fg="red")
+            sys.exit(1)
--- a/superset/commands/security/reset.py
+++ b/superset/commands/security/reset.py
@@ -1,0 +1,94 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import logging
+from typing import Any, Optional
+
+from superset import db, security_manager
+from superset.commands.base import BaseCommand
+from superset.connectors.sqla.models import SqlaTable
+from superset.key_value.models import KeyValueEntry
+from superset.models.core import Database, FavStar, Log
+from superset.models.dashboard import Dashboard
+from superset.models.slice import Slice
+
+logger = logging.getLogger(__name__)
+
+
+class ResetSupersetCommand(BaseCommand):
+    def __init__(
+        self,
+        confirm: bool,
+        user: Any,
+        exclude_users: Optional[str] = None,
+        exclude_roles: Optional[str] = None,
+    ) -> None:
+        self._user = user
+        self._confirm = confirm
+        self._users_to_exclude = ["admin"]
+        if exclude_users:
+            self._users_to_exclude.extend(exclude_users.split(","))
+        self._roles_to_exclude = ["Admin", "Public", "Gamma", "Alpha", "sql_lab"]
+        if exclude_roles:
+            self._roles_to_exclude.extend(exclude_roles.split(","))
+
+    def validate(self) -> None:
+        if not self._confirm:
+            raise Exception("Reset aborted.")  # pylint: disable=broad-exception-raised
+        if not self._user or not self._user.is_active:
+            raise Exception("User not found.")  # pylint: disable=broad-exception-raised
+
+    def run(self) -> None:
+        self.validate()
+        logger.debug("Resetting Superset Started")
+        db.session.query(SqlaTable).delete()
+        databases = db.session.query(Database)
+        for database in databases:
+            db.session.delete(database)
+        db.session.query(Dashboard).delete()
+        db.session.query(Slice).delete()
+        db.session.query(KeyValueEntry).delete()
+        db.session.query(Log).delete()
+        db.session.query(FavStar).delete()
+
+        logger.debug("Ignoring Users: %s", self._users_to_exclude)
+        users_to_delete = (
+            db.session.query(security_manager.user_model)
+            .filter(security_manager.user_model.username.not_in(self._users_to_exclude))
+            .all()
+        )
+        for user in users_to_delete:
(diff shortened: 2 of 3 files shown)


john-bodley:
Thanks @mknadh for the PR. Any reason you didn't opt for a CLI command for removing all assets? Additionally if you're after a factory reset why not just drop all the tables in the database and start from afresh?

mistercrunch:
Airflow used to have a "convenient" `resetdb` CLI subcommand, as in `airflow resetdb` (maybe it still does? I think we removed it). I thought it was a good idea early on to make my dev life easier to reset environments. Hell I think it even asked for confirmation. Well I changed my mind pretty quickly when someone internally at one of the companies I worked at used it against the wrong environment (PROD).

superset/reset/api.py
```diff
@@ -0,0 +1,128 @@
...
+            500:
+              $ref: '#/components/responses/500'
+        """
+        try:
+            item = self.reset_model_schema.load(request.json)
+        # This validates custom Schema with custom validations
+        except ValidationError as error:
+            return self.response_400(message=error.messages)
+        try:
+            if item.get("all") or item.get("datasets"):
```
craig-rueda, line 93: Can we make a command for this?

superset/initialization/__init__.py
```diff
@@ -221,6 +222,7 @@ def init_views(self) -> None:
         appbuilder.add_api(QueryRestApi)
         appbuilder.add_api(ReportScheduleRestApi)
         appbuilder.add_api(ReportExecutionLogRestApi)
+        appbuilder.add_api(ResetRestApi)
```
craig-rueda, line 225: Can we put this add behind a boolean config that defaults to `False`? Although useful for local dev work, this feels a little dangerous to have enabled by default

craig-rueda (commented):
Looks good for local testing, but probably want to lock this down by default, and allow Superset admins to opt-in

craig-rueda:
One more comment - I agree with @john-bodley that this would probably be better done as a CLI.

michael-s-molina:
I also agree this shouldn't be a new API. If we create a command, it should require many levels of confirmation to avoid the problem @mistercrunch mentioned.

giftig:
+1 for the suggestions to make it a cli command if anything, for reasons already stated. But what would this do which isn't already achieved by simply dropping the database?

mistercrunch:
Could we have a config setting in `superset/config.py` that disallows this by default or maybe base on DEBUG, something like `ALLOW_RESET = DEBUG`?

mknadh:
> +1 for the suggestions to make it a cli command if anything, for reasons already stated. But what would this do which isn't already achieved by simply dropping the database?

https://github.com/apache/superset/discussions/27207#discussioncomment-8564014

superset/initialization/__init__.py
mknadh: Sure, this is a good idea and will add it behind a feature flag, where the **default behavior** of the feature flag would be to disable this API (by default). That way, no one can run it against PROD, unless enabled.

CC: @john-bodley , @mistercrunch , @craig-rueda , @michael-s-molina

mknadh:
> One more comment - I agree with @john-bodley that this would probably be better done as a CLI.

Here are some reasons of implementing this as a REST API rather than a CLI command:

1. Better access control: REST APIs allow leveraging OAuth and complex permission policies to control access. CLI commands are harder to lock down.

2. Automation and integration: REST APIs lend themselves better to automation, configuration management, and integration with other systems.

3. Consistency: Adding to the existing CRUD REST APIs for other Superset model objects maintains a consistent interface for clients.

4. Reporting: REST APIs usage can be better tracked, monitored and audited for something as sensitive as a data reset.

5. Error handling: REST APIs provide better mechanisms for structured error handling compared to CLIs.

CC: @john-bodley, @mistercrunch, @craig-rueda, @michael-s-molina

mknadh:
> I also agree this shouldn't be a new API. If we create a command, it should require many levels of confirmation to avoid the problem @mistercrunch mentioned.

https://github.com/apache/superset/pull/27221#issuecomment-1962201344

mknadh:
> Could we have a config setting in `superset/config.py` that disallows this by default or maybe based on DEBUG, something like `ALLOW_RESET = DEBUG` or straight up `ALLOW_RESET = False`

https://github.com/apache/superset/pull/27221#discussion_r1501301049

giftig:
@mknadh I don't think I can agree with any of the points you listed [above](https://github.com/apache/superset/pull/27221#issuecomment-1962201344).

(1) CLI commands are inherently more locked down as you're running a command on the box running superset and this is typically very much locked down in a production environment, only available to devops engineers. Bear in mind, as well, anyone who could run a CLI command could also get a flask shell and simply perform the same operation by writing the code themselves.

(2) CLI commands can be automated, and it should be fairly straightforward in any context where you're likely to want to wipe out a whole database. Can you provide examples where you'd want to wipe out the database via REST API and a cli command would not be a viable option?

(3) I don't think this argument is valid; it's unreasonable and dangerous to expect Superset to provide a "wipe out all data" option as a production API.

(4) Monitoring and auditing can be applied to an environment in which a cli command could be run, and I'd make the same arguments as point (1).

(5) That's not true. CLI tools have structured error handling and reporting mechanisms too.

In my opinion an operation like this is too destructive and shouldn't be provided at all.

I'd be a little more comfortable with it provided as a cli command to be run in development environments only, and (1) warn very clearly about the destructive nature of the operation and (2) do some sanity checks of your config to make sure it's not applied in a production-like environment.

In terms of providing a rest API to do it, no, I don't see any reason why this could be desirable. You'd never want to run this in a production environment. And who should be able to perform this operation via a rest operation? Any admin user? Even in a large deployment where there are half a dozen administration staff of varying levels of familiarity with Superset or technical ability?

zhaoyongjie:
@mknadh Hey, I believe you can't success to call the "reset" REST API in a real production environment since the database should lock a table if other users read/write at same time.

BTW.....if you really need a new metadata of Superset, please create a new database instance and change `SQLALCHEMY_DATABASE_URI`.

dpgaspar (commented):
+1 on converting this into a CLI command

rusackas:
Agreed on the CLI approach, and either (a) locking that down by default in config, or (b) having some VERY thorough confirmations before it executes ("are you sure you want to drop your database? There's no going back. Solve this math problem to prove you're serious").

mknadh:
Thank you all. Is this looking fine?

# Command-Line Interface (CLI) for Factory Reset

## Specifications

1. **Default State**: By default, the factory-reset CLI command is disabled. It can be enabled by setting the boolean feature flag `ENABLE_FACTORY_RESET_COMMAND` to `true`.

2. **Authentication**: When the feature flag is enabled and the command is executed, the CLI will prompt the user to enter their username and password.

3. **Execution and Data Removal**: After validating the provided credentials and ensuring the user belongs to the "Admin" role, the command will perform the following actions (the order may vary):

   - Delete all users, except those with the "Admin" role.
   - Delete all roles, except "Admin", "Public", "Gamma", "Alpha", and "sql_lab".
   - Delete all dashboards.
   - Delete all slices/charts.
   - Delete all datasets.
   - Delete all databases.
   - Delete all favorites (dashboards/slices).
   - Delete all logs.
   - Delete all KeyValue entries (superset metastore cache).

4. **Audit Log**: A new audit log entry will be created, containing the username of the admin who executed the CLI command and the timestamp of execution.

mknadh:
Thank you all. 

I am transitioning the current REST API to a Command-Line Interface (CLI) with the following specifications. Is this specification for this CLI looking fine?

# Command-Line Interface (CLI) for Factory Reset

## Specifications

1. **Default State**: By default, the factory-reset CLI command is disabled. It can be enabled by setting the boolean feature flag `ENABLE_FACTORY_RESET_COMMAND` to `true`.

2. **Authentication**: When the feature flag is enabled and the command is executed, the CLI will prompt the user to enter their username and password.

3. **Execution and Data Removal**: After validating the provided credentials and ensuring the user belongs to the "Admin" role, the command will perform the following actions (the order may vary):

   - Delete all users, except those with the "Admin" role.
   - Delete all roles, except "Admin", "Public", "Gamma", "Alpha", and "sql_lab".
   - Delete all dashboards.
   - Delete all slices/charts.
   - Delete all datasets.
   - Delete all databases.
   - Delete all favorites (dashboards/slices).
   - Delete all logs.
   - Delete all KeyValue entries (superset metastore cache).

4. **Audit Log**: A new audit log entry will be created, containing the username of the admin who executed the CLI command and the timestamp of execution.

rusackas:
Sounds reasonable to me. It might just be good to add a confirmation (with a safe default) in the CLI to the effect of:

```
Please confirm that you want to:

• Delete all users, except those with the "Admin" role.
• Delete all roles, except "Admin", "Public", "Gamma", "Alpha", and "sql_lab".
• Delete all dashboards.
• Delete all slices/charts.
• Delete all datasets.
• Delete all databases.
• Delete all favorites (dashboards/slices).
• Delete all logs.
• Delete all KeyValue entries (superset metastore cache).

Proceed? [y/N]
```

michael-s-molina:
> 1. Default State: By default, the factory-reset CLI command is disabled. It can be enabled by setting the boolean feature flag ENABLE_FACTORY_RESET_COMMAND to true.

Assuming that this can only be executed by admins after layers of confirmations, I don't think we need another feature flag.

> 2. Authentication: When the feature flag is enabled and the command is executed, the CLI will prompt the user to enter their username and password.

I'm curious here. Do we currently have authentication for the CLI commands? Or are assuming that only admins can execute CLI commands? The reason I'm asking is because we already have CLI commands that require admin privileges but we don't require authentication. One example is the `superset db downgrade` that can break the application.

> 3. Execution and Data Removal: After validating the provided credentials and ensuring the user belongs to the "Admin" role, the command will perform the following actions (the order may vary):

There are many more entities such as saved queries, CSS templates, native filters, etc. Should we change the message to say that this is just a subset of entities that will be deleted? Or even revert the message saying that we'll reset to the factory state and list what we're going to preserve (the except parts)?

Are we going to preserve the example dashboards?

michael-s-molina:
FYI: The number of questions and replies to this PR clearly indicates that this should be a SIP.

rusackas:
I don't think it's too late for this to be a SIP. The PR isn't yet doing what's discussed on the thread, and even if the PR were ready, I think it's fine to put a `hold` label on it pending consensus.

mknadh:
Hi @craig-rueda, @dpgaspar - Updated the pull-request by converting the superset factory-reset _REST API_ to a _CLI command_. Please review and merge, when you get a chance. Thanks.

rusackas:
@mknadh you'll need to run the pre-commit hook (https://superset.apache.org/docs/contributing/development/#git-hooks-1) to run the linter(s) and get this to pass CI.

rusackas:
Looks like you'll need to run that pre-commit hook again. Then hopefully we can get this across the finish line :D

mknadh:
> Looks like you'll need to run that pre-commit hook again. Then hopefully we can get this across the finish line :D

Pre-commit hook is successful. Please check now.

CC: @rusackas

craig-rueda (approved):

Issue #13345, opened 2021-02-25
cannot delete user "Associated data exists, please delete them first"


maddyobrienjones:
When trying to delete a user, the error message "Associated data exists, please delete them first" appears. All charts and dashboards created by that account have been deleted. I have also set the user as inactive and deleted its role setting. I am able to edit the user but not delete it. 

I found the following similar issue which was marked as stale with no response: https://github.com/apache/superset/issues/8752 

### Expected results

Successful deletion of a user

### Actual results

"Associated data exists, please delete them first" error message

Logs when navigating to user list and trying to delete user (domain has been replaced with deploymentlink.com):
10.4.15.23 - - [25/Feb/2021:19:48:28 +0000] "GET /users/list/ HTTP/1.1" 200 26741 "deploymentlink.com/superset/welcome" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36"
10.4.15.23 - - [25/Feb/2021:19:48:34 +0000] "POST /users/delete/3 HTTP/1.1" 302 299 "deploymentlink.com/users/list" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36"
10.4.15.23 - - [25/Feb/2021:19:48:34 +0000] "GET /users/list/ HTTP/1.1" 200 26786 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36"

#### Screenshots

![Screen Shot 2021-02-25 at 2 42 56 PM](https://user-images.githubusercontent.com/35735580/109208087-e616ce80-7777-11eb-8b28-4ff3920c773b.png)

#### How to reproduce the bug

1. Go to 'List Users'
2. Navigate to user row
3. Click on 'Delete' icon

### Environment

(please complete the following information):

- superset version: `superset version`: Superset 0.999.0dev
- python version: `python --version`: 3.7.9
- node.js version: `node -v`: couldn't find

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [x] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [ ] I have reproduced the issue with at least the latest released version of superset.
[...]

PR #27223, opened 2024-02-23
feat(REST API): Adding REST API for advanced/force delete of a user #27207

mknadh:
feat(REST API): Adding REST API for advanced/force delete of a user #27207

### SUMMARY
This feature addresses existing issue since longtime. 
https://github.com/apache/superset/issues/13345

(cannot delete user "Associated data exists, please delete them first" #13345 

### ADDITIONAL INFORMATION

- [x ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ x] Introduces new feature or API

### DIFF
--- a/superset/security/api.py
+++ b/superset/security/api.py
@@ -24,10 +24,16 @@ from flask_appbuilder.security.decorators import permission_name, protect
 from flask_wtf.csrf import generate_csrf
 from marshmallow import EXCLUDE, fields, post_load, Schema, ValidationError
 
+from superset import db, security_manager
+from superset.connectors.sqla.models import SqlaTable
 from superset.commands.dashboard.embedded.exceptions import (
     EmbeddedDashboardNotFoundError,
 )
 from superset.extensions import event_logger
+from superset.key_value.models import KeyValueEntry
+from superset.models.core import Database, FavStar, Log
+from superset.models.dashboard import Dashboard
+from superset.models.slice import Slice
 from superset.security.guest_token import GuestTokenResourceType
 from superset.views.base_api import BaseSupersetApi, statsd_metrics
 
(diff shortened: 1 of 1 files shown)


john-bodley:
@mknadh I don't think deleting the data which was created by a user is actually desirable given that other users would most likely be consuming their charts, dashboards, etc. Furthermore just because a user is no longer active it can definitely be helpful (from a metadata perspective) to identify which assets they created, updated, etc.

At Airbnb we have a daily pipeline which marks previous employees as inactive Superset users and then on a monthly basis we delete unused assets and _try_ to delete inactive users—where the later only succeeds if and only if there's no assets associated with them.

superset/security/api.py
```diff
@@ -158,3 +164,64 @@ def guest_token(self) -> Response:
...
+              $ref: '#/components/responses/500'
+        """
+        try:
+            user = db.session.query(security_manager.user_model).get(pk)
+            if not user:
+                raise Exception("User not found")
+            db.session.query(FavStar).filter_by(user_id=pk).delete()
+            db.session.query(Log).filter_by(user_id=pk).delete()
+            db.session.query(KeyValueEntry).filter(KeyValueEntry.created_by_fk==pk).delete()
+            db.session.query(Slice).filter(Slice.created_by_fk==pk).delete()
```
john-bodley, line 214: Due to the foreign key constraints and non-cascading logic one simply cannot delete a table, database, etc. as there likely will be other charts, datasets, etc. which rely on these. This is a safeguard to ensure one doesn't wipe out vast swaths of data by mistake.

john-bodley (changes requested):
The TL;DR is I don't think this is a pattern we want to adopt. I understand the request, but life cycle management—be that charts, dashboards, users. etc.—is nuanced/complex and likely is something best managed outside of Superset given that policies vary greatly by institution.

rusackas:
Closing this since it's been blocked for about 9 months, and we haven't heard a comment since.

PR #27239, opened 2024-02-23
fix(alerts/reports): removing duplicate notification method options

fisjac:
### SUMMARY

When adding notification methods to an alert or report, a user can select duplicate methods (select email twice or slack twice). This PR updates the behavior to exclude notification methods that are already being used.

Additionally, the delete notification method for additional methods only appears after a user has selected a notification method. This PR changes this so that the ability to delete additional notification methods is always available, even if the method has not yet been selected.

Before:
https://www.loom.com/share/7bb024ba1d2a483eb6db44be6c137e32?sid=5620dc39-24aa-4165-8b12-39dddf6b48b5

After:
https://www.loom.com/share/c4c9c68f4f0446f18f5e4cbf1064d052?sid=3caccb4a-e2ec-4cd4-989e-d34ec31e6b7e
### TESTING INSTRUCTIONS

Ensure that within either your local superset_config.py or docker/pythonpath_dev/superset_config_docker.py files, you have a SLACK_API_TOKEN string filled in. It doesn't need to be an active api token, just any string.

Open superset and navigate to the alerts and reports section.
Create an alert or report and open the Notification method collapse section.
Observe that there are two options for notification method.
Add a new notification method.
Check the options of the newly added notification method which should exclude the previously added option.

Update the method of the first notification setting, and ensure that subsequent notification settings are being removed, and the add notification method button reappears.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: "ALERT_REPORTS": True
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -50,6 +50,7 @@ import { useCommonConf } from 'src/features/databases/state';
 import { InfoTooltipWithTrigger } from '@superset-ui/chart-controls';
 import {
   NotificationMethodOption,
+  NotificationSetting,
   AlertObject,
   ChartObject,
   DashboardObject,
@@ -393,12 +394,6 @@ const NotificationMethodAdd: FunctionComponent<NotificationMethodAddProps> = ({
         : t('Add delivery method')}
     </StyledNotificationAddButton>
   );
-};
-
-type NotificationSetting = {
-  method?: NotificationMethodOption;
-  recipients: string;
-  options: NotificationMethodOption[];
 };
 
 const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -492,15 +492,26 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     NotificationSetting[]
   >([]);
   const onNotificationAdd = () => {
-    const settings: NotificationSetting[] = notificationSettings.slice();
-    settings.push({
-      recipients: '',
-      options: allowedNotificationMethods,
-    });
+    setNotificationSettings([
+      ...notificationSettings,
+      {
+        recipients: '',
+        // options shown in the newly added notification method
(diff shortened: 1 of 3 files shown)


superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -548,8 +550,16 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
     const settings = notificationSettings.slice();
-
+    // if you've changed notification method
+    if (settings[index].method !== setting.method) {
+      // remove subsequent notification methods
+      while (settings.length - 1 > index) {
+        settings.pop();
+        setNotificationAddState('active');
+      }
+    }
```
eschutho, line 560: as much as possible, I would recommend using array.map and filter for most of this logic instead of making copies of arrays and then adding and removing elements with slice, pop, etc.. It's ok to rewrite the existing patterns here.

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -498,9 +493,16 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
   >([]);
   const onNotificationAdd = () => {
     const settings: NotificationSetting[] = notificationSettings.slice();
```
eschutho, line 495: I know this is existing code, but since we're reusing this pattern, if we need to make a copy for immutability, I would suggest using destructuring over slice. But if you can use map and filter on notificationSettings, that would be all the better. And those methods also make a copy, so would still work as a pattern for keeping notificationSettings immutable if you needed it.

geido:
Let's rebase this with master to unstuck CI

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: Changed the pattern to use filter rather than slice and pop

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: changed the pattern to use filter and reduce instead of slice

eschutho (approved):
Looks great @fisjac!

eschutho:
@fisjac I think you're going to need to rebase to bring in the other required CI checks.

PR #27252, opened 2024-02-26
chore: update docker base image from python 3.9-buster to 3.12-slim

nigzak:
### SUMMARY
The base image python 3.9 has findings which are fixed with a newer python base version
If we update to 3.12-slim 1high and 1medium would be getting fixed

image

### TESTING INSTRUCTIONS
newer base image is used

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/RELEASING/Dockerfile.from_local_tarball
+++ b/RELEASING/Dockerfile.from_local_tarball
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-FROM python:3.9-buster
+FROM python:3.12-slim
 
 RUN useradd --user-group --create-home --no-log-init --shell /bin/bash superset
 
--- a/RELEASING/Dockerfile.from_svn_tarball
+++ b/RELEASING/Dockerfile.from_svn_tarball
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-FROM python:3.9-buster
+FROM python:3.12-slim
 
 RUN useradd --user-group --create-home --no-log-init --shell /bin/bash superset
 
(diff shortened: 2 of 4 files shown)


dpgaspar (commented):
We will eventually bump our base python version to 3.10 most probably

rusackas:
Should we go to 3.10 first? Or do we want to make this big of a leap in one shot?

nigzak:
If you want I can change the pullrequest also to 3.10
3.10 would also reduce the findings

If you think this is "too bad" to swith to 3.12 I think it is totally fine to go to 3.10 and check again if relevant later

for your reverence the findings listed direct on dockerhub scan

// 3.9-buster (current used)
image

// 3.10-slim
image

// 3.11-slim
image

// 3.12-slim (latest stable available)
image

// 3.13 (NOT stable) => still has similar findings as 3.12 (even more)
image

nigzak:
FYI: even if you want to stay on python 3.9 we might at least change to "-slim" which has a high amount of findings less in comparision to "-buster" ("-buster" seems abandoned?)

// slim
image

// buster
image

mistercrunch:
Pointing out that we build on a future base here:
https://github.com/apache/superset/blob/master/scripts/build_docker.py#L152-L154

We could move the base to 3.10.x and the future to 3.12.x

I'm supportive to move to bookwork across the board too.

nigzak:
as it was pointed out to me no tickets based on public well-known and easy visible CVE shall be created anymore.
I already closed all my tickets (or they were closed) - this is the only one left.

=> please let me know if I shall close it now

mistercrunch:
Default base is now `3.10-slim-bookworm`, and while I'm not sure if the screenshot above is from DockerHub, it appears to be and I don't see any know vulnerabilities looking at DockerHub - though I may not be looking at the right place... Closing for now. If you have concerns around known vulnerabilities, please use the security@ mailing list.

PR #27253, opened 2024-02-26
feat: Add comparison time label to Big Number with Time Period plugin

kgabryje:
### SUMMARY
Add a custom control to Big Number with Time Period Comparison plugin, which adds a label with actual time range for comparison below the range for comparison selector.

https://github.com/apache/superset/assets/15073128/2829e113-c320-451f-aa27-45a24f8ec7cc

### TESTING INSTRUCTIONS
1. Create a Big Number with Time Period Comparison chart
2. Add some time filter
3. Select range for comparison
4. Verify that the label updates whenever the time filter or range for comparison changes
5. Run query
6. Verify that the label matches the time filter in second query

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/ComparisonRangeLabel.tsx
+++ b/superset-frontend/src/explore/components/controls/ComparisonRangeLabel.tsx
@@ -24,6 +24,7 @@ import moment from 'moment';
 import {
   BinaryAdhocFilter,
   css,
+  ensureIsArray,
   fetchTimeRange,
   SimpleAdhocFilter,
   t,
@@ -62,23 +63,27 @@ export const ComparisonRangeLabel = ({
   );
 
   useEffect(() => {
-    if (isEmpty(currentTimeRangeFilters) || (isEmpty(shifts) && !startDate)) {
+    const shiftsArray = ensureIsArray(shifts);
+    if (
+      isEmpty(currentTimeRangeFilters) ||
+      (isEmpty(shiftsArray) && !startDate)
+    ) {
       setLabels([]);
-    } else if (!isEmpty(shifts) || startDate) {
+    } else if (!isEmpty(shiftsArray) || startDate) {
       const promises = currentTimeRangeFilters.map(filter => {
         const startDateShift = moment(
           (filter as any).comparator.split(' : ')[0],
         ).diff(moment(startDate), 'days');
         const newShift = startDateShift
           ? [`${startDateShift} days ago`]
-          : shifts
-            ? shifts.slice(0, 1)
+          : shiftsArray
+            ? shiftsArray.slice(0, 1)
             : undefined;
 
         return fetchTimeRange(
           filter.comparator,
           filter.subject,
-          multi ? shifts : newShift,
+          multi ? shiftsArray : newShift,
         );
       });
       Promise.all(promises).then(res => {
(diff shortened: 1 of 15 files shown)


superset/views/api.py
```diff
@@ -108,6 +112,31 @@ def time_range(self, **kwargs: Any) -> FlaskResponse:
             error_msg = {"message": _("Unexpected time range: %(error)s", error=error)}
             return self.json_response(error_msg, 400)
 
+    @api
+    @handle_api_exception
+    @has_access_api
+    @rison(ger_relative_time_range_schema)
+    @expose("/v1/relative_time_range/", methods=("GET",))
+    def relative_time_range(self, **kwargs: Any) -> FlaskResponse:
```
Antonio-RiveroMartnez, line 120: Hey! Could we also accomplish the same adding a new optional field to the schema for `/v1/time_range/` and use that API instead?

superset-frontend/src/explore/components/controls/ComparisonRangeLabel.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return label ? (
+    <Tooltip title={t('Actual time range for comparison')}>
+      <span
+        css={theme => css`
+          font-size: ${theme.typography.sizes.m}px;
+          color: ${theme.colors.grayscale.base};
+        `}
+      >
+        {label}
```
Antonio-RiveroMartnez, line 67: Should we display the two periods being compared? Like:
Screenshot 2024-02-26 at 17 22 12

superset/views/api.py
kgabryje: I was thinking about this and I'm not sure, because currently the rison schema in `/time_range` is just `string`. That worked fine when we only expected a time range string there, but not sure if we could make it work with 2 fields (time range and shift)

rusackas:
I'm not up to speed on the controls changes happening here, but if they're JUST for this plugin, any new controls should be added in that plugin src. If it's a shared control and new functionality, it should be in the superset-ui package(s). In general, we should not be adding code to the superset-frontend codebase, as that should be migrated to external packages for enhanced reusability. Also, if the plugin is intended to be published as an `npm` package (which I would think we should do for all plugins) then this code can't be modified/published as part of that package release.

zhaoyongjie:
IMO, the time comparison hundred percent frontend changes enough, I post my thoughts in the PR: https://github.com/apache/superset/pull/27145/files#r1499727676

The key point of design for the date time expression is that just needs frontend to build a expression.

kgabryje:
@rusackas good points, thank you! yes this control is just for that 1 plugin. I moved it to `src/explore/components` so that we could use `dateFilterUtils`, but maybe we should simply move those utils to `@superset-ui/core`

zhaoyongjie:
@kgabryje I opened a PR for fetching multiple time range https://github.com/apache/superset/pull/27370/files

superset/views/api.py
```diff
@@ -41,6 +41,10 @@
     from superset.common.query_context_factory import QueryContextFactory
 
 get_time_range_schema = {"type": "string"}
+ger_relative_time_range_schema = {
```
eschutho, line 44: @kgabryje is this a typo?

PR #27255, opened 2024-02-26
feat: show more information when loading chart

betodealmeida:
### SUMMARY

This PR adds text to the chart spinner, informing the user:

1. When data is fetching fetched from the database (and which database)
2. When the chart is being rendered

Dashboards:

Screenshot 2024-02-26 at 15-00-00 Spinner test

Screenshot 2024-02-26 at 15-00-14 Spinner test

Charts:

Screenshot 2024-02-26 at 15-00-24 Superset

Screenshot 2024-02-26 at 15-00-45 Superset

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Chart/Chart.jsx
+++ b/superset-frontend/src/components/Chart/Chart.jsx
@@ -246,17 +246,16 @@ class Chart extends React.PureComponent {
     );
   }
 
-  renderSpinner(chartStatus, databaseName) {
-    const messages = {
-      loading: t('Fetching data from %s', databaseName),
-      success: t('Rendering chart'),
-      rendered: t('Success'),
-    };
+  renderSpinner(databaseName) {
+    const message =
+      databaseName === undefined
+        ? t('Waiting on database...')
+        : t('Waiting on %s', databaseName);
 
     return (
       <LoadingDiv>
         <Loading position="inline-centered" />
-        <MessageSpan>{messages[chartStatus]}</MessageSpan>
+        <MessageSpan>{message}</MessageSpan>
       </LoadingDiv>
     );
   }
@@ -347,7 +346,7 @@ class Chart extends React.PureComponent {
           width={width}
         >
           {isLoading
-            ? this.renderSpinner(chartStatus, databaseName)
+            ? this.renderSpinner(databaseName)
             : this.renderChartContainer()}
         </Styles>
       </ErrorBoundary>
--- a/superset-frontend/src/components/Chart/Chart.jsx
+++ b/superset-frontend/src/components/Chart/Chart.jsx
@@ -247,10 +247,9 @@ class Chart extends React.PureComponent {
   }
 
   renderSpinner(databaseName) {
-    const message =
-      databaseName
-        ? t('Waiting on %s', databaseName)
-        : t('Waiting on database...');
+    const message = databaseName
+      ? t('Waiting on %s', databaseName)
+      : t('Waiting on database...');
 
     return (
       <LoadingDiv>
--- a/superset-frontend/src/components/Chart/Chart.jsx
+++ b/superset-frontend/src/components/Chart/Chart.jsx
@@ -248,9 +248,9 @@ class Chart extends React.PureComponent {
 
   renderSpinner(databaseName) {
     const message =
-      databaseName === undefined
-        ? t('Waiting on database...')
-        : t('Waiting on %s', databaseName);
+      databaseName
+        ? t('Waiting on %s', databaseName)
+        : t('Waiting on database...');
 
     return (
       <LoadingDiv>
--- a/superset-frontend/src/components/Chart/Chart.jsx
+++ b/superset-frontend/src/components/Chart/Chart.jsx
@@ -19,6 +19,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import {
+  css,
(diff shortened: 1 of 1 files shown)


superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -232,16 +233,68 @@ class Chart extends React.PureComponent {
...
+        `}
+      >
+        <Loading position="inline-centered" />
+        <span
+          // eslint-disable-next-line theme-colors/no-literal-colors
+          css={css`
+            display: block;
+            margin: 16px auto;
+            width: fit-content;
+            color: #666666;
```
rusackas, line 259: We should use the theme colors here.

rusackas, line 259:
```suggestion

    const theme = useTheme();

    return (
      <div
        css={css`
          position: absolute;
          left: 50%;
          top: 50%;
          transform: translate(-50%, -50%);
        `}
      >
        <Loading position="inline-centered" />
        <span
          css={css`
            display: block;
            margin: 16px auto;
            width: fit-content;
            color: theme.colors.grayscale.base;
```

rusackas: You would also need to `import { useTheme } from '@superset-ui/core';`

betodealmeida: Thanks!

superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -232,16 +246,51 @@ class Chart extends React.PureComponent {
     );
   }
 
+  renderSpinner(chartStatus, databaseName) {
+    const messages = {
+      loading: t('Fetching data from %s', databaseName),
```
mistercrunch, line 251: Not sure if we got design input here, but maybe we want to make it more clear that we're waiting for the database to do its job. There's always been this perception of "superset is slow" when in fact it's the underlying datastore that we're waiting on. Maybe a more passive message like "Querying Presto".

Also curious if this is still an overlay (I think previously was transparent?) or just a opaque thing? Maybe a nice effect/animation would be slick here (?) I'm generally not too big on animation, but since this is so common and visible, it may make sense to have something subtle and sharp-looking here.

michael-s-molina: Agree with @mistercrunch here. A passive message will help improve Superset's performance perception.

superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -232,16 +246,51 @@ class Chart extends React.PureComponent {
...
       height,
       chartAlert,
       chartStatus,
+      datasource,
       errorMessage,
       chartIsStale,
       queriesResponse = [],
       width,
     } = this.props;
+    const databaseName = datasource?.database?.name;
```
michael-s-molina, line 293: According to Typescript the datasource/database can be `undefined` which will show in the interface. Can we handle this case?

michael-s-molina: Maybe use "Waiting on database..."

michael-s-molina:
Not in the scope of this PR but related. Should we make the same change on SQL Lab for the queries? @justinpark for awareness.

superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -297,20 +346,9 @@ class Chart extends React.PureComponent {
...
-                {...this.props}
-                source={this.props.dashboardId ? 'dashboard' : 'explore'}
-                data-test={this.props.vizType}
-              />
-            ) : (
-              <Loading />
-            )}
-          </div>
-          {isLoading && <Loading />}
+          {isLoading
```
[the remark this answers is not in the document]
betodealmeida: D'oh, you're right, let me fix this. Thanks!

superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -232,16 +246,50 @@ class Chart extends React.PureComponent {
     );
   }
 
+  renderSpinner(databaseName) {
+    const message =
```
michael-s-molina, line 250: Maybe `const message = databaseName ? t('Waiting on database...') : t('Waiting on %s', databaseName);` just to be safe as I don't know if the database name can be `null` or empty.

superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -232,16 +246,50 @@ class Chart extends React.PureComponent {
...
+      databaseName === undefined
+        ? t('Waiting on database...')
+        : t('Waiting on %s', databaseName);
```
betodealmeida, line 253:
```suggestion
      databaseName
        ? t('Waiting on %s', databaseName)
        : t('Waiting on database...');
```

betodealmeida, line 253: Let's do it like this since it gives more context for the translation:

```suggestion
    const message =
      databaseName
        ? t('Waiting on %s', databaseName)
        : t('Waiting on database...');
```

eschutho (approved):
Looks great @betodealmeida!

PR #27258, opened 2024-02-26
fix: Sorting charts/dashboards makes the applied filters ineffective

### LINKED ISSUE #14292: Sorting Charts/Dashboards makes the applied filters ineffective
Charts/Dashboard's card view: After applying any filter (e.x Favorite=Yes); if we select any sorting option from the sort dropdown menu, the filtered list get reset and the entire list is shown.

### Expected results

If any filter is applied to the list, then sorting that list should retain the applied filters.

### Actual results

The applied filter to the list gets reset to its original state after selecting any sort criteria.

In the video below, if we apply a filter (Favorite = Yes) to the list of charts, then selecting sort = Alphabetical will reset the filters and show the list with all charts (Starred and Un-Starred)

#### Screenshots

https://user-images.githubusercontent.com/57817186/115641067-353b3480-a2cd-11eb-8506-6a9607091393.mov

#### How to reproduce the bug

1. Go to Charts from toolbar and select the 'Card' view
2. Apply a filter by selecting Favorite = Any from the list headers. Make sure you have few favorite/starred charts to see the results
3. Select Sort = Alphabetical
4. You will see that the entire list is shown even though we have applied the filter to show only favorite charts 

### Environment

- superset version: commit# `c9a755f`
- python version: `python 2.7.16`
- node.js version: `v13.11.0`

### Checklist

michael-s-molina:
### SUMMARY
Fixes a bug in the charts and dashboards list where selected filters were not being applied when changing the sorting order. It also updates the URL with the applied sorting to preserve the order in case of a refresh.

Fixes https://github.com/apache/superset/issues/14292. 

https://github.com/apache/superset/assets/70410625/6661ba2e-2114-4b18-9677-94e91e35cbbf

https://github.com/apache/superset/assets/70410625/8e983e21-f94b-4d3d-868b-e1fe53d204ee

### TESTING INSTRUCTIONS
Check the original issue for instructions.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
@@ -173,6 +173,13 @@ describe('Charts list', () => {
       orderAlphabetical();
       cy.getBySel('styled-card').first().contains('% Rural');
     });
+
+    it('should preserve other filters when sorting', () => {
+      cy.getBySel('styled-card').should('have.length', 25);
+      setFilter('Type', 'Big Number');
+      setFilter('Sort', 'Least recently modified');
+      cy.getBySel('styled-card').should('have.length', 3);
+    });
   });
 
   describe('common actions', () => {
--- a/superset-frontend/src/components/ListView/types.ts
+++ b/superset-frontend/src/components/ListView/types.ts
@@ -23,8 +23,6 @@ export interface SortColumn {
   desc?: boolean;
 }
 
-export type SortColumns = SortColumn[];
-
 export interface SelectOption {
   label: string;
   value: any;
@@ -84,7 +82,7 @@ export interface FilterValue {
 export interface FetchDataConfig {
   pageIndex: number;
   pageSize: number;
-  sortBy: SortColumns;
+  sortBy: SortColumn[];
   filters: FilterValue[];
 }
 
--- a/superset-frontend/cypress-base/cypress/e2e/dashboard_list/list.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/dashboard_list/list.test.ts
@@ -117,6 +117,13 @@ describe('Dashboards list', () => {
       orderAlphabetical();
       cy.getBySel('styled-card').first().contains('Supported Charts Dashboard');
     });
+
+    it('should preserve other filters when sorting', () => {
+      cy.getBySel('styled-card').should('have.length', 5);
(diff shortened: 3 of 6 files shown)


superset-frontend/src/components/ListView/types.ts
```diff
@@ -84,7 +82,7 @@ export interface FilterValue {
 export interface FetchDataConfig {
   pageIndex: number;
   pageSize: number;
-  sortBy: SortColumns;
+  sortBy: SortColumn[];
```
villebro, line 85: Is there some reason why this is being changed from `SortColumns` to `SortColumn[]` here? While on this topic, I personally don't like these types of pluralization aliases, as it makes it seem like it's something more than just an array of the singular type. But if we want to refrain from making these types of sweeping changes during the 4.0 hardening phase, I'd recommend leaving this as-is for now, and following up later with a PR that removes the plural type.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -173,6 +173,13 @@ describe('Charts list', () => {
       orderAlphabetical();
       cy.getBySel('styled-card').first().contains('% Rural');
     });
+
+    it('should preserve other filters when sorting', () => {
+      cy.getBySel('styled-card').should('have.length', 25);
+      setFilter('Type', 'Big Number');
+      setFilter('Sort', 'Least recently modified');
+      cy.getBySel('styled-card').should('have.length', 3);
+    });
```
villebro, line 182: ❤️

villebro (approved):
Tested to work as expected. LGTM with a comment about the plural type alias. Kudos for squashing a 2 year old bug! ❤️

rusackas (approved):

superset-frontend/src/components/ListView/types.ts
giftig: Personally I find following the "boy scout" approach is usually a good way of driving incremental change and reduces the pain of having to invest a chunk of time and effort into doing big bang changes. That's especially true if there are concerns over adequate test coverage for the "big bang" change, for example.

Though that's assuming there's already an agreed policy to remove aliases like these, otherwise the risk is different groups enacting different boy scout policies, which might be more of a concern in a large open source project like this one than I'm typically used to.

giftig (approved):

PR #27260, opened 2024-02-26
fix(alerts/reports): implementing custom_width as an Antd number input

fisjac:
### SUMMARY

The behavior of the custom_width input on alerts and reports used to be a numeric input. This PR implements the antd NumberInput behavior. It also implements min and max at 600 and 2400x to prevent users from inputting invalid values.

Before:
https://www.loom.com/share/05786eb33da74f3385f8e08fc0bb6427?sid=21aa2e2b-5cc4-4075-aae4-72cc9ce13217

After:
https://www.loom.com/share/cadf7b979abb48e79afd75893fb80f0e?sid=81bae2c6-b553-4725-a8dc-f4104820f665

### TESTING INSTRUCTIONS

Create or open an existing alert or report. Ensure that the custom width input is a numeric input.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERTS_REPORTS=true
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -1549,7 +1549,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
                 <InputNumber
                   type="number"
                   name="custom_width"
-                  value={currentAlert?.custom_width ?? null}
+                  value={currentAlert?.custom_width || undefined}
                   min={600}
                   max={2400}
                   placeholder={t('Input custom width in pixels')}
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -873,7 +873,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     updateAlertState(name, parsedValue);
   };
 
-  const onCustomWidthChange = (value: number | null) => {
+  const onCustomWidthChange = (value: number | null | undefined) => {
     updateAlertState('custom_width', value);
   };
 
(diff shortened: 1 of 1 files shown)


superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -1542,12 +1546,14 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
             >
               <div className="control-label">{t('Screenshot width')}</div>
               <div className="input-container">
-                <Input
+                <InputNumber
                   type="number"
                   name="custom_width"
-                  value={currentAlert?.custom_width || ''}
+                  value={currentAlert?.custom_width ?? null}
```
michael-s-molina, line 1552: Given that we're setting 600 as the minimum value, should accept `null` values here or always start with 600?

geido: I think it makes sense to have 600 as a default value here. Perhaps we could export in a const for better visibility

fisjac: On the backend, custom_width as a field is nullable, and when null, it defaults to the WEBDRIVER_WINDOW in config.py

Default value is:
```
 WEBDRIVER_WINDOW = {
    "dashboard": (1600, 2000),
    "slice": (3000, 1200),
    "pixel_density": 1,
} 
```

eschutho: @fisjac for the ts error here, I think you're going to want the default to be `undefined`
`error TS2322: Type 'number | null' is not assignable to type 'number | undefined'.`

geido: This will never be below 600 though right and it won't ever be null or undefined for that reason too?

geido (approved):

PR #27262, opened 2024-02-27
fix(Alerts & Reports): Fixing bug that resets cron value to default when empty  

fisjac:
### SUMMARY

In either alerts or reports, when backspacing the value in cron schedule input field, the input will repopulate with the default cron value when the input becomes an empty string.

Before: 
https://github.com/apache/superset/assets/41238731/30be8978-5122-410e-84ec-fd170100ac1c

After:
https://www.loom.com/share/31dd1e0b67174cc38de376419e99c6ff?sid=0261ed05-48f4-49af-acae-7fe6280141ad

### TESTING INSTRUCTIONS

Navigate to an existing or new alert/report.
Open the Schedule section.
Under schedule type, select CRON schedule
Where it says a schedule (default is "0 0 * * *"), backspace until the field is empty.
If it remains empty, that is desired behavior.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERT_REPORT=true
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -1581,7 +1581,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
           key="schedule"
         >
           <AlertReportCronScheduler
-            value={currentAlert?.crontab || ALERT_REPORTS_DEFAULT_CRON_VALUE}
+            value={currentAlert?.crontab || ''}
             onChange={newVal => updateAlertState('crontab', newVal)}
           />
           <StyledInputContainer>


superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -1581,7 +1581,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
           key="schedule"
         >
           <AlertReportCronScheduler
-            value={currentAlert?.crontab || ALERT_REPORTS_DEFAULT_CRON_VALUE}
+            value={currentAlert?.crontab || ''}
```
michael-s-molina, line 1584: I don't have much context here but this does look problematic. Is the default being completely ignored? I don't see other uses of the variable. Maybe @eschutho can help here.

fisjac: The default is being set when a new alert/report is generated, so the user should always open the modal to a valid schedule input to start.

sadpandajoe:
@michael-s-molina heads up this is part of 4.0

michael-s-molina (approved):

superset-frontend/src/features/alerts/AlertReportModal.tsx
michael-s-molina: Given that we are validating empty CRON values, this change is fine.

PR #27274, opened 2024-02-27
fix(docs): removing meta tag CSP, poking more holes in htaccess

rusackas:
### SUMMARY

The site is not displaying things properly... and it might be because of the `htaccess` file OR the `meta` CSP tag... both are present. This removes the CSP  meta tag, and adds more holes in the `htaccess` file. Let's see if it works! Sadly, more PRs to come, I would guess.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -261,7 +261,6 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
-      metadata: [{name: 'Content-Security-Policy', content: "default-src 'self'; frame-src 'https://calendar.google.com/' 'https://preset.io/' 'https://sidebar.bugherd.com/';"}],
     }),
   scripts: [
     '/script/matomo.js',
--- a/docs/static/.htaccess
+++ b/docs/static/.htaccess
@@ -31,5 +31,5 @@ style-src 'self' https://fonts.googleapis.com; \
 script-src-elem 'self' https://www.googletagmanager.com https://www.google-analytics.com; \
 style-src-elem 'self' https://fonts.googleapis.com; \
 font-src 'self' https://fonts.gstatic.com; \
-frame-src 'self' https://calendar.google.com https://preset.io https://sidebar.bugherd.com 'https://unpkg.com; \
+frame-src 'self' https://calendar.google.com https://preset.io https://sidebar.bugherd.com https://unpkg.com; \
 "
--- a/docs/static/.htaccess
+++ b/docs/static/.htaccess
@@ -22,4 +22,14 @@ RewriteRule ^(.*)$ https://superset.apache.org/$1 [R,L]
 RewriteCond %{HTTP_HOST} ^superset.incubator.apache.org$ [NC]
 RewriteRule ^(.*)$ https://superset.apache.org/$1 [R=301,L]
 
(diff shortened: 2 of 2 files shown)


docs/docusaurus.config.js
```diff
@@ -261,7 +261,6 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
-      metadata: [{name: 'Content-Security-Policy', content: "default-src 'self'; frame-src 'https://calendar.google.com/' 'https://preset.io/' 'https://sidebar.bugherd.com/';"}],
```
rusackas, line 264: the quotes on URLs are part of the problem, I suspect... but we need more holes poked for other stuff as well since we're apparently overriding some apache server deafault.

docs/static/.htaccess
```diff
@@ -22,4 +22,14 @@ RewriteRule ^(.*)$ https://superset.apache.org/$1 [R,L]
 RewriteCond %{HTTP_HOST} ^superset.incubator.apache.org$ [NC]
 RewriteRule ^(.*)$ https://superset.apache.org/$1 [R=301,L]
 
-Header set Content-Security-Policy "default-src 'self'; frame-src 'https://calendar.google.com/' 'https://preset.io/' 'https://sidebar.bugherd.com/' 'https://unpkg.com/';"
+Header set Content-Security-Policy "default-src 'self'; img-src *;"
+
+Header set Content-Security-Policy "default-src 'self'; \
+script-src 'self'; \
+img-src 'self' https://static.scarf.sh *; \
```
rusackas, line 29: the wildcard here isn't ideal. When the bulk of the site looks OK, we can tighten the screws here.

docs/static/.htaccess
```diff
@@ -22,4 +22,14 @@ RewriteRule ^(.*)$ https://superset.apache.org/$1 [R,L]
 RewriteCond %{HTTP_HOST} ^superset.incubator.apache.org$ [NC]
 RewriteRule ^(.*)$ https://superset.apache.org/$1 [R=301,L]
 
-Header set Content-Security-Policy "default-src 'self'; frame-src 'https://calendar.google.com/' 'https://preset.io/' 'https://sidebar.bugherd.com/' 'https://unpkg.com/';"
+Header set Content-Security-Policy "default-src 'self'; img-src *;"
+
+Header set Content-Security-Policy "default-src 'self'; \
+script-src 'self'; \
```
rusackas, line 28: Trying to avoid 'unsafe-inline' wherever possible, but we may well have to resort to it here or elsewhere.

docs/static/.htaccess
```diff
@@ -22,4 +22,14 @@ RewriteRule ^(.*)$ https://superset.apache.org/$1 [R,L]
...
+img-src 'self' https://static.scarf.sh *; \
+style-src 'self' https://fonts.googleapis.com; \
+script-src-elem 'self' https://www.googletagmanager.com https://www.google-analytics.com; \
+style-src-elem 'self' https://fonts.googleapis.com; \
+font-src 'self' https://fonts.gstatic.com; \
+frame-src 'self' https://calendar.google.com https://preset.io https://sidebar.bugherd.com 'https://unpkg.com; \
```
rusackas, line 34: I don't think 'self' is needed here, but no harm no foul.

craig-rueda (approved):

PR #27282, opened 2024-02-28
fix(ci): docker builds don't work from remote forks

mistercrunch:
### SUMMARY
Fix issues relate to multi-platform builds not working with `--load`.

Lots of intricacies around multi-platform builds somehow (doesn't work on OSx, doesn't work with `--load`, and somehow can't `--push` from remote forks, ...)


michael-s-molina (approved):
LGTM

dpgaspar (approved):

PR #27285, opened 2024-02-28
fix(plugin-chart-echarts): calculate Gauge Chart intervals correctly when min value is set

### LINKED ISSUE #21328: Gauge chart shows incorrect interval colors when using minimum value
When using the Gauge chart, it is possible to define interval bounds and interval colors, e.g. for marking a range of "good" values. It is also possible to set minimum and maximum values to define the left and right boundaries of the Gauge.

When interval bounds and colors are in place and then a min value is set, the position of the interval color boundary does not change.

#### How to reproduce the bug

1. Create a new Gauge chart, e.g. with exported_stats example dataset
2. Select AVG(daily_members_posting_messages) as Metric
3. On Customize tab, set INTERVAL BOUNDS to 10, 16.53 and INTERVAL COLORS to 3, 6
4. Set MIN to 8
5. Observe how the boundary between green and red does not move when MIN is updated

### Expected results

The color boundary should move to a new position depending on the value of MIN.

### Actual results

The color boundary stays where it was when MIN was 0.

#### Screenshots

Without MIN value set:
Screenshot_20220905_115353

With MIN value set:
Screenshot_20220905_115410

### Environment

- browser type and version: Firefox 104.0.1
- superset version: 2.0.0

### Checklist

- [N/A] I have checked the superset logs for python stacktraces and included it here as text if there are any.

goto-loop:
### SUMMARY

Fixes #21328.

If you create a Gauge Chart with interval bounds and a minimum value set, the color boundaries are placed incorrectly. This PR changes the boundary calculation to take the minimum value into account during normalization by using the formula described [here](https://stats.stackexchange.com/a/70807).

I've also updated the part of the test file where the interval calculation gets tested.

Before (boundary between green and red should be at value 10 but isn't):
Screenshot from 2024-02-27 16-58-21
After:
Screenshot from 2024-02-27 16-59-06

### TESTING INSTRUCTIONS

1. Create a new Gauge chart, e.g. with exported_stats example dataset
2. Select AVG(daily_members_posting_messages) as Metric
3. On Customize tab, set INTERVAL BOUNDS to 10, 16.53 and INTERVAL COLORS to 3, 6
4. Set MIN to 8
5. Now the boundary between green and red gets updated with the rest of the chart

### ADDITIONAL INFORMATION

- [x] Has associated issue: #21328 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/test/Gauge/transformProps.test.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/test/Gauge/transformProps.test.ts
@@ -16,8 +16,15 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { ChartProps, SqlaFormData, supersetTheme } from '@superset-ui/core';
-import transformProps from '../../src/Gauge/transformProps';
+import {
+  CategoricalColorNamespace,
+  ChartProps,
+  SqlaFormData,
+  supersetTheme,
+} from '@superset-ui/core';
+import transformProps, {
+  getIntervalBoundsAndColors,
+} from '../../src/Gauge/transformProps';
 import { EchartsGaugeChartProps } from '../../src/Gauge/types';
 
 describe('Echarts Gauge transformProps', () => {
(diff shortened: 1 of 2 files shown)


villebro (commented):
I agree with the change, and I was able to validate the change in local testing. However, it would be nice if we could add a few unit tests for `setIntervalBoundsAndColors`, as the `transformProps` test doesn't seem to actually be catching this error.

Also, this is not your making, but calling a function `set` that is in fact not setting, but returning something, feels misleading. I think `getInvervalBoundsAndColors` would be more appropriate.

rusackas:
I was hoping to spin up a test environment here, but the docker build is failing. A rebase might fix that if warranted. Thanks @villebro for testing locally.

goto-loop:
Thanks @villebro for the feedback! I've renamed the function and added some tests specifically for it, please have a look and let me know what you think.

@rusackas I've rebased the branch, maybe you can try the docker build again.

villebro:
Thanks @goto-loop for the great tests and addressing the function name! Fingers crossed CI passes 🤞

superset-frontend/plugins/plugin-chart-echarts/test/Gauge/transformProps.test.ts
```diff
@@ -342,3 +350,43 @@ describe('Echarts Gauge transformProps', () => {
     );
   });
 });
+
+describe('getIntervalBoundsAndColors', () => {
+  it('should generate correct interval bounds and colors', () => {
```
villebro, line 355: nit/FYI: we're moving away from `describe` with nested `it`s towards simple `test` blocks. But we can convert these in a follow-up refactor pass.

villebro (approved):
LGTM, thanks for this great fix and quick follow-up on the review comments ❤️

superset-frontend/plugins/plugin-chart-echarts/test/Gauge/transformProps.test.ts
goto-loop: Good point, I wasn't aware of that. Will keep it in mind for the next PR :slightly_smiling_face:

PR #27290, opened 2024-02-28
docs(import_datasources): Remove legacy documentation and update current use

ddxv:
### SUMMARY
The documentation was still referencing the "legacy_import_datasources". I updated to the current expected use for ZIP files and `--username` option based on #27154.

I have removed the documentation which was for `legacy_import_datasources` based on the assumption that this is not expected to be supported. 

If we would like to keep supporting the `legacy_import_datasources` I can add it back into the docs with it's own titled section. 

Please let me know what works best.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/miscellaneous/importing-exporting-datasources.mdx
+++ b/docs/docs/miscellaneous/importing-exporting-datasources.mdx
@@ -61,42 +61,17 @@ superset export_datasource_schema
 
 As a reminder, you can use the `-b` flag to include back references.
 
-### Importing Datasources from YAML
+### Importing Datasources
 
-In order to import datasources from a YAML file(s), run:
-
-```
-superset import_datasources -p <path or filename>
-```
-
-If you supply a path all files ending with **yaml** or **yml** will be parsed. You can apply
-additional flags (e.g. to search the supplied path recursively):
+In order to import datasources from a ZIP file, run:
 
 ```
-superset import_datasources -p <path> -r
+superset import_datasources -p <path / filename>
 ```
 
-The sync flag **-s** takes parameters in order to sync the supplied elements with your file. Be
+The username flag **-u** sets the user used for the datasource import. The default is 'admin'. in order to sync the supplied elements with your file. Be
 careful this can delete the contents of your meta database. Example:
 
 ```
-superset import_datasources -p <path / filename> -s columns,metrics
-```
-
-This will sync all metrics and columns for all datasources found in the `<path /filename>` in the
-Superset meta database. This means columns and metrics not specified in YAML will be deleted. If you
-would add tables to columns,metrics those would be synchronised as well.
-
-If you don’t supply the sync flag (**-s**) importing will only add and update (override) fields.
-E.g. you can add a verbose_name to the column ds in the table random_time_series from the example
-datasets by saving the following YAML to file and then running the **import_datasources** command.
-
-```
-databases:
-- database_name: main
-  tables:
-  - table_name: random_time_series
-    columns:
-    - column_name: ds
-      verbose_name: datetime
+superset import_datasources -p <path / filename> -u 'admin'
 ```
--- a/docs/docs/miscellaneous/importing-exporting-datasources.mdx
+++ b/docs/docs/miscellaneous/importing-exporting-datasources.mdx
@@ -79,7 +79,7 @@ superset import_datasources -p <path / filename> -u 'admin'
 
 #### From older versions of Superset to current version
 
-When using Superset version 4.x.x to import from an older version (2.x.x or 3.x.x) importing is supported as the command `legacy_import_datasources` and expects a JSON or directory of JSONs. The options are `-r` for recursive and `-u` for specifying a user. Example:
+When using Superset version 4.x.x to import from an older version (2.x.x or 3.x.x) importing is supported as the command `legacy_import_datasources` and expects a JSON or directory of JSONs. The options are `-r` for recursive and `-u` for specifying a user. Example of legacy import without options:
 
 ```
 superset legacy_import_datasources -p <path or filename>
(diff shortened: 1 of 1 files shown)


sfirke:
I'm unfamiliar with the technical aspect of this but appreciate you keeping the docs updated!  My only concern would be if this only applies to the latest version (I see the PR is tagged 4.0 / 3.1) and we would want to keep the older info for users of older versions.   Did this documentation even work before your patch, it sounds like it was broken?  If it worked for a reasonably recent version (say 2.0.0 or later) I'd be inclined to preserve it.

Ideally there'd be a way to link to an older static version of the docs, but I don't think that's possible, so anything useful could be preserved in a section labeled as such.

ddxv:
@sfirke thanks for the input. I went ahead and undid the deletion of the legacy documentation, and instead labled it as `Legacy Import Datasources`.

 I also added clarification for using the `VERSIONED_EXPORT` which was used in an if / else switch in versions 2/3 for accepting either the ZIP file or YAML file(s).

docs/docs/miscellaneous/importing-exporting-datasources.mdx
```diff
@@ -61,15 +61,29 @@ superset export_datasource_schema
...
+
+The optional username flag **-u** sets the user used for the datasource import. The default is 'admin'. Example:
+
+```
+superset import_datasources -p <path / filename> -u 'admin'
+```
+
+### Legacy Importing Datasources from YAML
+
+In Superset 4 the older way of datasource importing is supported as `legacy_import_datasources`. In older versions (2 & 3) of Superset, to switch between ZIP and YAML files, the feature flag `VERSIONED_EXPORT` is used. When `VERSIONED_EXPORT` is `True`, `import_datasources` expects a ZIP file, otherwise YAML. Example:
```
sfirke, line 80:
```suggestion
In Superset version 4.x.x the older way of datasource importing is supported as the command `legacy_import_datasources`. In older versions (2.x.x & 3.x.x) of Superset, to switch between ZIP and YAML files, the feature flag `VERSIONED_EXPORT` is used. When `VERSIONED_EXPORT` is `True`, `import_datasources` expects a ZIP file, otherwise YAML. Example:
```

sfirke:
Thanks for the changes.  To check my understanding, are there three scenarios to cover:

1. Importing 4.0.0+ dashboards into a 4.0.0+ instance = use `import_datasources`
2. Importing <4.0.0 dashboards into a 4.0.0+ instance = use `legacy_import_datasources`
3. Importing <4.0.0 dashboards into a <4.0.0 instance = use `import_datasources`

Is that right?  Right now 1. is covered, but the legacy piece could be clearer in how it covers both 2 & 3.

Sorry if this feels like I'm belaboring this or if I'm misunderstanding.

ddxv:
> Thanks for the changes. To check my understanding, are there three scenarios to cover:
> 
>     1. Importing 4.0.0+ dashboards into a 4.0.0+ instance = use `import_datasources`
> 
>     2. Importing <4.0.0 dashboards into a 4.0.0+ instance = use `legacy_import_datasources`
> 
>     3. Importing <4.0.0 dashboards into a <4.0.0 instance = use `import_datasources`
> 
> 
> Is that right? Right now 1. is covered, but the legacy piece could be clearer in how it covers both 2 & 3.
> 
> Sorry if this feels like I'm belaboring this or if I'm misunderstanding.

Glad you brought that up, I hadn't considered upgrading across versions as I had gotten into this from just deploying a specific versions.  Your 3 points are definitely valid, so I added a couple more sections to the Legacy section to cover them.

docs/docs/miscellaneous/importing-exporting-datasources.mdx
```diff
@@ -61,15 +61,39 @@ superset export_datasource_schema
...
+
+When using Superset version 4.x.x to import from an older version (2.x.x or 3.x.x) importing is supported as the command `legacy_import_datasources` and expects a JSON or directory of JSONs. The options are `-r` for recursive and `-u` for specifying a user. Example:
+
+```
+superset legacy_import_datasources -p <path or filename>
+```
+
+#### From older versions of Superset to older versions
+
+When using an older Superset version (2.x.x & 3.x.x) of Superset, ZIP and YAML files are supported, and to switch between them the feature flag `VERSIONED_EXPORT` is used. When `VERSIONED_EXPORT` is `True`, `import_datasources` expects a ZIP file, otherwise YAML. Example:
```
sfirke, line 90:
```suggestion
When using an older Superset version (2.x.x & 3.x.x) of Superset, the command is `import_datasources`.  ZIP and YAML files are supported and to switch between them the feature flag `VERSIONED_EXPORT` is used. When `VERSIONED_EXPORT` is `True`, `import_datasources` expects a ZIP file, otherwise YAML. Example:
```

docs/docs/miscellaneous/importing-exporting-datasources.mdx
```diff
@@ -61,15 +61,39 @@ superset export_datasource_schema
...
+
+```
+superset import_datasources -p <path / filename> -u 'admin'
+```
+
+### Legacy Importing Datasources
+
+#### From older versions of Superset to current version
+
+When using Superset version 4.x.x to import from an older version (2.x.x or 3.x.x) importing is supported as the command `legacy_import_datasources` and expects a JSON or directory of JSONs. The options are `-r` for recursive and `-u` for specifying a user. Example:
```
sfirke, line 82: You say the options are `-r` and `-u` but then the example has `-p` -- is that correct?  Just checking that `-p` is supported here and that you don't want to demo `-r` or `-u` in the command.

sfirke (commented):
This is looking good!  Just one little suggestion and a little question.  Thank you @ddxv for addressing the various scenarios one can face here 🙏

docs/docs/miscellaneous/importing-exporting-datasources.mdx
ddxv: Hello sfirke, the command with -p is correct as that is a required parameter, but it is confusing in this example to have not included either the optionals. I think clarifying this will help.

sfirke (approved):

sfirke:
thank you @ddxv for this great update to the docs! 🙌

PR #27295, opened 2024-02-29
fix(sqllab): invalid dump sql shown after closing tab

justinpark:
### SUMMARY

When `SqllabBackendPersistence` is enabled, the tab_history only populates the most recent active tab. As a result, if a user closes a tab upon first loading, the tab_history disappears, yet the UI component still loads the latest tab. This unintentionally displays the "unloaded" tabstateview, which consists of the defaultSqlEditor value that includes a placeholder SQL like `SELECT * FROM WHERE`.
This recent commit addresses and rectifies this issue, ensuring the tabstateview loads correctly when an unloaded condition tab is activated. 

Before:

https://github.com/apache/superset/assets/1392866/28734ec7-cc48-497b-a9e2-b3ec85255897

After:

https://github.com/apache/superset/assets/1392866/18aab574-25ff-44dd-8255-1353d2bad4fe

### TESTING INSTRUCTIONS
Set SqllabBackendPersistence enable
Load SQL Lab which includes more than one tabs
Close the current tab immediately, and then observe if the query editor displays the correct SQL population.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
+++ b/superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
@@ -54,7 +54,7 @@ import Mousetrap from 'mousetrap';
 import Button from 'src/components/Button';
 import Timer from 'src/components/Timer';
 import ResizableSidebar from 'src/components/ResizableSidebar';
-import { AntdDropdown, AntdSwitch } from 'src/components';
+import { AntdDropdown, AntdSwitch, Skeleton } from 'src/components';
 import { Input } from 'src/components/Input';
 import { Menu } from 'src/components/Menu';
 import Icons from 'src/components/Icons';
@@ -77,6 +77,7 @@ import {
   setActiveSouthPaneTab,
   updateSavedQuery,
   formatQuery,
+  switchQueryEditor,
 } from 'src/SqlLab/actions/sqlLab';
 import {
   STATE_TYPE_MAP,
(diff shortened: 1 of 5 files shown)


superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
```diff
@@ -847,7 +859,17 @@ const SqlEditor: React.FC<Props> = ({
...
-      {showEmptyState ? (
+      {shouldLoadQueryEditor ? (
+        <div
+          data-test="sqlEditor-loading"
+          css={css`
+            flex: 1;
+            padding: ${theme.gridUnit * 4}px;
+          `}
+        >
+          <Skeleton active />
```
michael-s-molina, line 870: Given that the `Skeleton` refers to the whole query editor, could you make the spinner centralized? Right now, it's aligned to the left.

michael-s-molina (approved):
Thanks for the fix @justinpark. I left a non-blocking comment.

superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
justinpark: The skeleton appears visually better when aligned to the left. (By the way, the attached video used a different loading indicator.)

justinpark: _DEV__Superset

PR #24329, opened 2023-06-08
feat(sqllab): add shortcut for run current sql

justinpark:
### SUMMARY
The manual selection of the current query is time-consuming. (Longer the query more time it takes to select manually)
This commit creates a new shortcut(`ctrl + shift + enter`) which automatically detects and selects the range of the current query and then run the selection in order to reduce the manual selection work.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

After:

https://github.com/apache/superset/assets/1392866/d146c9a5-5819-4fe2-af22-6281f3bb0de6

Before:

https://github.com/apache/superset/assets/1392866/45f22a20-32b3-4eab-b075-95bf4e108df6

### TESTING INSTRUCTIONS
Hit "Ctrl+Shift+Enter" on sql editor

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27314, opened 2024-03-01
feat(sqllab): Allow running query without highlighting the code

soniagtm:
### SUMMARY

When a user enters multiple queries in one tab and wants to execute each of them, they would have to highlight the SQL they want to execute (in previous versions), which would not be ideal for extremely long queries. To solve this problem, I created the getCurrentQuery function to **allow executing each query without the need to highlight the specific query they wish to run**. Instead, users can **place the cursor at the intended query and press the RUN button**.

It's important to note that **queries must be separated by either at least one blank line or a semicolon**, and **no blank lines are allowed within the same query**. When the cursor is placed at the blank lines between two queries, the upper query will be executed. This also handles the case when a semicolon is part of the query.

This function was actually created a long time ago, and I found that in newer versions, there already exists a shortcut key to execute the current query. I wanted to share this implementation in case it might be useful or maybe created as another shortcut key.

**Before**

https://github.com/apache/superset/assets/96282005/ce96ec0b-9462-4f6d-9c49-01a91fe6b947

**After**

https://github.com/apache/superset/assets/96282005/755150a8-a985-4308-90a6-d584f5a3a711

### TESTING INSTRUCTIONS

1. Go to SQL Lab
2. Enter multiple queries in one tab (separate with semicolon or with a line break)
3. Move cursor to desired query to run
4. Press Run button

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -131,6 +131,126 @@ const fieldConverter = mapping => obj =>
 export const convertQueryToServer = fieldConverter(queryServerMapping);
 export const convertQueryToClient = fieldConverter(queryClientMapping);
 
+function splitQueries(input) {
+  let inQuote = false;
+  let inLineComment = false;
+  let inBlockComment = false;
+  const queries = [];
+  let lastIndex = 0;
+
+  for (let i = 0; i < input.length; i += 1) {
+    const char = input[i];
+    const lookahead = input[i + 1];
+
+    // Handle quotation marks
+    if ((char === "'" || char === '"') && !inBlockComment && !inLineComment) {
+      inQuote = !inQuote;
+    }
+
+    // Handle line comment: --
+    else if (!inQuote && !inBlockComment && char === '-' && lookahead === '-') {
+      inLineComment = true;
+    }
+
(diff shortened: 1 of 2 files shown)


justinpark:
@soniagtm Thank you for your suggestion. I appreciate your input, but I do have a few reservations regarding these changes.
Firstly, this modification will alter the behavior of the "Run" command. Users who employ multiple SQLs in a single run might experience some regression due to this change.
Secondly, I have already introduced a similar feature that runs the current query using a shortcut(ctrl + shift + enter #24329). This means users wishing to run a query without highlighting can achieve the same results using this shortcut.
I hope this information helps clarify my approach.
I'm always open to further discussion.

Thanks,

soniagtm:
Hi @justinpark,

Thank you for sharing your thoughts and concerns. Considering your reservations, I completely understand the importance of maintaining the current functionality and user experience. 

Regarding the existing feature for running current query, it would indeed be beneficial if it could support executing queries without relying solely on semicolons, as is commonly seen in other tools. Additionally, it would be helpful to address cases where semicolons are part of the query. Perhaps these could be explored as potential areas for improvement in the future.

Thank you once again for your feedback.

PR #27351, opened 2024-03-01
fix: Migration for single metric in Big Number with Time Comparison

kgabryje:
### SUMMARY
Due to changing `metrics` control to `metric` in Big Number with Time Comparison chart, the existing charts were broken due to missing `metric` in form data. This PR adds a migration which takes the first item from `metrics` and applies it to the new `metric` control.

### TESTING INSTRUCTIONS
1. Revert commit fd4f9ac (`feat: Use standardized controls in Big Number with Time Comparison`)
2. Create a Big Number with Time Comparison chart and save it
3. Un-revert the commit
4. Verify that now the chart is broken - no metric is added
5. Run the migration
6. Verify that the chart now works and `metric` control has a value

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
Current: 0.13 s
10+: 0.12 s
100+: 0.12 s
1000+: 0.20 s

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
@@ -143,7 +143,6 @@ export default function transformProps(chartProps: ChartProps) {
     width,
     height,
     data,
-    metric,
     metricName,
     bigNumber,
     prevNumber,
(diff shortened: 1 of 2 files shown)


Antonio-RiveroMartnez (approved):

mistercrunch:
We should move all to `metrics` and delete `metric` forever!

mistercrunch:
Also - apologies - super early in the history of Superset (pre-React!) I created multiple metrics-related control instead of creating a more versatile one. I think back then it was backend generate by wtforms and they had different components for single/multiple select...

sadpandajoe:
🏷️ preset:2024.8

PR #27364, opened 2024-03-01
fix(API): Updating assets via the API should preserve ownership configuration

Vitor-Avila:
### SUMMARY
The current behavior for updating assets via the API is inconsistent:
* When updating dashboards and reports, the `owners` information wouldn't be modified unless it's part of the update payload
* When updating datasets and charts, the `owners` information would be lost unless if included in the update payload

This scenario introduces issues as for example an admin updating the dataset description via the API would remove all owners.

This PR centralizes this logic to ensure consistency across charts, dashboards, datasets and reports (also DRYer).

No UI changes.

### TESTING INSTRUCTIONS
Tests added. To test through the UI:
1. Create a new chart. 
2. Update its description via the API.
3. Validate the `owners` configuration was not affected.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/base.py
+++ b/superset/commands/base.py
@@ -19,7 +19,7 @@ from typing import Any, Optional
 
 from flask_appbuilder.security.sqla.models import User
 
-from superset.commands.utils import populate_owners
+from superset.commands.utils import populate_owners, update_owner_list
 
 
 class BaseCommand(ABC):
@@ -70,3 +70,17 @@ class UpdateMixin:  # pylint: disable=too-few-public-methods
         :returns: Final list of owners
         """
         return populate_owners(owner_ids, default_to_user=False)
+
+    @staticmethod
+    def update_owner_list(
+        current_owners: list[User],
+        new_owners: Optional[list[int]],
+    ) -> list[User]:
+        """
+        Handle list of owners for update events.
+
+        :param current_owners: list of current owners
+        :param new_owners: list of new owners specified in the update payload
+        :returns: Final list of owners
+        """
+        return update_owner_list(current_owners, new_owners)
--- a/superset/commands/base.py
+++ b/superset/commands/base.py
@@ -19,7 +19,7 @@ from typing import Any, Optional
 
 from flask_appbuilder.security.sqla.models import User
 
-from superset.commands.utils import populate_owner_list, update_owner_list
+from superset.commands.utils import compute_owner_list, populate_owner_list
 
 
 class BaseCommand(ABC):
@@ -72,7 +72,7 @@ class UpdateMixin:  # pylint: disable=too-few-public-methods
         return populate_owner_list(owner_ids, default_to_user=False)
 
     @staticmethod
-    def update_owners(
+    def compute_owners(
         current_owners: Optional[list[User]],
         new_owners: Optional[list[int]],
     ) -> list[User]:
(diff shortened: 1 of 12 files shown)


superset/commands/base.py
```diff
@@ -70,3 +70,17 @@ def populate_owners(owner_ids: Optional[list[int]] = None) -> list[User]:
         :returns: Final list of owners
         """
         return populate_owners(owner_ids, default_to_user=False)
+
+    @staticmethod
+    def update_owner_list(
```
hughhhh, line 75: why are we wrapping this as `staticmethod`?

superset/commands/utils.py
```diff
@@ -62,6 +62,24 @@ def populate_owners(
...
+    """
+    Helper function for update commands, to properly handle the owners list.
+    Preserve the previous configuration unless included in the update payload.
+
+    :param current_owners: list of current owners
+    :param new_owners: list of new owners specified in the update payload
+    :returns: Final list of owners
+    """
+    owners_ids = (
+        [owner.id for owner in current_owners] if new_owners is None else new_owners
```
hughhhh, line 78: can you add this small unit test that will basically verify the that `populate_owners` will be called with the proper values?

Vitor-Avila: Sure! I'm going to create a `tests/unit_tests/commands/test_utils.py` file. I found some other unit test files using `pytest` so I'll go with that.

superset/commands/base.py
```diff
@@ -70,3 +70,17 @@ def populate_owners(owner_ids: Optional[list[int]] = None) -> list[User]:
...
+        new_owners: Optional[list[int]],
+    ) -> list[User]:
+        """
+        Handle list of owners for update events.
+
+        :param current_owners: list of current owners
+        :param new_owners: list of new owners specified in the update payload
+        :returns: Final list of owners
+        """
+        return update_owner_list(current_owners, new_owners)
```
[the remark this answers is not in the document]
Vitor-Avila: I had that same reaction when started looking into this haha I kept it this flow because this was how `populate_owners` was imported. I'll implement your first suggestion in a future commit

superset/commands/utils.py
Vitor-Avila: @hughhhh just added the unit tests 🙌 let me know if you have any feedback

superset/commands/base.py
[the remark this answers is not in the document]
Vitor-Avila: ended up renaming the functions to avoid importing `utils` entirely.

superset/commands/utils.py
```diff
@@ -62,6 +62,25 @@ def populate_owners(
     return owners
 
 
+def update_owner_list(
```
dpgaspar, line 65: nit: this is a pure function, but the name implies it's updating something but it's not, since it returns a newly created `list[User]`. why not call it `generate_owners` or `compute_owners`

dpgaspar (approved):
looks good, left a non blocking comment

superset/commands/utils.py
Vitor-Avila: that makes sense! Updated the functions to `compute_owners` (in the class), and `compute_owner_list` (in `utils`).

Vitor-Avila:
@michael-s-molina @hughhhh let me know if you have any other feedback, or if you think this one is good to merge. Thanks!

michael-s-molina:
> hey @michael-s-molina would you mind cherry-picking this PR into the 4.0 (or next possible) release?

All PRs with the `fix` prefix are cherry-picked every day. You don't need to request their addition 😉

Vitor-Avila:
that's amazing! thanks @michael-s-molina 🙏

PR #27370, opened 2024-03-03
feat: support to fetch multiple date time in time_range endpoint

zhaoyongjie:
### SUMMARY
Refactoring the `/v1/time_range/` endpoint to support multiple time range queries.

N/A

### TESTING INSTRUCTIONS
The unit test was covered.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/api.py
+++ b/superset/views/api.py
@@ -40,7 +40,15 @@ from superset.views.base import api, BaseSupersetView, handle_api_exception
 if TYPE_CHECKING:
     from superset.common.query_context_factory import QueryContextFactory
 
-get_time_range_schema = {"type": "string"}
+get_time_range_schema = {
+    "type": ["string", "array"],
+    "items": {
+        "type": "object",
+        "properties": {
+            "timeRange": {"type": "string"},
+        },
+    },
+}
 
 
 class Api(BaseSupersetView):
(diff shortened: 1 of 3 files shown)


superset/views/api.py
```diff
@@ -40,7 +40,15 @@
 if TYPE_CHECKING:
     from superset.common.query_context_factory import QueryContextFactory
 
-get_time_range_schema = {"type": "string"}
+get_time_range_schema = {
+    "type": ["string", "array"],
+    "items": {
+        "type": "object",
+        "properties": {
+            "timeRange": {"type": "string"},
```
zhaoyongjie, line 48: Let's use flexible data structure in the object so that support more config in time range request, e.g. timezone. Our forked version already supported timezone in time range, will open separated PR.

kgabryje (approved):

superset/views/api.py
```diff
@@ -40,7 +40,15 @@
 if TYPE_CHECKING:
     from superset.common.query_context_factory import QueryContextFactory
 
-get_time_range_schema = {"type": "string"}
+get_time_range_schema = {
+    "type": ["string", "array"],
```
michael-s-molina, line 44: @kgabryje @zhaoyongjie Isn't this a breaking change given that we're changing the type from string to array? If yes, we would need to make it backward compatible to avoid reverting this PR given that we're not in a breaking window anymore.

zhaoyongjie: it's not a breaking changing, the `['string', 'array']` means that time_range is able to adapt String and Array type.

michael-s-molina: Oh my bad. I missed the `string` type inside the array. Time to get some coffee ☕ !

zhaoyongjie: Perfect! Look forward to a cafe time!

PR #27368, opened 2024-03-03
feat: datediff in datetime_parser

zhaoyongjie:
### SUMMARY
The PR introduced a new function in date time expression that is `datediff(<from>, <to>, [time unit],)`, it returns a integer, so the return value should apply in the `dateadd` function.

The basic form is:
```
# return value is 9
datediff(datetime('2018-01-01T00:00:00'), datetime('2018-01-10T00:00:00'))

# return value is -9
datediff(datetime('2018-01-10T00:00:00'), datetime('2018-01-01T00:00:00'))
```

The `from` and `to` parameter is able to be set any date time expression:
```
# return value is -365
datediff(dateadd(datetime('2018-01-01T00:00:00'), 1, year), datetime('2018-01-10T00:00:00'))
```

The `datediff` also be in `dateadd` as a time delta
```
# return value is "datetime(2017, 12, 23, 0, 0, 0)"
dateadd(
    datetime('2018-01-01T00:00:00'),
    datediff(datetime('2018-01-10T00:00:00'), datetime('2018-01-01T00:00:00')),
    day,
)
```

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N/A

### TESTING INSTRUCTIONS

The unit test was covered 100% cases.

### ADDITIONAL INFORMATION

- [x] Has associated issue: the PR: https://github.com/apache/superset/pull/27145 and https://github.com/apache/superset/pull/27355 should make refactoring that build time_range in frontend.
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27386, opened 2024-03-04
refactor: build shifted time filter in frontend

zhaoyongjie:
### SUMMARY
This PR intend to build query object for the time comparison in frontend rather than backend. The `datediff` implemented in [PR](https://github.com/apache/superset/pull/27368), so `InheritedRange` is also supported in frontend.

In order to simplify making query object, created `getTimeComparisonFiltersByQueryObject` and `getShiftedTimeRange` utils and removed previous frontend and backend relative codes.

N/A

### TESTING INSTRUCTIONS

the new unit tests should cover all cases.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/package.json
```diff
@@ -109,7 +109,6 @@
     "@superset-ui/legacy-preset-chart-nvd3": "file:./plugins/legacy-preset-chart-nvd3",
     "@superset-ui/plugin-chart-echarts": "file:./plugins/plugin-chart-echarts",
     "@superset-ui/plugin-chart-handlebars": "file:./plugins/plugin-chart-handlebars",
-    "@superset-ui/plugin-chart-period-over-period-kpi": "file:./plugins/plugin-chart-period-over-period-kpi",
```
zhaoyongjie, line 112: This package was moved to superset-ui

superset-frontend/packages/superset-ui-core/src/index.ts
```diff
@@ -37,4 +37,3 @@ export * from './math-expression';
 export * from './ui-overrides';
 export * from './hooks';
 export * from './currency-format';
-export * from './time-comparison';
```
zhaoyongjie, line 40: Replaced with new utils.

Antonio-RiveroMartnez:
Hey @zhaoyongjie first, thanks for taking care of these changes. Now, as I see it we're duplicating existing logic in the frontend now that the backend already has in place, also, in this viz type it's ok to prepare two queries (original vs comparison) but vizs like table chart where you must use JOIN in the comparison queries, having them as separated queries is not really something we should do, and in order to process just one query and extract from it the "comparison" data, you still need to identify such queries apart from current ones.

Why would you need to identify them? think on pagination queries, having two isolated queries won't guarantee you to have the right result sets to compare against each other.

So, TLDR; IMO we still need the backend to process at some extend these ranges and  tell apart when you're running a regular query from a query that will generate subqueries and JOIN on them.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/buildQuery.ts
```diff
@@ -18,50 +18,38 @@
...
+    baseQueryObject,
+  ]);
 
-  const queryContextA = buildQueryContext(formData, baseQueryObject => [
+  let shiftedFormData = formData;
+  if (
+    formData?.time_comparison === ComparisonTimeRangeType.Custom &&
+    formData?.adhoc_custom
+  ) {
+    shiftedFormData = {
```
kgabryje, line 36: non-blocking: we've got some more viz plugins with time range comparison coming, so it might be worth it to move this logic to a reusable util. We can do that as a follow up @Antonio-RiveroMartnez

Antonio-RiveroMartnez:
> @Antonio-RiveroMartnez If you need a backend join, you need to use time_offset instead of adding a new concept in current query object.
> 
> This is what I'm worried about, now you've added a lot of redundant request load on the query object.

time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.

zhaoyongjie:
> time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at query pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.

Okay, Let's narrow the scope of the discussion, I'm not sure what join you are talking about. The goal of this PR to do one thing that build comparison time range in frontend rather than backend.

Antonio-RiveroMartnez:
> > time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at query pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.
> 
> Okay, Let's narrow the scope of the discussion, I'm not sure what join you are talking about. The goal of this PR to do one thing that build comparison time range in frontend rather than backend. So, let' discussion around this PR.

Yes, totally fine with that, just giving you context of why/when the query object needs to identify something "new" for time comparisons because again, you not always are able to compare using multiple queries sent from the frontend. Meaning the backend would be generating those ranges too, so that'w why I'm not worried but just pointing out the fact that some of the code deleted here would be actually required to handle those cases and that instead of having this in two places "front" and "back" I would rather having it in just one place.

If it's easier to resonate, we can move ahead with these changes the moment CI passes and when I'm adjusting the other PR I will point out and reintroduce the necessary code and we can discuss there.

Thanks as always.

zhaoyongjie:
> > > time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at query pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.
> > 
> > 
> > Okay, Let's narrow the scope of the discussion, I'm not sure what join you are talking about. The goal of this PR to do one thing that build comparison time range in frontend rather than backend. So, let' discussion around this PR.
> 
> Yes, totally fine with that, just giving you context of why/when the query object needs to identify something "new" for time comparisons because again, you not always are able to compare using multiple queries sent from the frontend. Meaning the backend would be generating those ranges too, so that'w why I'm not worried but just pointing out the fact that some of the code deleted here would be actually required to handle those cases and that instead of having this in two places "front" and "back" I would rather having it in just one place.
> 
> If it's easier to resonate, we can move ahead with these changes the moment CI passes and when I'm adjusting the other PR I will point out and reintroduce the necessary code and we can discuss there.
> 
> Thanks as always.

Thanks for your reply, I hope you are able to provide some examples for why/when/what is the exact `JOIN` in the your [PR](https://github.com/apache/superset/pull/27355), but may be current PR is not a good place to talk about `join`.

Theoretically, the frontend and backend build time range are exactly the same because all the time range will parse by the pyparser. 

I think we might need to think some new way to address like this https://github.com/apache/superset/pull/27355/files#diff-af5e15a55ff9c81f441eb6b9b10dae13b2ee17fec614d3e3ad0057ed0dc814aeR1434, I don't think such codes is maintable, but we can discussion in your PR.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/buildQuery.ts
zhaoyongjie: Sure, good point.

Antonio-RiveroMartnez:
> > > > time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at query pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.
> > > 
> > > 
> > > Okay, Let's narrow the scope of the discussion, I'm not sure what join you are talking about. The goal of this PR to do one thing that build comparison time range in frontend rather than backend. So, let' discussion around this PR.
> > 
> > 
> > Yes, totally fine with that, just giving you context of why/when the query object needs to identify something "new" for time comparisons because again, you not always are able to compare using multiple queries sent from the frontend. Meaning the backend would be generating those ranges too, so that'w why I'm not worried but just pointing out the fact that some of the code deleted here would be actually required to handle those cases and that instead of having this in two places "front" and "back" I would rather having it in just one place.
> > If it's easier to resonate, we can move ahead with these changes the moment CI passes and when I'm adjusting the other PR I will point out and reintroduce the necessary code and we can discuss there.
> > Thanks as always.
> 
> Thanks for your reply, I hope you could provide some examples for why/when/what is the exact `JOIN` in the your [PR](https://github.com/apache/superset/pull/27355), but may be current PR is not a good place to talk about `join`.
> 
> Theoretically, the frontend and backend build time range are exactly the same because all the time range will parse by the pyparser.
> 
> I think we might need to think some new way to address like this https://github.com/apache/superset/pull/27355/files#diff-af5e15a55ff9c81f441eb6b9b10dae13b2ee17fec614d3e3ad0057ed0dc814aeR1434, I don't think such codes is maintable, but we can discussion in your PR.

I tagged you in that other PR already if you want to discuss about the `JOIN`. But again, IMHO the backend still would need to build these ranges because as I said, it's not always a good idea/or even valid to have separate `queryObjects` to build the comparison. Meaning, the "frontend" doing it like in here only address one type of comparison scenario. While having it on the backend might serve all scenarios, or at least ease reusability for when needed.

TLDR; The code you're removing here will be needed after and reintroduced again very shortly so I see no point in doing it the first time.

I'm going to try to explain why building in the frontend is only "good" for one time of comparisons now:

Imagine your queries are like this is like this:

```
Query A: a metric from 1900-01-01 to 1950-01-01 
Results: 1000

QueryB: a metric from 1950-01-01 to 2000-01-01
Results: 100
```

Here having the frontend sending two queries is OK and totally fine like you have in this PR. Now, what's the limitation of this approach? Let's imagine another type of queries for another type of vizs. For example, Table Chart. Same exercise as before:

```
Query A: a list of users that have bought the more items along with the num of items from 1900-01-01 to 1950-01-01 ordered by # items bought desc
Results: User1: 100, User2: 90, User3:80

Query B: a list of users that have bought the more items along with the num of items from 1950-01-01 to 2000-01-01 ordered by # items bought desc
Results: User4:95, User1:70, User2:65
```

In that scenario it's also fine to build two queries in the frontend with individual time_ranges and process each queryObject individually to run the metrics etc. Again, just fine with your PR here.

Limitation? You can paginate a result set in Table chart (or any other chart that does the same or that would do the same later on). If you paginate using again just the frontend with isolated queries like you have here in this PR you would have something like:

```
Query A: a list of users that have bought the more items along with the num of items from 1900-01-01 to 1950-01-01 ordered by # items bought desc LIMIT 2
Results: User1: 100, User2: 90, User3:80

Query B: a list of users that have bought the more items along with the num of items from 1950-01-01 to 2000-01-01 ordered by # items bought desc LIMIT2
Results: User4:95, User1:70
```

When creating the comparison metrics you would assume "User2 in query A has 90 items and it's not present in the result set of query B so it's 0" which is as you can imagine, it's invalid. So, how to solve that? We could produce the queryObject needed in the backend itself (with the `JOIN` we can discuss in the other PR) and let the backend build the time ranges needed when constructing the queries.

But yet again why not just two queries from the frontend with different ranges?
1. It's redundant (as per the definition of redundant) to send you two queryObjects where we can produce both with just one: My mistake because I used Mixed time series as base when building the BigNumber with time comparison, so I introduced that pattern there.
2. Data might have changed between execution time of each query, which may affect your results with inconsistent data.

Hopefully this little exercise helped to clarify my point of view on why we should NOT bring that time_rage building to the frontend only (if not at all), and also can help to realize why time_offsets are not usable here neither.

Just want to be clear of the limitation and the fact that solving the comparison for good like in the other PR would require the backend to process these ranges at what's more adding an optional in the extras or similar in the queryObject.

But if others are happy with this we can move forward acknowledging what I stated here.

Thanks as always @zhaoyongjie for the help and helping to bring more robust solutions whenever possible.

zhaoyongjie:
> > > > > time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at query pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.
> > > > 
> > > > 
> > > > Okay, Let's narrow the scope of the discussion, I'm not sure what join you are talking about. The goal of this PR to do one thing that build comparison time range in frontend rather than backend. So, let' discussion around this PR.
> > > 
> > > 
> > > Yes, totally fine with that, just giving you context of why/when the query object needs to identify something "new" for time comparisons because again, you not always are able to compare using multiple queries sent from the frontend. Meaning the backend would be generating those ranges too, so that'w why I'm not worried but just pointing out the fact that some of the code deleted here would be actually required to handle those cases and that instead of having this in two places "front" and "back" I would rather having it in just one place.
> > > If it's easier to resonate, we can move ahead with these changes the moment CI passes and when I'm adjusting the other PR I will point out and reintroduce the necessary code and we can discuss there.
> > > Thanks as always.
> > 
> > 
> > Thanks for your reply, I hope you could provide some examples for why/when/what is the exact `JOIN` in the your [PR](https://github.com/apache/superset/pull/27355), but may be current PR is not a good place to talk about `join`.
> > Theoretically, the frontend and backend build time range are exactly the same because all the time range will parse by the pyparser.
> > I think we might need to think some new way to address like this https://github.com/apache/superset/pull/27355/files#diff-af5e15a55ff9c81f441eb6b9b10dae13b2ee17fec614d3e3ad0057ed0dc814aeR1434, I don't think such codes is maintable, but we can discussion in your PR.
> 
> I tagged you in that other PR already if you want to discuss about the `JOIN`. But again, IMHO the backend still would need to build these ranges because as I said, it's not always a good idea/or even valid to have separate `queryObjects` to build the comparison. Meaning, the "frontend" doing it like in here only address one type of comparison scenario. While having it on the backend might serve all scenarios, or at least ease reusability for when needed.
> 
> TLDR; The code you're removing here will be needed after and reintroduced again very shortly so I see no point in doing it the first time.
> 
> I'm going to try to explain why building in the frontend is only "good" for one time of comparisons now:
> 
> Imagine your queries are like this is like this:
> 
> ```
> Query A: a metric from 1900-01-01 to 1950-01-01 
> Results: 1000
> 
> QueryB: a metric from 1950-01-01 to 2000-01-01
> Results: 100
> ```
> 
> Here having the frontend sending two queries is OK and totally fine like you have in this PR. Now, what's the limitation of this approach? Let's imagine another type of queries for another type of vizs. For example, Table Chart. Same exercise as before:
> 
> ```
> Query A: a list of users that have bought the more items along with the num of items from 1900-01-01 to 1950-01-01 ordered by # items bought desc
> Results: User1: 100, User2: 90, User3:80
> 
> Query B: a list of users that have bought the more items along with the num of items from 1950-01-01 to 2000-01-01 ordered by # items bought desc
> Results: User4:95, User1:70, User2:65
> ```
> 
> In that scenario it's also fine to build two queries in the frontend with individual time_ranges and process each queryObject individually to run the metrics etc. Again, just fine with your PR here.
> 
> Limitation? You can paginate a result set in Table chart (or any other chart that does the same or that would do the same later on). If you paginate using again just the frontend with isolated queries like you have here in this PR you would have something like:
> 
> ```
> Query A: a list of users that have bought the more items along with the num of items from 1900-01-01 to 1950-01-01 ordered by # items bought desc LIMIT 2
> Results: User1: 100, User2: 90, User3:80
> 
> Query B: a list of users that have bought the more items along with the num of items from 1950-01-01 to 2000-01-01 ordered by # items bought desc LIMIT2
> Results: User4:95, User1:70
> ```
> 
> When creating the comparison metrics you would assume "User2 in query A has 90 items and it's not present in the result set of query B so it's 0" which is as you can imagine, it's invalid. So, how to solve that? We could produce the queryObject needed in the backend itself (with the `JOIN` we can discuss in the other PR) and let the backend build the time ranges needed when constructing the queries.
> 
> But yet again why not just two queries from the frontend with different ranges?
> 
> 1. It's redundant (as per the definition of redundant) to send you two queryObjects where we can produce both with just one: My mistake because I used Mixed time series as base when building the BigNumber with time comparison, so I introduced that pattern there.
> 2. Data might have changed between execution time of each query, which may affect your results with inconsistent data.
> 
> Hopefully this little exercise helped to clarify my point of view on why we should NOT bring that time_rage building to the frontend only (if not at all), and also can help to realize why time_offsets are not usable here neither.
> 
> Just want to be clear of the limitation and the fact that solving the comparison for good like in the other PR would require the backend to process these ranges at what's more adding an optional in the extras or similar in the queryObject.
> 
> But if others are happy with this we can move forward acknowledging what I stated here.
> 
> Thanks as always @zhaoyongjie for the help and helping to bring more robust solutions whenever possible.

Do you mind copy/paste this discussion in the [other PR](https://github.com/apache/superset/pull/27355) instead of one? IMO, this discussion goes far beyond the problem this PR wants to solve.

Antonio-RiveroMartnez:
> > > > > > time_offsets joins DataFrames which are not where you would need to apply the join, would you? You need the join to happen at query pre execution level, not when you get the results of individual queries. `time_offset` works when you have the total data at your disposal like for example when you're working with other vizs but not for example when working with paginated data like in tables.
> > > > > 
> > > > > 
> > > > > Okay, Let's narrow the scope of the discussion, I'm not sure what join you are talking about. The goal of this PR to do one thing that build comparison time range in frontend rather than backend. So, let' discussion around this PR.
> > > > 
> > > > 
> > > > Yes, totally fine with that, just giving you context of why/when the query object needs to identify something "new" for time comparisons because again, you not always are able to compare using multiple queries sent from the frontend. Meaning the backend would be generating those ranges too, so that'w why I'm not worried but just pointing out the fact that some of the code deleted here would be actually required to handle those cases and that instead of having this in two places "front" and "back" I would rather having it in just one place.
> > > > If it's easier to resonate, we can move ahead with these changes the moment CI passes and when I'm adjusting the other PR I will point out and reintroduce the necessary code and we can discuss there.
> > > > Thanks as always.
> > > 
> > > 
> > > Thanks for your reply, I hope you could provide some examples for why/when/what is the exact `JOIN` in the your [PR](https://github.com/apache/superset/pull/27355), but may be current PR is not a good place to talk about `join`.
> > > Theoretically, the frontend and backend build time range are exactly the same because all the time range will parse by the pyparser.
> > > I think we might need to think some new way to address like this https://github.com/apache/superset/pull/27355/files#diff-af5e15a55ff9c81f441eb6b9b10dae13b2ee17fec614d3e3ad0057ed0dc814aeR1434, I don't think such codes is maintable, but we can discussion in your PR.
> > 
> > 
> > I tagged you in that other PR already if you want to discuss about the `JOIN`. But again, IMHO the backend still would need to build these ranges because as I said, it's not always a good idea/or even valid to have separate `queryObjects` to build the comparison. Meaning, the "frontend" doing it like in here only address one type of comparison scenario. While having it on the backend might serve all scenarios, or at least ease reusability for when needed.
> > TLDR; The code you're removing here will be needed after and reintroduced again very shortly so I see no point in doing it the first time.
> > I'm going to try to explain why building in the frontend is only "good" for one time of comparisons now:
> > Imagine your queries are like this is like this:
> > ```
> > Query A: a metric from 1900-01-01 to 1950-01-01 
> > Results: 1000
> > 
> > QueryB: a metric from 1950-01-01 to 2000-01-01
> > Results: 100
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here having the frontend sending two queries is OK and totally fine like you have in this PR. Now, what's the limitation of this approach? Let's imagine another type of queries for another type of vizs. For example, Table Chart. Same exercise as before:
> > ```
> > Query A: a list of users that have bought the more items along with the num of items from 1900-01-01 to 1950-01-01 ordered by # items bought desc
> > Results: User1: 100, User2: 90, User3:80
> > 
> > Query B: a list of users that have bought the more items along with the num of items from 1950-01-01 to 2000-01-01 ordered by # items bought desc
> > Results: User4:95, User1:70, User2:65
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In that scenario it's also fine to build two queries in the frontend with individual time_ranges and process each queryObject individually to run the metrics etc. Again, just fine with your PR here.
> > Limitation? You can paginate a result set in Table chart (or any other chart that does the same or that would do the same later on). If you paginate using again just the frontend with isolated queries like you have here in this PR you would have something like:
> > ```
> > Query A: a list of users that have bought the more items along with the num of items from 1900-01-01 to 1950-01-01 ordered by # items bought desc LIMIT 2
> > Results: User1: 100, User2: 90, User3:80
> > 
> > Query B: a list of users that have bought the more items along with the num of items from 1950-01-01 to 2000-01-01 ordered by # items bought desc LIMIT2
> > Results: User4:95, User1:70
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When creating the comparison metrics you would assume "User2 in query A has 90 items and it's not present in the result set of query B so it's 0" which is as you can imagine, it's invalid. So, how to solve that? We could produce the queryObject needed in the backend itself (with the `JOIN` we can discuss in the other PR) and let the backend build the time ranges needed when constructing the queries.
> > But yet again why not just two queries from the frontend with different ranges?
> > 
> > 1. It's redundant (as per the definition of redundant) to send you two queryObjects where we can produce both with just one: My mistake because I used Mixed time series as base when building the BigNumber with time comparison, so I introduced that pattern there.
> > 2. Data might have changed between execution time of each query, which may affect your results with inconsistent data.
> > 
> > Hopefully this little exercise helped to clarify my point of view on why we should NOT bring that time_rage building to the frontend only (if not at all), and also can help to realize why time_offsets are not usable here neither.
> > Just want to be clear of the limitation and the fact that solving the comparison for good like in the other PR would require the backend to process these ranges at what's more adding an optional in the extras or similar in the queryObject.
> > But if others are happy with this we can move forward acknowledging what I stated here.
> > Thanks as always @zhaoyongjie for the help and helping to bring more robust solutions whenever possible.
> 
> Do you mind copy/paste this discussion in the [other PR](https://github.com/apache/superset/pull/27355) instead of one? IMO, this discussion goes far beyond the problem this PR wants to solve.



In my opinion it belong here because here you're removing it form the backend and will make us put it there again, and moving the time range only to the frontend is also something I don't agree and this PR is the one doing it. But again, open to bring more people in and I'm always up to whatever is best for the project 🎉 Thanks again.

zhaoyongjie:
> When creating the comparison metrics you would assume "User2 in query A has 90 items and it's not present in the result set of query B so it's 0" which is as you can imagine, it's invalid. So, how to solve that? We could produce the queryObject needed in the backend itself (with the JOIN we can discuss in the other PR) and let the backend build the time ranges needed when constructing the queries.

@Antonio-RiveroMartnez The use case here is entirely different from your `time comparison`. What you're aiming for is to `join` or `merge` two **query result sets** based on certain logical relationships along the **datetime index**, and this is precisely what the `time_offset`(aka rolling) is worked for.

However, the issue this PR aims to address is 'time comparison', where the essence lies in comparing two independent queries with differences in filters, including **time filters** or **custom adhoc filters**.

Thanks for your explanation.

zhaoyongjie:
@Antonio-RiveroMartnez @kgabryje the comprehensive frontend unit tests was added, please review this PR again.

superset-frontend/packages/superset-ui-core/src/query/extractTimeComparisonFilters.ts
```diff
@@ -0,0 +1,132 @@
...
+  let unit = 'DAY';
+  let delta = '-1';
+  if (timeShiftType === ComparisonTimeRangeType.Year) {
+    unit = 'YEAR';
+  }
+  if (timeShiftType === ComparisonTimeRangeType.Month) {
+    unit = 'MONTH';
+  }
+  if (timeShiftType === ComparisonTimeRangeType.Week) {
+    unit = 'WEEK';
```
eschutho, line 112: Hi @zhaoyongjie. A few of us took a look at this today and we're thinking that we can simplify this logic by using the existing "1 year ago", "1 month ago" and "1 week ago" human readable strings instead of using this enum. And possibly a way to calculate inherit and custom as well. We're going to do some work on Bar and Line which will help us get more familiar with the existing patterns and then we'll come back and update this. Does that work for you? cc @Antonio-RiveroMartnez

eschutho: I thought I would give you an early heads up in case it impacted whether or not you needed as many changes here and consequently if you needed to write a SIP.

zhaoyongjie: I think you need to read this function though and check how does time filter work before comments.

zhaoyongjie: The goal of this function(`getShiftedTimeRange`) is that generate a shifted time filter base on the previous timeFilter, and implemented in Frontend rather than backend.

eschutho: Hi, yeah, just saying that if you're trying to find a way around the extra code, we'll try to clean this up. If you still want to move the rest of the existing logic to the frontend, then yeah it seems like a SIP would still be worthwhile.

eschutho (changes requested):
It's been a while since we've been talking about this PR, but I think a lot has changed since and perhaps we don't need this.

zhaoyongjie:
Since we have another PR for the same feature, close this one, feel free reopen if we need it in future.

PR #27388, opened 2024-03-04
fix(utils): fix off-by-one error in how rolling window's min_periods truncates dataframe

sfirke:
### SUMMARY
In Advanced Analytics, if the Min Periods value is 5, the first row in the resulting `df_rolling` with a valid result will be row 4 (since counting starts at 0).  Right now the line `df_rolling = df_rolling[min_periods:]` truncates rows 0-4 -- that is, five rows -- throwing out the first rolling value that satisfies the min_periods requirement.

Currently the tooltip for Min Periods is misleading, this will make it true.

### SCREENSHOTS
Here in Superset 3.1.1 you can see that the 5th bar should be eligible for a rolling average since Min Periods is 5.  But it's only beginning on the 6th bar that the line appears.
image

When I tell Superset to use 5-Period averages for the rolling window and min of 4 periods, this truncates the first 4 rows (due to the buggy off-by-one I'm fixing) and serves what I wanted.  27.4 is indeed the average of the first 5 periods and a valid result.
image

### TESTING INSTRUCTIONS
Build a chart with rolling average in an ephemeral environment.

### DIFF
--- a/superset/utils/pandas_postprocessing/rolling.py
+++ b/superset/utils/pandas_postprocessing/rolling.py
@@ -97,5 +97,5 @@ def rolling(  # pylint: disable=too-many-arguments
     df_rolling = _append_columns(df, df_rolling, columns)
 
     if min_periods:
-        df_rolling = df_rolling[min_periods-1:]
+        df_rolling = df_rolling[min_periods - 1 :]
     return df_rolling
--- a/superset/utils/pandas_postprocessing/rolling.py
+++ b/superset/utils/pandas_postprocessing/rolling.py
@@ -97,5 +97,5 @@ def rolling(  # pylint: disable=too-many-arguments
     df_rolling = _append_columns(df, df_rolling, columns)
 
     if min_periods:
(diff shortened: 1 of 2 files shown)


sfirke:
I see I will have to update some tests.  This is my first time working with testing in Python so may take me a bit.  I still think this is the correct PR.

sfirke:
@michael-s-molina maybe a candidate for 4.0.0rc2 if there's time?  It is definitely a fix IMO, but it will change how certain charts display.  If people were adapted to the buggy behavior and were specifying N-1 for min periods as a workaround to actually get N to display, it could cause an unintended data point to display.  So bundling with a major release would be convenient.

Not critical though and IMO could be merged into master after 4.0.0 without waiting for a breaking window.

michael-s-molina (approved):

rusackas:
Merging, as I think this is a fix. If people start complaining, we can revert it. If anyone disagrees, I can revert it now. 

CC @sadpandajoe @Vitor-Avila FYI

PR #27390, opened 2024-03-04
fix: Re-enable CI checks on release branches

michael-s-molina:
### SUMMARY
https://github.com/apache/superset/pull/26766 restricted some CI workflows to run specifically on the `master` branch. This resulted in a problem for the release branches where running CI is really important when cherry-picking fixes into the branch. This PR adds a regex to allow CI to run on push events for release branches, restoring the previous behavior.

Check [filter-pattern-cheat-sheet](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) for details about the regex. Dots don't need to be escaped.

### TESTING INSTRUCTIONS
Restricted CI workflows should run only on `master` and branches that match the regex.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/check_db_migration_confict.yml
+++ b/.github/workflows/check_db_migration_confict.yml
@@ -4,7 +4,8 @@ on:
     paths:
       - "superset/migrations/**"
     branches:
-      - 'master'
+      - "master"
+      - "[0-9].[0-9]"
   pull_request:
     paths:
       - "superset/migrations/**"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,16 +2,16 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "master" ]
+    branches: ["master", "[0-9].[0-9]"]
     paths:
-      - 'superset/**'
+      - "superset/**"
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ "master" ]
+    branches: ["master"]
     paths:
-      - 'superset/**'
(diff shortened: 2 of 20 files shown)


rusackas:
Maybe the precommit hook isn't checking these files? The change in quotes seems unusual if so.

mistercrunch:
Oh I didn't think of release branches when I made the changes, but seems totally reasonable that we'd run CI for release branches. 

I had to review the `concurrency` setting, and currently it will run CI for every "push", which I believe is desired behavior (@michael-s-molina please confirm). Note that the way it's set up + GHA mechanics, it will run CI for each push (not for each commit) and **won't** kill previous jobs the same way that we do on `pull_request` event, where a new push will kill previously running jobs. If you wanted to get CI to run for each cherry you pick you can do so by pushing after each cherry to get a CI run for each one.

Total side note here, I'd love to keep things more DRY across actions (not repeating the branch name pattern for instance), but I don't think we're at the critical mass of complexity to justify something more dynamic just yet. It could be nice to have some sort of DSL that generates the actions more dynamically, but seems over-engineered for where we're at currently.

mistercrunch:
@rusackas I think the check-yaml pre-commit hook isn't sophisticated/opinionated like that, it just makes sure the yaml is valid, it doesn't care about how it's styled

dpgaspar (approved):

mistercrunch (approved):

PR #27392, opened 2024-03-04
fix: Missing sql_editor_id index

justinpark:
### SUMMARY
As titled, it adds the index for `sql_editor_id` in query table which seems to have caused a slowdown in the [sqllab_bootstrap](https://github.com/apache/superset/blob/d2f7dec208cfa31583310f96a9f387853af0fbc8/superset/sqllab/utils.py#L109-L113) query.

### TESTING INSTRUCTIONS
locally test `superset db upgrade`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
+++ b/superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
@@ -23,17 +23,21 @@ Create Date: 2024-03-04 14:48:16.998927
 """
 
 # revision identifiers, used by Alembic.
-revision = 'efb7a20ff8d8'
-down_revision = 'be1b217cd8cd'
+revision = "efb7a20ff8d8"
+down_revision = "be1b217cd8cd"
 
-from alembic import op
 import sqlalchemy as sa
+from alembic import op
+
 
 def upgrade():
     op.create_index(
-        op.f("ix_query_sql_editor_id"), "query", ["sql_editor_id"], unique=False
+        op.f("ix_query_user_id_sql_editor_id"),
+        "query",
+        ["user_id", "sql_editor_id"],
+        unique=False,
     )
 
 
 def downgrade():
-    op.drop_index(op.f("ix_query_sql_editor_id"), table_name="query")
+    op.drop_index(op.f("ix_query_user_id_sql_editor_id"), table_name="query")
--- a/superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
+++ b/superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
@@ -26,18 +26,24 @@ Create Date: 2024-03-04 14:48:16.998927
 revision = "efb7a20ff8d8"
 down_revision = "be1b217cd8cd"
 
-import sqlalchemy as sa
 from alembic import op
+
+from superset.migrations.shared.utils import table_has_index
+
+table = "query"
+index = "ix_query_user_id_sql_editor_id"
 
 
 def upgrade():
-    op.create_index(
-        op.f("ix_query_user_id_sql_editor_id"),
-        "query",
-        ["user_id", "sql_editor_id"],
-        unique=False,
-    )
+    if not table_has_index(table, index):
+        op.create_index(
+            op.f(index),
+            table,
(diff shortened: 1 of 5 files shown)


john-bodley:
@justinpark you'll also have to:

1. Update the model to reflect the parity between how the SQLAlchemy model is defined and the underlying schema.
2. Add a line item to `UPDATING.md`*.

Creating a new index to the `query` could be very expensive (given the size of the table) and may require downtime.

superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
```diff
@@ -0,0 +1,39 @@
...
+# revision identifiers, used by Alembic.
+revision = 'efb7a20ff8d8'
+down_revision = 'be1b217cd8cd'
+
+from alembic import op
+import sqlalchemy as sa
+
+def upgrade():
+    op.create_index(
+        op.f("ix_query_sql_editor_id"), "query", ["sql_editor_id"], unique=False
```
john-bodley, line 34: Should this index be on the (`user_id`, `sql_editor_id`) tuple instead?

mistercrunch:
From experience that table can get massive and depending on your SQL engine, and creating an index can be a locking operation, which can cascade into deadlock territory on this particular table (I think there's still a pretty crazy amount of polling happening in SQL Lab to look at query state). If you're using MySQL I'd treat this migration very very sensitively. I think there are ways to create index in a non-locking way in MySQL, but have to go out of your way to do it `(ALGORITHM=INPLACE, LOCK=NONE;)`. Postgres should do better. 

Maybe not a bad time to archive the query table table in your environment ... For most shops I think 90 days worth of query history is enough to go by, not a bad thing to schedule something that deletes 90+ day old queries daily... Note that `changed_on` is also indexed to help and support and efficient DELETE operation there.

superset/models/sql_lab.py
```diff
@@ -86,7 +86,7 @@ class Query(
     user_id = Column(Integer, ForeignKey("ab_user.id"), nullable=True)
     status = Column(String(16), default=QueryStatus.PENDING)
     tab_name = Column(String(256))
-    sql_editor_id = Column(String(256))
+    sql_editor_id = Column(String(256), index=True)
```
john-bodley, line 89: Given that it's now a index on two columns you'll have to use the [`Index`](https://docs.sqlalchemy.org/en/20/core/constraints.html#indexes) function.

mistercrunch:
About my earlier comment, the issue may be significant enough that you'd want to write a custom migration for MySQL for your own sake at Airbnb, you can probably go direct DDL `CREATE INDEX` with the proper flags just for MySQL. I remember fighting hard on some simple migration on that specific table. Connections / locks kept coming, had to silence all k8s deployments, kill MySQL sessions (which kept coming back with polling) and ultimately wait for things to get through. `(ALGORITHM=INPLACE, LOCK=NONE;)` may allow you to not schedule or incur downtime.

superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
```diff
@@ -0,0 +1,43 @@
...
+# revision identifiers, used by Alembic.
+revision = "efb7a20ff8d8"
+down_revision = "be1b217cd8cd"
+
+import sqlalchemy as sa
+from alembic import op
+
+
+def upgrade():
+    op.create_index(
```
michael-s-molina, line 34: Could we skip the upgrade if the index already exists?

michael-s-molina: I pushed a commit to check for index existence.

justinpark:
@michael-s-molina @john-bodley PTAL

michael-s-molina (approved):

superset/migrations/versions/2024-03-04_14-48_efb7a20ff8d8_add_query_sql_editor_id_index.py
justinpark: @john-bodley We have agreed to use a single key index rather than these composite keys. Could you please confirm the agreement here?

john-bodley (approved):

PR #27403, opened 2024-03-06
fix: missing shared color in mixed timeseries

justinpark:
### SUMMARY
When a shared label color is used in the mixed series chart, there seems to be an issue where the label color, particularly in the 2nd series label, is not accurately rendered. This occurs because the mixed series passes the label color name with a ` (1)` suffix, which is not necessary according to the color algorithm.
This commit aims to address this issue by removing this redundant suffix from the color scale key. 

|Screenshot 2024-03-05 at 4 28 04 PM|Screenshot 2024-03-05 at 4 30 14 PM|
|--|--|
|Screenshot 2024-03-05 at 4 31 26 PM|Screenshot 2024-03-05 at 4 31 14 PM|

### TESTING INSTRUCTIONS
storybook

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -411,7 +411,8 @@ export default function transformProps(
   rawSeriesB.forEach(entry => {
     const entryName = String(entry.name || '');
     const seriesName = `${inverted[entryName] || entryName} (1)`;
-    const colorScaleKey = getOriginalSeries(seriesName, array);
+    const colorScaleSeriesName = inverted[entryName] || entryName;
+    const colorScaleKey = getOriginalSeries(colorScaleSeriesName, array);
 
     const seriesFormatter = getFormatter(
       customFormattersSecondary,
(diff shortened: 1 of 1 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -411,7 +411,8 @@ export default function transformProps(
   rawSeriesB.forEach(entry => {
     const entryName = String(entry.name || '');
     const seriesName = `${inverted[entryName] || entryName} (1)`;
```
michael-s-molina, line 413:
```suggestion
    const seriesEntry = inverted[entryName] || entryName;
    const seriesName = `${seriesEntry} (1)`;
```

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -411,7 +411,8 @@ export default function transformProps(
...
-    const colorScaleKey = getOriginalSeries(seriesName, array);
+    const colorScaleSeriesName = inverted[entryName] || entryName;
+    const colorScaleKey = getOriginalSeries(colorScaleSeriesName, array);
```
michael-s-molina, line 415:
```suggestion
    const colorScaleKey = getOriginalSeries(seriesEntry, array);
```

michael-s-molina (approved):

PR #27405, opened 2024-03-06
chore: upgrade setuptools/pip in Dockerfile

mistercrunch:
Making sure that pip and setuptools are the latest in our Dockerfile

closes https://github.com/apache/superset/pull/27405


john-bodley:
@mistercrunch I don't see the `requirements/development.in` file in your PR?

mistercrunch:
I thought I was stuck and couldn't run pip-compile-multi since it's busted, but I just found its `--no-upgrade` flag which did work just now.

BTW keep running in some issue where pip-compile-multi replaces the `.` in `-e file:.` to the full local path in my local computer randomly and I have to replace back to a `.` manually. Luckily it doesn't mess up the hash that pip-compile-multi manages.

mistercrunch:
Oh btw, note that even after removing setuptools from `requirements/development.in`, all of our `requirements/*.txt` files still have that message:

```
# The following packages are considered to be unsafe in a requirements file:
# setuptools
```

mistercrunch:
@john-bodley I think I addressed your point, let me know if there's anything else

rusackas (approved):

PR #27411, opened 2024-03-06
fix(dashboard): Only fetch CSS templates for dashboard header menu when in edit mode

mskelton:
### SUMMARY

When embedding a superset dashboard with a guest token, opening the header menu will present an error toast due to failing to load CSS templates. These are only used when in `editMode`, so this PR ensures we only load them if you are in edit mode.

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina (changes requested):
Thanks for the PR @mskelton. I think the correct solution here is to move the logic to fetch the templates to the `CssEditor` component which will only be rendered in edit mode.

mskelton:
@michael-s-molina I've updated the PR with that change.

michael-s-molina (approved):

PR #27415, opened 2024-03-06
fix(maps): Rename Ukrainian City to Ukrainian format and deccomunization

0xFillin:
### SUMMARY
The city name is listed as “Kiev,” which is the russian variant. + Rename old city name Dnipropetrovsk to Dnipro

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


villebro:
Thanks for the PR @Fillin-Trader. It would be helpful if you can add references in the description to support the proposed changes, as these are likely not obvious to someone without in depth knowledge of the area and how the names are translated to English.

rusackas:
If we want this change to "stick" we should make an edit in the Jupyter notebook, rather than editing a generated file which might get overridden.

villebro:
> If we want this change to "stick" we should make an edit in the Jupyter notebook, rather than editing a generated file which might get overridden.

Very good point @rusackas - let's make sure at least the notebook has this update, and even better, requesting the update at the source, rather than as an override in our script.

rusackas:
Ooh, that's even better. If NaturalEarth would reflect this in their data, the fix would be far more impactful than affecting only Superset.

rusackas:
One bit of technical feedback here: This geojson will be overwritten whenever anyone runs the Jupyter Notebook, and pulls new data from Natural Earth. So if you want to make changes to the geojson file, it should be done in the notebook as part of that process/worflow.

To address the elephant in the room: This does indeed feel a little political in nature, which is something we generally try not to dabble in in general. For instance, when there is a boundary dispute, we both sides "win" in the maps. With a given name, we don't usually default to the region's own language for the name; for example, we have a Spain map... not an España map. However, in this case, things are a little different, since indeed Kiev was imposed as a standard the USSR, which clearly no longer exists in its own right. Ukraine has been trying to adjust this standard since its independence in 1991, which far predates the current geopolitical struggles it faces, so I think there's a strong case to change it. Additionally, in 2019, the United States Board on Geographic Names officially changed its official spelling to "Kyiv," so I believe this change has merit if we're going to refer to other regions/territories by their American-English names (e.g. Spain). In other words, while this feels like stepping in controversy, I feel like this is actually consistent, current, accurate, and I'd have a significantly harder time making a case for the opposite.

rusackas (approved):
Approving for the Examples dashboard change, even if the geojson is at risk of being overwritten/reverted.

rusackas:
Closing this since it's not fixed in the notebook, and at least half of it was covered by this PR: https://github.com/apache/superset/pull/28226

If you want to rename Dnipropetrovsk to Dnipro (fine with me!) then please open a new PR that updates the Jupyter Notebook, or file an [issue](https://github.com/apache/superset/issues/new/choose) so someone else can fix it when they have time.

PR #27422, opened 2024-03-07
test(Migration to RTL): Refactor ActivityTable.test.tsx from Enzyme to RTL

rtexelm:
### SUMMARY

Replace enzyme with RTL functions. Refactor the tests accordingly.

### TESTING INSTRUCTIONS

`npm test ActivityTable.test.tsx`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/home/ActivityTable.test.tsx
+++ b/superset-frontend/src/features/home/ActivityTable.test.tsx
@@ -68,70 +68,71 @@ fetchMock.get(dashboardsEndpoint, {
   ],
 });
 
-describe('ActivityTable', () => {
-  const activityProps = {
-    activeChild: TableTab.Created,
-    activityData: mockData,
-    setActiveChild: jest.fn(),
-    user: { userId: '1' },
-    isFetchingActivityData: false,
-  };
+const mockSetActiveChild = jest.fn();
 
-  const activityViewedTabProps = {
-    activeChild: TableTab.Viewed,
-    activityData: mockData,
-    setActiveChild: jest.fn(),
-    user: { userId: '1' },
-    isFetchingActivityData: false,
-  };
+const activityProps = {
+  activeChild: TableTab.Created,
+  activityData: mockData,
+  setActiveChild: mockSetActiveChild,
+  user: { userId: '1' },
+  isFetchingActivityData: false,
+};
 
-  const emptyActivityProps = {
-    activeChild: TableTab.Created,
-    activityData: {},
-    setActiveChild: jest.fn(),
-    user: { userId: '1' },
-    isFetchingActivityData: false,
-  };
+const activityViewedTabProps = {
+  activeChild: TableTab.Viewed,
+  activityData: mockData,
+  setActiveChild: mockSetActiveChild,
+  user: { userId: '1' },
+  isFetchingActivityData: false,
+};
 
-  const renderOptions = {
-    useRedux: true,
-    useRouter: true,
-  };
+const emptyActivityProps = {
+  activeChild: TableTab.Created,
+  activityData: {},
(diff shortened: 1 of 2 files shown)


superset-frontend/src/features/home/ActivityTable.test.tsx
```diff
@@ -85,55 +78,61 @@ describe('ActivityTable', () => {
...
-    act(() => {
-      const handler = wrapper.find('[role="tab"] a').at(1).prop('onClick');
-      if (handler) {
-        handler({} as any);
-      }
-    });
+    renderActivityTable(activityProps);
+    const editedButton = screen.getByText(/edited/i);
+    expect(editedButton).toBeInTheDocument();
+    userEvent.click(editedButton);
```
justinpark, line 121: Could you kindly make use of `import { fireEvent } from spec/helpers/testing-library` as opposed to userEvent for click events? We typically reserve userEvent for certain custom events, such as a file upload event.

```suggestion
    fireEvent.click(editedButton);
```

superset-frontend/src/features/home/ActivityTable.test.tsx
```diff
@@ -85,55 +78,61 @@ describe('ActivityTable', () => {
...
     const dashboardCall = fetchMock.calls(/dashboard\/\?q/);
     const chartCall = fetchMock.calls(/chart\/\?q/);
-    // waitforcomponenttopaint does not work here in this instance...
+    // waitFor() does not work here in this instance...
```
justinpark, line 124: could you try with
```
await waitFor(() => expect(fetchMock.calls(/chart\/\?q/).toHaveLength(1));
```

rtexelm: that was actually my original attempt but the test will not pass, seems to be the same problem the previous author ran into. Not sure what the diagnosis is but it is passing with setTimeout

rtexelm: attempted adding a 5s timeout to the `waitFor()` with no change in result but the test passes in <20ms when using `setTimeout`. the only feedback I'm getting is a deprecation warning from the `moment` library concerning use of `non RFC2822/ISO date formats` probably from the `changed_on_utc` value in the `fetchMock`s

justinpark: @rtexelm it appears that the API fetch is only triggered when the 
`activeChild` prop is updated to `TableTab.Edited`.
Hence, it's being called before the case, which seems to be incorrect.

https://github.com/apache/superset/blob/47ae9d4cc3ca94332cb14359fb8a306f91c2da60/superset-frontend/src/features/home/ActivityTable.tsx#L142-L146

Kindly consider using `rerender` with `activeChild={TableTab.Edited}` (which is anticipated to be propagated after `setActiveChild()` has been called). This method should be a right direction whether the fetch has been called correctly.

rtexelm: Fantastic! I didn't know the usefulness of `rerender`. That looks like it's worked. The fetchMock calls don't register but the mockData is rendered without specifying `TableTab.Edited` as a replete tab

rtexelm:
Added a small change in `ChartTable.test` that was a leftover from merging `ActivityTable.test`. It just moved `fireEvent` to `userEvent`

rtexelm:
@justinpark is this looking better?

mistercrunch (approved):
LGTM

mistercrunch:
looks like you'll have to rebase, we address the issue with the missing required checks.

justinpark (commented):
LGTM!

PR #23762, opened 2023-04-21
fix: repeated color in the same chart

lilykuang:
### SUMMARY

Currently, the charts are experiencing an issue where colors are being repeated even though the color palette has more colors than the chart requires. This is problematic because it can make it difficult for users to distinguish between different data points and can lead to confusion. the code will now check colors in shared label and remove the color for the list.
- add a feature flag 
- implement a function to remove shared label color from range  

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:
before
After:
after

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27439, opened 2024-03-08
fix(Dashboard): Color inconsistency on refreshes and conflicts

### LINKED ISSUE #16970: [dashboard][color theme] chart color are changed when refreshing dashboard
chart color are changed when refreshing dashboard

#### How to reproduce the bug

1. Go to dashboard with multiple charts
2. Applied color theme to dashboard, for example supers color
2. Refresh dashboard
3. Observe color of each chart

### Expected results

Refresh dashboard did not change color of chart

### Actual results

Refresh dashboard did not change color of chart

#### Screenshots
https://user-images.githubusercontent.com/81597121/135937535-016cd613-cbf5-48a4-9849-7b0010a68176.mov

### Environment

- superset version: master
- node.js version: `node -v`
- any feature flags active:

### Checklist

### LINKED ISSUE #25384: Different colours for the same metrics
When I update page with dashboard in browser, I can see different colours for the same metrics
- helm chart version: `0.10.8`
- superset version: `3.0.0`
- python version: `3.10`

There is no such problem In the previous version superset 2.1.0

image

image

### LINKED ISSUE #26456: [dashboard][color theme] chart color are changed when refreshing dashboard
chart color is changed when refreshing the dashboard
The chart Color Scheme  Keeps changing in the following scenarios :
1. For the Same User viewing the same dashboard.
2. For different users viewing the same dashboard.

#### How to reproduce the bug

1. Go to the dashboard with multiple charts
2. Applied color theme to the dashboard, for example, superset color
3. Refresh dashboard
4. Observe the color of each chart
5. See error

### Expected results

Refresh dashboard should not change the color of the chart

### Actual results

Refreshing the dashboard changes the color of chart

#### Screenshots

Image 1& 2 shows the same chart with different color schemes opened in a fraction of a few seconds by the same user.
image
image
Images 3 & 4 are color schemes of the same chart for different users.
image
image

### Environment

- browser type and version: Chrome Version 119.0.6045.199 
- superset version: `3.0.1`
- node.js version: `node -v`
- any feature flags active:
1. AVOID_COLORS_COLLISION = True (By default)

### Checklist

- [.] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [.] I have reproduced the issue with at least the latest released version of superset.
- [.] I have checked the issue tracker for the same issue and I haven't found one similar.

### Additional context
I saw a similar issue happening before and attaching the issue link which is closed for now.
https://github.com/apache/superset/issues/16970

### LINKED ISSUE #28104: Charts Changing Color with Each Refresh
image
In Superset We are disabled from modifying the color scheme of the charts, it says it will be following the dashboard color scheme,
and each time we refresh the dashboard a new color will be assigned to the chart,

Even if try to clear the dashboard color scheme it will be automatically rapper

geido:
### SUMMARY
After the changes in this PR https://github.com/apache/superset/pull/23762 color inconsistency on refreshes was reintroduced.

Fixes #16970
Fixes #26456
Fixes #28104
Fixes #25384 (specifically for metrics changing colors on refreshes. An additional fix is required for charts emitting only COUNT(*) as label and not specific ones)

This PR does the following:

- Fixes color inconsistency on refreshes
- Fixes color inconsistency from chart in hidden tab to Explore
- It helps avoiding color collisions by adding a usage tracker of the colors at the chart level
- Deprecates AVOID_COLORS_COLLISION feature flag
- Renames the shared labels color map to `LabelColorsMap` 
- Centralizes the logics of generating the colors map and color metadata

N.A.

### TESTING INSTRUCTIONS
1. Set a color scheme for a dashboard
2. Refresh the dashboard
3. Colors should not change
4. Going from any chart in any tab to Explore, should carry the same colors forward

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


RESOURCES/FEATURE_FLAGS.md
```diff
@@ -84,6 +84,7 @@ These features flags currently default to True and **will be removed in a future
 
 [//]: # "PLEASE KEEP THE LIST SORTED ALPHABETICALLY"
 
+- AVOID_COLORS_COLLISION
```
rusackas, line 87: Is this just for testing, and probably removal before the next version? Or do we expect this to live a while and need a deprecation cycle?

rusackas, line 87: We might technically need to get consensus on this. @michael-s-molina any problems getting consensus and doing this now, or should we wait until the 5.0 proposals open up?

superset-frontend/cypress-base/cypress/e2e/dashboard/editmode.test.ts
```diff
@@ -210,7 +211,7 @@ describe('Dashboard edit', () => {
         '[data-test-chart-name="Top 10 California Names Timeseries"] .line .nv-legend-symbol',
       )
         .first()
-        .should('have.css', 'fill', 'rgb(69, 78, 124)');
+        .should('have.css', 'fill', 'rgb(31, 168, 201)');
```
rusackas, line 214: I guess we could use the color palette here, if we want the test not to break when if screw with the palette, but that seems negligible ¯\\\_(ツ)_/¯

RESOURCES/FEATURE_FLAGS.md
michael-s-molina: I think waiting for 5.0 would be a good idea as it allows an orderly evaluation of proposed deprecations. We just need to add a card to the board to remind us to submit the proposal.

geido: @michael-s-molina @rusackas this feature flag introduced a regression, so I don't think it makes sense to discuss depreciation. I am keeping it here and not deleting it for now just because we don't have a breaking window open. Thoughts?

villebro:
@geido would it be possible to add a brief description of the mechanics of the changes in this PR? This would help grasp the chosen approach without having to dive directly into the diff.

superset-frontend/packages/superset-ui-core/src/color/CategoricalColorScale.ts
```diff
@@ -39,99 +38,172 @@ class CategoricalColorScale extends ExtensibleFunction {
...
+  }
+
+  /**
+   * Lower chances of color collision by returning the least used color
+   * Checks across colors of current slice within SharedLabelColorSingleton
+   *
+   * @param currentColor
+   * @returns the least used color that is not the excluded color
+   */
+  getNextAvailableColor(currentColor: string) {
```
geido, line 176: This is meant to help with color collisions. It orders the colors based on their usage count in the chart. Least used colors will be preferred. This does not guarantee full protection against color collisions, which should be implemented at the individual viz based on their shape.

superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx
```diff
@@ -256,10 +264,51 @@ export class Tabs extends React.PureComponent {
     }
   };
 
+  // the initial color map is generated on save and catches all rendered charts
```
geido, line 267: Currently the initial color map is stored in the dashboard metadata at save but fails to map colors of charts that are hidden within tabs as those are not rendered. This checks for new colors popping in the color map that are not in the stored metadata and updates the metadata accordingly. The stored color map is then used by the viz in Explore to respect the colors of the dashboard context.

geido: Ideally this would not be necessary. A chart could expose a post-render hook that will individually check whether its color map is correctly updated in the stored metadata. However, this might have performance implications as multiple requests would need to be made. A larger refactor of the implementation would be required to account for efficiency in updating the color map and performance.

geido:
> @geido would it be possible to add a brief description of the mechanics of the changes in this PR? This would help grasp the chosen approach without having to dive directly into the diff.

@villebro this is meant to fix the regression introduced in https://github.com/apache/superset/pull/23762 so there is not much more context to add. However, I added inline comments to the parts of the code that I think introduce new behaviours or that might require clarifications.

geido:
A few thoughts about the current state of the implementation and the way forward:

- `SharedLabelColorSingleton` was meant to store only those labels that are shared across multiple charts. However, that is not the case since this is used to store all label colors independently if shared or not. This should get a better name, probably `LabelColorsMap` and change its references everywhere.
- There are multiple places that are storing the color map in the dashboard metadata, including saving the properties and clicking on tabs (as of the changes in this PR), on `verifyUpdateColorScheme` and others. A larger refactoring work is required to coordinate and centralize changes to the metadata. For example, we might centralize updating the metadata when individual tabs load
- `CategoricalColorScale` and `CategoricalColorNamespace`, as well as `SharedLabelColorSingleton` should be refactored to improve the hierarchy and lower interdependencies. This PR attempts to improve on that but more refactoring work is required.

CC @eschutho

superset-frontend/packages/superset-ui-core/src/color/CategoricalColorScale.ts
```diff
@@ -39,99 +38,172 @@ class CategoricalColorScale extends ExtensibleFunction {
 
   scale: ScaleOrdinal<{ toString(): string }, string>;
 
-  parentForcedColors: ColorsLookup;
-
   forcedColors: ColorsLookup;
 
+  sharedColorMapInstance: ReturnType<typeof getSharedLabelColor>;
+
+  sliceMap: Map<string, string>;
```
kgabryje, line 45: I'm assuming it's a mapping of series/dimensions and colours assigned to them? If so, could we maybe change that name, since we're often calling saved charts "slices", which could be confusing here?

geido: Sure makes sense

betodealmeida (approved):
Looks good to me, and I really like your thoughts on moving forward — this seems very complicated as is today.

kgabryje (approved):
LGTM, sorry for the long delay after my first review!

mistercrunch:
Rebasing should fix the issues around required checks

geido:
Thanks everyone for the reviews. After some more thinking, I will continue with the refactoring work on this branch so that we can merge a stable solution to this problem.

mistercrunch:
I'd love to see a user-oriented FAQ entry "How do colors get assigned to my dashboard?"

sadpandajoe:
I think I may have broken a chart if giving too many data points
Screenshot 2024-06-06 at 10 11 23 AM

Repro steps:
* Create a pie chart
* Dimensions: name
* Metric: sum(rank)
* row limit: 100
* color scheme: red to yellow
* This will create a chart
* bump the row limit to 10000 
* This will show white as a color which is not part of the red to yellow palette and if you hover over the white slice, there are actually multiple items in there

sadpandajoe:
Color collision still happens sometimes on stacked bar charts
Screenshot 2024-06-06 at 10 23 15 AM

Steps to repro:
* Create a bar chart with video games sales dataset
* x-axis: platform
* metric: sum(rank)
* dimensions: name
* row limit: 100
* color theme: blue to green
* stacked style: stack

notice that the GC and PS4 bar have the same color on top of each other

geido:
> Color collision still happens sometimes on stacked bar charts Screenshot 2024-06-06 at 10 23 15 AM
> 
> Steps to repro:
> 
> * Create a bar chart with video games sales dataset
> * x-axis: platform
> * metric: sum(rank)
> * dimensions: name
> * row limit: 100
> * color theme: blue to green
> * stacked style: stack
> 
> notice that the GC and PS4 bar have the same color on top of each other

After a deeper investigation, it appears that this issue is due to how data is grouped. This is one of those issues that requires a dedicated color algorithm as the default one can only check for the color usage in the order as the data is received, not in the way it is visualized.

geido:
> I think I may have broken a chart if giving too many data points Screenshot 2024-06-06 at 10 11 23 AM
> 
> Repro steps:
> 
> * Create a pie chart
> * Dimensions: name
> * Metric: sum(rank)
> * row limit: 100
> * color scheme: red to yellow
> * This will create a chart
> * bump the row limit to 10000
> * This will show white as a color which is not part of the red to yellow palette and if you hover over the white slice, there are actually multiple items in there

This seems to be a problem with the Pie chart in general as I can reproduce it on master too. I'll track it for a separate investigation.

PR #27442, opened 2024-03-08
fix(drilling): drill by pagination works with MSSQL data source

### LINKED ISSUE #24072: MSSQL requires an order_by when using an OFFSET or a non-simple LIMIT clause
#### How to reproduce the bug

### Expected results

### Actual results

#### Screenshots

### Environment

- superset version: `2.1.0`
- node.js version: -

### Checklist

sfirke:
### SUMMARY
fixes #24072 in which Microsoft SQL Server requires an order-by column in order for pagination to succeed.  In doing so, now the drilled data is always sorted by the 1st column in ascending order.  Previously it retained the dataset's sort order.

Better fixes outside of my ability might be:
- Implement this on a per-database level
- Sort by row number or the equivalent to functionally do nothing and preserve the underlying data order
- Improve user controls for how drilled data displays, allowing the user to control which columns are shown and what the sort order is

But I think the harm introduced (re-sorting the drilled data) is outweighed by fixing this feature for MSSQL.

### TESTING INSTRUCTIONS
- CI
- Sam deployed and manually tested the docker image

**BEFORE (4.0.0rc1**

https://github.com/apache/superset/assets/7569808/18506aea-ea2b-4207-b97a-c1c9b7debf3b

**AFTER**

https://github.com/apache/superset/assets/7569808/0fad13b6-bb8f-4e35-8ae5-971d7fda6462

### ADDITIONAL INFORMATION

- [x] Has associated issue: #24072

Update the drill-to-detail functionality to ensure drill by pagination works with MSSQL data sources, including adjustments to test expectations and query ordering.

These changes fix an issue where the drill feature did not work correctly with MSSQL due to ordering issues, by integrating proper ordering logic in the query and aligning test data with the new expected behavior. This approach ensures compatibility with additional data sources such as MSSQL, while maintaining existing functionality.


john-bodley:
@sfirke once this is ready for review would you mind adding the relevant reviewers?

michael-s-molina:
@sfirke the CI error you're getting

`superset/common/query_actions.py:156: error: List item 0 has incompatible type "Tuple[Union[AdhocColumn, str], bool]"; expected "Tuple[Union[AdhocMetric, str], bool]"  [list-item]`

is because you're passing a `Column` here

query_obj.orderby = [(query_obj.columns[0], True)]

 instead of a `Metric` which is what `OrderBy` expects.

```
Column = Union[AdhocColumn, str]
Metric = Union[AdhocMetric, str]
OrderBy = tuple[Metric, bool]
```

superset/common/query_actions.py
```diff
@@ -154,6 +153,7 @@ def _get_samples(
             qry_obj_cols.append(o.get("column_name"))
         else:
             qry_obj_cols.append(o.column_name)
+    query_obj.orderby = [(query_obj.columns[0], True)]
```
michael-s-molina, line 156: I don't have much context here but given that we're explicitly setting:

```
query_obj.orderby = []
query_obj.metrics = None
```
and that the definition of `OrderBy` is:

```
OrderBy = tuple[Metric, bool]
```

Setting the order by to the first column might not be the correct answer given that it only accepts ordering by metric which is explicitly being set to empty. Maybe @zhaoyongjie or @villebro have more context and can help us.

sfirke: Thanks for the feedback @michael-s-molina 🙏 !  Someone in the associated issue thread said this one-liner was a solution for them so I hoped it would be this simple but it looks like not.

@zhaoyongjie gave a [high-level description of how he thinks this should be fixed](https://github.com/apache/superset/issues/24072#issuecomment-1859252019).  Implementing it is outside of my current understanding unfortunately.

sfirke:
I am seeing the Docker / docker-build (ci) step is succeeding, but I don't see a corresponding image on Docker Hub: https://hub.docker.com/r/apache/superset/tags  Am I missing it there when I look + search, or is it not pushing to Docker Hub?  It seems like other PR images are.  That would allow me to deploy the image and test it thoroughly.

rusackas:
I think that's correct, that PRs build the docker image (as a bit of a smoke test) but don't push them to dockerhub (to reduce clutter). I think @mistercrunch set up a command in his recent work to manually publish a branch's docker image, but I can't recall what it is :)

sfirke:
Ah indeed, the PR whose image I see on Docker Hub is one from @mistercrunch: https://github.com/apache/superset/pull/27465  Max, is there a command I can use to push a PR branch image to Docker Hub?

superset/common/query_actions.py
sfirke: Evan suggested changing the definition of OrderBy and that seems to have worked.

mistercrunch:
@sfirke as far as I know there isn't a way to do this in this context (PR from a fork) because of potential security issues. Docker creds just aren't available in this context, if they were, an attacker could push arbitrary code to the `latest` image for instance.

You should be able to `docker build` locally at will though, any reason why you want CI to build it for you?

rusackas:
I wonder, would it work on your fork if you added your own dockerhub secrets/credentials to your own github fork?

mistercrunch:
`supersetbot` acts as a neat wrapper around `docker build` and could be expanded to support other hub repos. The main win from using the wrapper is setting the read cache to the location used by `master` builds, meaning you potentially have a lot less to build locally. It also has a notion of build "presets" that point to combination of layers and settings. 

```
$ nvm use 20
$ npm i -g supersetbot
$ supersetbot docker --help
Usage: supersetbot docker [options]

Generates/run docker build commands use in CI

Options:
  -t, --preset                  Build preset
  -c, --context <context>       Build context (default: "local")
  -r, --context-ref <ref>       Reference to the PR, release, or branch
  -p, --platform <platform...>  Platforms (multiple values allowed)
  -f, --force-latest            Force the "latest" tag on the release
  -v, --verbose                 Print more info
  -h, --help                    display help for command
$ supersetbot docker --dry-run
docker buildx build \
      --load \
      -t apache/superset:8ed368c1fe624e6288d9270997098732d2289f7c-arm \
        -t apache/superset:8ed368c-arm \
      --cache-from=type=registry,ref=apache/superset-cache:3.9-slim-bookworm \
       \
      --target lean \
      --build-arg PY_VER=3.9-slim-bookworm \
      --platform linux/arm64 \
      --label sha=8ed368c1fe624e6288d9270997098732d2289f7c \
      --label target=lean \
      --label build_trigger= \
      --label base=3.9-slim-bookworm \
      --label build_actor=undefined \
      .
```
notice the `--cache-from`

sfirke:
I'm stuck, it's unclear to me why this test is failing:

```
___________ TestPostChartDataApi.test_chart_data_applied_time_extras ___________

self = <tests.integration_tests.charts.data.api_tests.TestPostChartDataApi testMethod=test_chart_data_applied_time_extras>

    @pytest.mark.usefixtures("load_birth_names_dashboard_with_slices")
    def test_chart_data_applied_time_extras(self):
        """
        Chart data API: Test chart data query with applied time extras
        """
        self.query_context_payload["queries"][0]["applied_time_extras"] = {
            "__time_range": "100 years ago : now",
            "__time_origin": "now",
        }
        rv = self.post_assert_metric(CHART_DATA_URI, self.query_context_payload, "data")
>       self.assertEqual(rv.status_code, 200)
E       AssertionError: 400 != 200
```

And why is `test-postgres-presto` passing but `test-postgres-hive` not?

rusackas:
Looks like this is pretty stuck. Maybe it needs a rebase? I'll convert it to draft for now, but mark it as ready if/when you think it might be.

sfirke:
"Draft" sounds right, I'd like to see this through but won't get to it
anytime soon

rusackas:
@sfirke is this still WIP?

sfirke:
@rusackas yes, alas. I will still get to this _sometime_, but not imminently.

sfirke:
@rusackas I have superseded this pull request with a new version, #34583, that is passing CI. It needs a review but maybe we can get this into 6.0 and 5.0.1!

PR #27448, opened 2024-03-08
fix(package-lock): node_modules/@react/types to 16.9.53

justinpark:
### SUMMARY
During the implementation of [#26011](https://github.com/apache/superset/pull/26011/files#diff-535054a5c3b273a1f8a602847fd9d9a0b881058725b3565b590f08c0dabac187R19622), @types/react was upgraded to `16.9.53`, and as a result, package-lock selected the `16.14.51` version.

feat_telemetry___Adding_Scarf_based_telemetry_to_Superset___26011__·_apache_superset_8437a23

Here at Airbnb, our internal dependencies are not yet compatible with `types/react@16.14.51`, thereby posing a hurdle for our release. We intend to explore compatibility with 16.14.51 in Superset 5.0 or later versions, although the timing is not ideal right now.

This commit modifies the `package-lock` to reflect the same version number as outlined in #26011.
Simultaneously, we've also lowered @ant-design/icons@5.2.6 (#26727) to 5.0.1, due to its incompatibility with 16.9.53.

### TESTING INSTRUCTIONS
npm run type

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/TelemetryPixel/index.tsx
+++ b/superset-frontend/src/components/TelemetryPixel/index.tsx
@@ -47,7 +47,6 @@ const TelemetryPixel = ({
   const pixelPath = `https://apachesuperset.gateway.scarf.sh/pixel/${PIXEL_ID}/${version}/${sha}/${build}`;
   return process.env.SCARF_ANALYTICS === 'false' ? null : (
     <img
-      // @ts-ignore
       referrerPolicy="no-referrer-when-downgrade"
       src={pixelPath}
       width={0}
--- a/superset-frontend/package.json
+++ b/superset-frontend/package.json
@@ -80,7 +80,7 @@
     "last 3 edge versions"
   ],
   "dependencies": {
-    "@ant-design/icons": "^5.2.6",
+    "@ant-design/icons": "^5.0.1",
     "@emotion/babel-preset-css-prop": "^11.11.0",
     "@emotion/cache": "^11.4.0",
     "@emotion/react": "^11.4.1",


superset-frontend/src/components/TelemetryPixel/index.tsx
```diff
@@ -47,7 +47,6 @@ const TelemetryPixel = ({
   const pixelPath = `https://apachesuperset.gateway.scarf.sh/pixel/${PIXEL_ID}/${version}/${sha}/${build}`;
   return process.env.SCARF_ANALYTICS === 'false' ? null : (
     <img
-      // @ts-ignore
```
justinpark, line 50: This `referrerPolicy` value is compatible with `16.9.53`

mistercrunch (changes requested):
Can guarantee that this is bound to come up again as next week @dependabot is going to try to push this up again. You should put some sort of comment around the pin around why it's pinned, that or a dependabot rule with associated comment/documentation as to why it needs to be pinned - or both

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

I'm upset that `json` and therefore `package.json` doesn't support comments! - but [there are hacks to bring comments in there](https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install). I wonder if we could move to a `package.js` but some tools don't support it (say unclear if dependabot works when using a package.js?).

justinpark:
> Can guarantee that this is bound to come up again as next week @dependabot is going to try to push this up again. You should put some sort of comment around the pin around why it's pinned, that or a dependabot rule with associated comment/documentation as to why it needs to be pinned - or both

I identified the compatible version for our internal dependencies, so we can proceed to close this pull request as it stands now. As previously discussed, my intention is to incorporate TypeScript checking into our CI step. Hopefully, this will help us to identify any Dependabot issues at an earlier stage.

PR #27454, opened 2024-03-10
# build(deps): bump antd from 4.19.3 to 5.14.2 in /docs

morpheus-lee:
same as #27330

### DIFF
--- a/docs/src/styles/main.less
+++ b/docs/src/styles/main.less
@@ -16,8 +16,6 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-@import '~antd/lib/style/themes/default.less';
-@import '~antd/dist/antd.less'; // Import Ant Design styles by less entry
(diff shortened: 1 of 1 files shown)


rusackas:
Hi... it looks like you deleted some LESS files, but didn't bump the package. It doesn't seem at all the same as the linked PR. Can you tell us more about what you're looking to do here?

i-love-thinking:
> Hi... it looks like you deleted some LESS files, but didn't bump the package. It doesn't seem at all the same as the linked PR. Can you tell us more about what you're looking to do here?

Base on [antd V4 to V5](https://ant.design/docs/react/migration-v5)
Because antd V5 has been migrated to css-in-js, and its related less files have been removed. So I just deleted the import statement that shouldn't be there. Then, I manually checked the compilation results, and there was no big problem.

Issue #26423, opened 2024-01-07
Countries map - Add Turkey Regions in addition to Turkey Departments


iskenderulgen:
### Discussed in https://github.com/apache/superset/discussions/26422

<div type='discussions-op-text'>

<sup>Originally posted by **iskenderulgen** January  7, 2024</sup>
### SCREENSHOT

The desired map output according to the region information of Turkey is as follows. 

![turkey_regions](https://github.com/apache/superset/assets/15057168/d2fa7984-65f6-482d-be3d-857e2f5cfa2c)

### Description 

- In addition to Turkey's departmens, there is also need for Turkey's Regional map.
- The map representation at NUTS1 (region) level is expected to enable Superset users to visualize high-level analyses according to their needs.

### Design Input

We could follow what has been done for Italy's and France's Regions (sample line in code base:
(https://github.com/apache/superset/blob/master/superset-frontend/plugins/legacy-plugin-chart-country-map/src/countries.ts#L210)

### Proposition

ISO 3166-2 code is presented for city level such as TR-34 (Istanbul). However, after some search, there seems to be less or no information about Turkey's ISO 3166-2 codes for regional information. Region information is obtained from Nomenclature of Territorial Units for Statistics (NUTS) which has been widely used for Turkey since 2002. Turkish Statistical Institute (TUIK) provides board level information using NUTS codes which is of great benefit to oraganizations and companies. In view of the lack of ISO 3166-2 code information for Turkey at the region level and the wide use of NUTS codes, we could associate 12 region polygons with NUTS representations for **ISO** and **NAME_1** keys of geojson. For future if or when we can find ISO 3166-2 on region level, we can create new PR to change the data structure.

Ref= https://en.wikipedia.org/wiki/ISO_3166-2:TR
Ref=https://en.wikipedia.org/wiki/NUTS_statistical_regions_of_Turkey</div>

iskenderulgen:
This issue is completed with pull request #27455

PR #27455, opened 2024-03-10
feat: Add Turkey's regions to country map visualization

iskenderulgen:
### SUMMARY
In addition to Turkey's departmens, there is also need for Turkey's Regional map for geospatial statistical inference with organizational data. The regional polygons are created from city level polygon using geopandas. 
* Original city-wise polygons are acquired from https://data.humdata.org/dataset/cod-ab-tur?
* ISO 3166-2 code is presented for city level such as TR-34 (Istanbul). However, after some search, there seems to be less or no information about Turkey's ISO 3166-2 codes for regional information. Region information is obtained from Nomenclature of Territorial Units for Statistics (NUTS) which has been widely used for Turkey since 2002. Turkish Statistical Institute (TUIK) provides board level information using NUTS codes which is of great benefit to oraganizations and companies. In view of the lack of ISO 3166-2 code information for Turkey at the region level and the wide use of NUTS codes, we associated 12 region polygons with NUTS representations for ISO and NAME_1 keys of geojson. For future if or when we can find ISO 3166-2 on region level, we can create new PR to change the data structure. Considering majority of the spatial data coming from TUIK, polygon will be easy to join with the TUIK data.

Ref= https://en.wikipedia.org/wiki/ISO_3166-2:TR
Ref=https://en.wikipedia.org/wiki/NUTS_statistical_regions_of_Turkey

* The data structure is as follows
Screenshot from 2024-03-09 18-22-37

* Original regions geojson was 13MB of size which could crash superset on production. To mitigate that problem, we used simplify(tolerance=0.005, preserve_topology=True). After simplification the accuracy of the polygons are checked againts the original file with the area functions and the difference was -0.01, which is acceptible considering the size reduction from 13mb to 448kb. 
* The geojson file follows the same structure with Italy's regions such as "urn:ogc:def:crs:OGC:1.3:CRS84". 
* Gejson file both tested with locally on python environment and with Apache Superset 3.1 version. Images are presented as follows.
* Upon acceptance of the Region based feature addition. Sub regional data (26 region) and district level (973) polygons will be PR'ed with NUTS codes from TUIK institution. With these features complete granularity will be added to Apache superset for Turkey's locations. 

## Turkey's cities geojson file which is used to create regional based polygons.
turkey cities

### Base plot from matplotlib
turkey_regions
### Colored plot from matplotlib
turkey_regions_colored
### Apache superset dev-environment test image on local test
* test is conducted by creating a new table and iserting toy data that matches the ISO (NUTS) codes. 
Screenshot from 2024-03-10 12-30-25

### ADDITIONAL INFORMATION

- [x] Has associated issue:  #26423
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Hi there... we're trying to manage all map contributions through the Jupyter notebook, to make this area of the codebase more maintainable. Would you be willing to contribute the changes through there, so that we can more easily maintain this (and other) GeoJSON files?

iskenderulgen:
Hi, I'll check the jupyter file to move the process there. Only problem might be the ISO_3166-2:TR code for Turkey since they are not exist. If i can tackle with custom dictionary for NUTS codes, i'll update the codebase otherwise i'll need to stick with the custom created json file.

iskenderulgen:
@rusackas hello, the additions you requested are appended to the map generator notebook. Since there is no ISO codes for the country, NUTS codes are used for regions and a dictionary constructured to map cities into regions. Notebook tested locally and faced with no error. As mentioned above if / when ISO codes made available for Turkey i will make neccessary changes with new PR, until than, NUTS codes from TUIK institue will be used for all granularity apart from city-level.

rusackas:
There's a seemingly unrelated unit test here that keeps failing... would you mind rebasing?

Also, when I run the notebook locally, the geojson file that's generated by the notebook is different (and smaller?) than the one on this PR. I'm not sure what's up with that.

iskenderulgen:
Hi, i updated the branch to be in sync with superset:master, but have no clue why npm unit test is failing. 

The reason maps sizes are different is that, the first commit that included geojson (which is tested with superset development using docker) comes from OCHA HDX database. HDX Geojson was updated 202103 and the geospatial analytics team keeps the hdx database updated as the new changes arrives (we also use these adm1 - adm2 - adm3 maps  in our local prod environments, wanted to share json files with rest of the users). Thus, the map has more detailed representation for borders and we used these adm1 level data to create regions. Jupyter notebook uses natural earth data, even though we used same code on jupyter notebook, natural earth data has some differences on adm1 level. In the picture below we compared Ocha hdx, Natural earth and [Wiki page](https://en.wikipedia.org/wiki/NUTS_statistical_regions_of_Turkey) region map data and marked some differences.  If possible we would like to contribute with the json file created from HDX and keep code base on jupyter notebook as requested. Size difference is approximately 150KB (HDX is 338KB and Natural earth data is 190KB).  Would like to help if you have any other questions. 

region data

rusackas:
Hi again, and thanks for the continued engagement here. Indeed, we want to keep things as accurate (and fair) as possible when it comes to map data. If HDX is license-compatible with Apache projects (which seems to depend on the file), geopolitically uncontroversial (i.e. both sides win in a border dispute) and more accurate, then that's great! I think its be great to have the Jupyter notebook pull that from HDX as part of the notebook workflow, and perhaps add a clear link to a compatible license in the notebook comments or something along those lines. We're just trying to have a single point of maintenance for all this, and it's not good if the notebook generates a map that overwrites something superior :)

rusackas:
Hi there, and thanks for keeping this progressing. There are a couple of issues to contend with:
* It's not clear that the licensing is compatible. Is [this](https://data.humdata.org/dataset/cod-ab-tur) the best reference regarding the license? All it says, really, is that it's for humanitarian use only. I'm not sure if we qualify? Maybe! We might have to get ASF to weigh in ¯\_(ツ)_/¯ 
* The PR is not passing the linter checks... I added a comment on the PR about this, but we might as well fix it in the notebook so it outputs properly formatted code, rather than relying on the linter. Feel free to either fix the notebook, or just run the [pre-commit hooks](https://superset.apache.org/docs/contributing/hooks-and-linting/). Either way can get the PR across the finish line.
• I just ran the updated notebook, and it appears the geojson file the notebook outputs doesn't match what's on the PR. Not sure what's up with that difference. 

I wish it were easier to collaborate on this notebook, so I didn't have to ask for these changes. The process is getting better, but is not yet perfect :)

iskenderulgen:
hi, i inspected the HDX project metadata and the only licence indication was "[other : Shared by the General Command of Mapping for humanitarian use only](https://data.humdata.org/dataset/cod-ab-tur?)"  not much to detail about it and also HDX was not clear what "[other](https://data.humdata.org/faqs/licenses)" license mean. So for compatibility i reverted back to natural earth polygon for regions location. 

Also updated the notebook lint section to have same export that we had on the older commits this should solve the problem you commented. Also overwritten the hdx polygon with the data coming from natural earth file is now 107kb size. After commit i run the notebook again and the file output seems to be same on my setup. If you can elaborate the file difference on your end, happy to help to that :)

rusackas:
Hmm... weird. The geojson that I get as a result from my own notebook run (after pulling your latest changes) is also 107k. It's not an easy diff to make sense of, so I'll add a ZIP of it here in case you're able to spot something.
[turkey_regions.geojson.zip](https://github.com/apache/superset/files/14597228/turkey_regions.geojson.zip)

It sure would be nice to use this HDX data (which keeps making me think of [Home Depot](https://www.homedepot.com/b/HDX/N-5yc1vZ9tk)) - maybe one of us can concat them and/or ASF to see if it is indeed usable. The more current and geopolitically acceptable our data is, the better.

iskenderulgen:
I'll inspect the zip file that you provided and make area comparison of each region in case there is difference that requires handling. It would be really nice if it could be confirmed from HDX. I'll keep the HDX codebase and if HDX is usable for the project I can update in quick pr. 

Since there is no district level data on natural earth apart from USA, it could be usefull to include HDX db for detailed granularity. Alternative is I can use natural earth for region and sub_region but will need to switch to openstreemap for district level polygons. Best case is, up on agreement with HDX i can create all 4 maps (region - sub region - city level - district level) from one URL and one database. thanks for the help in advance.

iskenderulgen:
Could there be a dependency problem with npm package, front-end check is in progress like 3 hours and 30 minutes. Also is there  a something else that i should be working on to merge-close pr. Until official approval arrives from HDX we can use natural earh polygons for regions map. After approval i can start to work on polygons for all granularity on jupyter to update maps.

rusackas:
Approving CI... I'll approve and merge when possible. I was going to do a little notebook tidying, but you're welcome to make your changes first if you'd like. 

I haven't looked into the `italy_regions` map either, by the way. That's the only one left that isn't pulling from NaturalEarth data, if you want to contribute that so that we're finally 100% notebook-derived in our maps.

rusackas (approved):
Really appreciate all your work/input here :)

rusackas:
Were you planning to reach out to the HDX team for clarity on the license, or shall I?

rusackas:
I'll reach out and follow up! Lemme open a quick PR on the notebook, and then you're welcome to pull it and keep going on Italy regions or anything else :) I very much appreciate the help!

PR #27464, opened 2024-03-11
fix: pass valid SQL to SM

betodealmeida:
### SUMMARY

When creating a virtual dataset we pass the unprocessed SQL to security manager. If the SQL contains Jinja it might not be valid and might fail to parse, which causes the security manager to allow the query to run (see https://github.com/apache/superset/pull/26748).

This PR fixes it by processing the templated SQL before passing it to the security manager. See the discussion in https://github.com/apache/superset/pull/26476.

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


john-bodley:
Thanks @betodealmeida for authoring this fix. Thinking a bit about this over the weekend, I think this—a two phase approach—is likely the right technique. 

Interestingly previously using `sqlparse` (3.0.3) the following queries returned different results,

```python
>>> from superset.sql_parse import ParsedQuery

>>> ParsedQuery("SELECT * FROM foo WHERE ds = '{{ presto.latest_partition('foo') }}'").tables
{Table(table='foo', schema=None, catalog=None)}

>>> ParsedQuery("SELECT '{{ presto.latest_partition('foo') }}'").tables
set()
```

i.e., Superset only identified the `foo` table from the `FROM` clause and didn't identify the table from within the Jinja macro.

Finally, @betodealmeida regarding your comment, 

> When creating a virtual dataset we pass the unprocessed SQL to security manager.

This could also be a physical dataset which contains a freeform SQL filter containing a Jinja macro.

superset/commands/dataset/create.py
```diff
@@ -85,10 +88,12 @@ def validate(self) -> None:
             exceptions.append(TableNotFoundValidationError(table_name))
 
         if sql:
+            processor = jinja_context.get_template_processor(database=database)
```
john-bodley, line 91: @betodealmeida I think we need this in more places than just virtual datasets per my comment in the PR, i.e., a chart (via an ad-hoc filter), SQL Lab query, etc. could contain a Jinja template.

Should this two-phase approach be part of the `ParsedQuery` class so it's applicable to all use cases?

superset/commands/dataset/create.py
```diff
@@ -85,10 +88,12 @@ def validate(self) -> None:
...
+            rendered_sql = processor.process_template(sql)
```
john-bodley, line 92: The additional challenge is that—depending on the macro—an actual query may need to be executed in order to process the template, i.e., the `[hive|presto|trino].latest_partition` Jinja macro actually executes a query in order to resolve the macro. This results in somewhat of a chicken-and-egg scenario, i.e., one has to _try_ to execute a query (on behalf of the user) in order to resolve whether the user has access to it.

Furthermore if the user doesn't have access to the tables which are used to resolve the `[hive|presto|trino].latest_partition` Jinja macro then line #92 will raise an `SupersetDBAPIProgrammingError` (or similar).

I'm still not sure what the ideal solution is here. Without doubt this is non-trivial.

john-bodley:
Regarding https://github.com/apache/superset/pull/27464#discussion_r1520265947 and https://github.com/apache/superset/pull/27464#discussion_r1520393408, I was thinking maybe an alternative approach would be a multi-phase approach for the `ParsedQuery` table extraction where:

- Phase I: Identify all the Jinja macros—hopefully by leveraging Jinja functionality—and extract any table names present, i.e., we would need to parse (likely via SQLGlot) the macros

```python
>>> from sqlglot import parse
>>> parse("trino.latest_partition('foo')")
[Dot(
  this=Identifier(this=trino, quoted=False),
  expression=Anonymous(
    this=latest_partition,
    expressions=[
      Literal(this=foo, is_string=True)]))]
```

- Phase II: Replace the Jinja macros with some SQL placeholder* and then parse the full query to extract any remaining table names.

The extracted set of tables is then the union of phase I and phase II.

\* I wonder if `NULL` would work for all scenarios. We don't need to generate SQL which is actually executable, but merely SQL which SQLGlot will be able to successfully parse.

superset/commands/dataset/create.py
betodealmeida: @john-bodley as far as I could tell, in those places it's already handled correctly. For SQL Lab I can confirm that the template is processed before being passed to the security manager for parsing. I'll test with ad-hoc filters, but when I inspected the code I only saw this place in need of fixing.

betodealmeida:
@john-bodley why not simply translate the macro (by executing queries, if necessary) in order to have the actual query that's going to run, so we can check for all the tables used? I think it's the best way to prevent a tight coupling between Jinja templates and SQL parsing, and I'm worried that with the approach you're proposing we might have errors in the future when people introduce new macros and the logic to extract tables misses them.

Is your concern strictly performance (ie, running a query that might be expensive only for the security manager to block the main query)? Or are you worried that it might leak metadata, eg, with a query like this, where the user doesn't have access to table `bar`:

```sql
SELECT '{{ latest_partition('bar') }}';
```

Is a user knowing the latest partition of a table they don't have access to a problem? One solution to this problem would be checking the macro queries against the security manager as well, so the process would be:

1. Render all macros
2. When macros like `latest_parititon` render, they run their queries against the security manager before executing them
3. Run the rendered query against the security manager
4. Run query

Though it again requires new macros to adopt this pattern (of doing step 2) if we want to keep Superset secure.

john-bodley:
@betodealmeida I raised the issue with simply rendering all macros [here](https://github.com/apache/superset/pull/27464#discussion_r1520393408). The challenge is the `[hive|presto|trino].latest[_sub]_partition` macros likely will resolve to Hive, Presto, or Trino queries which need to be executed, however the user may not have access to the underlying table(s)—hence my chicken-and-egg statement. I don't believe it's viable to first run a precursor query in order to then resolve which tables are associated with the query and whether said user has access to them.

The logic I outlined in https://github.com/apache/superset/pull/27470 doesn't render all macros, which @michael-s-molina, @tobyma, et. al have correctly pointed out could be problematic—especially if there's SQL injection which likely would come from the `url_param` macro.

Regarding your four steps I think they would need to be tweaked—in essence a hybrid of the logic in this PR and https://github.com/apache/superset/pull/27470 as the `latest_[_sub]_parititon` macros need special  treatment. Specifically reordering/tweaking steps 1 and 2 and tweaking steps 3 and 4:

1. Via Jinja extract the table referenced in the `latest_[_sub]_parititon` macro and replace the contents with some benign query.
2. Render all remaining macros.
3. Extract the tables from the rendered query and combine with 1.
4. Verify that via the security manager that the user has access to said tables.
5. Run query.

https://github.com/apache/superset/pull/27470 currently defined steps 1, 3, 4, and 5. The only missing part is 2, rendering the SQL query to resolve any remaining macros.

john-bodley:
For reference [here's](https://github.com/apache/superset/blob/85d0d88fc276e147c51056aec87854e5feb4176e/superset/sqllab/query_render.py#L61-L70) another place in code (for right or wrong) which renders the Jinja template (for validation purposes), yet faces the same chicken-and-egg situation where any `latest[_sub]_partition` macros are executed prior to checking if the user has access.

john-bodley (changes requested):
Thanks @betodealmeida for syncing up earlier. 

I've tested this PR—in conjunction with https://github.com/apache/superset/pull/26748 —and the logic outlined in this PR isn't quite right, i.e., when exploring a chart using a virtual dataset the SQL statement being parsed still contained Jinja templating.

Per my previous https://github.com/apache/superset/pull/26476#issuecomment-1991117528 there are a number of vectors from explore where we check whether the user has access to either the raw SQL or `Query` object (either of which may contain Jinja templating), i.e., rather than just when creating a dataset per this PR.

It's probably prudent to adopt the "shift left" mentality—thus adhering to the DRY principle—and add the rendering logic from this PR  upstream of the various commands.

I think we (@michael-s-molina, you, and myself) agreed that (for forward compatibility) the soon to be replaced `ParsedQuery` object should only handle real SQL statements, rather than SQL-esque statements containing Jinja templating, hence it feels like the security manager is the best place to add said logic, i.e., [this](https://github.com/apache/superset/blob/master/superset/security/manager.py#L1963) line would be replaced with, 

```python
get_template_processor(database).process_template(query.sql)
```

in addition to the following import:

```python
from superset.jinja_context import get_template_processor
```

superset/security/manager.py
```diff
@@ -1956,11 +1957,14 @@ def raise_for_access(
                 return
 
             if query:
+                # make sure the quuery is valid SQL by rendering any Jinja
```
john-bodley, line 1960:
```suggestion
                # make sure the query is valid SQL by rendering any Jinja
```

tests/unit_tests/commands/dataset/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
john-bodley, line 1: @betodealmeida I gather this file is now obsolete.

john-bodley (approved):
Thanks @betodealmeida for the changes. I left a few small comments, but otherwise this LGTM.

superset/security/manager.py
betodealmeida: Good catch! Let me fix this in another PR so we can merge this already.

tests/unit_tests/commands/dataset/__init__.py
betodealmeida: Ah, good point.

Issue #26786, opened 2024-01-24
[SIP-117] Improve SQL parsing


betodealmeida:
### Motivation

The current status of SQL parsing in Superset is not ideal. A few reasons:

1. We currently use 3 libraries for parsing:
  a. `sqlparse` is the original library used for SQL parsing. It's non-validating and non-dialect specific, which brings a few challenges.
  b. `sqloxide` was introduced as an optional dependency, where performance was critical.
  c. `sqlglot` was introduced in https://github.com/apache/superset/pull/26476 to fix security issues.
2. Superset has an interface to SQL parsing in `superset.sql_parse.ParsedQuery`, but many places in the code call `sqlparse` directly. This includes complex operations, like injecting RLS rules into SQL queries.
3. `ParsedQuery` is sometimes used for single statements, sometimes for multi-statement queries, which could lead to subtle bugs, since it has methods that assume a single statement.
4. In many functions a query will be repeatedly parsed and converted back to a string, which is inefficient.
5. Parsing is done using a generic parser that is not dialect specific.

Superset needs robust and reliable SQL parsing, so it can understand and modify queries being run in programmatic ways. Today, many of the manipulations to SQL are done using string functions, because the `sqlparse` AST is too low-level. When queries are manipulated programmatically, like in RLS injection, the code is complex and hard to follow, for the same reason.

### Proposed Change

My proposal is to create 2 new classes that should provide a clean interface to SQL parsing:

```python
class SQLScript:

    """
    Class representing a SQL script (or block) with multiple statements.

    Examples:

        "SELECT 1; SELECT 2"  # 2 statements
        "CREATE TABLE a (b INT)"  # single statement
        ""  # 0 statements

    """

    statements: List[SQLStatement]

    def __init__(self, query: str, engine: str):
        ...


class SQLStatement:

    """
    Class representing a single SQL statement.

    Optionally in the future we might want to have derived classes for
    more specific statements, like DML or query.
    """
     
    def __init__(self, statement: str, engine: str):
        ...
```

(This is a simplified version, see https://github.com/apache/superset/pull/26767 for more details.)

These classes will provide:

1. A clear differentiation between a parsed query (with multiple statements), and a single statement.
2. Dialect-specific parsing, via a required attribute indicating the DB engine spec `engine`.
3. All SQL-parsing related functionality needed by Superset, so that the Superset code will only use these classes for anything related to SQL parsing, introspection, and manipulation. No SQL parsing library should be imported anywhere else in Superset.
[...]

PR #27470, opened 2024-03-12
fix(sql_parse): Ensure table extraction handles Jinja templating

john-bodley:
### SUMMARY

This is somewhat of an alternative formulation to https://github.com/apache/superset/pull/27464 by performing a two-phase approach for extracting tables from SQL which may contain Jinja templating, resulting in non-parsable SQL from SQLGlot's perspective. 

The logic is outlined in https://github.com/apache/superset/pull/27464#issuecomment-1989433700.

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes https://github.com/apache/superset/issues/27581
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -24,7 +24,7 @@ from typing import Any, Callable, cast, Optional, TYPE_CHECKING, TypedDict, Unio
 import dateutil
 from flask import current_app, has_request_context, request
 from flask_babel import gettext as _
-from jinja2 import DebugUndefined
+from jinja2 import DebugUndefined, Environment
 from jinja2.sandbox import SandboxedEnvironment
 from sqlalchemy.engine.interfaces import Dialect
 from sqlalchemy.sql.expression import bindparam
@@ -479,11 +479,11 @@ class BaseTemplateProcessor:
         self._applied_filters = applied_filters
         self._removed_filters = removed_filters
         self._context: dict[str, Any] = {}
-        self._env = SandboxedEnvironment(undefined=DebugUndefined)
+        self.env: Environment = SandboxedEnvironment(undefined=DebugUndefined)
         self.set_context(**kwargs)
 
         # custom filters
-        self._env.filters["where_in"] = WhereInMacro(database.get_dialect())
+        self.env.filters["where_in"] = WhereInMacro(database.get_dialect())
 
     def set_context(self, **kwargs: Any) -> None:
         self._context.update(kwargs)
@@ -496,7 +496,7 @@ class BaseTemplateProcessor:
         >>> process_template(sql)
         "SELECT '2017-01-01T00:00:00'"
         """
-        template = self._env.from_string(sql)
+        template = self.env.from_string(sql)
         kwargs.update(self._context)
 
         context = validate_template_context(self.engine, kwargs)
@@ -643,7 +643,7 @@ class TrinoTemplateProcessor(PrestoTemplateProcessor):
     engine = "trino"
 
     def process_template(self, sql: str, **kwargs: Any) -> str:
-        template = self._env.from_string(sql)
+        template = self.env.from_string(sql)
         kwargs.update(self._context)
 
         # Backwards compatibility if migrating from Presto.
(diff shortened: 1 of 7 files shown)


superset/jinja_context.py
```diff
@@ -479,11 +479,11 @@ def __init__(
         self._applied_filters = applied_filters
         self._removed_filters = removed_filters
         self._context: dict[str, Any] = {}
-        self._env = SandboxedEnvironment(undefined=DebugUndefined)
```
john-bodley, line 482: Given that this PR (and other logic) references `_env` I thought it should be made "public"—in Presto's pseudo public/private naming sense.

superset/sql_parse.py
```diff
@@ -284,19 +286,64 @@ def _extract_tables_from_sql(self) -> set[Table]:
         Extract all table references in a query.
 
         Note: this uses sqlglot, since it's better at catching more edge cases.
+
+        Due to Jinja templating a multiphase approach is necessary as the referenced SQL
```
michael-s-molina, line 290:
```suggestion
        Due to Jinja templating, a multiphase approach is necessary as the referenced SQL
```

superset/sql_parse.py
```diff
@@ -284,19 +286,64 @@ def _extract_tables_from_sql(self) -> set[Table]:
...
+                    table
+                    for statement in parse(sql, dialect=self._dialect)
+                    for table in self._extract_tables_from_statement(statement)
+                    if statement
+                ]
+            )
         except SqlglotError:
-            logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
-            return set()
+            logger.warning("Unable to parse SQL (%s): %s", self._dialect, sql)
```
michael-s-molina, line 344: I think we should throw an exception if a sqlglot or Jinja parsing error occurs instead of returning an empty set. The main reason is because we can have a valid SQL query with no tables, which would also evaluate to an empty set. This is a distinct situation from the inability to determine the existence of tables, and the implications can vary. I think it's the responsibility of the caller to deal with both scenarios.

superset/sql_parse.py
```diff
@@ -284,19 +286,64 @@ def _extract_tables_from_sql(self) -> set[Table]:
...
+
+        tables = set()
+        sql = self.stripped()
+
+        # Mock the required database as the processor signature is exposed publically.
+        processor = get_template_processor(database=Mock(backend=self._dialect))
+        template = processor.env.parse(sql)
+
+        for node in template.find_all(nodes.Call):
+            if isinstance(node.node, nodes.Getattr) and node.node.attr in (
```
michael-s-molina, line 312: Do we need to consider other possible [Jinja macros](https://superset.apache.org/docs/installation/sql-templating/) such as  `{{ url_param('custom_variable') }}`?

michael-s-molina (commented):
Thank you for the PR @john-bodley. I left some first-pass comments.

superset/sql_parse.py
john-bodley: @michael-s-molina this case is being handled in @betodealmeida's https://github.com/apache/superset/pull/26748.

betodealmeida (changes requested):
Hey, @john-bodley, `ParsedQuery` is scheduled to be removed, as per https://github.com/apache/superset/issues/26786. Let's have a call and talk about the best way to move forward here.

betodealmeida:
See the approach I suggested here: https://github.com/apache/superset/pull/27464#issuecomment-1991531350

It does a better job decoupling Jinja rendering from SQL parsing, which is my main concern with this approach, together with the fact that `ParsedQuery` is going to be removed.

superset/sql_parse.py
john-bodley: @michael-s-molina see https://github.com/apache/superset/pull/27464#issuecomment-1992376478.

john-bodley:
Closing in favor of https://github.com/apache/superset/pull/27464.

john-bodley:
@betodealmeida and @michael-s-molina I reopened this PR to address https://github.com/apache/superset/issues/27581 and now should be ready for review.

betodealmeida (approved):
Awesome! Thank you so much for fixing this, John!

Issue #27177, opened 2024-02-20
Error on embedded dashboard after upgrading to 3.1.1: "Guest user cannot modify chart payload"


rscarborough1996:
### Bug description

After upgrading from 3.1.0 to 3.1.1, dashboards and filters no longer work and give this error: "Guest user cannot modify chart payload". I didn't change any permissions, and I even tried changing GUEST_ROLE_NAME to "Admin", but it still didn't work. I can see that the POST request to /api/v1/chart/data is coming back with 403, but I can't tell what is causing it.

### How to reproduce the bug

Using the Docker 3.1.1 image:
1. Create an embedded dashboard app
2. Try to use it

### Superset version

3.1.1

### Python version

I don't know

### Node version

I don't know

### Browser

Chrome

### Additional context

Logs:
SupersetErrorException
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/local/lib/python3.9/site-packages/flask_appbuilder/security/decorators.py", line 95, in wraps
    return f(self, *args, **kwargs)
  File "/app/superset/views/base_api.py", line 127, in wraps
    raise ex
  File "/app/superset/views/base_api.py", line 121, in wraps
    duration, response = time_function(f, self, *args, **kwargs)
  File "/app/superset/utils/core.py", line 1463, in time_function
    response = func(*args, **kwargs)
  File "/app/superset/utils/log.py", line 255, in wrapper
    value = f(*args, **kwargs)
  File "/app/superset/charts/data/api.py", line 235, in data
    command.validate()
  File "/app/superset/commands/chart/data/get_data_command.py", line 68, in validate
    self._query_context.raise_for_access()
  File "/app/superset/common/query_context.py", line 137, in raise_for_access
    self._processor.raise_for_access()
  File "/app/superset/common/query_context_processor.py", line 754, in raise_for_access
    security_manager.raise_for_access(query_context=self._query_context)
  File "/app/superset/security/manager.py", line 1960, in raise_for_access
    raise SupersetSecurityException(
superset.exceptions.SupersetSecurityException: Guest user cannot modify chart payload
2024-02-20 19:57:34,465:WARNING:superset.views.base:SupersetErrorException
[...]

PR #27484, opened 2024-03-12
fix: check if guest user modified query

betodealmeida:
### SUMMARY

Improve the logic that checks if a guest user has tampered with the query context or form data, preventing them from requesting metrics different from the ones saved in the chart.

Before:

Screenshot 2024-03-12 at 15-38-21 React App

After:

Screenshot 2024-03-12 at 15-35-44 React App

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -143,6 +143,56 @@ RoleModelView.edit_columns = ["name", "permissions", "user"]
 RoleModelView.related_views = []
 
 
+def query_context_modified(query_context: "QueryContext") -> bool:
+    """
+    Check if a query context has been modified.
+
+    This is used to ensure guest users don't modify the payload and fetch data
+    different from what was shared with them in dashboards.
+    """
+    form_data = query_context.form_data
+    stored_chart = query_context.slice_
+
+    # sanity checks
+    if form_data is None or stored_chart is None:
+        return True
+
+    # cannot request a different chart
+    if form_data.get("slice_id") != stored_chart.id:
+        return True
+
+    # compare form_data
+    requested_metrics = {
+        frozenset(metric.items()) if isinstance(metric, dict) else metric
+        for metric in form_data.get("metrics") or []
+    }
+    stored_metrics = {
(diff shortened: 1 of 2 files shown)


craig-rueda (approved):
LGTM

michael-s-molina:
@betodealmeida Does this fix https://github.com/apache/superset/issues/27177?

JurajBurian:
Hello @betodealmeida, 
@michael-s-molina, you closed #27177, but problem still exist.  We have embeded dashboards with rls and native filters. After latst commit, embedding is not working. Message is the same :
`superset_app          | superset.exceptions.SupersetSecurityException: Guest user cannot modify chart payload`

I patched manager.py on line 158  : 
```Python
    # sanity checks
    if form_data is None or stored_chart is None:
        return False #replaced from True
```
and it help a lot, but then I hit the wall with another problem :
```Log
superset_app          |   File "/app/superset/security/manager.py", line 178, in <setcomp>
superset_app          |     frozenset(metric.items()) if isinstance(metric, dict) else metric
superset_app          | TypeError: unhashable type: 'dict'
```
unfortunatelly I am not Python expert, but this problem looks as "type" problem. Strange is that exception happens only on tabs having some type of graphs, for example `mixed chart ` is affected , `heat map` or  `bar chart` works fine.

enzo-dechaene:
Hello @michael-s-molina the problem is also on the 3.0

betodealmeida:
```
superset_app          |   File "/app/superset/security/manager.py", line 178, in <setcomp>
superset_app          |     frozenset(metric.items()) if isinstance(metric, dict) else metric
superset_app          | TypeError: unhashable type: 'dict'
```

This error message is really weird, because we're explicitly checking if `metric` is a `dict`.

betodealmeida:
https://github.com/apache/superset/issues/27177 should fix the RLS and mixed chart problems.

PR #27487, opened 2024-03-13
fix(postprocessing): resample with holes

villebro:
### SUMMARY
Currently zero imputation, forward values and backwards values in advanced analytics only works correctly if the time index is completely missing - if the index is present, and there are null values for a specific series, they will not be imputed correctly.

When doing zero imputation for a single series and the time index for the intermediate values is missing, the imputation works as expected:
image

However, with multiple series, it no longer works if the intermediate values are missing in one series for certain indices, but are present for the other. Notice the missing imputation for "boy" series:

### BEFORE

Zero imputation:
image

Forward values:
image

Backward values:
image

### AFTER
Now the zero, backfill and forward fill are imputed correctly.

Zero imputation:
image

Forward values:
image

Backward values:
image

Note, other methods, like linear interpolation were already working as expected.

### TESTING INSTRUCTIONS
1. Create a temporal chart with multiple series, where at least one temporal index is missing for one series
2. Perform zero imputation, forward values or backward values filling
3. Notice it doesn't work.
4. Make the series with the missing indices the only series in the chart, and notice all above mentioned imputation methods work as expected.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/pandas_postprocessing/resample.py
+++ b/superset/utils/pandas_postprocessing/resample.py
@@ -43,13 +43,16 @@ def resample(
         raise InvalidPostProcessingError(_("Resample operation requires DatetimeIndex"))
     if method not in RESAMPLE_METHOD:
         raise InvalidPostProcessingError(
-            _("Resample method should in ") + ", ".join(RESAMPLE_METHOD) + "."
+            _("Resample method should be in ") + ", ".join(RESAMPLE_METHOD) + "."
         )
 
     if method == "asfreq" and fill_value is not None:
         _df = df.resample(rule).asfreq(fill_value=fill_value)
+        _df = _df.fillna(fill_value)
     elif method == "linear":
         _df = df.resample(rule).interpolate()
     else:
         _df = getattr(df.resample(rule), method)()
+        if method in ("ffill", "bfill"):
+            _df = _df.fillna(method=method)
     return _df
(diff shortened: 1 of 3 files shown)


superset/utils/pandas_postprocessing/resample.py
```diff
@@ -43,13 +43,16 @@ def resample(
...
 
     if method == "asfreq" and fill_value is not None:
         _df = df.resample(rule).asfreq(fill_value=fill_value)
+        _df = _df.fillna(fill_value)
     elif method == "linear":
         _df = df.resample(rule).interpolate()
     else:
         _df = getattr(df.resample(rule), method)()
+        if method in ("ffill", "bfill"):
+            _df = _df.fillna(method=method)
```
villebro, line 57: Check the docs for using `fillna` with `method`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.resample.Resampler.fillna.html#pandas.core.resample.Resampler.fillna

zhaoyongjie (approved):
The changes LGTM.

sfirke:
I can confirm that the bug is as you describe in 3.1.1, zero/forward/backward fill all fail to completely fill for a dimension that lacks data points where a different dimension has them.  This looks like a great fix!

sfirke (approved):

PR #27488, opened 2024-03-13
perf(sqllab): reduce bootstrap data delay by queries

justinpark:
### SUMMARY
The SQL Lab bootstrap data, which includes all queries associated with every open editor tab, is quite large in size.
This has been causing a noticeable delay in the initial loading time for SQL Lab. It's important to note that these queries are solely used for maintaining a query history, so this update is designed to bypass these queries in the bootstrap data.
Instead, we will use the Query API to fetch the same results when the Query History component loads, making the process more efficient.

Before:
Superset-2

After:
Screenshot 2024-03-11 at 3 38 15 PM

- Paginated /w infinite scrolling

https://github.com/apache/superset/assets/1392866/36c348e2-1367-4da8-b17c-eef47ef99c7b

### TESTING INSTRUCTIONS
Go to SQL Lab and run some queries in a tab and then switch the query history tab

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/QueryHistory/QueryHistory.test.tsx
+++ b/superset-frontend/src/SqlLab/components/QueryHistory/QueryHistory.test.tsx
@@ -17,7 +17,10 @@
  * under the License.
  */
 import React from 'react';
-import { render, screen } from 'spec/helpers/testing-library';
+import fetchMock from 'fetch-mock';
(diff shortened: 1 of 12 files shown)


michael-s-molina (approved):
LGTM

PR #27501, opened 2024-03-13
feat: chart telemetry

betodealmeida:
### SUMMARY

Some people think Superset is slow. 🙃 They wait on a chart to render, not realizing that Superset is mostly idle waiting on the analytical database to run the query and return the results.

YouTube has a nice feature called "Stats for nerds" when you right click on a [video](https://youtu.be/dQw4w9WgXcQ):

Screenshot 2024-03-13 at 16-49-15 Rick Astley - Never Gonna Give You Up (Official Music Video) - YouTube

We thought of adding a similar option for charts, so that users can see how much time is spent in different steps:

- generating a query
- running the query
- fetching results
- post-processing the results
- converting the results to JSON
- rendering chart

Collecting this data is not easy because the Superset callstack is pretty deep. In order to do this I introduced a new Flask global called `telemetry` that can be used as a context manager:

```python
from flask import g

with g.telemetry.add("Running query"):
    result = run_query(sql)
```

When instrumenting an exposed endpoint with the `show_telemetry` decorator the resulting JSON payload will include the telemetry data, in a format suitable to be rendered as a [flame graph](https://brendangregg.com/flamegraphs.html) or Gantt chart (formatted for readability):

```bash
curl 'http://localhost:8089/api/v1/chart/data?form_data=%7B%22slice_id%22%3A98%7D&dashboard_id=12' -X POST ... | jq
{
  ...
  "telemetry": [
    {
      "name": "Running command",
      "start": 1710362199.5533068,
      "children": [
        {
          "name": "Executing query",
          "start": 1710362200.416116,
          "children": [],
          "end": 1710362210.498135
        },
        {
          "name": "Fetching data from cursor",
          "start": 1710362210.4981928,
          "children": [],
          "end": 1710362210.498541
        },
        {
          "name": "Loading data into dataframe",
          "start": 1710362210.504464,
          "children": [],
          "end": 1710362210.5063
        }
      ],
      "end": 1710362210.512182
    },
    {
      "name": "JSON encoding",
      "start": 1710362210.512204,
      "children": [],
      "end": 1710362210.512265
    }
  ]
}
```

In a next PR I'll add a way to visualize the data in the interface.

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -359,7 +359,7 @@ class ChartDataRestApi(ChartRestApi):
         # This is needed for sending reports based on text charts that do the
         # post-processing of data, eg, the pivot table.
         if result_type == ChartDataResultType.POST_PROCESSED:
-            with g.telemetry("Post processing data"):
+            with g.telemetry.add("Post processing data"):
                 result = apply_post_process(result, form_data, datasource)
 
         if result_format in ChartDataResultFormat.table_like():
@@ -398,7 +398,7 @@ class ChartDataRestApi(ChartRestApi):
             )
 
         if result_format == ChartDataResultFormat.JSON:
-            with g.telemetry("JSON encoding"):
+            with g.telemetry.add("JSON encoding"):
                 response_data = simplejson.dumps(
                     {"result": result["queries"]},
                     default=json_int_dttm_ser,
@@ -418,7 +418,7 @@ class ChartDataRestApi(ChartRestApi):
         datasource: BaseDatasource | Query | None = None,
     ) -> Response:
         try:
-            with g.telemetry("Running command"):
+            with g.telemetry.add("Running command"):
                 result = command.run(force_cached=force_cached)
         except ChartDataCacheLoadError as exc:
             return self.response_422(message=exc.message)
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -53,7 +53,7 @@ from superset.utils.core import (
     get_user_id,
     json_int_dttm_ser,
 )
-from superset.utils.decorators import logs_context
+from superset.utils.decorators import logs_context, show_telemetry
 from superset.views.base import CsvResponse, generate_download_headers, XlsxResponse
 from superset.views.base_api import statsd_metrics
 
@@ -181,6 +181,7 @@ class ChartDataRestApi(ChartRestApi):
 
     @expose("/data", methods=("POST",))
     @protect()
+    @show_telemetry
     @statsd_metrics
     @event_logger.log_this_with_context(
         action=lambda self, *args, **kwargs: f"{self.__class__.__name__}.data",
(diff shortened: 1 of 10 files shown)


superset/charts/data/api.py
```diff
@@ -358,7 +359,8 @@ def _send_chart_response(
         # This is needed for sending reports based on text charts that do the
         # post-processing of data, eg, the pivot table.
         if result_type == ChartDataResultType.POST_PROCESSED:
-            result = apply_post_process(result, form_data, datasource)
+            with g.telemetry("Post processing data"):
```
betodealmeida, line 362: I'm not super happy with the name `telemetry` here and the use of Flask `g` (though we do need to use a per-request singleton for this to work). Open to suggestions.

rafehi:
Just on this, in our particular use case, Superset is the bottleneck, specifically sql_parse and the inefficiency of ParsedQuery. I know you're in the process of switching this out to sqlglot, and in our testing, this will make a big difference. This is mostly due to the long SQL queries that we're using.  Might be worth instrumenting `get_query_str_extended`,as this was a hotspot for us.

rusackas:
It's certainly not the responsibility of this PR, but I did hear a good idea today regarding Presto that could be made Superset-wide. That is: rather than just showing the Superset spinner as a chart loads, it could display two sequential loading animations... one indicating the DB is loading (the "can" icon filling up or something), followed by the Superset spinner once results are loaded. Heck, maybe one day it'd even be a multi-step progress bar. Anyway, nice PR... I'll move along quietly now.

betodealmeida:
> @rafehi: Just on this, in our particular use case, Superset is the bottleneck, specifically sql_parse and the inefficiency of ParsedQuery. I know you're in the process of switching this out to sqlglot, and in our testing, this will make a big difference. This is mostly due to the long SQL queries that we're using. Might be worth instrumenting `get_query_str_extended`,as this was a hotspot for us.

For sure! I instrumented a few of the steps, but later I want to add more (eg, SQL generation) when I'm working on the UI.

john-bodley:
@betodealmeida I love this idea, however I wonder if it could be more implicit, optional, and potentially piggy back off of existing statsD metrics, i.e., adhering to the DRY principle.

Specifically at Airbnb we use [`ddtrace`](https://ddtrace.readthedocs.io/en/stable/) which monkey patches Apache Superset and provides rich telemetry. I'm not sure whether `ddtrace` and integration with Datadog would be seen as a richer alternative or a pattern which we could adopt here.

betodealmeida:
> @betodealmeida I love this idea, however I wonder if it could be more implicit, optional, and potentially piggy back off of existing statsD metrics, i.e., adhering to the DRY principle.

That's a good point, specially because we already have profiling (in the backend), logging, metrics.

Making this optional is relatively easy, but making it implicit is hard, because the stack is very deep and it would log too much information to be useful. Let me see if I can reimplement this using statsd instead.

PR #27505, opened 2024-03-14
chore: 2nd try - simplify python dependencies

mistercrunch:
Keeping some of the good ideas from https://github.com/apache/superset/pull/27465 while putting that one on hold.

What's happening in this PR?
- keep `pip-compile-multi`, but go down to two bundles `base.in` and `development.in`, converging all of `testing.in`, `docker.in`, `integration.in` and `local.in` into `development.in`
- create a github reusable action `.github/actions/setup-backend/action.yml` for python/backend setup


craig-rueda (approved):
LGTM

.github/workflows/superset-cli.yml
```diff
@@ -52,22 +52,8 @@ jobs:
...
-      - name: Install dependencies
-        if: steps.check.outcome == 'failure'
-        uses: ./.github/actions/cached-dependencies
-        with:
-          run: |
-            apt-get-install
-            pip-upgrade
-            pip install wheel
-            pip install -r requirements/testing.txt
-            setup-postgres
```
eschutho, line 70: do we still need this here?

.github/workflows/superset-applitool-cypress.yml
```diff
@@ -58,23 +58,7 @@ jobs:
...
-      - name: Install python dependencies
-        uses: ./.github/actions/cached-dependencies
-        with:
-          run: |
-            pip-upgrade
-            pip install -r requirements/testing.txt
-      - name: Setup postgres
-        uses: ./.github/actions/cached-dependencies
-        with:
-          run: setup-postgres
```
eschutho, line 77: Do we still need `run: setup-postgres` here?

mistercrunch: oh good catch, I assumed we're good since CI is going through, but this is a job on a schedule

.github/workflows/superset-cli.yml
mistercrunch: mmmh, this one ran and succeedd without it. Looking at the code it seems semi-preventative stuff:
```
  sudo apt-get update && sudo apt-get install --yes libecpg-dev
  say "::group::Initialize database"
  psql "postgresql://superset:superset@127.0.0.1:15432/superset" <<-EOF
    DROP SCHEMA IF EXISTS sqllab_test_db CASCADE;
    DROP SCHEMA IF EXISTS admin_database CASCADE;
    CREATE SCHEMA sqllab_test_db;
    CREATE SCHEMA admin_database;
    ```
   I added it back just in case

eschutho (approved):
LGTM. But do we know if this could potentially break someone's deployment if they were relying on the dev dependencies for production? I'm thinking maybe the bigquery or oauth drivers.

mistercrunch:
I added a note in UGPRADING.md, but would love to get @john-bodley 's blessing before merging this one

eschutho:
Great, or rather, even with the notes in UPGRADING.md, would this break semver if this is released in 4.1 for example?

mistercrunch:
I don't think we should treat `requirements/*` as a public interface. Maybe, just maybe, we should treat `requirements/base.txt` as sensitive (renaming/deleting is would likely break some things somewhere). 

I could be great to define the things or type of things we want to treat as a public interface for semver purposes and beyond. I'd argue that pinned deps shouldn't make that list.

Now with the pinned reqs keeping the package together that's a bit different, though that's why I'm trying to get back to a place where the pinned requirements would not be vital.

mistercrunch:
Ok, merging this as I have other improvements lined up

PR #27507, opened 2024-03-14
fix: master docker build is broken

mistercrunch:
Following up on https://github.com/apache/superset/pull/27503 while
flying blind. I'm guessing that DOCKERHUB_USER is empty (?) No way for
me to check. Looking for DOCKERHUB_TOKEN should be enough though.


dpgaspar (approved):

mistercrunch:
Nice!

PR #27522, opened 2024-03-14
feat: support for KQL in `SQLScript`

betodealmeida:
### SUMMARY

This PR is part of [SIP-117](https://github.com/apache/superset/issues/26786):

- [x] Introduce `SQLScript` and `SQLStatement`, and replace any simple calls to `sqlparse` with calls to the new classes (https://github.com/apache/superset/pull/26767).
- [X] Implement a custom `SQLStatement` class for `Kusto`, since [it doesn't use conventional SQL](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/) (this PR).
- [ ] Implement `insert_rls_in_predicate` and `insert_rls_as_subquery` inside `SQLStatement`.
- [ ] Implement `get_cte_query` inside `SQLStatement`.
- [ ] Implement `apply_top_to_sql` in `SQLStatement`.
- [ ] Replace `ParsedQuery` with `SQLScript` and `SQLStatement`.
- [ ] Remove `sqlparse` from Superset.

One of the goals of SIP-117 is to encapsulate all SQL parsing and manipulation logic inside two classes: `SQLScript` and `SQLStatement` (mostly the latter; `SQLScript` is just a lightweight collection of `SQLStatement` objects). This includes logic that currently lives inside the DB engine specs (eg, `parse_sql`, `apply_top_to_sql`, `get_cte_query`, etc.).

Unfortunately, one of the DB engine specs currently supported by Superset is very unusual. The `KustoKqlEngineSpec` engine spec uses a language called KQL that is very different from SQL. In order to move the `parse_sql` method out of the DB engine spec this PR refactors the `SQLStatement` class, introducing a `BaseSQLStatement` base class that is used by `SQLStatement` (for dialects supported by `sqlglot`) and `KustoKQLStatement` (for KQL).

Note that this is not a path forward for supporting non-SQL databases! I only added support for `KustoKqlEngineSpec` because it's already there, but IMHO we should deprecate it in the future. Supporting non-SQL databases adds a lot of complexity to Superset, and is a problem that is better solved orthogonally by writing a SQL adapter.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -19,12 +19,13 @@
 
 from __future__ import annotations
 
+import enum
 import logging
 import re
 import urllib.parse
 from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, cast, Generic, TypeVar
 from unittest.mock import Mock
 
 import sqlglot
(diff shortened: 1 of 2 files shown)


superset/sql_parse.py
```diff
@@ -432,6 +519,192 @@ def get_settings(self) -> dict[str, str]:
...
+
+        """
+        set_regex = r"^set\s+(?P<name>\w+)(?:\s*=\s*(?P<value>\w+))?$"
+        if match := re.match(set_regex, self._parsed, re.IGNORECASE):
+            return {match.group("name"): match.group("value") or True}
+
+        return {}
+
+
+class SQLScript:
```
craig-rueda, line 666: I'm guessing this already has tests?

superset/sql_parse.py
```diff
@@ -432,6 +519,192 @@ def get_settings(self) -> dict[str, str]:
...
+    don't split the query in a semi-colon that's part of a string.
+    """
+
+    OUTSIDE_STRING = enum.auto()
+    INSIDE_SINGLE_QUOTED_STRING = enum.auto()
+    INSIDE_DOUBLE_QUOTED_STRING = enum.auto()
+    INSIDE_MULTILINE_STRING = enum.auto()
+
+
+def split_kql(kql: str) -> list[str]:
```
craig-rueda, line 536: I don't see a test for this

superset/sql_parse.py
betodealmeida: Right, this is an existing class that already has tests, it looks like new because the file changed too much.

superset/sql_parse.py
betodealmeida: It's tested indirectly here: 

https://github.com/apache/superset/pull/27522/files#diff-7e7fc3bcddf707b5ecdbd5b2f0968f957f49fe14fe082dafce51204726940081R1901

But I'll add some explicit tests.

craig-rueda (approved):
LGTM

PR #27523, opened 2024-03-14
fix: Hive integration test

betodealmeida:
### SUMMARY

Skipping the Hive test that is failing with a message that doesn't make sense. It complains about this generated query not being valid due to the `GROUP 1`, even though it is valud:

```sql
SELECT
  `name` AS `name`,
  SUM(num) AS `sum__num`
FROM `birth_names`
WHERE
  `ds` >= CAST('1924-03-14 00:00:00.000000' AS TIMESTAMP)
  AND `ds` < CAST('2024-03-14 21:34:33.000000' AS TIMESTAMP)
  AND `gender` = 'boy'
  AND NOT `num` IS NULL
  AND NOT (
    `name` IS NULL OR `name` IN ('"abc"')
  )
GROUP BY
  1
ORDER BY
  `sum__num` DESC
LIMIT 100;
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/charts/data/api_tests.py
+++ b/tests/integration_tests/charts/data/api_tests.py
@@ -450,6 +450,9 @@ class TestPostChartDataApi(BaseTestChartDataApi):
         """
         Chart data API: Test chart data query with applied time extras
         """
+        if backend() == "hive":
+            return
+
         self.query_context_payload["queries"][0]["applied_time_extras"] = {
             "__time_range": "100 years ago : now",
             "__time_origin": "now",


justinpark:
@betodealmeida would you mind to add some test description and summary?

betodealmeida:
> @betodealmeida would you mind to add some test description and summary?

Hey, this is just a test. I wanted to run the test suite without it stopping on the first error. But for some reason the Hive test didn't run?

geido (approved):

michael-s-molina:
@betodealmeida It looks like we'll [need to skip more tests](https://github.com/apache/superset/actions/runs/8363902025/job/22897962018?pr=27536).

`FAILED tests/integration_tests/charts/data/api_tests.py::TestPostChartDataApi::test_chart_data_async_results_type - AssertionError: 400 != 200`

betodealmeida:
@michael-s-molina huh, you're right... but I'm confused, the tests had passed.

I think the problem is with our Hive, I wonder if it was upgraded because we don't have a pinned version in the Docker image? The query is valid, but for some reason it's no longer understanding that `GROUP BY 1` means `GROUP BY name`. Maybe there's a config flag for enabling integer aliases?

betodealmeida:
Also, the tests passed on https://github.com/apache/superset/pull/27576, which I just rebased.

michael-s-molina:
> Also, the tests passed on #27576, which I just rebased.

It's weird, because the tests are failing for my [PR](https://github.com/apache/superset/pull/27536) which I also rebased. The same error but a different test.

michael-s-molina:
> Also, the tests passed on https://github.com/apache/superset/pull/27576, which I just rebased.

@betodealmeida You were able to merge this PR and https://github.com/apache/superset/pull/27576 only because `test-postgres-hive` resulted in no op.  https://github.com/apache/superset/pull/27536 is failing because it actually touches files that make the tests execute.

PR #27536, opened 2024-03-15
feat: Adds option to disable drill to detail per database

michael-s-molina:
### SUMMARY
This PR adds an option to disable the drill to detail feature per database. This is useful for databases that don't support atomic-level queries or contain large amounts of data not suitable for this type of operation.

During the planning phase of 4.0, the proposal to deprecate the `DRILL_TO_DETAIL` feature flag was rejected for the same reasons this PR is introducing this configuration. The hope is that with this new feature, we can deprecate the feature flag in the next major release. @mattitoo 

https://github.com/apache/superset/assets/70410625/73079873-3ca9-496d-b325-7c5b65562667

### TESTING INSTRUCTIONS
Check that the drill to detail feature can be disabled/enabled for a specific database.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
> Note the drill-to-detail feature doesn't strictly require atomic level data, i.e., it works with Druid where the underlying data is typically rolled-up, however users are only exposed to the (partially) pre-aggregated records.

I'm thinking more about databases that only contain aggregated data but not the atomic data that was used to compute the aggregations. In those cases, drill to detail won't make sense.

michael-s-molina:
> and whether this should be configured at the database or dataset level. I'm not sure if it's typical that non-viable use cases are typically engine specific or dataset specific, though—as you point out—access to the underly records is likely database specific.

I think they are not mutually exclusive. We could have the same setting at the dataset level. I decided to start at the database level because I think the more common scenario is to have databases with no atomic data or large amounts of data. We may also have many virtual datasets, making the configuration harder.

superset-frontend/src/features/databases/DatabaseModal/ExtraOptions.tsx
```diff
@@ -571,6 +571,23 @@ const ExtraOptions = ({
...
+              id="disable_drill_to_detail"
+              indeterminate={false}
+              checked={!!extraJson?.disable_drill_to_detail}
+              onChange={onExtraInputChange}
+              labelText={t('Disable drill to detail')}
+            />
+            <InfoTooltip
+              tooltip={t(
+                `Disables the drill to detail feature for this database. This might be useful if
+                the database does not support atomic-level queries or contains large amounts of data.`,
```
john-bodley, line 586:
```suggestion
                `Disables the drill to detail feature for this database.`,
```

john-bodley: @michael-s-molina I was trying to think about what to say here, i.e., making mention to the fidelity of the data, but couldn't find the right words. I think it's likely really a case by case basis and thus we probably don't need to provide suggestions as to why this might be needed.

john-bodley (approved):

PR #17202, opened 2021-10-22
feat(metrics): Provide override for disabling ad-hoc metrics

john-bodley:
### SUMMARY

There could be some instances where ad-hoc metrics for specific datasets should be disabled, i.e., the dataset owner specifically does not want users to create ad-hoc non-sanctioned metrics. This PR adds—at the dataset level—a custom override for disabling ad-hoc metrics, i.e., removing the `SIMPLE` and `CUSTOM SQL` tabs. The default behavior preserves the existing workflow.

Note if this override is enabled, any existing ad-hoc metric will persist and be deemed editable even though the tab is officially marked as disabled. 

For consistency, for the native Druid connector I also disabled the `CUSTOM SQL` tab (which housed a message as to why this was invalid—this proposed way seems cleaner as well) and added the same treatment for ad-hoc filters.

#### Disallowed

Screen Shot 2021-11-05 at 11 33 31 AM

Screen Shot 2021-11-05 at 11 33 39 AM

Screen Shot 2021-11-05 at 11 41 45 AM

#### Disallowed (pre-existing)

Interestingly the tab is disabled but the canvas remains operational.

Screen Shot 2021-10-28 at 2 58 27 PM

#### Allowed (default)

Screen Shot 2021-10-22 at 11 30 10 AM

### TESTING INSTRUCTIONS

Added unit tests and manually tested for a number of visualization types and ad-hoc metric components: metric, metrics, percentage metric, sort-by metric, etc.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #27539, opened 2024-03-15
fix(explore): Allow only saved metrics and columns

justinpark:
### SUMMARY
Airbnb had previously overridden an option to disable ad-hoc metrics with commit #17202.
However, with the D&D option now enabled, users can accidentally drop an ad-hoc metric, thus gaining access to the previously disabled (tab) option.
To address this, the commit includes a condition to check whether the item being dropped is among the saved metrics, preventing such cases.

Before:

https://github.com/apache/superset/assets/1392866/00519a56-9218-4847-b678-0498f39be6bf

After:

https://github.com/apache/superset/assets/1392866/eee213ce-d3ce-44c4-a608-d2a4cc88b771

### TESTING INSTRUCTIONS
Set a dataset extra with `{ "disallow_adhoc_metrics": true }'
Create a chart with the dataset and then try to drop a column to "Metric" which should be blocked 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.test.tsx
+++ b/superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.test.tsx
@@ -18,7 +18,6 @@
  */
 import React from 'react';
 import thunk from 'redux-thunk';
-import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
 
 import {
@@ -29,7 +28,13 @@ import {
 import { ColumnMeta } from '@superset-ui/chart-controls';
 import { TimeseriesDefaultFormData } from '@superset-ui/plugin-chart-echarts';
 
-import { render, screen } from 'spec/helpers/testing-library';
+import {
+  fireEvent,
+  render,
+  screen,
+  within,
+} from 'spec/helpers/testing-library';
+import type { AsyncAceEditorProps } from 'src/components/AsyncAceEditor';
 import AdhocMetric from 'src/explore/components/controls/MetricControl/AdhocMetric';
 import AdhocFilter from 'src/explore/components/controls/FilterControl/AdhocFilter';
 import {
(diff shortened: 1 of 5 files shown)


superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.tsx
```diff
@@ -384,14 +405,21 @@ const DndFilterSelect = (props: DndFilterSelectProps) => {
...
+      ) {
+        return false;
+      }
+
       if (item.type === DndItemType.Column) {
-        return props.columns.some(
-          col => col.column_name === (item.value as ColumnMeta).column_name,
-        );
+        const columnName = (item.value as ColumnMeta).column_name;
+        return availableColumnSet.has(columnName);
```
justinpark, line 418: @kgabryje I made a change from #27209 to improve the array iteration check by utilizing a hashset, which offers better performance and optimization.

superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.test.tsx
```diff
@@ -153,9 +170,187 @@ test('renders options with adhoc metric', async () => {
...
       useDnd: true,
+      store,
     },
   );
   expect(
     await screen.findByText('Drop columns/metrics here or click'),
   ).toBeInTheDocument();
 });
+
+test('cannot drop a column that is not part of the simple column selection', () => {
```
justinpark, line 181: @kgabryje I also added the spec for your case. Please let me know if you have any issues.

superset-frontend/src/explore/components/controls/DndColumnSelectControl/DndFilterSelect.tsx
kgabryje: Makes sense, thank you!

kgabryje (approved):

PR #27540, opened 2024-03-16
docs: make k8s top item in Installation section

mistercrunch:
Also clarify our stance against using our docker-compose assets in production

tested minikube and provided pointers in the docs


rusackas (approved):
LGTM. My only hesitation was the ordering of things. I almost feel like a (more painful) re-org is needed. Rather than `Install with {Tool}` being the nav options, it feels like we should re-org the content to be `Install for {Development/Production}` and go into the relevant options on those pages. CC @sfirke

PR #27542, opened 2024-03-16
feat(maps): Add Italy regions code to the map generator notebook

iskenderulgen:
### SUMMARY
Italy regions geojson file creation was constructed outside of the map generator jupyter notebook. Up on completion of #27455 pr it is [suggested](https://github.com/apache/superset/pull/27455#issuecomment-1998398195) to add italy regions process to the map generator notebook file. Regions are created from dataframe itself since it provides region - region code information. In 2019 [major changes](https://en.wikipedia.org/wiki/ISO_3166-2:IT) applied on italy region information thus natural earh data directly used to create regions. This is more dynamic than creating custom dictionary as the natural-earth data updated maps will follow (this is only valid for countries that has ISO codes for all granulartiy, countries such as Turkey uses NUTS codes thus it requires custom dictionary). 

Along with italy regions some code improvements also added. When plotting all the maps in earlier stages of notebook, process fails due to missing country-regions. This was fixed and those countries are shown in the center of the mini-plot for corresponding country.  New italy gejson file committed , ".ts" file remained same. 

#### The old italy regions geojson file plot

original_italy_region

#### New italy regions geojson file plot

new italy region

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


iskenderulgen:
Hi rusackas, added italy regions to the country map generator notebook. before merging I wanna do some tests on france regions in case i can use region - region_cod columns to create regions map for france w/o using custom dictionary. This would bring more dynamic approach and will be future proof. Also should we delete italy_regions file inside the countries_custom folder since it is added to the notebook flow.

rusackas:
Thank you for this! I was going to take a look at those early cell failures as well, and at least let them fail gracefully. Thanks for looking into that. I'll pull and review when I'm home after a weekend trip.

iskenderulgen:
hi, did some minor code change on italy region section to be more clear on the dataframe name. Also some code improvements applied on france regions. rather than using fixed region dictionary, now process infers regions codes and names from natural earth data. This will bring dynamic flow in case france region information changes. As the natural earth data updated, region map will be updated too. There was small file size difference but plots shows no majora diff with the old and new region files. 

### Previous france region map plot

original


### New plot after code changes

improved

rusackas:
This looks great, thank you so much!

rusackas (approved):

Issue #27544, opened 2024-03-17
[SIP-121] Proposal for Metrics Comparison base on the Temporal Dimension or/and Textual Dimension


zhaoyongjie:
### Background

The **period-over-period analysis** is a BI technique that compares metrics from a recent period to the same period in the past. We already have this feature in Advanced Analytics and apply it to time-series analytic visualization. This feature utilizes `time_offsets` and `post_processing` in the `query_object` to achieve **period-over-period analysis**. Practically, this feature, along with the entire Advanced Analytics section, should have supported all visualizations because the [legacy Time Section has been removed](https://github.com/apache/superset/pull/21767) and the Time Filter has already moved into Adhoc Filter.

As an experiment, as long as the time dimension is provided and corresponding time filter populate in Adhoc Filter, any charts should easily incorporate this feature. I've opened a toy [pull request](https://github.com/apache/superset/pull/27545) to demonstrate that [only 5 frontend lines](https://github.com/apache/superset/pull/27545/files) (excluding import statements) are needed to integrate this feature into the current Table chart.

![image](https://github.com/apache/superset/assets/2016594/f887be97-5233-4249-83e5-58fbf45f3b63)

### Motivation

Currently, there are some limitations in time_offsets and Advanced Analytics:

1. Theoretically, Period-over-period analysis is a method of analyzing different data slices base on specific dimensions within the same dataset. Therefore, the comparisons should not only be based on temporal dimensions, but also on textual dimensions.

2. According to the first perspective, flexable filters should be provided when users compare data slices, not just time filters.

3. By default, time_offset uses the first occurrence of a time column in `query_object.columns` as the basis for time shifted. It cannot specify which time column to use for generating shifted time range because there can be multiple time filters in current Adhoc Filter.

4. The actual shifted time-range values cannot be displayed on the UI and there isn't an approach to provide it.

Based on the above analysis, I propose to enhance/refactoring the `time_offsets` field to achieve a more generalized `period-over-period analysis`.

### Proposed Changes
To illustrate what needs to change, I will construct a time comparison using the current `time_offsets` field and the `cleaned_sales_data` dataset. This `query_object` will generate a time comparison based on the `order_date`, comparing data slices from `1 week ago`, `1 month ago`, and `1 year ago` :

```json
{
  "columns": [{
    "timeGrain": "P1D",
    "columnType": "BASE_AXIS",
    "sqlExpression": "order_date",
    "label": "order_date",
    "expressionType": "SQL"
  }],
[...]

PR #27545, opened 2024-03-17
[DO NOT MERGE] demonstrate time comparison in Table chart 

zhaoyongjie:
### SUMMARY

Time comparison in Table Viz and apply the AA panel in it.

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
@@ -28,6 +28,10 @@ import {
 } from '@superset-ui/core';
 import { PostProcessingRule } from '@superset-ui/core/src/query/types/PostProcessing';
 import { BuildQuery } from '@superset-ui/core/src/chart/registries/ChartBuildQueryRegistrySingleton';
+import {
+  isTimeComparison,
+  timeCompareOperator,
+} from '@superset-ui/chart-controls';
 import { TableChartFormData } from './types';
 import { updateExternalFormData } from './DataTable/utils/externalAPIs';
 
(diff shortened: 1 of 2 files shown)


eschutho:
Thanks for this example @zhaoyongjie!

eschutho:
@zhaoyongjie since we have the time column in the filter, do you think it's possible to allow users to build a chart without the time column as a dimension as well? Maybe we're doing something wrong, but we weren't able to remove the time dimension from this chart. 
Screenshot 2024-03-19 at 3 45 33 PM

zhaoyongjie:
> since we have the time column in the filter, do you think it's possible to allow users to build a chart without the time column as a dimension as well? Maybe we're doing something wrong, but we weren't able to remove the time dimension from this chart.

@eschutho may be not 100% percent understand why the time dimension should be removed from the chart, --- the filter only affect the where clause but the time dimension should be aggregated from the source table.

Hidden time dimension in the frontend might be more making sense? The custom tab in Table chart is able to set different format by columns, the same logic might add in it, for instance, the "hidden" checkbox should be added in there.

eschutho:
Gotcha. What if we don't want to aggregate/group by the time, though, but rather a different column? Here's an example of what we're trying to get to. It's grouped by name and filtered by date. The time column only serves as a filter for the selected metrics and the time comparison.
Screenshot 2024-03-20 at 4 28 52 PM

Another example from this PR, we want to group by product line, and remove the date. We could remove it from the frontend, but there will still be duplicate rows on the product lines. 
Screenshot 2024-03-20 at 4 33 48 PM

zhaoyongjie:
> What if we don't want to aggregate/group by the time, though, but rather a different column? Here's an example of what we're trying to get to. It's grouped by name and filtered by date. The time column only serves as a filter for the selected metrics and the time comparison.

Hey @eschutho, thank you for clarifying it. I've just realized this use case. In my opinion, it might not be a time comparison but rather a feature that sounds like "metrics comparison based on arbitrary dimensions".

What's interesting is that I've opened a [SIP](https://github.com/apache/superset/issues/27544) to refactor the time_offset field in the Query_Object in order to build such queries.

Could you please take a look at the SIP to see if it aligns with your requirements?

eschutho:
Thanks @zhaoyongjie! Taking a look.

PR #27551, opened 2024-03-18
fix(AlertReports): clearing custom_width when disabled

fisjac:
### SUMMARY

After selecting an custom_width value for a PNG screenshot, the value was stored, even though it was not needed. This would occasionally generate validation on the backend when the unused custom_screenshot value fell outside of the 600-2400px range.

The proposed fix clears the custom_screenshot value from the payload when a non-PNG option has been selected

https://www.loom.com/share/686f1ac272ca46b98854e8b5e0ad2d59?sid=a2b91872-008f-4eb1-8f06-89a653399839
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERT_REPORTS
- [x] Changes UI


john-bodley:
@fisjac I _believe_ there are frontend components where we preserve the view state to aid users when toggling between options. Instead of clearing this on the frontend could we simply just remove it from the payload when the type is not PNG?

geido (approved):

PR #27552, opened 2024-03-18
fix(AlertReports): defaulting grace period to undefined

fisjac:
### SUMMARY

When the grace period field is deleted, and the user attempts to save, the user receives a backend validation error stating "grace period cannot be null."

This fix changes the behavior to reset the grace period to the default value when deleted.

Before:
https://www.loom.com/share/605ed2522a274beaa80f27839915405c?sid=92cb121b-7117-43fa-831f-0162c5f1f50d

After:
https://www.loom.com/share/dbd25315f8f149a4b557a834dbb07d74
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERT_REPORTS
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -885,7 +885,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
 
     // Need to make sure grace period is not lower than TIMEOUT_MIN
     if (value === 0) {
-      updateAlertState(target.name, null);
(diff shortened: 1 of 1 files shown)


villebro (approved):
Thanks for the PR! LGTM and tested to work as expected. It seems slightly unintuitive that the value is zero when the field is empty. Also, it'd be great to have a test, but I'm not sure it's worth the hassle or added overhead for such a small fix. Thoughts @kgabryje ?

kgabryje (approved):

PR #27554, opened 2024-03-18
fix(AlertsReports): making log retention "None" option valid 

fisjac:
### SUMMARY

This PR implements two changes:
1. Default behavior of log retention was to never commit session changes when log_prune runs. This behavior has been changed to ensure that logs are in fact deleted when log_prune runs.
2. When a value of "None" was selected from the dropdown options, the backend would return a validation error "log_retention must be greater than 0." The changes attempt to fix this so that a log retention of 0 days results in all logs being deleted.

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERTS_REPORTS
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -1622,11 +1622,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
                 ariaLabel={t('Log retention')}
                 placeholder={t('Log retention')}
                 onChange={onLogRetentionChange}
-                value={
-                  typeof currentAlert?.log_retention === 'number'
-                    ? currentAlert?.log_retention
-                    : ALERT_REPORTS_DEFAULT_RETENTION
-                }
+                value={currentAlert?.log_retention}
                 options={RETENTION_OPTIONS}
                 sortComparator={propertyComparator('value')}
               />
--- a/superset/commands/report/log_prune.py
+++ b/superset/commands/report/log_prune.py
@@ -48,6 +48,7 @@ class AsyncPruneReportScheduleLogCommand(BaseCommand):
                     row_count = ReportScheduleDAO.bulk_delete_logs(
                         report_schedule, from_date, commit=False
                     )
+                    db.session.commit()
                     logger.info(
                         "Deleted %s logs for report schedule id: %s",
                         str(row_count),
--- a/superset/reports/schemas.py
+++ b/superset/reports/schemas.py
@@ -320,7 +320,7 @@ class ReportSchedulePutSchema(Schema):
     log_retention = fields.Integer(
         metadata={"description": log_retention_description, "example": 90},
         required=False,
-        validate=[Range(min=1, error=_("Value must be greater than 0"))],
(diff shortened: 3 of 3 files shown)


john-bodley:
@fisjac would you mind including a screenshot to aid reviewers in understanding the context of allowing log retention to be "None".

superset/commands/report/log_prune.py
```diff
@@ -46,7 +46,7 @@ def run(self) -> None:
                 )
                 try:
                     row_count = ReportScheduleDAO.bulk_delete_logs(
-                        report_schedule, from_date, commit=False
+                        report_schedule, from_date, commit=True
```
john-bodley, line 49: This is somewhat interesting. With `commit=False` I don't see where this command actually commits. My sense is there should be a `db.session.commit()` on line 60 to ensure we commit once rather than for every report schedule.

BTW I'm working on a PR which will ensure that we handle the "unit of work" correctly.

superset/reports/schemas.py
```diff
@@ -320,7 +320,7 @@ class ReportSchedulePutSchema(Schema):
     log_retention = fields.Integer(
         metadata={"description": log_retention_description, "example": 90},
         required=False,
-        validate=[Range(min=1, error=_("Value must be greater than 0"))],
+        validate=[Range(min=0, error=_("Value must be 0 or greater"))],
```
john-bodley, line 323: What does a log retention of `0` mean and how does this differ from `None`?

superset/commands/report/log_prune.py
john-bodley: @fisjac you marked this as resolved, yet I don't believe the comment was addressed, i.e., the `run` method should commit once—at line 60 per the previous suggestion—rather than on a per report schedule basis.

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -1610,11 +1610,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
                 ariaLabel={t('Log retention')}
                 placeholder={t('Log retention')}
                 onChange={onLogRetentionChange}
-                value={
-                  typeof currentAlert?.log_retention === 'number'
-                    ? currentAlert?.log_retention
-                    : ALERT_REPORTS_DEFAULT_RETENTION
-                }
```
geido, line 1617: Let's double check if this can actually go

superset/reports/schemas.py
fisjac: @john-bodley I had forgotten to push some changes up. This should be covered in the latest push.

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: The default log_retention value is specified as a number, and the dropdown options include only integers.

Additionally, there backend validation schema only allows integers. There shouldn't be a situation which calls for `typeof !== 'number'`

superset/commands/report/log_prune.py
fisjac: @john-bodley I had forgotten to push some changes up. This should be covered in the latest push.

geido (approved):

PR #27558, opened 2024-03-18
fix(explore): drag and drop indicator UX

justinpark:
### SUMMARY
The current explore drag-and-drop (DND) indicator only highlights the drop allow zone, and users can recognize the disallowed zone when they try to drop an item there. However, this UX can be challenging when the current dragging item is not droppable in the entire zone, as there is no highlighting to indicate this.

https://github.com/apache/superset/assets/1392866/68d359bf-e8b7-44ff-bafb-e1b2055a0d48

This lack of feedback can lead users to believe that the DND functionality is not working, until they mouse over the zone.

This commit aims to address this UX issue by implementing changes. Now, all droppable zones will be highlighted when dragging, and a different color scheme will indicate which zones are available and which are unavailable for dropping. 

https://github.com/apache/superset/assets/1392866/0f795241-9eeb-4a53-92c2-00a6a8d3d282

Additionally, the commit also updates the drag over feedback to align with the changes made to the dashboard's DND UX(#26699). This way, users can have a consistent and improved drag-and-drop experience, with clear usability feedback.

Before:

https://github.com/apache/superset/assets/1392866/7b0a92b6-78f3-4aec-b679-023a048f90ab

After:

https://github.com/apache/superset/assets/1392866/6c796977-ae3d-478e-acf7-b4485e6eebe9

### TESTING INSTRUCTIONS
Create a "Table" chart
Select RAW RECORDS and then drag the Metrics to check where is droppable zone

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/OptionControls/index.tsx
+++ b/superset-frontend/src/explore/components/controls/OptionControls/index.tsx
@@ -129,8 +129,8 @@ export const DndLabelsContainer = styled.div<{
     display: ${({ isDragging }) => (isDragging ? 'block' : 'none')};
     background-color: ${({ theme, canDrop }) =>
       canDrop ? theme.colors.primary.base : theme.colors.error.light1};
-    z-index: 10;
-    opacity: 0.2;
+    z-index: ${({ theme }) => theme.zIndex.aboveDashboardCharts};
+    opacity: ${({ theme }) => theme.opacity.light};
     top: 1px;
     right: 1px;
     bottom: 1px;
@@ -139,12 +139,12 @@ export const DndLabelsContainer = styled.div<{
   &:after {
     display: ${({ isOver, canDrop }) => (canDrop && isOver ? 'block' : 'none')};
     background-color: ${({ theme }) => theme.colors.primary.base};
-    z-index: 10;
-    opacity: 0.3;
-    top: -4px;
-    right: -4px;
-    bottom: -4px;
-    left: -4px;
+    z-index: ${({ theme }) => theme.zIndex.dropdown};
+    opacity: ${({ theme }) => theme.opacity.mediumLight};
+    top: ${({ theme }) => -theme.gridUnit}px;
+    right: ${({ theme }) => -theme.gridUnit}px;
+    bottom: ${({ theme }) => -theme.gridUnit}px;
+    left: ${({ theme }) => -theme.gridUnit}px;
   }
 `;
 
--- a/superset-frontend/src/explore/components/controls/OptionControls/index.tsx
+++ b/superset-frontend/src/explore/components/controls/OptionControls/index.tsx
@@ -104,8 +104,17 @@ export const LabelsContainer = styled.div`
 `;
 
 const borderPulse = keyframes`
-  to {
-    background-size: 100% 0px
+  0% {
+    right: 100%;
+  }
+  50% {
+    left: 4px;
+  }
+  90% {
+    right: 4px;
+  }
+  100% {
+    left: 100%;
   }
 `;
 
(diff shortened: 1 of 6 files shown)


justinpark:
cc: @kasiazjc @yousoph for UI feedback

mistercrunch:
Looks great!

superset-frontend/src/explore/components/controls/OptionControls/index.tsx
```diff
@@ -106,18 +106,46 @@ export const LabelsContainer = styled.div`
...
+    content: ' ';
+    position: absolute;
+    border-radius: ${({ theme }) => theme.gridUnit}px;
+  }
+  &:before {
+    display: ${({ isDragging }) => (isDragging ? 'block' : 'none')};
+    background-color: ${({ theme, canDrop }) =>
+      canDrop ? theme.colors.primary.base : theme.colors.error.light1};
+    z-index: 10;
+    opacity: 0.2;
```
rusackas, line 133: If it makes sense for these, the Superset theme has some built-in opacity stops (which we can expand on if needed)

image

superset-frontend/src/explore/components/controls/OptionControls/index.tsx
```diff
@@ -106,18 +106,46 @@ export const LabelsContainer = styled.div`
...
+  &:after {
+    content: ' ';
+    position: absolute;
+    border-radius: ${({ theme }) => theme.gridUnit}px;
+  }
+  &:before {
+    display: ${({ isDragging }) => (isDragging ? 'block' : 'none')};
+    background-color: ${({ theme, canDrop }) =>
+      canDrop ? theme.colors.primary.base : theme.colors.error.light1};
+    z-index: 10;
```
rusackas, line 132: There's a z-index layer for this in the Superset theme... you can uset `theme.zIndex.aboveDashboardCharts`

image

superset-frontend/src/explore/components/controls/OptionControls/index.tsx
```diff
@@ -106,18 +106,46 @@ export const LabelsContainer = styled.div`
...
+    right: 1px;
+    bottom: 1px;
+    left: 1px;
+  }
+  &:after {
+    display: ${({ isOver, canDrop }) => (canDrop && isOver ? 'block' : 'none')};
+    background-color: ${({ theme }) => theme.colors.primary.base};
+    z-index: 10;
+    opacity: 0.3;
+    top: -4px;
```
rusackas, line 144: Same for the z-index/opacity here, but also, if these positions are based on the "drop zone" having a gridUnit of space, you might want to use the `gridUnit` here as well.

rusackas:
This is looking awseome! Left some CSS comments about under-utilized zindex/opacity settings, which we ought to be propagating throughout code whenever we get the chance :D

superset-frontend/src/explore/components/controls/OptionControls/index.tsx
justinpark: Great

rusackas (approved):
Thanks for the touchups! LGTM!

mistercrunch:
wow-icegif-4

john-bodley:
@kasiazjc do you have any inputs with regards to the design change? Note this change is inline with how droppable zones function for dashboards.

superset-frontend/src/explore/components/controls/OptionControls/index.tsx
```diff
@@ -103,21 +103,86 @@ export const LabelsContainer = styled.div`
...
+        bottom: -${theme.gridUnit / 2}px;
+        height: ${theme.gridUnit / 2}px;
+      `}
+  }
+  &:after {
+    display: ${({ isOver, canDrop, isLoading }) =>
+      isLoading || (canDrop && isOver) ? 'block' : 'none'};
+    background-color: ${({ theme, isLoading }) =>
+      isLoading ? theme.colors.grayscale.light3 : theme.colors.primary.base};
+    z-index: ${({ theme }) => theme.zIndex.dropdown};
```
michael-s-molina, line 178: Could you move the variables to the top? Like:

```
const TooltipSectionWrapper = styled.div`
  ${({ theme }) => css`
    font-size: ${theme.typography.sizes.s}px;

    &:not(:last-of-type) {
      margin-bottom: ${theme.gridUnit * 2}px;
    }
  `}
`;
```

to avoid repeating:

```
${({ theme, isDragging, isLoading }) => ... };
```

michael-s-molina (commented):
Thank you for the improvement @justinpark!

michael-s-molina (approved):

PR #27567, opened 2024-03-19
fix(db_engine_specs): Update convert_dttm to work correctly with CrateDB

### LINKED ISSUE #27382: Timerange filter empty result
### Bug description

Importing a public data set or using a dataset on an external DB using  "ts" TIMESTAMP WITH TIME ZONE, works fine but using wizard to manage a simple timerange produce an empty result.

### How to reproduce the bug

Follow the first steps of the guide:
https://preset.io/blog/timeseries-cratedb-superset/
CREATE TABLE IF NOT EXISTS "doc"."power_consumption" (
   "ts" TIMESTAMP WITH TIME ZONE,
   "Global_active_power" REAL,
   "Global_reactive_power" REAL,
   "Voltage" REAL,
   "Global_intensity" REAL,
   "Sub_metering_1" REAL,
   "Sub_metering_2" REAL,
   "Sub_metering_3" REAL,
   "meter_id" TEXT,
   "location" GEO_POINT,
   "city" TEXT
)
CLUSTERED INTO 4 SHARDS

Populated with sample data.

Now set a filter on ts like that:
immagine

Looks like the generated query has a .0 at the ond of timestamps which is not understood by internal superset engine and by external DB.

This is the query generated by SuperSet:

SELECT DATE_TRUNC('day', ts) AS ts,
       sum("Global_active_power") AS "SUM(Global_active_power)"
FROM doc.power_consumption
WHERE ts >= 1172707200000.0
  AND ts < 1206748800000.0
GROUP BY DATE_TRUNC('day', ts)
ORDER BY "SUM(Global_active_power)" DESC
LIMIT 10000

I've tested on external CrateDB and on imported data.
The result is the same.

### Screenshots/recordings

_No response_

### Superset version

3.1.1

### Python version

3.9

### Node version

16

### Browser

Firefox

### Additional context

_No response_

### Checklist

hlcianfagna:
### SUMMARY
[By design](https://cratedb.com/docs/crate/reference/en/5.6/general/ddl/data-types.html#timestamp), when CrateDB casts `BIGINT` values to `TIMESTAMP` it interprets them as being in milliseconds but when it casts `DOUBLE` values it interprets them as being in seconds.
This means that when Superset was trying to filter on a value such as 1679184000000.0 this was interpreted as a date far in the future instead of 2024.

Fixes #27382

### TESTING INSTRUCTIONS

- Spin up a CrateDB instance
- Create a table with a `TIMESTAMP WITH TIME ZONE` column
- Spin up Superset with `docker-compose -f docker-compose-non-dev.yml up`
- Browse http://localhost:8088/ and login with admin / admin
- Create a database connection from Superset to CrateDB using `crate://username:password@hostname:4200/?ssl=true`
- Create a dataset pointing to the table created earlier
- Create a new table chart
- Define dimensions and metrics
- Add a filter on the timestamp column for last week or a custom interval of dates
- Check in CrateDB the query that was run with `SELECT * FROM sys.jobs_log ORDER BY started DESC;`
- Edit `/app/superset/db_engine_specs/crate.py` as in this PR
- Restart the `superset_app` container
- Try again creating the chart with the timestamp filter

### ADDITIONAL INFORMATION

- [X] Has associated issue: https://github.com/apache/superset/issues/27382


villebro (approved):
LGTM, but I think you will need to update the unit test in `tests/unit_tests/db_engine_specs/test_crate.py`.

superset/db_engine_specs/crate.py
```diff
@@ -59,7 +59,7 @@ def convert_dttm(
         sqla_type = cls.get_sqla_column_type(target_type)
 
         if isinstance(sqla_type, types.TIMESTAMP):
-            return f"(({dttm.timestamp() * 1000})::BIGINT)"
+            return f"{dttm.timestamp()}"
```
proddata, line 62: `f"""CAST('{dttm.isoformat()}' AS TIMESTAMP)"""`

is more in line with other solutions:
https://github.com/apache/superset/blob/2944d7b32ac6aaf1aaf33b9591911fdcbae25547/superset/db_engine_specs/postgres.py#L195

villebro: @hlcianfagna do you agree? @proddata would you be able to help out reviewing and testing this PR? I don't use CrateDB right now, so any help validating this would be helpful.

proddata: I think it is debatable, without knowing how Superset exactly uses this method. 

From the description of the [base class](https://github.com/apache/superset/blob/2944d7b32ac6aaf1aaf33b9591911fdcbae25547/superset/db_engine_specs/base.py#L1059-L1070) it seems like this method should convert a Python `datetime` object to the database target data type, which would be `TIMESTAMP`, however CrateDB can implicitly cast from a floating point value to `TIMESTAMP`, which is why the proposed solution of @hlcianfagna most likely also works.

Vitor-Avila:
Thanks for working on this fix @hlcianfagna 🙏

superset/db_engine_specs/crate.py
hlcianfagna: I tested @proddata 's suggestion and it works well, it also looks safer/more elegant so I made a new commit

proddata (approved):

villebro:
@hlcianfagna there's a stubborn flaky test that's blocking merging here that's been disabled since on master branch. Can you rebase the PR? This should help resolve the flaky test..

villebro:
Finally!

PR #27582, opened 2024-03-20
feat(jinja): metric macro

Vitor-Avila:
### SUMMARY

This PR adds a new Jinja macro: ``{{ metric('metric_key', dataset_id) }}``. This macro can be used to retrieve metric syntax from a dataset, to be re-used in charts/SQL Lab/etc.

This is useful to avoid copy/paste and provide a centralized place for metric management. Some use-cases:
* **Overwriting a metric label in the chart level:** Users can use the macro as a custom SQL metric to define a different label in the chart without changing the dataset configuration.
* **Combining metrics:** Users could combine metrics (for example `{{ metric('revenue') }} - {{ metric('cost') }}`.
* **Using metrics in SQL Lab:** Retrieve the metric syntax in SQL Lab for ad-hoc querying.
* **Reusing metrics across datasets:** When creating virtual datasets that are "variations" of a curated physical dataset, this macro can help retrieving metric syntax syntax from the curated dataset into the virtual one.

The `dataset_id` parameter is optional and if not provided Superset will attempt to get it from the current context.

I also noticed there were both `test_jinja_context.py` and `jinja_context_test.py`, so I unified the two.

**After**

[Video attached to Google Drive (Exceeds 10MB)](https://drive.google.com/file/d/14yqTiyUABAXWp2kQGZByGN3nV_89tMjT/view?usp=sharing)

### TESTING INSTRUCTIONS

Both unit and integration tests added. For manual testing:

1. Create a metric in any dataset.
2. Create a chart using this dataset.
3. Define a metric using custom SQL: `{{ metric("metric_key") }}`, replacing `metric_key` with the key used to create the metric in step 1.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [x] Required feature flags: `ENABLE_TEMPLATE_PROCESSING`.
- [ ] Changes UI


superset/jinja_context.py
```diff
@@ -722,3 +723,69 @@ def dataset_macro(
...
+def get_dataset_id_from_context(metric_key: str) -> int:
+    """
+    Retrives the Dataset ID from the request context.
+
+    :param metric_key: the metric key.
+    :returns: the dataset ID.
+    """
+    # pylint: disable=import-outside-toplevel
+    from superset.daos.chart import ChartDAO
+    from superset.views.utils import get_form_data
```
dpgaspar, line 737: do we really need to import these outside of the toplevel? even get_form_data?

superset/jinja_context.py
```diff
@@ -722,3 +723,69 @@ def dataset_macro(
...
+
+    exc_message = _(
+        "Please specify the Dataset ID for the ``%(name)s`` metric in the Jinja macro.",
+        name=metric_key,
+    )
+    if form_data := get_form_data()[0]:
```
dpgaspar, line 743: nit: you could:
``` python
if not form_data, _ := get_form_data():
    raise SupersetTemplateException(exc_message)
...
```
to reduce nesting

btw, would the `slice_id` optionally returned by `get_form_data` help here?

superset/jinja_context.py
Vitor-Avila: @dpgaspar I was following the pattern from other methods in this file -- happy to move them to the top. Going to do that in my next commit.

Vitor-Avila: @dpgaspar I tried moving these local imports to the top but they result in circular import errors (even the `get_form_data`. I was only able to move the test imports to the top.

superset/jinja_context.py
Vitor-Avila: Thanks for the suggestion! Just added a commit taking advantage of the `Slice` returned from `get_form_data()[1]` and also reducing nesting.

superset/jinja_context.py
```diff
@@ -722,3 +723,72 @@ def dataset_macro(
...
+    )
+
+    form_data, chart = get_form_data()
+    if not (form_data or chart):
+        raise SupersetTemplateException(exc_message)
+
+    if chart:
+        return chart.datasource_id
+    if dataset_id := form_data.get("url_params", {}).get("datasource_id"):
+        return dataset_id
```
dpgaspar, line 751: this could be `None`

superset/jinja_context.py
```diff
@@ -722,3 +723,72 @@ def dataset_macro(
...
+    if chart_id := (
+        form_data.get("slice_id") or form_data.get("url_params", {}).get("slice_id")
+    ):
+        chart_data = ChartDAO.find_by_id(chart_id)
```
dpgaspar, line 755: `chart_id` could be `None`

superset/jinja_context.py
Vitor-Avila: that's a good catch! I was thinking that you must associate a dataset to a chart when creating one, but forgot it could be deleted after. Improved that `if chart` to `if chart and chart.datasource_id` to avoid returning `None`.

On the second part tho (with the walrus operator) if `url_params["datasource_id"]` is `None`, this `if` block condition won't be met and therefore it won't return. I added a test case covering this specific scenario.

superset/jinja_context.py
Vitor-Avila: Similarly, the walrus operator condition here shouldn't be met so we wouldn't return `None`. Also added a test case for this scenario.

dpgaspar (approved):

PR #27593, opened 2024-03-20
fix(Dashboard): Add border to row when hovering HoverMenu in edit mode

rtexelm:
### SUMMARY

This adds a distinction to the context of the hover menu icons being a full row of the dashboard editor rather than the left most chart. It's purpose is to avoid confusion as to what these icons would affect.

B:
row icons
A:
new-hover

### TESTING INSTRUCTIONS

Edit a dashboard then hover over the menu icons to the left of any row

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
+++ b/superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
@@ -99,6 +99,7 @@ export default class HoverMenu extends React.PureComponent<HoverMenuProps> {
           )}
           onMouseEnter={this.handleMouseEnter}
           onMouseLeave={this.handleMouseLeave}
+          data-test="hover-menu"
         >
           {children}
         </div>
--- a/superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
+++ b/superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
@@ -25,6 +25,7 @@ interface HoverMenuProps {
   position: 'left' | 'top';
   innerRef: RefObject<HTMLDivElement>;
   children: React.ReactNode;
+  onHover?: (isHovered: boolean) => void;
 }
 
 const HoverStyleOverrides = styled.div`
@@ -71,21 +72,16 @@ export default class HoverMenu extends React.PureComponent<HoverMenuProps> {
     children: null,
   };
 
-  constructor(props: HoverMenuProps) {
-    super(props);
-    // Disabling unused state rule as it is used in Row.jsx by hoverRef
-    // eslint-disable-next-line
-    this.state = {
-      hovered: false,
-    };
-  }
-
   handleMouseEnter = () => {
-    this.setState({ hovered: true });
+    if (this.props.onHover) {
+      this.props.onHover(true);
+    }
   };
 
   handleMouseLeave = () => {
-    this.setState({ hovered: false });
+    if (this.props.onHover) {
+      this.props.onHover(false);
+    }
   };
 
   render() {
--- a/superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
+++ b/superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-unused-state */
 /**
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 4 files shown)


superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
```diff
@@ -81,6 +82,8 @@ export default class HoverMenu extends React.PureComponent<HoverMenuProps> {
             position === 'left' && 'hover-menu--left',
             position === 'top' && 'hover-menu--top',
           )}
+          onMouseEnter={onHover}
+          onMouseLeave={onHover}
```
eschutho, line 86: Technically this component knows the state of the hover. OnMouseEnter = hovered onMouseLeave = not hovered. Instead of having the parent storing and toggling the state (which could get out of sync), I think we can be more explicit by letting this HoverMenu component save the state of the hover, and then there's already a ref getting passed in. We should be able to use that ref to get the state of the hover instead of having the onHover callback.

rtexelm: @eschutho I've tried to add a boolean property of `hoveredState` to `inneRef.current` dependent on the `onMouseEnter/Leave` handlers soley inside `HoverMenu` but any changes are not registering when trying to add a class to the `GridRow` in `Row.jsx`. It might be that the `innerRef` is to `Draggable` so I would need to create some logic in there to handle the `hoveredState`? I think I need a more detailed rundown of what you mean in this case...

john-bodley:
@rtexelm is there merit in in matching the line color/thickness to be the same as when you hover over a specific dashboard element?

rtexelm:
@john-bodley I thought making it 2px instead of 1 would hopefully differentiate it from a regular dashboard element, inspired by the styling from [WithPopoverMenu](https://github.com/apache/superset/blob/master/superset-frontend/src/dashboard/components/menu/WithPopoverMenu.tsx) but using a dashed border to avoid confusion with that context.
It could go further in that direction, changing the border style and color to be more distinct but I don't think it should be closer to the style of a chart or other dashboard element because rows represent a separate role as a container of those

superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
rtexelm: @eschutho Yikes, after an extensive rabbit hole it looks like there isn't really a great way of using a ref to update the state of `Row` based on hovered state of `HoverMenu` without a context provider. I'm thinking a callback is the better choice here but just in case, I made a commit with my attempt at handling things with a ref.

superset-frontend/src/dashboard/components/menu/HoverMenu.tsx
```diff
@@ -70,6 +72,18 @@ export default class HoverMenu extends React.PureComponent<HoverMenuProps> {
...
 
+  handleMouseEnter = () => {
+    if (this.props.onHover) {
+      this.props.onHover(true);
+    }
+  };
+
+  handleMouseLeave = () => {
+    if (this.props.onHover) {
+      this.props.onHover(false);
```
eschutho, line 83: This looks great. Another small nit, I'd suggest being more explicit when defining this onHover function above. When calling it here, there's no context about what this boolean means. If possible, I usually like to pass a dictionary instead of a boolean argument, so `this.props.onHover({isHovered:false})` for example.

rtexelm: Thank you for this suggestion by the way, details like this really help me get better!

rtexelm:
Came across the problem that `getByTestId` was not finding `data-testid` but will find `data-test`. There is an override I found [here](https://testing-library.com/docs/queries/bytestid/) but I couldn't find it utilized in the codebase. I did find this in `babel.config.js`:

```json
production: {
      plugins: [
        [
          'babel-plugin-jsx-remove-data-test-id',
          {
            attributes: 'data-test',
          },
        ],
      ],
    },
```
Does this make sense for the substitution during testing?

eschutho:
@rtexelm Yeah, this all looks good to me. Thanks for the improvements!

eschutho (approved):
Lgtm!

PR #27601, opened 2024-03-21
fix: Persist query params appended to permalink

kgabryje:
### SUMMARY
When adding a query param to an existing permalink, that query param was being removed by the redirect logic.

### TESTING INSTRUCTIONS
1. Create a permalink to a dashboard
2. Paste that permalink to browser URL input and add `?standalone=3` before navigating
3. Verify that the additional query param was persisted and the dashboard opened in standalone mode

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/core.py
+++ b/superset/views/core.py
@@ -843,6 +843,8 @@ class Superset(BaseSupersetView):
         if url_params := state.get("urlParams"):
             params = parse.urlencode(url_params)
             url = f"{url}&{params}"
+        if original_params := request.query_string.decode():
+            url = f"{url}&{original_params}"
         if hash_ := state.get("anchor", state.get("hash")):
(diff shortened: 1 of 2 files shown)


michael-s-molina (approved):

villebro (approved):
LGTM - I suspect adding an IT for that `/dashboard/p/` endpoint is difficult, and would arguably just add to the existing tech debt. But it would be great if we could move this logic out of `/ivews/core.py` at some point.

Edit: never mind the test comment "sleepy eyes* 😆

michael-s-molina:
https://github.com/apache/superset/pull/27605 should fix CI.

PR #27608, opened 2024-03-21
fix(docker): error around missing requirements/base.txt

### LINKED ISSUE #27606: Problem Building dev Docker image
### Bug description

When building the Docker images on macOS, there is an issue when installing the dependencies on the Dev image.
Looks like it might be a result of #27505 

M2 Pro
macOS Sonoma 14.2.1

### How to reproduce the bug

Fresh repo clone
Before building, I pruned Docker's cache.

From the project's root dir
`docker compose build`

Output
```sh
[superset-worker dev 2/2] RUN --mount=type=bind,target=./requirements/development.txt,src=./requirements/development.txt     --mount=type=cache,target=/root/.cache/pip     pip install -r requirements/development.txt:

0.345 ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/base.txt'
```

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

Not applicable

### Node version

Not applicable

### Browser

Not applicable

### Additional context

_No response_

### Checklist

mistercrunch:
### SUMMARY

I'm unclear on why exactly CI succeeded while this was an issue running
locally. I'm guessing it has to do with `--mount` and how it works
around caching, so I'm moving to a COPY-based approach, which is more
standard and easy to reason about. It also leaves a useful artefact on
the machine as to how it was built. It does add a tiny layer and payload
on the image, but it's super cheap for the utility/introspection it provides.

Closes https://github.com/apache/superset/issues/27606

----
I found the issue as to why this defect in the dockerfile wasn't caught in CI, it's because the `--preset dev` parameter was not getting passed down properly, and resulting in targeting the `lean` layer, avoid that issue altogether. I fixed `supersetbot`, added unit tests and published a new version on npm.

### TESTING/VALIDATION

tested `docker-compose up` and validated that CI is now building the right target as opposed to always be building the `lean` target.

### DIFF
--- a/.github/supersetbot/package.json
+++ b/.github/supersetbot/package.json
@@ -1,6 +1,6 @@
 {
   "name": "supersetbot",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "description": "A bot for the Superset GitHub repo",
   "type": "module",
   "main": "src/index.js",
--- a/.github/supersetbot/src/cli.js
+++ b/.github/supersetbot/src/cli.js
@@ -154,16 +154,16 @@ export default function getCLI(context) {
 
     program.command('docker')
       .description('Generates/run docker build commands use in CI')
-      .option('-t, --preset', 'Build preset', /^(lean|dev|dockerize|websocket|py310|ci)$/i, 'lean')
+      .option('-t, --preset <preset>', 'Build preset', /^(lean|dev|dockerize|websocket|py310|ci)$/i, 'lean')
       .option('-c, --context <context>', 'Build context', /^(push|pull_request|release)$/i, 'local')
       .option('-r, --context-ref <ref>', 'Reference to the PR, release, or branch')
       .option('-p, --platform <platform...>', 'Platforms (multiple values allowed)')
       .option('-f, --force-latest', 'Force the "latest" tag on the release')
       .option('-v, --verbose', 'Print more info')
-      .action(function (preset) {
-        const opts = context.processOptions(this);
+      .action(function () {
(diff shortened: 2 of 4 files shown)


rusackas (approved):
I'll admit I had GPT walk me through some of this, but it seems like it all ought to be fine from what I/we can tell.

guizsantos:
I think the issue here was that `requirements/development.txt` references `requirements/base.txt` and since `base.txt` is mounted during runtime, it is not available when the second pip install is running. I believe adding `--mount=type=bind,target=./requirements/base.txt,src=./requirements/base.txt \` in addition to the `development.txt` mount fixes the issue. I liked the mount base approach, much prefer it over the `COPY`-based one.

mistercrunch:
Oh right, totally, that was the issue.

Originally in my PR I was confused as to why CI didn't fail / catch this issue in the first place which I found along the way and commented on the PR body ->

> I found the issue as to why this defect in the dockerfile wasn't caught in CI, it's because the --preset dev parameter was not getting passed down properly, and resulting in targeting the lean layer, avoid that issue altogether. I fixed supersetbot, added unit tests and published a new version on npm.

xyy94813:
Hi, 
I want to confirm if this PR fixes follow bug in CI of my PR??

```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/testing.txt'
```

https://github.com/apache/superset/actions/runs/8235126931/job/22862070545?pr=26180

mistercrunch:
@xyy94813 rebase and you should be fine

PR #27614, opened 2024-03-21
chore: rm references to .github/actions/cached-dependencies

mistercrunch:
### SUMMARY
Doing some spring cleaning here where I noticed this empty folder is
referenced a lot and presumably does nothing. Clearing it.


mistercrunch:
Given up, not worth it

PR #27616, opened 2024-03-22
fix(chart): Sort By (Backend sorting)

nelsondrew:
### **SUMMARY**

Currently , if you use the sort by from chart editor control panel it has no effect , the Sort by control attatches an order by to the sql query which actually sorts the data on sql level but not in chart editor . You can see that it gives proper sorted data in sql lab.
The reason for this is that  the post processing applied on the dataframe causes it to lose the sorting applied by the order by query. 
Screenshot 2024-03-23 at 3 13 24 PM

Screenshot 2024-03-23 at 3 13 52 PM

Solution : What we basically do is we preserve the order of the dataframe with respect to X axis and we store it in a variable and then re use that variable to re order the dataframe again 

Example : 

In the above screenshots we can see that the output data comes in order "No : Not Willing To " , "Yes willing to" , "No answer" in sql lab if sorted correctly , what we can do is that we can preserve this order in an array as [ "No : Not Willing To " , "Yes willing to" , "No answer"] and we will later use this order to reorder the data after postprocessing so that the data gets correct soring order 

Before: 
screen-capture (11)
After: 
screen-capture (12)

With this fix , can apply sorting on any column available in the dataset, for example we apply sorting by Max(Age), a column which is not applied in the present chart , currently the x axis sort by only allows the applied metrics or columns.. with this fix we can use any column in the dataset

screen-capture (14)

Or can write extreme edge cases where the user can write custom sql for ordering data , this is actually the use case in some organisations where the data needs to be sorted in specific order that makes no sense by metric or category name , so currently they make dummy columns and assign them numbers .. this whole process can be avoided if sorting with custom sql can be done :

Screenshot 2024-03-22 at 6 23 20 PM (2)

Screenshot 2024-03-22 at 6 23 37 PM (1)

### TESTING INSTRUCTIONS
Try using Sort By on a chart on Superset 4.0 by any metric and check if the data gets sorted or not, if necessary compare it with sql lab results as it shows the correct order

Then try using sort by ,in chart editor in this fix & try sorting by any metric and compare it with the sql lab results

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/query/types/QueryResponse.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/types/QueryResponse.ts
@@ -39,10 +39,6 @@ export interface DataRecord {
   [key: string]: DataRecordValue;
 }
 
-export interface InitialOrder {
-  sort_order: string[];
-}
-
 /**
  * Queried data for charts. The `queries` field from `POST /chart/data`.
  * See superset/charts/schemas.py for the class of the same name.
@@ -82,11 +78,6 @@ export interface ChartDataResponseResult {
     | 'timed_out';
   from_dttm: number | null;
   to_dttm: number | null;
-  /**
-   * Initial Order of the pandas dataframe for query which is used to re order the data
-   * when time series limit metric is present
-   */
-  initial_order: InitialOrder;
 }
 
 export interface TimeseriesChartDataResponseResult
--- a/superset-frontend/packages/superset-ui-core/src/query/types/QueryResponse.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/types/QueryResponse.ts
@@ -39,6 +39,10 @@ export interface DataRecord {
   [key: string]: DataRecordValue;
 }
 
+export interface InitialOrder {
+  sort_order: string[];
+}
+
 /**
  * Queried data for charts. The `queries` field from `POST /chart/data`.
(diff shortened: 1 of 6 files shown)


john-bodley:
Thanks @nelsondrew for the PR. I've kicked off the CI checks but could you please complete the PR description? This helps to provide the necessary context to the reviewers.

zhaoyongjie:
@nelsondrew please add more detailed description for the fix. We should care about which sort do you want to change. e.g. server side order by or Pandas stage sort by, which viz face this issue. 

Thanks!

rusackas:
I'll pile on that in addition to clarity, it might be good to include a test that passes with your PR and fails without it... that can help explain what the code does, and make sure whatever's broken doesn't break again one day :D

nelsondrew:
Added all the necessary details , screenshots and Changes , Please Review once
@john-bodley  @zhaoyongjie  @rusackas  @sfirke

yousoph:
Hi @nelsondrew , thanks for the PR! Would you be able to add a test for this change? Thanks!

geido:
Hey @nelsondrew curious if you would like to bring this to the finish lines. Also, pinging @villebro as I'd like your opinion on this implementation. Thanks both!

PR #27625, opened 2024-03-22
perf(explore): virtualized datasource field sections

justinpark:
### SUMMARY
At Airbnb, we have a dataset that contains > 10k columns and metrics.
Given the size of this dataset, the current rendering of the show all datasource target list is not realistic.

To address this issue, this commit introduces virtualized rendering for the field sections, which significantly improves performance and supports handling large-sized metadata datasets.

Before:

https://github.com/apache/superset/assets/1392866/4402d60f-ef57-45d7-aaa6-aacae28507d7

After:

https://github.com/apache/superset/assets/1392866/bd463655-0ff0-433a-972a-2d7b6c817c34

### TESTING INSTRUCTIONS
Go to "create a chart" and then 
- test drag and drop 
- test show/hide all
- test collapse/expand header

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/DatasourcePanel/DatasourcePanelItem.tsx
+++ b/superset-frontend/src/explore/components/DatasourcePanel/DatasourcePanelItem.tsx
@@ -44,13 +44,13 @@ type Props = {
     totalColumns: number;
     width: number;
     showAllMetrics: boolean;
-    setShowAllMetrics: (showAll: boolean) => void;
+    onShowAllMetricsChange: (showAll: boolean) => void;
     showAllColumns: boolean;
-    setShowAllColumns: (showAll: boolean) => void;
+    onShowAllColumnsChange: (showAll: boolean) => void;
     collapseMetrics: boolean;
-    setCollapseMetrics: (collapse: boolean) => void;
+    onCollapseMetricsChange: (collapse: boolean) => void;
     collapseColumns: boolean;
-    setCollapseColumns: (collapse: boolean) => void;
+    onCollapseColumnsChange: (collapse: boolean) => void;
   };
 };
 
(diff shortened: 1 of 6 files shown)


superset-frontend/src/explore/components/DatasourcePanel/DatasourcePanelItem.tsx
```diff
@@ -0,0 +1,230 @@
...
+  index: number;
+  style: CSSProperties;
+  data: {
+    metricSlice: Metric[];
+    columnSlice: DataSourcePanelColumn[];
+    totalMetrics: number;
+    totalColumns: number;
+    width: number;
+    showAllMetrics: boolean;
+    setShowAllMetrics: (showAll: boolean) => void;
```
michael-s-molina, line 47: I think from the component's viewpoint `setShowAllMetrics` should be named `showAllMetricsChanged` or `onShowAllMetrics`. I know that the parent component will set a state with this value but the component's API shouldn't reflect that. This component can be used elsewhere and the event can be handled differently.

superset-frontend/src/explore/components/DatasourcePanel/DatasourcePanelItem.tsx
```diff
@@ -0,0 +1,230 @@
...
+`;
+
+const SectionHeader = styled.span`
+  ${({ theme }) => `
+    font-size: ${theme.typography.sizes.m}px;
+    line-height: 1.3;
+  `}
+`;
+
+const DatasourcePanelItem: React.FC<Props> = ({ index, style, data }) => {
```
michael-s-molina, line 133: Nice work extracting the component into its own file and adding tests 👏🏼

michael-s-molina (commented):
Thank you for the improvement @justinpark!

michael-s-molina (approved):
LGTM. I left a non-blocking comment about nomenclature.

superset-frontend/src/explore/components/DatasourcePanel/DatasourcePanelItem.tsx
justinpark: Done!

Issue #20300, opened 2022-06-07
[SIP-85] OAuth2 for databases


betodealmeida:
### Motivation

Modern cloud databases (eg, Sowflake, BigQuery, and Google Sheets) support OAuth2 for user authentication and authorization. Connecting Superset to these databases today, specially Google Sheets, is a complicated multi-step process, requiring uses to create a service account that can potentially give access to all the sheets inside an organization (if "impersonate user" is not enabled).

Superset could offer a much better experience to users by supporting OAuth2 directly. Users would simply have to grant the necessary permissions to Superset, and Superset would store and use the access tokens when running queries. This would remove the complexity of setting up these databases, and solve the problem of risking giving access to confidential tables and/or sheets.

### Proposed Change

We propose supporting OAuth2 directly in Superset. When accessing a resource (table or Google sheet) that requires OAuth2 the user would be redirected to the authorization server, where they can grant the necessary permissions. After OAuth2 the access and refresh tokens would be stored in table with grain of `(database, user)`, and reused in subsequent connections.

The following video shows the process in action:

https://user-images.githubusercontent.com/1534870/172488708-06f99cf3-499f-4323-91cd-fa426259e06c.mov

### New or Changed Public Interfaces

This proposal adds a new table and a new endpoint. It also modifies `modify_url_for_impersonation` to pass a new parameter, the OAuth2 `access_token`. The SIP will be first implemented for Google Sheets, but the implementation is database agnostic, with all database specific logic living in the engine spec.

To store the user tokens a new table `database_user_oauth2_tokens` would be created, with a corresponding SQLAlchemy model:

| user_id | database_id | access_token | access_token_expiration | refresh_token |
| --- | --- | --- | --- | --- |
| 1 | 1 | XXX | 2022-06-06 09:00:00Z | YYY |

(The access token and refresh token are stored in encrypted columns, similar to other database credentials.)

Here the user with ID 1 has an access token `XXX` on the database with ID 1. When they run a query in that database Superset will pass the access token to the database engine spec via `modify_url_for_impersonation`. For Google Sheets to use this token it's as simple as:

```python
class GSheetsEngineSpec(SqliteEngineSpec):
    ...
    @classmethod
    def modify_url_for_impersonation(
        cls,
        url: URL,
        impersonate_user: bool,
        username: Optional[str],
        access_token: Optional[str] = None,  # added in this SIP
    ) -> None:
        if impersonate_user and access_token:
            url.query["access_token"] = access_token
```

[...]

Issue #9298, opened 2020-03-12
[SIP-41] Proposal for Alignment of Backend Error Handling


john-bodley:
### Motivation

[[SIP-40] Proposal for Custom Error Messages](https://github.com/apache/incubator-superset/issues/9194) proposes a unified pattern for handling API errors on the frontend however currently it is not apparent what error types (surfacing from the backend) need to be supported. 

Reading through the code it is not super apparent:

- What exceptions are raised.
- How errors are handled and by whom, i.e., via the Flask or FAB error-handler, re-wrapped, etc. 

Superset defines a [number](https://github.com/apache/incubator-superset/blob/master/superset/exceptions.py) of exceptions each deriving from the `SupersetException` base class which contains an HTTP response status code in the [400, 600) range. Additionally there are a [number](https://github.com/apache/incubator-superset/blob/52c59d689063bfac9cba73b416d2b1bdc6a1275e/superset/datasets/commands/exceptions.py) of exceptions related to dataset commands.

#### Raised Exceptions

It can be quite difficult to understand exactly where and when an exception will be handled. The Python community as a whole does a fairly poor job of documenting (sadly I sense a shortage of typing was not adding support for exceptions) what exception types a method can raise and thus  from a development perspective tracing the exception flow can be challenging. 

#### Error Handling

A common approach in Flask for [handling errors](https://flask.palletsprojects.com/en/1.1.x/errorhandling/#error-handlers) is registering error handlers which can defined at either the application or blueprint level. This pattern is also supported by a number of the Flask RESTful extensions. Currently there only exists [one](https://github.com/apache/incubator-superset/blob/406ad8778c4b83461e534b03d165c279484b2d2c/superset/views/core.py#L2628) registered error handler in Superset and [one](https://github.com/dpgaspar/Flask-AppBuilder/blob/cfb324007ca12c6d5939e05521880ca7ba888236/examples/simpleform/app/views.py#L36) in FAB.

Even though Superset has a number of defined exceptions there is no consistent way in how these are propagated/handled and thus it is hard to do an audit on which errors exists. For example:

- [Here](https://github.com/apache/incubator-superset/blob/406ad8778c4b83461e534b03d165c279484b2d2c/superset/views/core.py#L325) is an example of returning a Flask Response with an error status. It seems this pattern is mostly used for non-API calls.
[...]

Issue #9194, opened 2020-02-24
[SIP-40] Proposal for Custom Error Messages


etr2460:
### Motivation

The current user experience for error cases in Superset does not provide actionable feedback to users. Error messages often are displayed straight from the backend and are fairly inscrutable. Additionally, there’s no way to customize or add other functionality to these messages on a deployment by deployment basis. A key part of the user experience is providing friendly, actionable messaging when something goes wrong.

### Proposed Change

We propose a new, unified pattern for error handling in API endpoints across all of Superset. This will consist of the following work:
1. Unify the backend around a standard error payload for all API errors returned from Superset of the format:
```
{
  errors: [
    {
      message: string,
      error_type: Enum,
      level: ‘info’ | ‘warning’ | ‘error’,
      extra: Dict[string, Any],
    }, ...
  ]
}
```
2. Create an Enum on both the server and client side for uniquely identifying error types. The specific type of an error is included in the standard error payload (e.g. ACCESS_DENIED, DB_ENGINE_SYNTAX_ERROR, etc.)
3. Refactor Superset’s frontend to go through a single `SupersetAlert` component for handling all error messages. This component will import a renderer file that defaults to some generic error customization messages (ex. Adding Request Access links).
4. Allow a Superset admin to specify an override renderer file that gets used instead of the default renderer. This override renderer is injected by leveraging the Registry class from @superset-ui/core to handle the custom renderers registration.

### New or Changed Public Interfaces

- All APIs within Superset will return the above error payload on 4xx and 5xx status code responses.
- Client side options will be made public for injecting your own error rendering file

### New dependencies

N/A 
### Migration Plan and Compatibility

No migrations should be necessary, but the new API may not be backwards compatible. Any changes to the current API will be notated in UPDATING.md and should take place prior to v1.0.0.

### Rejected Alternatives

Server side rendering of errors was rejected as we’re trying to remove the existing templates and move all rendering to the client side

cc: @kristw @rusackas @nytai @graceguo-supercat @ktmud

kristw:
@etr2460 The error code and `SupersetAlert` component seems useful for usage outside of Superset app as well. Do you want to implement the front-end error handling as a new package `@superset-ui/superset-error` to provide typings and components? 

> Allow a Superset admin to specify an override renderer file that gets used instead of the default renderer. This override renderer is injected with a Webpack Plugin.

[...]

PR #27631, opened 2024-03-23
feat(SIP-85): OAuth2 for databases

betodealmeida:
### SUMMARY

This PR introduces a new table called `database_user_oauth2_tokens`. The table is used for storing personal user tokens associated with a given database:

```sql
# \d database_user_oauth2_tokens
                                               Table "public.database_user_oauth2_tokens"
         Column          |            Type             | Collation | Nullable |                         Default
-------------------------+-----------------------------+-----------+----------+---------------------------------------------------------
 created_on              | timestamp without time zone |           |          |
 changed_on              | timestamp without time zone |           |          |
 id                      | integer                     |           | not null | nextval('database_user_oauth2_tokens_id_seq'::regclass)
 user_id                 | integer                     |           | not null |
 database_id             | integer                     |           | not null |
 access_token            | bytea                       |           |          |
 access_token_expiration | timestamp without time zone |           |          |
 refresh_token           | bytea                       |           |          |
 created_by_fk           | integer                     |           |          |
 changed_by_fk           | integer                     |           |          |
Indexes:
    "database_user_oauth2_tokens_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "database_user_oauth2_tokens_changed_by_fk_fkey" FOREIGN KEY (changed_by_fk) REFERENCES ab_user(id)
    "database_user_oauth2_tokens_created_by_fk_fkey" FOREIGN KEY (created_by_fk) REFERENCES ab_user(id)
    "database_user_oauth2_tokens_database_id_fkey" FOREIGN KEY (database_id) REFERENCES dbs(id)
    "database_user_oauth2_tokens_user_id_fkey" FOREIGN KEY (user_id) REFERENCES ab_user(id)
```

Whenever a SQLAlchemy engine is instantiated, the personal user token (or `None`) will be passed to the `get_url_for_impersonation` method in the DB engine spec, so that a custom URL can be built for the user. For example, for GSheets:

```python
def get_url_for_impersonation(
    cls,
    url: URL,
    impersonate_user: bool,
    username: str | None,
    access_token: str | None,  # <== here
) -> URL:
    if not impersonate_user:
        return url

    if username is not None:
        user = security_manager.find_user(username=username)
        if user and user.email:
            url = url.update_query_dict({"subject": user.email})

    if access_token:
        url = url.update_query_dict({"access_token": access_token})

    return url
```

The change allows users to login to databases like BigQuery, Snowflake, Dremio, Databricks, Google Sheets, etc. using their own credentials. This makes it easier to set up databases, since service accounts are no longer required, and provides better isolation of data between users. Only support for Google Sheets is implemented in this PR, and it's considered the reference implementation. Note that a newer version of Shillelagh is required, since a change in the Google Auth API introduced a [regression](https://github.com/betodealmeida/shillelagh/pull/438).

In order to populate the table with personal access tokens, the DB engine spec checks for a specific exception that signals that OAuth2 should start:

```python
class BaseEngineSpec:
    def execute(...) -> None:
        try:
            cursor.execute(query)
        except cls.oauth2_exception as ex:  # <== here
            if cls.is_oauth2_enabled() and g.user:
                cls.start_oauth2_dance(database_id)
            raise cls.get_dbapi_mapped_exception(ex) from ex
        except Exception as ex:
            raise cls.get_dbapi_mapped_exception(ex) from ex
```

When called, the `start_oauth2_dance` method will return the error `OAUTH2_REDIRECT` to the frontend. The error is captured by the `ErrorMessageWithStackTrace` component, which provides a link to the user so they can start the OAuth2 authentication. Since this is implemented at the DB engine spec level, any query will trigger it — in SQL Lab, Explore, or dashboards — see the screenshots below for the UX.

Note that while the current implementation triggers OAuth2 when a query needs authorization, we could also implement affordances in the database UI to manually trigger OAuth2 to store the personal access tokens. This could be done in the future.

SQL Lab. Note that the query runs automatically once OAuth is completed:

https://github.com/apache/superset/assets/1534870/1c328d8e-e2ad-48d9-8c82-b04279d93dad

Explore. Note that the chart is automatically updated after OAuth:

https://github.com/apache/superset/assets/1534870/e9b787f0-34fd-40c1-b700-0c6f49b5fdd9

Same thing for dashboards:

https://github.com/apache/superset/assets/1534870/8ab5dd51-9bea-437f-b22d-f68364b72308

### TESTING INSTRUCTIONS

1. Create a Google Sheets database.
2. Create a Google OAuth2 application at https://console.cloud.google.com/apis/credentials/oauthclient/ of type "Web application"
3. Edit `superset_config.py` and add the client ID and secret:

```python
DATABASE_OAUTH2_CREDENTIALS = {
    "Google Sheets": {
       "CLIENT_ID": "XXX.apps.googleusercontent.com",
       "CLIENT_SECRET": "GOCSPX-YYY",
    },
}
```

4. In SQL Lab, try to query a sheet that is not shared publicly. It should trigger OAuth2.
5. Add the sheet as a dataset and create a chart.
6. Delete the tokens from the database:

```sql
DELETE FROM database_user_oauth2_tokens;
```

7. Reload the chart. It should trigger OAuth2.
8. Add the chart to a dashboard, delete the tokens, and reload the dashboard. It should trigger OAuth2.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/20300
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2024-03-20_16-02_678eefb4ab44_add_access_token_table.py
```diff
@@ -0,0 +1,76 @@
...
+        ),
+        sa.ForeignKeyConstraint(
+            ["database_id"],
+            ["dbs.id"],
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["ab_user.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
```
mistercrunch, line 71: It would be great to add an index (btree) on `(user_id, database_id)` as we'll be accessing this much like a key value store so it should be fast!

superset/exceptions.py
```diff
@@ -312,3 +312,47 @@ def __init__(self, sql: str, engine: Optional[str] = None):
             extra={"sql": sql, "engine": engine},
         )
         super().__init__(error)
+
+
+class OAuth2RedirectError(SupersetErrorException):
```
mistercrunch, line 317: NIT(optional): OAuth2RedirectError could inherit from `OAuth2Error`

superset/utils/oauth2.py
```diff
@@ -0,0 +1,69 @@
...
+
+from __future__ import annotations
+
+from datetime import datetime, timedelta
+
+from superset import db
+from superset.db_engine_specs.base import BaseEngineSpec
+
+
+def get_oauth2_access_token(
```
mistercrunch, line 26: Optional: I think in theory a lot of the methods around this could be defined as `async`, which I don't think we're using much around the codebase. Eventually if/when we work on performance improvements for the backend we could look into whether having more async functions defined could help improve perf. Might be great to be able to keep doing other work while we oauth handshake async.

superset/db_engine_specs/hive.py
```diff
@@ -547,7 +551,10 @@ def update_impersonation_config(
 
     @staticmethod
     def execute(  # type: ignore
-        cursor, query: str, async_: bool = False
+        cursor,
+        query: str,
+        database_id: int,
```
mistercrunch, line 556: NIT: New field could have been added as last arg, but `db_engine_specs` isn't a public API so it doesn't matter. I think in any case we have to touchup all the classes that are overriding this method (like you did). I noticed that some overrides for this method define `**kwargs` and really don't think we should have different headers in derived classes, but that's probably outside the scope of this PR.

mistercrunch (commented):
Overall seems solid. My most important point is probably around adding a database index on the new model, the rest are comments/notes.

john-bodley:
@betodealmeida do we perceive there could/would be other authorization frameworks other than OAuth 2.0? If so I was wondering if there was merit in renaming `database_user_oauth2_tokens` to be something more generic and adding additional column(s) which define said frameworks.

craig-rueda (commented):
Looks good so far! I think there's quite a few cases that could be tested. (Token delete/insert/ etc., the updated engine spec definitions, etc...)

betodealmeida:
> @betodealmeida do we perceive there could/would be other authorization frameworks other than OAuth 2.0? If so I was wondering if there was merit in renaming `database_user_oauth2_tokens` to be something more generic and adding additional column(s) which define said frameworks.

@john-bodley I'm not sure, to be honest. The beauty of OAuth2 is that the same flow is shared across multiple providers, all you need is an access token and a refresh token, so the same foundation works for BigQuery/GSheets/Snowflake/Dremio/Databricks.

The one case I can think of is if at some point we'd want users to be able to input their own username/password, but I think that trying to address potential future uses would increase the complexity without clear benefits.

superset/exceptions.py
betodealmeida: Ah, good idea!

superset/connectors/sqla/models.py
```diff
@@ -1744,6 +1746,11 @@ def assign_column_label(df: pd.DataFrame) -> pd.DataFrame | None:
 
         try:
             df = self.database.get_df(sql, self.schema, mutator=assign_column_label)
+        except (SupersetErrorException, SupersetErrorsException) as ex:
```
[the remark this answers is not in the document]
betodealmeida: Yeah, we need to improve the exception handling. We never agreed on how to implement [SIP-41](https://github.com/apache/superset/issues/9298), and the code is a mix of exception bubbling and explicit error responses.

For context, today if you let an exception of type `SupersetErrorException` or `SupersetErrorsException` bubble up Flask will format it nicely in a [SIP-40](https://github.com/apache/superset/issues/9194) compliant error payload, with codes that are shared between the backend and frontend. It greatly simplifies the flow in this PR.

betodealmeida:
> Looks good so far! I think there's quite a few cases that could be tested. (Token delete/insert/ etc., the updated engine spec definitions, etc...)

@craig-rueda I did test inserting/deleting the token in the API test. I'll add tests for the new DB engine spec methods.

superset/utils/oauth2.py
```diff
@@ -0,0 +1,69 @@
...
+        token_response = db_engine_spec.get_oauth2_fresh_token(token.refresh_token)
+
+        # store new access token; note that the refresh token might be revoked, in which
+        # case there would be no access token in the response
+        if "access_token" in token_response:
+            token.access_token = token_response["access_token"]
+            token.access_token_expiration = datetime.now() + timedelta(
+                seconds=token_response["expires_in"]
+            )
+            db.session.add(token)
```
villebro, line 62: If I'm reading this correctly, this logic is being handled by the webworker, and will cause multiple token refreshes to trigger if multiple queries are executed at the same time. I touch on how we solved this for our own OAuth solution here, where a distributed lock ensures that only a single worker executes the refresh, and other workers wait until that's completed: https://github.com/apache/superset/issues/20300#issuecomment-2018638419

betodealmeida: Oh, great point, @villebro! But I don't think we have a way of creating a global lock in a default install of Superset, no? Since Redis is not mandatory. We could use the metadata DB for that, but that brings some additional complication (we need to manage timeouts ourselves).

How did you implement it?

superset/db_engine_specs/hive.py
betodealmeida: I had to add here because only Hive has the `async_` parameter, which gets unpacked from `**kwargs` in the base class.

superset/utils/oauth2.py
villebro: @betodealmeida I implemented it as a context manager that stores the lock in the key-value table, i.e. in the metadata db. This makes it globally accessible from all workers with minimal overhead. Here's a snippet for how it happens in the code flow ("profile" here denotes the internal OAuth app id):

```python
    try:
        raise_for_invalid_token(tokens=tokens, verify=False)
        return tokens
    except AuthTokenValidationException:
        pass

    try:
        logger.debug(f"Tokens on profile `{profile}` have expired, refreshing")
        with AuthLock(username=username, profile=profile) as lock_id:
            tokens = _do_token_refresh(
                username=username,
                tokens=tokens,
                profile=profile,
            )

            try:
                raise_for_invalid_token(tokens=tokens)
            except TokenValidationException as ex:
                raise TokenActionException(
                    f"Refreshed tokens for profile `{profile}` and user `{username}` are invalid: {ex.message}",
                ) from ex

            logger.debug(
                "Updating tokens on profile `superset` for user `%s` (lock id: %s)",
                username,
                lock_id,
            )
            expiration_dttm = datetime.now() + timedelta(days=TOKEN_EXPIRATION_DAYS)
            UpdateKeyValueCommand(
                resource=JWT_RESOURCE,
                key=get_token_key(
                    resource=JWT_RESOURCE, username=username, profile=profile
                ),
                codec=AuthKeyValueCodec(username),
                value=tokens,
                expires_on=expiration_dttm,
            ).run()
            return tokens
    except CreateAuthLockFailedException:
        logger.debug(
            "Auth lock taken for profile `%s` and user `%s`, retrying fetching of superset tokens",
            profile,
            username,
        )
        raise AuthBackoffException()
```

(I did some redacting here, so there may be typos)

That `AuthLock` tries to create an entry in the key-value table under a specific resource with a specific key, and if it's already there, it raises an exception that we reraise to an exception that backoff can handle. Also, the lock has an expiry, meaning that if anything bad happens, it'll get removed after a short while (we seem to have 20 seconds, which should be ample time to do a refresh round trip).

betodealmeida: That looks awesome, I'll use it here to refresh the token. Thanks, @villebro!

villebro: Sounds good - I DM'ed the context manager via Slack so you don't have to reinvent the wheel

superset/utils/oauth2.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    pair_str = f"{a}-{b}"
+    return uuid.uuid5(uuid.NAMESPACE_DNS, pair_str)
+
+
+@contextmanager
+def AuthLock(  # pylint: disable=invalid-name
+    user_id: int,
+    database_id: int,
+) -> Iterator[None]:
```
villebro, line 96: One thing to consider is making the lock a generic construct, so that it can be utilized wherever locks are needed. It could look something like this (just spitballing, the contract could look different):

```python
@contextmanager
def KeyValueDistributedLock(
    namespace: str,
    params: dict[str, any],
) -> Iterator[None]:
```

This way it could be used for other use cases without having to create new dedicated context managers.

superset/db_engine_specs/gsheets.py
```diff
@@ -136,6 +153,89 @@ def extra_table_metadata(
...
+        config = current_app.config["DATABASE_OAUTH2_CREDENTIALS"]["Google Sheets"]
+        redirect_uri = current_app.config.get(
+            "DATABASE_OAUTH2_REDIRECT_URI",
+            state["default_redirect_uri"],
+        )
+
+        encoded_state = jwt.encode(
+            payload=state,
+            key=current_app.config["SECRET_KEY"],
+            algorithm=current_app.config["DATABASE_OAUTH2_JWT_ALGORITHM"],
```
eschutho, line 179: should we put an expiration on this jwt?

betodealmeida: Yeah, good idea!

superset/utils/oauth2.py
betodealmeida: Great idea!

superset/models/core.py
```diff
@@ -982,6 +989,27 @@ def make_sqla_column_compatible(
...
+    __tablename__ = "database_user_oauth2_tokens"
+    __table_args__ = (sqla.Index("idx_user_id_database_id", "user_id", "database_id"),)
+
+    id = Column(Integer, primary_key=True)
+
+    user_id = Column(Integer, ForeignKey("ab_user.id"), nullable=False)
+    user = relationship(security_manager.user_model, foreign_keys=[user_id])
+
+    database_id = Column(Integer, ForeignKey("dbs.id"), nullable=False)
+    database = relationship("Database", foreign_keys=[database_id])
```
dpgaspar, line 1006: if a database connection is deleted it would be nice to set a cascade delete on it's tokens

superset/utils/lock.py
```diff
@@ -0,0 +1,88 @@
...
+    from superset.commands.key_value.delete import DeleteKeyValueCommand
+    from superset.commands.key_value.delete_expired import DeleteExpiredKeyValueCommand
+
+    key = uuid.uuid5(uuid.uuid5(uuid.NAMESPACE_DNS, namespace), serialize(kwargs))
+    logger.debug("Acquiring lock on namespace %s for key %s", namespace, key)
+    try:
+        DeleteExpiredKeyValueCommand(resource=KeyValueResource.LOCK).run()
+        CreateKeyValueCommand(
+            resource=KeyValueResource.LOCK,
+            codec=PickleKeyValueCodec(),
```
dpgaspar, line 75: the dreaded pickle, hard to imagine an attack vector on this case

superset/utils/oauth2.py
dpgaspar: for reference: https://flask.palletsprojects.com/en/3.0.x/async-await/#performance

superset/config.py
```diff
@@ -1394,6 +1394,24 @@ def EMAIL_HEADER_MUTATOR(  # pylint: disable=invalid-name,unused-argument
...
 TEST_DATABASE_CONNECTION_TIMEOUT = timedelta(seconds=30)
 
+# Details needed for databases that allows user to authenticate using personal
+# OAuth2 tokens. See https://github.com/apache/superset/issues/20300 for more
+# information
+DATABASE_OAUTH2_CREDENTIALS: dict[str, dict[str, Any]] = {
+    # "Google Sheets": {
+    #    "CLIENT_ID": "XXX.apps.googleusercontent.com",
+    #    "CLIENT_SECRET": "GOCSPX-YYY",
+    # },
```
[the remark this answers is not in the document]
betodealmeida: So far I only tested Google OAuth2 (GSheets & BigQuery), but I'm trying to account for implementations that might be slightly different.

superset/databases/api.py
```diff
@@ -1050,6 +1054,86 @@ def validate_sql(self, pk: int) -> FlaskResponse:
...
+              content:
+                text/html:
+                  schema:
+                    type: string
+            400:
+              $ref: '#/components/responses/400'
+            500:
+              $ref: '#/components/responses/500'
+        """
+        parameters = request.args.to_dict()
```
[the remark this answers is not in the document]
betodealmeida: Yeah, good point, I'll add that.

I do think the validation should ignore unknown fields, and just check that all fields needed are present. I had a problem in the past where [PKCE](https://oauth.net/2/pkce/) was introduced to an identity provider, and even though it was optional my website stopped working because I was validating the response against a schema that didn't have the field.

superset/databases/api.py
```diff
@@ -1050,6 +1054,86 @@ def validate_sql(self, pk: int) -> FlaskResponse:
...
+          summary: >-
+            Receive personal access tokens from OAuth2
+          description: ->
+            Receive and store personal access tokens from OAuth for user-level
+            authorization
+          parameters:
+          - in: query
+            name: state
+            schema:
+              type: string
```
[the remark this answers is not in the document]
betodealmeida: But it's not. 😊 The response could have just an `error` parameter.

superset/databases/api.py
```diff
@@ -1050,6 +1054,86 @@ def validate_sql(self, pk: int) -> FlaskResponse:
...
+          - in: query
+            name: code
+            schema:
+              type: string
```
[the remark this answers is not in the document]
betodealmeida: Same as above.

superset/utils/lock.py
betodealmeida: Yeah, the stored value is just `True`, so it should be fine.

superset/models/core.py
betodealmeida: Oh, great point!

superset/db_engine_specs/gsheets.py
```diff
@@ -136,6 +153,81 @@ def extra_table_metadata(
...
+            "scope": " ".join(SCOPES),
+            "access_type": "offline",
+            "include_granted_scopes": "false",
+            "response_type": "code",
+            "state": encode_oauth2_state(state),
+            "redirect_uri": redirect_uri,
+            "client_id": config["CLIENT_ID"],
+            "prompt": "consent",
+        }
+        return "https://accounts.google.com/o/oauth2/v2/auth?" + urlencode(params)
```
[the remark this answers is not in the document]
betodealmeida: Great idea, specially because it's versioned.

superset/databases/api.py
```diff
@@ -1050,6 +1054,86 @@ def validate_sql(self, pk: int) -> FlaskResponse:
...
+              $ref: '#/components/responses/500'
+        """
+        parameters = request.args.to_dict()
+        if "error" in parameters:
+            raise OAuth2Error(parameters["error"])
+
+        state = decode_oauth2_state(parameters["state"])
+
+        # exchange code for access/refresh tokens
+        database = db.session.query(Database).filter_by(id=state["database_id"]).one()
```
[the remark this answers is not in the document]
betodealmeida: You're right, I'll add one.

superset/databases/api.py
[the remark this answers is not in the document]
dpgaspar: good point, we can do something like:
``` python
class OAuth2ArgumentsSchema(Schema):
    class Meta:
        unknown = EXCLUDE
...

```

superset/databases/api.py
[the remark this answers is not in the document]
dpgaspar: true

superset/db_engine_specs/base.py
```diff
@@ -397,6 +425,79 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
...
+    @classmethod
+    def start_oauth2_dance(cls, database_id: int) -> None:
+        """
+        Start the OAuth2 dance.
+
+        This method will raise a custom exception that is captured by the frontend to
+        start the OAuth2 authentication. The frontend will open a new tab where the user
+        can authorize Superset to access the database. Once the user has authorized, the
+        tab sends a message to the original tab informing that authorization was
+        successfull (or not), and then closes. The original tab will automatically
```
villebro, line 444: typo
```suggestion
        successful (or not), and then closes. The original tab will automatically
```

villebro (commented):
Dropping intermediate review, will continue later

requirements/base.txt
```diff
@@ -319,7 +322,7 @@ rsa==4.9
     # via google-auth
 selenium==3.141.0
     # via apache-superset
-shillelagh[gsheetsapi]==1.2.10
+shillelagh[gsheetsapi]==1.2.18
```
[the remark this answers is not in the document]
betodealmeida: Happy to! But this is a feature that long existed (since I created the SIP), but there was a regression in the API — apparently Google stopped returning a status code of 200 when authorization was needed, and started returning a 403, which ironically broke the flow.

superset/db_engine_specs/base.py
betodealmeida: Thanks, I fixed this in a few other places as well.

superset/db_engine_specs/base.py
```diff
@@ -1891,7 +2008,7 @@ def cancel_query(  # pylint: disable=unused-argument
         :param query: Query instance
         :param cancel_query_id: Value returned by get_cancel_query_payload or set in
         other life-cycle methods of the query
-        :return: True if query cancelled successfully, False otherwise
+        :return: True if query cancelled successfuly, False otherwise
```
villebro, line 2011: It seems the `sed` operation caught this one by mistake 😆

villebro:
@betodealmeida I don't want to throw a wrench in the gears, but I'm wondering if this is sufficiently general.. Consider the following:
- You have two OAuth2 systems on different base urls using different clients.
- For a single database type, say Trino, you need to use different OAuth2 servers. Let's say "Trino db 1" uses client 1, and "Trino db 2" uses client 2. Furthermore assume that "Trino db 3" uses basic auth (current use case).

If we only have the ability to store these creds in the `DATABASE_OAUTH2_CREDENTIALS` config param, and use a single key per db type, I don't think we could support the above model. Furthermore, adding or changing clients will require a redeploy.

For this reason I would propose making it more generic as follows:
- Add a model/UI where one can create and manage client creds.
- Change the database model so that it has a foreign key to the client model above, and add a dropdown in the database modal for choosing which client to use (if any).

Thoughts?

betodealmeida:
@villebro that sounds like a very specific use case to me, is that common? For the databases that I'm planning to add support for, there's usually just a single endpoint:

- GSheets
- BigQuery
- Snowflake
- Dremio
- Databricks

Since these are all commercial cloud "databases" (in quotes because of GSheets).

For the use case you describe, a simple workaround would be to add the custom endpoint to the database `extra` field, and then modify the DB engine spec to check for the URL there, before using the default one from the config. If we want to make the solution more permanent we could then add a UI option to the DB modal allowing users to specify a custom, per-DB, endpoint.

vfauth:
@betodealmeida We also need to use Oauth2 with Trino as a DB, so we would love to have it supported.

However we don't need multiple clients (for now).

betodealmeida:
> @betodealmeida We also need to use Oauth2 with Trino as a DB, so we would love to have it supported.

@vfauth thanks, it's good to know!

I'm not planning to add support for Trino, but it should be relatively straightforward for someone else to add. If you look at the GSheets implementation it's basically (1) building the redirect URI and posting requests to get the (2) access token and (3) refresh a token. I assume in most implementations the only change will be the base URLs.

betodealmeida:
@villebro I gave it some more thought, and I really think the 2-layer approach that I mentioned (have a configuration at the DB engine spec level, and optional configuration at the DB level) is the best solution.

At Preset we need the DB engine spec level configuration, so that we can configure our own OAuth2 applications only once, and allow our customers to OAuth2 without having to set up their own applications. If we only had the per-DB configuration we'd either have to expose our client ID/secret (which we can't, for obvious reasons), or our customers would have to create their own client applications, which is not always trivial.

I think with the 2-layer approach we have the best of both worlds.

zhaoyongjie:
@betodealmeida 
> This PR introduces a new table called database_user_oauth2_tokens. 

Maybe a stupid question, but I think we should consider the implementation details. Why introduce a new table and a super complex foreign key topology? I believe this requirement is straightforward. We should store some tokens in the user table and keep the corresponding token in the user session. The metadata in Superset is extremely messy right now. Why do you guys always create new tables and create foreign keys for a super simple requirement?

villebro:
@betodealmeida the two layer approach would likely solve the main blocking painpoints. However, there's still some issues with this approach:
- In our use case, users that create new database connections don't tend to have access to the client details (these are created by another process). Requiring them to enter the client secrets is problematic from a governance perspective.
- We have lots of database connections referencing the same client. If the client needed to be changed, one would need to manually update each database entry manually. If we had a separate model for clients, updating the client entity would automatically apply to all databases that leverage it. For this reason, this really feels more like a foreign key, rather than a property on the Database model.

If we'd have regular API endpoints for the OAuth client resources, it should be possible to make sure that only the correct parties are able to CRUD them, solving the access problem.

While this approach (=foreign key in the database entity to the client entity) would create an additional step for Cloud based databases, where one has to choose which client to use as opposed to that being automatically selected, I don't see this as a big issue. We could even define which specs a client is valid for, and if only a single client exists for a particular spec, that would be autoselected when enabling OAuth2 in the db modal.

villebro:
> Maybe a stupid question, but I think we should consider the implementation details. Why introduce a new table and a super complex foreign key topology? I believe this requirement is straightforward. We should store some tokens in the user table and keep the corresponding token in the user session. The metadata in Superset is extremely messy right now. Why do you guys always create new tables and create foreign keys for a super simple requirement?

@zhaoyongjie when implementing our internal OAuth solution, we initially considered storing it in the User object. However, we decided against it, as it would create potential race conditions, if two threads wanted to update two separate user tokens at the same time (this typically happens if you open a dashboard that's referencing charts that use different OAuth clients). For this reason we ended up storing them in the key-value table. This has worked very well for us. I personally don't see any issue using a dedicated table for these, but I also think the key-value table would be a perfectly viable option, as it fits that use case well.

betodealmeida:
> Maybe a stupid question, but I think we should consider the implementation details. Why introduce a new table and a super complex foreign key topology? I believe this requirement is straightforward. We should store some tokens in the user table and keep the corresponding token in the user session. 

We can't store the tokens in the user table because it's a 1:n relationship. Storing in a table has the benefit that it's easy to make sure it's always encrypted, which is important since these are credentials, and we don't have to worry about session hijacking.

I wouldn't call it a super complex key topology, it's two foreign keys to the user and the database tables, with cascade deletes to clean them up.

> The metadata in Superset is extremely messy right now. Why do you guys always create new tables and create foreign keys for a super simple requirement?

I don't know, why do people raise these discussions only after the SIP has been approved?

betodealmeida:
@villebro ah, I see, I think I misunderstood your original point. Let me think about it.

zhaoyongjie:
> We can't store the tokens in the user table because it's a 1:n relationship. Storing in a table has the benefit that it's easy to make sure it's always encrypted, which is important since these are credentials, and we don't have to worry about session hijacking.

The `1:n relationship` can be maintained in the business codes and stories as JSON or any other preferred data structure. Credentials can also be managed within the business codes. The same data structure is used in dashboards.json_metadata and elsewhere.

> I wouldn't call it a super complex key topology, it's two foreign keys to the user and the database tables, with cascade deletes to clean them up.

If you have experienced the Sueprset migration process, you can experience how "poorly" designed foreign keys and cascade deletes are. :)

betodealmeida:
> The `1:n relationship` can be maintained in the business codes and stories as JSON or any other preferred data structure. Credentials can also be managed within the business codes. The same data structure is used in dashboards.json_metadata and elsewhere.

Why move the logic to the application when DBs are designed to handle this for us? With the benefit of indexes, atomicity, transactions, and more? And like Ville said, that would leave the flow vulnerable to race conditions.

> If you have experienced the Sueprset migration process, you can experience how "poorly" designed foreign keys and cascade deletes are. :)

I'm not sure I understand your point. Are you saying we normally don't implement foreign keys and cascade deletes correctly? If so, I'm open to suggestions on how to improve them in this PR.

betodealmeida:
@villebro what are your thoughts on introducing later the concept of the OAuth2 client, that can be associated with a database in the dropdown? And for now we'd have only the DB engine spec default client. I don't think adding the client model later would require any significant refactor, the only difference is changing the logic here:

https://github.com/apache/superset/pull/27631/files#diff-462c756528595a2be4a0a3c3063077f984384c7858ecdb5bd3e9603163c3678bR469-R473

villebro:
> what are your thoughts on introducing later the concept of the OAuth2 client, that can be associated with a database in the dropdown? And for now we'd have only the DB engine spec default client. I don't think adding the client model later would require any significant refactor, the only difference is changing the logic here: ...

I feel in general these types of configurations should never require changes to `superset_config.py` + subsequent redeploys. Also, if we do end up introducing the OAuth2 client model later on to support non-cloud db OAuth2 flow, then the config based approach will anyway become redundant, and will cause both maintenance burden, deprecation and removal at some point. This is based on the assumption that the best UX for adding cloud db clients would still be through the UI, where we would have a Test button that validates the client credentials, and can e.g. render a decoded token for the user to inspect and validate claims etc.

So in the interest of reducing throwaway work, maintenance burden and deprecation tasks in forthcoming Superset versions, I'd personally vote for doing the full monty here. This would mean introducing the OAuth2 client model, and making sure it works both for cloud dbs (Snowflake, Gsheets etc) and the more traditional OAuth2 based db setups, especially Trino, which I suspect is the one a lot of enterprise users are eagerly looking forward to. I can of course pitch in with the Presto/Trino logic, but it's quite trivial.

betodealmeida:
@villebro I feel like you're asking me to solve a very specific problem you have — handling multiple OAuth providers for the same database type — with a solution that doesn't solve the problem I have. 🙃 I don't want my users to have to select an OAuth client when they add a database; I want it to just work for them with a standard client, so that everyone who adds Snowflake will have it automatically configured to use the Snowflake endpoint.

At Preset, specifically, we don't want to store the client secret in the metadata database, because it creates a huge vector for attacks. If it gets exposed accidentally in the API it could lead to customers accessing resources from other customers, so we really need to have the secret stored in the config. So it would force us to have our clients creating their own OAuth applications, which is something that would be nice to have as optional, but not as a requirement.

There is precedence for configs that can be overridden on a per-asset basis; `CACHE_DEFAULT_TIMEOUT` comes to mind, with a default value that can be overridden per database or dataset. So I don't the model of having a generic OAuth2 config is at odds with also having a client model — specially because for Snowflake, BigQuery, GSheets, Clickhouse, Dremio, the users will never have to specify a different client at the DB level, since there's always a single provider.

villebro:
Don't get me wrong, my intention is not to block your use case, I'm simply trying to bring up the fact that this solution wouldn't work in a use case that I would need it to work in if I were to use this feature. I feel that the following are all common OAuth2 use cases:
- A single client shared for all connections of a single database type
- A single client shared for connections across multiple database types
- Multiple clients used for connections on a single database type

While defining a single client per spec in the config would solve the first use case, it does not generalize well into the latter use cases. For this reason I feel defining the clients in the config isn't a very user friendly or scalable solution.

Btw, I read up on the Snowflake OAuth2 documentation, and I get the impression that each Snowflake account will have an OAuth2 authorization endpoint of its own:

image

https://docs.snowflake.com/en/user-guide/oauth-custom#call-the-oauth-endpoints

If I'm reading this correctly, a single client for Snowflake connections on different accounts would not work. Or am I reading this wrong?

harksin:
Hey there, 
thanks for this so usefull PR !
I can confirm that when we use superset and Trino we are interested in a support for Oauth in superset with Trino to get rid of the user impersonation.

betodealmeida:
> I'm simply trying to bring up the fact that this solution wouldn't work in a use case that I would need it to work in if I were to use this feature.

The solution I'm proposing is a hybrid `superset_config.py` (this PR) + new model (future PR), which would work for **both** of of our use cases. We'd have to introduce the new model and UI elements, but most of the logic of this PR would remain unchanged. I'm not against your proposal, on the contrary, I think it's a great idea and an elegant solution.

But what I'm hearing from you is that we should go with only the new model, where the client ID/secret live in the metadata database. That solution is very suboptimal for my use case, since it would require users to create and manage their own applications.

You mentioned:

> This is based on the assumption that the best UX for adding cloud db clients would still be through the UI

But that is definitely not true in my use case. Google Sheets is one of our most popular databases used at Preset, and it would be great if users could connect to private sheets without having to figure out how to create an OAuth2 application first. BigQuery is probably **the** most popular, and the faster our users can start exploring their data, the better.

You also mentioned:

> then the config based approach will anyway become redundant, and will cause both maintenance burden, deprecation and removal at some point.

I don't see why the config approach would have to removed, since it complements the approach you're proposing. The default configuration for the feature is just an empty dictionary. We just need to add logic that checks the for the client information in two different places, as we already do for many things in Superset — this is IMHO the **main feature** of Superset, how it's adaptable and configurable to so many different use cases.

And in fairness, I do think that if we have the hybrid approach most people will prefer to use the UI to configure OAuth2, because as you said, it's easier to update the client in a single place and changes don't require a redeploy. But that doesn't mean that it's a solution that works for everyone.

(As for Snowflake, my understanding is that we can derive the OAuth2 URL from the account name, which is part of the SQLAlchemy URL.)

villebro:
> The solution I'm proposing is a hybrid `superset_config.py` (this PR) + new model (future PR), which would work for **both** of of our use cases. We'd have to introduce the new model and UI elements, but most of the logic of this PR would remain unchanged. I'm not against your proposal, on the contrary, I think it's a great idea and an elegant solution.
>
> But what I'm hearing from you is that we should go with only the new model, where the client ID/secret live in the metadata database. That solution is very suboptimal for my use case, since it would require users to create and manage their own applications.

I'm totally ok with the hybrid approach if we make sure we're not painting ourselves into a corner with it. The reason I feel uneasy with implementing a "one client for all connections of type x" approach as the initial implementation is that I feel it's inherently atypical for this type of flow: While it may be optimal for this specific use case, I don't think it works generally for most other use cases.

> But that is definitely not true in my use case. Google Sheets is one of our most popular databases used at Preset, and it would be great if users could connect to private sheets without having to figure out how to create an OAuth2 application first. BigQuery is probably **the** most popular, and the faster our users can start exploring their data, the better.

I understand this. Again, my reservation stems from the fact that this is not a typical OAuth2 flow, i.e. you likely can't use this for the majority of OAuth2 connectivity use cases, but the alternative I'm proposing works for this one, too. But as I said, I'm not against being able to provide static creds in the config where applicable.

> I don't see why the config approach would have to removed, since it complements the approach you're proposing. The default configuration for the feature is just an empty dictionary. We just need to add logic that checks the for the client information in two different places, as we already do for many things in Superset — this is IMHO the **main feature** of Superset, how it's adaptable and configurable to so many different use cases.

> (As for Snowflake, my understanding is that we can derive the OAuth2 URL from the account name, which is part of the SQLAlchemy URL.)

But even if you dynamically generate the URI, you would still need to store the client creds somewhere, right? In other words, the user would still need to pass the client id and secret somehow for the backend to be able to use them. Quoting from https://docs.snowflake.com/en/user-guide/oauth-custom#request-header:

image

betodealmeida:
@villebro for Snowflake the admin would add to `config.py`:

```python
DATABASE_OAUTH2_CREDENTIALS = {
    "Snowflake": {
        "CLIENT_ID": "XXX",
        "CLIENT_SECRET": "YYY",
    },
}
```

Then once a Snowflake database is added the DB engine spec can determine the OAuth2 URL from the SQLAlchemy URI, and use the information from the config to authorize the user. We could also optionally have the URL in the config, for deployments where that is possible.

I think we have two **_very_** different use cases, which is why we need these two workflows. For Preset, most users are  using cloud DBs, for the same reason they're using cloud Superset — they don't want to run their own infrastructure. And we want to have them connected to their data as quickly and easily as possible. We want to allow them to use our own application, but the `Admin` users shouldn't have access to the application details.

In your case, you're running your own very dynamic infrastructure, you have dozens (hundreds?) of databases, and multiple custom OAuth2 providers. Your admins are tech-savvy and have no trouble connecting Superset to complex database deployments, creating OAuth2 applications, and so on.

You mentioned "reducing throwaway work". I don't think there's a lot of work that will be thrown away if later we implement your suggestion. Most of the logic will remain, the only thing that would change would be how we pass the configuration to the DB engine specs — currently they (and by "they" I mean GSheets) read it from the config, but we could pass it explicitly so it come either from the config or from the assigned client.

I'm happy to add that abstraction in the near future, before someone starts working on a `n:n` between databases and OAuth2 client.

villebro:
> for Snowflake the admin would add to `config.py`:
> 
> ```python
> DATABASE_OAUTH2_CREDENTIALS = {
>     "Snowflake": {
>         "CLIENT_ID": "XXX",
>         "CLIENT_SECRET": "YYY",
>     },
> }
> ```
> 
> Then once a Snowflake database is added the DB engine spec can determine the OAuth2 URL from the SQLAlchemy URI, and use the information from the config to authorize the user. We could also optionally have the URL in the config, for deployments where that is possible.

This inherently means that you'll only be able to connect to a single Snowflake account using OAuth2. I'm sure that may be fine for many deployments, but as a general pattern that's unnecessarily restrictive.

Again I want to reiterate that my purpose here isn't to block this work. Rather, I'm trying to bring up issues with this approach that I'm expecting to surface after this PR is merged when orgs start rolling this out to their deployments. For that reason, it would be great to have an understanding of when these follow-up tasks are expected to be done to call OAuth2 support complete. Maybe we could discuss this over a meeting to see which orgs can commit to this work, and when?

betodealmeida:
> This inherently means that you'll only be able to connect to a single Snowflake account using OAuth2. I'm sure that may be fine for many deployments, but as a general pattern that's unnecessarily restrictive.

No, you could add as many different Snowflake databases as you wanted. Each one would connect to a different OAuth2 URL because each one would have a different SQLAlchemy URI. The client ID and the client are not tied to a single account, unless I'm missing something.

I'm planning to add support for Snowflake next, so if we need any kind of refactoring to support it, or if we need to implement the full model that you're proposing, I would be the one doing that work.

villebro:
> No, you could add as many different Snowflake databases as you wanted. Each one would connect to a different OAuth2 URL because each one would have a different SQLAlchemy URI. The client ID and the client are not tied to a single account, unless I'm missing something.

I would be surprised if one client works across multiple accounts. I haven't tried this personally, but I understand the process as follows:
- Login to your Snowflake account
- Issue a `CREATE SECURITY INTEGRATION TYPE = OAUTH` as per [here](https://docs.snowflake.com/en/user-guide/oauth-custom#integration-example)
- Use the creds from `SYSTEM$SHOW_OAUTH_CLIENT_SECRETS` when executing queries on your account as per [here](https://docs.snowflake.com/en/sql-reference/functions/system_show_oauth_client_secrets)

And if you'd want to integrate with another account, you'd redo those steps on that account, and then use those creds when executing queries against that one.

betodealmeida:
> I would be surprised if one client works across multiple accounts. I haven't tried this personally, but I understand the process as follows [...]

I haven't tried it personally yet, and if it doesn't, it's OK — I'll then implement your proposal next.

superset/utils/oauth2.py
mistercrunch: Interesting, my comment is out of scope here, resolving

superset/db_engine_specs/README.md
```diff
@@ -542,6 +543,70 @@ The method `get_url_for_impersonation` updates the SQLAlchemy URI before every q
 
 Alternatively, it's also possible to impersonate users by implementing the `update_impersonation_config`. This is a class method which modifies `connect_args` in place. You can use either method, and ideally they [should be consolidated in a single one](https://github.com/apache/superset/issues/24910).
 
+### OAuth2
```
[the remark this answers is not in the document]
mistercrunch: NIT: personally I like to break lines at 80-100 in docs/md, but there's not standard enforced there at the moment, so fine either way.

mistercrunch (approved):
Minor comments but overall LGTM. Stamping my approval but we may want another stamp from @villebro since he got deep in here already and seemed likely to push this further in the future.

superset/db_engine_specs/README.md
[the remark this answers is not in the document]
betodealmeida: Right, this README documents all the functionality of the DB engine specs and is targeting developers, explaining the methods that are needed. I'm happy to write additional docs about OAuth2 in the main website, I'll do that.

villebro (approved):
LGTM with a request for making it possible to define clients per database in the future, preferably via a dedicated client model.

betodealmeida:
@joaoferrao yeah, I think this is because we run the query in a separate thread for Trino. I'm away from my computer right now and traveling tomorrow, do you want to hop on a call on Wednesday to discuss this?

betodealmeida:
Basically, I think we need this fix: https://github.com/apache/superset/pull/29981

enryls:
Hello guys,

I tried this feature but I have a doubt about creating a Google Sheets Database as it's still asking me for the service account data and Sheets informations. Maybe I'm missing something?

Thanks

enryls:
> > I tried this feature but I have a doubt about creating a Google Sheets Database as it's still asking me for the service account data and Sheets informations. Maybe I'm missing something?
> 
> Can you give more details on what you did? Did you create an application, populate `DATABASE_OAUTH2_CREDENTIALS`, and enable user impersonation?

Hi @betodealmeida thanks for the reply.
I followed this steps:
- I create a Google Sheets Database in Superset with user impersonation (I created it with a public Google Sheets because if not I were not able to complete the creation)
- I setup the variable 
    ```py 
    DATABASE_OAUTH2_CLIENTS = {
            # "Google Sheets": {
            #     "id": "XXX.apps.googleusercontent.com",
            #     "secret": "GOCSPX-YYY",
            #     "scope": " ".join(
            #         [
            #             "https://www.googleapis.com/auth/drive.readonly",
            #             "https://www.googleapis.com/auth/spreadsheets",
            #             "https://spreadsheets.google.com/feeds",
            #         ]
            #     ),
            #     "authorization_request_uri": "https://accounts.google.com/o/oauth2/v2/auth",
            #     "token_request_uri": "https://oauth2.googleapis.com/token",
            # },
        }
    ```
- Then I tried to query in SQL Editor a private Google Sheets

enryls:
> > > I tried this feature but I have a doubt about creating a Google Sheets Database as it's still asking me for the service account data and Sheets informations. Maybe I'm missing something?
> > 
> > 
> > Can you give more details on what you did? Did you create an application, populate `DATABASE_OAUTH2_CREDENTIALS`, and enable user impersonation?
> 
> Hi @betodealmeida thanks for the reply. I followed this steps:
> 
> * I create a Google Sheets Database in Superset with user impersonation (I created it with a public Google Sheets because if not I were not able to complete the creation)
> * I setup the variable
>   ```python
>   DATABASE_OAUTH2_CLIENTS = {
>           # "Google Sheets": {
>           #     "id": "XXX.apps.googleusercontent.com",
>           #     "secret": "GOCSPX-YYY",
>           #     "scope": " ".join(
>           #         [
>           #             "https://www.googleapis.com/auth/drive.readonly",
>           #             "https://www.googleapis.com/auth/spreadsheets",
>           #             "https://spreadsheets.google.com/feeds",
>           #         ]
>           #     ),
>           #     "authorization_request_uri": "https://accounts.google.com/o/oauth2/v2/auth",
>           #     "token_request_uri": "https://oauth2.googleapis.com/token",
>           # },
>       }
>   ```
> * Then I tried to query in SQL Editor a private Google Sheets

After some tentatives I was able to login, but there is a strange behaviour, the first time I query the sheet I receive a "Server Error" then I refresh the page and I receive "Provide authorization" message. I'll investigate more and let you know.

PR #27641, opened 2024-03-25
chore: fix master builds + bump python library "cryptography"

mistercrunch:
Originally generated by @supersetbot 🤖

@mistercrunch here taking over after noticing that the `master` builds were having some issue, and addressing here while bumping this library.

This is pushing the reusable action `setup-backend` I started in a previous PR to bring things together in a simpler way.

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -24,6 +24,7 @@ runs:
         cache: ${{ inputs.cache }}
     - name: Install dependencies
       run: |
+        sudo apt-get update && sudo apt-get -y install libldap2-dev libsasl2-dev
         pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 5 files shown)


.github/actions/setup-backend/action.yml
```diff
@@ -24,6 +24,7 @@ runs:
         cache: ${{ inputs.cache }}
     - name: Install dependencies
       run: |
+        sudo apt-get update && sudo apt-get -y install libldap2-dev libsasl2-dev
```
mistercrunch, line 27: I think it makes sense to bring python-related os dependencies in this `setup-backend` reusable action instead of within the `bashlib.sh` that I still don't understand how it gets discovered and executed.

lilykuang (approved):

geido (approved):

PR #27647, opened 2024-03-26
feat: move supersetbot out of repo

mistercrunch:
This PR
- moves `supersetbot` out of here and onto here https://github.com/apache-superset/supersetbot
- improves the GHA to set it up


mistercrunch:
I decided to move the bot out of this repo and onto here -> https://github.com/apache-superset/supersetbot

rusackas:
Just curious, is this out of necessity, or just easier to work on it? It might be nice to move it back into Superset/ASF control once the dust settles if it's just to make it easier to get things spun up and tested for a while.

mistercrunch:
Yes the main reason is it doesn't require the same level of scrutiny and rigor as the main repo, and it can have it's own CI/release cycles/tests, .

rusackas (approved):
Sounds reasonable... perhaps one day it can make a grand homecoming or something. I think the main thing is we need docs on the wiki so we can evangelize its use.

mistercrunch:
Right, it can always move back here once things settle.

PR #27700, opened 2024-03-27
fix: row limits & row count labels are confusing

mistercrunch:
### SUMMARY

Currently, in explore & dashboard, we usually apply a row limit
on the query we issue against the database.

Following this, some visualization backend code does some
post-processing on data frames using pandas, typically pivoting
some things, which affects the result set's row count in
intricate capacities.

Now from a UX standpoint the user is exposed with:
- row limits in the control panels
- row count in various areas of the UI where visualization, preview,
  samples and raw results are shown

Currently we show the rowcount that's from row-processing. So maybe
a hard limit was applied at 1000 rows on the database, but we
pivot and it goes does to say 532, and we show the user that
we haven't hit the limit when we actually did.

Also note that the component that shows rowcount is supposed
to turn red if limit is hit, letting the user know they are looking at
truncated data.

This PR carries the right row count to the `RowCountLabel` above the visualization in explore so that it can turn red if the limit is reached reliably. 

For now, this PR does not do anything new around the other RowCountLabel located in the `RESULTS` (south) pane. This area show the post-processed rows from cache, and seem like is has to show its matching post-processed row count. Note that the row count here is not guaranteed to match the one above.

Passing the actual row count returned from the database and using it to surface the [improved] tooltip
Screenshot 2024-03-28 at 5 43 43 PM


villebro:
This has definitely caused confusion over the years. Great initiative @mistercrunch !

betodealmeida (approved):
Looks good and it's a great improvement!

I wonder if we should standardize the name, though, using `sql_rowcount` in the backend and something like `sqlRowCount` in the frontend? (instead of just `rowcount`)

mistercrunch:
Yeah I wasn't sure either and there was not clear pattern established for me to follow. Looked that the person who wrote the code before went out of their way to go flat-case (coltypes, colnames, rowcount, ....).

I'm not sure what best practices are and what the most common on our repo is. If I was to take on more refactoring I think i'd tackle something around grouping objects under say a `queryMetadata` object and passing the object around as much as possible instead of spelling out all the attributes at every step of the way... That would allow say augmenting `queryMetadata` in the backend with an extra attribute and having it flow all the way through a bunch of objects. Merging for now, and we could revisit the camel-case-on-frontend-policy

PR #27701, opened 2024-03-27
fix: useTruncation infinite loop, reenable dashboard cross links on ChartList

kgabryje:
### SUMMARY
Due to a bug in `useChildElementTruncation`, there was an edge case which could trigger an infinite loop of rerenders. The bug occurs when the difference in width of subsequent numbers after the plus sign (e.g. +4 and +5) is just big enough to change the truncation state to the other number. This PR ensures that the calculation is performed only once, and if the edge case is triggered, we simply accept the off-by-one error.
This PR also implements a ResizeObserver which recalculates the truncations if the container size changes. Previously we relied on the width values from references in `useLayoutEffect`'s dependency array, which was not a correct approach, since the changes of those values would not trigger a rerender.

This change allowed us to bring back the feature that relied on the truncation logic, which is dashboard cross links on Chart List page. The feature adds a column to the list which displays the dashboards that given chart is added to.

Before (occurred long ago when "dashboards added to" column was not yet hidden):

image

After:

image

### TESTING INSTRUCTIONS
1. Visit ChartList
2. Verify that "Dashboards added to" column is visible and that it correctly displays the dashboards. Dashboard names should be clickable and navigate to given dashboard. Truncated dashboards should appear in a tooltip

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
@@ -112,7 +112,7 @@ describe('Charts list', () => {
       cy.getBySel('sort-header').eq(1).contains('Name');
       cy.getBySel('sort-header').eq(2).contains('Type');
       cy.getBySel('sort-header').eq(3).contains('Dataset');
-      cy.getBySel('sort-header').eq(4).contains('Dashboards added to');
+      cy.getBySel('sort-header').eq(4).contains('On dashboards');
       cy.getBySel('sort-header').eq(5).contains('Owners');
       cy.getBySel('sort-header').eq(6).contains('Last modified');
       cy.getBySel('sort-header').eq(7).contains('Actions');
--- a/superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
@@ -54,7 +54,7 @@ function visitChartList() {
 }
 
 describe('Charts list', () => {
-  describe.skip('Cross-referenced dashboards', () => {
+  describe('Cross-referenced dashboards', () => {
     beforeEach(() => {
       cy.createSampleDashboards([0, 1, 2, 3]);
       cy.createSampleCharts([0]);
@@ -112,9 +112,10 @@ describe('Charts list', () => {
       cy.getBySel('sort-header').eq(1).contains('Name');
       cy.getBySel('sort-header').eq(2).contains('Type');
       cy.getBySel('sort-header').eq(3).contains('Dataset');
-      cy.getBySel('sort-header').eq(4).contains('Owners');
-      cy.getBySel('sort-header').eq(5).contains('Last modified');
-      cy.getBySel('sort-header').eq(6).contains('Actions');
+      cy.getBySel('sort-header').eq(4).contains('Dashboards added to');
+      cy.getBySel('sort-header').eq(5).contains('Owners');
+      cy.getBySel('sort-header').eq(6).contains('Last modified');
+      cy.getBySel('sort-header').eq(7).contains('Actions');
     });
 
     it('should sort correctly in list mode', () => {
--- a/superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
+++ b/superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
@@ -83,7 +83,9 @@ const useChildElementTruncation = () => {
 
     onResize();
 
-    return obs.disconnect;
+    return () => {
+      obs.disconnect();
+    };
   }, [plusRef.current]); // plus is rendered dynamically - the component rerenders the hook when plus appears, this makes sure that useLayoutEffect is rerun
 
   return [elementRef, plusRef, elementsTruncated, hasHiddenElements] as const;
--- a/superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
+++ b/superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { RefObject, useLayoutEffect, useState, useRef } from 'react';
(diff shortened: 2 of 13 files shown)


superset-frontend/src/pages/ChartList/index.tsx
```diff
@@ -157,6 +157,20 @@ const StyledActions = styled.div`
   color: ${({ theme }) => theme.colors.grayscale.base};
 `;
 
+const DashboardCrossLinks = React.memo(
```
justinpark, line 160: It appears to be more beneficial to create a dedicated file specifically for this component.

kgabryje: Agreed 👍

kgabryje: done

kgabryje:
@justinpark would you be able to do another pass?

superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
```diff
@@ -27,92 +27,68 @@ import { RefObject, useLayoutEffect, useState, useRef } from 'react';
...
 
-    if (scrollWidth > clientWidth) {
-      // "..." is around 6px wide
-      const truncationWidth = 6;
-      const plusSize = plusRefElement?.offsetWidth || 0;
-      const maxWidth = clientWidth - truncationWidth;
-      const elementsCount = childNodes.length;
+      if (scrollWidth > clientWidth) {
+        // "..." is around 6px wide
+        const truncationWidth = 6;
```
betodealmeida, line 47: I thought we had a function for computing the width of a given element? Can we use something like `element.getBoundingClientRect().width` to get the exact width of the ellipsis rendered in an invisible element?

superset-frontend/src/dashboard/components/nativeFilters/FilterCard/NameRow.tsx
```diff
@@ -31,8 +31,7 @@ export const NameRow = ({
   hidePopover,
 }: FilterCardRowProps & { hidePopover: () => void }) => {
   const theme = useTheme();
-  const filterNameRef = useRef<HTMLElement>(null);
-  const [elementsTruncated] = useTruncation(filterNameRef);
+  const [filterNameRef, , elementsTruncated] = useTruncation();
```
betodealmeida, line 34: #TIL

superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
```diff
@@ -27,92 +27,68 @@ import { RefObject, useLayoutEffect, useState, useRef } from 'react';
...
-    elementRef.current?.clientWidth,
-    elementRef,
-  ]);
 
-  return [elementsTruncated, hasHiddenElements];
+    onResize();
+
+    return () => {
+      obs.disconnect();
+    };
```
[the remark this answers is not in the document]
kgabryje: Oh good point! Though to be fair, I’m not sure about readability - I was halfway through writing a comment about how we should return a cleanup function rather than returning `obs.disconnect()`… but it’s early morning, so maybe it’s my fault 😉

eschutho (approved):

superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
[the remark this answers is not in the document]
villebro: I'm almost 50-50 here, but I'm leaning in the direction of @betodealmeida here, so I'd say +1 for changing to `return obs.disconnect;`

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -112,9 +112,10 @@ describe('Charts list', () => {
       cy.getBySel('sort-header').eq(1).contains('Name');
       cy.getBySel('sort-header').eq(2).contains('Type');
       cy.getBySel('sort-header').eq(3).contains('Dataset');
-      cy.getBySel('sort-header').eq(4).contains('Owners');
-      cy.getBySel('sort-header').eq(5).contains('Last modified');
-      cy.getBySel('sort-header').eq(6).contains('Actions');
+      cy.getBySel('sort-header').eq(4).contains('Dashboards added to');
```
villebro, line 115: This is maybe out of scope, but I feel `Dashboards added to` is too verbose, and I think just going with `Dashboards` or `On dashboards` makes more sense.

villebro (approved):
In addition to the fix, very nice cleanup work here 👍 A non-blocking comment on the column name + supporting Beto's change request.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
kgabryje: @yousoph @kasiazjc wdyt?

superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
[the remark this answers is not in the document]
kgabryje: 2 vs 1, request for change has passed 🙂

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
kgabryje: @kasiazjc says that "On dashboards" sounds fine. I changed it here (in charts list) and also in Explore for consistence

superset-frontend/packages/superset-ui-core/src/hooks/useTruncation/useChildElementTruncation.ts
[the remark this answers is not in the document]
kgabryje: @villebro @betodealmeida With `return obs.disconnect`, the page is crashing with `illegal invocation` error. To be honest I don't know why it doesn't work, but it does work with `return () => { obs.disconnect(); }` so I'm going with that

image

PR #27706, opened 2024-03-27
fix: Select onChange is fired when the same item is selected in single mode

### LINKED ISSUE #27668: Select component fires onChange when selected value has not changed
### Bug description

When rendering the Select component in singleMode (only one selection can be selected), selecting the currently selected value from the dropdown options triggers the onChange function to fire, even though the selection has not changed.

### How to reproduce the bug

Go to alerts and reports
Add an alert or report
Under notification settings, select Email
Add some text to the recipients section
Select Email again
See that the recipients field has been cleared

### Screenshots/recordings

https://www.loom.com/share/9f9829e8877a444db074e6f8ce0f669f?sid=1e4fe88d-4075-4b6a-9dc0-50749316baa2

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

ALERT_REPORTS=True

### Checklist

michael-s-molina:
### SUMMARY
Prevents Select's `onChange` from being fired when the same item is selected in single mode. 

Fixes https://github.com/apache/superset/issues/27668.

### TESTING INSTRUCTIONS
CI.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Select/utils.tsx
+++ b/superset-frontend/src/components/Select/utils.tsx
@@ -49,9 +49,9 @@ export function getValue(
   return isLabeledValue(option) ? option.value : option;
 }
 
-export function equalValues(a: V | LabeledValue, b: V | LabeledValue) {
-  const actualA = isObject(a) && 'value' in a ? a.value : a;
-  const actualB = isObject(b) && 'value' in b ? b.value : b;
+export function isEqual(a: V | LabeledValue, b: V | LabeledValue, key: string) {
+  const actualA = isObject(a) && key in a ? a[key] : a;
+  const actualB = isObject(b) && key in b ? b[key] : b;
   // When comparing the values we use the equality
   // operator to automatically convert different types
   // eslint-disable-next-line eqeqeq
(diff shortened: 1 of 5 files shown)


superset-frontend/src/components/Select/utils.tsx
```diff
@@ -49,22 +49,25 @@ export function getValue(
...
   return optionsArray.find(
     x =>
-      // eslint-disable-next-line eqeqeq
-      x == value ||
-      (isObject(x) &&
-        // eslint-disable-next-line eqeqeq
-        (('value' in x && x.value == value) ||
-          (checkLabel && 'label' in x && x.label === value))),
+      equalValues(x, value) ||
+      (checkLabel && isObject(x) && 'label' in x && x.label === value),
```
kgabryje, line 70: Do you think it makes sense to extend `equalValues` to accept "key" argument, so you could call `equalValues(x, checkLabel, "label")` here?

kgabryje (approved):
LGTM, added 1 non blocking comment

superset-frontend/src/components/Select/utils.tsx
michael-s-molina: Good call. Done.

PR #27716, opened 2024-03-27
feat(plugin): color option for table with time comparison

lilykuang:
### SUMMARY

Adding color option for table with time comparison
- Create a new collapsible section titled "Visual formatting."
- Inside, include a subsection called "Cell bars" with:
   - Cell bars
   - Color +/- (renamed to "Add colors to cell bars for +/-")
   - Align +/-
- Add a "Conditional formatting" section:
   - Enable a checkbox for "Basic conditional formatting for comparison" when comparison is active.
   - Upon selecting the checkbox, reveal "Color Scheme for comparison" option.
      - Default color scheme: "Green for increase, red for decrease."
      - Alternative option: "Red for decrease, green for increase."

time-comparison-color

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
@@ -50,7 +50,8 @@ import {
   getStandardizedControls,
 } from '@superset-ui/chart-controls';
 
-import { PAGE_SIZE_OPTIONS } from './consts';
+import { COMPARISON_PREFIX, PAGE_SIZE_OPTIONS } from './consts';
+import { ColorSchemeEnum } from './types';
 
 function getQueryMode(controls: ControlStateMapping): QueryMode {
   const mode = controls?.query_mode?.value;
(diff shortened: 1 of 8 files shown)


john-bodley:
My only comment regarding this is I hope we're thinking in a holistic manner in terms of how best to define color options across all chart types as opposed to a specific chart. It's critical that we strive for a coherent UX whereas historically Superset has felt rather fragmented in nature. The migration to ECharts definitely helped provide a layer of visual consistency which would be good to adhere to.

zhaoyongjie:
@Antonio-RiveroMartnez It makes sense! Thanks for the clarification.

superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -520,41 +598,41 @@ const config: ControlPanelConfig = {
...
-                    | ChartDataResponseResult
-                    | undefined,
-                };
-              },
+              choices: [
+                [ColorSchemeEnum.Green, 'Green for increase, red for decrease'],
+                [ColorSchemeEnum.Red, 'Red for increase, green for decrease'],
+              ],
+              visibility: ({ controls }) =>
+                controls?.comparison_color_enabled?.value === true,
```
Antonio-RiveroMartnez, line 631: Non blocking: Should we make this control also visible only if user has the `Enable Time Comparison` set as true only? if false then ignore this setting entirely? SO we don't affect existing vizs?

Antonio-RiveroMartnez (approved):

PR #27717, opened 2024-03-27
chore(explore): Hide non-droppable metric and column list

justinpark:
### SUMMARY
There can be certain items that cannot be accepted in all options. (For instance, metrics cannot be accepted in the columns selection. In cases where the controls solely consist of the columns selection, none of the metrics will be accepted in the overall dropzone, as shown in the screenshot below.)

To address this, the commit takes a step forward by checking the eligibility of each filter and hiding the item when there are no eligible options available.

Before:

https://github.com/apache/superset/assets/1392866/3d3e6b92-c461-49ba-ad24-f23528acf22c

After:

https://github.com/apache/superset/assets/1392866/7bc72d2d-f9b8-4603-b607-926c49af9004

### TESTING INSTRUCTIONS

- Create a table chart with the dataset
- Switch to "Raw Records" and then "Metrics" should be hidden

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/DatasourcePanel/index.tsx
+++ b/superset-frontend/src/explore/components/DatasourcePanel/index.tsx
@@ -176,6 +176,8 @@ export default function DataSourcePanel({
     );
   }, [dropzones, metrics]);
 
+  const hiddenColumnCount = _columns.length - allowedColumns.length;
+  const hiddenMetricCount = metrics.length - allowedMetrics.length;
   const [showSaveDatasetModal, setShowSaveDatasetModal] = useState(false);
   const [inputValue, setInputValue] = useState('');
   const [showAllMetrics, setShowAllMetrics] = useState(false);
@@ -346,6 +348,8 @@ export default function DataSourcePanel({
                   onCollapseMetricsChange: setCollapseMetrics,
                   collapseColumns,
                   onCollapseColumnsChange: setCollapseColumns,
+                  hiddenMetricCount,
+                  hiddenColumnCount,
                 }}
                 overscanCount={5}
               >
--- a/superset-frontend/src/explore/components/DatasourcePanel/index.tsx
+++ b/superset-frontend/src/explore/components/DatasourcePanel/index.tsx
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useContext, useMemo, useState } from 'react';
 import {
   css,
   DatasourceType,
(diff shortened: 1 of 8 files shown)


superset-frontend/src/explore/components/DatasourcePanel/index.tsx
```diff
@@ -122,18 +127,48 @@ const StyledInfoboxWrapper = styled.div`
...
+      } catch {} // eslint-disable-line no-empty
+    }
+    return extra;
+  }, [datasource.extra]);
+  const allowlistOnly = extra.disallow_adhoc_metrics;
+
+  const allowedColumns = useMemo(() => {
+    const validators = Object.values(dropzones);
+    if (!isArray(_columns)) return [];
+    return allowlistOnly
```
michael-s-molina, line 156: @justinpark If we have the concept of validators where every item is checked to see if it's allowed, similar to a security check, wouldn't be the responsibility of the validator to check for `disallow_adhoc_metrics`? In other words, could we always validate the items with the default implementation being that all items are allowed?

michael-s-molina (commented):
Thank you for the PR @justinpark.

superset-frontend/src/explore/components/DatasourcePanel/index.tsx
justinpark: > In other words, could we always validate the items with the default implementation being that all items are allowed?

@michael-s-molina As mentioned earlier, in the adhoc query, all columns and metrics are accepted. Therefore, applying additional filtering in cases where non-`disallowed-adhoc` case can be considered unnecessary.

justinpark: We synced offline and will ensure that these filters are not affected by the 'disallow-adhoc' condition.

michael-s-molina:
@justinpark I think it would be important to show some kind of message/feedback to the users when we have columns or metrics hidden because they can't be used anywhere. Maybe @kasiazjc can help us here.

justinpark:
@michael-s-molina If you are interested in considering feedback, this proposal suggests adopting a Gmail-style approach.
_DEV__Superset

cc: @kasiazjc

michael-s-molina:
This looks nice @justinpark. We won't have the Undo link right? I'm assuming that you can't undo the applicability of columns because they depend on what's already configured in the controls.

justinpark:
> I would only keep the background white to not mix it with the draggable items and because this looks like a secondary information that does not need to be highlighted frequently. Once users view this for the first time, they will get used to it.

Sounds good. I'll update accordingly.

> We won't have the Undo link right? I'm assuming that you can't undo the applicability of columns because they depend on what's already configured in the controls.

I agree with your opinion. Let me remove "undo" now.

michael-s-molina:
@justinpark If you resize the left panel, it's cutting the text.

Screenshot 2024-04-04 at 08 48 03

justinpark:
> justinpark If you resize the left panel, it's cutting the text.

@michael-s-molina I fixed by the text ellipsis.

eclipsis

michael-s-molina (approved):

PR #27718, opened 2024-03-27
refactor(plugins):  BigNumber Time Comparison with existing time_offset API

Antonio-RiveroMartnez:
### SUMMARY
Previously we implemented the time comparison in BigNumber viz using some new custom properties in the queryObject to compute the offsets. We are moving away from that approach in order to be more consistent with the way Superset handles time comparison in other visualizations.

Some of the changes of this PR are:
- Use existing time_offset APIs instead of the instant comparison ones
- Create new shared section to include the controls for time comparison, some taken from Advanced Analytics
- Create a new control label to display the comparison ranges using existing time_range API and not a new one

There will be a follow up PR cleaning up the customizations from previous approach, that clean up is left out of this PR to ease the review process.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [X] Required feature flags: FEATURE_CHART_PLUGINS_EXPERIMENTAL
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
+++ b/superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
@@ -57,7 +57,7 @@ export const parseDttmToDate = (dttm: string): Date => {
     now.setUTCFullYear(now.getUTCFullYear() - 1, 0, 1);
     return now;
   }
-  if (dttm.includes('ago')) {
+  if (dttm?.includes('ago')) {
     const parts = dttm.split(' ');
     const amount = parseInt(parts[0], 10);
     const unit = parts[1];
--- a/superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
+++ b/superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
@@ -16,53 +16,77 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import moment, { Moment } from 'moment';
 import { ensureIsArray } from '../utils';
 
-export const parseDttmToMoment = (dttm: string): Moment => {
-  if (dttm === 'now') {
-    return moment().utc().startOf('second');
-  }
-  if (dttm === 'today' || dttm === 'No filter') {
-    return moment().utc().startOf('day');
+const DAY_IN_MS = 24 * 60 * 60 * 1000;
+export const parseDttmToDate = (dttm: string): Date => {
+  const now = new Date();
+  now.setUTCHours(0, 0, 0, 0);
+
+  if (dttm === 'now' || dttm === 'today' || dttm === 'No filter') {
+    return now;
   }
   if (dttm === 'Last week') {
-    return moment().utc().startOf('day').subtract(7, 'day');
+    now.setUTCDate(now.getUTCDate() - 7);
+    return now;
   }
   if (dttm === 'Last month') {
-    return moment().utc().startOf('day').subtract(1, 'month');
+    now.setUTCMonth(now.getUTCMonth() - 1);
+    now.setUTCDate(1);
+    return now;
   }
   if (dttm === 'Last quarter') {
-    return moment().utc().startOf('day').subtract(1, 'quarter');
+    now.setUTCMonth(now.getUTCMonth() - 3);
+    now.setUTCDate(1);
+    return now;
   }
   if (dttm === 'Last year') {
-    return moment().utc().startOf('day').subtract(1, 'year');
+    now.setUTCFullYear(now.getUTCFullYear() - 1);
+    now.setUTCDate(1);
+    return now;
   }
   if (dttm === 'previous calendar week') {
-    return moment().utc().subtract(1, 'weeks').startOf('isoWeek');
+    now.setUTCDate(now.getUTCDate() - (now.getUTCDay() || 7));
+    return now;
   }
   if (dttm === 'previous calendar month') {
-    return moment().utc().subtract(1, 'months').startOf('month');
+    now.setUTCMonth(now.getUTCMonth() - 1, 1);
+    return now;
   }
   if (dttm === 'previous calendar year') {
-    return moment().utc().subtract(1, 'years').startOf('year');
+    now.setUTCFullYear(now.getUTCFullYear() - 1, 0, 1);
+    return now;
   }
   if (dttm.includes('ago')) {
     const parts = dttm.split(' ');
     const amount = parseInt(parts[0], 10);
-    const unit = parts[1] as
-      | 'day'
-      | 'week'
-      | 'month'
-      | 'year'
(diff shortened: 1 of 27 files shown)


superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
```diff
@@ -0,0 +1,101 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import moment, { Moment } from 'moment';
+import { ensureIsArray } from '../utils';
+
+export const parseDttmToMoment = (dttm: string): Moment => {
+  if (dttm === 'now') {
+    return moment().utc().startOf('second');
```
eschutho, line 24: you should be able to do this with just vanilla javascript instead of Moment. We're currently trying to remove moment wherever possible.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/controlPanel.ts
```diff
@@ -180,14 +108,12 @@ const config: ControlPanelConfig = {
         ],
       ],
     },
+    sections.timeComparisonControls(false, false, false),
```
eschutho, line 111: I'd suggest passing an object here instead of arguments so that we can tell which values are being set to false here for example.

superset/common/query_context_processor.py
```diff
@@ -488,21 +511,17 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
-                index = (get_base_axis_labels(query_object.columns) or [DTTM_ALIAS])[0]
-                if not dataframe_utils.is_datetime_series(offset_metrics_df.get(index)):
-                    raise QueryObjectValidationError(
-                        _(
-                            "A time column must be specified "
-                            "when using a Time Comparison."
-                        )
+                if dataframe_utils.is_datetime_series(offset_metrics_df.get(index)):
+                    # modifies temporal column using offset
+                    offset_metrics_df[index] = offset_metrics_df[index] - DateOffset(
```
michael-s-molina, line 516: Hi @Antonio-RiveroMartnez. This logic to calculate the offsets is not correct and I'm fixing it in https://github.com/apache/superset/pull/28432. I think it would be prudent to wait for that fix and rebase your PR to see if it will affect this new feature in any way.

Antonio-RiveroMartnez: Hey @michael-s-molina , thanks for the heads up! Ok, will do, it's very likely that your PR gets merged before this one so it should be ok for rebasing.

eschutho (approved):
lgtm

michael-s-molina:
@Antonio-RiveroMartnez I added the `hold:testing` label to this PR. I'll wait for your rebase and will test the changes using Airbnb's production data, the same way I did for https://github.com/apache/superset/pull/28432 just to confirm that the time-shifts remain in sync with legacy charts.

superset-frontend/packages/superset-ui-core/test/time-comparison/getTimeOffset.test.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { getTimeOffset, parseDttmToDate } from '@superset-ui/core';
+
+describe('parseDttmToDate', () => {
```
michael-s-molina, line 21: Can you avoid `describe` and just use `test` to follow this [best practice](https://github.com/apache/superset/wiki/Testing-Guidelines-and-Best-Practices#avoid-nesting-when-youre-testing)? We're slowly refactoring the tests to move in that direction. 

I noticed that this file name is `getTimeOffset.test.ts` but it's also testing `parseDttmToDate`. Can you rename the file to something more generic or breaking it into 2 test files?

superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
```diff
@@ -0,0 +1,121 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { ensureIsArray } from '../utils';
+
+const DAY_IN_MS = 24 * 60 * 60 * 1000;
+export const parseDttmToDate = (dttm: string): Date => {
```
michael-s-molina, line 22: Isn't `parseDttmToDate` similar to [parse_human_datetime](https://github.com/apache/superset/blob/821c7d7f2c430c4a4294883a66128ba98fd949c5/superset/utils/date_parser.py#L57)? If yes, is there a way we can reuse that so we don't have 2 implementations to maintain?

superset-frontend/packages/superset-ui-core/test/time-comparison/fetchTimeRange.test.ts
```diff
@@ -116,3 +117,54 @@ it('returns a formatted error message from response', async () => {
     error: 'Network error',
   });
 });
+
+it('fetchTimeRange with shift', async () => {
```
michael-s-molina, line 121: Same comment to use `test` instead of `it`.

Antonio-RiveroMartnez: Done!

superset-frontend/packages/superset-ui-core/test/time-comparison/getTimeOffset.test.ts
Antonio-RiveroMartnez: Done!

superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
lilykuang: Thank you for pointing this out. parseDttmToDate and parse_human_datetime are very similar. However, parse_human_datetime is in the backend, while parseDttmToDate is also for validating a [DatePicker component](https://github.com/apache/superset/pull/27718/files#diff-76b79c56cfbdda04728e5590bd1927760383d5c44c9d265a828da8804472a271R65-R81) in the frontend. This separation makes unifying them tricky. I'm open to any suggestions you might have.

michael-s-molina:
Thanks for rebasing and addressing the comments @Antonio-RiveroMartnez. I'll execute the tests using Airbnb's production data.

PR #27725, opened 2024-03-27
chore(sqllab): Do not strip comments when executing SQL statements

john-bodley:
### SUMMARY

I realize there is merit for stripping SQL comments when parsing SQL to extract tables etc., i.e., to aid with simplification, however I'm not _entirely_ sure why we do this when executing SQL especially given that numerous SQL dialects supports  optimizer hints in the form of comments:

- [MySQL](https://dev.mysql.com/doc/refman/8.3/en/optimizer-hints.html)
- [Spark SQL](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-hints.html)

This PR ensures when running SQL statements in SQL Lab we don't strip away the comments.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_lab.py
+++ b/superset/sql_lab.py
@@ -419,7 +419,6 @@ def execute_sql_statements(
     # Breaking down into multiple statements
     parsed_query = ParsedQuery(
         rendered_query,
-        strip_comments=True,
         engine=db_engine_spec.engine,
     )
     if not db_engine_spec.run_multiple_statements_as_one:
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Oh yeah, we should definitely not be stripping comments. There are multiple reasons to leave them — in addition to optimization like you mentioned I've also seen them used for auditing. Thanks for fixing this!

betodealmeida:
(I restarted the flaky MySQL test for you.)

PR #27773, opened 2024-03-28
feat(alert report tabs): adding feature flag

fisjac:
### SUMMARY

Adds a feature flag for implementing a new feature enabling the user to select which tab of the dashboard the user would like to send as an alert or report.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERT_REPORT_TABS
- [ ] Changes UI

### DIFF
--- a/RESOURCES/FEATURE_FLAGS.md
+++ b/RESOURCES/FEATURE_FLAGS.md
@@ -27,6 +27,7 @@ These features are considered **unfinished** and should only be used on developm
 
 [//]: # "PLEASE KEEP THE LIST SORTED ALPHABETICALLY"
 
+- ALERT_REPORT_TABS
 - ENABLE_ADVANCED_DATA_TYPES
 - PRESTO_EXPAND_DATA
 - SHARE_QUERIES_VIA_KV_STORE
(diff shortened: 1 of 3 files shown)


eschutho (approved):

michael-s-molina (changes requested):
I left a comment about the feature flag. If needed, we also need to update `FEATURE_FLAGS`.md.

fisjac:
@michael-s-molina could you clarify what you mean. It seems we have a couple options here:
1. build the feature flag and merge in. This would enable building smaller PRs to build out the feature, but possibly creates and unused feature flag if the feature isn't merged.
2. Build the feature as a separate branch and merge in together alongside the feature flag, creating a large PR.

We can go either way, but it would seem smaller PRs will be much easier to implement/review.

rusackas:
Hey @fisjac - I think @michael-s-molina means that you just need to add an entry to [this file](https://github.com/apache/superset/blob/master/RESOURCES/FEATURE_FLAGS.md). Presumably under the "In Development" section. Note that if the feature flag gets added and then removed in-between minor versions (i.e. before Superset 4.1 is cut) you can just take it right out when you're done with development. If the flag gets released, removing it would constitute a breaking change, so we could remove it in ~6 months as part of Superset 5.0.

michael-s-molina:
Hi @fisjac. Both options are valid but if you opt for the feature flag (useful if multiple developers), you need to update the `FEATURE_FLAGS.md` file.

michael-s-molina (approved):

rusackas (approved):

PR #27779, opened 2024-03-29
chore: Fix/remove hardcode of admin role

Always-prog:
### SUMMARY
Just removed hardcode of admin role from perm utils and using `AUTH_ROLE_ADMIN` constant in the Superset config instead.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Always-prog:
@supersetbot orglabel

Always-prog:
How I can add orglabel?)

superset/config.py
```diff
@@ -314,8 +314,7 @@ def _try_json_readsha(filepath: str, length: int) -> str | None:
 # AUTH_REMOTE_USER : Is for using REMOTE_USER from web server
 AUTH_TYPE = AUTH_DB
 
-# Uncomment to setup Full admin role name
-# AUTH_ROLE_ADMIN = 'Admin'
+AUTH_ROLE_ADMIN = "Admin"
```
john-bodley, line 317: I don't think we need to uncomment this as it's the default in Flask-AppBuilder.

superset-frontend/src/dashboard/util/permissionUtils.ts
```diff
@@ -25,16 +25,14 @@ import {
 import { Dashboard } from 'src/types/Dashboard';
 import { findPermission } from 'src/utils/findPermission';
 
-// this should really be a config value,
-// but is hardcoded in backend logic already, so...
-const ADMIN_ROLE_NAME = 'admin';
+const ADMIN_ROLE_NAME = window.adminRole || 'Admin';
```
john-bodley, line 28:
```suggestion
const ADMIN_ROLE_NAME = window.adminRole;
```

john-bodley: See https://github.com/apache/superset/pull/27779/files#r1544723819 as to why this will _always_ be set.

superset/config.py
Always-prog, line 317:
```suggestion
# AUTH_ROLE_ADMIN = "Admin"
```

mistercrunch:
@supersetbot unlabel TechAudit-BI

mistercrunch:
@supersetbot orglabel

mistercrunch:
more about the bot here -> https://github.com/apache-superset/supersetbot . It even has a logo now 🤖 🤖 🤖 
Screenshot 2024-04-02 at 11 03 46 PM

superset/config.py
Always-prog: Commented, and replaced ' with ".

Always-prog:
Thank you @mistercrunch!

Always-prog:
@john-bodley Hi! Fixed code by your review. Please take a look!

Always-prog:
@rusackas Test cases has beed passed without rebase

rusackas:
The main thing I'm wondering about this PR now that I look more closely at the code, is if there's a better place to store this information. I'm a little nervous about storing bootstrap data on the window object, which opens up a new "junk drawer" pattern where people might store anything there. I'm not sure if we should have the Feature Flags there either, really. Is it possible to access this via the `getBootstrapData()` like we do so many other things in the codebase already. It looks like the role is indeed in there.

Always-prog:
@rusackas I see, ok. I have pushed commit with the fix :+1:, please take a look!

Always-prog:
I'll fix the frontend soon

rusackas:
Looks like this one just... never finished CI for some reason. Closing/reopening to kick start this now... let's see if we can get it merged!

superset-frontend/src/dashboard/util/permissionUtils.ts
```diff
@@ -24,17 +24,17 @@ import {
 } from 'src/types/bootstrapTypes';
 import { Dashboard } from 'src/types/Dashboard';
 import { findPermission } from 'src/utils/findPermission';
+import getBootstrapData from 'src/utils/getBootstrapData';
 
-// this should really be a config value,
-// but is hardcoded in backend logic already, so...
-const ADMIN_ROLE_NAME = 'admin';
+const bootstrapData = getBootstrapData();
+const ADMIN_ROLE_NAME = bootstrapData.common.conf.AUTH_ROLE_ADMIN || 'Admin';
```
mistercrunch: Mmmmh, seems `ADMIN_ROLE_NAME != AUTH_ROLE_ADMIN`. AUTH_ROLE_ADMIN probably exist do it can diverge from the global ADMIN_ROLE - assuming this exists. Maybe FAB exposes it somewhere? Would have to be carried to the frontend?

rusackas:
This is coming together nicely, @Always-prog. Pulling `AUTH_ROLE_ADMIN` from `getBootstrapData()` is exactly what I was hoping for. Mind reverting the seemingly accidental file-mode change on `shim.tsx` and I'll review/approve?

rusackas:
Thanks @Always-prog, the getBootstrapData() approach looks right. CI's failing though, `bootstrapData.common.conf` is undefined in the CrudThemeProvider test suite so `permissionUtils.ts` throws on import. Might be worth guarding that access and adding a test for the custom `AUTH_ROLE_ADMIN` path, Bito's thread above is flagging the same gap.

rusackas (approved):
~~CI's green now was green (those cypress/npm failures were infra flakes, unrelated to this diff — reran and they cleared). The `getBootstrapData()` guard and tests look solid. Thanks for sticking with this one, @Always-prog.~~

I stand corrected... CI still needs a little love, it seems... but LGTM otherwise.

PR #27788, opened 2024-03-29
feat(explore): Clear temporal filter value

kgabryje:
### SUMMARY
Currently when there is 1 temporal filter added and user tries to remove it, we display a message saying that there must be at least 1 temporal filter in the chart. We use the same flow regardless if the filter has a value or not, so simply clearing the filter requires many additional clicks.
This PR improves that behaviour:
- if there's more than 1 temporal filter, remove the filter on "x" click
- if there's 1 temporal filter and it has a value other than "No filter", set the value to "No filter" on "x" click
- if there's 1 empty temporal filter, display a warning message

https://github.com/apache/superset/assets/15073128/ba3897ab-3005-4e9e-9d9a-47e9ebd8e29c

### TESTING INSTRUCTIONS
1. Go to explore and add 2 or more temporal filters with some values
2. Delete all but the last one
3. Try to delete the last one - instead of being removed, the filter should clear its value
4. Try to delete it again - a warning message should be displayed

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/ControlPanelsContainer.tsx
+++ b/superset-frontend/src/explore/components/ControlPanelsContainer.tsx
@@ -52,7 +52,7 @@ import {
 } from '@superset-ui/chart-controls';
 import { useSelector } from 'react-redux';
 import { rgba } from 'emotion-rgba';
-import { kebabCase } from 'lodash';
+import { kebabCase, isEqual } from 'lodash';
 
 import Collapse from 'src/components/Collapse';
 import Tabs from 'src/components/Tabs';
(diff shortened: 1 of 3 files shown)


betodealmeida:
This seems a bit confusing... if we allow a temporal filter that says "No filter", why not allow having no filters? It seems like we're leaking an implementation detail to the user, where the API requires a filter (even if it says "No filter") so we require it in the UI.

It might be easier to allow having no filters, and simply sending "No filter" to the API if there are no filters, though ideally we'd fix this in the backend. Having the X do two different things — remove the filter or changing the filter time range depending on the number of filters — will confuse users.

cc: @yousoph

yousoph:
> This seems a bit confusing... if we allow a temporal filter that says "No filter", why not allow having no filters? It seems like we're leaking an implementation detail to the user, where the API requires a filter (even if it says "No filter") so we require it in the UI.

Yeah, agreed with what @michael-s-molina mentioned that it's not quite ideal but we need to know what the temporal column is. 

You're right that it is a little inconsistent when removing multiple filters but the idea was to remove a bit of friction from the user and reduce the frequency at which they see what feels like an error popup when the goal is to clear the filter values for now, until this flow is improved further.

eschutho (approved):

PR #27840, opened 2024-04-01
feat: new CSV upload form and API

dpgaspar:
### SUMMARY

Continuing the work on deprecating/removing server side rendered pages, this adds a new React/Antd Form and backend API for CSV uploads.

maintains all the current functionality and adds the following benefits:
Backend:
- better backend parameter validation.
- Aligned API with `api/v1` and command pattern.
- Improved error handling.

Frontend:
- Better looking ;)
- `schemas` is a drop down related with the databases dropdown
- Will initially try to infer the CSV header and load the recognised columns.

Screenshot 2024-04-12 at 14 08 00

- Columns to read is a multi select dropdown (source from the loaded CSV header)
- Columns as dates is as multi select dropdown (source from the loaded CSV header)
- Column index is dropdown (source from the loaded CSV header)
- Null values is a multi select dropdown pre populated
- delimiter is now a dropdown

https://github.com/apache/superset/assets/4025227/973897d4-3214-45ae-a0ef-4bc756d54c64

## Still todo on this PR:
- Deprecate or remove the old view?
- migrate the old view permission (depends on the previous item)
- Remove the old view from the menu
- After we settle on the UI add more frontend tests

## Before:
Screenshot 2024-04-05 at 15 42 51

## After
Screenshot 2024-04-12 at 16 59 11

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+      ]}
+      primaryButtonLoading={isLoading}
+      name="database"
+      data-test="csvupload-modal"
+      onHandledPrimaryAction={form.submit}
+      onHide={onClose}
+      primaryButtonName="Upload"
+      centered
+      show={show}
+      title={<h4>CSV Upload</h4>}
```
eschutho, line 394: maybe we can wrap this string in a translate function?

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+                    allowClear
+                  />
+                </StyledFormItem>
+                <p className="help-block">
+                  {t('Select a database to upload the file to')}
+                </p>
+              </Col>
+              <Col span={11}>
+                <StyledFormItem
+                  label={t('Table Name')}
```
eschutho, line 482: @yousoph Do you know if we have guidelines on capitalization around labels? Should everything be sentence case?

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+                  )}
+                </p>
+              </Col>
+            </Row>
+          </Collapse.Panel>
+          <Collapse.Panel
+            header={
+              <div>
+                <h4>{t('Columns')}</h4>
+                <p className="helper">{t('Adjust column settings.')}</p>
```
eschutho, line 650: Another small nit question on consistency in punctuation. Should we be using periods at the end of the helper text or no?

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+              <Col span={24}>
+                <StyledFormItem
+                  label={t('Column Data Types')}
+                  name="column_data_types"
+                >
+                  <Input aria-label={t('Column data types')} type="text" />
+                </StyledFormItem>
+                <p className="help-block">
+                  {t(
+                    'A dictionary with column names and their data types if you need to change the defaults. Example: {"user_id":"int"}. Check Python\'s Pandas library for supported data types.',
```
eschutho, line 747: @yousoph wdyt about linking to some documentation here so that we don't need to confuse people about what Pandas is?

superset-frontend/src/features/databases/CSVUploadModal/styles.ts
```diff
@@ -0,0 +1,77 @@
...
+    padding-bottom: ${theme.gridUnit * 2.5}px;
+    .anticon.ant-collapse-arrow {
+      top: calc(50% - ${anticonHeight / 2}px);
+    }
+    .helper {
+      color: ${theme.colors.grayscale.base};
+    }
+  }
+  h4 {
+    font-size: 16px;
```
eschutho, line 41: can we use the theme values here?

superset-frontend/src/pages/DatabaseList/index.tsx
```diff
@@ -230,6 +233,16 @@ function DatabaseList({
     {
       label: t('Upload file to database'),
       childs: [
+        {
+          label: t('New Upload CSV'),
+          name: 'Upload CSV file',
+          url: '#',
+          onClick: () => {
+            setCsvUploadModalOpen(true);
+          },
```
eschutho, line 242: is there a simple way to make this a button instead of an href?

superset/commands/database/csv_import.py
```diff
@@ -0,0 +1,164 @@
...
+                    "if_exists": self._options.get("already_exists", "fail"),
+                    "index": self._options.get("index_column"),
+                    "index_label": self._options.get("column_labels"),
+                },
+            )
+        except ValueError as ex:
+            raise DatabaseUploadFailed(
+                message=_(
+                    "Table already exists. You can change your"
+                    " 'if table already exists' strategy to append or replace."
```
eschutho, line 127: Is another option here for them to rename the table. Should we mention that @yousoph?

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
rusackas: Ahh yes, of course, the Pandas Library
image

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
dpgaspar: I think so, followed what we have on database modal

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+  t,
+} from '@superset-ui/core';
+import Modal from 'src/components/Modal';
+import Collapse from 'src/components/Collapse';
+import {
+  Upload,
+  AntdForm,
+  Col,
+  Row,
+  AntdButton as Button,
```
michael-s-molina, line 33: We shouldn't import a button directly from Ant Design but use one of our Button styles.

sfirke:
This looks great!  It gets me wondering about "upload columnar file to database" and "upload Excel file to database."  My users mostly use the Excel one.  Would some aspects of this overhaul be applicable to those upload options?

michael-s-molina (changes requested):
Thank you so much for working on this @dpgaspar. It's great that we continue to replace old functionality with improved interfaces. This first-pass review will only include some broad points. I can dive deeper into the code once we resolve these comments.

1 - I know it's not part of this PR, but could you use this opportunity to improve how the menu is displaying the disabled state? Right now, when hovering, it's showing the text as clickable, and when you click, nothing happens. In my opinion, disabled items should be always grey. @kasiazjc 

https://github.com/apache/superset/assets/70410625/805cbc68-7dac-40b7-98f3-aee4d4fe4668

2 - Can you match the modal design with the alerts/reports? 

I believe it's important to match the same width, dropdown arrow positions, etc. @kasiazjc 

Screenshot 2024-04-08 at 16 34 36
Screenshot 2024-04-08 at 16 34 59

3 - I believe these toggles would be better represented if we use the description directly instead of having a title + description.

Screenshot 2024-04-08 at 16 54 25
Screenshot 2024-04-08 at 16 58 23

It would also be nice to rearrange the fields order to group the toggles.

4 - The Select button should be one of our predefined buttons and it looks a little lost on the screen. I think once we change the modal width and rearrange the fields, it will look better. @kasiazjc 

Screenshot 2024-04-08 at 17 00 39

5 - How does loaded columns work with 30+ columns? @kasiazjc can we make sure we cover this case?

6 - About

> Deprecate or remove the old view?
> migrate the old view permission (depends on the previous item)
> Remove the old view from the menu

If the new interface provides the same functionality as the old one, I don't view it as a breaking change and would vote for removing old assets. We just need to be careful about permissions. I left a comment about that asking if they can be perceived as breaking changes.

It's nice to see you contributing frontend code! ❤️

superset/commands/database/csv_import.py
yousoph: Table already exists. You can change your 'if table already exists' strategy to append or replace or provide a different Table Name to use.

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
yousoph: That's a good idea... I didn't find a list we could reference in Pandas docs, is this something we need to come up with our own list of for our own docs?

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+                </p>
+              </Col>
+            </Row>
+          </Collapse.Panel>
+          <Collapse.Panel
+            header={
+              <div>
+                <h4>{t('File Settings')}</h4>
+                <p className="helper">
+                  {t('Adjust how the file will be imported.')}
```
yousoph, line 535: Suggested text change 
`Adjust how the file will be imported.` -> 
`Adjust how spaces, blank lines, null values are handled and other file wide settings.`

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
yousoph: Suggested text change 
`Adjust column settings.` -> 
`Adjust column settings such as specifying the columns to read, how duplicates are handled, column data types, and more.`

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
```diff
@@ -0,0 +1,805 @@
...
+                  )}
+                </p>
+              </Col>
+            </Row>
+          </Collapse.Panel>
+          <Collapse.Panel
+            header={
+              <div>
+                <h4>{t('Rows')}</h4>
+                <p className="helper">{t('Adjust row settings.')}</p>
```
yousoph, line 757: Suggested text change
`Adjust row settings.` -> 
`Set header rows and the number of rows to read or skip.`

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
mistercrunch: Screenshot 2024-04-08 at 4 55 38 PM

superset-frontend/src/features/databases/CSVUploadModal/index.tsx
dpgaspar: we endup capitalising all input form labels on CSS, so I don't think it matters

kasiazjc:
Thanks @michael-s-molina for the review and @dpgaspar for working on this! 
> 
> 1 - I know it's not part of this PR, but could you use this opportunity to improve how the menu is displaying the disabled state? Right now, when hovering, it's showing the text as clickable, and when you click, nothing happens. In my opinion, disabled items should be always grey. @kasiazjc

agreed, disabled items should always be grayed out and show 🚫 cursor instead of pointy finger, not sure if it can be easily done globally though?  

> 2 - Can you match the modal design with the alerts/reports?
> 
> I believe it's important to match the same width, dropdown arrow positions, etc. @kasiazjc
> 
> Screenshot 2024-04-08 at 16 34 36 Screenshot 2024-04-08 at 16 34 59

yes, all of the properties should be the same as in the alerts/reports. Additionally, each select/input should be in its own row (like in alerts modal)

> 4 - The Select button should be one of our predefined buttons and it looks a little lost on the screen. I think once we change the modal width and rearrange the fields, it will look better. @kasiazjc
> 
> Screenshot 2024-04-08 at 17 00 39

I think it should be tertiary button (outlined) so that it's still visible and easy to spot.

> 5 - How does loaded columns work with 30+ columns? @kasiazjc can we make sure we cover this case?
> 
@dpgaspar I think columns should be under the loaded file, with ability to uncollapse/collapse the columns (if more than fits in a row) so it would be something like this (perfect scenario):

image

I would also skip "basic" header (if it's not needed from you point of view) + descriptions under the fields in general section (for schema we can move it to the tooltip)

What do you think? Is it doable? (also do not mind things like labels not looking like what we have in the code)

michael-s-molina:
@kasiazjc @dpgaspar One thing I forgot to mention was about the loaded columns. Apart from their names, their types are equality important because they will inform users if they need to apply any transformation before loading them. It's the same principle we have when we create a dataset:

Screenshot 2024-04-09 at 13 10 03

We also have a Column Data Types field that allows type manipulation:

Screenshot 2024-04-09 at 13 13 14

@kasiazjc I think we can resolve both requirements if we present a similar table that allows users to edit their types as they would using the Column Data Types field.

dpgaspar:
> @kasiazjc @dpgaspar One thing I forgot to mention was about the loaded columns. Apart from their names, their types are equality important because they will inform users if they need to apply any transformation before loading them. It's the same principle we have when we create a dataset:
> 
> Screenshot 2024-04-09 at 13 10 03
> We also have a Column Data Types field that allows type manipulation:
> 
> Screenshot 2024-04-09 at 13 13 14
> @kasiazjc I think we can resolve both requirements if we present a similar table that allows users to edit their types as they would using the Column Data Types field.

On the loaded columns we can't infer the type yet, as for the Column Data Types field I would vote on improving that field on another PR

dpgaspar:
@michael-s-molina @kasiazjc I've addressed basically all your comments:

@michael-s-molina 
1 - Fixed the display disabled at the component level
2 - fixed the width, did not fix the dropdown icons etc because I want to avoid customizing the colapse panels (so it follows database modal)
3 - done
4 - done
5 - Followed @kasiazjc comment

Screenshot 2024-04-10 at 14 43 23

6 - Agree removed the code

superset-frontend/src/pages/DatabaseList/index.tsx
dpgaspar: could not find an easy way without refactoring the component

superset/migrations/versions/2024-04-08_15-43_5ad7321c2169_mig_new_csv_upload_perm.py
```diff
@@ -0,0 +1,85 @@
...
+PVM_MAP = {
+    Pvm("CsvToDatabaseView", "can_this_form_post"): (
+        Pvm("Database", "can_csv_upload"),
+    ),
+    Pvm("CsvToDatabaseView", "can_this_form_get"): (Pvm("Database", "can_csv_upload"),),
+}
+
+
+def do_upgrade(session: Session) -> None:
+    add_pvms(session, NEW_PVMS)
```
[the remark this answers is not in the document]
dpgaspar: I guess you could say it's a breaking change, on `UPDATING.md` should I add a note to `## Next` or `### Breaking changes` ?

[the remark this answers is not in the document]
michael-s-molina: The problem is that if this PR contains breaking changes, it will be blocked until the next breaking window. There are two ways to avoid this:
- Make it backward compatible
- Remove the permission changes from this PR and do them during the breaking window

michael-s-molina (changes requested):
Thank you for the improvements @dpgaspar. Here are the comments for the second review:

> 1 - Fixed the display disabled at the component level

The cursor is correctly displaying the forbidden sign but the items are still becoming blue when hovered which I don't think should be the case because it gives the impression that they leave the forbidden state and are now clickable. I also noticed that the first item is not disabled.
Screenshot 2024-04-11 at 09 43 34  | Screenshot 2024-04-11 at 09 43 45
<hr>

> 2 - fixed the width, did not fix the dropdown icons etc because I want to avoid customizing the colapse panels (so it follows database modal)

@kasiazjc @eschutho @dpgaspar We just need to decide which design is the correct one and apply it consistently: 

<table>
<tr>
<td>
Screenshot 2024-04-11 at 09 36 38
</td>
<td>
Screenshot 2024-04-11 at 09 37 14
</td>
</tr>
</table>

We need to match spacing between header and description, borders, arrow position, the X color. These modals should all be using the same component to render the collapse panels.
<hr>

> 3 - done

Would be possible to move all toggles to last, grouping them, to improve the design? I think you can do this for all collapse panels.

Screenshot 2024-04-11 at 09 52 38

<hr>

> 5 - Followed @kasiazjc comment

The columns are breaking instead of overflowing to the next line. I also think we need some space between the lines.

Screenshot 2024-04-11 at 10 02 59

@kasiazjc Shouldn't we use the +N pattern that we use for chart's list and the Select component instead of the black arrow?

Screenshot 2024-04-11 at 10 09 05

<hr>

7 - I think these should be in just one row and the tooltip icon needs to be vertically aligned.

Screenshot 2024-04-11 at 09 56 07

<hr>

> On the loaded columns we can't infer the type yet, as for the Column Data Types field I would vote on improving that field on another PR

I'm ok with doing that in another PR. Inferring column types would be a great addition.

dpgaspar:
> Thank you for the improvements @dpgaspar. Here are the comments for the second review:
> 
> > 1 - Fixed the display disabled at the component level
> 
> The cursor is correctly displaying the forbidden sign but the items are still becoming blue when hovered which I don't think should be the case because it gives the impression that they leave the forbidden state and are now clickable. I also noticed that the first item is not disabled. Screenshot 2024-04-11 at 09 43 34 | Screenshot 2024-04-11 at 09 43 45
> 
> > 2 - fixed the width, did not fix the dropdown icons etc because I want to avoid customizing the colapse panels (so it follows database modal)
> 
> @kasiazjc @eschutho @dpgaspar We just need to decide which design is the correct one and apply it consistently:
> 
> Screenshot 2024-04-11 at 09 36 38	Screenshot 2024-04-11 at 09 37 14
> We need to match spacing between header and description, borders, arrow position, the X color. These modals should all be using the same component to render the collapse panels.
> 
> > 3 - done
> 
> Would be possible to move all toggles to last, grouping them, to improve the design? I think you can do this for all collapse panels.
> 
> Screenshot 2024-04-11 at 09 52 38
> > 5 - Followed @kasiazjc comment
> 
> The columns are breaking instead of overflowing to the next line. I also think we need some space between the lines.
> 
> Screenshot 2024-04-11 at 10 02 59
> @kasiazjc Shouldn't we use the +N pattern that we use for chart's list and the Select component instead of the black arrow?
> 
> Screenshot 2024-04-11 at 10 09 05
> 7 - I think these should be in just one row and the tooltip icon needs to be vertically aligned.
> 
> Screenshot 2024-04-11 at 09 56 07
> > On the loaded columns we can't infer the type yet, as for the Column Data Types field I would vote on improving that field on another PR
> 
> I'm ok with doing that in another PR. Inferring column types would be a great addition.

1 - The first item is disabled, I'll remove the hovered blue.
Screenshot 2024-04-11 at 15 26 51

2 - Totally agree, I think that the correct design should be imposed at the component level, then we can remove most of the custom `styles.ts`. So, I don't think that doing more style tweaking on this PR makes sense

3 - Sure!

5 - Will do, @kasiazjc do you agree?

7 - Will do.

Thank you for the great reviews!

michael-s-molina:
> Agreed @dpgaspar. We can definitely do a separate PR just to make all these modals look the same and use the same component.

@dpgaspar Assuming this 👆🏼, I think the only design changes left would be to fix the spacing between inputs 

<table>
<tr>
<td>
Screenshot 2024-04-12 at 10 14 43
</td>
<td>
Screenshot 2024-04-12 at 10 14 28
</td>
</tr>
</table>

and make the tooltip icons vertically aligned.

Screenshot 2024-04-12 at 10 19 02

dpgaspar:
> > Agreed @dpgaspar. We can definitely do a separate PR just to make all these modals look the same and use the same component.
> 
> @dpgaspar Assuming this 👆🏼, I think the only design changes left would be to fix the spacing between inputs
> 
> Screenshot 2024-04-12 at 10 14 43	Screenshot 2024-04-12 at 10 14 28
> and make the tooltip icons vertically aligned.
> 
> Screenshot 2024-04-12 at 10 19 02

ok, done!

eschutho (approved):
LGTM, thanks @dpgaspar!

michael-s-molina (approved):
Thank you for all the work @dpgaspar and for addressing the comments! This is a great improvement to the project 👏🏼

PR #27843, opened 2024-04-01
chore: Default to engine specification regarding using wildcard

john-bodley:
### SUMMARY

The `BaseEngineSpec.select_star` method provides an option (with a default) with regards to whether or not one should use a wildcard (`show_cols=False`) or enumerate the columns (`show_cols=True`) when generating a SQL statement for fetching sample records from a dataset.

Some of the underlying engines—Hive, Presto, BigQuery—override this method including the default for the `show_cols` function argument. When creating the SQL statement rather than forcing `show_cols=True` we should instead fallback to the engine specific preference.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -871,7 +871,7 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
         self.incr_stats("init", self.select_star.__name__)
         try:
             result = database.select_star(
-                table_name, schema_name, latest_partition=True, show_cols=True
+                table_name, schema_name, latest_partition=True
             )
         except NoSuchTableError:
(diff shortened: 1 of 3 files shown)


john-bodley:
After further research I opted to close this. There’s some weird patterns we use, especially given that the function arguments seem over prescribed.

betodealmeida (approved):
👍🏼

michael-s-molina (approved):
Nice. Thanks for fixing this @john-bodley.

PR #27846, opened 2024-04-01
fix: alembic's 'superset db migrate' fails with CompileError

mistercrunch:
When running on the latest `master`
running a simple `superset db migrate` within `docker-compose exec superset bash` to generate a new
database migration fails with error

```
sqlalchemy.exc.CompileError: PostgreSQL ENUM type requires a name.
```

This PR addresses the issue by giving a name to that particular enum.

From my understanding, this particular enum wasn't created in
a given migration script, so it may not be enforced at the database
level, probably by choice/design. Having the name here simply
allows for alembic's migrate subcommand to run.

### DIFF
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -2055,7 +2055,10 @@ class RowLevelSecurityFilter(Model, AuditMixinNullable):
     name = Column(String(255), unique=True, nullable=False)
     description = Column(Text)
     filter_type = Column(
-        Enum(*[filter_type.value for filter_type in utils.RowLevelSecurityFilterType])
+        Enum(
+            *[filter_type.value for filter_type in utils.RowLevelSecurityFilterType],
+            name="filter_type_enum",
+        ),
     )
     group_key = Column(String(255), nullable=True)
     roles = relationship(


superset/connectors/sqla/models.py
```diff
@@ -2055,7 +2055,10 @@ class RowLevelSecurityFilter(Model, AuditMixinNullable):
     name = Column(String(255), unique=True, nullable=False)
     description = Column(Text)
     filter_type = Column(
-        Enum(*[filter_type.value for filter_type in utils.RowLevelSecurityFilterType])
+        Enum(
+            *[filter_type.value for filter_type in utils.RowLevelSecurityFilterType],
+            name="filter_type_enum",
+        ),
```
dpgaspar, line 2061: implemented here as a varchar: https://github.com/apache/superset/blob/master/superset/migrations/versions/2020-09-15_18-22_e5ef6828ac4e_add_rls_filter_type_and_grouping_key.py
probably by choice

dpgaspar (approved):
implemented as a varchar here: https://github.com/apache/superset/blob/master/superset/migrations/versions/2020-09-15_18-22_e5ef6828ac4e_add_rls_filter_type_and_grouping_key.py

probably by design

mistercrunch:
@john-bodley this is an issue with `superset db migrate` (not `superset db upgrade`) which auto-creates new migration. AFAIK this is done manually when there's a need to create new migrations, never in CI.

Also note that auto-migrate has picks up the diff between the declared model and what's in the database post applying all migrations, and currently there are quite a bit of discrepancies that could be good to address.

PR #27847, opened 2024-04-01
feat(db): Adding DB_SQLA_URI_VALIDATOR

craig-rueda:
### SUMMARY
Quick PR that adds a new callback hook, `DB_SQLA_URI_VALIDATOR ` which can be optionally set in Superset configs. When set, it should point to a callable that is invoked in order to validate db_engine_spec's URIs.

**Example:**
```python
def DB_SQLA_URI_VALIDATOR(sqlalchemy_uri: URL):
    if not <some condition>:
        raise Exception("URI invalid")
```

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -43,6 +43,7 @@ from flask_appbuilder.security.manager import AUTH_DB
 from flask_caching.backends.base import BaseCache
 from pandas import Series
 from pandas._libs.parsers import STR_NA_VALUES  # pylint: disable=no-name-in-module
+from sqlalchemy.engine.url import URL
 from sqlalchemy.orm.query import Query
 
 from superset.advanced_data_type.plugins.internet_address import internet_address
@@ -1214,7 +1215,7 @@ DB_CONNECTION_MUTATOR = None
 #       if not <some condition>:
 #           raise Exception("URI invalid")
 #
-DB_ENGINE_URI_VALIDATOR = None
+DB_SQLA_URI_VALIDATOR: Callable[[URL], None] | None = None
 
 
 # A function that intercepts the SQL to be executed and can alter it.
--- a/superset/config.py
+++ b/superset/config.py
@@ -1206,6 +1206,17 @@ DASHBOARD_TEMPLATE_ID = None
 DB_CONNECTION_MUTATOR = None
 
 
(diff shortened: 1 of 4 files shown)


mistercrunch (commented):
2 thoughts:
- about naming `DB_SQLA_URI_VALIDATOR` might be a slightly better name as it validates a sqlalchemy URI more than an "engine" uri
- I thought we could have a more generic `DB_VALIDATOR` instead or in complement to this so that someone could look at other `DatabaseConnection` attributes too, as opposed to the uri only. Password complexity, extra attributes or anything else.

craig-rueda:
Aren't all of the connection attributes captured in the parsed url? Also, this is intended to be invoked pre-connect, so we likely wouldn't want to perform validations on a "connection".

mistercrunch:
I think the bigger exception is the way we do BigQuery for instance cramming many things into a blob, but unclear how common/uncommon it is. This probably a good starting point. Can add more hooks if/as needed.

betodealmeida:
Yeah, in general there are two parts:

engine = create_engine(uri, **params)

Both are needed for a full connection, since in some cases the params are too big to build a URI (like BigQuery credentials).

superset/config.py
```diff
@@ -1206,6 +1206,17 @@ def CSV_TO_HIVE_UPLOAD_DIRECTORY_FUNC(  # pylint: disable=invalid-name
...
 
+# A callable that is invoked for every invocation of DB Engine Specs
+# which allows for custom validation of the engine URI.
+# See: superset.db_engine_specs.base.BaseEngineSpec.validate_database_uri
+# Example:
+#   def DB_ENGINE_URI_VALIDATOR(sqlalchemy_uri: URL):
+#       if not <some condition>:
+#           raise Exception("URI invalid")
+#
+DB_ENGINE_URI_VALIDATOR = None
```
dpgaspar, line 1217: nit: `DB_ENGINE_URI_VALIDATOR: Callable[[URL], None] | None = None`

dpgaspar (approved):

mistercrunch (approved):

john-bodley:
@craig-rueda out of interest why is this surfacing as an issue now? Did we not gracefully handle SQLAlchemy URI errors in the past?

PR #27849, opened 2024-04-02
feat: Slack Avatar integration

mistercrunch:
### SUMMARY
Our current Avatars are a little dry, using a limited variety of colors and showing the user's initials. The app could use a little more personality. Now given we already have a slack integration and administrators can configure their `SLACK_API_TOKEN`, I was thinking we can use this integration to get Slack's avatars integrated in Superset. Given the popularity of Slack and users familiarity with the avatars there, I thought it'd be neat to bring a bit more color in Superset.

Note that I'm also introducing the config `SLACK_ENABLE_AVATARS` which is `False` by default, and added a note to `UPDATING.md` for release managers to switch it on if desired.

The approach here is centered around a new endpoint `/api/v1/user/{id}/avatar.png`, which fetches from a new column in
UserAttribute (our one-to-one extension to FAB's User model) name `avatar_url`, and redirects (301) to the destination. When loading for the first time, if information is empty and Slack is configured, we call Slack for the avatar URL, which has a secret location, but serves the image from outside the authentication part of their API once known. If the feature is off or no avatar url is set for the user, we simply return a `204` (no-content), preventing from logging errors or anything negative as this is expected. 

The `avatar_url` field could be used programmatically in environments to serve other urls, but at this time I did not implement a GUI for users to self-set their avatar for instance. I think in most cases people will want to integrate with some external service (MSFT Team, Discord, ...). The approach I took here could be expanded to work with Gravatar for instance. 

#### About Perf
This approach while unconventional (is it?) should perform very well.  Say in an environment with 1000 users, the Slack API should trigger once-ish for each user in the database to fill the `avatar_url` for that user (1000 api hits). For each user navigating the website, a single hit to `/api/v1/users/{some_user_id}/avatar.png` where the database is looked up is sufficient (up to 1M hits, but super low-key endpoint), after which the permanent redirect is cached client-side. The 204 should get cached like any other request following the configured cache-control in place.

Screenshot 2024-04-01 at 10 07 03 PM

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -1330,6 +1330,11 @@ EMAIL_REPORTS_CTA = "Explore in Superset"
 SLACK_API_TOKEN: Callable[[], str] | str | None = None
 SLACK_PROXY = None
 
+# Whether Superset should use Slack avatars for users.
+# If on, you'll want to add "https://avatars.slack-edge.com" to the list of allowed
+# domains in your TALISMAN_CONFIG
+SLACK_ENABLE_AVATARS = False
+
 # The webdriver to use for generating reports. Use one of the following
 # firefox
 #   Requires: geckodriver and firefox installations
@@ -1454,6 +1459,7 @@ TALISMAN_CONFIG = {
             "data:",
             "https://apachesuperset.gateway.scarf.sh",
             "https://static.scarf.sh/",
+            # "https://avatars.slack-edge.com", # Uncomment when SLACK_ENABLE_AVATARS is True
         ],
         "worker-src": ["'self'", "blob:"],
         "connect-src": [
@@ -1483,6 +1489,7 @@ TALISMAN_DEV_CONFIG = {
             "data:",
             "https://apachesuperset.gateway.scarf.sh",
             "https://static.scarf.sh/",
+            "https://avatars.slack-edge.com",
         ],
         "worker-src": ["'self'", "blob:"],
         "connect-src": [
--- a/superset/migrations/versions/2024-04-11_00-49_bbf146925528_.py
+++ b/superset/migrations/versions/2024-04-11_00-49_bbf146925528_.py
@@ -1,0 +1,38 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""empty message
+
+Revision ID: bbf146925528
+Revises: ('678eefb4ab44', 'c22cb5c2e546')
+Create Date: 2024-04-11 00:49:51.592325
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "bbf146925528"
+down_revision = ("678eefb4ab44", "c22cb5c2e546")
+
+import sqlalchemy as sa
+from alembic import op
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    pass
--- a/superset/migrations/versions/2024-04-15_16-06_0dc386701747_.py
+++ b/superset/migrations/versions/2024-04-15_16-06_0dc386701747_.py
@@ -1,0 +1,38 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""empty message
+
+Revision ID: 0dc386701747
+Revises: ('5ad7321c2169', 'bbf146925528')
+Create Date: 2024-04-15 16:06:29.946059
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "0dc386701747"
+down_revision = ("5ad7321c2169", "bbf146925528")
+
+import sqlalchemy as sa
+from alembic import op
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    pass
--- a/superset/models/user_attributes.py
+++ b/superset/models/user_attributes.py
@@ -14,8 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
 from flask_appbuilder import Model
-from sqlalchemy import Column, ForeignKey, Integer
+from sqlalchemy import Column, ForeignKey, Integer, String
 from sqlalchemy.orm import relationship
 
 from superset import security_manager
@@ -39,6 +40,6 @@ class UserAttribute(Model, AuditMixinNullable):
     user = relationship(
         security_manager.user_model, backref="extra_attributes", foreign_keys=[user_id]
     )
-
     welcome_dashboard_id = Column(Integer, ForeignKey("dashboards.id"))
     welcome_dashboard = relationship("Dashboard")
+    avatar_url = Column(String(100))
--- a/superset/utils/slack.py
+++ b/superset/utils/slack.py
@@ -1,0 +1,59 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+from flask import current_app
+from slack_sdk import WebClient
+
+
+class SlackClientError(Exception):
+    pass
+
+
+def get_slack_client() -> WebClient:
+    token: str = current_app.config["SLACK_API_TOKEN"]
+    if callable(token):
+        token = token()
+    return WebClient(token=token, proxy=current_app.config["SLACK_PROXY"])
+
+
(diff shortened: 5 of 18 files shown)


.gitignore
```diff
@@ -110,5 +110,6 @@ release.json
 messages.mo
 
 docker/requirements-local.txt
+docker/.env-local
```
mistercrunch, line 113: Note: I was working with `docker-compose` and had to set my `SLACK_API_KEY` somewhere, and was afraid that I'd accidentally push it to the repo while working in `docker./env` so I added a new **gitignored** docker/.env-local

docker/pythonpath_dev/superset_config.py
```diff
@@ -99,6 +99,7 @@ class CeleryConfig:
 WEBDRIVER_BASEURL_USER_FRIENDLY = WEBDRIVER_BASEURL
 
 SQLLAB_CTAS_NO_LIMIT = True
+SLACK_API_TOKEN = os.getenv("SLACK_API_TOKEN")
```
mistercrunch, line 102: It could be nice to have a gitignored `superset_config_local.py` (similar to `.env-local` above)

superset/dashboards/api.py
```diff
@@ -179,6 +179,9 @@ def ensure_thumbnails_enabled(self) -> Optional[Response]:
         "owners.id",
         "owners.first_name",
         "owners.last_name",
+        "owners.email",
+        "owners.id",
+        # "owners.user_attributes.avatar_url",
```
mistercrunch, line 184: @dpgaspar I think I might need your help to get this to work or to find some workaround. Here I need a handle on `avatar_url`  which I stored in the `UserAttributes` model, which is one-to-one with FAB's `User` model, and looks like there's some sweet magic here that's a bit beyond me - at least for now. 

I considered other approaches, like extending FAB's User model, but I don't think we have a handle on it in Superset RN. Other approach would be to support multiple dots here.

Curious on your thoughts

superset/config.py
```diff
@@ -1328,6 +1328,8 @@ def EMAIL_HEADER_MUTATOR(  # pylint: disable=invalid-name,unused-argument
 # Slack API token for the superset reports, either string or callable
 SLACK_API_TOKEN: Callable[[], str] | str | None = None
 SLACK_PROXY = None
+# Whether Superset should use Slack avatars for users
+SLACK_ENABLE_AVATARS = True
```
rusackas, line 1332: At one point we discussed and _might have completed_ moving all boolean configs to Feature Flags. If that's going to be a policy (which I don't think has _official_ consensus) then this should be too.  CC @eschutho and @michael-s-molina in case they remember differently. I'm just a fan of consistency, in any case ;)

superset/config.py
```diff
@@ -1453,6 +1455,7 @@ def EMAIL_HEADER_MUTATOR(  # pylint: disable=invalid-name,unused-argument
             "data:",
             "https://apachesuperset.gateway.scarf.sh",
             "https://static.scarf.sh/",
+            "https://avatars.slack-edge.com",
```
rusackas, line 1458: I'm not sure anymore if the ASF will care about this or not.

superset/config.py
michael-s-molina: We need to have a discussion about feature flags in a specific meeting. I was researching a little bit about feature flags best practices and [this](https://docs.getunleash.io/topics/feature-flags/short-lived-feature-flags) caught my attention. Specifically, the following:

> The most common use case for feature flags is to protect new functionality. That means that when the roll-out of new functionality is complete, the feature flag should be removed from the code and archived.

> Feature flags should not be used for static application configuration.

If we followed a similar model, we would never have the categorization we currently have in [FEATURE_FLAGS.md](https://github.com/apache/superset/blob/master/RESOURCES/FEATURE_FLAGS.md). We would just assume that all feature flags should be removed once the feature is complete and what we have as `STABLE` would actually be application configurations, and ideally [changeable at runtime](https://docs.getunleash.io/topics/feature-flags/runtime-control). What I like about this model is the clarity/simplicity.

mistercrunch: Yes there's a fine line between configuration setting and feature flag. Happy to move to FF if we think it belong there

superset/models/user_attributes.py
```diff
@@ -39,6 +41,22 @@ class UserAttribute(Model, AuditMixinNullable):
     user = relationship(
         security_manager.user_model, backref="extra_attributes", foreign_keys=[user_id]
     )
-
     welcome_dashboard_id = Column(Integer, ForeignKey("dashboards.id"))
     welcome_dashboard = relationship("Dashboard")
+    avatar_url = Column(String(100))
+
+
+class OwnershipMixin:  # pylint: disable=too-few-public-methods
```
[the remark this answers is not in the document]
mistercrunch: I meant to add a note about this, it's likely to be useful in the future if/when we decide to incorporate more UserAttributes. Thought it may come handy. Maybe I add a `NOTE:`?

superset/config.py
mistercrunch: They shouldn't, especially as these are just default conf settings, and we don't store anything. Personally can't think of a rule anyone would be breaking by turning this on, plus it's off by default.

tests/integration_tests/users/api_tests.py
```diff
@@ -62,3 +63,23 @@ def test_get_me_anonymous(self, mock_g):
...
+        assert response.status_code == 404  # Assuming no user found leads to 404
+
+    @patch("superset.views.users.api.get_user_avatar", return_value=AVATAR_URL)
+    def test_avatar_with_valid_user(self, mock):
+        self.login(ADMIN_USERNAME)
+
+        response = self.client.get("/api/v1/user/1/avatar.png")
+        mock.assert_called_once_with("admin@fab.org")
+        assert response.status_code == 301
+        assert response.headers["Location"] == self.AVATAR_URL
```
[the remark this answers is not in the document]
mistercrunch: Maybe I can just `git mv` the file

superset/views/users/api.py
```diff
@@ -93,3 +97,85 @@ def get_my_roles(self) -> Response:
...
+            # fetch from the one-to-one relationship
+            if len(user.extra_attributes) > 0:
+                avatar_url = user.extra_attributes[0].avatar_url
+
+            should_fetch_slack_avatar = app.config.get(
+                "SLACK_ENABLE_AVATARS"
+            ) and app.config.get("SLACK_API_TOKEN")
+            if not avatar_url and should_fetch_slack_avatar:
+                # Fetching the avatar url from slack
+                avatar_url = get_user_avatar(user.email)
```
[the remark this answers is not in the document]
mistercrunch: I like the idea of adding a configuration hook here, though I'd have to define the interface prematurely (say guessing `def GET_USER_AVATAR_FUNC(user) -> str` ), and probably think of support for non `png` file types. I'd punt and say we shouldn't prematurely generalize the feature. Makes me think the main issue with the PR is the `.png` assumption. I have to test whether the jpg redirect happens to work or whether I can make it work.

tests/integration_tests/users/api_tests.py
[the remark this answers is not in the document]
mistercrunch: Here I just followed what was there before, but may be an opportunity to improve this, let me see.

superset/config.py
rusackas: I think it's that when any user traffic hits a 3rd party service, they could get IPs/analytics from it, in which case they'll want to have a DPA in place. They might already have one with Slack, which would make this a moot point, but if not, we can probably get one one, if this is even a concern.

mistercrunch: Think I should tie this to the feature flag?

mistercrunch: Ok I made SLACK_ENABLE_AVATARS = False by default (at least for now), and commented the `TALISMAN_CONFIG` changes out, with a note about de-commenting the line if you set SLACK_ENABLE_AVATARS to True. Altered the UPDATING.md comment to reflect that.

superset/utils/slack.py
```diff
@@ -0,0 +1,33 @@
...
+    token: str = current_app.config["SLACK_API_TOKEN"]
+    if callable(token):
+        token = token()
+    return WebClient(token=token, proxy=current_app.config["SLACK_PROXY"])
+
+
+def get_user_avatar(email: str, client: WebClient = None) -> str:
+    client = client or get_slack_client()
+    response = client.users_lookupByEmail(email=email)
+    avatar_url = response.data.get("user").get("profile").get("image_192")
```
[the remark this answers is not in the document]
mistercrunch: ok I improved the error handling here

superset/config.py
mistercrunch: Made all defaults more conservative,  `SLACK_ENABLE_AVATARS` is False by default and users have to alter their `TALISMAN_CONFIG` manually, resolving this thread.

tests/integration_tests/users/api_tests.py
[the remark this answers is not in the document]
mistercrunch: @dpgaspar I ended up creating a very simple DAO and writing basic unit tests for it, and expanded on the api tests here.

superset/views/users/api.py
```diff
@@ -93,3 +97,85 @@ def get_my_roles(self) -> Response:
...
+                    db.session.query(UserAttribute)
+                    .filter_by(user_id=user_id_as_int)
+                    .first()
+                    or
+                    # create new
+                    UserAttribute(user_id=user_id_as_int)
+                )
+                user_attrs.avatar_url = avatar_url
+                db.session.add(user_attrs)
+                db.session.commit()
```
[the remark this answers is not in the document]
mistercrunch: This is the only element I did not fully address because it's effectively the feature here. Note that mutations will be very very sparse here. Only if:
- feature is on (SLACK_ENABLE_AVATAR + SLACK_API_TOKEN)
- it's the **very first time** the avatar is requested for a given target user in the whole environment, updating the UserAttributes.avatar_url once with a short string

In all normal day-to-day
- upon a new browser requesting an avatar, the GET does as a normal get (a one cell database lookup)
- 301 gets cached, target image gets cached

mistercrunch:
@dpgaspar I addressed pretty much all feedback. Personally prefer avoiding the CELERY_BEAT configuration, but open to consider that approach too if we decide it's preferable.

dpgaspar (approved):

superset/migrations/versions/2024-04-11_00-49_bbf146925528_.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""empty message
+
+Revision ID: bbf146925528
```
sadpandajoe, line 19: What does this migration do? See that upgrade and downgrade are just pass

superset/migrations/versions/2024-04-15_16-06_0dc386701747_.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""empty message
+
+Revision ID: 0dc386701747
```
sadpandajoe, line 19: Same as above, what would this migration do?

mistercrunch: Hey, sorry about the confusion, but in alembic the migration "chain" can split and you end up with multiple `heads`, typically when you rebase and there's been some other migration. This migration merges the 2 heads. In theory it's possible to do the equivalent of a rebase, but that requires a bit of work. In this case I "git rebased" twice during the lifecycle of this PR and had to merge the heads 2 times, leaving this 2 empty-ish migrations.

mistercrunch: more about it here -> https://alembic.sqlalchemy.org/en/latest/branches.html

PR #27857, opened 2024-04-02
feat(dashboard): Add metadata bar to the header

justinpark:
### SUMMARY
As the dashboard currently lacks metadata information such as owners and last modified time, users have to navigate to edit mode and click "edit properties" to access this information. Unfortunately, there is no way to check the last modified time at all.
To address this issue, this commit integrates the metadata bar UI from the chart header into the dashboard header and updates the API properties accordingly. This enhancement will provide users with convenient access to the metadata information they need.

Screenshot 2024-04-02 at 1 24 09 PM

https://github.com/apache/superset/assets/1392866/2956ea12-1e23-42ee-bd6c-cec8cead739e

### TESTING INSTRUCTIONS
Go to a dashboard and then check the header

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


mistercrunch:
Do we / should we hide this in print mode (standalone=true) or in an embedded context? I'm not sure what the behavior should be or what the options settings should be and generally think the more information the better, but I'm unclear if everyone would agree.

rusackas:
+1 on the more the merrier. Knowing how fresh the dashboard is (the edit timestamp) is valuable even in an embedded context. The owners might not be something you want to see in an embedded context, but if anyone barks about that, I think that's also addressable — probably by means of a new config option (prop?) in the embedded component, just like you'd tweak other features (filter bar collapsing, etc)

rusackas (approved):
Code LGTM. Further discussion about the feature's utility in embedding may be warranted (CC @kasiazjc @yousoph) but otherwise, I'm onboard :D

superset/models/dashboard.py
```diff
@@ -250,6 +250,16 @@ def changed_by_name(self) -> str:
...
 
+    @property
+    def created_by_name(self) -> str:
+        if not self.created_by:
+            return ""
+        return str(self.created_by)
+
+    @property
+    def owners_by_name(self) -> list[str]:
+        return [owner.get_full_name() for owner in self.owners]
```
michael-s-molina, line 261: This looks like more of frontend code than a new model property. If you check other instances of the Metadata component, this is being handled on the frontend as we may have different representations of the possible values.

michael-s-molina (commented):
Thank you for the improvement @justinpark.

justinpark:
> Do we / should we hide this in print mode (standalone=true) or in an embedded context? I'm not sure what the behavior should be or what the options settings should be and generally think the more information the better, but I'm unclear if everyone would agree.

Since we currently do not hide any items in the header in standalone mode, let's keep the metadata bar as it is for the time being.

superset/models/dashboard.py
justinpark: Done

superset/models/dashboard.py
```diff
@@ -242,6 +242,12 @@ def changed_by_name(self) -> str:
             return ""
         return str(self.changed_by)
 
+    @property
```
michael-s-molina, line 245: I think this is the same case as the owners and should be handled by the frontend.

justinpark: @michael-s-molina This case is different. The property 'created_by' does not currently exist, therefore it is being added as a new property. The interface used for this addition is 'changed_by_name', which closely resembles the desired shape.

justinpark: Sounds good

michael-s-molina (approved):

nemccarthy:
> Do we / should we hide this in print mode (standalone=true) or in an embedded context? I'm not sure what the behavior should be or what the options settings should be and generally think the more information the better, but I'm unclear if everyone would agree.

This is a great call out and we 100% need this. In the embedded context why would someone need to know about metadata thats really useful for a superset user? 

It should be an option and totally breaks embedding for us as we do not want to show this to end users in the embedded use case.

PR #27858, opened 2024-04-02
chore(sql_parse): Provide more meaningful SQLGlot errors

john-bodley:
### SUMMARY

https://github.com/apache/superset/pull/26767 replaced the non-validating `sqlparse` parser with the pseudo-validating `SQLGlot` parser to aid with SQL parsing. The main use case for SQL parsing—from a security perspective—is to identify/extract the underlying tables/views referenced in the statement and checking whether the user is authorized to access them.

Previously, given the non-validating nature of `sqlparse`, virtually all SQL statements could be parsed—irrespective of whether they were valid—which meant that any syntax errors were reported by the underlying engine in a pseudo-meaningful manner. Now however, given that `SQLGlot` is more opinionated, syntax errors are detected by `SQLGlot` during the table/view extraction phase meaning that the query is never run and thus the engine's parser is never run resulting in a non-actionable error, i.e., it merely stated that the SQL could not be parsed. This is especially problematic in SQL Lab where freeform SQL is the norm and can be rather lengthy.

The TL;DR is now SQL statements which contain a syntax error are _likely_ going to be captured by `SQLGlot` upstream of being executed by the engine.

This PR proposes leveraging the messaging from `SQLGlot`'s `ParseError` or `TokenError` ([example](https://github.com/tobymao/sqlglot/blob/main/README.md#parser-errors)) to provide a more meaningful (and thus actionable) error message to the user.

#### PREVIOUS 

Prior to https://github.com/apache/superset/pull/26767 `sqlparse` was able to "parse" the SQL statement—missing an `AND` after the `key = 'foobar'` clause—which was then executed by the Trino engine (given that the user had access to the `datainfra.one_row` table) resulting in the following Trino error,

Screenshot 2024-04-02 at 10 47 11 AM

which detected the error was around the `1` token.

#### CURRENT

After https://github.com/apache/superset/pull/26767 `SQLGlot` was (rightfully) unable to parse the SQL statement and thus could not extract the underlying tables which halted the execution flow. The resulting error message isn't overly helpful as it doesn't inform the user where the actual syntax error is.

Screenshot 2024-04-02 at 12 09 58 PM

#### PROPOSED

Leveraging `SQLGlot`'s `ParseError` et al. exceptions to provide a more meaningful error message,

Screenshot 2024-04-02 at 3 42 47 PM

which correctly detected the error was around the `1` token. The messaging is loosely based on how MySQL presents errors.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -752,11 +752,21 @@ class ParsedQuery:
             statements = parse(self.stripped(), dialect=self._dialect)
         except SqlglotError as ex:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
-            dialect = self._dialect or "generic"
+
+            message = (
+                "Error parsing near '{highlight}' at line {line}:{col}".format(  # pylint: disable=consider-using-f-string
+                    **ex.errors[0]
+                )
+                if isinstance(ex, ParseError)
+                else str(ex)
+            )
+
             raise SupersetSecurityException(
                 SupersetError(
                     error_type=SupersetErrorType.QUERY_SECURITY_ACCESS_ERROR,
-                    message=__(f"Unable to parse SQL ({dialect}): {self.sql}"),
+                    message=__(
+                        f"You may have an error in your SQL statement. {message}"
+                    ),
                     level=ErrorLevel.ERROR,
                 )
             ) from ex
--- a/tests/unit_tests/sql_parse_tests.py
+++ b/tests/unit_tests/sql_parse_tests.py
@@ -277,26 +277,30 @@ def test_extract_tables_illdefined() -> None:
     with pytest.raises(SupersetSecurityException) as excinfo:
         extract_tables("SELECT * FROM schemaname.")
     assert (
-        str(excinfo.value) == "Unable to parse SQL (generic): SELECT * FROM schemaname."
+        str(excinfo.value)
+        == "You may have an error in your SQL statement. Error parsing near '.' at line 1:25"
     )
 
     with pytest.raises(SupersetSecurityException) as excinfo:
         extract_tables("SELECT * FROM catalogname.schemaname.")
     assert (
         str(excinfo.value)
-        == "Unable to parse SQL (generic): SELECT * FROM catalogname.schemaname."
+        == "You may have an error in your SQL statement. Error parsing near '.' at line 1:37"
     )
 
     with pytest.raises(SupersetSecurityException) as excinfo:
         extract_tables("SELECT * FROM catalogname..")
     assert (
         str(excinfo.value)
-        == "Unable to parse SQL (generic): SELECT * FROM catalogname.."
+        == "You may have an error in your SQL statement. Error parsing near '.' at line 1:27"
     )
 
     with pytest.raises(SupersetSecurityException) as excinfo:
         extract_tables('SELECT * FROM "tbname')
-    assert str(excinfo.value) == 'Unable to parse SQL (generic): SELECT * FROM "tbname'
+    assert (
+        str(excinfo.value)
+        == "You may have an error in your SQL statement. Error tokenizing 'SELECT * FROM \"tbnam'"
+    )
 
     # odd edge case that works
     assert extract_tables("SELECT * FROM catalogname..tbname") == {


superset/sql_parse.py
```diff
@@ -752,11 +752,21 @@ def _extract_tables_from_sql(self) -> set[Table]:
             statements = parse(self.stripped(), dialect=self._dialect)
         except SqlglotError as ex:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
-            dialect = self._dialect or "generic"
+
+            if isinstance(ex, ParseError):
+                error = ex.errors[0]
+                suffix = "Check the {dialect}manual for the right syntax to use near '{highlight}' at line {line}".format(
```
john-bodley, line 758: Simply doing `str(ex)` doesn't work as `SQLGlot` [includes](https://github.com/tobymao/sqlglot/blob/e5e5ba0c4921f99e063e6d1f47ffaa2181bf26e1/sqlglot/parser.py#L1264-L1273) ANSI codes which are helpful when printing in the terminal, but simply render as `[4m` and `[0m` in HTML.

superset/sql_parse.py
```diff
@@ -752,11 +752,21 @@ def _extract_tables_from_sql(self) -> set[Table]:
...
+                    line=f"{error['line']}:{error['col']}",
+                )
+            else:
+                suffix = ex.args[0]
+
             raise SupersetSecurityException(
                 SupersetError(
                     error_type=SupersetErrorType.QUERY_SECURITY_ACCESS_ERROR,
-                    message=__(f"Unable to parse SQL ({dialect}): {self.sql}"),
+                    message=__(f"You may have an error in your SQL syntax. {suffix}"),
```
john-bodley, line 769: I included the work _may_ as `SQLGlot` may result in false positives whilst parsing. See [here](https://github.com/tobymao/sqlglot/issues/3247) as an example.

superset/sql_parse.py
```diff
@@ -752,11 +752,21 @@ def _extract_tables_from_sql(self) -> set[Table]:
             statements = parse(self.stripped(), dialect=self._dialect)
         except SqlglotError as ex:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
-            dialect = self._dialect or "generic"
+
+            if isinstance(ex, ParseError):
+                error = ex.errors[0]
+                suffix = "Check the {dialect}manual for the right syntax to use near '{highlight}' at line {line}".format(  # pylint: disable=consider-using-f-string,line-too-long
```
john-bodley, line 758: Simply doing `str(ex)` doesn't work as `SQLGlot` [includes](https://github.com/tobymao/sqlglot/blob/e5e5ba0c4921f99e063e6d1f47ffaa2181bf26e1/sqlglot/parser.py#L1264-L1273) ANSI codes which are helpful when printing in the terminal, but simply render as `[4m` and `[0m` in HTML.

mistercrunch:
It feels to me that the SQL IDE should really just pass the SQL to the underlying connection and return the results or error message in most cases, and not get in the way of things.  Now I'm wondering if `_extract_tables_from_sql` needs to run at all times or only if there are schema/tables restrictions in place (?)

john-bodley:
@mistercrunch that was the other option I considered and had discussed internally, i.e., the error being somewhat context aware. Rather than `SQLGlot`'s error being wrapped in a `SupersetSecurityException` it would simply bubble up and be handled in a context aware case-by-case basis. 

Thus in the context of SQL Lab—which doesn't leverage a cache and where institutions _may_ have a third party security manager, e.g. Apache Ranger—then allowing the `SqlglotError` error to pass through to the engine (irrespective of if the user has access to the underlying tables/views) is likely rather safe—especially given that the engine's parser will most likely detect the error.

The challenge arrises when:

1. There is no underlying authorization checks at the engine level.
2. `SQLGlot` returns a false positive ([example](https://github.com/tobymao/sqlglot/issues/3247)) which can be exploited by an actor if (1) is true.

Furthermore the current implement of `raise_for_access()` isn't context aware. Given the enumerated challenges it might be best for individual institutions to augment their security manager as they see fit. For example at Airbnb if we feel that either,

1. `SQLGlot` has a non-acceptable false positive rate ,or
2. `SQLGlot`'s error messaging is too terse (when compared with the underlying engine)

then we may augment our security manager to simply not raise when the `raise_for_access()` method results in a `SqlglotError` when invoked from SQL Lab (leveraging either the referrer or call stack) as we have the peace of mind that Apache Ranger will prevent unauthorized access.

betodealmeida:
> Now I'm wondering if `_extract_tables_from_sql` needs to run at all times or only if there are schema/tables restrictions in place.

That's a good idea! Part of the problem today is that `_extract_tables_from_sql` is called by simply accessing the `.tables` property in the parsed query, we could make it more explicit so it's only called whenever really needed and not by mistake.

superset/sql_parse.py
```diff
@@ -752,11 +752,19 @@ def _extract_tables_from_sql(self) -> set[Table]:
             statements = parse(self.stripped(), dialect=self._dialect)
         except SqlglotError as ex:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
-            dialect = self._dialect or "generic"
+
+            if isinstance(ex, ParseError):
+                error = ex.errors[0]
+                suffix = f"""Error parsing near '{error["highlight"]}' at line {error["line"]}:{error["col"]}"""  # pylint: disable=line-too-long
```
betodealmeida, line 758: Could simplify to:

```suggestion
                suffix = "Error parsing near '{highlight}' at line {line}:{col}".format(**error)
```

Thought the linter usually complain of using `.format()`.

betodealmeida (approved):
❤️

michael-s-molina (approved):
Wow! That's what I call a well-defined PR description 👏🏼 

Thanks for the improvement and explanations.

PR #27884, opened 2024-04-03
chore: consolidate the Superset python package metadata

mistercrunch:
### SUMMARY

Following up on https://github.com/apache/superset/pull/27859, where
I introduced a new file `pyproject.toml` to set up the
python project metadata, here I'd like to consolidate and modernize
further by pushing as much as possible into pyproject.toml,
except where things need to be dynamic, for which we'll keep using
setup.py:

- get rid of setup.cfg, everything there should be able to move to
  pyproject.toml
- remove duplicate keys from setup.py and pyproject.toml
- minimize/simplify setup.py
----

NOTE: we may be able to chisel a bit more at `setup.py`, but it's going to me needed in some capacity since we have some dynamic parameters, notably the `version` number we read and align from `package.json`. There's a tricky thing related to `entry-points` we could tackle.

Also note that having `setup.py` provides continuity for commands like `python setup.py sdist` which is kind of nice.

If we decided to make that a goal, we could fully kill `setup.py` and centralize on pyproject.toml, we're well on our way there.


mypy.ini
```diff
@@ -0,0 +1,33 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
john-bodley, line 1: Why `mypy.ini` and not just add these to `pyproject.toml`. Per [here](https://mypy.readthedocs.io/en/stable/config_file.html) this is a supported configuration setting for Mypy and it would ensure consistency with how we configure `isort` et al.

mistercrunch: Ok let me try it

mistercrunch: ok I did this and pushed a bit further to bring tox.ini into pyproject.toml as well

mistercrunch:
@john-bodley let me know what you think as to whether `tox.ini` should be consolidate as well. I'm fine either way, but note that I had to use "legacy_tox_ini" since not all tox.ini configs have a direct translation in toml, and we have quite a bit of things going on in there...

betodealmeida (approved):
Nice!!

PR #27891, opened 2024-04-04
chore(AlteredSliceTag): Migrate to functional

rtexelm:
### SUMMARY

Refactor component to functional and update tests

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -98,48 +94,83 @@ function alterForComparison(value?: string | null | []): string | null {
...
-    const diffs = this.getDiffs(props);
-    const controlsMap: ControlMap = getControlsForVizType(
-      props.origFormData.viz_type,
-    ) as ControlMap;
-    const rows = this.getRowsFromDiffs(diffs, controlsMap);
+};
 
-    this.state = { rows, hasDiffs: !isEmpty(diffs), controlsMap };
+export const formatValueHandler = (
+  value: any,
```
kgabryje, line 100: Why did we change the type to `any`?

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -98,48 +94,83 @@ function alterForComparison(value?: string | null | []): string | null {
...
+    if (!value.length) {
+      return '[]';
     }
-    const diffs = this.getDiffs(newProps);
-    this.setState(prevState => ({
-      rows: this.getRowsFromDiffs(diffs, prevState.controlsMap),
-      hasDiffs: !isEmpty(diffs),
-    }));
+    return value
+      .map((v: any) => {
```
kgabryje, line 115: same here - wherever possible, we should prefer more specific types

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -98,48 +94,83 @@ function alterForComparison(value?: string | null | []): string | null {
...
-    return Object.entries(diffs).map(([key, diff]) => ({
-      control: controlsMap[key]?.label || key,
-      before: this.formatValue(diff.before, key, controlsMap),
-      after: this.formatValue(diff.after, key, controlsMap),
-    }));
+  if (controlsMap[key]?.type === 'BoundsControl') {
+    return `Min: ${value[0]}, Max: ${value[1]}`;
+  }
+  if (controlsMap[key]?.type === 'CollectionControl' && Array.isArray(value)) {
+    return value.map((v: any) => safeStringify(v)).join(', ');
```
kgabryje, line 128: same here

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -98,48 +94,83 @@ function alterForComparison(value?: string | null | []): string | null {
...
+  }
+  if (
+    controlsMap[key]?.type === 'MetricsControl' &&
+    value.constructor === Array
+  ) {
+    const formattedValue = value.map((v: any) => v?.label ?? v);
```
kgabryje, line 134: same here

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -98,48 +94,83 @@ function alterForComparison(value?: string | null | []): string | null {
...
+    return formattedValue.length ? formattedValue.join(', ') : '[]';
+  }
+  if (typeof value === 'boolean') {
+    return value ? 'true' : 'false';
   }
+  if (Array.isArray(value)) {
+    const formattedValue = value.map((v: any) => v?.label ?? v);
```
kgabryje, line 141: same here

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -149,72 +180,39 @@ class AlteredSliceTag extends React.Component<
...
-    return isEqual(alterForComparison(val1), alterForComparison(val2));
-  }
+  useEffect(() => {
+    const diffs = getDiffs();
+    const controlsMap = getControlsForVizType(
+      props.origFormData?.viz_type,
+    ) as ControlMap;
+    setRows(getRowsFromDiffs(diffs, controlsMap));
+    setHasDiffs(!isEmpty(diffs));
+  }, [getDiffs, props]);
```
kgabryje, line 197: Can we replace `props` with `props.origFormData?.viz_type` in deps array?

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -149,72 +180,39 @@ class AlteredSliceTag extends React.Component<
...
-    return safeStringify(value);
-  }
+      setRows(prevRows =>
+        getRowsFromDiffs(diffs, prevRows as unknown as ControlMap),
+      );
+      setHasDiffs(!isEmpty(diffs));
+    };
+
+    updateStateWithDiffs();
+  }, [getDiffs, props, prevProps]);
```
kgabryje, line 213: How is that useEffect different than the one above? It looks like both will run when props change and both call the same setters

kgabryje: Also, the logic in this useEffect looks a bit different than in original code, where we call `this.getRowsFromDiffs(diffs, prevState.controlsMap)` instead of `getRowsFromDiffs(diffs, prevRows)`. Is that intentional?

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -149,72 +180,39 @@ class AlteredSliceTag extends React.Component<
...
 
-  renderModalBody(): React.ReactNode {
+  const renderModalBody = useCallback(() => {
```
kgabryje, line 215: I think this could be cleaner as a `useMemo`

superset-frontend/src/components/AlteredSliceTag/index.tsx
```diff
@@ -235,39 +233,35 @@ class AlteredSliceTag extends React.Component<
...
         className="table-condensed"
         columnsForWrapText={columnsForWrapText}
       />
     );
-  }
+  }, [rows]);
 
-  renderTriggerNode(): React.ReactNode {
-    return (
+  const renderTriggerNode = useCallback(
```
kgabryje, line 244: this too: 
```suggestion
  const triggerNode = useMemo(
```

superset-frontend/src/components/AlteredSliceTag/AlteredSliceTag.test.jsx
```diff
@@ -17,325 +17,164 @@
...
-        wrapper.instance().formatValue(filters, 'adhoc_filters', controlsMap),
-      ).toBe(expected);
-    });
+test('formatValueHandler formats string values correctly', () => {
+  const value = 'test';
+  const result = formatValueHandler(value, 'key5', controlsMap);
+  expect(result).toEqual('test');
+});
 
-    it('correctly formats filters with string values', () => {
```
kgabryje, line 288: Could we add tests for specific controls like this one? I don't see them in the refactored tests and I think they're valuable

superset-frontend/src/components/AlteredSliceTag/index.tsx
rtexelm: this is an oversight from trying to shoehorn all the changes from the AI into the file, the second seems useEffect like a less helpful approach so I'll delete it

superset-frontend/src/components/AlteredSliceTag/index.tsx
rtexelm: an oversight from the AI changes, meshing a number of different file revisions from different sources

superset-frontend/src/components/AlteredSliceTag/AlteredSliceTag.test.jsx
rtexelm: This was actually being handled but only for string filters by the AdhocFilter test. It was pretty abstract so I made it more explicit in the revision

eschutho (approved):

PR #27905, opened 2024-04-04
fix: retain non-ascii characters in result sets

### LINKED ISSUE #19388: nested column cannot display correctly
A table with some array type columns, Chinese cannot display correctly.

#### How to reproduce the bug
As I already add a dataset named 'array_table'.
1. Go to 'SqlLab'
    `select * from * from array_table limit 10`
2. Click on 'RUN'
3. Scroll down to 'RESULTS'
截屏2022-03-28 上午10 05 44

### Expected results
Chinese array columns display correctly.

### Actual results

截屏2022-03-28 上午10 05 44

#### Screenshots

截屏2022-03-28 上午10 05 44

### Environment

- superset version: 1.4
- python version: 3.7.9
- node.js version: 14.15.5
- any feature flags active:

### Checklist

### LINKED ISSUE #19982: [sqllab] Array output in sqllab displays Chinese characters as utf-8 codes
Returning an array of Chinese characters in sqllab shows up as an array of utf-8 codes. Returning the data as a string and not an array shows up fine. Downloading to CSV also has the same issue of showing up as utf-8 codes. 

#### How to reproduce the bug

1. Run a query that returns an array of Chinese characters
2. Try to view it in sqllab

### Expected results

Shows the characters correctly

### Actual results

Shows utf8 characters, when you click on the data it shows a popup window that displays the characters correctly, but when you download it to csv it shows up incorrectly.

### Environment

- browser type and version:
- superset version: `superset version`
- node.js version: `node -v`
- any feature flags active:

### Checklist

- [] I have checked the superset logs for python stacktraces and included it here as text if there are any.

### LINKED ISSUE #22904: Russian (Cyrillic) text is improperly displayed when using array_agg in the SQL query
#### How to reproduce the bug

1. Store Russian text in a Postgres database.
2. Make sure you can fetch that textual data into Superset and properly display it via SQL lab or chart explorer.
3. Modify the request: add a `group by` clause and use the `array_agg` function to create an array of strings of Russian text.
4. **Problem**: the strings become "unicode gibberish". Example: `["\u041b\u043e\u043d\u0433\u0441\u043b\u0438\u0432\u044b", "\u0421\u0432\u0438\u0442\u0448\u043e\u0442\u044b"]`. They get displayed like this in both the SQL Lab and the Chart Explorer.

### Environment

- browser type and version: updated Brave (Chome-base) and updated Libre Office (Firefox-based)
- helm info:
  - chart: `superset-0.7.7`
  - app version: `1.0`

### Checklist

rusackas:
### SUMMARY

_Hopefully_ this will leave Chinese characters alone, rather than converting them to ascii representations of unicode characters. 

Fixes: https://github.com/apache/superset/issues/19388
Fixes: https://github.com/apache/superset/issues/19982
Fixes: https://github.com/apache/superset/issues/22904

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/result_set.py
+++ b/superset/result_set.py
@@ -69,16 +69,16 @@ def stringify_values(array: NDArray[Any]) -> NDArray[Any]:
 
     with np.nditer(result, flags=["refs_ok"], op_flags=[["readwrite"]]) as it:
         for obj in it:
-            if na_obj := pd.isna(obj):
-                # pandas <NA> type cannot be converted to string
-                obj[na_obj] = None
+            if pd.isna(obj.item()):  # Correctly check for pandas NA or numpy NaN values
+                obj[...] = None  # Directly assign None to handle missing values appropriately
             else:
                 try:
-                    # for simple string conversions
-                    # this handles odd character types better
-                    obj[...] = obj.astype(str)
-                except ValueError:
-                    obj[...] = stringify(obj)
+                    # Convert only if the string is fully ASCII, otherwise leave as is
+                    if obj.item().encode('ascii', 'ignore').decode() == obj.item():
+                        obj[...] = obj.astype(str)
+                except (ValueError, UnicodeEncodeError):
+                    # Catch exceptions for non-ASCII or other conversion issues, leaving them unchanged
+                    continue
 
     return result
 
--- a/superset/result_set.py
+++ b/superset/result_set.py
@@ -69,15 +69,14 @@ def stringify_values(array: NDArray[Any]) -> NDArray[Any]:
 
     with np.nditer(result, flags=["refs_ok"], op_flags=[["readwrite"]]) as it:
         for obj in it:
-            if pd.isna(obj.item()):  # Correctly check for pandas NA or numpy NaN values
-                obj[
-                    ...
-                ] = None  # Directly assign None to handle missing values appropriately
+            item = obj.item()  # Extract the scalar value for manipulation
+            if pd.isna(item):  # Handle missing values by setting them to None
+                obj[...] = None
             else:
                 try:
                     # Convert only if the string is fully ASCII, otherwise leave as is
                     if obj.item().encode("ascii", "ignore").decode() == obj.item():
-                        obj[...] = obj.astype(str)
+                        obj[...] = str(item)
                 except (ValueError, UnicodeEncodeError):
                     # Catch exceptions for non-ASCII or other conversion issues,
                     # leaving them unchanged
--- a/superset/result_set.py
+++ b/superset/result_set.py
@@ -72,21 +72,11 @@ def stringify_values(array: NDArray[Any]) -> NDArray[Any]:
             item = obj.item()  # Extract the scalar value for manipulation
             if pd.isna(item):  # Handle missing values by setting them to None
                 obj[...] = None
+            elif isinstance(item, (dict, list)):  # Serialize dicts and lists to JSON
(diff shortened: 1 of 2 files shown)


superset/result_set.py
```diff
@@ -69,16 +69,16 @@ def stringify_values(array: NDArray[Any]) -> NDArray[Any]:
 
     with np.nditer(result, flags=["refs_ok"], op_flags=[["readwrite"]]) as it:
         for obj in it:
-            if na_obj := pd.isna(obj):
-                # pandas <NA> type cannot be converted to string
-                obj[na_obj] = None
+            if pd.isna(obj.item()):  # Correctly check for pandas NA or numpy NaN values
+                obj[...] = None  # Directly assign None to handle missing values appropriately
```
rusackas, line 73: I can take this part out, but GPT says:

> The original line uses a pattern that's generally seen in contexts where na_obj might be used to index or conditionally modify parts of a data structure. However, this specific syntax doesn't correctly apply to this situation because pd.isna(obj) returns a boolean value, and obj[na_obj] = None suggests an attempt to use boolean indexing or similar logic, which is not applicable here.

> The correct approach, and more streamlined for handling missing values (like NaN or <NA> in pandas data types) in a numpy array while iterating with np.nditer, would be directly assigning None (or np.nan if you're working within numpy's domain and want to keep it as a float array) to elements that are detected as NA/NaN. This is because np.nditer provides a way to iterate over each element of the array and modify it in place if necessary. The obj[...] = None syntax ensures that the current element of the iteration is replaced with None, correctly handling missing values by explicitly setting them as Python's None object, which numpy treats as np.nan when the array's dtype is float.

john-bodley (approved):

superset/result_set.py
```diff
@@ -69,17 +69,24 @@ def stringify_values(array: NDArray[Any]) -> NDArray[Any]:
...
-                    obj[...] = stringify(obj)
-
+                if isinstance(item, str):  # Apply ASCII check only to strings
+                    try:
+                        if item.encode("ascii", "ignore").decode() == item:
+                            obj[...] = item  # It's fully ASCII, leave as is
+                        else:
+                            obj[...] = item  # It's a non-ASCII string, leave as is
+                    except UnicodeEncodeError:
+                        obj[...] = item  # Handle potential encoding issues
```
cccs-tom, line 83: I may be misunderstanding this code, but with lines 79, 81 and 83 doing the exact same thing, can't this whole block be replaced with one assignment?
```suggestion
                    obj[...] = item
```

rusackas: You're quite right... that was a mid-stream revision. I pushed another update that seems closer to my goal, but it's not working as expected, it seems. I'm dabbling in things I'm not an expert at, so I've just converted this to a draft and will fiddle with it when I have spare time unless someone wants to steer me in a better direction :P

cccs-tom: Sounds good! I happened upon this PR while searching for why strings with French accents might be represented in `\u` notation, so we'll be very interested in your fix once it's available!

rusackas: Awesome! I'm not even 100% sure I'm on the right path here, as the tests are clearly not happy. I thought "thisl'll be easy, and I can close a few bugs" but it's not as trivial as I'd thought, it seems.

rusackas:
image

PR #27906, opened 2024-04-04
chore: [proposal] de-matrix python-version in GHAs

mistercrunch:
### SUMMARY

In this PR:
- introducing the notion of a python-version named 'current' and 'next' for the `setup-backend` GHA, and using those semantics in the matrix + required checks.  `setup-backend` is the only place where we install python now, so it's DRY and allowing us to bump in one place in the future, with required checks names staying consistent
- run pytest against the next version of python as @villebro suggested, same as we've been doing for integration tests
- simplifying the single-item matrix (python-version) to NOT using a matrix. I'm guessing the reason we currently have a single-item matrix is an artifact of supporting multiple version in the past, and/or making it easy to go multi-python-version checks in the future, but there's a burden associated, especially around how this relates to "required checks" specified in .asf.yml
- fixing/simplifying the related no-op workflows. We'll need new ones, but will be able to deprecate a bunch and simplify things. For instance, when we migrate to 3.11 in the future, we won't have to manage a bunch of python-version-specific no-ops

About supporting multiple/future version of python, I'd argue that we should focus on a single one for a given CI run, and that if/when we need to CI against multiple version, we run a FULL test suite punctually in a dedicate PR/branch/ref. Point being, it's expensive for every commit to validate multiple versions of python and in many ways its not necessary.

Currently our multi-python-version support is dubious at best, with only few checks that run against multiple versions. I really think we should pick a single version and support it very well. If/when we want to upgrade python version, we'd cut a PR and run CI for that purpose.

If we want to continuously, actively support multiple python versions (and I don't think we should!), I'd suggest either
a release-specific procedure (release manager using release branch, running full CI for that version/release) and/or a nightly job that would keep an eye on that version of python.

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -13,6 +13,10 @@ inputs:
     description: 'Type of requirements to install. Options: base, development, default'
     required: false
     default: 'dev'
+  install-superset:
+    description: 'Whether to install Superset itself. If false, only python is installed'
+    required: false
+    default: 'true'
 
 runs:
   using: 'composite'
@@ -24,11 +28,13 @@ runs:
         cache: ${{ inputs.cache }}
     - name: Install dependencies
       run: |
-        sudo apt-get update && sudo apt-get -y install libldap2-dev libsasl2-dev
-        pip install --upgrade pip setuptools wheel
-        if [ "${{ inputs.requirements-type }}" = "dev" ]; then
-          pip install -r requirements/development.txt
-        elif [ "${{ inputs.requirements-type }}" = "base" ]; then
-          pip install -r requirements/base.txt
+        if [ "${{ inputs.install-superset }}" = "true" ]; then
+          sudo apt-get update && sudo apt-get -y install libldap2-dev libsasl2-dev
+          pip install --upgrade pip setuptools wheel
+          if [ "${{ inputs.requirements-type }}" = "dev" ]; then
+            pip install -r requirements/development.txt
+          elif [ "${{ inputs.requirements-type }}" = "base" ]; then
+            pip install -r requirements/base.txt
+          fi
         fi
       shell: bash
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -2,7 +2,7 @@ name: 'Setup Python Environment'
 description: 'Set up Python and install dependencies with optional configurations.'
 inputs:
   python-version:
-    description: 'Python version to set up. Accepts a version number, "current", or "future".'
+    description: 'Python version to set up. Accepts a version number, "current", or "next".'
     required: true
     default: 'current'
   cache:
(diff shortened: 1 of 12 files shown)


.github/workflows/no-op.yml
```diff
@@ -26,89 +26,103 @@ jobs:
...
+## TO BE DEPRECATED after https://github.com/apache/superset/pull/27906
+## is merged
+## These no-ops were used to deprecate/renaming checks
+## and can be safely delete if seen on `master`
+## ------------------------------------------------------
   pre-commit:
     strategy:
       matrix:
-        python-version: ["3.9"]
+        python-version: ["3.9", "3.10", ""]
```
craig-rueda, line 75: Do we need to include `3.9`? Let's just do `3.10`

craig-rueda:
I'm all for just "officially" supporting a single PY version going forward. Most other projects do this AFAIK, so what's stopping us? :)

raphaelauv:
the only benefit of supporting multiple python version is to increase the chance to offer more compatible client/driver libraries for source data to the end users

example I use the database XXX , the python client is not compatible with python 3.11 , I want use latest version of Apache Superset version X.Y.Z that only support 3.11

a policy like supporting the last 2 version of python with active support could e a first step, wdyt ?

mistercrunch:
Right, though we need to define what "support" means. Some criteria:
- providing an official Docker release for the version of python - probably running the CI/test suite at release time
- running all CI unit/integration tests on that version of python at all times, insuring that say `master` is compatible with that version of python at most times
- not doing pro-active tests, but also not restricting the python package letting people "run at their own risk" (this seems never desirable, and is kind of what we're doing currently)

What we're doing now:
- python package let's user install on 3.9, 3.10 and 3.11 I believe, based on the pyproject.toml
- we run the full test suite on 3.10
- we run one or two test suite on 3.11
- Official docker images are all 3.10

mistercrunch:
Capturing the conversation from the **Release Strategy Group** this morning:

- let's support a single python version officially per release, and run CI against that one mainly
- let's keep preventing "future regressions" on future python version by running some minimal CI against the future version of python (currently test-postgres against 3.11). For now this is the only place where we need a proper matrix. I might try to rename to "current" and "future" to address the required check issues

.github/workflows/no-op.yml
```diff
@@ -0,0 +1,96 @@
+# no-op.yml
```
mistercrunch, line 1: I'll be able to delete this file after we merge this

.github/actions/setup-backend/action.yml
```diff
@@ -13,22 +13,39 @@ inputs:
...
 runs:
   using: 'composite'
   steps:
-    - name: Set up Python ${{ inputs.python-version }}
+    - name: Interpret Python Version
+      id: set-python-version
+      shell: bash
+      run: |
+        if [ "${{ inputs.python-version }}" = "current" ]; then
+          echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
```
rusackas, line 29: No problem for this PR, but I'm thinking at some point we might want to keep our npm/node/python/whatever version numbers in a JSON file at the repo's root, so they can be pulled into scripts, documentation, etc. as needed.

.github/workflows/no-op.yml
```diff
@@ -0,0 +1,96 @@
...
+name: no-op Checks
+on: pull_request
+
+jobs:
+  frontend-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: No-op for frontend-build
+        run: exit 0
+
```
rusackas, line 31: There are some inconsistent breaks between blocks if we want to get all OCD here.

I wonder: would it be a problem to add ALL required checks here? If so, I wonder if we could ingest and iterate through the list from asf.yaml rather than having to maintain this.

rusackas (approved):
Left a couple nits/suggestions, but seems fine to me in general.

PR #27907, opened 2024-04-04
chore: remove fixed psycopg2 version

raphaelauv:
### SUMMARY
~bump psycopg2~

allow upper versions of psycopg2

### DIFF
--- a/docs/docs/installation/running-on-kubernetes.mdx
+++ b/docs/docs/installation/running-on-kubernetes.mdx
@@ -155,7 +155,7 @@ connect to those datasources in your Superset installation:
 ```yaml
 bootstrapScript: |
   #!/bin/bash
-  pip install psycopg2==2.9.6 \
+  pip install psycopg2==2.9.9 \
     sqlalchemy-bigquery==1.6.1 \
     elasticsearch-dbapi==0.2.5 &&\
   if [ ! -f ~/bootstrap ]; then echo "Running Superset with uid {{ .Values.runAsUser }}" > ~/bootstrap; fi
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,7 +150,7 @@ ocient = [
 oracle = ["cx-Oracle>8.0.0, <8.1"]
 pinot = ["pinotdb>=0.3.3, <0.4"]
 playwright = ["playwright>=1.37.0, <2"]
-postgres = ["psycopg2-binary>=2.9.6, <2.10"]
+postgres = ["psycopg2-binary>=2.9.6, <3.0.0"]
 presto = ["pyhive[presto]>=0.6.5"]
 trino = ["trino>=0.328.0"]
 prophet = ["prophet>=1.1.5, <2"]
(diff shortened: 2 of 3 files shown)


requirements/development.txt
```diff
@@ -192,7 +192,7 @@ protobuf==4.23.0
     #   googleapis-common-protos
     #   grpcio-status
     #   proto-plus
-psycopg2-binary==2.9.6
+psycopg2-binary==2.9.9
```
mistercrunch, line 195: this file is dynamically generated from running something like `pip-compile-multi -P psycopg2-binary`, but first you'll have to unpin it here -> https://github.com/apache/superset/blob/master/pyproject.toml#L153

I'm not sure why it's pinned currently, but I'd suggest changing that line to:
`postgres = ["psycopg2-binary>=2.9.6"]` and then running the command above.

docs/docs/installation/running-on-kubernetes.mdx
```diff
@@ -155,7 +155,7 @@ connect to those datasources in your Superset installation:
 ```yaml
 bootstrapScript: |
   #!/bin/bash
-  pip install psycopg2==2.9.6 \
+  pip install psycopg2==2.9.9 \
```
mistercrunch, line 158: this should probably reference the optional dependencies I referenced in another comment. Maybe should say something like `pip install -e .[postgres]`, which would make it consistent and in sync with the ranges defined in pyproject.toml

pyproject.toml
```diff
@@ -150,7 +150,7 @@ ocient = [
 oracle = ["cx-Oracle>8.0.0, <8.1"]
 pinot = ["pinotdb>=0.3.3, <0.4"]
 playwright = ["playwright>=1.37.0, <2"]
-postgres = ["psycopg2-binary==2.9.6"]
+postgres = ["psycopg2-binary==2.9.9"]
```
mistercrunch, line 153: Personally I'd vote for leaving this open as in `psycopg2-binary>=2.9.6`

mistercrunch:
Curious on the motives for the bump, just common hygiene or is there an issue / bug with 2.9.6? 

Generally I like to have code comments when we put an upper bound or a pin on a version so that we know what was keeping us from upgrading when the bound was set.

raphaelauv:
hey , thanks for the review 

2.9.9 support python 3.12 , so it's a first step on superset 3.12 support

mistercrunch:
About python support, I wrote this recently, kind of a request for comment around the topic https://github.com/apache/superset/pull/27906

pyproject.toml
raphaelauv: what about 

`psycopg2-binary>=2.9.6,<2.10` ?

since postgres is the most famous backend for superset itself

and also cause superset do not associate constraint files with release 

so  I don't want to be the guy that make fail future installation in case of breaking change on psycopg2 ( that is very unlikely cause psycopg3 is already out, but we never know )

john-bodley: Requirements are ideally intervals rather than specific versions—which _will_ get pinned for CI et al. when you run `pip-compile-multi`. The concern with pinning a version here is it could severely reduce the feasibly space when `pip-compile-multi` tries to freeze all packages.

john-bodley:
@raphaelauv could you please complete the PR description? This helps to provide the necessary context to the reviewers.

pyproject.toml
raphaelauv: I don't understand your message @john-bodley , what are you talking about ?

PR #27908, opened 2024-04-05
feat(dbview): Add token request button to DuckDB and MotherDuck database modal

guenp:
### SUMMARY
- Add a `DuckDBParametersSchema` to break out the DuckDB connection URI into database, access token (for MotherDuck) and query
- Add a button to go to MotherDuck and copy an access token to paste directly into Superset.

The latter is implemented by adding support for the `metadata` `description` and `load_default` values from the parameters schema fields.

Before:
image

After:
ebebe569-b32b-4c8b-9b0e-98bf7c1d9b6b

### TESTING INSTRUCTIONS
See animated GIF.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset/databases/utils.py
+++ b/superset/databases/utils.py
@@ -104,7 +104,7 @@ def get_table_metadata(
     }
 
 
-def make_url_safe(raw_url: Union[str, URL]) -> Any:
+def make_url_safe(raw_url: Union[str, URL]) -> URL:
     """
     Wrapper for SQLAlchemy's make_url(), which tends to raise too detailed of
     errors, which inevitably find their way into server logs. ArgumentErrors
--- a/superset/databases/utils.py
+++ b/superset/databases/utils.py
@@ -117,11 +117,6 @@ def make_url_safe(raw_url: Union[str, URL]) -> Any:
         url = raw_url.strip()
         try:
             return make_url(url)  # noqa
-        except ValueError:
-            if "duckdb" in url:
-                from superset.db_engine_specs.duckdb import MDURI
-                return MDURI.from_str(url)
-            raise DatabaseInvalidError()
         except Exception:
             raise DatabaseInvalidError()  # pylint: disable=raise-missing-from
 
--- a/superset/databases/utils.py
+++ b/superset/databases/utils.py
@@ -117,6 +117,11 @@ def make_url_safe(raw_url: Union[str, URL]) -> URL:
         url = raw_url.strip()
         try:
             return make_url(url)  # noqa
+        except ValueError:
+            if "duckdb" in url:
+                from superset.db_engine_specs.duckdb import MDURI
+                return MDURI.from_str(url)
+            raise DatabaseInvalidError()
         except Exception:
             raise DatabaseInvalidError()  # pylint: disable=raise-missing-from
 
--- a/superset/databases/utils.py
+++ b/superset/databases/utils.py
@@ -104,7 +104,7 @@ def get_table_metadata(
     }
 
 
-def make_url_safe(raw_url: Union[str, URL]) -> URL:
+def make_url_safe(raw_url: Union[str, URL]) -> Any:
     """
     Wrapper for SQLAlchemy's make_url(), which tends to raise too detailed of
     errors, which inevitably find their way into server logs. ArgumentErrors
--- a/superset/db_engine_specs/duckdb.py
+++ b/superset/db_engine_specs/duckdb.py
@@ -16,6 +16,7 @@
 # under the License.
 from __future__ import annotations
 
+import os
 import re
 from datetime import datetime
 from re import Pattern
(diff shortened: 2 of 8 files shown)


superset/databases/utils.py
```diff
@@ -104,7 +104,7 @@ def get_table_metadata(
     }
 
 
-def make_url_safe(raw_url: Union[str, URL]) -> URL:
+def make_url_safe(raw_url: Union[str, URL]) -> Any:
```
betodealmeida, line 107: This shouldn't be necessary?
```suggestion
def make_url_safe(raw_url: Union[str, URL]) -> URL:
```

superset/db_engine_specs/duckdb.py
```diff
@@ -38,9 +45,208 @@
...
+
+    @classmethod
+    def validate_parameters(
+        cls, properties: DuckDBPropertiesType
+    ) -> list[SupersetError]:
+        """
+        Validates any number of parameters, for progressive validation.
+
+        If only the hostname is present it will check if the name is resolvable. As more
+        parameters are present in the request, more validation is done.
```
betodealmeida, line 208: This doesn't apply to MotherDuck, as there's no hostname validation.

```suggestion
```

superset/db_engine_specs/duckdb.py
```diff
@@ -38,9 +45,208 @@
...
+            return None
+
+        spec = APISpec(
+            title="Database Parameters",
+            version="1.0.0",
+            openapi_version="3.0.2",
+            plugins=[MarshmallowPlugin()],
+        )
+        spec.components.schema(cls.__name__, schema=cls.parameters_schema)
+        return spec.to_dict()["components"]["schemas"][cls.__name__]
```
betodealmeida, line 243: Note to self, we should not require this, it's the same for every database.

superset/db_engine_specs/duckdb.py
```diff
@@ -38,9 +45,208 @@
...
+        if self.motherduck_token:
+            query += f"?motherduck_token={self.motherduck_token}"
+        else:
+            query += "?"
+        query += "&".join([f"{key}={value}" for key, value in self.query.items()])
+        return f"{DuckDBEngineSpec.engine}:///{self.database}{query}"
+
+    @classmethod
+    def from_str(cls, name: str) -> DuckDBURI:
+        pattern = re.compile(
```
betodealmeida, line 72: Not sure if it works for every MotherDuck URL, but you can try using `make_url` from SQLAlchemy here to parse the URL, instead of a regex. Or even `urllib.parse.urlparse`. Regular expressions make me nervous. 😬

betodealmeida (commented):
This is really nice, I love how easy it makes to connect to MotherDuck!

I left a few comments... can we also add some unit tests for `build_sqlalchemy_uri`, `get_parameters_from_uri`, and for the serialization/deserialization of `DuckDBURI`?

michael-s-molina:
@guenp @betodealmeida What's our policy for this screen regarding specific database configs? I thought this was a generic database interface given that it would be really difficult to maintain this modal if we add UI-specific elements for each database type which seems the case here. Are we planning to support specific versions of this modal for each database type?

betodealmeida:
> @guenp @betodealmeida What's our policy for this screen regarding specific database configs? I thought this was a generic database interface given that it would be really difficult to maintain this modal if we add UI-specific elements for each database type which seems the case here. Are we planning to support specific versions of this modal for each database type?

@michael-s-molina the frontend components are mostly generic, and the configuration lives in the DB engine spec (eg, https://github.com/apache/superset/pull/27908/files#diff-651962bab25f2e142c7d1701b15bed4e3c055374d855f1340b85724b573966b7R116-R129). In this PR they're adding some optional parameters to show the URL to get the token, which is useful in general, <acronym title="in my humble opinion">IMHO</acronym>.

That being said, I actually think that the frontend components should be more specific. I think we should have a `MotherDuckDatabaseConfig` component, a `BigQueryDatabaseConfig`, and so on. The current approach of building a UI based on generic components that are shared between databases doesn't scale, and as you mentioned, and the databases will start stepping on each other toes very quickly.

michael-s-molina:
> That being said, I actually think that the frontend components should be more specific. I think we should have a MotherDuckDatabaseConfig component, a BigQueryDatabaseConfig, and so on. The current approach of building a UI based on generic components that are shared between databases doesn't scale, and as you mentioned, and the databases will start stepping on each other toes very quickly.

That was exactly my thinking. It's not a blocker for this PR but something we should consider given the flexibility we'll gain. It's similar to database engine specs where you have common elements but can also extend them to include specific things that will improve overall UX for users.

superset/databases/utils.py
guenp: Good catch, thanks!

superset/db_engine_specs/duckdb.py
guenp: I'm also not sure if we need this function, let me try to refactor this.

guenp: I've removed this class altogether and am using `URL` as per your suggestion.

betodealmeida (approved):
Awesome!

PR #27921, opened 2024-04-05
docs: add more warnings for default secrets and docker-compose

dpgaspar:
### SUMMARY

Add more warnings on our default for docker-compose and stress out the importance of having a unique random `SECRET_KEY`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/installation/configuring-superset.mdx
+++ b/docs/docs/installation/configuring-superset.mdx
@@ -99,6 +99,10 @@ SECRET_KEY = 'YOUR_OWN_RANDOM_GENERATED_SECRET_KEY'
 
 You can generate a strong secure key with `openssl rand -base64 42`.
 
+> Your secret key will be used for securely signing session cookies
+  and encrypting sensitive information on the database
+  Make sure you are changing this key for your deployment with a strong key.
+
 #### Rotating to a newer SECRET_KEY
 
 If you wish to change your existing SECRET_KEY, add the existing SECRET_KEY to your `superset_config.py` file as
--- a/docs/docs/installation/configuring-superset.mdx
+++ b/docs/docs/installation/configuring-superset.mdx
@@ -100,7 +100,7 @@ SECRET_KEY = 'YOUR_OWN_RANDOM_GENERATED_SECRET_KEY'
 You can generate a strong secure key with `openssl rand -base64 42`.
 
 :::caution Your secret key will be used for securely signing session cookies
-  and encrypting sensitive information on the database
+and encrypting sensitive information stored in Superset's application metadata database.
   Make sure you are changing this key for your deployment with a strong key.
 
 #### Rotating to a newer SECRET_KEY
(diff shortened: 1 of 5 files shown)


docs/docs/installation/configuring-superset.mdx
```diff
@@ -99,6 +99,10 @@ SECRET_KEY = 'YOUR_OWN_RANDOM_GENERATED_SECRET_KEY'
 
 You can generate a strong secure key with `openssl rand -base64 42`.
 
+> Your secret key will be used for securely signing session cookies
```
sfirke, line 102:
```suggestion
:::caution Your secret key will be used for securely signing session cookies
```

docs/docs/installation/configuring-superset.mdx
```diff
@@ -99,6 +99,10 @@ SECRET_KEY = 'YOUR_OWN_RANDOM_GENERATED_SECRET_KEY'
...
+  and encrypting sensitive information on the database
```
sfirke, line 103:
```suggestion
and encrypting sensitive information stored in Superset's application metadata database.
```

docs/docs/installation/configuring-superset.mdx
```diff
@@ -99,6 +99,10 @@ SECRET_KEY = 'YOUR_OWN_RANDOM_GENERATED_SECRET_KEY'
...
+  Make sure you are changing this key for your deployment with a strong key.
```
sfirke, line 104:
```suggestion
Ensure your deployment uses a strong, unique key.
Store this key securely. Without this key, you will be unable to access your Superset metadata database. That is, you will lose all of your Superset dashboards, charts, etc. :::
```

sfirke (approved):
@dpgaspar I left a suggestion to change from quote format to caution format (you can see an example of how that renders on the [docker compose install page](https://superset.apache.org/docs/installation/installing-superset-using-docker-compose/)).  And I put a note in there about keeping your key, what do you think about that?  People have written in Slack about getting locked out of their Superset instance when they don't know their SECRET_KEY that they randomly generated.  Maybe they store it as an env variable and then delete a file or wipe out a VM or something.

mistercrunch:
One thing I was thinking about recently around `ENVIRONMENT_TAG_CONFIG` and `SUPERSET_ENV` is we could either force a very visible tag or refuse to start when use a bad combination of things (default `SECRET_KEY` and "SUPERSET_ENV=production").

Also was thinking that in-your-face, "things you should know" in dev would be useful, notices like "hey we noticed you're on python 3.9 - the supported version is now 3.10, you should upgrade"

docs/docs/installation/configuring-superset.mdx
csummers: This `:::caution` block as merged is not closed with a `:::`, so this page renders everything after like so:

image

The suggestion did have a closing `:::`, though.

sfirke: Thanks @csummers! I have a PR open now to fix this.

Issue #27504, opened 2024-03-14
Time comparison does not always capture the correct data when the result set is limited


eschutho:
### Bug description

When fetching a set of data over a period of time and adding a time comparison, if the result set is truncated, some of the comparison information will also be missing. 
**Here is an example with missing data when a small row limit is applied:**
<img width="2324" alt="Screenshot 2024-03-13 at 4 55 31 PM" src="https://github.com/apache/superset/assets/5186919/87759e39-ef7e-4288-8dae-5082860951ce">

In the above example, I am trying to fetch the top 10 years from the years 1989 - 2004 where the name Alex was the most popular and compare that count to the period 10 years prior. As you can see, the values for 1981, 1982, 1983, and 1984 are missing, although I would expect to get those values, because as you can see, if I fetch more data by removing the row limit, those values exist. 

**Here is an example with all the data when a larger row limit is applied**
<img width="2324" alt="Screenshot 2024-03-13 at 4 55 59 PM" src="https://github.com/apache/superset/assets/5186919/b58663c7-f38f-424f-995e-f7c25ce0e4b5">

I believe the reason why this bug exists is because we are making two separate db requests, one for each time period when the name is "Alex", and limiting each request by 10 and sorting on the count. Because the count for each response is going to be different, when we make the comparison, the values for the dates in the original time series may not exist. 

**Here are the current db requests when a row limit is applied:** 
<img width="894" alt="Screenshot 2024-03-13 at 5 01 33 PM" src="https://github.com/apache/superset/assets/5186919/77f9b08c-a8b6-498b-8268-2eba5c2d6024">
<img width="894" alt="Screenshot 2024-03-13 at 5 01 59 PM" src="https://github.com/apache/superset/assets/5186919/f4647d8c-528e-480b-b702-b3a9aea8c3c5">

I believe what needs to happen instead is that the first query with timerange A needs to join the data from timerange B on a left outer join, and then apply the limit, and that would give the correct results. 

**This is an example of a query that would fetch all the data required:**
```
SELECT 
  query1.ds AS ds1,
  query1.count AS count1,
  query2.ds AS ds2,
  query2.count AS count2
FROM 
  (
    SELECT 
      DATE_TRUNC('year', ds) AS ds,
      COUNT(*) AS count
    FROM 
      public.birth_names
    WHERE 
      ds >= TO_TIMESTAMP('1989-03-13 00:00:00.000000', 'YYYY-MM-DD HH24:MI:SS.US') AND
[...]

PR #27934, opened 2024-04-08
fix(time_offset): improved LIMIT-handling in advanced analytics

Antonio-RiveroMartnez:
### SUMMARY
When applying a time_offset in Advanced Analytics, if the row_limit is too small, data will not be present when the join occurs, so this PR prevent such behavior by not using limit nor offset in the inner query (the shifted one).

For example this query shown in the following video has all the records for a period of 10+ years:

https://github.com/apache/superset/assets/38889534/8dca8b16-5c05-4b2d-8fd6-67387c8708e7

We can see that the result of shifted queries not always include the same periods than the original query, so when the join gets applied, it won't match some records, thus giving you incorrect output.

Original query with limit and sorting:
Captura de pantalla 08 04 2024 a 11 05 13 AM

1 year ago query with limit and sorting:
Captura de pantalla 08 04 2024 a 11 05 50 AM

2 years ago query with limit and sorting:
Captura de pantalla 08 04 2024 a 11 06 08 AM

Error:
In the following image some datapoints are missing in the shifted columns because they are not part of the result dataframe after sorting and limiting it like the original query.
error

Fix:
Now we can see that the query is able to find the records that belong to those offsets regardless, because it's joining using the complete data.
test

### ADDITIONAL INFORMATION

- [X] Has associated issue: https://github.com/apache/superset/issues/27504
- [ ] Required feature flags:
- [ ] Changes UI


superset/common/query_context_processor.py
```diff
@@ -455,6 +455,13 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
                 for metric in metric_names
             }
 
+            # When the original query has limit or offset we wont apply those
+            # to the subquery so we prevent data inconsistency due to missing records
+            # in the dataframes when performing the join
+            if query_object.row_limit or query_object.row_offset:
+                query_object_clone_dct["row_limit"] = None
```
zhaoyongjie, line 462: We should apply `row_limit` in config.py, in order to avoid a heavy query.

Antonio-RiveroMartnez: makes sense, will do. Thanks

john-bodley:
@Antonio-RiveroMartnez are there any open issues we can link to this? I'm somewhat surprised that no one has reported an issue with this in the past.

Antonio-RiveroMartnez:
Hey @john-bodley I just linked it in the PR description, thanks.

superset/common/query_context_processor.py
```diff
@@ -455,6 +455,13 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
                 for metric in metric_names
             }
 
+            # When the original query has limit or offset we wont apply those
```
mistercrunch, line 458: Wow this method has gotten large! Thinking about refactors here and factoring out a `processing_time_offset` (a method for each time offset) might help a little in reading this, but probably breaking down further would be good too. 

Doing the refactor in this PR as part of this PR would make it harder to spot the diff, so I'll approve this one, but hoping we can refactor some of these into more bite-sized methods. Maybe a goal could be to remove the `disable=too-many-locals,too-many-statements` above.

mistercrunch (approved):
Changes + tests LGTM. Let's follow up with a PR breaking this large method in the right places.

eschutho (approved):

PR #27943, opened 2024-04-08
feat: improve event logging for queries + refactor

mistercrunch:
### SUMMARY

The main driver for this PR was to enrich event logging around database engine and database drivers for events that interact directly with analytics databases. While touching this up, I did some refactor along the way.

- Adding a new analytics event/action `execute_sql` that targets just that, that **isn't** called when hitting the cache for instance. Using the context manager insures capturing the duration of the call as well.
- refactored logic duplication around config items `SQL_QUERY_MUTATOR` and `MUTATE_AFTER_SPLIT`, where code had been copy/pasted and reimplemented in various places. Here I move it as a method of the Database model.
- side mission: I realized that `request.form` is empty when posting JSON payload, and `request.json` should be used instead. This should automatically capture more automated logging that parses context out of the request object proactively
- factored out `post_process_df` as it seemed logical to isolate this out of `get_df`
- fix cypress glitch + improved cypress docs

### DIFF
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -66,6 +66,7 @@ from superset.db_engine_specs.base import MetricType, TimeGrain
 from superset.extensions import (
     cache_manager,
     encrypted_field_factory,
+    event_logger,
     security_manager,
     ssh_manager_factory,
 )
@@ -564,6 +565,20 @@ class Database(
         """
         return self.db_engine_spec.get_default_schema_for_query(self, query)
 
+    @staticmethod
+    def post_process_df(df: pd.DataFrame) -> pd.DataFrame:
+        def column_needs_conversion(df_series: pd.Series) -> bool:
+            return (
+                not df_series.empty
+                and isinstance(df_series, pd.Series)
+                and isinstance(df_series[0], (list, dict))
+            )
+
+        for col, coltype in df.dtypes.to_dict().items():
+            if coltype == numpy.object_ and column_needs_conversion(df[col]):
+                df[col] = df[col].apply(utils.json_dumps_w_dates)
+        return df
+
     @property
     def quote_identifier(self) -> Callable[[str], str]:
         """Add quotes to potential identifier expressions if needed"""
@@ -581,15 +596,15 @@ class Database(
         sqls = self.db_engine_spec.parse_sql(sql)
         with self.get_sqla_engine(schema) as engine:
             engine_url = engine.url
-        mutate_after_split = config["MUTATE_AFTER_SPLIT"]
-        sql_query_mutator = config["SQL_QUERY_MUTATOR"]
 
-        def needs_conversion(df_series: pd.Series) -> bool:
-            return (
-                not df_series.empty
-                and isinstance(df_series, pd.Series)
-                and isinstance(df_series[0], (list, dict))
-            )
+        def _mutate_sql_if_needed(sql_: str) -> str:
+            if config["MUTATE_AFTER_SPLIT"]:
+                return config["SQL_QUERY_MUTATOR"](
+                    sql_,
+                    security_manager=security_manager,
+                    database=None,
+                )
+            return sql_
 
         def _log_query(sql: str) -> None:
             if log_query:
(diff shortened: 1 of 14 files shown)


superset/models/core.py
```diff
@@ -598,42 +627,30 @@ def _log_query(sql: str) -> None:
...
-            else:
-                _log_query(sqls[-1])
-                self.db_engine_spec.execute(cursor, sqls[-1], self)
-
-            data = self.db_engine_spec.fetch_data(cursor)
-            result_set = SupersetResultSet(
-                data, cursor.description, self.db_engine_spec
-            )
-            df = result_set.to_pandas_df()
+                with event_logger.log_context(
```
mistercrunch, line 634: NOTE: this new event (along with the SQL Lab one bellow) was the driver for this PR. I wanted to have a clear event everytime we run SQL against a database, capture the engine name, driver and duration.

superset/sql_lab.py
```diff
@@ -281,21 +276,26 @@ def execute_sql_statement(  # pylint: disable=too-many-statements
...
-                query.id,
-                str(query.to_dict()),
-            )
-            data = db_engine_spec.fetch_data(cursor, increased_limit)
-            if query.limit is None or len(data) <= query.limit:
-                query.limiting_factor = LimitingFactor.NOT_LIMITED
-            else:
-                # return 1 row less than increased_query
-                data = data[:-1]
+        with event_logger.log_context(
```
mistercrunch, line 279: NOTE: the other event capturing sql execution with database details and duration (the context manager handles calculating and logging `duration`)

superset/utils/log.py
```diff
@@ -47,10 +47,12 @@ def collect_request_payload() -> dict[str, Any]:
...
     payload: dict[str, Any] = {
         "path": request.path,
-        **request.form.to_dict(),
-        # url search params can overwrite POST body
-        **request.args.to_dict(),
     }
+    payload.update(**request.form.to_dict())
+    payload.update(**request.args.to_dict())
+    if request.is_json:
+        json_payload = request.get_json(cache=True, silent=True)
```
mistercrunch, line 54: NOTE: here I discovered 2 important things:
- the logging framework, that inspects the `flask.request` object to gather context, was missing out on `request.json` which in many cases has rich context to read from
- `request.json` is a stream and can only be read once reliably, after which it'll show as empty. Here using `get_json(cache=True)` ensures the process can read it later in the execution flow.

superset/utils/log.py
```diff
@@ -359,3 +365,29 @@ def log(  # pylint: disable=too-many-arguments,too-many-locals
         except SQLAlchemyError as ex:
             logging.error("DBEventLogger failed to log event(s)")
             logging.exception(ex)
+
+
+class StdOutEventLogger(AbstractEventLogger):
```
mistercrunch, line 370: NOTE: side mission, creating this to use in my `superset_config` made it easier to test/debug. This `StdOutEventLogger` could be a decent default overall, or at least for when we're in development / debug mode. Though changing the default would be a breaking change so I'm holding back here for now.

superset/sql_lab.py
```diff
@@ -73,7 +73,6 @@
 SQLLAB_HARD_TIMEOUT = SQLLAB_TIMEOUT + 60
 SQL_MAX_ROW = config["SQL_MAX_ROW"]
 SQLLAB_CTAS_NO_LIMIT = config["SQLLAB_CTAS_NO_LIMIT"]
-SQL_QUERY_MUTATOR = config["SQL_QUERY_MUTATOR"]
```
john-bodley, line 76: Removing this configuration variable would be deemed a breaking change.

mistercrunch: Oh I'm not removing this config flag, just centralizing it's usage into the Database model -> 
https://github.com/apache/superset/pull/27943/files#diff-462c756528595a2be4a0a3c3063077f984384c7858ecdb5bd3e9603163c3678bR590

superset/sql_lab.py
```diff
@@ -264,11 +263,7 @@ def execute_sql_statement(  # pylint: disable=too-many-statements
         sql = apply_limit_if_exists(database, increased_limit, query, sql)
 
     # Hook to allow environment-specific mutation (usually comments) to the SQL
-    sql = SQL_QUERY_MUTATOR(
-        sql,
-        security_manager=security_manager,
-        database=database,
-    )
+    sql = database.mutate_sql_based_on_config(sql)
```
mistercrunch, line 266: @john-bodley here's the call to the centralized method that applies the same logic consistently across `SQL_QUERY_MUTATOR` and `MUTATE_AFTER_SPLIT` in one place

superset/sql_lab.py
mistercrunch: I'll go ahead and remove the labels as this is not a breaking change

superset/models/core.py
```diff
@@ -572,7 +587,30 @@ def quote_identifier(self) -> Callable[[str], str]:
...
+        - `MUTATE_AFTER_SPLIT`: If True, the SQL query mutator is only called after the
+          sql is broken down into smaller queries. If False, the SQL query mutator applies
+          on the group of queries as a whole. Here the called passes the context
+          as to whether the SQL is split or already.
+        """
+        sql_mutator = config["SQL_QUERY_MUTATOR"]
+        if sql_mutator and (
+            (is_splitted and config["MUTATE_AFTER_SPLIT"])
+            or (not is_splitted and not config["MUTATE_AFTER_SPLIT"])
+        ):
```
betodealmeida, line 605: Small suggestion to simplify the logic:

```suggestion
        if sql_mutator and (is_splitted == config["MUTATE_AFTER_SPLIT"]):
```

I'd also rename `is_splitted` to `is_split` ([ref](https://english.stackexchange.com/questions/467873/the-past-participle-of-split-split-or-splitted#467898)).

superset/models/core.py
```diff
@@ -603,42 +632,30 @@ def _log_query(sql: str) -> None:
...
-            df = result_set.to_pandas_df()
+                with event_logger.log_context(
+                    action="execute_sql",
+                    database=self,
+                    object_ref=__name__,
+                ):
+                    self.db_engine_spec.execute(cursor, sql_, self)
+                    if i < len(sqls) - 1:
+                        # If it's not the last, we don't keep the results
+                        cursor.fetchall()
```
betodealmeida, line 647: It would be nice in the future if we captured all results and surfaced them in the UI, it's a feature that similar tools have.

superset/utils/log.py
```diff
@@ -47,10 +47,12 @@ def collect_request_payload() -> dict[str, Any]:
 
     payload: dict[str, Any] = {
         "path": request.path,
-        **request.form.to_dict(),
-        # url search params can overwrite POST body
-        **request.args.to_dict(),
     }
+    payload.update(**request.form.to_dict())
+    payload.update(**request.args.to_dict())
```
betodealmeida, line 52:
```suggestion
```

betodealmeida (approved):
Looks great! Left a few suggestions.

superset/utils/log.py
mistercrunch: the 2 lines look identical at first glance but they're different. Actually now I'm not sure why I changed lines 50-52... Let me roll this back to what it was before

PR #27950, opened 2024-04-09
feat: Utility function to render chart tooltips

michael-s-molina:
### SUMMARY
This PR adds an utility function to render chart tooltips with the following objectives:
- Add the ability to display total and percentage values for our tooltips. This improvement will also allow us to migrate legacy timeseries charts, which currently display that information, to ECharts.
- Standardize the way we display tooltips. Currently, every plugin displays tooltips using different information and designs, which decreases the overall user experience. By creating a reusable function, we can ensure consistency across plugins.
- Reduce code duplication.

This PR will only format ECharts tooltips to limit the scope of work but other plugins can also benefit from the utility function.

Screenshot 2024-04-09 at 10 40 12
Screenshot 2024-04-09 at 09 27 11
Screenshot 2024-04-09 at 10 38 21
Screenshot 2024-04-16 at 15 17 05
Screenshot 2024-04-09 at 10 42 09
Screenshot 2024-04-09 at 09 37 36
Screenshot 2024-04-10 at 10 19 15
Screenshot 2024-04-10 at 11 23 46
Screenshot 2024-04-16 at 10 55 53
Screenshot 2024-04-16 at 11 37 02
Screenshot 2024-04-16 at 15 02 27
Screenshot 2024-04-16 at 15 08 56
Screenshot 2024-04-16 at 15 23 41
Screenshot 2024-04-16 at 15 34 21
Screenshot 2024-04-18 at 13 37 20
Screenshot 2024-04-18 at 13 41 16
Screenshot 2024-04-18 at 14 12 42
Screenshot 2024-04-18 at 14 48 27
Screenshot 2024-04-19 at 09 12 08
Screenshot 2024-04-19 at 09 24 32
Screenshot 2024-04-19 at 11 35 04
Screenshot 2024-04-19 at 11 50 08
Screenshot 2024-04-19 at 13 22 01
Screenshot 2024-04-19 at 13 25 55
Screenshot 2024-04-19 at 13 30 42
Screenshot 2024-04-19 at 13 43 55
Screenshot 2024-04-19 at 13 49 11
Screenshot 2024-04-19 at 13 56 33

### TESTING INSTRUCTIONS
Test consists in checking how the tooltip is displayed for all ECharts plugins and their different control combinations.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/transformProps.ts
```diff
@@ -249,7 +229,27 @@ export default function transformProps(
...
+              ],
+            };
+            return new TooltipRenderer(
+              data,
+              0,
+              false,
+              false,
+              formatTime(params[0].data[0]),
+            ).renderHtml();
+          },
```
villebro, line 252: It appears we're re-instantiating `TooltipRenderer` on each call to the formatter, which may (or may not) cause an unnecessary perf hit. Did you consider implementing this so that we can define the renderer only once, essentially memoizing it, and then call that in a callback that is able to render the html using `params` as the input? Note, I may be way off here, but thinking out loud..

villebro (commented):
This is super exciting, I'm really looking forward to standardizing the tooltip renderer! First pass comment regarding potential readability/performance.

michael-s-molina:
@justinpark @villebro @rusackas Thank you for the comments. I'm still working on the component's API and implementation as I continue to iterate on each plugin. I'm still not sure what will be the final API and implementation because each plugin has a unique set of requirements. For that reason, I'll first first make sure all ECharts plugins are using the renderer, to have a clear view of all the requirements, and then work on optimizations. Feel free to review this PR while it evolves or wait for when it leaves the draft state.

villebro (commented):
This is looking great! Maybe add a few tests, but other than that, once CI clears this looks much better than the current state of affairs.

michael-s-molina:
Thanks for linking the resources @sfirke!

superset-frontend/packages/superset-ui-core/src/utils/tooltip.ts
```diff
@@ -0,0 +1,59 @@
...
+          ${data.length === 0 ? `<tr><td>${t('No data')}</td></tr>` : ''}
+          ${data
+            .map((row, i) => {
+              const rowStyle =
+                i === focusedRow ? 'font-weight: 700;' : 'opacity: 0.8;';
+              let padding = 0;
+              const cells = row.map((cell, j) => {
+                const cellStyle = `
+                  text-align: ${j > 0 ? 'right' : 'left'};
+                  padding-left: ${padding}px;
```
justinpark, line 48: nit: It appears you intended the following way
```suggestion
                  padding-left: ${j === 0 ? 0 : 16}px;
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/transformProps.ts
```diff
@@ -119,10 +123,11 @@ export default function transformProps(
 
   let [min, max] = (valueBounds || []).map(parseAxisBound);
   if (min === undefined) {
-    min = minBy(data, row => row[colorColumn])?.[colorColumn] as number;
+    min = (minBy(data, row => row[colorColumn])?.[colorColumn] as number) || 0;
   }
   if (max === undefined) {
-    max = maxBy(data, row => row[colorColumn])?.[colorColumn] as number;
+    max =
+      (maxBy(data, row => row[colorColumn])?.[colorColumn] as number) || 200;
```
justinpark, line 130: Where is the value of `200` coming from?

michael-s-molina: This is coming from the [default values](https://echarts.apache.org/en/option.html#visualMap-continuous.max) of ECharts. I added a constant to better describe the value.

justinpark (approved):
LGTM!

igamat:
Really nice feature indeed ! Will this feature be also available for the other types of charts (Country Map, World Map, Bar chart...) any time soon ? Thanks

rusackas:
> Really nice feature indeed ! Will this feature be also available for the other types of charts (Country Map, World Map, Bar chart...) any time soon ? Thanks

I think we all WANT it to get there, but it's probably not a high priority at the moment. If you have the means to contribute that as a PR, we'd be super excited to review it!

PR #22623, opened 2023-01-06
chore: skip puppeteer chromium download

nisheldo:
### SUMMARY
Skipping puppeteer chromium download in the build docker stage along with https://github.com/apache/superset/pull/22615 will allow arm based users to finally build locally 

Adding this will fix this error when building on m1 macs

 > [superset-node 7/8] RUN /frontend-mem-nag.sh && cd /app/superset-frontend && npm ci:
https://github.com/apache/superset/issues/18 0.141 Memory check Ok [12.1407 GB free]
https://github.com/apache/superset/issues/18 108.2 npm notice
https://github.com/apache/superset/issues/18 108.2 npm notice New major version of npm available! 7.24.2 -> 9.2.0
https://github.com/apache/superset/issues/18 108.2 npm notice Changelog: https://github.com/npm/cli/releases/tag/v9.2.0
https://github.com/apache/superset/issues/18 108.2 npm notice Run npm install -g npm@9.2.0 to update!
https://github.com/apache/superset/issues/18 108.2 npm notice
https://github.com/apache/superset/issues/18 108.2 npm ERR! code 1
https://github.com/apache/superset/issues/18 108.2 npm ERR! path /app/superset-frontend/node_modules/puppeteer
https://github.com/apache/superset/issues/18 108.2 npm ERR! command failed
https://github.com/apache/superset/issues/18 108.2 npm ERR! command sh -c node install.js
https://github.com/apache/superset/issues/18 108.2 npm ERR! The chromium binary is not available for arm64.
https://github.com/apache/superset/issues/18 108.2 npm ERR! If you are on Ubuntu, you can install with:
https://github.com/apache/superset/issues/18 108.2 npm ERR!
https://github.com/apache/superset/issues/18 108.2 npm ERR! sudo apt install chromium
https://github.com/apache/superset/issues/18 108.2 npm ERR!
https://github.com/apache/superset/issues/18 108.2 npm ERR!
https://github.com/apache/superset/issues/18 108.2 npm ERR! sudo apt install chromium-browser
https://github.com/apache/superset/issues/18 108.2 npm ERR!
https://github.com/apache/superset/issues/18 108.2 npm ERR! /app/superset-frontend/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:115
https://github.com/apache/superset/issues/18 108.2 npm ERR! throw new Error();
https://github.com/apache/superset/issues/18 108.2 npm ERR! ^
[...]

PR #27953, opened 2024-04-09
feat: optimize docker-compose up for faster boot time

mistercrunch:
### SUMMARY

When firing the `docker-compose up` command, which is pretty much the first step of most developer workflows, we kick off up 2 chromium downloads: one for PLAYWRIGHT and one for PUPPETEER by default. Those are used for optional features like alerts/reports/thumbnails/ and for CI.

My approach here to prevent all this work when running `docker-compose up` is to:
- playwright: move the download/install to the `Dockerfile` in the `dev` layer
- pupeteer: turn off by default. From my understanding this is only really required for a CI step, and is implemented there. I don't think developers are meant to have this locally. But note that you can still turn it on by switching a env var.

As a side mission, since I needed it while working on this, I'm also introducing support for defining environment variables overrides for docker-compose in `docker/.env-local` while adding this to `.gitignore`, which developers can use without the fear/confusion around committing local environment specific settings to the repo.

-----------------

About the core feature in this PR here, this saves minutes upon firing `docker-compose up`, and reduce confusion
around "what the heck is it doing!?", in a phase where we should just be pulling and starting docker images.

#### More about PUPPETEER
About PUPPETEER: in https://github.com/apache/superset/pull/22623, I found that it is used by Applitool in two GHAs:
- (superset-applitool-cypress)[https://github.com/apache/superset/blob/master/.github/workflows/superset-applitool-cypress.yml]
- (superset-applitools-storybook)[https://github.com/apache/superset/blob/master/.github/workflows/superset-applitools-storybook.yml]

Seems PUPPETEER is probably never really useful locally.

#### More about PLAYWRIGHT 

About PLAYWRIGHT, this is what powers alerts and reports as well thumbnail-generation.

### Potential follow up work
- decide on Chromium VS Firefox as the primary webdriver to power those features, and remove the other one from the docker image as it bloats it with whichever one is not in use
- use playwright to download the Firefox webdriver as right now it's a hot mess of wget tarballs, and it seemed playwright wraps this up nicely
- Dockerfile refactor: currently this layer is low in the docker pipeline, any layer rebuild before this one will retrigger the download of the browser. Eventually we could compose docker images to prevent this. I've been talking about setting up some "fit" and "fat" images for a while, which would include the web driver(s). Currently only `dev` has this.


mistercrunch:
@supersetbot rebase

mistercrunch:
@supersetbot rebase

mistercrunch:
@supersetbot rebase

mistercrunch:
@supersetbot rebase

.github/workflows/docker.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
       - id: set_matrix
         run: |
-          MATRIX_CONFIG=$(if [ "${{ github.event_name }}" == "pull_request" ]; then echo '["ci"]'; else echo '["dev", "lean", "py310", "websocket", "dockerize"]'; fi)
+          MATRIX_CONFIG=$(if [ "${{ github.event_name }}" == "pull_request" ]; then echo '["dev"]'; else echo '["dev", "lean", "py310", "websocket", "dockerize"]'; fi)
```
mistercrunch, line 24: this is a bit cryptic, but the short story is we built only the `ci` layer in PRs (and a wider list on `master` push events), and I decided that building `dev` makes much more sense. I thought `ci` was all-encompassing when I first set this matrix, but turns out it's just a small layer on top of lean. Testing `dev` captures the bulk of the dockerfile.

Dockerfile
```diff
@@ -119,9 +119,15 @@ RUN apt-get update -qq \
...
-    && wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz -O - | tar xfz - -C /usr/local/bin \
+        git \
+        pkg-config
+
+RUN pip install playwright
+RUN playwright install-deps
+RUN playwright install chromium
+
+# Install GeckoDriver WebDriver
+RUN wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz -O - | tar xfz - -C /usr/local/bin \
```
mistercrunch, line 130: I wonder if we could simply `playright install firefox` here instead of this line and the following. For now I decided not to touch it.

rusackas (approved):

docker-compose-image-tag.yml
```diff
@@ -33,7 +33,11 @@ services:
       - redis:/data
 
   db:
-    env_file: docker/.env
+    env_file:
```
shaoxp, line 36: these changes for env_file seems make i can not start superset following the documentation by
`docker-compose -f docker-compose-image-tag.yml up`

the error:
```The Compose file './docker-compose-image-tag.yml' is invalid because:
services.db.env_file contains {"path": "docker/.env", "required": true}, which is an invalid type, it should be a string
services.superset.env_file contains {"path": "docker/.env", "required": true}, which is an invalid type, it should be a string
services.superset-init.env_file contains {"path": "docker/.env", "required": true}, which is an invalid type, it should be a string
services.superset-worker.env_file contains {"path": "docker/.env", "required": true}, which is an invalid type, it should be a string
services.superset-worker-beat.env_file contains {"path": "docker/.env", "required": true}, which is an invalid type, it should be a string```

I am on Ubuntu 22.04.4 LTS

mistercrunch: hey @shaoxp , this got fixed in https://github.com/apache/superset/pull/28039 , but note that you need docker-compose 2.24+ for this config to work. I tried to lookup if there's a version of the yaml I can setup to inform users to upgrade but couldn't find anything.

shaoxp: Thanks for the info. i am able to start locally. 

maybe you can update the document to include the required version.

PR #27956, opened 2024-04-09
docs: deploy docs when merging to master

mistercrunch:
### SUMMARY

I believe the intention is to have the docs
aligned with what is on `master`, currently pushing on a release
branch will trigger a potential regression.

It could be reasonable for the docs to be aligned with the latest
release, but if that was the intention we should deploy the
docs as part of the release process, and avoid previous release branch
from deploying the docs.

I think this happened by mistake on
https://github.com/apache/superset/pull/27390/files#diff-1f3b6eaac4370772e5f0d6cebab7906e0a83ce78e07f6ae8a239c06b33a420d4R9
when the intention was to trigger CI on release branches, not to deploy
the docs.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -72,6 +72,7 @@ jobs:
         env:
           DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


rusackas (approved):
LGTM!

villebro (approved):
Good catch, this must have caused lots of flip-floppery..

mistercrunch:
@rusackas I'll let you merge this one since you wanted to use it for troubleshooting the `.htaccess` stuffs

PR #27962, opened 2024-04-10
feat: Dashboard tabs api endpoint

fisjac:
### SUMMARY

This PR implements a new property on the Dashboard model granting access to the tabs associated with a dashboard. A new API endpoint for accessing the associated dashboard tabs has been created at `/api/v1/dashboard/[id]/tabs`.

The payload of the api endpoint contains two properties: `all_tabs` which contains an array of all associated tabs, and `tab_tree` which contains a nested object representing the nested structure of tabs within the dashboard. tab_tree follows the data structure necessary for consumption by the Antd Tree and TreeSelect components on the frontend.

The primary use case for this implementation is for enabling the functionalities described in SIP: https://github.com/apache/superset/issues/26183.

The provided proof of concept provides an alternative approach, centralizing the business logic on the backend to process prior to serving to the client.

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -76,6 +76,7 @@ from superset.dashboards.schemas import (
     get_fav_star_ids_schema,
     GetFavStarIdsSchema,
     openapi_spec_methods_override,
+    TabsPayloadSchema,
     thumbnail_query_schema,
 )
 from superset.extensions import event_logger
@@ -141,6 +142,7 @@ class DashboardRestApi(BaseSupersetModelRestApi):
         "remove_favorite",
         "get_charts",
         "get_datasets",
+        "get_tabs",
         "get_embedded",
         "set_embedded",
         "delete_embedded",
@@ -236,6 +238,7 @@ class DashboardRestApi(BaseSupersetModelRestApi):
     chart_entity_response_schema = ChartEntityResponseSchema()
     dashboard_get_response_schema = DashboardGetResponseSchema()
     dashboard_dataset_schema = DashboardDatasetSchema()
+    tab_schema = TabsPayloadSchema()
     embedded_response_schema = EmbeddedDashboardResponseSchema()
     embedded_config_schema = EmbeddedDashboardConfigSchema()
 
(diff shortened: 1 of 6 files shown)


tests/integration_tests/dashboards/api_tests.py
```diff
@@ -869,6 +869,185 @@ def test_get_dashboards_created_by_me(self):
...
+                    },
+                ],
+            }
+        }
+        self.assertEqual(rv.status_code, 200)
+        self.assertEqual(response, expected_response)
+        db.session.delete(dashboard)
+        db.session.commit()
+
+    @pytest.mark.usefixtures("load_world_bank_dashboard_with_slices")
```
geido, line 1040: Should we add other negative paths, such as testing for permissions or invalid schema?

superset/dashboards/api.py
```diff
@@ -395,6 +399,64 @@ def get_datasets(self, id_or_slug: str) -> Response:
...
+        """
+        try:
+            tabs = DashboardDAO.get_tabs_for_dashboard(id_or_slug)
+            result = self.tab_schema.dump(tabs)
+            return self.response(200, result=result)
+
+        except (TypeError, ValueError) as err:
+            return self.response_400(
+                message=gettext(
+                    "Tab schema is invalid, caused by: %(error)s", error=str(err)
```
geido, line 452: I don't think we need to surface the specific error here in the response. We might want to log it instead.

fisjac: I borrowed this error handler from the `GET dashboard/<id>/datasets` endpoint, which throws when an invalid schema is detected. My thought was to ensure consistent validation behavior for similar endpoints.

tests/integration_tests/dashboards/api_tests.py
fisjac: currently the only tests that explicitly check permissions for the dashboard api are for PUT and DELETE requests. My understanding is that the FlaskAppbuilder @protect decorator is doing most of the work for ensuring permissions for GET requests. Presumably the @protect decorator is being tested somewhere else within the codebase?

geido (approved):

PR #27969, opened 2024-04-10
feat: add option to disable rendering of html in sql lab and table chart

soniagtm:
### SUMMARY

Currently, the platform renders any HTML command in the result table, which can sometimes obscure the full HTML in the request body. To address this, an option button that enables users to control the rendering of HTML in the result table is added. This option would provide users with more flexibility and clarity when viewing HTML content within SQLLab, as well as in table chart. This addition is also discussed in [this lazy consensus](https://lists.apache.org/thread/1y4rxnply1m7xonnwro2b12f276vc2hm).

**Before** 

_**SQL Lab**_
Screenshot 2567-04-10 at 15 35 32<br/>
_**Table Chart**_
Screenshot 2567-04-10 at 15 36 02

**After**

_**SQL Lab**_

https://github.com/apache/superset/assets/96282005/b8e5161c-c7bb-4c72-b977-277cf52de259

_**Table Chart**_

https://github.com/apache/superset/assets/96282005/24c58851-c461-4b7d-bd47-49b9436b576b

### TESTING INSTRUCTIONS

**SQL Lab**
1. Enter a query that contains an HTML string.
2. Press the "Run" button.
3. The result should be rendered as HTML.
4. Try disabling the "Render HTML" option.
5. Run the query again.
6. The result should be rendered as a normal string.

**Table Chart**
1. Create a table chart with the column that contains HTML strings.
2. The result should be rendered as HTML.
3. Uncheck the "Render columns in HTML format" option in Customize panel.
4. Update the chart.
5. The result should be rendered as a normal string.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/ResultSet/index.tsx
+++ b/superset-frontend/src/SqlLab/components/ResultSet/index.tsx
@@ -53,6 +53,7 @@ import FilterableTable from 'src/components/FilterableTable';
 import CopyToClipboard from 'src/components/CopyToClipboard';
 import { addDangerToast } from 'src/components/MessageToasts/actions';
 import { prepareCopyToClipboardTabularData } from 'src/utils/common';
+import { getItem, LocalStorageKeys } from 'src/utils/localStorageHelpers';
 import {
   addQueryEditor,
   clearQueryResults,
@@ -579,6 +580,10 @@ const ResultSet = ({
       const expandedColumns = results.expanded_columns
         ? results.expanded_columns.map(col => col.column_name)
         : [];
+      const allowHTML = getItem(
+        LocalStorageKeys.SqllabIsRenderHtmlEnabled,
+        true,
+      );
       return (
         <ResultContainer>
           {renderControls()}
@@ -626,6 +631,7 @@ const ResultSet = ({
             height={rowsHeight}
             filterText={searchText}
             expandedColumns={expandedColumns}
+            allowHTML={allowHTML}
           />
         </ResultContainer>
       );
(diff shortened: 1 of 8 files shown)


superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -579,6 +580,10 @@ const ResultSet = ({
       const expandedColumns = results.expanded_columns
         ? results.expanded_columns.map(col => col.column_name)
         : [];
+      const allowHTML = getItem(
+        LocalStorageKeys.SqllabIsRenderHtmlEnabled,
+        true,
```
justinpark, line 585: @soniagtm The default fallback value for this option should be `false` since it's an optional allowance.

superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
```diff
@@ -284,6 +284,9 @@ const SqlEditor: React.FC<Props> = ({
   const [autocompleteEnabled, setAutocompleteEnabled] = useState(
     getItem(LocalStorageKeys.SqllabIsAutocompleteEnabled, true),
   );
+  const [renderHTMLEnabled, setRenderHTMLEnabled] = useState(
+    getItem(LocalStorageKeys.SqllabIsRenderHtmlEnabled, true),
```
justinpark, line 288: same here. Please set `false` for fallback case.
```suggestion
    getItem(LocalStorageKeys.SqllabIsRenderHtmlEnabled, false),
```

superset-frontend/src/SqlLab/components/ResultSet/index.tsx
soniagtm: I thought that we wouldn't want to change the current behavior of the result table (FilterableTable), as it currently always renders results in HTML if applicable. That's why I set the default value to true.

justinpark: That makes sense. LGTM

justinpark (approved):

PR #27979, opened 2024-04-11
chore(ci): make pre-commit step faster by skipping superset install

mistercrunch:
### SUMMARY
Title - turns out there's no need to pip install superset and its deps when simply running pre-commit

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -24,6 +24,8 @@ jobs:
           submodules: recursive
       - name: Setup Python
         uses: ./.github/actions/setup-backend/
+        with:
+          install-superset: 'false'
       - name: Enable brew and helm-docs
         # Add brew to the path - see https://github.com/actions/runner-images/issues/6283
         run: |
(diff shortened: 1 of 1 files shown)


villebro (approved):
Nice! LGTM!

.github/workflows/pre-commit.yml
```diff
@@ -35,6 +37,7 @@ jobs:
           brew install norwoodj/tap/helm-docs
       - name: pre-commit
         run: |
+          pip install pre-commit
```
john-bodley, line 40: Sadly @mistercrunch this is not right. We should never run `pip install` without a specific version as it _could_ result in a non-deterministic build when running CI. Instead we should install the frozen requirements in `requirements/development.txt`. 

Granted this brings in a lot of extra packages, but this was less of an issue when we had more environments including `requirements/integration.[in|txt]` which included only the subset of packages needed to run various CI commands.

villebro: @john-bodley maybe we should break out the light weight linting packages into a separate requirements file and depend on that in the development one?

mistercrunch: we can revert and load the whole bundle, not a biggy

mistercrunch: here -> https://github.com/apache/superset/pull/27999

mistercrunch: Or how about this!? https://github.com/apache/superset/pull/28000

Issue #20159, opened 2022-05-23
[SIP-82] - Improving Superset Theming


michael-s-molina:
### Motivation

Recently we did a POC with the objective of changing the colors of the Superset theme to analyze how it affected the application. The reason was to understand how far we are from allowing Superset users to customize the theme. The objective of this SIP is to document the findings of the POC and to propose some efforts to improve Superset theming capabilities. We understand that some of the efforts may need additional SIPs or are linked with existing SIPs. It’s not the objective of this SIP to dive into the details of all the efforts but to propose an overview of the work necessary and to propose solutions to some of them.

The Superset theme configuration has the following structure:

```
const defaultTheme = {
  borderRadius: 4,
  colors: {
    text: {...},
    primary: {...},
    secondary: {...},
    grayscale: {...},
    error: {...},
    warning: {...},
    alert: {...},
    success: {...},
    info: {...},
  },
  opacity: {...},
  typography: {
    families: {...},
    weights: {...},
    sizes: {...},
  },
  zIndex: {...},
  transitionTiming: 0.3,
  gridUnit: 4,
};
```

We started by changing the theme's `primary`, `secondary`, and `grayscale` colors to green, purple, and gold respectively using colors from Ant Design [color palettes](https://ant.design/docs/spec/colors).

Primary            |  Secondary            |  Grayscale
:-------------------------:|:-------------------------:|:-------------------------:
<img width="396" alt="Screen Shot 2022-03-16 at 1 03 27 PM" src="https://user-images.githubusercontent.com/70410625/169547836-42879303-6372-4c1e-a74e-c3c6acc91cd6.png">  |  <img width="393" alt="Screen Shot 2022-03-16 at 1 05 54 PM" src="https://user-images.githubusercontent.com/70410625/169547941-5634d980-aaf3-4a3b-bd63-7cee7216a141.png">  |  <img width="389" alt="Screen Shot 2022-03-16 at 1 04 36 PM" src="https://user-images.githubusercontent.com/70410625/169547983-7be5abb8-6673-4382-9c26-f772c925ec91.png">

We also changed text colors to magenta. Here's the result of these changes on some of our main screens:

<img width="1767" alt="Screen Shot 2022-03-16 at 1 15 00 PM" src="https://user-images.githubusercontent.com/70410625/169550483-f6cd9caf-7549-472d-926f-bb22a8e8b29f.png">
[...]

PR #27995, opened 2024-04-11
feat(explore): Color scheme groups, new color schemes

kgabryje:
### SUMMARY
- adds a bunch of new color schemes
- groups the color schemes in categories: Custom (for color schemes added by users in config), Featured and Other
- Adds option group handling to Select and AsyncSelect

https://github.com/apache/superset/assets/15073128/99ff3c75-4424-472d-b91d-6bf5e46157d7

### TESTING INSTRUCTIONS
1. Test Select and AsyncSelect components with groups in storybook
2. Add some custom color schemes in config
3. Go to Explore and open a viz that has categorical color scheme control (for example Pie chart)
4. Open color scheme dropdown
5. Verify that your custom color scheme is in Custom group, the rest of the color schemes should be in groups Featured or Other

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
@kgabryje I have some questions about how this behavior will work with some of the requirements/features of the Select component:
- Selected items are always sorted first. How will that work with groups?
- We allow new items. Where will they be placed?
- What's the interaction with Select All?

Have we considered implementing this outside of the Select component? Like a radio group that changes the values of the Select depending on the option that is selected?

If we are thinking specifically about color palettes where new values and select all are disabled, maybe this should be a specific component? 

If you ask me, given the complexity involved, I would implement this with a combination of radio group + Select.

michael-s-molina:
Maybe another question would be, does it really make a difference separating color pallets by featured / others / etc? If so, another possible alternative would be to render the elements with a `featured` label as we do for database types.

Screenshot 2024-04-11 at 16 43 39

kgabryje:
> * Selected items are always sorted first. How will that work with groups?
Good question. Currently I left it unsorted, but I was thinking about sorting the selected within the groups. The groups remain in their initial positions.
> * We allow new items. Where will they be placed?
New items do not belong to any group - they will be treated as regular (ungrouped) options
> * What's the interaction with Select All?
It simply selects all options

> 
> Have we considered implementing this outside of the Select component? Like a radio group that changes the values of the Select depending on the option that is selected?
> 
> If we are thinking specifically about color palettes where new values and select all are disabled, maybe this should be a specific component?
> 
> If you ask me, given the complexity involved, I would implement this with a combination of radio group + Select.

The option groups are supported by the native Antd Select, so in my opinion we should also support them in our customised Select. Right now the color scheme picker is our only use case for the groups, but I feel like grouping the options in categories is not such an uncommon thing to do and the need for it might pop up in the future

michael-s-molina:
> The option groups are supported by the native Antd Select, so in my opinion we should also support them in our customised Select. Right now the color scheme picker is our only use case for the groups, but I feel like grouping the options in categories is not such an uncommon thing to do and the need for it might pop up in the future

We can go down that road but keep in mind that complexity will increase given the other requirements we have and that the Ant Design select does not have such as sorting selected items and new items. If we can avoid that, it would be better. That's why I'm trying to see if there's an alternative before starting to analyze how to resolve all the questions I posted above.

kgabryje:
> If there's no alternative, and we really need to implement this behavior, then we need to think about how to break the select component because I don't how this behavior will work with:
> 
> > * Selected items are always sorted first. How will that work with groups?
> > * We allow new items. Where will they be placed?
> > * What's the interaction with Select All?

Ugh sorry, I answered those questions in my first response but messed up the formatting 🤦 

Pasting it here again:

> * Selected items are always sorted first. How will that work with groups?

Good question. Currently I left it unsorted, but I was thinking about sorting the selected within the groups. The groups remain in their initial positions.

> * We allow new items. Where will they be placed?

New items do not belong to any group - they will be treated as regular (ungrouped) options

> * What's the interaction with Select All?

It simply selects all options

michael-s-molina:
@kgabryje @geido Let's do a meeting to discuss this change as I'm worried about complexity increase.

kgabryje:
@michael-s-molina @geido Following our discussion from last week, I opted for "low effort" solution for now, which is using Antd Select for ColorSchemeControl instead of our custom Select component. We'll revisit this component in the near future and try to figure out if we can provide a better user experience by rewriting this control with different UI

michael-s-molina:
@kgabryje Did you forget to commit something? I'm still seeing some changes in the Select component such as group options.

kgabryje:
@michael-s-molina Thanks for spotting, I updated the PR. I left the styling for groups in `StyledSelect`, so that the component can be reused for grouped options

superset-frontend/src/components/Select/styles.tsx
```diff
@@ -55,6 +55,13 @@ export const StyledSelect = styled(AntdSelect, {
     .select-all {
       border-bottom: 1px solid ${theme.colors.grayscale.light3};
     }
+    .ant-select-item.ant-select-item-group {
```
michael-s-molina, line 58: I believe these should be moved to the `ColorSchemeControl`:

```
<StyledSelect
  css={css`
    width: 100%;
    <HERE>
  `}
```

superset-frontend/packages/superset-ui-core/src/color/colorSchemes/categorical/modernSunset.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import CategoricalScheme from '../../CategoricalScheme';
+import { ColorSchemeGroup } from '../../types';
+
+const schemes = [
+  {
+    id: 'modernSunset',
+    label: 'Modern sunset',
+    group: ColorSchemeGroup.Featured,
+    colors: [
```
michael-s-molina, line 28: I'm not sure if these should go into the `theme` object. Could you add a `TODO` comment to revisit this file when working on [SIP-82 Improving Superset Theming](https://github.com/apache/superset/issues/20159)?

superset-frontend/src/components/Select/styles.tsx
kgabryje: 👍

superset-frontend/packages/superset-ui-core/src/color/colorSchemes/categorical/modernSunset.ts
kgabryje: 👍

kgabryje: Do you think all the other colour scheme files (including the old ones) should be revisited? I can add todos to all of them so that we don't miss them

michael-s-molina: That would be great! Thank you!

michael-s-molina (approved):
LGTM. Thanks for addressing the comments @kgabryje!

geido (approved):

PR #27996, opened 2024-04-11
fix: Incorrect onChange value when an unloaded value is pasted into AsyncSelect

michael-s-molina:
### SUMMARY
Fixes a bug where the `onChange` event was being fired with an incorrect value when pasting a text that corresponds to an existing option in the database, which was not loaded previously due to pagination.

https://github.com/apache/superset/assets/70410625/b6f340ec-0e4a-450b-be38-e4b1181c5dda

https://github.com/apache/superset/assets/70410625/305eb0f0-74ac-4053-a836-f877df2bc951

### TESTING INSTRUCTIONS
1 - Use an async select that has enough values to trigger pagination. You can also reduce the Select's default page size to test.
2 - Paste a value that is not loaded yet by the Select component
3 - Check that the `onChange` event is called with the correct value. It's easier to check this when the options have numeric values. In that case, `onChange` should receive the numeric value instead of the pasted text as the value property.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Select/AsyncSelect.test.tsx
+++ b/superset-frontend/src/components/Select/AsyncSelect.test.tsx
@@ -384,12 +384,14 @@ test('removes duplicated values', async () => {
     },
   });
   fireEvent(input, paste);
-  const values = await findAllSelectValues();
-  expect(values.length).toBe(4);
-  expect(values[0]).toHaveTextContent('a');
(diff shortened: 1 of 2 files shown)


justinpark (approved):

kgabryje (approved):
LGTM

PR #27997, opened 2024-04-11
fix(ListView): avoid firing event on change of selection

justinpark:
### SUMMARY
When you paste a value in the list view filter, the input field automatically submits the input value right away.
This happens because the `onSelect` event is triggered when the `onChange` event occurs, instead of when the actual selection is made.

Before:

https://github.com/apache/superset/assets/1392866/1b319b5a-e339-4ab4-988b-990576b0b8d8

After:

https://github.com/apache/superset/assets/1392866/ee9bde52-6e6e-4632-ba60-d2582f1fa323

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/ListView/Filters/Select.tsx
+++ b/superset-frontend/src/components/ListView/Filters/Select.tsx
@@ -50,7 +50,7 @@ function SelectFilter(
 ) {
   const [selectedOption, setSelectedOption] = useState(initialValue);
 
-  const onChange = (selected: SelectOption) => {
+  const handleOnSelect = (selected: SelectOption) => {
     onSelect(
       selected ? { label: selected.label, value: selected.value } : undefined,
     );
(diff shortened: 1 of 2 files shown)


rusackas (approved):
LGTM! The failing test looks like an easy-peasy fix ;)

michael-s-molina (changes requested):
@justinpark When you paste something on the Select component, it's already selected. The change you did might confuse users because if you paste and then close the select, the filter won't be applied but the value will be selected.

What is the problem of triggering the query when you paste?

michael-s-molina:
> @justinpark When you paste something on the Select component, it's already selected. The change you did might confuse users because if you paste and then close the select, the filter won't be applied but the value will be selected.

What is the problem of triggering the query when you paste?

justinpark:
Closed as #28017 solved the issue

PR #27998, opened 2024-04-11
docs: move mp4 video to superset-site/tree/lfs

mistercrunch:
This big video file doesn't belong in the main repo! 

Moved the file here -> https://github.com/apache/superset-site/tree/lfs

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,8 +70,8 @@ Superset provides:
 ## Screenshots & Gifs
 
 **Video Overview**
-
-https://github.com/apache/superset-site/raw/lfs/superset-video-4k.mp4
+<!-- File hosted here https://github.com/apache/superset-site/raw/lfs/superset-video-4k.mp4 -->
+https://user-images.githubusercontent.com/64562059/234390129-321d4f35-cb4b-45e8-89d9-20ae292f34fc.mp4
 
 <br/>
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Superset provides:
 
 **Video Overview**
 
-https://user-images.githubusercontent.com/64562059/234390129-321d4f35-cb4b-45e8-89d9-20ae292f34fc.mp4
+https://github.com/apache/superset-site/raw/lfs/superset-video-4k.mp4
 
 <br/>
 
--- a/docs/src/pages/index.tsx
+++ b/docs/src/pages/index.tsx
@@ -644,7 +644,7 @@ export default function Home(): JSX.Element {
               </div>
             </Carousel>
             <video autoPlay muted controls loop>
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -70,8 +70,8 @@ Superset provides:
 ## Screenshots & Gifs
 
 **Video Overview**
-
-https://github.com/apache/superset-site/raw/lfs/superset-video-4k.mp4
+<!-- File hosted here https://github.com/apache/superset-site/raw/lfs/superset-video-4k.mp4 -->
```
rusackas, line 73: The comment helps immensely, but I still wonder if we should just maintain this in a subdirectory on the main branch.

rusackas: We would (perhaps) need to modify the docs-deploy script to not nuke that directory.

mistercrunch: I'm not sure if it matters where it lives, currently there's no coupling between the CDN (effectively githubusercontent.com) and the location in the repo. In some ways it's just a backup.

I don't think anyone would ever edit this file, maybe there's another file format with the source of the different clips/transitions/audio somewhere we could put in source control, but really this asset is probably fine in a CDN + backed in any location where it could be restored.

rusackas: One wrinkle (which could play out either way) is that the Superset/docs site has the video file included by filesystem path, not via CDN. We should either update the docs site to use Githubusercontent as part of this, or some other means to point the website to its new location, before the site breaks.

mistercrunch: I see the full link here-> https://github.com/apache/superset/blob/master/docs/docs/intro.mdx?plain=1#L23

rusackas: Oh... weird... well, ok. Nonetheless, we just need to make sure to update the docs site with the updated URL, whichever branch we stash large files on.

mistercrunch: Oh actually I think we don't need to change anything else. Videos only serve from the CDN, and the CDN is de-coupled from the repos. When you drag-n-drop a picture or a video in an issue comment, it'll upload to the CDN and serve in perpetuity as AFAIU. The move to the branch is just a backup in case. I haven't tried but I'm pretty sure if I re-upload the same file to the CDN it may have the same url it it's smart. They don't need to store the same binary twice, guessing the address is based on a hash of some kind.

rusackas: Right, that's fine for the readme, but when the file disappears, the docs site (Docusaurus) will break. It's this line right [here](https://github.com/apache/superset/blob/master/docs/src/pages/index.tsx#L647).

rusackas (approved):

mistercrunch:
I'll babysit the deploy and make sure the video works from CDN

PR #28000, opened 2024-04-11
feat: action-specific py deps pinning

mistercrunch:
Had this idea around addressing the comment here: https://github.com/apache/superset/pull/27979#pullrequestreview-1994777679 , introducing action-specific pinning.

I tested that this works nicely when running `pip-compile-multi -P pre-commit`, allowing us to bump pre-commit or any lib across the bundles.

Submitting for comment, but both reverting and installing `requirements/development.txt` seem reasonable, curious what others think.

Some thoughts:
- we want to insure consistency between development.txt and this new pin file (I think pip-complile-multi would ensure that from my understanding)
- we wouldn't want to install all of superset and its deps to support that one action ...


john-bodley:
@mistercrunch I'm not opposed to this, but it feels like we're returning to our previous state where we have multiple environments for exactly this reason. I think it might be good to take a step back and think about this problem holistically.

mistercrunch:
> going around in circles

I think we made some progress though along the way though
- we identified the blocking bump that was breaking CI
- upgraded ~50 python libs that were way behind
- we have a solution to replace dependabot with supersetbot - though we're holding back on the next breaking change window to continuously manage and upgrade python dependenciesi 
- untangled the `setup-backend` logic to a single reusable GHA that's aligned with the bundles we have
- aligned Dockerfile layers with reqs bundles

I agree that we still haven't settled on the ideal set of bundles. Maybe we went from too many to too few, but seems we can find the right balance. Likely will need a new bundle.

PR #28018, opened 2024-04-12
fix: bump client side chart timeouts to use the SUPERSET_WEBSERVER_TIMEOUT

eschutho:
### SUMMARY

We're seeing a lot of client side timeouts in annotation layers and for the chart/data endpoints. I'm not sure why we had the timeout set to only 1 minute, since the chart/data endpoints are traditionally some of the longest running queries. 
Screenshot 2024-04-12 at 1 58 22 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Chart/chartAction.js
+++ b/superset-frontend/src/components/Chart/chartAction.js
@@ -248,17 +248,20 @@ export async function getChartDataRequest({
 
 export function runAnnotationQuery({
   annotation,
-  timeout = 300,
+  timeout,
   formData = null,
   key,
   isDashboardRequest = false,
   force = false,
 }) {
   return function (dispatch, getState) {
-    const sliceKey = key || Object.keys(getState().charts)[0];
+    const { charts, common } = getState();
+    const sliceKey = key || Object.keys(charts)[0];
+    const queryTimeout = timeout || common.conf.SUPERSET_WEBSERVER_TIMEOUT;
+
     // make a copy of formData, not modifying original formData
     const fd = {
-      ...(formData || getState().charts[sliceKey].latestQueryFormData),
+      ...(formData || charts[sliceKey].latestQueryFormData),
     };
 
     if (!annotation.sourceType) {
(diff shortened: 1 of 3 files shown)


superset-frontend/src/components/Chart/chartAction.js
```diff
@@ -248,7 +248,7 @@ export async function getChartDataRequest({
 
 export function runAnnotationQuery({
   annotation,
-  timeout = 60,
+  timeout = 300,
```
eschutho, line 251: Another option is for us to make this timeout configurable in config.py like we do with the server timeouts.

sadpandajoe: +1 on putting this in config such that we can see timeouts in one place and if someone wants to configure it for their instance they just have to update in config.py vs having to find all places and update it there.

craig-rueda: Yea, would be nice to make this a config :)

eschutho: kk, I'll make that change then. 👍

superset-frontend/src/components/Chart/chartAction.js
```diff
@@ -248,17 +248,20 @@ export async function getChartDataRequest({
...
   formData = null,
   key,
   isDashboardRequest = false,
   force = false,
 }) {
   return function (dispatch, getState) {
-    const sliceKey = key || Object.keys(getState().charts)[0];
+    const {charts, common} = getState()
+    const sliceKey = key || Object.keys(charts)[0];
+    const queryTimeout = timeout || common.conf.SUPERSET_WEBSERVER_TIMEOUT;
```
eschutho, line 260: This config was already used elsewhere, so I decided to reuse it instead of creating a new one. Maybe in a follow up PR if we feel want the webserver and client timeouts to be different, we can update all of them. But for now, it's consistent with other parts of the js codebase.

superset-frontend/src/components/Chart/chartAction.js
rusackas: Would it be sensible to use one of the existing timeout configs rather than adding a new one?

dpgaspar (approved):

PR #28028, opened 2024-04-15
chore: stabilize MySQL tests by aligning isolation levels

mistercrunch:
I've been keeping an eye on our CI and noticed that a couple of Python tests like `test_query_cache_key_changes_when_datasource_is_updated` and `test_chart_data_async` keep tripping up on MySQL. I think it might be the different isolation levels messing things up compared to PostgreSQL.

So, I've tweaked our MySQL setup in the GitHub Actions to use 'READ-COMMITTED', just like Postgres. Hopefully, this gets rid of those annoying inconsistencies and smooths things out for our test runs.

following https://dev.mysql.com/doc/refman/8.3/en/set-transaction.html#:~:text=To%20set%20the%20global%20isolation,REPEATABLE%2DREAD%20%2C%20or%20SERIALIZABLE%20.

### DIFF
--- a/.github/workflows/bashlib.sh
+++ b/.github/workflows/bashlib.sh
@@ -89,6 +89,7 @@ EOF
 setup-mysql() {
   say "::group::Initialize database"
   mysql -h 127.0.0.1 -P 13306 -u root --password=root <<-EOF
+    SET GLOBAL transaction_isolation='READ-COMMITTED';
     DROP DATABASE IF EXISTS superset;
     CREATE DATABASE superset DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci;
     DROP DATABASE IF EXISTS sqllab_test_db;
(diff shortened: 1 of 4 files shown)


mistercrunch:
Didn't work! :( Still feel like there's something here, the intuition around "why is mysql the only one that flakes on these specific tests!?" and something around read uncommitted and/or isolation levels does seem to make sense. Could it be that I'm not setting the right flag? Not early enough? Doesn't stick somehow?

.github/workflows/bashlib.sh
```diff
@@ -89,6 +89,8 @@ EOF
 setup-mysql() {
   say "::group::Initialize database"
   mysql -h 127.0.0.1 -P 13306 -u root --password=root <<-EOF
+    SET GLOBAL transaction_isolation='READ-COMMITTED';
```
mistercrunch, line 92: NOTE: this didn't work first try, so trying `SET GLOBAL TRANSATION` bellow instead...

rusackas:
Thanks for looking, this has been driving me nuts.

mistercrunch:
I think the `time.sleep` did the trick, at least I was able to run manually 4 times without seeing an error. Might have been luck, but willing to try our luck with merging this, and keeping an eye on CI

rusackas (approved):

PR #28031, opened 2024-04-15
chore(translations): fix translations order

### LINKED ISSUE #27894: Extracted Translation Order
### Bug description

Hello, I’ve noticed that when following the translations extraction process described in https://superset.apache.org/docs/contributing/translations/ the order of the items in the .pot file changes. This leads to a huge git diff, that is basically impossible to merge, even for small translation changes. Would you accept a PR that simply runs the pybabel extract command and updates the translation .pot and .po files with the what I assume to be correct order?

### How to reproduce the bug

1. Run `pybabel extract -F superset/translations/babel.cfg -o superset/translations/messages.pot -k _ -k __ -k t -k tn -k tct .` as stated [in the docs](https://superset.apache.org/docs/contributing/translations)
2. Notice how the .pot file has a long list of changes even if no new messages were added.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.10

### Node version

16

### Browser

Chrome

### Additional context

I couldn't find any information on an expected order of the translation items but went through the git history and saw that for the most part it was in a different order.

### Checklist

lscheibel:
### SUMMARY
This PR does two things:
1. fixes a use of f-strings in sql_parse.py which causes pybabel to error. (See https://github.com/python-babel/babel/issues/715 for why.)
2. reorders the translation messages within the translation files by simply running the commands described in https://superset.apache.org/docs/contributing/translations/ which otherwise would have lead to large merge conflicts even for small translation changes.

This PR does not include any changes to the translations themselves.

### TESTING INSTRUCTIONS
When extracting the messages from using the [documented pybabel commands](https://superset.apache.org/docs/contributing/translations/), without any translation changes, the message.pot file should not change (except for the license information).

### ADDITIONAL INFORMATION

- [x] Has associated issue: #27894
- [ ] Required feature flags:
- [ ] Changes UI

Fixes #27894


superset/sql_parse.py
```diff
@@ -765,8 +765,8 @@ def _extract_tables_from_sql(self) -> set[Table]:
                 SupersetError(
                     error_type=SupersetErrorType.QUERY_SECURITY_ACCESS_ERROR,
                     message=__(
-                        f"You may have an error in your SQL statement. {message}"
-                    ),
```
rusackas, line 769: I wonder if @john-bodley knows a good way to lint against this. According to GPT, we can write a custom pylint checker, but it seems like that would be a first for this codebase. Something off the shelf to prevent this would be nice.

john-bodley: @rusackas we have custom (WIP) Pylint checkers. I guess one would need to add one to check whether there's an f-string being used within a translation `__(...)`.

sfirke (approved):

superset/sql_parse.py
rusackas: Oh, could you point me to where those live, @john-bodley, and I'll try to GPT us out of this issue?

rusackas:
Running CI 🤞

rusackas (approved):

lscheibel:
Thanks for taking a look everyone. I fixed the line endings of the .po and .pot files, which were failing the CI.

rusackas:
Running Ci... then we'll merge it!

PR #28035, opened 2024-04-15
feat: Add Czech Republic country map.

martinspudich:
### SUMMARY

Add Czech Republic country map

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI
- [ ] Includes DB Migration (follow approval process in https://github.com/apache/superset/issues/13351)


rusackas:
Hi! These need to be added to the Jupyter notebook (which might be a one-line edit, adding the country name), as the geojson files and TS files are generated from there. If we don't do that, these files won't be maintained, and may be overwritten.

martinspudich:
@rusackas Hi, I add 'czech republic' to Jupyter notebook.

mistercrunch (approved):

rusackas:
> Hi, I add 'czech republic' to Jupyter notebook.

Thanks! There's another PR I probably ought to merge first... gimme a little while here...

rusackas (approved):

RIS3cz:
No Czech Republic map in v 4.0.2. Why?

sfirke:
4.0.2 is a patch/bugfix release, it only pulls in cherries that fix bugs.  This is considered a feature and so will be in the 4.1.0 release, which should be cut this week for testing.  Please stayed tuned to the 4.1.0 Release Candidate voting period during which you can test out 4.1.0 RC1 and make sure this works how you expect.

RIS3cz:
> 4.0.2 is a patch/bugfix release, it only pulls in cherries that fix bugs. This is considered a feature and so will be in the 4.1.0 release, which should be cut this week for testing. Please stayed tuned to the 4.1.0 Release Candidate voting period during which you can test out 4.1.0 RC1 and make sure this works how you expect.

Dear @sfirke, well, when I think about the versions and I know the way it works with the releases, I can understand it. But when I think in the way of common sense and practical usage, I do not understand why in the same time Krymea has returned back to Ukraine (which is another matter and I do not comment) at same time the suvereign countries, members of the EU, Czechia and Slovakia, being previously Czechoslovakia have not managed to be added to the countires folder of superset-frontend. I believe it will be repaired in close future. Thank You.

PR #28039, opened 2024-04-15
fix: support docker/.env-local for docker-compose

### LINKED ISSUE #28026: docker\.env-local not found
### Bug description

quickstart 

docker-compose -f docker-compose-image-tag.yml up

docker\.env-local not found

### How to reproduce the bug

https://superset.apache.org/docs/quickstart

git clone https://github.com/apache/superset

docker-compose -f docker-compose-image-tag.yml up

docker\.env-local not found

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

mistercrunch:
As reported here https://github.com/apache/superset/issues/28026, in https://github.com/apache/superset/pull/27953 I added support for a supplemental `docker/.env-local` override file, not thinking it was going to be required upon starting docker-compose. I thought I had tested that while switching branches but somehow didn't test that docker-compose would simply boot up in the absence of that file.

BUT! luckily there's a new feature as of docker-compose `2.24.0` where there's a way to specify required and not-required files here, so I'm taking advantage of that in this correction PR. Kudos to @artofcomputing for finding this new feature and for pointing this out!

I also added a bit of documentation to point to an existing mechanism around providing a `docker/pythonpath_dev/superset_config_docker.py` file, which existed in the logic but wasn't clearly documented. This
solves for my use case where I was working a PR adding features around Slack, where I wanted to provide an API key configuration without risks of commiting/pushing it back to the repo.

closes https://github.com/apache/superset/issues/28026

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,7 +110,6 @@ release.json
 messages.mo
 
 docker/requirements-local.txt
-docker/.env-local
 
 cache/
 docker/*local*
--- a/docker-compose-image-tag.yml
+++ b/docker-compose-image-tag.yml
@@ -34,10 +34,7 @@ services:
 
   db:
     env_file:
-      # defaults
       - docker/.env
-      # gitignored overrides
-      - docker/.env-local
     image: postgres:15
     container_name: superset_db
     restart: unless-stopped
@@ -47,10 +44,7 @@ services:
 
   superset:
     env_file:
-      # defaults
       - docker/.env
-      # gitignored overrides
-      - docker/.env-local
     image: *superset-image
     container_name: superset_app
     command: ["/app/docker/docker-bootstrap.sh", "app-gunicorn"]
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -110,7 +110,6 @@ release.json
 messages.mo
 
 docker/requirements-local.txt
-docker/.env-local
```
mistercrunch, line 113: turns out that's covered by `docker/*local*` bellow

rusackas (approved):
LGTM - I like the new syntax/optionality model better :)

mistercrunch:
docker-compose is getting very usable for dev workflows - having it boot fast from my previous PR makes it much more viable too

artofcomputing (commented):
LGTM, ran locally to test it, environment was up fast 🚀

mistercrunch:
NICE! I'm feeling like CI / docker / docker-compose is getting to a pretty good place.

PR #28054, opened 2024-04-16
fix(Dashboard): Support "Edit chart" click on a new window

geido:
### SUMMARY
This PR https://github.com/apache/superset/pull/26138 removed the link from the "Edit chart" menu item on dashboards to enhance accessibility but it caused the chart to open in a new window. The goal of this PR is to support opening the chart in the same window by default while still allowing for opening a chart in a separate window with `cmd + click`.

It also removes an `onBlur` handler that was causing the interruption of some menu item actions by closing the dropdown abruptly. 

N.A.

### TESTING INSTRUCTIONS
1. Click on "Edit chart" in a Dashboard
2. The chart should open in the same window
3. cmd + click, the chart should open in a new window

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
+++ b/superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
@@ -529,6 +529,7 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
   const [openScopingModal, scopingModal] = useCrossFiltersScopingModal(
     props.slice.slice_id,
   );
+  const history = useHistory();
 
   const queryMenuRef: RefObject<any> = useRef(null);
   const menuRef: RefObject<any> = useRef(null);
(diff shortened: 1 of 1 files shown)


eschutho (approved):

superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -590,7 +591,12 @@ const SliceHeaderControls = (props: SliceHeaderControlsPropsWithRouter) => {
       case MenuKeys.ExploreChart:
         // eslint-disable-next-line no-unused-expressions
         props.logExploreChart?.(props.slice.slice_id);
-        window.open(props.exploreUrl);
+        if (domEvent.metaKey || domEvent.ctrlKey) {
```
rusackas, line 594: A few questions here:
1) Are we sure this works for mac and PC browsers as expected? Seems like it should, but ¯\\\_(ツ)_/¯ 
2) Do we need tests to prevent this regressing again? 
3) Is there a reason we can't just use an HTML link here? I feel like we had that solved at one point.
4) Is it not amazing that this particular link has been the subject of like 20 PRs and Issues?

geido: 1. Yes it should
2. Let me see what kind of tests I can add here
3. We now enable dropdown navigation for accessibility by controlling onClick handlers
4. Tests should hopefully fix this

geido: I will add the unit tests in a follow-up PR to unblock this bug fix

Issue #22862, opened 2023-01-26
[SIP-95] Proposal for Catalog Support in SQL Lab


JELGT2011:
### Motivation

The current SQL Lab explorer is not very intuitive as there is no way to explore catalogs, and can only explore the `default` catalog.

### Proposed Change

Add a new `Catalog` dropdown in the sqllab editor (similar to schema).

<img width="406" alt="image" src="https://user-images.githubusercontent.com/4592187/214722967-cb4ad879-d460-402c-a799-395fbaf6b01a.png">

### New or Changed Public Interfaces

This should only affect the sqllab query tab. The `Catalog` dropdown will be similar enough to the `Schema` explorer. The word `schema` is kind of overloaded throughout Superset, so it's not entirely clear to me yet whether a new endpoint is necessary.

### New dependencies

None

### Migration Plan and Compatibility

None

### Rejected Alternatives

None

betodealmeida:
I would love to see this, but there are few complications. The biggest one is that we need support from the SQL Alchemy dialects, and I'm not sure how many support dynamic catalogs. For example, both Hive and Trino require the catalog to be specified when the connection is created:

- `trino://<username>:<password>@<host>:<port>/<catalog>/<schema>`
- `hive://<host>:<port>/<catalog>`

For engines like this the user would have to choose a default catalog when the database is added, and each DB engine spec in Superset would have to know how to replace the catalog in the URL, in case a user chooses one different from the default after the DB has been created.

It's definitely doable, and something we should do. We've been slowly adding support for catalogs:

https://github.com/apache/superset/blob/383313b105b0e82bea0f38cc971630eded5affe0/superset/sql_parse.py#L172

Are you planning to work on this, or are you just proposing it? If you're not going to work on it I can start taking a look.

JELGT2011:
I was planning to work on this, but of course I’d love help as I’ve never made a contribution to Superset. I was planning on doing a smaller task first (adding a configurable header/footer) since I figured this one might be complex.

betodealmeida:
> I was planning to work on this, but of course I’d love help as I’ve never made a contribution to Superset. I was planning on doing a smaller task first (adding a configurable header/footer) since I figured this one might be complex.

Ah, great! From the wording of the SIP I wasn't sure, and people are welcome to propose changes even if they're not going to work on it. I'm happy to help in any way you need, let me know!

afilipchik:
hey @betodealmeida, good to see you are still pushing Superset forward!

[...]

PR #28063, opened 2024-04-16
feat(SIP-95): new endpoint for extra table metadata

betodealmeida:
### SUMMARY

I started working on [SIP-95](https://github.com/apache/superset/issues/22862), which introduces catalogs to Superset. The introduction of a new hierarchical level, catalog → schema → table, means that some API endpoints need to be updated to support a new parameter.

One of these API endpoints is used to fetch extra metadata about a given table, and the route looks like this:

`/<int:pk>/table_extra/<path:table_name>/<schema_name>/`

There are a few problems with the current route:

- `table_name` is a `path`, because table names can contain slashes.
- `schema_name` is often passed as `null` or `undefined`, so we need special logic to convert it to `None` in the backend.
- `schema_name` should never contain slashes, otherwise the parsing will be wrong.

Because of this, instead of enhancing the existing endpoint to add support for catalogs, I decided to create a new one:

`/<int:pk>/table_metadata/extra/?table=t[&schema=s[&catalog=c]]`

In this new endpoint, `table`, `schema`, and `catalog` are all passed as query parameters. This allows slashes in all of them, and prevents polluting logs with confusing URLs such as `/api/v1/databses/1/table_extra/my_table/undefined/`. Only `table` is required, the other two are optional.

The API calls a new method `get_extra_table_metadata` in the corresponding DB engine spec. The new method takes a `Table` object, which contains a name and optional schema/catalog. To prevent breaking any 3rd-party DB engine specs the new method will issue a warning and call `extra_table_metadata` if it exists and no catalog was specified.

Note that while the old endpoint is not removed in this PR, to prevent breaking changes, the frontend code was updated to use the new endpoint.

In the future, we should remove the old endpoint and also the `extra_table_metadata` method. @michael-s-molina, what do you think of this approach of introducing a new API? Should we mark it as deprecated now and remove it in the next release, or should we mark is as deprecated in 5.0 and remove it in 6.0?

N/A

### TESTING INSTRUCTIONS

I updated existing tests and added new unit tests.

### ADDITIONAL INFORMATION

- [X] Has associated issue: https://github.com/apache/superset/issues/22862
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
> Should we mark it as deprecated now and remove it in the next release, or should we mark is as deprecated in 5.0 and remove it in 6.0?

You can mark it as deprecated now. Please add a proposal [here](https://github.com/orgs/apache/projects/345) to remove the old endpoint in 5.0. 

> In this new endpoint, table, schema, and catalog are all passed as query parameters. This allows slashes in all of them, and prevents polluting logs with confusing URLs such as /api/v1/databses/1/table_extra/my_table/undefined/. 

Nesting resources or not is a architectural decision, with [trade-offs](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Best-Practices-for-Sub-and-Nested-Resources/), that we need to make consistently in our API. Because we built the API one endpoint at a time, we don't have consistency in nomenclature, response types, params, etc. I'm hoping to fix all these problem with v2 where we would analyze all existing endpoints and propose a consistent API in a SIP. For that reason, any pattern you choose now is acceptable and will be reviewed in v2.

betodealmeida:
Thanks, @michael-s-molina!

> Nesting resources or not is an architectural decision, with [trade-offs](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Best-Practices-for-Sub-and-Nested-Resources/), that we need to make consistently in our API. Because we built the API one endpoint at a time, we don't have consistency in nomenclature, response types, params, etc. I'm hoping to fix all these problems with v2 where we would analyze all existing endpoints and propose a consistent API in a SIP. For that reason, any pattern you choose now is acceptable and will be reviewed in v2.

Oh, I'm not against nesting resources in general. The problem here is that table names can contain slashes, and I think we should assume that schemas and catalogs can too. Imagine we have this route:

`/some/api/<path:catalog>/<path:schema>/<path:table>/`

And we have a table with slashes in the catalog, schema, and name, which we escape with `%2F` when doing a request:

`/some/api/my%2Fcatalog/my%2Fschema/my%2Ftable/`

What the Flask router sees in the request is:

`/some/api/my/catalog/my/schema/my/table/`

Which gets mapped to:

- catalog: `my`
- schema: `catalog`
- table: `my/schema/my/table`

So we'd have to **double escape** the slashes:

`/some/api/my%252Fcatalog/my%252Fschema/my%252Ftable/`

And the route would have unescape them back. For cases like this I think it's easier to pass the parameters as query arguments:

`/some/api/?table=my/table&schema=my/schema&catalog=my/catalog`

While longer, I think it's easier to read. And when some of those parameters are missing it gets better:

`/some/api/?table=my/table`

vs.

`/some/api/undefined/undefined/my%252Ftable/`

So something to think about when designing v2.

michael-s-molina:
> Oh, I'm not against nesting resources in general. The problem here is that table names can contain slashes, and I think we should assume that schemas and catalogs can too.
> So something to think about when designing v2.

Yep! That's indeed a requirement we should consider when designing v2.

superset/databases/api.py
```diff
@@ -812,13 +831,91 @@ def table_extra_metadata(
...
+              $ref: '#/components/responses/401'
+            404:
+              $ref: '#/components/responses/404'
+            500:
+              $ref: '#/components/responses/500'
+        """
+        self.incr_stats("init", self.table_extra_metadata.__name__)
+
+        database = DatabaseDAO.find_by_id(pk)
+        if database is None:
```
john-bodley, line 902:
```suggestion
        if not (database := DatabaseDAO.find_by_id(pk)):
```

superset/databases/schemas.py
```diff
@@ -1195,3 +1195,27 @@ class OAuth2ProviderResponseSchema(Schema):
...
+
+class QualifiedTableSchema(Schema):
+    """
+    Schema for a qualified table reference.
+
+    Catalog and schema can be ommited, to fallback to default values. Table name must be
+    present.
+    """
+
+    table = fields.String(
```
john-bodley, line 1208: I know I used the term `table` in the `Table` class, but I wonder if this should be name given it's referring to the name of the table. This is also consistent with your description on line 1210.

superset/db_engine_specs/base.py
```diff
@@ -1034,21 +1035,33 @@ def normalize_indexes(cls, indexes: list[dict[str, Any]]) -> list[dict[str, Any]
         return indexes
 
     @classmethod
-    def extra_table_metadata(  # pylint: disable=unused-argument
+    def get_extra_table_metadata(  # pylint: disable=unused-argument
         cls,
         database: Database,
-        table_name: str,
-        schema_name: str | None,
+        table: Table,
```
john-bodley, line 1041: Nice!

superset/db_engine_specs/presto.py
```diff
@@ -1224,14 +1225,20 @@ def expand_data(  # pylint: disable=too-many-locals
...
+        database: Database,
+        table: Table,
     ) -> dict[str, Any]:
         metadata = {}
 
-        if indexes := database.get_indexes(table_name, schema_name):
+        if indexes := database.get_indexes(table.table, table.schema):
             col_names, latest_parts = cls.latest_partition(
-                table_name, schema_name, database, show_first=True, indexes=indexes
+                table.table,
```
john-bodley, line 1237: Should we be passing a `Table` instance everywhere?

betodealmeida: We should! I'm doing that in the next PR! 😊

superset/databases/api.py
betodealmeida: Ah, thanks!

superset/databases/schemas.py
betodealmeida: Yeah, I think the current terminology is confusing, there's a lot of `table.table` in the codebase. Let me change it.

superset/databases/api.py
```diff
@@ -812,13 +831,92 @@ def table_extra_metadata(
...
+        try:
+            parameters = QualifiedTableSchema().load(request.args)
+        except ValidationError as ex:
+            raise InvalidPayloadSchemaError(ex) from ex
+
+        table = Table(parameters["name"], parameters["schema"], parameters["catalog"])
+        try:
+            security_manager.raise_for_access(database=database, table=table)
+        except SupersetSecurityException as ex:
+            # instead of raising 403, raise 404 to hide table existence
```
dpgaspar, line 915: nice!

superset/db_engine_specs/base.py
```diff
@@ -1034,21 +1035,33 @@ def normalize_indexes(cls, indexes: list[dict[str, Any]]) -> list[dict[str, Any]
...
 
         :param database: Database instance
-        :param table_name: Table name
-        :param schema_name: Schema name
+        :param table: A Table instance
         :return: Engine-specific table metadata
         """
-        # TODO: Fix circular import caused by importing Database
+        # old method that doesn't work with catalogs
+        if hasattr(cls, "extra_table_metadata"):
```
dpgaspar, line 1051: optional: 
do:
``` python
if not hasattr(cls, "extra_table_metadata"):
    return {}
...
```
to reduce nesting

dpgaspar (approved):

superset/db_engine_specs/base.py
betodealmeida: Nice, I'll fix it in the next PR!

PR #28095, opened 2024-04-16
fix(superset-frontend): resolve installation step problem by using node > 18

### LINKED ISSUE #28094: Can't install superset-frontend dependencies by using node v20.11.1 and yarn v1.22.22
## Introduction
I use node v20.11.1 and yarn v1.22.22 for my project. And represented the superset-frontend directory as a workspace module, by using a custom bash script (which called on the preinstall step):
* `clone-subdir.sh`
```bash
#!/bin/bash
REPO_NAME="$(echo $2 | grep -oE '[^/]+$')";
git clone --filter=blob:none --no-checkout $2 $REPO_NAME
cd $REPO_NAME;
git sparse-checkout set --no-cone "$1/*"
if [ -n "$3" ]; then
    git pull origin $3;
    git checkout $3;
else
    git fetch origin;
    git checkout main
    [[ $? != 0 ]] && git checkout master;
fi
```
* `package.json`
```json
{
  "name": "superset_wrapper",
  "workspaces": ["superset/superset-frontend"],
  "private": true,
  "scripts": {
    "preinstall": "./scripts/clone-subdir.sh superset-frontend git@github.com:apache/superset master"
  },
  "dependencies": {
    "axios": "^1.6.8"
  }
}
```

## Problem
It works fine. But the problem happens on the installation step. I got the next error log at the installation of the superset-frontend dependencies:
```bash
error superset@0.0.0-dev: The engine "node" is incompatible with this module. Expected version "^18.19.1". Got "20.11.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

## Solution
The reason why it happens is the concrete node's version which is set at the `engine`'s field. It will be better to replace the current value with a logical expression like: `"node": ">= 18"`. It will save backward compatibility with the v18 and allow to work with the frontend part by using newer versions.

### How to reproduce the bug

1. Use node v20 and yarn v1 for this case
2. Create new project
3. Initialize package.json (with `"private": true` field)
4. Set `preinstall` script
5. Set workspaces field
6. Run `yarn install`

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

Not applicable

### Node version

18 or greater

### Browser

Not applicable

### Additional context

_No response_

### Checklist

corocoto:
### SUMMARY

Sets a logical expression `"node": ">= 18"` at the `engines` field. It will save backward compatibility with the v18 and allow to work with the frontend part by using newer versions.

### ADDITIONAL INFORMATION

Fixes #28094


rusackas:
Closing this out since we're on node 20 currently, and many packages won't work with 18.

PR #28105, opened 2024-04-17
feat: new Excel upload form and API

dpgaspar:
### SUMMARY
Continuing the work on deprecating/removing server side rendered pages, this adds a new React/Antd Form and backend API for Excel uploads.

Leverages the work already done here: #27840

Important to note that, we continue to offer a file preview functionality done on the browser. On the CSV case we load a limited initial chunk of the file and infer the column names from there. This time we have to load the entire Excel file, this unfortunately blocks the browser while loading on big files.

Another approach would be to immediately upload the file to the backend to a new API that would return all existing sheets and columns from the Excel file.

There are pros and cons on each approach, processing the file on the browser avoids uploading and processing the file twice on the server. Doing an initial upload offers a better user experience on big files.

I've added a switch so that users can opt out of the preview. Users can still manually insert column names and sheet names.

https://github.com/apache/superset/assets/4025227/c1e88eac-34a3-4131-9af7-717512f0aecc

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/views/database/views.py
```diff
@@ -154,141 +152,6 @@ def this_form_post(self) -> Any:
         )
 
 
-class ExcelToDatabaseView(SimpleFormView):
-    form = ExcelToDatabaseForm
-    form_template = "superset/form_view/excel_to_database_view/edit.html"
```
EugeneTorap, line 159: You can remove this `html` file and also `csv` `html` files

betodealmeida (approved):
Looks great, it's an awesome improvement!

It would be nice if there was just one action for uploading files, and we'd figure out the filetype and handle it accordingly. Especially if we add support for more types in the future.

superset-frontend/package-lock.json
```diff
@@ -15574,6 +15585,82 @@
         "url": "https://opencollective.com/storybook"
       }
     },
+<<<<<<< HEAD
```
betodealmeida, line 15588: Oops!

craig-rueda (approved):
Looks good - still a little weary about uploading/parsing on the client-side, but I think we can just see how it goes...

PR #28106, opened 2024-04-17
fix: db migrations on downgrade

dpgaspar:
### SUMMARY
Currently when downgrading migrations we get the following error:

```
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
ERROR [flask_migrate] Error: Ambiguous walk
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


craig-rueda (approved):

mistercrunch:
I think I generated those, my bad. Guessing the `alembic merge` isn't the right way to go (?) I'll try and line up my migrations  in the future. 

Should we add a unit GHA that tests the downgrade?

dpgaspar:
> I think I generated those, my bad. Guessing the `alembic merge` isn't the right way to go (?) I'll try and line up my migrations in the future.
> 
> Should we add a unit GHA that tests the downgrade?

no problem! a downgrade test sounds good to me

PR #28130, opened 2024-04-18
docs: add dynamic entity-relationship diagram to docs

mistercrunch:
Screenshot 2024-04-18 at 11 38 26 AM

Recently I stumbled upon plantuml and its ability to render entity relationship diagrams (and other various UML) defined as code. Originally started as a PoC one-off, but it's looking pretty good and could be added to our doc builds.

I wanted to visualize the ERD, and thought it'd be a great asset for contributors to be able to refer to it.

In this PR:
- I wrote a script that introspects the models in Superset
  and use jinja2 to render a `erd.plantuml`, and a `erd.svg`
- Adding a interactive (zoom/pan) browser to browse the svg.

Looking for input as to whether we want this in the repo / docs.


rusackas:
You'll have to add an exception (and comment) in the Depndendy Review yaml file that states we're using this under the BSD license, which I think is Apache 2.0 compatible.

.github/workflows/superset-docs-deploy.yml
```diff
@@ -39,6 +39,10 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: '18'
+      - name: Setup Python
```
mistercrunch, line 42: planning on YOLO this script as it runs on master merges

.rat-excludes
```diff
@@ -66,3 +66,8 @@ google-big-query.svg
 google-sheets.svg
 postgresql.svg
 snowflake.svg
+
+# docs-related
+docs/docs/.htaccess*
```
rusackas, line 71: doesn't seem like we should need this line... wondering what might've changed here.

.rat-excludes
```diff
@@ -66,3 +66,8 @@ google-big-query.svg
...
+erd.plantuml
+erd.svg
```
rusackas, line 73: We could probably add the license boilerplate to the SVG as part of the action, if that makes sense to do so.

docs/package.json
```diff
@@ -39,6 +39,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-github-btn": "^1.4.0",
+    "react-svg-pan-zoom": "^3.12.1",
```
rusackas, line 42: Ya know what might make a neat/easy Superset plugin?

scripts/erd.py
```diff
@@ -0,0 +1,204 @@
...
+def erd(output: Optional[str] = None) -> None:
+    """
+    Generates a PlantUML ERD of the models/database
+
+    Parameters:
+    -----------
+    output: str, optional
+        File to write the ERD to, defaults to erd.plantuml if not provided
+    """
+    output = output or "./erd.plantuml"
```
rusackas, line 193: Maybe splice in the ASF license here?

scripts/erd.py
```diff
@@ -0,0 +1,204 @@
...
+    file_path: str
+        File path to write the ERD to
+    """
+    data = introspect_models()
+    templates_path = os.path.join(os.path.dirname(__file__), "templates")
+    env = jinja2.Environment(loader=jinja2.FileSystemLoader(templates_path))
+
+    # Load the template
+    template = env.get_template("erd.plantuml.template")
+    rendered = template.render(data=data)
```
rusackas, line 171: Maybe splice in the ASF license boilerplate here?

scripts/erd/erd.py
```diff
@@ -0,0 +1,204 @@
...
+    Returns:
+    --------
+    Dict[str, List[Dict[str, Any]]]: data structure for jinja2 template
+    """
+    data: dict[str, list[dict[str, Any]]] = defaultdict(list)
+    seen_models: set[str] = set()
+    for model in db.Model.registry.mappers:
+        group_name = (
+            TABLE_TO_GROUP_MAP.get(model.mapper.persist_selectable.name)
+            or "Other Models"
```
rusackas, line 149: Should it be "other" (a bit of a junk drawer) or "Uncategorized Models" implying tech debt to be swept up?

rusackas (approved):
Seems generally good to me, with the caveat that we might want to avoid rat exclusions if possible. Also wondering if there's a way to build the diagram only if files with models/migrations change. But these are optimizations, not blockers.

J2085isa (approved):

PR #28132, opened 2024-04-18
fix: call render operator function

eschutho:
### SUMMARY
Found a small bug where the render operator function was being passed in instead of executed and so the operator was undefined. I just renamed the variable so that it would sound more like a method to make it easier to catch in the future. Also added some tests. 

Before: 
image

After: 
Screenshot 2024-04-18 at 2 30 39 PM

### TESTING INSTRUCTIONS
Add a color condition to a column on the table viz

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sadpandajoe (approved):
LGTM

rusackas (approved):
Thanks!

PR #28134, opened 2024-04-18
chore: clean up console upon firing up the CLI

mistercrunch:
### Summary

Currently there's some irrelevant DEBUG-level type logging happening when using the CLI, here I'm doing the following things:
- setting the default logging level to INFO, which is the normal/sane default usually used in software
- downgrading unimportant messages in the console from INFO to DEBUG
- adding a dash of color on the title (bolded-white) and config loading INFO (cyan)
- turning off load limiter `RATELIMIT_ENABLED` to `False` when not in a `production` environment (added a note to `UPDATING.md` for that)

### Screenshots
Screenshot 2024-04-18 at 3 34 24 PM
Screenshot 2024-04-18 at 3 30 46 PM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -95,17 +95,3 @@ repos:
     hooks:
       - id: helm-docs
         files: helm
-  - repo: local
-    hooks:
-      - id: pylint
-        name: pylint
-        entry: pylint
-        language: system
-        types: [python]
-        args:
-          [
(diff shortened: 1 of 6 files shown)


eschutho (approved):
LGTM

jessie-ross:
Hi, I have a similar PR which is a more comprehensive take on defaulting to info logs - we hard-code debug log level outside of the `config.py` script, so we need to clean that up too.

PR #28135, opened 2024-04-18
chore: get websocket service to start in docker-compose

mistercrunch:
When starting vanilla docker-compose, this shows up ->

Screenshot 2024-04-18 at 4 03 15 PM

What's happening is upon start up, the superset-websocket service requires a config.json that contains a secret key (much like we set a SECRET_KEY up for flask on the python side). In normal operation, the administrator needs to provide such a file by copying and altering `superset-websocket/config.example.json`.

For `docker-compose`, now that we're 100% clear in our docs and files that our provided setup should not be used for production, we want to provide a dummy config file so that the service can start, as opposed to the confusion of starting a service that crashes upon start and spits a wall of confusing JSON.

Note that I've also:
- improved the error message for when the secret is too simple
- copied the example file into `docker/superset-websocket/config.json`
- altered the password with a dummy secret

### DIFF
--- a/superset-websocket/src/index.ts
+++ b/superset-websocket/src/index.ts
@@ -100,7 +100,9 @@ if (startServer && opts.jwtSecret.length < 32) {
 }
 
 if (startServer && opts.jwtSecret.startsWith('CHANGE-ME')) {
-  console.warn('WARNING: it appears you secret in your config.json is insecure');
+  console.warn(
+    'WARNING: it appears you secret in your config.json is insecure',
+  );
   console.warn('DO NOT USE IN PRODUCTION');
 }
 
--- a/superset-websocket/src/index.ts
+++ b/superset-websocket/src/index.ts
@@ -94,8 +94,15 @@ export const statsd = new StatsD({
 });
 
 // enforce JWT secret length
-if (startServer && opts.jwtSecret.length < 32)
-  throw new Error('Please provide a JWT secret at least 32 bytes long');
+if (startServer && opts.jwtSecret.length < 32) {
+  console.error('ERROR: Please provide a JWT secret at least 32 bytes long');
+  process.exit(1);
+}
+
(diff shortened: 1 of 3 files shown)


superset-websocket/src/index.ts
```diff
@@ -94,8 +94,17 @@ export const statsd = new StatsD({
...
-if (startServer && opts.jwtSecret.length < 32)
-  throw new Error('Please provide a JWT secret at least 32 bytes long');
+if (startServer && opts.jwtSecret.length < 32) {
+  console.error('ERROR: Please provide a JWT secret at least 32 bytes long');
+  process.exit(1);
+}
+
+if (startServer && opts.jwtSecret.startsWith('CHANGE-ME')) {
+  console.warn(
+    'WARNING: it appears you secret in your config.json is insecure',
```
dpgaspar, line 104: On our first attempt to tackle the default `SECRET_KEY` issue, we did a similar thing on the Flask backend, and it was proven that many admins ignore these warnings even when the log warning is a large banner. 

Can we do it in such a way that is a specific config for docker-compose only? `docker/.env` holds many secret defaults but it's only used by docker-compose

mistercrunch: The way I did things here this file in the repo `./docker/superset-websocket/config.json` is only mounted in docker-compose. If you go raw docker or helm it wouldn't get picked up.

For clarity though now I wonder if we should create a new `docker/compose` folder for things meant only for docker-compose to use (?). Currently it's mostly resources for docker-compose, but some things might be reused in Dockerfile as entrypoints or by Helm(?)

dpgaspar: oh right, my comment was too early in the morning ;)

`docker/compose` seems like a good idea

dpgaspar (approved):

PR #28137, opened 2024-04-19
chore: add pylint to pre-commit hook

mistercrunch:
I got bit by the old pylint failing in CI again recently. pylint is too slow for IDEs
so I turn it off since it made vim too sluggish to my taste. Feels about
right as a pre-commit hook, I'd rather get that too take a moment than
waiting for a CI cycle.

Not sure why it wasn't there in the first place, too slow?

We should also consider moving to `ruff` and streamlining all this, but ruff
doesn't support all the pylint rules [yet].

### DIFF
--- a/.github/workflows/superset-python-misc.yml
+++ b/.github/workflows/superset-python-misc.yml
@@ -31,10 +31,6 @@ jobs:
       - name: Setup Python
         uses: ./.github/actions/setup-backend/
         if: steps.check.outputs.python
-      - name: pylint
-        if: steps.check.outputs.python
-        # `-j 0` run Pylint in parallel
-        run: pylint -j 0 superset
 
   babel-extract:
     runs-on: ubuntu-20.04
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,6 +84,7 @@ repos:
         entry: pylint
         language: system
         types: [python]
+        exclude: ^(tests/|superset/migrations/)
         args:
           [
             "-rn", # Only display messages
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -88,5 +88,4 @@ repos:
           [
             "-rn", # Only display messages
             "-sn", # Don't display the score
-            "--rcfile=.pylintrc", # Link to your config file
           ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,6 +89,4 @@ repos:
             "-rn", # Only display messages
             "-sn", # Don't display the score
             "--rcfile=.pylintrc", # Link to your config file
-            "--load-plugins=pylint.extensions.docparams", # Load an extension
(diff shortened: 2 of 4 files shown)


.github/workflows/superset-python-misc.yml
```diff
@@ -31,10 +31,6 @@ jobs:
       - name: Setup Python
         uses: ./.github/actions/setup-backend/
         if: steps.check.outputs.python
-      - name: pylint
```
mistercrunch, line 34: NOTE: this other action here https://github.com/apache/superset/blob/master/.github/workflows/pre-commit.yml#L38 will cover for this one.

john-bodley:
@mistercrunch per [here](https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html) it seems like this pre-commit hook will only work locally and will be ineffective in a remote CI setting.

rusackas:
Seems good to me, but I think @john-bodley might be the right stakeholder approver here.

mistercrunch:
@john-bodley I think it should be fine as long as I install it in the `setup-backend` reusable action, it'll install the right/same version as what should be on every dev env too.

I just realized it's only pylinting the `superset/` folder, not the `tests/` or script. It's really resource-intensive though, brings my M2 macbook to its knees. I think I'll give `ruff` a shot as it's blazing fast and catches (and autofixes) a bunch of things that other tools don't. They are crushing it on their roadmap and eventually should fully replace pylint/isort/black/pydantic/ .... Coming up in another PR soo

mistercrunch:
I think turning this on as pre-commit is a clear win if we're going to enforce pylint in CI. It's a bit slow, but pre-commit is a good place to pay the cost of verification (CI is too late, and pylint IDE integrations are sluggish in my experience)

mistercrunch:
Can I get the @john-bodley seal of approval :)?

betodealmeida (approved):

john-bodley:
@mistercrunch regarding your comment, 

> We should also consider moving to ruff and streamlining all this, but ruff
doesn't support all the pylint rules [yet].

There's a few WIP PRs—by @betodealmeida (I believe) and myself—which have custom Pylint extensions so I'm not sure if/when we'll actually can switch off Pylint as I _believe_ Ruff [doesn't support custom lint rules](https://github.com/astral-sh/ruff/discussions/8409).

mistercrunch:
@john-bodley I don't think we should deprecate `pylint`, it seems like it will be a long time before ruff gets to covering most of pylint -> https://github.com/astral-sh/ruff/issues/689 . This means it'll probably take years before it's mostly covered.

One thing that we could consider though is disabling the specific (200+) pylint rules where ruff already covers currently. This could allow pylint to perform better as it's pretty slow currently. This could accelerate commit hooks execution (usually that's not bad as is currently) but significantly speed up `pre-commit run --all`, which runs in CI and should probably run when doing big rebase/merges.

PR #28152, opened 2024-04-19
fix(sql_parse): Provide more lenient logic when extracting latest[_sub]_partition

john-bodley:
### SUMMARY

This is a follow up to https://github.com/apache/superset/pull/28117, which sadly wasn't suffice when variables as opposed to constants were being used with the `format()` Jinja macro.

Sadly the `as_const()` method has only been implemented for a subset of node types and given that the plethora of logic which one could define is vast it's likely best to just acknowledge there are times when we wont be able to accurately extract the table referenced by the `latest[_sub]_partition` Jinja macro.

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -1554,16 +1554,19 @@ def extract_tables_from_jinja_sql(sql: str, database: Database) -> set[Table]:
             "latest_partition",
             "latest_sub_partition",
         ):
-            # Extract the table referenced in the macro.
-            tables.add(
-                Table(
-                    *[
-                        remove_quotes(part.strip())
-                        for part in node.args[0].as_const().split(".")[::-1]
-                        if len(node.args) == 1
-                    ]
+            # Try to extract the table referenced in the macro.
+            try:
+                tables.add(
+                    Table(
+                        *[
+                            remove_quotes(part.strip())
+                            for part in node.args[0].as_const().split(".")[::-1]
+                            if len(node.args) == 1
+                        ]
+                    )
                 )
-            )
+            except nodes.Impossible:
+                pass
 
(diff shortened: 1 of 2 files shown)


tests/unit_tests/sql_parse_tests.py
```diff
@@ -1938,36 +1938,40 @@ def test_sqlstatement() -> None:
...
-            "SELECT '{{{{ {engine}.{macro} }}}}'",
+            "latest_partition('foo.bar')",
+            {Table(table="bar", schema="foo")},
+        ),
+        (
+            "latest_partition(' foo.bar ')",  # Non-atypical user error which works
             {Table(table="bar", schema="foo")},
         ),
         (
-            "SELECT * FROM foo.baz WHERE quux = '{{{{ {engine}.{macro} }}}}'",
```
john-bodley, line 1957: This example no longer exists. It's a trivial extension from the previous logic, but was making the nested parametrization challenging.

superset/sql_parse.py
```diff
@@ -1554,16 +1554,19 @@ def extract_tables_from_jinja_sql(sql: str, database: Database) -> set[Table]:
...
-            tables.add(
-                Table(
-                    *[
-                        remove_quotes(part.strip())
-                        for part in node.args[0].as_const().split(".")[::-1]
-                        if len(node.args) == 1
-                    ]
+            # Try to extract the table referenced in the macro.
+            try:
+                tables.add(
```
john-bodley, line 1559: Same logic as before, just nested within a try/except block.

michael-s-molina (approved):

PR #28156, opened 2024-04-19
fix(sqllab): invalid css scope for ace editor autocomplete

justinpark:
### SUMMARY
The issue has been fixed in this commit as the autocomplete container is now properly appended to the body instead of the ace container. This commit addresses and fixes the issue by applying the autocomplete style to the global.

Before:

Screenshot 2024-04-19 at 4 08 48 PM

After:

Screenshot 2024-04-19 at 4 08 30 PM

### TESTING INSTRUCTIONS
Go to SQL Lab and type anything for autocomplete

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/AceEditorWrapper/index.tsx
+++ b/superset-frontend/src/SqlLab/components/AceEditorWrapper/index.tsx
@@ -22,6 +22,7 @@ import { useDispatch } from 'react-redux';
 import { css, styled, usePrevious, useTheme } from '@superset-ui/core';
 import { Global } from '@emotion/react';
 
+import { SQL_EDITOR_LEFTBAR_WIDTH } from 'src/SqlLab/constants';
 import { queryEditorSetSelectedText } from 'src/SqlLab/actions/sqlLab';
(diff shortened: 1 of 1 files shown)


superset-frontend/src/SqlLab/components/AceEditorWrapper/index.tsx
```diff
@@ -22,6 +22,7 @@ import { useDispatch } from 'react-redux';
 import { css, styled, usePrevious, useTheme } from '@superset-ui/core';
 import { Global } from '@emotion/react';
 
+import { SQL_EDITOR_LEFTBAR_WIDTH } from 'src/SqlLab/constants';
```
rusackas, line 25: Needn't tackle it in this PR, but I would think layout constants like these should be global (or in the theme) since they likely could/should apply to other views in the app.

michael-s-molina: `src/SqlLab/constants.ts` looks like a good place for this constant.

michael-s-molina (approved):

PR #28121, opened 2024-04-18
docs: Refactoring structure of docs and sidebar

artofcomputing:
### SUMMARY
This DRAFT PR attempts to refactor the current documentation structure by doing the following things:
- Create a custom Docusaurus sidebar for handling docs positions and overall organization.
- Split Installation and Configuration apart, giving a focused approach for each topic.
- Create a new topic called "Using Superset" which will focus on day-to-day usage (creating dashboards, charts, using tags, creating reports and alerts).
- Renaming of previous topics to simpler names.
- Setting up the option of "hideable" sidebars.

Since there are changes within the document structure, there's a few broken links, will fix them before being ready to review.

PR #28161, opened 2024-04-20
docs: Refactor Documentation Structure

artofcomputing:
### SUMMARY
This PR aims to perform the same as #28121 (without redoing a bunch of commits) which does the following things:

- Create a custom Docusaurus sidebar for better handling documents positions via sidebar.js (no more manual reshuffling of files)
- Split Installation and Configuration apart, giving a focused approach for each topic.
- Create a new topic called "Using Superset" which will focus on day-to-day usage (creating dashboards, charts, using tags, creating reports and alerts).
- Renaming of previous topics to simpler names while renaming files for better organization.
- Setting up the option of "hideable" sidebars.

Multiple links have been broken due to renaming of files and new directories, bulk of commits were done in order to fix links.


rusackas (approved):
LGTM, but wouldn't mind a second pair of eyes here :)

sfirke (approved):
Outstanding!

sfirke:
Thank you so much!  This is a nice improvement.  Re: the changes in `docs/sidebars.js`, I want to make sure I get it -- before this was autogenerated, now this is a manual list, which gives us better control over that ordering?  And then for the subpages, like databases, those are autogenerated?

sfirke:
Will we still have manual reshuffling of individual database subpages, in that section?

rusackas:
Merging it before crazy conflicts happen, but I guess the last question still stands :)

artofcomputing:
> Re: the changes in docs/sidebars.js, I want to make sure I get it -- before this was autogenerated, now this is a manual list, which gives us better control over that ordering? And then for the subpages, like databases, those are autogenerated?

It's a good question, I'm thinking of adding the structure and how to configure it under the contributing section.

The way it works now is a mix of both, for sections we're using a manual configuration, that way we don't need to create _category_.json within each directory to define the order, that save us time. For documents, we're using the autogenerated feature, so we don't need to manually add all of the documents on the configuration file (i.e manually adding all DB pages would take time).

If we need to change page order, we can do it either in the configuration file by specifying its location, or using the sidebar_position attribute at the start of the file

PR #28164, opened 2024-04-20
chore: refactor file upload commands

dpgaspar:
### SUMMARY
Small refactor on `CSVImportCommand` and `ExcelImportCommand` to DRY and make it more easily extendable and maintainable.

Using dependency injection on a generic `UploadCommand` of `BaseReaders` that will be able to read a certain type of file.
Pandas lib is able to read from multiple file types, CSV, HTML, Excel, XML, Latex, Feather, SAS XPORT etc. So with this refactor it will be much simpler and clean to add support for new file types.

Also this PR fixes the following bugs on reader options:
- `overrite_duplicates`: was an existing option that wasn't actually implemented, removed since I found no option for it on pandas `read_csv`
- `decimal`: was not actually implemented

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/databases/schemas.py
```diff
@@ -1116,12 +1116,6 @@ class CSVUploadPostSchema(BaseUploadPostSchema):
             "description": "DD/MM format dates, international and European format"
         }
     )
-    overwrite_duplicates = fields.Boolean(
-        metadata={
-            "description": "If duplicate columns are not overridden,"
-            "they will be presented as 'X.1, X.2 ...X.x'."
-        }
-    )
```
eschutho, line 1124: @dpgaspar will this break an existing api if someone is passing this field? Should we deprecate it instead?

eschutho (approved):
left one comment but it looks great!

superset/databases/schemas.py
dpgaspar: this was only used on the new API introduced here: https://github.com/apache/superset/pull/27840
so this API was not yet released

PR #28176, opened 2024-04-22
feat(reports): Set a minimum interval for each report's execution

Vitor-Avila:
### SUMMARY
It's currently possible to create alerts and reports configured to be executed every minute. While it might be a suitable implementation for some use-cases, some users might create assets with a more recurrent frequency needed, leading to additional load to the DB which can cause slowness, timeouts, etc. 

This PR introduces two configs that can be specified in `superset_config.py`:
``` python
# Set a minimum interval threshold between executions (for each Alert/Report)
# Value should be the amount of minutes (integer). Min: 1, Max: 60
ALERT_MINIMUM_INTERVAL_MINUTES = None
REPORT_MINIMUM_INTERVAL_MINUTES = None
```

This minimum interval is validated for both creation of new reports and also the modification of existing ones.

Note that this PR doesn't include a migration, so changing this config won't automatically change existing reports.

No UI changes.

### TESTING INSTRUCTIONS
Both integration and unit tests added. For manual testing:
1. Specify a minimum interval in config:
``` python
ALERT_MINIMUM_INTERVAL_MINUTES = 2
REPORT_MINIMUM_INTERVAL_MINUTES = 2
```
2. Try creating a report/alert that exceeds this limit.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/commands/report/base.py
```diff
@@ -76,3 +78,43 @@ def validate_chart_dashboard(
...
+
+        :param cron_schedule: The cron schedule configured.
+        :param report_type: The report type (Alert/Report).
+        """
+        config_key = (
+            "ALERT_MINIMUM_INTERVAL_MINUTES"
+            if report_type == ReportScheduleType.ALERT
+            else "REPORT_MINIMUM_INTERVAL_MINUTES"
+        )
+        minimum_interval = current_app.config.get(config_key)
```
eschutho, line 99: you may want to also perform some validation to make sure that this value is a number, or it will fail below at line 119.

tests/unit_tests/commands/report/base_test.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+
+
+@app_custom_config(alert_minimum_interval=10)
+def test_validate_report_frequency_alert_only() -> None:
+    """
+    Test the ``validate_report_frequency`` method when there's
+    only a configuration for alerts and user is creating report.
+    """
+    for schedule in TEST_SCHEDULES:
```
eschutho, line 137: I think @pytest.mark.parametrize would work well for you on these for loop tests.

superset/commands/report/base.py
```diff
@@ -76,3 +78,43 @@ def validate_chart_dashboard(
...
+        if not minimum_interval or minimum_interval < 2:
```
betodealmeida, line 100: Why 2?

betodealmeida (commented):
I'm a bit worried about manipulating the crontab as a string to figure out the shortest interval — it can miss edge cases, and it forces constraints to the user due to the implementation details (in this case, the interval has to be in minutes, and has a maximum value of 60, which are artificial requirements).

An alternative solution would be to allow any minimum interval between reports:

```python
ALERT_MINIMUM_INTERVAL_MINUTES = timedelta(seconds=90)
```

Or, as it seems to be more common for intervals in `config.py`:

```python
ALERT_MINIMUM_INTERVAL_MINUTES = int(timedelta(seconds=90).total_seconds())
```

Then you can compute a few execution times programmatically and see if it violates the config (untested):

```python
def is_valid_schedule(cron_schedule: str, config_value: int) -> bool:
    it = croniter(cron_schedule)
    previous = next(it)
    for _ in range(1000):  # how many? I have no idea...
        current = next(it)
        interval, previous = current - previous, current
        if interval.total_seconds() < config_value:
            return False

    return True
```

Vitor-Avila:
> I'm a bit worried about manipulating the crontab as a string to figure out the shortest interval — it can miss edge cases, and it forces constraints to the user due to the implementation details (in this case, the interval has to be in minutes, and has a maximum value of 60, which are artificial requirements).
> 
> An alternative solution would be to allow any minimum interval between reports:
> 
> ```python
> ALERT_MINIMUM_INTERVAL_MINUTES = timedelta(seconds=90)
> ```
> 
> Or, as it seems to be more common for intervals in `config.py`:
> 
> ```python
> ALERT_MINIMUM_INTERVAL_MINUTES = int(timedelta(seconds=90).total_seconds())
> ```
> 
> Then you can compute a few execution times programmatically and see if it violates the config (untested):
> 
> ```python
> def is_valid_schedule(cron_schedule: str, config_value: int) -> bool:
>     it = croniter(cron_schedule)
>     previous = next(it)
>     for _ in range(1000):  # how many? I have no idea...
>         current = next(it)
>         interval, previous = current - previous, current
>         if interval.total_seconds() < config_value:
>             return False
> 
>     return True
> ```

hey @betodealmeida that was the route I was initially going for too, but "how many repetitions?" (in the `for _ in range(x)`) is what made me think of another alternative. There's a lot of flexibility in the configuration (like the user could make it every 3 minutes until minute 58, but then re-execute on 59, select multiple days with a different interval between each in a month, etc). 

I can see benefits and concerns in both routes:
* **PR approach:** Biggest advantage here is that its more limited scope can validate all executions, since it only cares to the minute piece. Biggest disadvantage is that it's more restrictive (limit can't be higher than 60 minutes), and that it's handling a cron schedule as a string.

* **Implementing cron validation:** No limit on the configuration side (user could create any interval needed like days, hours, etc). The solution might not iterate on all executions.

Since I believe most Orgs would like to limit this in the minutes range (like at least 5/10 minutes interval), I thought it made more sense to go with the first route. But I'm happy to migrate to the second approach -- we could start with a loop with `1440` as the repetition count (amount of minutes in a day) and go from there.

Vitor-Avila:
thanks for the feedback, @eschutho 🙏 I'll put this on hold until we align on best route to move forward, before working on these.

john-bodley:
@Vitor-Avila how do these setting reflected in the UI given one can specifying whatever cron schedule they desire?

superset/commands/report/base.py
Vitor-Avila: great catch! Just fixed that

superset/commands/report/base.py
```diff
@@ -76,3 +78,43 @@ def validate_chart_dashboard(
...
+            return
+
+        minutes_interval = cron_schedule.split(" ")[0]
+        # Check if cron is set to every minute (*)
+        # or if it has an every-minute block (1-5)
+        if "-" in minutes_interval or minutes_interval == "*":
+            raise ReportScheduleFrequencyNotAllowed(report_type, minimum_interval)
```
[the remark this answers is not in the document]
Vitor-Avila: Let me know if the new implementation is better -- happy to further change it as well.

tests/unit_tests/commands/report/base_test.py
Vitor-Avila: ohh, wasn't familiar with it! just added it

Vitor-Avila:
> @Vitor-Avila how do these setting reflected in the UI given one can specifying whatever cron schedule they desire?

@john-bodley I haven't implemented any UI validation, so the user would face the toast error message when trying to save:

image

superset/commands/report/base.py
```diff
@@ -76,3 +79,45 @@ def validate_chart_dashboard(
...
+            if report_type == ReportScheduleType.ALERT
+            else "REPORT_MINIMUM_INTERVAL"
+        )
+        minimum_interval = current_app.config.get(config_key, 0)
+
+        if not isinstance(minimum_interval, int):
+            logger.error(
+                "Invalid value for %s: %s", config_key, minimum_interval, exc_info=True
+            )
+            return
```
Vitor-Avila, line 106: I decided to avoid raising a `ValidationError` here to prevent blocking users from creating alerts/reports until the configuration is fixed by an admin. Open to feedback, tho

superset/commands/report/exceptions.py
```diff
@@ -93,6 +93,34 @@ def __init__(self) -> None:
...
+    def __init__(
+        self,
+        report_type: str = "Report",
+        minimum_interval: int = 120,
+    ) -> None:
+        interval_in_minutes = (
+            minimum_interval / 60
+            if not minimum_interval % 60
+            else minimum_interval // 60 + 1
+        )
```
betodealmeida, line 111: I think this does it:

```suggestion
        interval_in_minutes = math.ceil(minimum_interval / 60)
```

(you need to import `math` from the stdlib, though)

superset/config.py
```diff
@@ -1325,6 +1325,10 @@ def EMAIL_HEADER_MUTATOR(  # pylint: disable=invalid-name,unused-argument
 # Custom width for screenshots
 ALERT_REPORTS_MIN_CUSTOM_SCREENSHOT_WIDTH = 600
 ALERT_REPORTS_MAX_CUSTOM_SCREENSHOT_WIDTH = 2400
+# Set a minimum interval threshold between executions (for each Alert/Report)
+# Value should be an integer i.e. int(timedelta(minutes=5).total_seconds())
+ALERT_MINIMUM_INTERVAL = 0
+REPORT_MINIMUM_INTERVAL = 0
```
betodealmeida, line 1331: Maybe this would be clearer about the expected units of `ALERT_MINIMUM_INTERVAL` and `REPORT_MINIMUM_INTERVAL`.

```suggestion
ALERT_MINIMUM_INTERVAL = int(timedelta(seconds=0).total_seconds())
REPORT_MINIMUM_INTERVAL = int(timedelta(seconds=0).total_seconds())
```

betodealmeida (approved):
Awesome! I just left a couple suggestions.

superset/config.py
Vitor-Avila: ohh that's a great suggestion! I think I'll use `int(timedelta(minutes=0).total_seconds())` since seconds are not really exposed in the report configuration, but it's def better than `0`. 🙌

eschutho (approved):

PR #28189, opened 2024-04-24
fix(docs): ERD docs fail on master

mistercrunch:
Screenshot 2024-04-23 at 6 21 42 PM

In https://github.com/apache/superset/pull/28130 I YOLOed the docs and failed.

This time around I tested the CI workflow with this command : `gh workflow run superset-docs-deploy.yml -r fix_docs2`. Logs are here -> https://github.com/apache/superset/actions/runs/8809186198/job/24179578667


mistercrunch:
@rusackas !

.github/workflows/superset-docs-deploy.yml
```diff
@@ -41,19 +40,28 @@ jobs:
           node-version: '18'
       - name: Setup Python
         uses: ./.github/actions/setup-backend/
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '21'
+      - name: Install Graphviz
```
rusackas, line 47: Note that this is not an Apache-compatible license, but for this use case, I _think_ it's Ok.

rusackas (approved):

PR #28205, opened 2024-04-24
fix: all_database_access should enable access to all datasets/charts/dashboards

mistercrunch:
When granting `all_database_access`, one would expect to get access to all charts and datasets.

From my understanding `all_database_access` is practically identical to `all_datasource_access`, and perhaps the two perms should be collapsed into one. In the meantime, this addresses user confusion around this perm not working according to its name.

I'm guessing that the difference between the two could be that `all_database_access` would also allow you to CRUD database objects(?)

Side mission:

Introducing a nifty context manager for tests to create temp user:
```python 
with self.temporary_user(clone_user=gamma_user, extra_pvms=[('all_database_access', 'all_database_access')],login=True) as user:
    user.do_stuff()
```

The neat thing is it can clone a user, will delete it on exit, optionally can log user in/out. It's not mutating anything so it should work well for running tests in parallel and/or for tests that fail half way through.

I'm planning on a follow up refactoring PR to standardize on using `temporary_user` everywhere where the current create_user is used.

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -403,6 +403,14 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
 
         return self.can_access("all_query_access", "all_query_access")
 
+    def can_access_all_databases(self) -> bool:
+        """
+        Return True if the user can access all the databases, False otherwise.
+
+        :returns: Whether the user can access all the databases
+        """
+        return self.can_access("all_database_access", "all_database_access")
+
     def can_access_all_datasources(self) -> bool:
         """
         Return True if the user can access all the datasources, False otherwise.
@@ -410,16 +418,9 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
         :returns: Whether the user can access all the datasources
         """
 
-        return self.can_access("all_datasource_access", "all_datasource_access")
-
-    def can_access_all_databases(self) -> bool:
-        """
-        Return True if the user can access all the databases, False otherwise.
-
-        :returns: Whether the user can access all the databases
-        """
-
-        return self.can_access("all_database_access", "all_database_access")
+        return self.can_access_all_databases() or self.can_access(
+            "all_datasource_access", "all_datasource_access"
+        )
 
     def can_access_database(self, database: "Database") -> bool:
         """
@@ -2433,7 +2434,6 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
 
         if self.is_admin():
             return
-
         orig_resource = db.session.query(resource.__class__).get(resource.id)
         owners = orig_resource.owners if hasattr(orig_resource, "owners") else []
 
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -403,14 +403,6 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
 
         return self.can_access("all_query_access", "all_query_access")
 
-    def can_access_all_databases(self) -> bool:
-        """
-        Return True if the user can access all the databases, False otherwise.
-
-        :returns: Whether the user can access all the databases
-        """
-        return self.can_access("all_database_access", "all_database_access")
-
     def can_access_all_datasources(self) -> bool:
         """
         Return True if the user can access all the datasources, False otherwise.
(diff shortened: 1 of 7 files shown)


superset/security/manager.py
```diff
@@ -409,15 +409,16 @@ def can_access_all_datasources(self) -> bool:
         :returns: Whether the user can access all the datasources
         """
 
-        return self.can_access("all_datasource_access", "all_datasource_access")
+        return self.can_access_all_databases() or self.can_access(
+            "all_datasource_access", "all_datasource_access"
+        )
```
betodealmeida, line 414: This makes sense and looks fine, I'm just worried about unintended consequences. Would be nice to get a signoff from @dpgaspar and/or @villebro and/or @john-bodley here, just in case.

Also, it would be really nice if we could decouple the data permissions from the API permissions. Maybe we should create a workgroup to brainstorm how this should look like.

mistercrunch: I agree, grepping for `all_database_access` there's very little justifying its existence as AFAIU, two choices here:
- make it work as "expected" based on name which would be what's in this PR, essentially making `all_database_access` == `all_datasource_access`
- deprecate `all_database_access` and [maybe] "migrate"/reassign to `all_datasource_access` where it's defined

mistercrunch: Digging a bit deeper I found a reference to `SecurityManager.can_access_all_databases` in `superset/databases/filters.py`, which seems potentially at odds with the model-based perms as in `Database.can_write` and `Database.can_read`. If we were to deprecate `all_database_access` we could transfer it to `Database.can_read` in roles where it's specified to try to match the current behavior.

mistercrunch:
So, what do we do? Deprecate or fix?

superset/security/manager.py
betodealmeida: @eschutho mentioned the case where we want to give someone access to all the tables for them to query in SQL Lab, but we don't want them modifying the database configuration. In that case, we'd want them to have `all_datasource_access` but not `all_database_access`, no?

eschutho:
Also a few more places like [here](https://github.com/apache/superset/blob/master/superset/security/manager.py#L433) where it could break. Plus as @mistercrunch mentioned the filters present more complexity where we can remove specific access through [configs](https://github.com/apache/superset/blob/6657353bcafbfd4dcbd6596bfb97f5ace179d7e4/superset/config.py#L1610) even with all database access.  

We can fix any outliers, so maybe the bigger question, is how do we want it to work? Most people probably logically assume that permissions work like Database -> Dataset -> Dashboard -> Chart, but they don't. I'm usually for small incremental changes, but the permissions models are so complex, that I expect even a small change could break quite a few things. Maybe a 5.0 project?

superset/security/manager.py
mistercrunch: I think the CRUD on Database is tied to the model / modelview, as opposed to this special term.

I have a spreadsheet of all the perms, and the relevant ones for database CRUD should be these ->

<google-sheets-html-origin><style type="text/css"></style>
Database | can_export | MODEL_VIEW
-- | -- | --
Database | can_read | MODEL_VIEW
Database | can_write | MODEL_VIEW
Databases | menu_access | MENU

mistercrunch:
I think this specific special permission is redundant with `all_datasource_access` on one side (data access policy) and redundant with `Database - can_read` on the CRUD side of managing databases. 

I vote for deprecating this perm, though I'd rather get consensus prior to doing the bit of work.

superset/security/manager.py
john-bodley: We (Airbnb) tend not to use these access controls and thus I don't really have much to add. My only question would be, why this hasn't been an issue in the past as it _potentially_ looks like a fairly egregious oversight from a consistency standpoint—if in fact the proposed behavior is correct.

mistercrunch:
After conversations with @eschutho and @yousoph , we decided that `all_database_access` should remain and provide both access to all data (through all databases connections) and allow listing all said database connections. Adding a note to UPDATING.md to notify people upgrading, but it should't come as a surprise that `all_database_access` give access to all databases.

dpgaspar:
> After conversations with @eschutho and @yousoph , we decided that `all_database_access` should remain and provide both access to all data (through all databases connections) and allow listing all said database connections. Adding a note to UPDATING.md to notify people upgrading, but it should't come as a surprise that `all_database_access` give access to all databases.

I agree with this change, to my understanding before 2.1, `all database access` or a specific `database access on X` would not list any chart, dashboards, users would have database, schema and table access on SQLLab only. Currently granting `database access on X` will give access to all charts and dashboards based on that database, but not `all database access` so it does seem like a bug.

dpgaspar (approved):

PR #28207, opened 2024-04-25
docs: depricated Marking on outdated docs - Local dev setup superset

surapuramakhil:
### SUMMARY
I have spent multiple hours in figuring out setting up my local superset (development) setup.

I have followed both docs https://github.com/apache/superset/blob/master/CONTRIBUTING.md
and https://github.com/apache/superset/blob/master/docs/docs/contributing/local-backend.mdx

Ended up with a bunch of errors. (tried fixing, but chain it continuing) lot of breaking stuff.

I have even tried PyPI installation methods (even that method breaks) while delivering frontend content. Marked deprecated

Current way is https://superset.apache.org/docs/installation/docker-compose#option-1---for-an-interactive-development-environment

Updated same in other places. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -424,7 +424,7 @@ Should you decide that reverting is desirable, it is the responsibility of the C
 - **Provide concise reproduction steps:** Ensure that the issue can be clearly understood and duplicated by the original author of the PR.
 - **Put the revert through code review:** The revert must be approved by another committer.
 
-## Setup Local Environment for Development
+## Setup Local Environment for Development (Depricated use docker compose https://superset.apache.org/docs/installation/docker-compose#option-1---for-an-interactive-development-environment)
 
 First, [fork the repository on GitHub](https://help.github.com/articles/about-forks/), then clone it. You can clone the main repository directly, but you won't be able to send pull requests.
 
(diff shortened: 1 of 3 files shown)


rusackas:
While docker-compose is probably easiest (and smoother/faster than ever), the other methods of installation are not deprecated, and people use them all the time. Maybe you can file issues about the problems you ran into, or connect with some folks on [slack](http://bit.ly/join-superset-slack) to troubleshoot?

rusackas:
If we can help whatever situation is happening on your machine, maybe we can add some helpful documentation steps rather than kill off these installation methods.

surapuramakhil:
When I left installation steps, these were issues

1. From local setup docs - 1 errors was superset_config.py should be set in evn variable - that instruction was mission

It would install apache-superset as pip dependency.
1 was specific to brew - where is expected mysqClient should be brew. (it was complaining since I have installed MySQL via official installer)

2. when I did https://superset.apache.org/docs/installation/pypi, when did 
`superset run -p 8088 --with-threads --reload --debugger`

Server starts - (UI is not identical). After you give creds which you created while init, you see blank screen only two boxes are show.

sfirke:
@surapuramakhil I made some updates to the PyPI instructions, do you have feedback on this PR: https://github.com/apache/superset/blob/58c25238fee96c89203184dd2badda7a4f06aa35/docs/docs/installation/pypi.mdx

sfirke:
Last week I tried to install from PyPI following Superset's docs and it did not work.  That is a bad state of affairs, I agree.  And this PR helped bring my attention to that.  But I think the appropriate fix is to improve those instructions, not to mark those installation methods as deprecated.  I'm going to close this PR in light of that.  @surapuramakhil your feedback on my linked PR was already helpful, thank you!  And I welcome further suggestions from you or anyone else as to how to improve those installation instructions.

PR #28225, opened 2024-04-25
fix(Dev-Server): Edit ChartPropsConfig reexport to be a type object

rtexelm:
### SUMMARY

This should solve the warning that `ChartPropsConfig` does not exist in `superset-ui-core/src/chart/models/ChartProps`
See [here](https://github.com/rtexelm/superset-RTM/pull/new/fix/dev-server-warnings/chartpropsconfig/sc-80076) and [here](https://github.com/microsoft/TypeScript/issues/28481#issuecomment-479965467)

Example of warning fixed:

```zsh
WARNING in ./packages/superset-ui-core/src/chart/index.ts 24:0-28
export 'ChartPropsConfig' (reexported as 'ChartPropsConfig') was not found in './models/ChartProps' (possible exports: default)
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/chart/index.ts
+++ b/superset-frontend/packages/superset-ui-core/src/chart/index.ts
@@ -20,7 +20,11 @@
 export { default as ChartClient } from './clients/ChartClient';
 export { default as ChartMetadata } from './models/ChartMetadata';
 export { default as ChartPlugin } from './models/ChartPlugin';
-export { default as ChartProps, ChartPropsConfig } from './models/ChartProps';
+export { default as ChartProps } from './models/ChartProps';
+/* reexporting this interface as a type solves the dev-server
+ * warning the named export does not exist. See https://github.com/webpack/webpack/issues/7378#issuecomment-492641148
+ */
+export type { ChartPropsConfig } from './models/ChartProps';
 
 export { default as createLoadableRenderer } from './components/createLoadableRenderer';
 export { default as reactify } from './components/reactify';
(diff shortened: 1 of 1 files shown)


superset-frontend/packages/superset-ui-core/src/chart/index.ts
```diff
@@ -20,7 +20,11 @@
 export { default as ChartClient } from './clients/ChartClient';
 export { default as ChartMetadata } from './models/ChartMetadata';
 export { default as ChartPlugin } from './models/ChartPlugin';
-export { default as ChartProps, ChartPropsConfig } from './models/ChartProps';
+export { default as ChartProps } from './models/ChartProps';
+/* reexporting this interface as a type solves the dev-server
+ * warning the named export does not exist. See https://github.com/webpack/webpack/issues/7378#issuecomment-492641148
+ */
+export type { ChartPropsConfig } from './models/ChartProps';
```
eschutho, line 27: can you do something like this? It seems a bit more logical reading it this way, and may not need a comment then to explain why?
```
import { default as ChartProps, ChartPropsConfig } from './models/ChartProps';


export { ChartProps };
export type { ChartPropsConfig }; 
```

rtexelm: certainly, the comment is there to justify the transfer from `interface` to `type` for future pedantic devs like me

geido (approved):

geido (approved):

PR #28241, opened 2024-04-26
fix(explore): temporal column mixin

justinpark:
### SUMMARY
In order to address the issue of the `isTemporal` property not being calculated from the mapping state in temporalColumnMixin, this commit migrates the property to the override object property.
This change ensures that the 3rd party controlOverrides can be properly overridden.

### TESTING INSTRUCTIONS
Big number with Trendline works as before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/mixins.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/mixins.tsx
@@ -23,7 +23,12 @@ import {
   t,
   validateNonEmpty,
 } from '@superset-ui/core';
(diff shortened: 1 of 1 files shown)


michael-s-molina (approved):

lokeshwar9:
I wonder, if this has anything related to the below issue.
if we deselect few temporal columns in edit dataset and sync columns at any later point of time.
The configuration i applied for temporal columns are all gone and superset is mapping the all the time columns in dataset as temporal.
does this merge retains my configuration?

PR #28236, opened 2024-04-26
chore(docs): add closing ::: to caution tag

sfirke:
Right now there is no closing for the `::: caution` section so several screens of [this page ](https://superset.apache.org/docs/configuration/configuring-superset)are formatted as caution.

PR #28243, opened 2024-04-26
chore(docs): Move ::: onto its own line to fix caution formatting

sfirke:
### SUMMARY
Tried to fix this in #28236, it didn't work, trying again.  I'm modeling the fix on the caution block on the page https://raw.githubusercontent.com/apache/superset/master/docs/docs/quickstart.mdx which renders correctly.

### DIFF
--- a/docs/docs/configuration/configuring-superset.mdx
+++ b/docs/docs/configuration/configuring-superset.mdx
@@ -101,7 +101,8 @@ You can generate a strong secure key with `openssl rand -base64 42`.
 
 :::caution Use a strong secret key
 This key will be used for securely signing session cookies and encrypting sensitive information stored in Superset's application metadata database.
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):

mistercrunch:
Do you run `yarn run start`? It works super well for me and updates the pages instantly as I save files.

sfirke:
I haven't - sounds like I should, thanks for the tip!  My current dev setup leaves a lot to be desired, I need to improve it...

mistercrunch:
Oh yeah it's super neat - starts a process that monitors and auto-refreshes, you don't even need to `cmd-r` to refresh the page, otherwise you're flying blind!

rusackas:
Just "yarn start" to be pedantic ;)

Issue #25830, opened 2023-11-01
[SIP-99D] Proposal for consolidating Command/DAO create, update, and upsert operations


john-bodley:
## [SIP-99D] Proposal for consolidating Command/DAO create, update, and upsert operations

This SIP is part of the [[SIP-99] Proposal for correctly handling business logic](https://github.com/apache/superset/issues/25048) series. Specifically, it proposes consolidating the Command/DAO create, update, and upsert* operations in a single upsert operation to adhere to the DRY principle which should significantly reduce the code footprint/complexity.

\* upsert is a _portmanteau_ via the combination of the words "update" and "insert". These operations are already supported per the [`UpsertKeyValueCommand`](https://github.com/apache/superset/blob/1e37f0b41782d35d68f657bfa87aeb9055d1e6e7/superset/key_value/commands/upsert.py) (which interestingly references only the `CreateKeyValueCommand` command and not the `UpdateKeyValueCommand` command) and the [`EmbeddedDashboardDAO.upsert()`](https://github.com/apache/superset/blob/1e37f0b41782d35d68f657bfa87aeb9055d1e6e7/superset/daos/dashboard.py#L373C1-L385) method.

### Primer

[[SIP-35] Proposal for Improving Superset’s Python Code Organization](https://github.com/apache/superset/issues/9077) introduced the concept of a Commands and Data Access Objects (DAOs) and suggested that CRUD operations would be represented via create, update, and delete Commands and DAOs.

The challenge however is the create and update operations are rather similar—they typically mostly differ on the absence or presence of a primary key respectively—which leads to significant code/test bloat and copypasta (thus violating the DRY principle) which is prone to lead to result in code inconsistency.

#### Examples

- The [`CreateRLSRuleCommand`](https://github.com/apache/superset/blob/a3686459a9ac4a4b916c111971de85800c0e7dee/superset/row_level_security/commands/create.py) and [`UpdateRLSRuleCommand`](https://github.com/apache/superset/blob/a3686459a9ac4a4b916c111971de85800c0e7dee/superset/row_level_security/commands/update.py) classes are virtually copypasta of one another (per the FileMerge screenshot).

<img width="1440" alt="Screenshot 2023-11-01 at 4 59 52 PM" src="https://github.com/apache/superset/assets/4567245/5ec649af-5827-43d4-b684-e30971fe6ae7">

[...]

PR #28246, opened 2024-04-27
chore: clean up DB create command

betodealmeida:
### SUMMARY

I noticed the DB update command has logic to update permissions when the database name changes, but that's already being done by a [SQLAlchemy hook](https://github.com/apache/superset/blob/db5edb3a424e8cc9976c0b8b714119b55336fd93/superset/security/manager.py#L1158).

This PR removes the duplicate logic, and cleans up the SSH tunnel logic as well.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/update.py
+++ b/superset/commands/database/update.py
@@ -74,14 +74,9 @@ class UpdateDatabaseCommand(BaseCommand):
                 commit=False,
             )
             database.set_sqlalchemy_uri(database.sqlalchemy_uri)
-
-            try:
-                self._handle_ssh_tunnel(database)
-            except SSHTunnelError:
-                raise
-            except Exception as ex:
-                raise DatabaseUpdateFailedError() from ex
-
+            self._handle_ssh_tunnel(database)
+        except SSHTunnelError:
+            raise
         except (DAOUpdateFailedError, DAOCreateFailedError) as ex:
             raise DatabaseUpdateFailedError() from ex
 
--- a/superset/commands/database/update.py
+++ b/superset/commands/database/update.py
@@ -14,15 +14,19 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+from __future__ import annotations
+
 import logging
 from typing import Any, Optional
 
 from flask_appbuilder.models.sqla import Model
 from marshmallow import ValidationError
 
-from superset import is_feature_enabled
+from superset import is_feature_enabled, security_manager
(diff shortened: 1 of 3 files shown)


superset/commands/database/update.py
```diff
@@ -76,126 +68,47 @@ def run(self) -> Model:  # pylint: disable=too-many-statements, too-many-branche
...
-                        ) as ex:
-                            # So we can show the original message
-                            raise ex
-                        except Exception as ex:
-                            raise DatabaseUpdateFailedError() from ex
-
-            # adding a new database we always want to force refresh schema list
-            # TODO Improve this simplistic implementation for catching DB conn fails
-            try:
-                schemas = database.get_all_schema_names(ssh_tunnel=ssh_tunnel)
```
dpgaspar, line 135: this is one of the few places where we re fetch current schemas from the analytical dbs, and we take the chance to update any missing schemas permissions also. I don't think this is being done on the sqlalchemy hooks

john-bodley:
Thanks @betodealmeida. [[SIP-99D] Proposal for consolidating Command/DAO create, update, and upsert operations](https://github.com/apache/superset/issues/25830) proposes using upsert which should help remove redundant code.

superset/commands/database/update.py
betodealmeida: Talked offline, Daniel doesn't think this would work inside the hook, since it requires querying analytical databases. Leaving it here.

superset/commands/database/update.py
```diff
@@ -75,127 +73,106 @@ def run(self) -> Model:  # pylint: disable=too-many-statements, too-many-branche
...
-        old_view_menu_name: str, new_view_menu_name: str
+    def _handle_ssh_tunnel(self, database: Database) -> SSHTunnel | None:
+        """
+        Delete, create, or update an SSH tunnel.
+        """
+        if "ssh_tunnel" not in self._properties:
+            return None
+
+        if not is_feature_enabled("SSH_TUNNELING"):
+            db.session.rollback()
```
mistercrunch, line 105: are there any hanging transactions here?

mistercrunch (approved):
Minor comment, otherwise LGTM.

superset/commands/database/update.py
betodealmeida: There are, the database has been added before the SSH tunnel. I have a ticket to improve this and make it atomic.

PR #28247, opened 2024-04-28
docs: set up redirects

### LINKED ISSUE #28245: Superset Kubernetes documentation resource offline
### Bug description

Just by way of FYI:

Searching for your Kubernetes installation documentation brings me to this resource:

image

With URL:

https://superset.apache.org/docs/installation/running-on-kubernetes/

But I get this error:

image

Same behavior for the indexed documentation for configuration in general (also on the superset.apache.org domain):

https://superset.apache.org/docs/installation/configuring-superset/

### How to reproduce the bug

As above

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

mistercrunch:
### SUMMARY
Someone filled an issue about search index referring to dead links in our
documentation. Turns out many of the top links when googling `apache
superset documentation` point to 404s now.

In this PR I'm trying to set up client-side redirects to the related pages.

Note that the docs for this plugin state that server-side redirects are
better, and that the redirects don't work in dev, only prod, which
makes it hard to test

In the future, we'll need more rigor around catching dead links,
either during code review for the docs, or through some docusaurus
automation magic forcing us to come up with redirects whenever we kill
routes.

 [Docusaurus redirect plugin documentation](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) for reference

### Testing
I took the freedom to run `gh workflow run superset-docs-deploy.yml -r docs_redirect`, [logs here](https://github.com/apache/superset/actions/runs/8863837953/job/24338367534) since there's no way to test in dev and wanted this fix asap.

closes https://github.com/apache/superset/issues/28245

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -160,7 +160,7 @@ const config = {
             from: '/docs/contributing/creating-viz-plugins/',
           },
           {
-            to: '/docs/installation/kubernetes/',
+            to: '/docs/installation/kubernetes',
             from: '/docs/installation/',
           },
         ],
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -135,6 +135,34 @@ const config = {
             to: '/docs/faq',
             from: '/docs/frequently-asked-questions',
           },
+          {
+            to: '/docs/installation/kubernetes',
+            from: '/docs/installation/running-on-kubernetes/',
+          },
+          {
+            to: '/docs/contributing/howtos',
+            from: '/docs/contributing/testing-locally/',
+          },
+          {
+            to: '/docs/using-superset/creating-your-first-dashboard',
+            from: '/docs/creating-charts-dashboards/creating-your-first-dashboard/',
+          },
+          {
+            to: '/docs/using-superset/creating-your-first-dashboard',
+            from: '/docs/creating-charts-dashboards/exploring-data/',
+          },
+          {
+            to: '/docs/quickstart',
+            from: '/docs/installation/installing-superset-using-docker-compose/',
+          },
+          {
+            to: '/docs/contributing/howtos',
+            from: '/docs/contributing/creating-viz-plugins/',
+          },
+          {
+            to: '/docs/installation/kubernetes/',
+            from: '/docs/installation/',
+          },
         ],
       },
     ],
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -152,7 +152,7 @@ const config = {
             from: '/docs/creating-charts-dashboards/exploring-data/',
           },
           {
-            to: '/docs/quickstart',
+            to: '/docs/installation/docker-compose',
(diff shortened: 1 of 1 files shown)


mistercrunch:
Ok looks like this work, there may be more dead links indexed, but should cover the first few pages of google results when googling `apache superset documentation`

mistercrunch:
Unclear how much better are server-side redirects for this purpose, but search engines may do better with a clear permanent redirect (http 301)

docs/docusaurus.config.js
```diff
@@ -135,6 +135,34 @@ const config = {
...
+          {
+            to: '/docs/using-superset/creating-your-first-dashboard',
+            from: '/docs/creating-charts-dashboards/creating-your-first-dashboard/',
+          },
+          {
+            to: '/docs/using-superset/creating-your-first-dashboard',
+            from: '/docs/creating-charts-dashboards/exploring-data/',
+          },
+          {
+            to: '/docs/quickstart',
```
sfirke, line 155:
```suggestion
            to: '/docs/installation/docker-compose',
```

docs/docusaurus.config.js
```diff
@@ -135,6 +135,34 @@ const config = {
...
+          },
+          {
+            to: '/docs/contributing/howtos',
+            from: '/docs/contributing/creating-viz-plugins/',
+          },
+          {
+            to: '/docs/installation/kubernetes',
+            from: '/docs/installation/',
+          },
         ],
```
sfirke, line 166:
```suggestion
          {
            to: '/docs/installation/pypi/',
            from: '/docs/installation/installing-superset-from-pypi/',
          },
          {
            to: '/docs/configuration/configuring-superset/',
            from: '/docs/installation/configuring-superset/',
          },
          {
            to: '/docs/configuration/cache/',
            from: '/docs/installation/cache/',
          },
          {
            to: '/docs/configuration/async-queries-celery/',
            from: '/docs/installation/async-queries-celery/',
          },
          {
            to: '/docs/configuration/event-logging/',
            from: '/docs/installation/event-logging/',
          },
        ],
```

sfirke (approved):
Important fix to the docs, thank you.  I went through this recent PR #28161 and added additional redirects where:
- The result was broken coming from a search engine, and
- It wasn't yet fixed in the file on your branch

mistercrunch:
In terms of preventing dead link while allowing us to do more of these much needed refactors, I see two options:
- rigor and caution in docs PR reviews, make sure all orphaned routes are rewired some place
- set up some automation that prevents dead link. While working with the redirect plugin, I noticed it'll check and complain when there's a redirect to a link that's unknown to docusaurus. That means we could we could set up dummy redirect to every single existing route that would fail the build when becoming orphaned

sfirke:
Failing the build upon a broken link seems reasonable to me and more likely to succeed than diligent reviews.

PR #28260, opened 2024-04-29
fix(Dashboard): PDF/Image download via endpoint

geido:
PR has been moved here https://github.com/apache/superset/pull/29272


superset/views/img_proxy.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @expose("/")
+    @event_logger.log_this
+    def img_proxy(self) -> FlaskResponse:
+        """
+        Proxy to an external URL, to overcome CORS restrictions.
+        Returns a HTTP response containing the resource fetched from the external URL.
+        """
+        if not get_user_id():
+            abort(403)
```
dpgaspar, line 45: this seems a bit out of our current pattern. This would mean that no specific permission is required only a user must exist. Any reason to not use `has_access` decorator here?

superset/views/img_proxy.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if not request.referrer or (
+            request.referrer and not same_origin(request.referrer, request.url_root)
+        ):
+            abort(403)
+
+        response = self.fetch_resource(url)
+
+        if not response:
+            abort(500)
```
dpgaspar, line 65: What do you think about aborting with 404 instead?
Alternatively we could just let `fetch_resource` raise with `raise_for_status` and let flask handle the exception. We can also wrap RequestException with a Superset specific exception with status 500 or 404

superset/views/img_proxy.py
```diff
@@ -0,0 +1,105 @@
...
+        headers: dict[str, Any] = dict(response.headers.items())
+
+        return Response(response.content, response.status_code, headers)
+
+    def validate_image(self, image_content: bytes) -> bool:
+        """
+        Load the image from bytes, Pillow will raise an IOError if the file is not an image
+        """
+        try:
+            with Image.open(BytesIO(image_content)) as img:
```
dpgaspar, line 90: this may be potentially dangerous, ex: https://github.com/farisv/PIL-RCE-Ghostscript-CVE-2018-16509

superset/views/img_proxy.py
geido: It seems that we have no control over how these requests are being made via `dom-to-pdf` and the `has_access` decorator fails to recognize it as a logged in request

superset/views/img_proxy.py
```diff
@@ -0,0 +1,105 @@
...
+        parsed_url = urlparse(url)
+        if parsed_url.scheme not in ["http", "https"]:
+            abort(400)
+
+        if not request.referrer or (
+            request.referrer and not same_origin(request.referrer, request.url_root)
+        ):
+            abort(403)
+
+        response = self.fetch_resource(url)
```
craig-rueda, line 62: This looks super dangerous. I could just write a `curl` that spoofs the `referrer`, etc. That would allow me to then make requests to any other server (internal and external)

geido: The request needs to be authenticated but I agree there is potential risk here. We could limit the requests to certain allowed domains, however, that would require that the users of a dashboard know upfront the domains that will host images that they might want to attach to a dashboard via markdown.

geido:
Going with a completely different approach here due to the security concerns involved with using a proxy. There will be an endpoint implemented to generate a screenshot via the backend. The frontend will request the screenshot and wait for its execution to enable the download for the client.

geido:
Moved here https://github.com/apache/superset/pull/29272

PR #28265, opened 2024-04-29
feat(maps): Adding ALL the countries to the Country Map plugin! 🌎

rusackas:
### SUMMARY

So... after much ado of getting the Jupyter Notebook up to snuff, this PR adds as many of the countries as I could support. It pulls things from Natural Earth data just... as they are. We might need to adjust some disputed territories, move some islands closer to the mainland, etc., but in broad strokes, this should make us truly global, and we can continue to make overrides to Natural Earth in the notebook as we always have.

This does even more now!
* Strips out countries that have only one region, and would therefore be useless as a choropleth
* Standardizes display of maps (now with more colors!)
* Organizes a bit and cleans out some useless cells
* Pretties up the country preview grid
* Simplifies geometry for the country preview grid
* Skips the first preview_all_countries, since it takes a while and you don't _ususally_ need it
* Adds a TODO list of countries that could use layout adjustments due to distant islands.
* Removes the legacy countries_custom folder of geojson files... let's not have that!
* adds a "speed run" mode that skips rendering charts, and makes it WAAAAAY faster to just update maps locally or as part of CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


john-bodley (approved):
Woot!

rusackas:
Just need a codeowner review from @villebro to get this done :)

villebro (approved):
This is amazingly valuable work, BIG thanks for all the work here! Validated locally, LGTM!

PR #28266, opened 2024-04-29
fix: use pessimistic json encoder in SQL Lab

### LINKED ISSUE #28001: BYTEA column value decoding breaks preview and select. "'utf-8' codec can't decode byte 0xe6 in position 0: invalid continuation byte"
### Bug description

Preview and select functionalities break in SuperSet UI on columns of BYTEA type.

```
Database error
Failed to execute query '15' - 'SELECT 
  addressid,
  addressline1,
  city,
  stateprovinceid,
  postalcode,
  spatiallocation, -- <- BYTEA type triggering the error
  rowguid,
  modifieddate
from address 
limit 1
;': 'utf-8' codec can't decode byte 0xe6 in position 0: invalid continuation byte
```

Same query completes without error in Python console (following is ran inside an already running superset worker container):

```
>>> connection = engine.connect()
>>> q = 'select * from adventureworks.person.address limit 1'
>>> rows = connection.execute(text(q)).fetchall()
>>> rows
[(1, '1970 Napa Ct.', None, 'Bothell', 79, '98011', b'\xe6\x10\x00\x00\x01\x0c\xae\x8b\xfc(\xbc\xe4G@g\xa8\x91\x89\x89\x8a^\xc0', UUID('9aadcb0d-36cf-483f-84d8-585c2d4ec6e9'), datetime.datetime(2007, 12, 4, 0, 0))]
```

Table def in Postgres:

```
addressid INTEGER
addressline1 VARCHAR
addressline2 VARCHAR
city VARCHAR
stateprovinceid INTEGER
postalcode VARCHAR
spatiallocation BYTEA <- including this column in select causes error
rowguid UUID
modifieddate TIMESTAMP
```

That same table def as seen through Trino (presto):

```
addressid INTEGER
addressline1 VARCHAR
addressline2 VARCHAR
city VARCHAR
stateprovinceid INTEGER
postalcode VARCHAR
spatiallocation VARBINARY. <- including this column in select causes same error.
rowguid NULL
modifieddate TIMESTAMP
```

Superset version 4.0.0 (deployed per official Helm chart)

Per Chrome's Developer Tools > Network, the error is retuned with following call:

POST `api/v1/sqllab/execute/`

500 INTERNAL SERVER ERROR

Payload:

```
{"client_id":"kc-rU1IK7","database_id":2,"json":true,"runAsync":false,"schema":"public","sql":"SELECT\n  *\nfrom table_with_bytea \n;","sql_editor_id":"1","tab":"Untitled Query 1","tmp_table_name":"","select_as_cta":false,"ctas_method":"TABLE","queryLimit":1000,"expand_data":true}
```

Response:
```
{
    "errors": [
        {
            "message": "Failed to execute query '22' - 'SELECT\n  *\nfrom table_with_bytea \n;': 'utf-8' codec can't decode byte 0xbe in position 2: invalid start byte"
        }
    ]
}
```

# Expected Behavior

Superset should not try to decode values of binary types for display purposes.
Returning stringified hex is probably most meaningful. 

For display purposes, showing first ~32 chars of HEX representation of the value is acceptable. Or, replace with something like `<binary data>`

### How to reproduce the bug

In Postgres:

```
create table table_with_bytea
(
    working  varchar(10),
    breaking bytea
);

insert into table_with_bytea
values ('asdf', '\xDEADBEEF'::bytea)
;

-- run this in SuperSet:
select * from table_with_bytea;
```

### Screenshots/recordings

Screenshot 2024-04-11 at 16 38 48

### Superset version

4.0.0

### Python version

I don't know

### Node version

I don't know

### Browser

Chrome

### Additional context

official 4.0.0 images from docker hub apache/superset

### Checklist

mistercrunch:
### SUMMARY
An issue reported in https://github.com/apache/superset/issues/28001 shows SQL Lab failing hard when a database driver returns a type that can't be decoded for serialization. Switching to this pessimistic helper function for json serialization will prevent a hard failure, but will obfuscate the data for the specific cells where things can't be serialized.

Not being in a situation where I can reproduce the issue, I'm operating with the stacktrace, meaning I can't really handle the specific cell, but can prevent SQL Lab from failing hard on a single cell issue.

Few related notes:
- maybe there's something around the driver being in utf8 mode of some kind. In theory we could say that driver / superset should agree on utf8 for things to work
- driver should probably giving us a `bytes` type from a `VARBINARY` and from the stacktrace it appears to be something else, presumably `str`, though that's unclear
- could be good to refactor out everything json-related out of superset/utils/core.py and into a new superset/utils/json.py, and make sure we always use thesame utility functions throughout the codebase as opposed to raw `import json` across the codebase. From my understanding there's some magic around simplejson injecting itself everywhere that happens too.

closes https://github.com/apache/superset/issues/28001

### DIFF
--- a/superset/sqllab/api.py
+++ b/superset/sqllab/api.py
@@ -340,13 +340,15 @@ class SqlLabRestApi(BaseSupersetApi):
         key = params.get("key")
         rows = params.get("rows")
         result = SqlExecutionResultsCommand(key=key, rows=rows).run()
+
+        payload = json.dumps(
+            result,
+            default=utils.pessimistic_json_iso_dttm_ser,
+            ignore_nan=True,
+        )
         # return the result without special encoding
         return json_success(
-            json.dumps(
-                result,
-                default=utils.json_iso_dttm_ser,
-                ignore_nan=True,
-            ),
+            payload,
             200,
         )
 
--- a/superset/sqllab/api.py
+++ b/superset/sqllab/api.py
@@ -341,16 +341,14 @@ class SqlLabRestApi(BaseSupersetApi):
         rows = params.get("rows")
         result = SqlExecutionResultsCommand(key=key, rows=rows).run()
 
+        # Using pessimistic json serialization since some database drivers can return
+        # unserializeable types at times
         payload = json.dumps(
             result,
             default=utils.pessimistic_json_iso_dttm_ser,
             ignore_nan=True,
         )
-        # return the result without special encoding
-        return json_success(
-            payload,
-            200,
-        )
+        return json_success(payload, 200)
 
     @expose("/execute/", methods=("POST",))
     @protect()
--- a/superset/utils/core.py
+++ b/superset/utils/core.py
@@ -507,8 +507,8 @@ def json_iso_dttm_ser(obj: Any, pessimistic: bool = False) -> Any:
         return base_json_conv(obj)
     except TypeError as ex:
         if pessimistic:
+            logger.error("Failed to serialize %s", obj)
             return f"Unserializable [{type(obj)}]"
-
         raise ex
 
 
--- a/tests/unit_tests/utils/test_core.py
+++ b/tests/unit_tests/utils/test_core.py
@@ -18,6 +18,8 @@ import os
 from dataclasses import dataclass
 from typing import Any, Optional
 from unittest.mock import MagicMock, patch
+import datetime
+import json
 
 import pandas as pd
 import pytest
@@ -35,6 +37,8 @@ from superset.utils.core import (
     parse_boolean_string,
     QueryObjectFilterClause,
     remove_extra_adhoc_filters,
+    pessimistic_json_iso_dttm_ser,
+    json_iso_dttm_ser,
 )
 
 ADHOC_FILTER: QueryObjectFilterClause = {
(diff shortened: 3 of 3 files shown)


superset/sqllab/api.py
```diff
@@ -340,13 +340,15 @@ def get_results(self, **kwargs: Any) -> FlaskResponse:
...
         result = SqlExecutionResultsCommand(key=key, rows=rows).run()
+
+        payload = json.dumps(
+            result,
+            default=utils.pessimistic_json_iso_dttm_ser,
+            ignore_nan=True,
+        )
         # return the result without special encoding
         return json_success(
-            json.dumps(
```
john-bodley, line 345: Why not keep this inline given `payload` is a single use variable.

john-bodley (commented):
@mistercrunch would you mind adding some unit tests to i) help provide context, and ii) prevent future regressions.

Additionally what will be be rendered in the case of non-encodable values?

michael-s-molina:
@mistercrunch Should we show some kind of warning to the users when we obfuscate data? I'm trying to avoid a scenario where data is missing because of quality issues and users are not aware of the problem.

mistercrunch:
+1 totally agree on the unit test + logging something to the user and probably logging something to the administrator. Marking this as DRAFT as it needs more work, but want to share potential solution with the person who reported the issue.

mistercrunch:
Digging a bit it appears the pessimistic serializer would fill the unserializable cell with `f"Unserializable [{type(obj)}]"` https://github.com/apache/superset/blob/master/superset/utils/core.py#L508-L510 , which seems reasonable.

Though maybe catching `TypeError` is not broad enough here too as I think given the diversity and maturity spectrum of drivers we deal with, we should assume that it's likely we might get objects we cannot serialize and should fail gracefully (let bot user know and leave a good trace in the logs)

So left TODO here:
- try and get more info from the issue reported as to what type isn't handed, 
  -  handle that specific type as gracefully as possible
  - write a unit test specific to that type
- assume we may get ANY object from a driver
  - write a unit test representing a fully "unhandlable" type
  - make sure we log something useful
  - fill in the cell something that makes some sense to the user

superset/sqllab/api.py
mistercrunch: on this I'm happy either way, but thought the change made things a bit more readable, maybe making it a tiny bit easier to read the stacktrace or debug. But I don't have strong opinions on single-use variable

betodealmeida (approved):

Habeeb556:
I have the same problem encoder here #29457 but with `TIMESTAMP` datatype.

`Version 4.0.1`

rusackas:
@Habeeb556 looks like 4.0.2 brings this cheery in if that helps.

PR #5178, opened 2018-06-12
[sql] Correct SQL parameter formatting 

john-bodley:
TL;DR This PR removes the need for double escaping the percent character both in SQL Lab and explorer and ensures both approaches interface with the SQLA engine in a consistent manner.

This PR came about due to an error when exporting a SQL Lab result to CSV. For example in Presto one can successfully run the following query:
```
SELECT '%'
```
which executes directly via the SQLAlchemy engine, however an error was thrown when trying to export to CSV which uses the Pandas interface where the `parameters` argument is `{}` which causes an exception in [PyHive](https://github.com/dropbox/PyHive/blob/master/pyhive/presto.py#L172) which uses the `pyformat` SQL [parameterstyle](https://www.python.org/dev/peps/pep-0249/#paramstyle). In essence, 
```
SELECT '%' %  {}
```
is invalid and thus one would need to escape the `%` character, i.e., 
```
SELECT '%%' %  {}
```

The  reason for the error is the Pandas SQL uses the engine connection (which doesn't expose the execute arguments) whereas SQL Lab uses a raw connection (which exposes the execute arguments). 

We should ensure that both SQL Lab and explore use the same interface, which means that either i) the compiled SQL statements include escaping of `%` to satisfy `pyformat` and enforce the parameter args to be either `()` or `{}`, or ii) the compiled SQL statements remove all escaping and enforce the parameter args to be `None`. Note in both cases one can copy-and-paste explorer queries directly in SQL Lab, though this PR implements the second option as i) one doesn't need to escape the `%` character for the appropriate dialects, and ii) the raw SQL is what one would run in the native DB CLI.

 The fix is simply to bypass the Pandas API for querying and mimic the logic that SQL Lab uses which ensures that for presto `parameters` is `None` and thus it will not re-format the string. 

Note this PR also fixes the Hive and MySQL engine specs inline with their API, i.e., both [Hive](https://github.com/dropbox/PyHive/blob/master/pyhive/hive.py#L337) and [MySQL](https://github.com/PyMySQL/mysqlclient-python/blob/master/MySQLdb/cursors.py#L204) explicitly test whether the `parameters` and `args` respectively are `None` (rather than doing a boolean check). 

This means for MySQL in SQL Lab one can run queries of the form, 
```
SELECT * FROM table WHERE column LIKE '%foo%'
```
as opposed to:
```
SELECT * FROM table WHERE column LIKE '%%foo%%'
```
[...]

PR #28270, opened 2024-04-29
fix: druid double percent

betodealmeida:
### SUMMARY

Some DB API 2.0 drivers use `pyformat` for `paramstyle`. This means that queries should be parameterized with `%s` for the placeholders, like this:

```python
cursor.execute("SELECT * FROM t WHERE role = '%s'", ("engineer",))
```

The driver then performs "old-school" string interpolation:

```python
sql = "SELECT * FROM t WHERE role = '%s'"
parameters = ("engineer",)
escaped_parameters = escape_parameters(parameters)  # important to prevent SQL injection
final_query = sql % escaped_parameters
```

Because of the percent interpolation, when SQL compiles a query for [a database that uses `pyformat` or `format`](https://github.com/sqlalchemy/sqlalchemy/blob/6888cf79db79d5e5660300ccf2a2a91f1eecf75f/lib/sqlalchemy/sql/compiler.py#L7361-L7364), any percent symbols in the query are escaped by being replaced with `%%`:

https://github.com/sqlalchemy/sqlalchemy/blob/6888cf79db79d5e5660300ccf2a2a91f1eecf75f/lib/sqlalchemy/sql/compiler.py#L2652-L2653

For some reason we undo that process (introduced in https://github.com/apache/superset/pull/5178):

https://github.com/apache/superset/blob/76d897eaa2f9e137102bc194c2e3109c29d0348f/superset/models/core.py#L668-L669

The code above doesn't make sense. If SQLAlchemy is replacing `%` with `%%` for databases where `dialect.identifier_preparer._double_percents` is true, why would we reverse it when we compile the query for the same databases?

One clue can be found in another codepath were we compile the query, but that replacement is missing. In `values_for_column`:

https://github.com/apache/superset/blob/44690fb299ab3b7adc24e84eeec73bccdde14420/superset/models/helpers.py#L1380-L1384

@Vitor-Avila noticed that here, when the column is a calculated column containing a percent symbol, like:

```sql
case when column like '%a%' then 1 else 0 end
```

Then the generated SQL being sent to the database is:

```sql
case when column like '%%a%%' then 1 else 0 end
``` 

Note that the query above is completely valid and syntactically equivalent to the original one, so everything works as expected when we run it... except that in Druid, the query performs extremely poorly, compared to the one with a single percent. This suggests that the fix is to add `sql = sql.replace("%%", "%")` to the `values_for_column` method as well.

But again... why are we undoing what SQLAlchemy is doing?

Looking deeper into the problem, I found a bug in pydruid:

https://github.com/druid-io/pydruid/blob/1d72d26c3e14bc9a7c6725dfa877c98a7afbe6f3/pydruid/db/api.py#L430-L435

Note that in the code above, when no parameters are passed to `execute` — which is the case when Superset calls the method — the string interpolation never happens, because the SQL is returned early! This means that any escaped percent symbols (`%%`) **will not** be unescaped to `%`.

I've fixed pydruid in https://github.com/druid-io/pydruid/pull/317, and made a new release. This PR bumps the version to the fixed one, and removes the `sql = sql.replace("%%", "%")` logic completely. This way, when double percents are passed to pydruid, they will be unescaped **by the driver**, as expected.

### TESTING INSTRUCTIONS

I tested with Postgres, since psycopg2 uses `pyformat`. Queries run as expected:

Screenshot 2024-04-29 at 16-42-51 Superset

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,7 +119,7 @@ databricks = [
 db2 = ["ibm-db-sa>0.3.8, <=0.4.0"]
 dremio = ["sqlalchemy-dremio>=1.1.5, <1.3"]
 drill = ["sqlalchemy-drill>=1.1.4, <2"]
-druid = ["pydruid>=0.6.5,<0.7"]
+druid = ["pydruid>=0.6.7,<0.7"]
 duckdb = ["duckdb-engine>=0.9.5, <0.10"]
 dynamodb = ["pydynamodb>=0.4.2"]
 solr = ["sqlalchemy-solr >= 0.2.0"]
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -206,7 +206,7 @@ pure-sasl==0.6.2
     #   thrift-sasl
 pydata-google-auth==1.7.0
     # via pandas-gbq
-pydruid==0.6.6
+pydruid==0.6.7
     # via apache-superset
 pyee==11.0.1
     # via playwright
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -664,10 +664,6 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
         with self.get_sqla_engine(catalog=catalog, schema=schema) as engine:
             sql = str(qry.compile(engine, compile_kwargs={"literal_binds": True}))
 
-            # pylint: disable=protected-access
-            if engine.dialect.identifier_preparer._double_percents:  # noqa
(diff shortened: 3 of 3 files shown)


betodealmeida:
Ugh, reading https://github.com/apache/superset/pull/5178 again, it seems that this behavior is desirable for MySQL/Preseto/Hive. Will revisit this.

PR #28275, opened 2024-04-30
docs(upgrading): clarify upgrade process

mboegl:
Differentiate between upgrading superset docker/native itself and the metadata database to indicate that upgrading the database is necessary in either case.

### SUMMARY
Indicate that `superset db upgrade` and `superset init` are necessary after upgrading superset, independent of the deployment method.

Personally it was obvious that `superset db upgrade` was needed, but at that point I already felt the documentation could be clearer. I didn't execute `superset init` though, as it was not obvious from the `Docker Compose` section of the upgrade guide. I thought some weird HTTP 403 errors in the SQLLAB were a regression in 4.0.0 until I stumbled upon this line by accident, which, after execution, fixed everything.

### DIFF
--- a/docs/docs/installation/upgrading-superset.mdx
+++ b/docs/docs/installation/upgrading-superset.mdx
@@ -33,6 +33,13 @@ To upgrade superset in a native installation, run the following commands:
 
 ```bash
 pip install apache-superset --upgrade
+```
+
+## Upgrading Metadata Database
+
+Migrate the metadata database by running:
+
+```bash
 superset db upgrade
 superset init
 ```


rusackas (approved):
Running CI 🤞 

Changes look good to me, but I'll let others chime in before merging.

sfirke (approved):

sfirke:
Seems like a straightforward improvement to me.  Thank you @SaTae66 for the contribution!

sfirke:
Hm @SaTae66 are you running Docker with a single container or Docker Compose?

I upgraded yesterday using Docker Compose.  When I run `docker compose up`, it runs the script at https://github.com/apache/superset/blob/master/docker/docker-init.sh which includes `superset_db_upgrade`.  And when I look at `docker logs superset_init | head -n 100` I see that the DB migrations were applied.

So I don't think `superset db upgrade` is needed necessarily for Docker Compose, but I'd like to learn more.  For instance if someone is running with Docker Compose should we make it clearer they need to have this `docker-init.sh` file in the appropriate path?

Also did you have to enter your container first with `docker exec -it ...` in order to run `superset db upgrade` ?

PR #28276, opened 2024-04-30
chore: use depth=1 for cloning

rantoniuk:
It's not needed to clone full history of the repository just to get and spin up docker compose.

### DIFF
--- a/docs/docs/installation/docker-compose.mdx
+++ b/docs/docs/installation/docker-compose.mdx
@@ -50,7 +50,7 @@ Note that this documentation assumes that you have [Docker](https://www.docker.c
 following command:
 
 ```bash
-git clone https://github.com/apache/superset.git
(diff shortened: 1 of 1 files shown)


mistercrunch:
I'm not against the idea, but there's something nice in using vanilla commands. Say a more junior git user may be confused when trying to checkout a tag after making a shallow clone. More senior users know when to shallow-clone - say if they intent to make small contribution and such.

grepping for git clone:
```bash
$ git grep "git clone" docs/
docs/docs/contributing/development.mdx:git clone git@github.com:your-username/superset.git
docs/docs/installation/docker-compose.mdx:git clone https://github.com/apache/superset.git
docs/docs/quickstart.mdx:$ git clone https://github.com/apache/superset
```

`quickstart.mdx` is probably the place where shallow-cloning would make most sense.

craig-rueda (approved):

PR #28277, opened 2024-04-30
fix(ci): adding codecov token

rusackas:
### SUMMARY

Codecov token has been added to GitHub secrets. I _think_ this is all that's needed to leverage it, but I'm not highly confident.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/superset-frontend.yml
+++ b/.github/workflows/superset-frontend.yml
@@ -82,3 +82,5 @@ jobs:
         if: steps.check.outputs.frontend
         working-directory: ./superset-frontend
         run: ../.github/workflows/codecov.sh -c -F javascript
+        env:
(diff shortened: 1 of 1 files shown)


michael-s-molina (approved):

mistercrunch (approved):

rusackas:
@mistercrunch to answer your question about intent here, this was to resolve the ratelimit problem seen in a whole bunch o' PRs, like so:

image

rusackas:
Also note that I'm adding this new token to the PMC's 1Password vault in case anyone needs it.

michael-s-molina:
@rusackas It looks like we need this change for all workflows that connect to Codecov such as [database tests](https://github.com/apache/superset/actions/runs/8909934029/job/24468165629). It would be better if we can add the token globally.

PR #28280, opened 2024-04-30
fix(ci): run frontend actions based on master, not the fork.

rusackas:
### SUMMARY

This should make it so forks can leverage our secrets (e.g. codecov key) and not need to rebase.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/superset-frontend.yml
+++ b/.github/workflows/superset-frontend.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - "master"
       - "[0-9].[0-9]"
-  pull_request:
+  pull_request_target:
     types: [synchronize, opened, reopened, ready_for_review]
 
 # cancel previous workflow jobs for PRs
--- a/.github/workflows/superset-python-integrationtest.yml
+++ b/.github/workflows/superset-python-integrationtest.yml
@@ -6,7 +6,7 @@ on:
     branches:
       - "master"
       - "[0-9].[0-9]"
-  pull_request:
+  pull_request_target:
     types: [synchronize, opened, reopened, ready_for_review]
 
 concurrency:
--- a/.github/workflows/superset-python-presto-hive.yml
+++ b/.github/workflows/superset-python-presto-hive.yml
@@ -6,7 +6,7 @@ on:
     branches:
       - "master"
       - "[0-9].[0-9]"
-  pull_request:
+  pull_request_target:
     types: [synchronize, opened, reopened, ready_for_review]
 
 # cancel previous workflow jobs for PRs
--- a/.github/workflows/superset-python-unittest.yml
+++ b/.github/workflows/superset-python-unittest.yml
@@ -6,7 +6,7 @@ on:
     branches:
       - "master"
       - "[0-9].[0-9]"
-  pull_request:
+  pull_request_target:
     types: [synchronize, opened, reopened, ready_for_review]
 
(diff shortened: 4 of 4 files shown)


betodealmeida (approved):

mistercrunch (changes requested):
Pointing out that codecov is referenced in 7 places in `.github/workflows`
```bash
$ git grep codecov
.codecov.yml:codecov:
.github/workflows/superset-frontend.yml:        uses: codecov/codecov-action@v4
.github/workflows/superset-python-integrationtest.yml:        uses: codecov/codecov-action@v4
.github/workflows/superset-python-integrationtest.yml:        uses: codecov/codecov-action@v4
.github/workflows/superset-python-integrationtest.yml:        uses: codecov/codecov-action@v4
.github/workflows/superset-python-presto-hive.yml:        uses: codecov/codecov-action@v4
.github/workflows/superset-python-presto-hive.yml:        uses: codecov/codecov-action@v4
.github/workflows/superset-python-unittest.yml:        uses: codecov/codecov-action@v4
```
Also looked like there's some magic with the GHA:
https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov

rusackas:
@mistercrunch fixed! 

(cc @michael-s-molina since we discussed this earlier)

.github/workflows/superset-frontend.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - "master"
       - "[0-9].[0-9]"
-  pull_request:
+  pull_request_target:
```
dpgaspar, line 8: as per docs:
```
Warning: For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch. To help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered. For more information, see "[Keeping your GitHub Actions and workflows secure: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests)" on the GitHub Security Lab website.
```

secrets can leak to forks, this trigger runs github workflow code from the base branch (master) but we are checking out code from the PR ref on these workflows, that can open other vectors for attacks, for example by changing code on `/scripts` that would get executed by the workflow.

dpgaspar (changes requested):

.github/workflows/superset-frontend.yml
mistercrunch: That's right, the person calling from the fork can't change the workflow itself (the one from the main branch runs), but if that workflow calls say a script (or evolves into calling one down the line), then they could extract the env vars and/or secrets specified on the main branch for that action. We should review our `pull_request_target ` and put them in a special place where we have extra caution.

rusackas: Yep... I'll audit more closely for risk here, fix what I can, and add comments on what doesn't need fixing. Hopefully we can add linting/enforcement after that so other people that don't understand these complexities don't make the same mistake.

PR #28282, opened 2024-05-01
feat: accelerate webpack builds with filesystem cache

mistercrunch:
# NOW 5 TIMES FASTER 🔥 🍌🔥 🍌🔥 🍌🔥 🍌 

the only concern is there won't be enough time to fix a coffee ☕  when firing up webpack

speedy-gonzales-running-t0uvc86wu0vlqdwv

```
real    0m50.233s
user    1m56.371s
sys     0m10.192s
```

```
real    0m9.892s
user    0m11.070s
sys     0m2.034s
```

Works in docker-compose too
unusual-whale-typing-x2q3860nvo2i3hgz

### DIFF
--- a/superset-frontend/.gitignore
+++ b/superset-frontend/.gitignore
@@ -2,3 +2,4 @@ coverage/*
 cypress/screenshots
 cypress/videos
 src/temp
+.temp_cache/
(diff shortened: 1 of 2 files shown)


lilykuang (approved):
🔥

rusackas (approved):

PR #28283, opened 2024-05-01
fix: silence docker-compose useless warnings

mistercrunch:
When firing up docker-compose, 3 warning message are shown that are not useful / necessary

```
WARN[0000] The "CYPRESS_CONFIG" variable is not set. Defaulting to a blank string.
WARN[0000] The "SCARF_ANALYTICS" variable is not set. Defaulting to a blank string.
WARN[0000] The "CYPRESS_CONFIG" variable is not set. Defaulting to a blank string.
WARN[0000] /Users/max/code/superset/docker-compose.yml: `version` is obsolete
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,6 @@ x-common-build: &common-build
   cache_from:
     - apache/superset-cache:3.10-slim-bookworm
 
-version: "4.0"
 services:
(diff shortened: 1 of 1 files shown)


rusackas (approved):
Awesomesauce!

PR #28317, opened 2024-05-02
feat(SIP-95): permissions for catalogs

betodealmeida:
### SUMMARY

Continuing the work on https://github.com/apache/superset/issues/22862. This PR:

- Adds catalog support to the Postgres DB engine spec (in Postgres, a catalog is called a "database").
- Adds a new endpoint `/catalogs/` under the DB API, to list catalogs.
- Adds an optional parameter `catalog` to the `/schemas/` DB API, to specify a non-default catalog.
- Updates the security manager and model to manage catalog-level permissions (automatically creating, renaming, and deleting them, similar to how we do for for schemas).

These are all backend changes that shouldn't affect the current behavior of Superset, other than introducing the new permissions. Note that as of today users can query different catalogs in SQL Lab, and even create (virtual) datasets in different catalogs, and our SQL parser is able to handle catalogs correctly; because of that, introducing these new permissions can be considered a security improvement.

In the next PR I'll start working on the UI to expose the catalog functionality, and enabling support for multi-catalog in a given database will be optional and default to false.

<del>One thing that I'm not sure about is when we should rename existing schema permissions from `[db].[schema]` to `[db].[default-catalog].[schema]`. In this PR I'm doing it when the database gets updated. I don't think doing it in a migration is a good solution, because we'd have to add a new migration every time a DB engine spec is updated to support catalogs. Another option would be to do it during `superset init` — @dpgaspar do you have any thoughts here?</de>

<b>Edit:</b> added a migration for the schema permissions, done by a helper function, and a comment in the base DB engine spec informing that when `supports_catalog` is set to `True` a migration is necessary.

Screen Shot 2024-05-01 at 1 34 33 PM

### TESTING INSTRUCTIONS

I've updated existing tests, and I'm adding new tests for all the changes. The PR is draft so it's easier to track which changes have been covered with tests.

### ADDITIONAL INFORMATION

- [X] Has associated issue: https://github.com/apache/superset/issues/22862
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -73,10 +73,12 @@ from superset.daos.database import DatabaseDAO, DatabaseUserOAuth2TokensDAO
 from superset.databases.decorators import check_table_access
 from superset.databases.filters import DatabaseFilter, DatabaseUploadEnabledFilter
 from superset.databases.schemas import (
+    CatalogsResponseSchema,
     ColumnarMetadataUploadFilePostSchema,
     ColumnarUploadPostSchema,
     CSVMetadataUploadFilePostSchema,
     CSVUploadPostSchema,
+    database_catalogs_query_schema,
     database_schemas_query_schema,
     database_tables_query_schema,
     DatabaseConnectionSchema,
@@ -146,6 +148,7 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
         "table_extra_metadata",
         "table_extra_metadata_deprecated",
         "select_star",
+        "catalogs",
         "schemas",
         "test_connection",
         "related_objects",
@@ -266,6 +269,7 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
     edit_model_schema = DatabasePutSchema()
 
     apispec_parameter_schemas = {
+        "database_catalogs_query_schema": database_catalogs_query_schema,
         "database_schemas_query_schema": database_schemas_query_schema,
         "database_tables_query_schema": database_tables_query_schema,
         "get_export_ids_schema": get_export_ids_schema,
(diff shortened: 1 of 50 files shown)


superset/db_engine_specs/bigquery.py
```diff
@@ -127,7 +127,7 @@ class BigQueryEngineSpec(BaseEngineSpec):  # pylint: disable=too-many-public-met
 
     allows_hidden_cc_in_orderby = True
 
-    supports_catalog = True
+    supports_catalog = False
```
betodealmeida, line 130: Changing to false here and in other DB engine specs, since the specs should only have it set to true if they implement their own `get_default_catalog`. I'll update the DB engine specs in the near future.

superset/databases/api.py
```diff
@@ -589,6 +594,70 @@ def delete(self, pk: int) -> Response:
...
+            400:
+              $ref: '#/components/responses/400'
+            401:
+              $ref: '#/components/responses/401'
+            404:
+              $ref: '#/components/responses/404'
+            500:
+              $ref: '#/components/responses/500'
+        """
+        database = self.datamodel.get(pk, self._base_filters)
```
dpgaspar, line 639: `database = DatabaseDAO.find_by_id(pk)`

dpgaspar (approved):
LGTM, I vote for db migrations instead of superset init

superset/db_engine_specs/base.py
```diff
@@ -1412,24 +1426,24 @@ def get_catalog_names(  # pylint: disable=unused-argument
...
     @classmethod
-    def get_schema_names(cls, inspector: Inspector) -> list[str]:
+    def get_schema_names(cls, inspector: Inspector) -> set[str]:
         """
         Get all schemas from database
 
         :param inspector: SqlAlchemy inspector
         :return: All schemas in the database
         """
-        return sorted(inspector.get_schema_names())
```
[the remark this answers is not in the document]
betodealmeida: Ah, thanks! I'll update the logic in SQL Lab to show it sorted there then.

PR #28322, opened 2024-05-02
fix(sql_parse): Add Apache Spark to SQLGlot dialect mapping

john-bodley:
### SUMMARY

Though we support [Apache Spark](https://github.com/apache/superset/blob/master/superset/db_engine_specs/spark.py) there's no associated entry in the engine to SQLGlot dialect mapping, meaning that certain Spark SQL statements are non parsable when [no dialect is specified](https://github.com/apache/superset/blob/e94360486e3db9488119f805f4c118ba5021754a/superset/sql_parse.py#L464-L467):

```python
>>> from sqlglot import parse_one
>>>
>>> parse_one("SELECT * FROM `foo`.`bar`", dialect=None)
  ...
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 19.
  SELECT * FROM `foo`.`bar`
```

compared to 

```python
>>> from sqlglot import parse_one
>>>
>>> parse_one("SELECT * FROM `foo`.`bar`", dialect="spark")
Select(
  expressions=[
    Star()],
  from=From(
    this=Table(
      this=Identifier(this=bar, quoted=True),
      db=Identifier(this=foo, quoted=True))))
``` 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -137,6 +137,7 @@ SQLGLOT_DIALECTS = {
     "shillelagh": Dialects.SQLITE,
     "snowflake": Dialects.SNOWFLAKE,
     # "solr": ???
+    "spark": Dialects.SPARK,
     "sqlite": Dialects.SQLITE,
     "starrocks": Dialects.STARROCKS,
     "superset": Dialects.SQLITE,


michael-s-molina (approved):

superset/sql_parse.py
```diff
@@ -137,6 +137,7 @@
     "shillelagh": Dialects.SQLITE,
     "snowflake": Dialects.SNOWFLAKE,
     # "solr": ???
+    "spark": Dialects.SPARK,
```
betodealmeida, line 140: Nice!

PR #28340, opened 2024-05-03
fix: database logos look stretched

mistercrunch:
in previous documentation PR I messed up the look of some logos


rusackas (approved):
Yay! No more typo :)

PR #28341, opened 2024-05-04
fix: Remedy logic for UpdateDatasetCommand uniqueness check

john-bodley:
### SUMMARY

At Airbnb we detected a number of duplicate datasets even though, per the `UpdateDatasetCommand.validate()` method, there is a check to ensure that the new dataset name is unique. The issue was that it wrongfully used the schema of the existing (source) table as opposed to the target schema.

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/dataset/create.py
+++ b/superset/commands/dataset/create.py
@@ -61,17 +61,16 @@ class CreateDatasetCommand(CreateMixin, BaseCommand):
     def validate(self) -> None:
         exceptions: list[ValidationError] = []
         database_id = self._properties["database"]
-        table_name = self._properties["table_name"]
         schema = self._properties.get("schema")
         catalog = self._properties.get("catalog")
         sql = self._properties.get("sql")
         owner_ids: Optional[list[int]] = self._properties.get("owners")
 
-        table = Table(table_name, schema, catalog)
+        table = Table(self._properties["table_name"], schema, catalog)
 
         # Validate uniqueness
         if not DatasetDAO.validate_uniqueness(database_id, table):
-            exceptions.append(DatasetExistsValidationError(table_name))
+            exceptions.append(DatasetExistsValidationError(table))
 
         # Validate/Populate database
         database = DatasetDAO.get_database_by_id(database_id)
@@ -86,7 +85,7 @@ class CreateDatasetCommand(CreateMixin, BaseCommand):
             and not sql
             and not DatasetDAO.validate_table_exists(database, table)
         ):
-            exceptions.append(TableNotFoundValidationError(table_name))
+            exceptions.append(TableNotFoundValidationError(table))
 
         if sql:
             try:
--- a/superset/commands/dataset/update.py
+++ b/superset/commands/dataset/update.py
@@ -86,15 +86,21 @@ class UpdateDatasetCommand(UpdateMixin, BaseCommand):
         except SupersetSecurityException as ex:
             raise DatasetForbiddenError() from ex
 
-        database_id = self._properties.get("database", None)
-        table_name = self._properties.get("table_name", None)
+        database_id = self._properties.get("database")
+
+        table = Table(
+            self._properties.get("table_name"),  # type: ignore
+            self._properties.get("schema"),
+            self._properties.get("catalog"),
+        )
+
         # Validate uniqueness
         if not DatasetDAO.validate_update_uniqueness(
             self._model.database_id,
-            Table(table_name, self._model.schema, self._model.catalog),
+            table,
             self._model_id,
         ):
-            exceptions.append(DatasetExistsValidationError(table_name))
+            exceptions.append(DatasetExistsValidationError(table))
         # Validate/Populate database not allowed to change
         if database_id and database_id != self._model:
             exceptions.append(DatabaseChangeValidationError())
--- a/tests/integration_tests/datasets/api_tests.py
+++ b/tests/integration_tests/datasets/api_tests.py
@@ -42,6 +42,7 @@ from superset.datasets.models import Dataset  # noqa: F401
 from superset.extensions import db, security_manager
 from superset.models.core import Database
 from superset.models.slice import Slice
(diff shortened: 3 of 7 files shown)


superset/commands/dataset/create.py
```diff
@@ -61,17 +61,16 @@ def run(self) -> Model:
...
         sql = self._properties.get("sql")
         owner_ids: Optional[list[int]] = self._properties.get("owners")
 
-        table = Table(table_name, schema, catalog)
+        table = Table(self._properties["table_name"], schema, catalog)
 
         # Validate uniqueness
         if not DatasetDAO.validate_uniqueness(database_id, table):
-            exceptions.append(DatasetExistsValidationError(table_name))
+            exceptions.append(DatasetExistsValidationError(table))
```
john-bodley, line 73: We should be using a `Table` object rather than the table name which is too terse, i.e., doesn't contain the schema or catalog.

superset/commands/dataset/update.py
```diff
@@ -86,15 +86,21 @@ def validate(self) -> None:
...
+        table = Table(
+            self._properties.get("table_name"),
+            self._properties.get("schema"),
+            self._model.catalog,
+        )
+
         # Validate uniqueness
         if not DatasetDAO.validate_update_uniqueness(
             self._model.database_id,
-            Table(table_name, self._model.schema, self._model.catalog),
```
john-bodley, line 94: Here's the bug. The schema should be the schema defined in the properties and not the schema associated with the model.

tests/unit_tests/commands/dataset/test_update.py
```diff
@@ -0,0 +1,43 @@
+from unittest.mock import MagicMock
```
john-bodley, line 1: I was _blown away_ that there were [zero unit or integration tests](https://github.com/search?q=repo%3Aapache%2Fsuperset+UpdateDatasetCommand+path%3A%2F%5Etests%5C%2F%2F&type=code) for the `UpdateDatasetCommand`.

Though this test isn't pretty, it's a start.

tests/unit_tests/dao/dataset_test.py
```diff
@@ -51,7 +51,7 @@ def test_validate_update_uniqueness(session: Session) -> None:
     db.session.add_all([database, dataset1, dataset2])
     db.session.flush()
 
-    # same table name, different schema
```
john-bodley, line 54: These comments were ordered wrongly and thus misleading. It's probably best to not have comments if the logic is clear from the code.

tests/integration_tests/datasets/api_tests.py
```diff
@@ -660,7 +661,11 @@ def test_create_dataset_validate_uniqueness(self):
         assert rv.status_code == 422
         data = json.loads(rv.data.decode("utf-8"))
         assert data == {
-            "message": {"table_name": ["Dataset energy_usage already exists"]}
+            "message": {
+                "table": [
+                    f"Dataset {Table(energy_usage_ds.table_name, schema)} already exists"
```
john-bodley, line 666: The schema differs between SQLite and PostgreSQL and thus this is parameterized.

superset/commands/dataset/update.py
```diff
@@ -86,15 +86,21 @@ def validate(self) -> None:
         except SupersetSecurityException as ex:
             raise DatasetForbiddenError() from ex
 
-        database_id = self._properties.get("database", None)
```
john-bodley, line 89: The default for `dict.get()` is `None` and thus there's no need to explicitly define it.

tests/integration_tests/datasets/api_tests.py
```diff
@@ -682,7 +687,11 @@ def test_create_dataset_with_sql_validate_uniqueness(self):
...
```
john-bodley, line 692: See previous comment.

superset/commands/dataset/update.py
```diff
@@ -86,15 +86,21 @@ def validate(self) -> None:
...
-        table_name = self._properties.get("table_name", None)
+        database_id = self._properties.get("database")
+
+        table = Table(
+            self._properties.get("table_name"),  # type: ignore
+            self._properties.get("schema"),
+            self._properties.get("catalog"),
```
john-bodley, line 94: This doesn't exist in the schema, though is consistent with other logic. See https://github.com/apache/superset/commit/6cf681df6808c9b612cff1e53ddb6925a9b28ebf#r141673924 for details.

superset/commands/dataset/create.py
```diff
@@ -61,17 +61,16 @@ def run(self) -> Model:
...
         exceptions: list[ValidationError] = []
         database_id = self._properties["database"]
-        table_name = self._properties["table_name"]
         schema = self._properties.get("schema")
         catalog = self._properties.get("catalog")
         sql = self._properties.get("sql")
         owner_ids: Optional[list[int]] = self._properties.get("owners")
 
-        table = Table(table_name, schema, catalog)
+        table = Table(self._properties["table_name"], schema, catalog)
```
michael-s-molina, line 69: I think the previous version is more readable as it groups variables definition in just one place.
```suggestion
        table = Table(table_name, schema, catalog)
```

john-bodley: @michael-s-molina personally I try to remove single use variables (where possible) as I find it easier to read/grok the code.

tests/unit_tests/dao/dataset_test.py
```diff
@@ -51,7 +51,7 @@ def test_validate_update_uniqueness(session: Session) -> None:
...
+    #
```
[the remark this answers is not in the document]
john-bodley: 🤦

michael-s-molina (approved):

dpgaspar (approved):

PR #28349, opened 2024-05-06
fix: Add back description column to saved queries #12431

### LINKED ISSUE #12431: [sql_lab]Add back description column to saved queries
**Is your feature request related to a problem? Please describe.**
The description column was removed from the saved queries. We would like to add it back. 
Styling TBD

**Describe the solution you'd like**
Show the description of the saved query if it exists on the saved queries page.

**Describe alternatives you've considered**
Not showing the description, but it seems helpful.

**Additional context**
Banners_and_Alerts_and__DEV__Superset

imancrsrk:
Fixes https://github.com/apache/superset/issues/12431

**Add back Description column**
Screenshot 2024-05-05 at 6 08 34 PM

**Add tooltip for search box**
Screenshot 2024-05-06 at 4 25 32 PM

### DIFF
--- a/superset-frontend/src/components/InfoTooltip/index.tsx
+++ b/superset-frontend/src/components/InfoTooltip/index.tsx
@@ -83,7 +83,11 @@ export default function InfoTooltip({
       overlayStyle={overlayStyle}
       color={bgColor}
     >
-      <Icons.InfoSolidSmall className="info-solid-small" viewBox={viewBox} />
+      <Icons.InfoSolidSmall
+        className="info-solid-small"
+        viewBox={viewBox}
+        iconSize="m"
+      />
     </StyledTooltip>
   );
 }
--- a/superset-frontend/src/components/InfoTooltip/index.tsx
+++ b/superset-frontend/src/components/InfoTooltip/index.tsx
@@ -83,11 +83,7 @@ export default function InfoTooltip({
       overlayStyle={overlayStyle}
       color={bgColor}
     >
-      <Icons.InfoSolidSmall
-        className="info-solid-small"
-        viewBox={viewBox}
-        iconSize="m"
-      />
+      <Icons.InfoSolidSmall className="info-solid-small" viewBox={viewBox} />
     </StyledTooltip>
   );
 }
--- a/superset-frontend/src/pages/SavedQueryList/index.tsx
+++ b/superset-frontend/src/pages/SavedQueryList/index.tsx
@@ -298,6 +298,10 @@ function SavedQueryList({
         Header: t('Name'),
       },
       {
+        accessor: 'description',
+        Header: t('Description'),
+      },
+      {
         accessor: 'database.database_name',
         Header: t('Database'),
         size: 'xl',
@@ -447,6 +451,12 @@ function SavedQueryList({
         id: 'label',
         key: 'search',
         input: 'search',
+        operator: FilterOperator.AllText,
+      },
+      {
+        Header: t('Description'),
+        id: 'description',
+        key: 'description',
         operator: FilterOperator.AllText,
       },
       {
--- a/superset-frontend/src/pages/SavedQueryList/index.tsx
+++ b/superset-frontend/src/pages/SavedQueryList/index.tsx
@@ -447,7 +447,7 @@ function SavedQueryList({
   const filters: Filters = useMemo(
     () => [
       {
-        Header: t('search'),
+        Header: t('Search'),
         id: 'label',
         key: 'search',
         input: 'search',
--- a/superset-frontend/src/pages/SavedQueryList/index.tsx
+++ b/superset-frontend/src/pages/SavedQueryList/index.tsx
@@ -447,16 +447,10 @@ function SavedQueryList({
   const filters: Filters = useMemo(
     () => [
       {
-        Header: t('Name'),
+        Header: t('search'),
         id: 'label',
         key: 'search',
         input: 'search',
-        operator: FilterOperator.AllText,
-      },
-      {
-        Header: t('Description'),
-        id: 'description',
-        key: 'description',
         operator: FilterOperator.AllText,
       },
       {
--- a/superset-frontend/src/pages/SavedQueryList/index.tsx
+++ b/superset-frontend/src/pages/SavedQueryList/index.tsx
@@ -452,6 +452,8 @@ function SavedQueryList({
         key: 'search',
         input: 'search',
         operator: FilterOperator.AllText,
+        toolTipDescription:
+          'Searches all text fields: Name, Description, Database & Schema',
       },
       {
         Header: t('Database'),
--- a/superset-frontend/src/components/ListView/Filters/Search.tsx
+++ b/superset-frontend/src/components/ListView/Filters/Search.tsx
@@ -29,7 +29,7 @@ interface SearchHeaderProps extends BaseFilter {
   Header: string;
   onSubmit: (val: string) => void;
   name: string;
-  toolTipDescription: string;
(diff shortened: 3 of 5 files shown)


superset-frontend/src/pages/SavedQueryList/index.tsx
```diff
@@ -449,6 +453,12 @@ function SavedQueryList({
         input: 'search',
         operator: FilterOperator.AllText,
       },
+      {
+        Header: t('Description'),
+        id: 'description',
+        key: 'description',
+        operator: FilterOperator.AllText,
```
nytai, line 460: The search filter should already be checking the description column, so is this really necessary?

superset-frontend/src/pages/SavedQueryList/index.tsx
```diff
@@ -297,6 +297,10 @@ function SavedQueryList({
         accessor: 'label',
         Header: t('Name'),
       },
+      {
+        accessor: 'description',
```
nytai, line 301: What happens if there's a really long description?

superset-frontend/src/pages/SavedQueryList/index.tsx
nytai: Instead we can rename the "Name" filter to be "Name and Description"

nytai: it's actually searching all text columns (name, description, sql, label), so it can probably just be called "search"

https://github.com/apache/superset/blob/f5843fe588112c5c303215945a36fda0845c9e99/superset/queries/saved_queries/filters.py#L30

superset-frontend/src/pages/SavedQueryList/index.tsx
imancrsrk: It's getting truncated which I believe is ok.
Screenshot 2024-05-05 at 7 19 16 PM

superset-frontend/src/pages/SavedQueryList/index.tsx
imancrsrk: @nytai I believe the ask is to show query description in a separate second column.

https://github.com/apache/superset/issues/12431#issuecomment-1310895547

imancrsrk:
@nytai All checks have passed. Is it good to merge this ?

superset-frontend/src/pages/SavedQueryList/index.tsx
```diff
@@ -443,7 +447,7 @@ function SavedQueryList({
   const filters: Filters = useMemo(
     () => [
       {
-        Header: t('Name'),
+        Header: t('search'),
```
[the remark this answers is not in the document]
rusackas: What's the impetus for this change? All the other list views say "Name" since they only search the Name. Does this one also search descriptions? Obviously it ought to be truthful, but otherwise I'd err on the side of consistency.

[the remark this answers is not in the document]
imancrsrk: Yes, it searches all text columns (name, description, sql, label), so it can probably just be called "search"

https://github.com/apache/superset/pull/28349#discussion_r1590479993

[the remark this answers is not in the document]
rusackas: If you have the wherewithal, it might be nice to add an info tooltip next to the header to state exactly that, so people know exactly what it's searching. Then (on another PR) it'd be good to carry this through to other list views, if any of them indeed search more than name. 

CC @yousoph @kasiazjc in case this merits tracking.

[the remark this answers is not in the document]
imancrsrk: ack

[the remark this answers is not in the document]
imancrsrk: @rusackas I've added the tooltip & updated the screenshot in description - 

Screenshot 2024-05-06 at 4 25 32 PM

imancrsrk:
@rusackas Hoping you can take another look at the changes we've so far and comment on whether anything else is needed before this pr can be merged and issue resolved?

rusackas:
The tooltip looks strangely high up, otherwise we're looking good, and thanks for the touchup! I'll try to find time in a bit to pull your branch and see about using the tooltip component's props to align it, but I'm working on something else at the moment.

imancrsrk:
> The tooltip looks strangely high up, otherwise we're looking good, and thanks for the touchup! I'll try to find time in a bit to pull your branch and see about using the tooltip component's props to align it, but I'm working on something else at the moment.

ack. I'll try to align it as well later in the evening.

imancrsrk:
> > The tooltip looks strangely high up, otherwise we're looking good, and thanks for the touchup! I'll try to find time in a bit to pull your branch and see about using the tooltip component's props to align it, but I'm working on something else at the moment.
> 
> ack. I'll try to align it as well later in the evening.

@rusackas Hope it looks better now.
Screenshot 2024-05-07 at 9 20 12 PM

superset-frontend/src/components/InfoTooltip/index.tsx
```diff
@@ -83,7 +83,11 @@ export default function InfoTooltip({
       overlayStyle={overlayStyle}
       color={bgColor}
     >
-      <Icons.InfoSolidSmall className="info-solid-small" viewBox={viewBox} />
+      <Icons.InfoSolidSmall
+        className="info-solid-small"
+        viewBox={viewBox}
+        iconSize="m"
```
rusackas, line 89: This part seems risky... I'm not sure how many places in the app this would affect. I suspect a lot. @geido do you remember our little alignment tricks here?

rusackas: At a minimum, we could set up the InfoTooltip to take an optional iconSize parameter and take it in here. Then, in the other file, instead of `<InfoTooltip tooltip={toolTipDescription} />` you could use `<InfoTooltip tooltip={toolTipDescription} iconSize="m" />`. Then it would only affect the one instance instead of all InfoTooltips in the app.

I swear we had another trick to vertically align these icons, though, besides making them teensy.

rusackas: You can avoid altering the InfoTooltip alltogether, and play with the viewbox attribute... it seems like `<InfoTooltip tooltip={toolTipDescription} viewBox='0 -7 28 28'/>` works OK. It's kind of an awkward-feeling attribute to fiddle with, though.

imancrsrk: I've pushed this viewBox change. But I'm curious to learn how it works. I tried fiddling with x, y, width, height values but I wasn't able to understand what was going on.

michael-s-molina:
@villebro worked on making all lists consistent (using Name) and may have some input here.

rusackas:
OK, I think we're good to go otherwise... I'll await feedback before unilaterally merging. I think Truthfulness outweighs Consitency, which outweighs pretty much any other random change ;)

imancrsrk:
> OK, I think we're good to go otherwise... I'll await feedback before unilaterally merging. I think Truthfulness outweighs Consitency, which outweighs pretty much any other random change ;)

@rusackas Just following up to find out if we need anything else in this PR or if we are good to merge this change.

rusackas (approved):
I'm not sure if there's a better way to fix an icon's size/position like this. This _feels_ wrong, but it works without tweaking the underlying component. Still curious if @geido has any better ideas, or if @villebro has any arguments against doing this. Meanwhile, I think it LGTM and will merge if we don't hear back soon.

rusackas:
Done!

Issue #28218, opened 2024-04-25
[ENABLE_TEMPLATE_PROCESSING] BaseTemplateProcessor interface (process_template) is misused


tseruga:
### Bug description

Unsure of when this regression happened, but it likely happened part of #26476 or #27470 

Within the `BaseTemplateProcessor` base class, we have the following method interface for `process_template()`:

https://github.com/apache/superset/blob/52f8734662cdeadba1a4ae80be3e2c74ad8f85a9/superset/jinja_context.py#L493-L504

Specifically, the type of `sql` should be a `str`.

It appears that this contract is being broken after one or more of the above changes. Instead, this method (when implemented and assigned to an engine) will sometimes be passed a `jinja2.nodes.Template` type object.

One example of this is here:
https://github.com/apache/superset/blob/52f8734662cdeadba1a4ae80be3e2c74ad8f85a9/superset/sql_parse.py#L1524-L1579

We see that we grab the template processor, then convert the raw `str` to a `Template` object:
```
    processor = get_template_processor(database)
    template = processor.env.parse(sql)
```

This `Template` object is then passed to the `process_template()` method:

```

    return (
        tables
        | ParsedQuery(
            sql_statement=processor.process_template(template),
            engine=database.db_engine_spec.engine,
        ).tables
    )
```

Which breaks the contract set by the `BaseTemplateProcessor` interface.

The example implementation of a custom template processor also makes the assumption that the argument will always be a string as shown here:

https://github.com/apache/superset/blob/master/docs/docs/configuration/sql-templating.mdx?plain=1#L123-L149

I believe this example implementation will also break upon receiving a `Template` object instead of a string.

### How to reproduce the bug

1. Enable `ENABLE_TEMPLATE_PROCESSING` in `config.py`
2. Implement a custom template processor extending the `BaseTemplateProcessor`
3. Within `process_template()` perform a string operation
Ex:
```
class ExampleTemplateProcessor(BaseTemplateProcessor):
    def process_template(self, sql: str, **kwargs: Any) -> str:
        return sql.upper()
```
4. Wire up the new template processor an engine that you can query (I used `sqllite` since that's what the example data uses)
```
DEFAULT_PROCESSORS = {
    "presto": PrestoTemplateProcessor,
    "hive": HiveTemplateProcessor,
    "trino": TrinoTemplateProcessor,
    "sqlite": ExampleTemplateProcessor,
}
```
[...]

PR #28357, opened 2024-05-06
fix(sqllab): sqllab/execute returns 500 when user only has schema access

### LINKED ISSUE #28145: Query history page is not usable in 3.1.2 and 4.0.0, caused by bug in sql_parse.py
### Bug description

Cannot accesse 'Query history' page, and get empty list with this error:

image

the error comes from sql_parse.py. check the logs below.

### How to reproduce the bug

Minimum sample:

1. Start superset with following docker-compose.yml
```docker-compose
version: "3.7"

services:
  superset:
    image: apachesuperset.docker.scarf.sh/apache/superset:3.1.2
    volumes:
      - superset_home:/app/superset_home
    environment:
      SUPERSET_SECRET_KEY: secretkey
      SUPERSET_DB_URI: sqlite:///superset_home/superset.db?check_same_thread=false
      SUPERSET_CONFIG_PATH: /app/superset_home/superset_config.py
    ports:
      - 8088:8088
    command:
      - "sh"
      - "-c"
      - |-
        # allow sqlite for easier reproduce
        echo 'PREVENT_UNSAFE_DB_CONNECTIONS=False' > /app/superset_home/superset_config.py
        # init database
        superset db upgrade
        # init admin user
        superset fab create-admin --username admin --firstname Superset --lastname Admin --email admin@superset.com --password admin
        # init roles
        superset init
        # run server
        /usr/bin/run-server.sh

volumes:
  superset_home:
    external: false

```

2. Go localhost:8088, login as admin/admin
3. Add one database, can just use 'sqlite:///superset_home/superset.db?check_same_thread=false'
4. Goto SQL->SQLLab, run some queries.
5. Goto SQL->Query hisory, and the error appears.

### Screenshots/recordings

image

### Superset version

3.1.2

### Python version

3.10

### Node version

Not applicable

### Browser

Chrome

### Additional context

tried both 3.1.2 and 4.0.0, and got same error with slightly different line numbers.

logs from 3.1.2:
```log
  File "/app/superset/views/base.py", line 252, in wraps
    return f(self, *args, **kwargs)
  File "/app/superset/views/base_api.py", line 487, in get_list_headless
    duration, response = time_function(super().get_list_headless, **kwargs)
  File "/app/superset/utils/core.py", line 1463, in time_function
    response = func(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/flask_appbuilder/api/__init__.py", line 1613, in get_list_headless
    response[API_RESULT_RES_KEY] = list_model_schema.dump(lst, many=True)
  File "/usr/local/lib/python3.10/site-packages/marshmallow/schema.py", line 557, in dump
    result = self._serialize(processed_obj, many=many)
  File "/usr/local/lib/python3.10/site-packages/marshmallow/schema.py", line 519, in _serialize
    return [
  File "/usr/local/lib/python3.10/site-packages/marshmallow/schema.py", line 520, in <listcomp>
    self._serialize(d, many=False)
  File "/usr/local/lib/python3.10/site-packages/marshmallow/schema.py", line 525, in _serialize
    value = field_obj.serialize(attr_name, obj, accessor=self.get_attribute)
  File "/usr/local/lib/python3.10/site-packages/marshmallow/fields.py", line 344, in serialize
    return self._serialize(value, attr, obj, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/marshmallow/fields.py", line 1991, in _serialize
    return self._serialize_method(obj)
  File "/app/superset/queries/schemas.py", line 76, in get_sql_tables
    return obj.sql_tables
  File "/app/superset/models/sql_lab.py", line 75, in sql_tables
    extract_tables_from_jinja_sql(
  File "/app/superset/sql_parse.py", line 1110, in extract_tables_from_jinja_sql
    | ParsedQuery(
  File "/app/superset/sql_parse.py", line 280, in __init__
    self._parsed = sqlparse.parse(self.stripped())
  File "/app/superset/sql_parse.py", line 517, in stripped
    return self.sql.strip(" \t\r\n;")
AttributeError: 'Template' object has no attribute 'strip'
```

### Checklist

toniphan21:
### SUMMARY

When a user has no admin access and is trying to run a query in SQL Lab, Superset needs to check:

- Database access
- Schema access
- Data source access

To be able to check schema access, Superset needs to know which table the user is trying to execute on and use `extract_tables_from_jinja_sql()` to get it. The function extract tables in both ways:

- Extract any tables referenced within the confines of specific Jinja macros.
- Parse SQL and get tables.

In the line I changed, there is simply a bug that uses `template` which has type Template. The correct one should be `sql`, which is a string sent by the user.

### TESTING INSTRUCTIONS

Reproduce steps:

1. Create a new user
2. Create a role with permission: "schema access on [examples].[main]"
3. Assign the user to the role and "sql_lab" role
4. Log in with the new user and go to SQL Lab.
5. Write a simple query and execute it. You will see that the `/api/v1/sqllab/execute/` endpoint returns 500

### Impact

- Fixes: https://github.com/apache/superset/issues/28145

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -1622,7 +1622,7 @@ def extract_tables_from_jinja_sql(sql: str, database: Database) -> set[Table]:
     return (
         tables
         | ParsedQuery(
-            sql_statement=processor.process_template(template),
+            sql_statement=processor.process_template(sql),
             engine=database.db_engine_spec.engine,
         ).tables
     )


rusackas:
If this does indeed resolve https://github.com/apache/superset/issues/28145 then please feel free to add "Fixes: #28145" do your description block and it'll auto-close that issue when merged. 

Meanwhile, running CI 🤞

tseruga:
Relevant to this issue as well: https://github.com/apache/superset/issues/28218

michael-s-molina (changes requested):
> Then, in this part, the given sql should be processed via the processor to create a raw sql_statement, which is then put into the ParsedQuery class to extract tables. If that is the purpose, hence the variable passed to processor.process_template(...) should be a given SQL, which may include Jinja macros.

@toniphan21 Given the following:

> Actuality for (2) we need to parse "sanitized" SQL, i.e., we the Jinja templates rendered or removed so that SQLGlot is able to parse it.

We can't pass `sql` to `processor.process_template` because we need the processed template with the following:

```
# Replace the potentially problematic Jinja macro with some benign SQL.
node.__class__ = nodes.TemplateData
node.fields = nodes.TemplateData.fields
node.data = "NULL" 
```

The question that we need to answer is why [Template.render](https://github.com/apache/superset/blob/226b75579835ba340b7314c2a935833943df42ae/superset/jinja_context.py#L504) is not returning a `str` in accordance to its signature:

```
def render(self, *args: t.Any, **kwargs: t.Any) -> str:
```

rusackas:
This has been awating response to change requests for a long time, and now needs a rebase. We'd love to see it through, but I'll mark it as draft for now. Please mark it as "ready for review" when the time comes, otherwise it may eventually be closed due to inactivity.

PR #28343, opened 2024-05-04
fix(ci): correct input type for `allow-dependencies-licenses` in Dependency Review GH action

hainenber:
fix(ci): correct input type for `allow-dependencies-licenses` for Dependency Review GH action

### SUMMARY

Some of my PRs got failed in the GH action due to incorrect input type for said GH action. This change is to restore the CI job's functionality.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before
image

After
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #28358, opened 2024-05-06
chore: add a github "action-validator" in CI

mistercrunch:
### SUMMARY

https://github.com/apache/superset/pull/28343 fixes an issue where I had broken a specific github action workflow in a previous PR without realizing I did. If/when altering the yaml, if the yaml itsefl is valid (we have a check for that) yet the schema is not - either through passing the right data structure or required types, the action just won't run and the PR is still mergeable. 

NOTE: it's really easy to break a github action (yaml is brittle, and schema enforcement for actions is loose), and it's also really easy to miss the error message about it. The failure gets somewhat buried on the Actions tab (at the bottom of https://github.com/apache/superset/actions), which is super easy to miss.

Originally I tried the pre-commit hook, but bailed on that approach as it requires installing `action-validator` in PATH, which is not super convenient as it's a Rust package. There's a node wrapper we leverage in this action, but didn't want to extend the dev deps beyond python/pip.

Find out more about the hook here: https://github.com/mpalmer/action-validator

### TESTING

For testing, I reverted the change from #28343, and tested the bash script fails. It did and provided a nice json blob with everything wrong with the schema provided.

### SCREENSHOT
Screenshot 2024-05-06 at 11 02 18 AM

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -19,8 +19,8 @@ on:
         default: 'false'
         description: Whether to force a latest tag on the release
         options:
-          - true
-          - false
+          - 'true'
+          - 'false'
 jobs:
   config:
     runs-on: "ubuntu-latest"
--- a/.github/workflows/github-action-validator.sh
+++ b/.github/workflows/github-action-validator.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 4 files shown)


mistercrunch:
Converting to DRAFT as that action requires a Rust package, which requires a Rust package manager, which isn't convenient for developers - probably not worth it.

Either I:
- find some sort of easy pip wrapper for the package
- make it a CI only thing, so it gets caught by CI instead of as a commit hook

.github/workflows/tag-release.yml
```diff
@@ -19,8 +19,8 @@ on:
         default: 'false'
         description: Whether to force a latest tag on the release
         options:
-          - true
-          - false
+          - 'true'
```
mistercrunch, line 22: it caught this as an error so I fixed it

eschutho (approved):
nice!

betodealmeida (approved):

PR #28361, opened 2024-05-06
chore: fix master build by merging alembic migration heads

mistercrunch:
See title, this happens when merging two database migration without rebasing first

## ERROR
```
ERROR [flask_migrate] Error: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
```

Screenshot 2024-05-06 at 11 39 57 AM

### DIFF
--- a/superset/migrations/versions/2024-05-06_11-37_645bb206f96c_.py
+++ b/superset/migrations/versions/2024-05-06_11-37_645bb206f96c_.py
@@ -1,0 +1,35 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


john-bodley (approved):
Thanks for the fix.

rusackas (approved):

john-bodley:
BTW @mistercrunch https://github.com/apache/superset/pull/28361 was just merged today and we could likely just change the revision keys in the associated migration file if we didn't want to add another migration.

john-bodley:
@mistercrunch I wonder if there could be a GitHub action—something which I'm not overly well versed with—that could enforce that any PR which touches a migration file must be rebased against `master` before merging?

mistercrunch:
> I wonder if there could be a GitHub action ...

I think there's a timing issue here where while things may be ok at the time the PR is update (and CI runs), it doesn't mean things are ok after when the main branch has changed. We'd need something like pre-merge hook but I don't think they exist.

mistercrunch:
One approach would be to try and "auto-fix" with a script that would run when `push`ing on `master`, but not great to push this straight on the main branch...

That or maybe merge queues? https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

PR #28363, opened 2024-05-07
fix: don't strip SQL comments in Explore

mistercrunch:
Following up in the footsteps of https://github.com/apache/superset/pull/27725 as comments can be used as a way to pass the database engine hints, or for logging purposes.

### DIFF
--- a/scripts/tests/run.sh
+++ b/scripts/tests/run.sh
@@ -138,6 +138,5 @@ fi
 
 if [ $RUN_TESTS -eq 1 ]
 then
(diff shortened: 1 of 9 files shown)


john-bodley (approved):
@mistercrunch I agree with this approach. I think historically we did this to aid with traversing the `sqlparse` AST, but as we migrate to SQLGlot I suspect this requirement can likely be relaxed.

mistercrunch:
Found some comments in the code around stripping comments prior to checkit `is_select()` and other similar assertions and leaving those untouched.

PR #28364, opened 2024-05-07
fix: Enable explore button on SQL Lab view when connected to Apache Pinot as a database

### LINKED ISSUE #13623: 'Explore' button on SQL Lab view disabled when connected to Apache Pinot as a database
I connected superset to pinot successfully and was able to build SQL lab queries only to find out that Superset does not support Exploring of SQL Lab virtual data as a chart if the connected database is Apache Pinot. (The “Explore” button is disabled)

I am using Superset version 1.0.1

#### Screenshot
image

soumitra-st:
Older versions of Pinot did not support subquery, hence the explore view on SQL Lab was disabled. [Apache Pinot 1.0](https://docs.pinot.apache.org/reference/multi-stage-engine) added support for subquery, join, etc. in multi-stage query engine, so this PR enables the explore button.

### Create chart button is disabled before the fix
create-chart-disabled

### Create chart button is enabled after the fix
Screenshot 2024-05-06 at 6 32 59 PM

### Query in explore view
explore-view

### Query with alias in SQL Lab
alias-in-query

### Query with alias in explore
explore-query-with-alias

### ADDITIONAL INFORMATION
Fixes #13623 issue
- [x] Changes UI

### DIFF
--- a/superset/db_engine_specs/pinot.py
+++ b/superset/db_engine_specs/pinot.py
@@ -25,10 +25,10 @@ from superset.db_engine_specs.base import BaseEngineSpec
 class PinotEngineSpec(BaseEngineSpec):
     engine = "pinot"
     engine_name = "Apache Pinot"
-    allows_subqueries = False
-    allows_joins = False
-    allows_alias_in_select = False
-    allows_alias_in_orderby = False
+    allows_subqueries = True
+    allows_joins = True
+    allows_alias_in_select = True
+    allows_alias_in_orderby = True
 
     # https://docs.pinot.apache.org/users/user-guide-query/supported-transformations#datetime-functions
     _time_grain_expressions = {
--- a/superset/db_engine_specs/pinot.py
+++ b/superset/db_engine_specs/pinot.py
@@ -25,10 +25,6 @@ from superset.db_engine_specs.base import BaseEngineSpec
 class PinotEngineSpec(BaseEngineSpec):
     engine = "pinot"
     engine_name = "Apache Pinot"
-    allows_subqueries = True
-    allows_joins = True
-    allows_alias_in_select = True
-    allows_alias_in_orderby = True
 
     # https://docs.pinot.apache.org/users/user-guide-query/supported-transformations#datetime-functions
     _time_grain_expressions = {
--- a/docs/docs/configuration/databases.mdx
+++ b/docs/docs/configuration/databases.mdx
@@ -1035,6 +1035,11 @@ The expected connection string using username and password is formatted as follo
 pinot://<username>:<password>@<pinot-broker-host>:<pinot-broker-port>/query/sql?controller=http://<pinot-controller-host>:<pinot-controller-port>/verify_ssl=true``
 ```
 
+If you want to use explore view or joins, window functions, etc. then enable [multi-stage query engine](https://docs.pinot.apache.org/reference/multi-stage-engine).
+Add below argument while creating database connection in Advanced -> Other -> ENGINE PARAMETERS
+```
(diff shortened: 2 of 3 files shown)


superset/db_engine_specs/pinot.py
```diff
@@ -25,10 +25,10 @@
 class PinotEngineSpec(BaseEngineSpec):
     engine = "pinot"
     engine_name = "Apache Pinot"
-    allows_subqueries = False
-    allows_joins = False
-    allows_alias_in_select = False
-    allows_alias_in_orderby = False
+    allows_subqueries = True
```
john-bodley, line 28: These might not be necessary if they're the same as the base version.

john-bodley (commented):
@soumitra-st thanks for the change. Per [here](https://github.com/python-pinot-dbapi/pinot-dbapi) it seems like the current version of the DB-API only supports Pinto version 0.9.3 (I'm not sure if this is a single version or a limit bound) and thus I don't think we can merge this.

Additionally this change would be deemed breaking for anyone using an older version of Pinot and thus cannot be merged (as is) until the breaking window for 5.0 opens.

In the interim you can locally override your version of the `PinotEngineSpec` to include the relevant settings.

soumitra-st:
Thanks @john-bodley for your comments. IMO, the PR is fixing a bug not breaking existing functionality. Wondering why is it a breaking change?

I tested the changes using `pinotdb==5.2.0` and Pinot version `1.1.0`. I will work on updating the supported version of [pinotdb module](https://github.com/python-pinot-dbapi/pinot-dbapi?tab=readme-ov-file#python-db-api-and-sqlalchemy-dialect-for-pinot) to 1.1.0. If I can get a release of pinotdb that is compatible with Pinot version 0.9.3 onwards, then would this change still be considered as breaking?

soumitra-st:
@john-bodley I checked internally, the latest pinotdb module supports Pinot version 1.1.0, and we have merged the PR to update the version [here](https://github.com/python-pinot-dbapi/pinot-dbapi?tab=readme-ov-file#python-db-api-and-sqlalchemy-dialect-for-pinot).

rusackas:
Running CI 🤞 Please feel free to open a commit adding your org to our [In the Wild page](https://github.com/apache/superset/blob/master/RESOURCES/INTHEWILD.md) if you want to get a trivial PR under your belt and remove the need to manually trigger CI on every change.

soumitra-st:
> breaking window for 5.0 opens

@john-bodley , shall we merge this PR as it is? If not, what is the timeline for 5.0 release?

xiangfu0:
Latest [pinotdb](https://pypi.org/project/pinotdb/) version: 5.3.0 supports newer version of Pinot.
Can you please also update the pinotdb version to `pinot = ["pinotdb>=5.0.0, <6.0.0"]` here: https://github.com/apache/superset/blob/master/pyproject.toml#L154

sfirke:
Does pinotdb 5.3.0 support earlier versions of pinot than 1.1.0?  If so then it seems like this would not be a breaking change and could be merged now in my opinion.

Even if not, I expect this is worthwhile to merge at 5.0.0 and pinot users would need to update their pinot version along with the Superset upgrade.

xiangfu0:
> Does pinotdb 5.3.0 support earlier versions of pinot than 1.1.0?  If so then it seems like this would not be a breaking change and could be merged now in my opinion.
> 
> 
> 
> Even if not, I expect this is worthwhile to merge at 5.0.0 and pinot users would need to update their pinot version along with the Superset upgrade.

Correct, this is not a breaking change. New client library supports both old and new query endpoints/protocols.

sfirke (approved):

PR #28346, opened 2024-05-05
refactor: Migration of AnnotationLayerControl to TypeScript

EnxDev:
### SUMMARY

Migrate AnnotationLayerControl to TypeScript

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N.A

### TESTING INSTRUCTIONS

- All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #28370, opened 2024-05-07
refactor: Migration of Chart to TypeScript

EnxDev:
### SUMMARY

Migrate Chart to TypeScript

N.A.

### TESTING INSTRUCTIONS

- All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -40,69 +41,71 @@ import ChartRenderer from './ChartRenderer';
...
-  onFilterMenuClose: PropTypes.func,
-  ownState: PropTypes.object,
-  postTransformProps: PropTypes.func,
-  datasetsStatus: PropTypes.oneOf(['loading', 'error', 'complete']),
-  isInView: PropTypes.bool,
-  emitCrossFilters: PropTypes.bool,
-};
+export interface ChartProps {
+  annotationData?: object;
+  actions: any;
```
kgabryje, line 46: Can we use more specific types? In general we try to avoid using very generic types like `any` or `object`

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -40,69 +41,71 @@ import ChartRenderer from './ChartRenderer';
...
+  onFilterMenuClose?: Function;
+  ownState?: any;
+  postTransformProps?: Function;
+  datasetsStatus?: 'loading' | 'error' | 'complete';
+  isInView?: boolean;
+  emitCrossFilters?: boolean;
+  renderStartTime?: Date;
+}
 
+export type queryResponse = {
```
kgabryje, line 85:
```suggestion
export type QueryResponse = {
```

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -40,69 +41,71 @@ import ChartRenderer from './ChartRenderer';
...
-  dashboardId: null,
-  chartStackTrace: null,
+  dashboardId: undefined,
+  chartStackTrace: undefined,
   force: false,
   isInView: true,
 };
 
-const Styles = styled.div`
+const Styles = styled.div<{ height: number; width?: number }>`
```
kgabryje, line 108: We don't use `width` prop in `Styles` so maybe let's just remove it?

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -150,9 +153,16 @@ const MonospaceDiv = styled.div`
...
   white-space: pre-wrap;
 `;
+class Chart extends React.PureComponent<ChartProps, {}> {
+  static defaultProps = defaultProps;
+
+  renderStartTime: any;
 
-class Chart extends React.PureComponent {
-  constructor(props) {
+  renderContainerStartTime: number;
```
kgabryje, line 161: Is `renderContainerStartTime` actually used anywhere?

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -150,9 +153,16 @@ const MonospaceDiv = styled.div`
   overflow-x: auto;
   white-space: pre-wrap;
 `;
+class Chart extends React.PureComponent<ChartProps, {}> {
+  static defaultProps = defaultProps;
+
+  renderStartTime: any;
```
kgabryje, line 159: Where does `renderStartTime` come from? In the old version, we're only reading it, but we don't assign any value to it. In your changes you put `renderStartTime` in ChartProps, but also as class's field. I don't think it's actually passed as a prop anywhere, and as a class field it's also only read, not assigned.
So maybe we should just remove it?

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -150,9 +153,16 @@ const MonospaceDiv = styled.div`
...
 
-class Chart extends React.PureComponent {
-  constructor(props) {
+  renderContainerStartTime: number;
+
+  static propTypes: any;
```
kgabryje, line 163: I don't think that's necessary?

kgabryje (changes requested):
Thanks for the contribution! Added a few comments

superset-frontend/src/components/Chart/Chart.tsx
EnxDev: I removed it, thank you

superset-frontend/src/components/Chart/Chart.tsx
EnxDev: I removed it, thank you

superset-frontend/src/components/Chart/Chart.tsx
EnxDev: I removed it, thank you

superset-frontend/src/components/Chart/Chart.tsx
EnxDev: I removed this from ChartProps. 
You mean we should remove him from line 198 and 200 ( start_offset and duration keys)?

kgabryje: Yeah, unless I misunderstand something here, renderStartTime is always undefined because we don’t set its value anywhere, so maybe we could remove it from the code. @rusackas any objections?

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -214,7 +209,7 @@ class Chart extends React.PureComponent<ChartProps, {}> {
     });
   }
 
-  renderErrorMessage(queryResponse: queryResponse) {
+  renderErrorMessage(QueryResponse: QueryResponse) {
```
kgabryje, line 212: The argument should be lowercase, only the type is PascalCase

superset-frontend/src/components/Chart/Chart.tsx
kgabryje: Unless we want to actually fix it and assign a timestamp

EnxDev: Yes, I think we should understand why it was put there.
I remain waiting for instructions, thank you.

superset-frontend/src/components/Chart/ChartErrorMessage.tsx
```diff
@@ -21,10 +21,16 @@ import React from 'react';
...
 import ErrorMessageWithStackTrace from 'src/components/ErrorMessage/ErrorMessageWithStackTrace';
+import { ChartSource } from 'src/types/ChartSource';
 
-interface Props {
+export interface Props {
   chartId: string;
   error?: SupersetError;
+  subtitle: JSX.Element;
+  copyText: string;
+  link: string | undefined;
```
justinpark, line 31: It's equivalent to 
```suggestion
  link?: string;
```

rusackas: There's gotta be a lint rule for that we can add if there's a preference here.

superset-frontend/src/components/Chart/Chart.tsx
EnxDev: For the annotation `property`, I believe the interface I created in this [PR](https://github.com/apache/superset/pull/28346)  should be suitable.
For the `actions` property I created a type.

What do you think?

superset-frontend/src/components/Chart/Chart.tsx
rusackas: Well, if it's not wired up, that's an odd problem. It sure seems like it'd be useful for logging, which seems to be its intent. There's one in `ChartRenderer.jsx` that IS wired up, for reference. Maybe we don't need both though? 

It WAS set in the [original commit](https://github.com/apache/superset/commit/5c02e3199b924cc8a45622e660096d6fdc8c5857), actually. I haven't checked to see when its initialization was _removed_ but clearly there wasn't much outcry.

I'd err on the side of more logging, not less. Maybe @eschutho would be interested in taking more advantage of this insight :)

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -36,73 +38,102 @@ import { getUrlParam } from 'src/utils/urlUtils';
...
+  chartStackTrace?: string;
+  queriesResponse?: QueryResponse[];
+  triggerQuery?: boolean;
+  chartIsStale?: boolean;
+  errorMessage?: React.ReactNode;
+  addFilter?: Function;
+  onQuery?: MouseEventHandler<HTMLSpanElement>;
+  onFilterMenuOpen?: Function;
+  onFilterMenuClose?: Function;
+  ownState?: any;
```
geido, line 79: Can we add a better type for this?

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -201,7 +238,7 @@ class Chart extends React.PureComponent {
     });
   }
 
-  renderErrorMessage(queryResponse) {
+  renderErrorMessage(QueryResponse: QueryResponse) {
```
geido, line 241:
```suggestion
  renderErrorMessage(queryResponse: QueryResponse) {
```

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -36,73 +38,102 @@ import { getUrlParam } from 'src/utils/urlUtils';
...
+  onFilterMenuOpen?: Function;
+  onFilterMenuClose?: Function;
+  ownState?: any;
+  postTransformProps?: Function;
+  datasetsStatus?: 'loading' | 'error' | 'complete';
+  isInView?: boolean;
+  emitCrossFilters?: boolean;
+}
+
+export type QueryResponse = {
```
geido, line 86: This assumes that the response will always be an error. I think we already have a correct type for this. We either need to adjust that to account for the error or create one specifically for when get an error:

Instead of this QueryResponse see:` superset-frontend/packages/superset-ui-core/src/query/types/QueryResponse.ts` `ChartDataResponse`

For the error maybe `superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts` `ClientErrorObject`

superset-frontend/src/components/Chart/ChartErrorMessage.tsx
```diff
@@ -21,10 +21,16 @@ import React from 'react';
...
 
-interface Props {
+export interface Props {
   chartId: string;
   error?: SupersetError;
+  subtitle: JSX.Element;
+  copyText: string;
+  link?: string;
+  source: ChartSource;
+  stackTrace?: string;
```
geido, line 33: I believe this could be an extension of `superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts` `ClientErrorObject`

superset-frontend/src/components/Chart/Chart.tsx
EnxDev: renderErrorMessage(queryResponse: QueryResponse)

superset-frontend/src/components/Chart/Chart.tsx
eschutho: We can prob remove it and add it back in later if we find that we need it.

kgabryje (approved):

PR #28378, opened 2024-05-08
chore: module scope should not require the app context

mistercrunch:
This is a major issue that's been plaguing the backend for a long time. Currently you can't simply run a simple `from superset import models` (or pretty much any other module import) without getting an error about the app context missing, and/or circular imports

This DRAFT PR tries to evaluate what's getting in the way. So far I've identified:

- app.config being referenced in module scope, this is mostly easy to fix for common configuration, where we can rely on flask.current_app and avoid module scope
- dynamic/configurable model: this seems to be the core issue, where say we let people configure their EncryptedField's type
- some reliance on SecurityManager, and SecurityManager.user_model used as a relationship, I think this can be worked around using sqlalchemy's `relationship("use-a-string-reference")` call
- ???

TODO:
- I think we'd have to deprecate `SQLALCHEMY_ENCRYPTED_FIELD_TYPE_ADAPTER` altogether (?) unless we can find some solution
- **utils** should never depend on the app, the app depends on utils

TO-LOOK-INTO
- Having a different way to parse the config that doesn't depend on the app context (?)


superset/models/core.py
```diff
@@ -1071,9 +1069,22 @@ def get_oauth2_config(self) -> OAuth2ClientConfig | None:
         return self.db_engine_spec.get_oauth2_config()
 
 
-sqla.event.listen(Database, "after_insert", security_manager.database_after_insert)
```
dpgaspar, line 1074: Alternatively we can move these registrations into here: https://github.com/apache/superset/blob/master/superset/initialization/__init__.py#L414

I'm personally not in favour of stuff happening at the root level of modules

mistercrunch: I like the idea of centralizing them there, but screams circular imports, let me see

mistercrunch:
Sad to close this. I may give it another shot with a series of small PR one day

PR #28387, opened 2024-05-08
chore: remove and deprecate old CSS templates endpoints

dpgaspar:
### SUMMARY
Removing server side rendered pages endpoints from CSS templates view.
Also deprecating `csstemplateasyncmodelview/api/read` endpoint. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/css_templates.py
+++ b/superset/views/css_templates.py
@@ -15,14 +15,18 @@
 # specific language governing permissions and limitations
 # under the License.
 from flask_appbuilder.api import expose
+from flask_appbuilder.baseviews import expose_api
 from flask_appbuilder.models.sqla.interface import SQLAInterface
-from flask_appbuilder.security.decorators import has_access
-from flask_babel import lazy_gettext as _
+from flask_appbuilder.security.decorators import (
+    has_access,
+    has_access_api,
+    permission_name,
+)
 
 from superset.constants import MODEL_VIEW_RW_METHOD_PERMISSION_MAP, RouteMethod
 from superset.models import core as models
 from superset.superset_typing import FlaskResponse
-from superset.views.base import DeleteMixin, SupersetModelView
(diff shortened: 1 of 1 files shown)


superset/views/css_templates.py
```diff
@@ -54,8 +48,15 @@ def list(self) -> FlaskResponse:
...
+    include_route_methods = RouteMethod.API_READ
     class_permission_name = "CssTemplate"
     method_permission_name = MODEL_VIEW_RW_METHOD_PERMISSION_MAP
 
     list_columns = ["template_name", "css"]
+
+    @expose_api(name="read", url="/api/read", methods=["GET"])
+    @has_access_api
+    @permission_name("list")
+    @deprecated(eol_version="5.0.0")
```
dpgaspar, line 60: Could we just remove this? not used on the frontend and a REST API exists for this for some time now

Antonio-RiveroMartnez (approved):

rusackas:
@mistercrunch might wanna scope out the CI issue.

superset/views/css_templates.py
michael-s-molina: @dpgaspar We discussed this during the release meeting and we agreed to remove these legacy endpoints during a breaking window even if they are not under `v1`. That's what we did in the past and we just want to be mindful that someone out there might still be using the endpoint.

mistercrunch:
mmmh, ran the same command as CI but locally and it worked:
```
$ supersetbot orglabel --issue 28387 --repo apache/superset
🟢 SUCCESS: added label(s) preset-io to issue 28387
```
re-triggered CI. Looked like what happens when trying to reference an issue that does not exist yet... Not sure what happened.

mistercrunch:
This is not a required check, feel free to merge with that check failing while I look into it

PR #28393, opened 2024-05-08
chore(Databricks): New Databricks driver

Vitor-Avila:
### SUMMARY
The `sqlalchemy-databricks` package ([reference](https://github.com/crflynn/sqlalchemy-databricks)) was deprecated in favor of the updated version of `databricks-sql-python` ([reference](https://github.com/databricks/databricks-sql-python)). This new version allows users to connect to Databricks using:

```
databricks://token:dapi***@***.cloud.databricks.com?http_path=/sql/***&catalog=**&schema=**
```

As a result, some more recent features are only supported with this driver. For example, if you create a table in Databricks with a `CLUSTER BY` configuration, it's not possible to create a dataset using `databricks+connector://` as it doesn't successfully handle this metadata.

This PR introduces a new engine spec for the updated driver version, exposing a dynamic form. I believe long-term would be beneficial to deprecate the other engine specs (migrating existing connections to the new one) but for now this PR only implements support for the driver to validate with users on compatibility.

The dynamic form also includes fields for `catalog` and `schema`, so this could also help with catalog support.

New dynamic form:

image

### TESTING INSTRUCTIONS
1. Connect to Databricks using the **Databricks Python Connector** option in the database dropdown.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/CommonParameters.tsx
```diff
@@ -116,6 +116,69 @@ export const databaseField = ({
...
+    value={db?.parameters?.default_schema}
+    validationMethods={{ onBlur: getValidation }}
+    errorMessage={validationErrors?.default_schema}
+    placeholder={t('e.g. default')}
+    label={t('Default Schema')}
+    onChange={changeMethods.onParametersChange}
+    helpText={t('The default schema that should be used for the connection.')}
+  />
+);
+export const httpPathField = ({
```
Vitor-Avila, line 159: Had to create a new field for **HTTP Path** as the existing one included logic to retrieve/insert information in the `engine_params`

Vitor-Avila:
Going to implement tests in a future commit

betodealmeida:
@Vitor-Avila you need to rebase to fix the failing tests, @dpgaspar fixed them in `master` today.

betodealmeida (approved):
Looks good!

I think we're at a point where it might be useful to split the `databricks.py` file into multiple, perhaps something like:

- `databricks/odbc.py`
- `databricks/hive.py`
- `databricks/native.py`
- `databricks/python.py`

(Not sure about the names.)

We also need to start looking in organizing the frontend components a little bit, so it's easier to build more dynamic forms.

Vitor-Avila:
thank you @betodealmeida 🙏 I think long term would be interesting to at least deprecate the `connector` driver and migrate existing connections to this new driver. I'm not too familiar with Databricks to tell if we still need to keep `hive` and `odbc` or if those could also be migrated. 

Feel free to merge this PR if you want to continue working on catalogs -- I'll create a follow up PR to implement unit tests for this new driver.

PR #28396, opened 2024-05-09
fix: type annotation breaking on py3.9

dpgaspar:
### SUMMARY
A recent change as broke Superset on python 3.9. To my understanding we still support python 3.9 as per https://github.com/apache/superset/blob/master/pyproject.toml#L27

I do think we should run our tests on the minimal python version as well to avoid these issues, but I'm not sure on how to change it on github actions, or we don't set it or we use python-version: ["current", "next"]

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/pandas_postprocessing/contribution.py
+++ b/superset/utils/pandas_postprocessing/contribution.py
@@ -14,6 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from __future__ import annotations
+
 from decimal import Decimal
 from typing import Any
 


michael-s-molina:
Thanks for the fix @dpgaspar!

> I do think we should run our tests on the minimal python version as well to avoid these issues, but I'm not sure on how to change it on github actions, or we don't set it or we use python-version: ["current", "next"]

That's a fair point. I didn't even noticed that we were missing the `from __future__ import annotations` because neither the tests or pre-commit hook complained about it. I believe we'll have many more of these problems if these checks don't catch these types of errors.

@mistercrunch Is it possible to fix this?

michael-s-molina (approved):

mistercrunch:
https://github.com/apache/superset/pull/28419 should help prevent this

PR #28419, opened 2024-05-10
fix: run some CI tests against previous python version

mistercrunch:
As currently python3.10 is our main version, we do run a bit of CI (not
the full matrix) against 3.11.

This is introducing doing the same for 3.9 as we do for 3.11. Check
isn't required, but should show as failed and inform something broke 3.9
support

Now that refactor/cleaned up some of the CI / GHAs, it's pretty easy to get this going.

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -29,6 +29,8 @@ runs:
           echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
           echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
+        elif [ "${{ inputs.python-version }}" = "previous" ]; then
+          echo "PYTHON_VERSION=3.9" >> $GITHUB_ENV
         else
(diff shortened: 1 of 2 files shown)


mistercrunch:
Validated it installed python3.9 and ran integration tests here -> https://github.com/apache/superset/actions/runs/9026120305/job/24802927512?pr=28419#step:5:49

michael-s-molina (approved):

michael-s-molina:
Thank you! ❤️

.github/actions/setup-backend/action.yml
```diff
@@ -29,6 +29,8 @@ runs:
           echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
           echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
+        elif [ "${{ inputs.python-version }}" = "previous" ]; then
```
john-bodley, line 32: @mistercrunch I wonder if `previous`, `current`, and `next` are somewhat abstract, i.e., what does "previous" actually mean? Would it be simpler to just use the version name?

dpgaspar (approved):

PR #28422, opened 2024-05-10
fix: Update migration logic in #27119

john-bodley:
### SUMMARY

I realize it's generally perceived as poor form to change an existing migration, however per [here](https://github.com/apache/superset/blob/f29e1e4c29a46f7d607cfa59adb8bb21d107091c/superset/migrations/versions/2019-05-06_14-30_afc69274c25a_alter_sql_column_data_type_in_query_mysql_table.py#L36-L46) the `query. executed_sql` and `query. select_sql` were previously defined as type `LONGTEXT` rather than `TEXT`, thus  https://github.com/apache/superset/pull/27119 resulted in:

1. Downsizing these columns to `MEDIUMTEXT` could result in data loss.
2. Altering the `query` table unnecessarily can be rather expensive given the size of the table. 

This PR remove mutating these columns unnecessarily.

### TESTING INSTRUCTIONS

Tested the migration using a MySQL metadata database.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/sql_lab.py
+++ b/superset/models/sql_lab.py
@@ -59,7 +59,13 @@ from superset.models.helpers import (
 )
 from superset.sql_parse import CtasMethod, extract_tables_from_jinja_sql, Table
 from superset.sqllab.limiting_factor import LimitingFactor
-from superset.utils.core import get_column_name, MediumText, QueryStatus, user_label
+from superset.utils.core import (
+    get_column_name,
+    LongText,
+    MediumText,
+    QueryStatus,
+    user_label,
+)
 
 if TYPE_CHECKING:
     from superset.connectors.sqla.models import TableColumn
@@ -110,11 +116,11 @@ class Query(
     sql_editor_id = Column(String(256), index=True)
     schema = Column(String(256))
     catalog = Column(String(256), nullable=True, default=None)
-    sql = Column(MediumText())
+    sql = Column(LongText())
     # Query to retrieve the results,
     # used only in case of select_as_cta_used is true.
-    select_sql = Column(MediumText())
-    executed_sql = Column(MediumText())
+    select_sql = Column(LongText())
+    executed_sql = Column(LongText())
     # Could be configured in the superset config.
     limit = Column(Integer)
     limiting_factor = Column(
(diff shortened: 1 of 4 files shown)


mistercrunch:
Tagging as "may require downtime" knowing that mysql takes table-level lock for this type of operation and that this table tend to get large, and is busy with polling (sqllab client poking at query status).

michael-s-molina:
Thanks for fixing that @john-bodley 🙏🏼

michael-s-molina (approved):

michael-s-molina:
@john-bodley Before merging this PR, can you do the additional improvement of adding a `LongText` type similar to [MediumText](https://github.com/apache/superset/blob/7da6147bf8c3894a5f34a6c6b40ba14da6256400/superset/utils/core.py#L1496)? I believe this error was caused because the [models](https://github.com/apache/superset/pull/27119/files#diff-94d3bd02afcbd1983033ed9c33e9efb466cb9a64e3b1f3954e6e0d627f70d0ef) were defined as `Text` so you could only see the real type in the migration scripts. This type or error can be avoided if we have `LongText`.

superset/models/sql_lab.py
```diff
@@ -110,11 +116,11 @@ class Query(
     sql_editor_id = Column(String(256), index=True)
     schema = Column(String(256))
     catalog = Column(String(256), nullable=True, default=None)
-    sql = Column(MediumText())
+    sql = Column(LongText())
```
michael-s-molina, line 119: @john-bodley Just to confirm.. is `sql` `LongText` or `MediumText`? The reason I'm asking is because it's not being affected by your script.

john-bodley: Per [here]([here](https://github.com/apache/superset/blob/f29e1e4c29a46f7d607cfa59adb8bb21d107091c/superset/migrations/versions/2019-05-06_14-30_afc69274c25a_alter_sql_column_data_type_in_query_mysql_table.py#L36-L46)) `sql` is `LongText` and isn't impacted by any of the aforementioned migrations.

michael-s-molina (approved):

john-bodley:
Given we can't cherry-pick new migrations, I thought it would be prudent to break this PR into two,

1. _This_ PR which "fixes" the existing migration by preventing the downsizing.
2. https://github.com/apache/superset/pull/28482 which adds a new migration to resizes any wrongfully downsized columns.

as that way we can cherry-pick (1) into the 4.0 branch.

Issue #28353, opened 2024-05-06
Slack API `files.upload` v1 deprecation


asaf400:
### Bug description

The API superset uses for slack reports has been deprecated, and needs to be replaced with the new API:

[As described here](https://api.slack.com/methods/files.upload)
`files.upload is deprecated and will stop functioning on March 11, 2025. Use [files.getUploadURLExternal](https://api.slack.com/methods/files.getUploadURLExternal) and [files.completeUploadExternal](https://api.slack.com/methods/files.completeUploadExternal) to upload files instead. Newly created apps will be unable to use files.upload beginning May 8, 2024. See [Uploading files](https://api.slack.com/messaging/files#uploading_files) for more details on the process and [this changelog](https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay) for more on the deprecation.
`

Seems like new slack-sdk version required minimum is [v3.19.0](https://github.com/slackapi/python-slack-sdk/releases/tag/v3.19.0)
Which Superset already uses https://github.com/apache/superset/blob/f5843fe588112c5c303215945a36fda0845c9e99/pyproject.toml#L88

```
[root@LAPTOP-ASAF-T14 ~]# pip install 'slack_sdk>=3.19.0, <4'
Collecting slack_sdk<4,>=3.19.0
...
Installing collected packages: slack_sdk
Successfully installed slack_sdk-3.27.1
```

additional resources: 
https://github.com/slackapi/python-slack-sdk/pull/1272

### How to reproduce the bug

Deprecation

### Screenshots/recordings

As a slack organization administrator I received the following message by email:
 
<img width="422" alt="2024-05-06 12_40_45-Gmail - Upcoming changes to support for the files upload API method - asaf levy@" src="https://github.com/apache/superset/assets/14169006/d61cd2b6-50f3-4896-b361-7f3ba98664c3">

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [x] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

mistercrunch:
https://github.com/apache/superset/pull/28423 should do it, not sure how to test/validate though

eschutho:
[...]

PR #28423, opened 2024-05-10
fix: move to slack-sdk files_upload_v2

mistercrunch:
Randomly caught https://github.com/apache/superset/issues/28353 and decided to propose a fix. This seemed straightforward when I looked at the docs here https://github.com/slackapi/python-slack-sdk/releases/tag/v3.19.0

Confirmed we're on the latest client and that stuff is pretty well covered by unit tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ dependencies = [
     "shortid",
     "sshtunnel>=0.4.0, <0.5",
     "simplejson>=3.15.0",
-    "slack_sdk>=3.19.0",
+    "slack_sdk>=3.19.0, <4",
     "sqlalchemy>=1.4, <2",
     "sqlalchemy-utils>=0.38.3, <0.39",
     "sqlglot>=23.0.2,<24",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ dependencies = [
     "shortid",
     "sshtunnel>=0.4.0, <0.5",
     "simplejson>=3.15.0",
-    "slack_sdk>=3.19.0, <4",
+    "slack_sdk>=3.19.0",
     "sqlalchemy>=1.4, <2",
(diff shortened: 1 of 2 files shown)


geido (approved):

pyproject.toml
```diff
@@ -85,7 +85,7 @@ dependencies = [
     "shortid",
     "sshtunnel>=0.4.0, <0.5",
     "simplejson>=3.15.0",
-    "slack_sdk>=3.19.0, <4",
+    "slack_sdk>=3.19.0",
```
john-bodley, line 88: Why remove the upper bound? Typically we try to bind dependencies to their current major version.

mistercrunch: Let's talk about this! I wanted to start by noting that those boundaries are used by raw "pip install apache-superset" (where having an upper bound is probably a good idea, fully pinned for that use case might be an even better approach...), and by @supersetbot to auto-generate PRs bumping pinned libs, within the boundaries set here.

First question is what should happen when people in the wild just go `pip install apache-superset`, all of our installation docs / docker / CI is pretty clear about running the pinned dep installs first as in `pip install -r requirements/base.txt` FIRST to get to something that's actually likely to work. Now knowing that people may not do that, should we 1. pin stuff in the package ? or 2. identify ranges of what's likely to work (kind of what we have now with semver upper bounds). Note that the approach of "what's likely to work" ranges is fairly subjective, and there's a fair amount of things that can go wrong with this approach. Say if  the `cryptography` lib has a major fix I NEED in semver+1, the semver package upper bound restriction is actually preventing me from getting what I need. All sorts of other non-pinned stuff can still go wrong. 

One thing is I think we need to differentiate semantics around real breaking bounds (as in "we know library `>=N.x` won't work!") and hopeful ones (as in "given what we know about semver, library > N.x could have breaking changes...").

Anyhow, that's a complex topic, let's chat.

mistercrunch:
For now let me remove that line

PR #28425, opened 2024-05-10
feat: clarify that 'Text' supports markdown

mistercrunch:
Making it more clear that the "Text" layout element in dashboards supports Markdown by relabeling to "Text / Markdown"

Screenshot 2024-05-09 at 9 54 20 PM

### DIFF
--- a/superset-frontend/src/dashboard/components/gridComponents/new/NewMarkdown.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/new/NewMarkdown.jsx
@@ -28,7 +28,7 @@ export default function DraggableNewDivider() {
     <DraggableNewComponent
       id={NEW_MARKDOWN_ID}
       type={MARKDOWN_TYPE}
-      label={t('Text')}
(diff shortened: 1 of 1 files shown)


sfirke (approved):

rusackas:
It was called Markdown before... I'm not sure why we moved away from that in the first place, but perhaps there was a good reason. Maybe @kasiazjc or @yousoph remember?

rusackas:
I could even see a world where it's a "Rich Text" component and we eventually support Markdown/Handlebars/Jinja, or even add proper WYSIWYG rich text controls that use markdown without the user knowing.

mistercrunch:
GPT suggested "Rich Text" too but had to assert a momentary superiority over AI to try and preserve some self-worth :)

rusackas:
Oh, fine :P I look forward to the "Text / WYSIWYG / Markdown / Jinja / Handlebars / HTML / Liquid / Mustache" component ;)

PR #28430, opened 2024-05-10
chore(docs): fix two broken Docusaurus redirect links

sfirke:
One per Evan in Slack, one I believe was a typo in the previous fix (was email-reports instead of alerts-reports, so the redirect from alerts-reports doesn't happen).


rusackas (approved):

rusackas:
image

LOL... I know that feeling all too well.

PR #28441, opened 2024-05-12
chore(ci): use latest Ubuntu runner for CI

hainenber:
chore(ci): use latest Ubuntu runner for CI

### SUMMARY

Chore PR to keep CI up-to-date and hopefully some additional efficiency from running CI with latest Ubuntu flavor

N/A

### TESTING INSTRUCTIONS

All green CI can be considered as acceptance threshold 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/bump-python-package.yml
+++ b/.github/workflows/bump-python-package.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   bump-python-package:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     permissions:
       actions: write
       contents: write
--- a/.github/workflows/cancel_duplicates.yml
+++ b/.github/workflows/cancel_duplicates.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   cancel-duplicate-runs:
     name: Cancel duplicate workflow runs
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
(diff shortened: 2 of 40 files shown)


dpgaspar (commented):
I personally prefer to run on a pinned version, can you update to 24.04, latest LTS?

mistercrunch:
I agree the decision to pin, though it'd be nice to be able to keep this DRY by referencing a variable, though I don't think there's any obvious way to do this short or implementing our own templating solution around GHA yaml (let's not do that just yet). Also would be great if something like dependabot was to submit a PR when there's a new version up. 

Oh one more thing is I think we should align the base docker with CI too, but looks like GHA is mostly Ubuntu and base python images we use are Debian. I guess it show lining this up doesn't matter too much. It's kind of crazy we do all this testing on a Ubuntu base and then ship a Debian-based build into the world through docker.

.github/workflows/bump-python-package.yml
```diff
@@ -17,7 +17,7 @@ on:
 
 jobs:
   bump-python-package:
-    runs-on: ubuntu-latest
```
john-bodley, line 20: What's wrong with using `ubuntu-latest` everywhere? Is the idea that we fully pin/freeze our CI environments?

Issue #28372, opened 2024-05-07
SQL Lab hard fails with "'utf-8' codec can't decode byte 0xe6 in position 0: invalid continuation byte" when reading certain binary types


whitleykeith:
### Bug description

See https://github.com/apache/superset/issues/28001 for context

I attempted to validate the fix from https://github.com/apache/superset/pull/28266 but I'm still getting the same error as described in the issue.

### How to reproduce the bug

1. Have a Trino with VARBINARY column
2. Add binary, non-utf8 encodable data
3. Try to query it from SQL Lab

### Superset version

`467e612-dev` 

### Python version

3.10

### Node version

I don't know

### Browser

Chrome

### Additional context

Stack trace:

```
superset   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1484, in full_dispatch_request
superset     rv = self.dispatch_request()
superset   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1469, in dispatch_request
superset     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
superset   File "/usr/local/lib/python3.10/site-packages/flask_appbuilder/security/decorators.py", line 95, in wraps
superset     return f(self, *args, **kwargs)
superset   File "/app/superset/views/base_api.py", line 127, in wraps
superset     raise ex
superset   File "/app/superset/views/base_api.py", line 121, in wraps
superset     duration, response = time_function(f, self, *args, **kwargs)
superset   File "/app/superset/utils/core.py", line 1491, in time_function
superset     response = func(*args, **kwargs)
superset   File "/usr/local/lib/python3.10/site-packages/flask_appbuilder/api/__init__.py", line 182, in wraps
superset     return f(self, *args, **kwargs)
superset   File "/app/superset/utils/log.py", line 263, in wrapper
superset     value = f(*args, **kwargs)
superset   File "/app/superset/sqllab/api.py", line 346, in get_results
superset     payload = json.dumps(
superset   File "/usr/local/lib/python3.10/site-packages/simplejson/__init__.py", line 395, in dumps
superset     **kw).encode(obj)
superset   File "/usr/local/lib/python3.10/site-packages/simplejson/encoder.py", line 298, in encode
superset     chunks = self.iterencode(o)
superset   File "/usr/local/lib/python3.10/site-packages/simplejson/encoder.py", line 379, in iterencode
superset     return _iterencode(o, 0)
[...]

PR #28486, opened 2024-05-14
fix: utf-16 json encoder support

eyalezer:
### SUMMARY

Following the [pessimistic json encoder](https://github.com/apache/superset/pull/28266) fix and an issue reported in https://github.com/apache/superset/issues/28372 which fails to decode UTF-16 bytes in SQL Lab. Added UTF-16 encoding support for json serialization.

### DIFF
--- a/superset/sqllab/execution_context_convertor.py
+++ b/superset/sqllab/execution_context_convertor.py
@@ -52,13 +52,22 @@ class ExecutionContextConvertor:
 
     def serialize_payload(self) -> str:
         if self._exc_status == SqlJsonExecutionStatus.HAS_RESULTS:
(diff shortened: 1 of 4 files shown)


mistercrunch (approved):
LGTM - thanks for the fix

mistercrunch:
Looks like tests are failing ...

eyalezer:
indeed missed one integration test and also i see now some lint issues, will fix both

eyalezer:
@mistercrunch - Fixed and tested locally both formats and integration tests

eyalezer:
**Don't merge yet!**

Discovered additional instances throughout the system that are still uncovered.

PR #28497, opened 2024-05-14
fix: Correction translation

aehanno:
Correction french translation

### SUMMARY
Remove to fuzzy to add the right translation

### BEFORE
image

### TESTING INSTRUCTIONS
Check that after we got the new translation


aehanno:
@rusackas There is my new PR
I don't if we waiting the end of the discussion about the .json file before merging mine ?

aehanno:
@rusackas and @villebro, can we go forward with this PR ?

rusackas (approved):

aehanno:
Hi @rusackas, I see the tag 4.0.2 has been release but without this PR
How I can put it in the next one ?

rusackas:
@michael-s-molina would have the most context on the patch releases, but I can say that 4.1 (coming soon, pending timeline from @sadpandajoe) should include this, since it'll be a fresh cut from the `master` branch.

PR #28496, opened 2024-05-14
fix(docs): fix broken indexed link from Google search

sfirke:
https://superset.apache.org/docs/contributing/translations/ is the first Kagi search hit for "superset translations contribute", but it's broken.  Here's a redirect.

PR #28499, opened 2024-05-14
docs: creating a redirect for a legacy link about pre-commit hook

rusackas:
### SUMMARY

I google for the pre-commit hook docs all the time, and Google thinks it's a dead-end. Hopefully this fixes it.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch (approved):

sfirke:
When should one redirect via the .htaccess file vs. in Docusaurus?  Compare this to my PR here: https://github.com/apache/superset/pull/28496 

It looks like this way supports linking to a subheading link, which had my pending PR failing CI, so maybe this is better?

rusackas:
I'm actually not sure... as it stands, this PR did not actually work. 

https://superset.apache.org/docs/contributing/hooks-and-linting/
^^ the link here is a dead-end, as of this writing

I'm also not sure if a frontend redirect _eventually_ teaches the googlebots where to point. One would think so.

PR #28521, opened 2024-05-15
feat: unpack payload into log function

mistercrunch:
Unpack the payload (merged from the ? querystring and POST data, depending on context) and pass it to the log function. In my specific case, I was looking to do analysis on around how much of the calls to `ChartDataRestApi.data` are using "force=true", and realize while this is in context, it doesn't flow through my current logger. By adding this one line of code, it should allow for many more (in fact ALL querystring and POST payload) parameters to flow through the logger. From there custom loggers can have access to much more context. Note that by default, the loggers we have won't do anything with the new arguments passed, and essentially flush it.

### DIFF
--- a/tests/integration_tests/event_logger_tests.py
+++ b/tests/integration_tests/event_logger_tests.py
@@ -141,15 +141,18 @@ class TestEventLogger(unittest.TestCase):
             with logger(action="foo", engine="bar"):
                 pass
 
-        assert logger.records == [
-            {
-                "records": [{"path": "/", "engine": "bar"}],
-                "database_id": None,
-                "user_id": 2,
-                "payload": {"engine": "bar", "path": "/"},
-                "duration": 15000,
-            }
-        ]
+        self.assertEquals(
+            logger.records,
+            [
+                {
+                    "records": [{"path": "/", "engine": "bar"}],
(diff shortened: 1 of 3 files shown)


dpgaspar:
POST payload can have sensitive data

mistercrunch:
> POST payload can have sensitive data

Totally - though that variable is just passed to the logger, not logged. Actually let me improve this a bit. I'll pass more explicitely a `curated_payload` and have a class-level attribute with an allow-list of `curated_payload_params`

mistercrunch:
@dpgaspar I took a different approach, let me know what you think

tests/integration_tests/event_logger_tests.py
```diff
@@ -183,20 +188,25 @@ def log(
...
+                        {
+                            "path": "/",
+                            "object_ref": {"baz": "food"},
+                            "payload_override": {"engine": "sqlite"},
+                        }
+                    ],
+                    "database_id": None,
+                    "user_id": 2,
+                    "duration": 5558756000,
+                    "curated_payload": {},
```
dpgaspar, line 205: would be nice to actually have tests with the curated payload

dpgaspar (approved):

PR #28567, opened 2024-05-17
fix: Revert "fix: don't strip SQL comments in Explore (#28363)"

michael-s-molina:
### SUMMARY
This reverts commit c618767c6b4b7c1c921807e4f942586353110114.

After deploying Superset 4.0 to production a number of users reported issues where the rendered SQL statements were invalid if the underlying virtual dataset contained comments—likely trailing comments. Specifically the subquery lacked the trailing `)`.

We were able to determine that #28363 was the problematic PR and reverting it remedied the issue.

### TESTING INSTRUCTIONS

CI and confirmed that the issue was mitigated after reverting.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -1134,8 +1134,9 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         cte = None
         sql_remainder = None
         sql = sql.strip(" \t\n;")
+        sql_statement = sqlparse.format(sql, strip_comments=True)
         query_limit: int | None = sql_parse.extract_top_from_query(
-            sql, cls.top_keywords
+            sql_statement, cls.top_keywords
         )
         if not limit:
             final_limit = query_limit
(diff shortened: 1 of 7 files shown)


john-bodley (approved):

mistercrunch:
Mmmmh. Seems we'll need to start over and write some unit tests with trailing comments (?)

villebro:
@mistercrunch this seems easy enough to fix this by adding a newline at the end if the query contains `--`. I've done something similar for adhoc metrics/columns back in the day, and there should be unit tests for those.

michael-s-molina:
> Seems we'll need to start over and write some unit tests with trailing comments (?) Did we save some of that SQL that was demonstrated to fail? Were they dash-dash comments or dash-star comments?

@mistercrunch We had problems with dash-dash comments. We can definitely test a future PR using our production data as multiple dashboards failed.

mistercrunch:
A problem with dash-star is they don't embed into one-another well. My use case was around looking for a user hint (in the form of a dash-dash comment) and based on that altering the SQL. Now I wanted for the "view query" panel to show them the before and after transformation, but the dash-dash turns slash-star, and the end of their hint closes the opening /*.

Solution is either to keep dash-dash as is, or have the whole SQL_QUERY_MUTATOR feature have a way to keep track of the before and after and surface both (as opposed to doing this as a comment)

PR #28568, opened 2024-05-18
fix: add listener to repaint on visibility change for canvas

eschutho:
### SUMMARY

We're seeing with the latest release of Chrome, some canvas elements on the dashboard will disappear, and only reappear on mouseover. I believe that there was an experimental feature in Chrome that saved memory by removing canvas elements when the browser tab wasn't active, and it just went GA. 

Echart's library zrender is responsible for animating the canvas on mouseover which is making the canvas visible again, but we need another solution to repaint the canvas when the document is visible again. 

Before:

https://github.com/apache/superset/assets/5186919/450cfd71-e5ee-4796-9e6f-eb46ae5424fd

After:

https://github.com/apache/superset/assets/5186919/5dab80b9-657f-4a71-b47c-f74e6312f3db

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/package.json
+++ b/superset-frontend/package.json
@@ -316,6 +316,7 @@
     "ignore-styles": "^5.0.1",
     "imports-loader": "^3.1.1",
     "jest": "^26.6.3",
+    "jest-canvas-mock": "^2.5.2",
     "jest-environment-enzyme": "^7.1.2",
     "jest-enzyme": "^7.1.2",
     "jest-websocket-mock": "^2.2.0",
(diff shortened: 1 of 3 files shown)


mistercrunch:
This appears to fix the behavior in the Dashboard app, but not in explore (?) should we do this lower level in the Chart component that's ?

eschutho:
@mistercrunch it doesn't seem to affect explore or individual charts, but only dashboards. One issue with adding it to explore is that we need to assign an event listener to the window, and only one event listener is allowed at a time, or rather each time we create a new event listener, it will overwrite the last one. This change uses the already existing `visibilitychange` event listener used for logging.

mistercrunch:
Oh interesting, I'm positive I've seen the behavior in explore too. I'm on Mobile now so I can't tell reproduce... I think it was big number, blank after switching tab, reappears on hover.

eschutho:
Ok, I was just able to repro on explore. I can add another event listener there, but outside of the slice so that we don't wind up with multiple event listeners.

lilykuang (approved):

michael-s-molina:
@eschutho It looks like they [disabled the feature](https://issues.chromium.org/issues/328755781#comment107) that was causing the problem in 125.0.6422.76 which is already available. Can we revert this PR to avoid introducing logic that fixes a dependency problem?

PR #28570, opened 2024-05-18
chore(tags): Handle tagging as part of asset update call

Vitor-Avila:
### SUMMARY
Currently, if a change is made to the Tags configuration for an asset via the PropertiesModal, the modification is handled in a separate request. This causes two issues:
* Content owners with `write access on $asset_type` have the impression they can change the tags configuration (since they can save it).
* The operation silently fails -- no error in the UI.

This PR moves the tag handling to the API call to update asset, providing an error to the user if permissions are missing. The PR also introduces a new permission so that users that don't have `can write on Tag` can still add existing tags to the assets they own (and have permission to modify).

**BEFORE**

https://github.com/apache/superset/assets/96086495/916eab85-63a8-4425-ae05-fdef310400e2

**AFTER**

Had to add to GDrive due to filesize

https://drive.google.com/file/d/1141h-OqqvOFH4Ok4yXISwKCzIs3TP9gZ/view?usp=sharing

### TESTING INSTRUCTIONS
Added both integration and unit tests. For manual testing:
1. Log in with a role that doesn't have `can write on Tag` but has `can write on Dashboard`. 
2. Edit the properties for a dashboard, changing the Tags config.
3. Validate the change happens in the same network call and that errors are properly displayed.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/schemas.py
+++ b/superset/charts/schemas.py
@@ -122,6 +122,7 @@ description_markeddown_description = "Sanitized HTML version of the chart descri
 owners_name_description = "Name of an owner of the chart."
 certified_by_description = "Person or group that has certified this chart"
 certification_details_description = "Details of the certification"
+tags_description = "Tags to be associated with the chart"
 
 openapi_spec_methods_override = {
     "get": {"get": {"summary": "Get a chart detail information"}},
@@ -284,7 +285,9 @@ class ChartPutSchema(Schema):
     )
     is_managed_externally = fields.Boolean(allow_none=True, dump_default=False)
     external_url = fields.String(allow_none=True)
-    tags = fields.Nested(TagSchema, many=True)
+    tags = fields.List(
+        fields.String(metadata={"description": tags_description}, allow_none=True)
+    )
 
 
 class ChartGetDatasourceObjectDataResponseSchema(Schema):
--- a/superset/charts/schemas.py
+++ b/superset/charts/schemas.py
@@ -27,7 +27,6 @@ from marshmallow.validate import Length, Range
 from superset import app
 from superset.common.chart_data import ChartDataResultFormat, ChartDataResultType
 from superset.db_engine_specs.base import builtin_time_grains
-from superset.tags.models import TagType
 from superset.utils import pandas_postprocessing, schema as utils
 from superset.utils.core import (
     AnnotationType,
@@ -142,12 +141,6 @@ openapi_spec_methods_override = {
         }
     },
 }
-
-
-class TagSchema(Schema):
-    id = fields.Int()
-    name = fields.String()
-    type = fields.Enum(TagType, by_value=True)
 
 
 class ChartEntityResponseSchema(Schema):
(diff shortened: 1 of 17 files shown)


superset/charts/schemas.py
```diff
@@ -284,7 +285,9 @@ class ChartPutSchema(Schema):
     )
     is_managed_externally = fields.Boolean(allow_none=True, dump_default=False)
     external_url = fields.String(allow_none=True)
-    tags = fields.Nested(TagSchema, many=True)
```
hughhhh, line 287: Where are we removing the `TagSchema` reference here?

superset/commands/utils.py
```diff
@@ -102,3 +107,79 @@ def get_datasource_by_id(datasource_id: int, datasource_type: str) -> BaseDataso
         )
     except DatasourceNotFound as ex:
         raise DatasourceNotFoundValidationError() from ex
+
+
+def validate_tags(
+    object_type: str,
```
hughhhh, line 113: couldn't this be ENUM of resource `CHART || DASHBOARD || QUERY`

superset/charts/schemas.py
Vitor-Avila: that's a good catch -- I scanned for un-used imports but didn't realize that `TagSchema` was actually defined in this same file. I'll remove its definition in a follow up commit.

superset/commands/utils.py
Vitor-Avila: def could! I was already using `ObjectType` for the `update_tags` method, but here thought about going with `string` because I'm using `security_manager.can_access` to check if the user has the required perm, which is case sensitive. We have two options here:
1. Create a new `enum` with:
``` python
class AssetTypes(enum.Enum):
    CHART = "Chart"
    DASHBOARD = "Dashboard"
```
2. Use the existing `ObjectType` and then call `can_access` using `capitalize()`:
``` python
if not security_manager.can_access("can_tag", object_type.name.capitalize()):
```

@hughhhh would you have any preferences?

superset/charts/schemas.py
Vitor-Avila: I'll hold on committing this change ☝️ until I get your feedback on the other comment.

superset/commands/utils.py
hughhhh: Let's go with #2 using `ObjectType` enum

superset/charts/schemas.py
hughhhh: So we should prolly centralize `TagSchema` somewhere it could be used in all the resource update schemas

hughhhh:
Also @Vitor-Avila we need to update the calls for `Query` and `Dataset` to make sure they have the same ability to update the tags as well

Vitor-Avila:
> Also @Vitor-Avila we need to update the calls for `Query` and `Dataset` to make sure they have the same ability to update the tags as well

@hughhhh for Saved Queries I'm not sure if it makes sense since you can't really assign a Tag via the UI to a single query -- you can only assign tags to Saved Queries via the **BULK SELECT** (and in this case I think it makes sense to use the `bulk_create` approach which is already the case).

Tags are not fully supported with Datasets yet (not sure if the association works, but the Datasets list doesn't show Tags, the Datasets properties modal doesn't show tags, Datasets schemas don't have tags, etc) so this would be a bigger change a bit out of the scope of this PR.

hughhhh (approved):
🚢 LGTM

PR #28597, opened 2024-05-20
feat: add Nightingale chart support for echarts pie chart

ghost:
### SUMMARY

Add Nightingale Chart Support for Echarts Pie Chart

Echarts now supports Nightingale charts in pie charts, with two optional modes:

1. `radius`: Uses the central angle to show the percentage of data and the radius to show data size.
2. `area`: All sectors share the same central angle, with data size shown only through radiuses.

I have added a new option, 'ROSE TYPE,' to support this feature.

Before:
image

After:
image
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


rusackas:
Pretty cool! Are you aware of any feature-parity issues that would prevent a migration from the legacy rose chart to this? I'm always excited to deprecate/remove a legacy plugin in favor of a newer option :)

Also, would you mind adding this to the React Storybook entry? We're making a renewed effort to capture as many variants as possible there (interactive or static).

superset-frontend/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
```diff
@@ -87,6 +88,23 @@ const config: ControlPanelConfig = {
...
+            name: 'roseType',
+            config: {
+              type: 'SelectControl',
+              label: t('Rose Type'),
+              default: roseType,
+              renderTrigger: true,
+              choices: [
+                ['area', t('Area')],
+                ['radius', t('Radius')],
+                [false, t('None')],
```
michael-s-molina, line 102:
```suggestion
                [null, t('None')],
```

superset-frontend/plugins/plugin-chart-echarts/src/Pie/types.ts
```diff
@@ -45,6 +45,7 @@ export type EchartsPieFormData = QueryFormData &
     numberFormat: string;
     dateFormat: string;
     showLabelsThreshold: number;
+    roseType?: 'radius' | 'area';
```
michael-s-molina, line 48:
```suggestion
    roseType: 'radius' | 'area' | null;
```

superset-frontend/plugins/plugin-chart-echarts/src/Pie/types.ts
```diff
@@ -78,6 +79,7 @@ export const DEFAULT_FORM_DATA: EchartsPieFormData = {
   labelsOutside: true,
   showLabelsThreshold: 5,
   dateFormat: 'smart_date',
+  roseType: undefined,
```
michael-s-molina, line 82:
```suggestion
  roseType: null,
```

michael-s-molina (commented):
Thank you for the PR @hexcafe. Could you also add a screenshot of a Nightingale chart to the Pie viz gallery examples and make sure the Pie chart appears if you search for Nightingale? This will allow people to discover the chart. 

Right now we have 4 screenshots. I think you can remove the last 2 and add yours.

Screenshot 2024-05-21 at 08 44 10

michael-s-molina:
> Pretty cool! Are you aware of any feature-parity issues that would prevent a migration from the legacy rose chart to this? I'm always excited to deprecate/remove a legacy plugin in favor of a newer option :)

@rusackas I checked and the legacy plugin has no additional feature that would prevent us from migrating.

Screenshot 2024-05-21 at 08 54 38
Screenshot 2024-05-21 at 08 54 49

hexcafe:
@michael-s-molina Thanks! I have updated the metadata. Now, we can discover the chart by searching for "Nightingale".
image

superset-frontend/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
hexcafe: I think it's better to keep the value as false; otherwise, the select control will not behave correctly, and we will never get a chance to choose the None option.

hexcafe:
@rusackas @michael-s-molina  Should we upgrade to ECharts v5.5.0 soon? please take a look at #27341 , Thanks!

superset-frontend/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
michael-s-molina: @hexcafe `null` [is a valid value](https://github.com/search?q=repo%3Aapache%2Fsuperset++%5Bnull%2C+t%28%27None%27%29%5D%2C&type=code) for the select control. If you prefer to use `False`, then you need to change the Typescript definition accordingly as it's allowing `undefined`.

hexcafe:
Hi @michael-s-molina ,

Sorry for the delay. I have fixed the type errors. I hope this PR can be merged soon.

Thank you!

michael-s-molina (approved):
@hexcafe Thank you for the feature and for addressing all comments!

rusackas (approved):
LGTM2! Thank you for this! Approving CI to run, and will spin up a test environment just to give it one last test drive :)

rusackas:
Works great! Thanks again!

PR #28609, opened 2024-05-20
fix: dashboard performance

dpgaspar:
### SUMMARY
This PR introduces a substantial performance improvement to the dashboard list page.
We have noticed that `json_metadata` and `position_json` can contain huge payloads on dashboards that contain a large amount of charts (hundreds). This can heavily impact performance.

The easier path would be to just remove those columns from `list_columns` on the backend, but thats a breaking change. Also `thumbnail_url` uses `position_json` and `json_metadata` to compute the MD5 digest. So currently it's not possible to improve the REST API performance itself by default.

it's possible on FAB to just request the columns we want using the `columns` query parameter, but the SQL selected columns would still use the configures `list_columns` or `list_select_columns`. On 4.5.0 I've introduced a new query parameter named `select_columns` that works exactly like `columns` but will actually on query the requested columns (without never "escaping" the `list_columns` set configured on the backend, for security reasons).

Using the example dashboards:

Requesting just the columns we need defined on this PR, avg API response is ~16ms

Using the API by default (backend defined columns), avg API response is ~120ms

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


requirements/base.txt
```diff
@@ -15,8 +15,6 @@ apispec[yaml]==6.3.0
     # via flask-appbuilder
 apsw==3.42.0.1
     # via shillelagh
-async-timeout==4.0.3
```
mistercrunch, line 18: just noting that this particular pinned dep has been flickering, meaning `pip-compile-multi` isn't as deterministic as I'd like for it to be, for reasons I don't really understand

mistercrunch (approved):
LGTM overall. It's nice to be able to specify the columns needed for the list view

PR #28613, opened 2024-05-20
fix: revert fix(presto preview): re-enable schema previsualization for Trino/Presto table/schemas"

john-bodley:
### SUMMARY

This PR reverts apache/superset#26782 which was merged last week. Regrettably said logic seems to be problematic and throws a error:

```
  ...
  File "superset/db_engine_specs/presto.py", line 512, in where_latest_partition
    if col_type_class := getattr(types, col_type_name, None):
TypeError: getattr(): attribute name must be string
```

given that `col_type_name` isn't a string but rather a SQLAlchemy SQL type representing the column type—as denoted by the `column_type_by_name` mapping.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

cc: @brouberol

### DIFF
--- a/superset/db_engine_specs/presto.py
+++ b/superset/db_engine_specs/presto.py
@@ -536,10 +536,7 @@ class PrestoBaseEngineSpec(BaseEngineSpec, metaclass=ABCMeta):
         }
 
         for col_name, value in zip(col_names, values):
-            col_type = None
-            if col_type_name := column_type_by_name.get(col_name):
-                if col_type_class := getattr(types, col_type_name, None):
-                    col_type = col_type_class()
+            col_type = column_type_by_name.get(col_name)
 
             if isinstance(col_type, types.DATE):
                 col_type = Date()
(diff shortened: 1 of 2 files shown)


brouberol:
Feel free to revert, obviously! However I believe that the initial state was buggy in itself (at least that's what we experienced). If you can, could you add the values of the following variables?

- `col_names`
- `values`
- `column_type_by_name`

If somehow something changed in the types exposed by the `col_names` and `values` variable, we could adapt (or even remove) the logic I added. My only doubt is whether reverting will get us back to a working state, or into the initial buggy state.

john-bodley:
@brouberol during testing I saw values of the form (santized):

- `col_names`: `['ds']`
- `values`: `('2024-05-20',)`
- `column_type_by_name` `{'ds': VARCHAR(), ...}`

michael-s-molina (approved):

PR #28627, opened 2024-05-21
fix(engine): update databend engine spec for dialect version >=0.4.6

rad-pat:
### SUMMARY

`databend-sqlalchemy` moved from the pure-Python `databend-py` client to a Rust core, and the two disagree about TLS. The old client read `secure` and defaulted to an `http` scheme; the Rust core reads `sslmode` and defaults to `https`. An unrecognised parameter is not rejected — it is kept as a session variable — so a stored `?secure=false` is silently ignored and the connection switches to TLS against a server that may not speak it.

This PR emits `sslmode` in both directions, still reads the legacy `secure` spelling so existing connections repopulate the form, and migrates stored URIs.

While in the same two methods, it fixes some long-standing faults that made the dynamic form unusable for Databend:

- `get_parameters_from_uri` was declared with `*_args`, so the keyword call from `Database.parameters` raised `TypeError`; and it mutated `url.query`, which is an `immutabledict`. Both were swallowed by that property's bare `except`, so `Database.parameters` returned `{}` for every Databend connection and the form came back empty. This also meant the legacy-`secure` fallback could never actually run.
- `sslmode=enable` (which the driver accepts alongside `require`) read as *unencrypted*, so saving the form would rewrite a TLS connection to `sslmode=disable`.
- Values were compared case-sensitively, but `databend-py` parsed `secure` with `asbool` — `secure=True` genuinely meant encrypted.

The two methods now use the same signatures and the `URL.create` idiom as the other engine specs.

Per @betodealmeida's review, `encryption_disable_parameters` is also added to `BasicParametersMixin`, for drivers that need an explicit opt-out rather than treating absence of the encryption parameters as "off". It defaults to an empty dict, so no other spec changes behaviour.

The `pyproject.toml` bump this PR originally carried is gone — master is already on `databend-sqlalchemy>=0.5.5`, and `sslmode` is still the parameter the current driver reads.

#### On the migration

`upgrade()` rewrites `secure=true|false` to `sslmode=require|disable`. It also pins connections that carry **no** TLS parameter to `sslmode=disable`: those were plaintext under `databend-py`'s `http` default (Superset only ever wrote `secure=true`, never `secure=false`), so leaving them alone would let the new `https` default break exactly the connections the migration exists to protect.

The query string is edited one parameter at a time rather than parsed and re-rendered, because `URL.render_as_string` sorts and re-encodes the whole query and would rewrite unrelated parameters on every touched row.

### TESTING INSTRUCTIONS

Validate against a Databend instance, over both a TLS and a plaintext endpoint, creating a connection through the dynamic form and reopening it to confirm the fields repopulate.

Unit tests cover the URI build/parse round trip, the legacy `secure` fallback, the `encrypted_extra` keyword regression, the mixin's opt-out parameters, and the migration's upgrade/downgrade including the no-TLS-parameter case.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

Migration runtime: a single filtered pass over `dbs` (`sqlalchemy_uri LIKE 'databend%'`) via `paginated_update`, touching only Databend rows. On any realistic instance this is a handful of rows and completes in well under a second; no downtime expected. `downgrade()` restores `secure`, which is semantically but not textually exact — a row that had no TLS parameter before `upgrade()` returns as `secure=false` rather than bare, and casing is normalised. Both forms mean the same thing to `databend-py`.

### DIFF
--- a/superset/db_engine_specs/databend.py
+++ b/superset/db_engine_specs/databend.py
@@ -195,10 +195,15 @@ class DatabendEngineSpec(BasicParametersMixin, DatabendBaseEngineSpec):
     supports_file_upload = False
 
     sqlalchemy_uri_placeholder = (
-        "databend://user:password@host[:port][/dbname][?secure=value&=value...]"
+        "databend://user:password@host[:port][/dbname][?sslmode=value&=value...]"
     )
     parameters_schema = DatabendParametersSchema()
-    encryption_parameters = {"secure": "true"}
+    encryption_parameters = {"sslmode": "require"}
+    encryption_disable_parameters = {"sslmode": "disable"}
+
+    # every ``sslmode`` the driver resolves to an https scheme; it accepts both
+    # spellings, so a hand-written ``sslmode=enable`` must not read as plaintext
+    encryption_sslmodes = frozenset({"require", "enable"})
 
     metadata = {
         "description": (
@@ -214,7 +219,7 @@ class DatabendEngineSpec(BasicParametersMixin, DatabendBaseEngineSpec):
         ],
         "pypi_packages": ["databend-sqlalchemy"],
         "connection_string": (
-            "databend://{username}:{password}@{host}:{port}/{database}?secure=true"
+            "databend://{username}:{password}@{host}:{port}/{database}?sslmode=require"
         ),
         "default_port": 443,
         "parameters": {
@@ -259,29 +264,73 @@ class DatabendEngineSpec(BasicParametersMixin, DatabendBaseEngineSpec):
 
     @classmethod
     def build_sqlalchemy_uri(
-        cls, parameters: BasicParametersType, *_args: dict[str, str] | None
+        cls,
+        parameters: BasicParametersType,
+        encrypted_extra: dict[str, str] | None = None,
     ) -> str:
-        url_params = parameters.copy()
-        if url_params.get("encryption"):
-            query = parameters.get("query", {}).copy()
-            query.update(cls.encryption_parameters)
-            url_params["query"] = query
-        if not url_params.get("database"):
-            url_params["database"] = "__default__"
-        url_params.pop("encryption", None)
-        return str(URL(f"{cls.engine}", **url_params))
+        """
+        Build a Databend URI, always stating the TLS mode explicitly.
+
+        The driver honours ``sslmode`` rather than inferring TLS from the port,
+        so an unencrypted connection needs ``sslmode=disable`` spelled out
+        rather than simply omitting the encryption parameters.
+        """
+        query = parameters.get("query", {}).copy()
+        query.update(
+            cls.encryption_parameters
+            if parameters.get("encryption")
+            else cls.encryption_disable_parameters
+        )
+
+        return str(
+            URL.create(
+                cls.engine,
+                username=parameters.get("username"),
+                password=parameters.get("password"),
+                host=parameters.get("host"),
+                port=parameters.get("port"),
+                database=parameters.get("database") or "__default__",
+                query=query,
+            )
+        )
+
+    @classmethod
+    def _encryption_from_tls_parameters(
+        cls, sslmode: str | None, secure: str | None
+    ) -> bool:
+        """
+        Resolve whether a connection is encrypted from either TLS spelling.
+
+        ``databend-py`` parsed the legacy ``secure`` value with ``asbool``, so
+        casing is not significant in either parameter.
+        """
+        if sslmode is not None:
+            return sslmode.lower() in cls.encryption_sslmodes
+        if secure is not None:
+            return secure.lower() == "true"
+        return False
 
     @classmethod
     def get_parameters_from_uri(
-        cls, uri: str, *_args: dict[str, Any] | None
+        cls,
+        uri: str,
+        encrypted_extra: dict[str, Any] | None = None,
     ) -> BasicParametersType:
+        """
+        Decompose a Databend URI into individual connection parameters.
+
+        The legacy ``secure`` parameter is still recognised so that connections
+        stored before the move to ``sslmode`` repopulate the form correctly. Its
+        values were parsed as booleans by the previous driver, so casing is not
+        significant in either parameter. Both are always removed, so a URI
+        carrying the legacy and the current spelling at once cannot leak one of
+        them back into the connection as a user-supplied extra parameter.
+        """
         url = make_url_safe(uri)
-        query = url.query
-        if "secure" in query:
-            encryption = url.query.get("secure") == "true"
-            query.pop("secure")
-        else:
-            encryption = False
+        query = dict(url.query)
+        encryption = cls._encryption_from_tls_parameters(
+            query.pop("sslmode", None), query.pop("secure", None)
+        )
         return BasicParametersType(
             username=url.username,
             password=url.password,
--- a/docs/user_docs_versioned_docs/version-6.0.0/configuration/databases.mdx
+++ b/docs/user_docs_versioned_docs/version-6.0.0/configuration/databases.mdx
@@ -423,7 +423,7 @@ echo "sqlalchemy-cratedb" >> ./docker/requirements-local.txt
 #### Databend
 
 The recommended connector library for Databend is [databend-sqlalchemy](https://pypi.org/project/databend-sqlalchemy/).
-Superset has been tested on `databend-sqlalchemy>=0.2.3`.
+Superset has been tested on `databend-sqlalchemy>=0.5.5`.
 
 The recommended connection string is:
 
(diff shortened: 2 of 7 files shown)


rad-pat:
@hantmac - as discussed, here is PR for updating superset with latest dialect.

john-bodley:
@hantmac it seems like you maybe added the the Databend engine spec and thus I was wondering whether you would mind reviewing the PR?

hantmac:
> @hantmac it seems like you maybe added the the Databend engine spec and thus I was wondering whether you would mind reviewing the PR?

Sorry for late response. I will review as soon as possible. Thanks!

hantmac (approved):
Nice work! LGTM!

rad-pat:
@betodealmeida , any chance of this progressing? Do you want me to rebase against master?

hantmac (approved):

superset/db_engine_specs/databend.py
```diff
@@ -273,7 +276,10 @@ def get_parameters_from_uri(
     ) -> BasicParametersType:
         url = make_url_safe(uri)
         query = url.query
-        if "secure" in query:
```
betodealmeida, line 276: Can we add a migration updating existing URLs to the new format? This way upgrading Superset won't break any existing installations.

rad-pat: I was hoping the if...elseif... would cope with URLs of both types, so existing connections using `secure=true/false` would still work anyway

mistercrunch:
Pending changes + @betodealmeida approval ... Switching to draft to get off of our pile temporarily

rusackas:
@rad-pat any intent to follow through on this one? We'll likely close it soon otherwise.

rad-pat:
> @rad-pat any intent to follow through on this one? We'll likely close it soon otherwise.

@rusackas, this is a blast from the past - we apply this change to every release we build so that we connect to Databend ok. There were some comments, and I replied but never got any further response. This works for us as it stands, should not require any database migration to upgrade existing connections. If you're happy, I can rebase on master.

rusackas:
Thanks again @rad-pat - trying to get caught up on PRs again :P

The rebase is still a good idea here, and we'd love to see it through, but the Databend dialect has moved on well past 0.4.6 since. If the connection-param change is still needed against the current driver, the touchups here would be welcomed, or a fresh PR against master might be the easiest path.

rusackas (changes requested):
@rad-pat thanks for sticking with this one. The one thing from @betodealmeida's review that's still open is the migration to update existing connection URLs.

It matters because the parsing here only repopulates the form... the stored `sqlalchemy_uri` (e.g. `?secure=false`) still gets handed to the driver as-is, so on existing installs the `>=0.4.6` bump could break those connections unless we rewrite them.

Shouldn't be much: a migration that loads the `databend` databases and swaps `secure=true`→`sslmode=require` / `secure=false`→`sslmode=disable` in the stored URI. Happy to write it up and push the commit to this branch if you'd rather I take it... just say the word and I'll take a crack at it. Then we'll re-review :)

rusackas:
Turns out I _can't_ actually push to this fork/branch, so I can't help here. Would love to see this through though, one way or another.

superset/db_engine_specs/databend.py
```diff
@@ -217,10 +217,11 @@ class DatabendConnectEngineSpec(BasicParametersMixin, DatabendEngineSpec):
     _function_names: list[str] = []
 
     sqlalchemy_uri_placeholder = (
-        "databend://user:password@host[:port][/dbname][?secure=value&=value...]"
+        "databend://user:password@host[:port][/dbname][?sslmode=value&=value...]"
     )
     parameters_schema = DatabendParametersSchema()
-    encryption_parameters = {"secure": "true"}
+    encryption_parameters = {"sslmode": "require"}
+    encryption_disable_parameters = {"sslmode": "disable"}
```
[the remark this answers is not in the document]
rad-pat: Done — `encryption_disable_parameters` is now on `BasicParametersMixin`, with `build_sqlalchemy_uri` applying it on the `else` branch and `get_parameters_from_uri` stripping it. It defaults to `{}`, so `query.update({})` is a no-op and no other spec changes behaviour.

To answer my own question from last time: there's no need for an inverse boolean. `encryption` still derives from `encryption_parameters` alone — the disable set's only job in the read path is to be *removed*, so it doesn't survive into `query` and reappear as a user-supplied extra parameter. Once I framed it that way the asymmetry I was stuck on went away.

One subtlety worth flagging for whoever reviews it: the stripping can't merge the two dicts. They typically share a key with *different* values (`sslmode=require` / `sslmode=disable`), so `{**enable, **disable}` would keep only one and leak the other. It builds a list of `.items()` pairs from both instead.

`DatabendEngineSpec` still overrides both methods, because it needs the `__default__` database substitution, the legacy `secure` fallback, and a bare `databend://` scheme (the mixin builds `engine+default_driver`, which here would be `databend+databend://` and would change every stored URI). So the mixin change is generic groundwork rather than something Databend routes through — happy to take it back out if you'd rather it landed with its first real consumer.

superset/db_engine_specs/databend.py
rad-pat: You were right and my earlier reply was wrong, on two counts. Migration added.

The `if/elif` I was leaning on only repopulates the **connection form** — it never touches the connection. `DatabendDialect.create_connect_args` does `parameters = dict(url.query)` and concatenates every parameter straight into the DSN handed to the Rust client, so a stored `?secure=false` reaches the driver verbatim.

Worse, that branch could never actually run. `get_parameters_from_uri` was declared `(cls, uri, *_args)` while `Database.parameters` calls it as `get_parameters_from_uri(masked_uri, encrypted_extra=...)` — `TypeError`, swallowed by that property's bare `except` into `parameters = {}`. Even called positionally it then did `url.query.pop(...)` on an `immutabledict`. So there was no runtime fallback at all; both are fixed here with regression tests.

What the driver actually does with the stale parameter (`core/src/client.rs` in `databendlabs/bendsql`):

```rust
let mut scheme = "https";
...
"sslmode" => match v.as_ref() {
    "disable" => scheme = "http",
    "require" | "enable" => scheme = "https",
    _ => return Err(Error::BadArgument(format!("Invalid value for sslmode: {v}"))),
},
...
_ => { session_state.set(k, v); }   // unknown keys are not an error
```

So `secure=false` is absorbed as a session variable and the scheme defaults to `https` — plaintext installs stop connecting, while `secure=true` survives by luck.

One thing that surprised me while writing it: connections with **no** TLS parameter need migrating too. `databend-py` took `secure=False` and set `self.schema = "http"`, and Superset only ever wrote `secure=true` (never `secure=false`), so every connection saved with the encryption box unticked has no parameter at all and was plaintext. Under the new `https` default those break in exactly the way this migration exists to prevent, so `upgrade()` pins them to `sslmode=disable`. That isn't a TLS downgrade — there was never any TLS on those rows.

rad-pat:
Rebased onto master and pushed — both outstanding review points are addressed, and the branch is conflict-free again.

@rusackas — thanks for the nudge, and sorry it sat so long. Two things resolved themselves on the rebase: master is **already** on `databend-sqlalchemy>=0.5.5`, so the `pyproject.toml` bump this PR carried is simply gone, and I confirmed `sslmode` is still the parameter the current driver reads. Master had also merged `DatabendConnectEngineSpec` into `DatabendEngineSpec` since; the changes moved across cleanly.

The migration you asked for is in. Digging into it turned up something I hadn't expected — connections with *no* TLS parameter need migrating too, not just `secure=*` ones. `databend-py` defaulted to an `http` scheme and Superset only ever wrote `secure=true`, so every connection saved with the encryption box unticked has no parameter at all and was plaintext; the Rust core defaults to `https` and breaks them. Details in the thread above.

@betodealmeida — `encryption_disable_parameters` is on `BasicParametersMixin` as you asked, defaulting to `{}` so nothing else changes behaviour, and I've answered my own question from that thread (short version: no inverse boolean is needed — the disable set only ever needs *stripping* on the read path).

I also fixed some pre-existing faults in the two methods this PR touches, which is the bulk of the extra diff. `Database.parameters` was returning `{}` for every Databend connection — `get_parameters_from_uri` took `*_args` and so raised `TypeError` on the keyword call, and it mutated an `immutabledict`; both were swallowed by that property's bare `except`. Net effect: you could create a Databend connection through the dynamic form but reopening it showed an empty form, and the legacy-`secure` fallback was unreachable. Also `sslmode=enable` read as unencrypted (the driver accepts it alongside `require`), which meant a save could silently rewrite a TLS connection to `sslmode=disable`. Happy to split those into a separate PR if you'd prefer them reviewed apart from the `secure`→`sslmode` change — say the word.

@hantmac — you approved this a while back; the scope has grown since, so it's worth another look from a Databend angle if you have time, particularly the migration's treatment of parameter-less URIs.

Tests: unit coverage added for the build/parse round trip, the legacy fallback, the `encrypted_extra` keyword regression, the mixin's opt-out parameters, and the migration's upgrade/downgrade. There were none for these paths before, which is how all of the above survived.

rusackas (approved):
Thanks for the deep dive, @rad-pat — this is exactly the fix, migration and test coverage the earlier review was asking for. Verified the sslmode/secure round-trip logic and the migration's in-place query rewrite, all green. Merging.

PR #28628, opened 2024-05-21
chore: Set isolation level to READ COMMITTED for testing et al. 

john-bodley:
### SUMMARY

The default isolation level for PostgreSQL is `READ COMMITTED` whereas for MySQL it is `REPEATABLE READ`. Ideally we would strive to have these the same (for consistency reasons) and the former is likely preferred given the async nature of Superset (especially SQL Lab) where the backend is polling to determine whether a query has been cancelled from another Flask-SQLAlchemy session.

This PR:

1. Adds a comment in `config.py` to try to persuade institutions to set the isolation level. It's actually hard/difficult to impose this given that the default config is configured for SQLite which doesn't support `READ COMMITTED`.
2. Updates the test configs to ensure that—for MySQL and PostgreSQL—the isolation level is `READ COMMITTED`. 

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -196,6 +196,13 @@ SQLALCHEMY_DATABASE_URI = (
 # SQLALCHEMY_DATABASE_URI = 'mysql://myapp@localhost/myapp'
 # SQLALCHEMY_DATABASE_URI = 'postgresql://root:password@localhost/myapp'
 
+# The default MySQL isolation level is REPEATABLE READ whereas the default PostgreSQL
+# isolation level is READ COMMITTED. All backends should use READ COMMITTED (or similar)
+# to help ensure consistent behavior.
+SQLALCHEMY_ENGINE_OPTIONS = {
+    "isolation_level": "SERIALIZABLE",  # SQLite does not support READ COMMITTED.
+}
+
 # In order to hook up a custom password store for all SQLALCHEMY connections
 # implement a function that takes a single argument of type 'sqla.engine.url',
 # returns a password and set SQLALCHEMY_CUSTOM_PASSWORD_STORE.
--- a/tests/integration_tests/superset_test_config.py
+++ b/tests/integration_tests/superset_test_config.py
@@ -20,6 +20,8 @@ import math
 from copy import copy
 from datetime import timedelta
 
+from sqlalchemy.engine import make_url
+
 from superset.config import *  # noqa: F403
 from superset.config import DATA_DIR
 from tests.integration_tests.superset_test_custom_template_processors import (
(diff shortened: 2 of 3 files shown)


tests/integration_tests/superset_test_config.py
```diff
@@ -52,12 +54,15 @@
 if "UPLOAD_FOLDER" in os.environ:  # noqa: F405
     UPLOAD_FOLDER = os.environ["UPLOAD_FOLDER"]  # noqa: F405
 
-if "sqlite" in SQLALCHEMY_DATABASE_URI:
+if make_url(SQLALCHEMY_DATABASE_URI).get_backend_name() == "sqlite":
```
john-bodley, line 57: Safer than doing a string match.

michael-s-molina:
> Set isolation level to READ COMMITTED 

@john-bodley looking at the changed files I was not able to see where you set the isolation level to READ COMMITTED in this PR. Did you forget to commit something?

john-bodley:
@michael-s-molina I updated the PR title. It's actually hard to set this globally given that not all of our backends support the `READ COMMITTED`. I did add a comment in the `config.py` and updated the test configs so both MySQL and PostgreSQL will have similar behavior.

michael-s-molina (approved):

superset/config.py
```diff
@@ -196,6 +196,13 @@ def _try_json_readsha(filepath: str, length: int) -> str | None:
 # SQLALCHEMY_DATABASE_URI = 'mysql://myapp@localhost/myapp'
 # SQLALCHEMY_DATABASE_URI = 'postgresql://root:password@localhost/myapp'
 
+# The default MySQL isolation level is REPEATABLE READ whereas the default PostgreSQL
+# isolation level is READ COMMITTED. All backends should use READ COMMITTED (or similar)
+# to help ensure consistent behavior.
+SQLALCHEMY_ENGINE_OPTIONS = {
+    "isolation_level": "SERIALIZABLE",  # SQLite does not support READ COMMITTED.
```
sadpandajoe, line 203: @john-bodley Reading this it looks like everything will be `SERIALIZABLE` right? we're overriding it in the tests but are we overriding it as `READ COMMITTED` anywhere else for the other databases?

mistercrunch: @sadpandajoe followed the rabbit hole here, and I think you're right, this will bump all envs who don't set their SQLALCHEMY_ENGINE_OPTIONS to a higher isolation level.

mistercrunch: How about this draft? https://github.com/apache/superset/pull/30174

PR #28639, opened 2024-05-22
fix: adds the ability to disallow SQL functions per engine

dpgaspar:
### SUMMARY
Adds a new configuration key named `DISALLOWED_SQL_FUNCTIONS` that defines disallowed function per engine on SQL statements. These functions will be disallowed on SQLLab and Charts.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/db_engine_specs/trino.py
```diff
@@ -219,19 +219,22 @@ def execute_with_cursor(
...
-        def _execute(results: dict[str, Any], event: threading.Event) -> None:
+        def _execute(
+            results: dict[str, Any], event: threading.Event, app: Flask
+        ) -> None:
             logger.debug("Query %d: Running query: %s", query_id, sql)
 
             try:
-                cls.execute(cursor, sql, query.database)
+                with app.app_context():
+                    cls.execute(cursor, sql, query.database)
```
dpgaspar, line 229: threads don't have an app context: https://stackoverflow.com/questions/72541670/why-flask-app-context-is-lost-in-child-thread-when-application-factory-pattern-i

betodealmeida: Oh, interesting.

superset/sql_parse.py
```diff
@@ -223,6 +224,19 @@ def get_cte_remainder_query(sql: str) -> tuple[str | None, str]:
...
+    sql: str, function_list: set[str], engine: str | None = None
+) -> bool:
+    """
+    Check if the SQL statement contains any of the specified functions.
+
+    :param sql: The SQL statement
+    :param function_list: The list of functions to search for
+    :param engine: The engine to use for parsing the SQL statement
+    """
+    return ParsedQuery(sql, engine=engine).check_functions_exist(function_list)
```
betodealmeida, line 237: We (probably me) will have to convert this to use `sqlglot` and the `SQLStatement` class (https://github.com/apache/superset/issues/26786) but I'm happy to do it, seems simple enough.

betodealmeida (approved):

superset/config.py
```diff
@@ -1225,6 +1225,15 @@ def CSV_TO_HIVE_UPLOAD_DIRECTORY_FUNC(  # pylint: disable=invalid-name
 #
 DB_SQLA_URI_VALIDATOR: Callable[[URL], None] | None = None
 
+# A set of disallowed SQL functions per engine. This is used to restrict the use of
+# unsafe SQL functions in SQL Lab and Charts. The keys of the dictionary are the engine
+# names, and the values are sets of disallowed functions.
+DISALLOWED_SQL_FUNCTIONS: dict[str, set[str]] = {
```
john-bodley, line 1231: I'm unsure where the best place for this deny list, i.e., here in the configuration or within the extra JSON payload of the database. 

Additionally should this be engine (dialect) specific or database specific? If it's the later then maybe the extra JSON payload field is preferable.

dpgaspar: JSON payload at the database level is more dynamic and would avoid having to change the config to add remove disallowed functions. But on the other hand the user that actually registers the db could have intentions to "abuse" these functions.

Qadeem411 (approved):

PR #28652, opened 2024-05-23
feat: Adds the ECharts Histogram chart

### LINKED ISSUE #23066: Histogram Chart does not apply groupby /dimension grouping to the generated query
#### How to reproduce the bug

1. Go to charts
2. select any dataset and click on histogram
3. provide a dimention and also  a metric
4. See the query generated

### Expected results
a histogram of the grouped data should be seen

### Actual results
"Aggregate functions can't co-exist without GROUP BY clause\n"
query does not have a group by clause in it
### Environment
master branch

- browser type and version:  chrome ,Version 109.0.5414.120 (Official Build) (64-bit)
- superset version: master branch
- python version: super set python version (unchanged)
- node.js version: `node -v`  super set node version (unchanged)
- any feature flags active:

### Checklist

### LINKED ISSUE #24064: Histogram bars aren't always stacking on dimension
I have a dataset like:

```
 category_a | category_b | value
------------+------------+--------
 x          |  a         |  6.988
 x          |  b         |  6.994
 y          |  b         |  6.995
 z          |  b         |  6.996
 z          |  a         |  6.997
 z          |  c         |  7.001
...
```

I generated a Histogram Chart of `value` using `category_b` as a Dimension.  Here is the result, where the categories (a, b, c) are represented in 3 different colors:

image

The problem is, bars are sometimes hidden/overlapping, not stacked.  This can be detected when mousing over some of the bars where data appears to be missing.  For example:

image

The opacity changes on mouseover, to reveal hidden bars behind the first one.  It happens for at least 3 boxes in this example, not just the final one in the screenshot.

#### How to reproduce the bug

The process was described above.  I can provide a CSV of my actual data if it helps.

### Expected results

No bars are hidden, they are all stacked vertically when using Dimensions.

### Actual results

Some bars are randomly hidden.

### Environment

- browser type and version: tested in Firefox 113 and Chrome 113
- superset version: 2.1.0
- python version: (docker)
- node.js version: (docker)
- any feature flags active:

### Checklist

michael-s-molina:
### SUMMARY
This PR creates the ECharts Histogram chart which is used to display the distribution of a dataset by representing the frequency or count of values within different ranges or bins. It helps visualize patterns, clusters, and outliers in the data and provides insights into its shape, central tendency, and spread.

This plugin uses [numpy.histogram](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html#numpy.histogram) to compute the distribution on the server side. It supports a single numeric column and multiple dimensions. Aggregations are not supported.

The legacy Histogram is one of the plugins that are currently preventing us from [upgrading to React 17](https://github.com/orgs/apache/projects/345?pane=issue&itemId=59020521). With this PR, we should be able to create a migration script for legacy charts and remove another blocker for the React upgrade. 

I added the minimum required controls to allow the migration but many more controls can be added in the future to improve the chart's customizability.

Fixes https://github.com/apache/superset/issues/24064
Fixes https://github.com/apache/superset/issues/23066

Screenshot 2024-06-03 at 13 15 40
Screenshot 2024-06-03 at 15 54 54
Screenshot 2024-06-03 at 15 55 33

### TESTING INSTRUCTIONS
Check all features.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/explore/components/ExploreViewContainer/index.jsx
```diff
@@ -460,15 +460,21 @@ function ExploreViewContainer(props) {
...
-        key =>
-          typeof lastQueriedControls[key] !== 'undefined' &&
-          !areObjectsEqual(
-            props.controls[key].value,
-            lastQueriedControls[key].value,
-            { ignoreFields: ['datasourceWarning'] },
-          ),
-      );
+      const { controls } = props;
+      const changedControlKeys = Object.keys(controls).filter(key => {
```
michael-s-molina, line 464: This change fixes a problem with the algorithm that detect a stale chart. Previously, if the values being compared were not objects, we had wrong results. For example, if a user changed the row limit, which is a number, the banner indicating a staled chart was not shown. This should significantly improve user experience when changing control values.

mistercrunch:
200w

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/controlPanel.tsx
```diff
@@ -0,0 +1,131 @@
...
+            },
+          },
+        ],
+        [
+          {
+            name: 'normalize',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Normalize'),
+              description: t('Whether to normalize the histogram'),
```
kgabryje, line 77: I wonder if such descriptions are valuable - it feels a bit repetitive with the label. Maybe we should either skip the description, or (preferably) describe how normalization works, or provide a link to a resource that explains normalization?

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/controlPanel.tsx
```diff
@@ -0,0 +1,131 @@
...
+            },
+          },
+        ],
+        [
+          {
+            name: 'cumulative',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Cumulative'),
+              description: t('Whether to make the histogram cumulative'),
```
kgabryje, line 88: Same comment as above

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/controlPanel.tsx
```diff
@@ -0,0 +1,131 @@
...
+        [showLegendControl],
+        [
+          {
+            name: 'x_axis_title',
+            config: {
+              type: 'TextControl',
+              label: t('X Axis Title'),
+              renderTrigger: true,
+              default: '',
+              description: t('Changing this control takes effect instantly'),
```
kgabryje, line 110: I think that tooltip is already displayed when hovering the `renderTrigger: true` icon?

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/controlPanel.tsx
```diff
@@ -0,0 +1,131 @@
...
+        ],
+        [
+          {
+            name: 'y_axis_title',
+            config: {
+              type: 'TextControl',
+              label: t('Y Axis Title'),
+              renderTrigger: true,
+              default: '',
+              description: t('Changing this control takes effect instantly'),
```
kgabryje, line 122: here too

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/transformProps.ts
```diff
@@ -0,0 +1,203 @@
...
+    normalize,
+    showLegend,
+    showValue,
+    sliceId,
+    xAxisTitle,
+    yAxisTitle,
+  } = formData;
+  const { data } = queriesData[0];
+  const colorFn = CategoricalColorNamespace.getScale(colorScheme);
+  const formatter = getNumberFormatter(
```
kgabryje, line 66: Should we also support currency formatter here?

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/transformProps.ts
```diff
@@ -0,0 +1,203 @@
...
+    normalize ? NumberFormats.FLOAT_3_POINT : NumberFormats.INTEGER,
+  );
+  const percentFormatter = getPercentFormatter(NumberFormats.PERCENT_2_POINT);
+  const groupbySet = new Set(groupby);
+  const xAxisData: string[] = Object.keys(data[0]).filter(
+    key => groupbySet.has(key) === false,
```
kgabryje, line 72:
```suggestion
    key => !groupbySet.has(key),
```

superset-frontend/packages/superset-ui-chart-controls/src/utils/columnChoices.ts
```diff
@@ -16,24 +16,29 @@
...
   datasource?: Dataset | QueryResponse | null,
+  type?: GenericDataType,
 ): [string, string][] {
   if (isDataset(datasource) || isQueryResponse(datasource)) {
-    return datasource.columns
-      .map((col): [string, string] => [
+    const { columns } = datasource;
+    const filteredColumns = (
+      columns as { type_generic?: GenericDataType }[]
+    ).filter(col => (type !== undefined ? col.type_generic === type : true));
```
kgabryje, line 33:
```suggestion
    ).filter(col => (type === undefined || col.type_generic === type));
```

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/controlPanel.tsx
michael-s-molina: That's a good point. I'll see if I can generate some helpful tooltips using GPT.

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/transformProps.ts
michael-s-molina: Let me give you some context to answer this question. Now that we're migrating legacy plugins, it becomes clear that we have a lot of duplicated code and many inconsistencies between ECharts plugins regarding the controls that are available. I'm planning to tackle this horizontally, making sure that features, such as currency formatting, are available consistently among the plugins. For that reason, for this PR:

> I added the minimum required controls to allow the migration but many more controls can be added in the future to improve the chart's customizability.

kgabryje: Sorry, missed that part of the pr description. Makes sense!

michael-s-molina: No problem! You correctly spotted the missing control.

michael-s-molina:
Thanks for the review @kgabryje! I think I addressed all your comments, including adding helpful tooltips. Thanks for pointing that out! It was a copy/paste from the legacy chart.

kgabryje:
@michael-s-molina It looks like normalised histogram doesn't reach 1 for tallest bar - 1st screenshot is the new plugin, 2nd is the old one

image

image

kgabryje:
1 more comment - why do we have a select component for num column instead of drag&drop? We could limit the accepted column by type in dnd

michael-s-molina:
> @michael-s-molina It looks like normalised histogram doesn't reach 1 for tallest bar - 1st screenshot is the new plugin, 2nd is the old one

@kgabryje I changed the normalize calculation and updated the tooltip. It should be correct now. Keep in mind that it will not match the legacy version because that was wrong too 😅 .

kgabryje (approved):
LGTM, thanks for creating a proper histogram plugin!

michael-s-molina:
> 1 more comment - why do we have a select component for num column instead of drag&drop? We could limit the accepted column by type in dnd

@kgabryje I changed the control to use drag-and-drop as you suggested.

PR #28703, opened 2024-05-24
chore: remove ipython from development dependencies

mistercrunch:
# Summary

removing ipython as it's simply a nice-to-have, but not required for Superset development

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -178,7 +178,6 @@ development = [
     "freezegun",
     "greenlet>=2.0.2",
     "grpcio>=1.55.3",
-    "ipython",
     "openapi-spec-validator",
     "parameterized",
     "pip-compile-multi",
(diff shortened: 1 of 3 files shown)


betodealmeida (approved):

mistercrunch:
Yes for this particular one the "likely hood to be used" vs "dependency burden" isn't great. Currently the criteria is more or less "commonly used by humans and/or CI for development workflows".

Now that we have a workflow to upgrade things (supersetbot as a pip-compile-multi-aware dependabot replacement), it seems reasonable / manageable if we wanted to bring back more segmentation/complexity under `requirements/` if we wanted to.

mistercrunch:
One more thought it with segmentation in `requirements/` we can have different level of rigor for freshness for different files. Likely I'll set up supersetbot in different ways for `base.in` VS `development.in`

PR #17543, opened 2021-11-24
feat: new dataset/table/column models

betodealmeida:
### SUMMARY

This PR implements new models for SIP-68 (https://github.com/apache/superset/issues/14909):

- `Column`
- `Table`
- `Dataset`

The associated tables are created with a migration, and the existing datasets (`SqlaTable`, `SqlMetric`, `TableColumn`) are migrated to the new models:

- A physical dataset is mapped into a `Table` instance and a `Dataset` instance.
- A virtual dataset is mapped into a `Dataset` instance.
- Metrics and columns are mapped into `Column` instances (in the new model it represents physical columns, derived columns, and metrics).

The models are kept up-to-date via hooks added to `SqlaTable`, `SqlMetric`, and `TableColumn`. Every time these models are created, deleted, or updated, the new models are updated correspondingly. The sync is unidirectional, from the old models to the new ones.

The next step is to modify the backend to **read** from the new models. Initially the API will remain unmodified, but hopefully in the future we might be able to clean it up and modernize it (in some places we haven't migrated to the v1 API yet).

After that, the backend will be modified to **write** to the new models. Once the backend is reading and writing from the new models we can get rid of the old models (including the Druid models).

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

No visual changes.

### TESTING INSTRUCTIONS

Reviewers can run `superset load-examples` and see the new models being populated:

```sql
mysql> SELECT `schema`, name FROM sl_tables;
+----------+------------------------+
| schema   | name                   |
+----------+------------------------+
| examples | wb_health_population   |
| examples | birth_names            |
| examples | long_lat               |
| examples | birth_france_by_region |
| examples | sf_population_polygons |
| examples | flights                |
| examples | bart_lines             |
| examples | channel_members        |
| examples | messages               |
| examples | channels               |
| examples | threads                |
| examples | unicode_test           |
| examples | users                  |
| examples | users_channels         |
| examples | cleaned_sales_data     |
+----------+------------------------+
15 rows in set (0.00 sec)

mysql> SELECT name, is_physical FROM sl_datasets;
+---------------------------+-------------+
| name                      | is_physical |
+---------------------------+-------------+
| wb_health_population      |           1 |
| birth_names               |           1 |
| long_lat                  |           1 |
| birth_france_by_region    |           1 |
| sf_population_polygons    |           1 |
| flights                   |           1 |
| bart_lines                |           1 |
| FCC 2018 Survey           |           0 |
| messages_channels         |           0 |
[...]

Issue #14909, opened 2021-05-28
[WIP] A better model for Datasets


betodealmeida:
## [SIP] A better model for Datasets

This document proposes the creation of new set of SQLAlchemy models to better represent **datasets** in Superset and build the foundation for a powerful semantic layer:

- A `Column` model, for table columns, dataset metrics, and dataset derived columns;
- A `Table` model, representing physical tables or VIEWs in a given database;
- A `Dataset` model, representing the current concept of _physical_ and _virtual_ datasets.

These would replace the following existing 9 models:

- `BaseDatasource`, `DruidDatasource`, and `SqlaTable`. These models currently are used to represent Druid datasources (to be deprecated), as well as physical and virtual datasets built on SQLAlchemy.
- `BaseColumn`, `DruidColumn`, `TableColumn`, representing columns in Druid or SQLAlchemy.
- `BaseMetric`, `DruidMetric`, `SqlMetric`, representing metrics in Druid or SQLAlchemy.

(Note that Superset currently doesn't have a distinction between a table in a database and a dataset enriched with metadata. Because of this, it's hard to infer relationships between objects such as a virtual dataset and the tables it references, even though Superset is already able to extract that relationship from the SQL.)

There's an additional 10th model that is out-of-scope for this SIP: `AnnotationDatasource`. Eventually it would be nice to allow any dataset to be used as an annotation, removing the need for a custom model.

### Motivation

The current semantic model in Superset is very simple. Originally, Superset was built as a UI to explore Druid *datasources* (tables). At the time Druid didn't support SQL, and queries had to be done using a native JSON-based interface. Because of the lack of SQL support, when adding a datasource to Superset the column types had to be inferred from their names, and users could manually override them. Also, because of the nature of Druid's storage, users also had to indicate a temporal column, which columns were filterable, and which ones were groupable, in order to prevent expensive queries from running.

Users could also add new metrics and derived columns to a datasource, which was an important feature because they couldn't simply be defined on-the-fly through SQL. The metrics, columns, and metadata describing columns were (and still are) stored in a model called `DruidDatasource`. **This description of the the underlying table and the additional metadata became Superset's semantic layer**.

[...]

PR #28704, opened 2024-05-24
chore: remove sl_ tables

mistercrunch:
### SUMMARY

I don't know the full history but this collection of models ('sl_.*') were created to support semantic layer, but the work to build on top of it never really started. This PR removes those models and the unfinished things build on top of it.

I dug out this related SIP https://github.com/apache/superset/pull/17543

 It includes:
- sl_dataset_columns
- sl_table_columns
- sl_dataset_tables
- sl_columns
- sl_tables
- sl_dataset_users
- sl_datasets


john-bodley:
Adding @betodealmeida as a reviewer as he was the original author of said logic.

betodealmeida (approved):
> I don't know the full history but this collection of models ('sl_.*') were created to support semantic layer, but the work to build on top of it never really started. This PR removes those models and the unfinished things build on top of it.

For context, SIP-68 (https://github.com/apache/superset/issues/14909) proposed to clean up out existing semantic layer, making relationships explicit and differentiating physical datasets from underlying tables.

Since changing the existing models was too big of an effort, the initial code created the new models, and performed a double-write to keep them in sync with the old models. The idea was to update the code to read from the new models, and then drop the double-write and the old models, but this never happened because still the effort was too much.

mistercrunch:
Could always revert or find inspiration here. But feels good to simplify the erd in the meantime

PR #28745, opened 2024-05-28
fix(reports): Update the element class to wait for when taking a screenshot

Vitor-Avila:
### SUMMARY
In the process of taking screenshots for an alert/report, Selenium/Playwright is configured to wait a certain time (controlled via `superset_config.py`) until all chart elements are drawn. The logic was configured to monitor the `slice_container` class, however with https://github.com/apache/superset/pull/27255 this class is only used once the chart data is retrieved.

This PR changes the class to be monitored to the `chart-container` wrapper.

No UI changes.

### TESTING INSTRUCTIONS
1. Configure `SCREENSHOT_LOCATE_WAIT` to be 30 seconds, and `SCREENSHOT_LOAD_WAIT` to be 200 seconds.
2. Create a chart that takes longer than 30 seconds to load. You can use `pg_sleep()` with Postgres for example. 
3. Configure a report with this chart. 

Before this PR the report would fail, as the chart wouldn't load within 30 seconds.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/json.py
+++ b/superset/utils/json.py
@@ -211,7 +211,7 @@ def dumps(  # pylint: disable=too-many-arguments
             cls=cls,
         )
     except UnicodeDecodeError:
-        results_string = simplejson.dumps(  # type: ignore[call-overload]
+        results_string = simplejson.dumps(
             obj,
             default=default,
             allow_nan=allow_nan,
--- a/superset/utils/json.py
+++ b/superset/utils/json.py
@@ -211,7 +211,7 @@ def dumps(  # pylint: disable=too-many-arguments
             cls=cls,
         )
     except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


superset/utils/json.py
```diff
@@ -211,7 +211,7 @@ def dumps(  # pylint: disable=too-many-arguments
             cls=cls,
         )
     except UnicodeDecodeError:
-        results_string = simplejson.dumps(  # type: ignore[call-overload]
+        results_string = simplejson.dumps(
```
Vitor-Avila, line 214: Had to remove the `ignore` to make `mypy` happy.

kgabryje (approved):
LGTM

PR #28746, opened 2024-05-28
fix(Dashboard row): Reveal resize handle underneath droppable component

rtexelm:
### SUMMARY

This edits the class added to the last droppable. Previously the absolute positioning obscured the resize handle so the user couldn't decrease width once it was at max. I've left in the conditional in case specific styling is needed in the future.

B:
ChartResize-ezgif com-optimize (1)
old-resize

A:
row_resizable-ezgif com-optimize
new-resize

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


justinpark:
@rtexelm I created a new PR that fixes the issue while retaining the display of the last drop zone's position.
(fyi - The solution ensures that the position only appears during dragging, which resolves the problem.)

PR #25017, opened 2023-08-17
fix: Downgrades Prophet to 1.1.1 and Holidays to 0.23

michael-s-molina:
### SUMMARY
https://github.com/apache/superset/pull/24129 bumped Prophet to 1.1.4 but there's a [nasty bug](https://github.com/facebook/prophet/issues/2354) affecting all versions following 1.1.1. This PR downgrades Prophet to 1.1.1 and pins the version in `setup.py`. We can relax the version constraint once that bug is fixed.

This PR also reverts the [holidays bump](https://github.com/apache/superset/pull/24647) given that version 0.28 is incompatible with `prophet` 1.1.1 even with a declared dependency of `holidays>=0.14.2`. I tried using version 0.28 but it broke CI.

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #28747, opened 2024-05-28
chore: document upper bound for python lib 'holidays' >= 0.26

mistercrunch:
Change of plans, adding a note documenting the upper bound

~~I simply removed the upper bound for the holidays lib and ran `pip-compile-multi -P holidays`. Note that `pip-compile-multi` isn't as' deterministic as we'd like for it to be, but things should be fine as "# vias" are just informational here.~~

~~For the record: https://github.com/vacanza/python-holidays/blob/dev/CHANGES~~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ dependencies = [
     "geopy",
     "gunicorn>=22.0.0; sys_platform != 'win32'",
     "hashids>=1.3.1, <2",
+    # known issue with holidays 0.26.0 and above related to prophet lib #25017
     "holidays>=0.25, <0.26",
     "humanize",
     "importlib_metadata",
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -80,7 +80,7 @@ cron-descriptor==1.4.3
     # via apache-superset
 croniter==2.0.5
     # via apache-superset
-cryptography==42.0.7
+cryptography==42.0.8
     # via
     #   apache-superset
     #   paramiko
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -57,7 +57,7 @@ dependencies = [
     "geopy",
     "gunicorn>=22.0.0; sys_platform != 'win32'",
     "hashids>=1.3.1, <2",
-    "holidays>=0.25, <0.26",
+    "holidays>=0.25",
```
john-bodley, line 60: Should we place an upper bound of `1.0.0` as well?

john-bodley (commented):
@michael-s-molina or @villebro do either of you know why we previously had an upper bound of `0.26` here?

john-bodley:
@mistercrunch it would be good to determine if the issue in https://github.com/apache/superset/pull/25017 has been resolved.

mistercrunch:
@john-bodley great catch! I totally would have missed that. I think we really need to add comments for all real upper bounds in pyroject.toml, including if it's just "based on semantic versioning and/or stress levels around this specific dependency".

I'm planning on implementing something like `supersetbot bump-python --disregard-upper-bound` which would do as specified and remove the upper bound in the process. We may need more parsable hints as smart comments so that @supersetbot knows what to do with it say:

```
"holidays>=0.25, <0.26", #known-upper-bound-issue #couple-with-lib=prophet
```

or something like
```
"holidays>=0.25, <1", #semver-upper-bound 
```

mistercrunch:
@rusackas closing/cleaning some old PRs if you don't mind stamping this

sadpandajoe (approved):

PR #28749, opened 2024-05-29
fix(sqllab): Sort db selector options by the API order

justinpark:
### SUMMARY
Currently, the order of the db selector dropdown in SQL Lab is always sorted by record ID, regardless of the intended sort order.
This issue arises due to two reasons: 
1. an incorrect order_column name is being passed
2. AsyncSelect component's [default sortComparator](https://github.com/apache/superset/blob/07b2449bd777cf04a266f1de3d72af6874b9c305/superset-frontend/src/components/Select/AsyncSelect.tsx#L138) is sorting by ID (Since label is not a string, the next ordering will be by id).

This commit fixes these issues, ensuring that the db selector is sorted by the API order as intended.

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
+++ b/superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
@@ -154,6 +154,12 @@ const fakeDatabaseApiResult = {
   ],
 };
 
+const fakeDatabaseApiResultInReverseOrder = {
+  ...fakeDatabaseApiResult,
+  ids: [2, 1],
+  result: [...fakeDatabaseApiResult.result].reverse(),
+};
+
 const fakeSchemaApiResult = {
   count: 2,
   result: ['information_schema', 'public'],
@@ -168,7 +174,8 @@ const fakeFunctionNamesApiResult = {
   function_names: [],
 };
 
-const databaseApiRoute = 'glob:*/api/v1/database/?*';
+const databaseApiRoute =
+  'glob:*/api/v1/database/?*order_column:database_name,order_direction*';
 const catalogApiRoute = 'glob:*/api/v1/database/*/catalogs/?*';
 const schemaApiRoute = 'glob:*/api/v1/database/*/schemas/?*';
 const tablesApiRoute = 'glob:*/api/v1/database/*/tables/*';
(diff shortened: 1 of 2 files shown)


michael-s-molina (changes requested):
@justinpark if you look at the default [DEFAULT_SORT_COMPARATOR](https://github.com/apache/superset/blob/07b2449bd777cf04a266f1de3d72af6874b9c305/superset-frontend/src/components/Select/constants.ts#L30C14-L30C37) implementation, you can see that it will sort items alphabetically if `label` is of type `string`. In this case, it's not sorting alphabetically because `label` is being set to a React component [here](https://github.com/apache/superset/blob/07b2449bd777cf04a266f1de3d72af6874b9c305/superset-frontend/src/components/DatabaseSelector/index.tsx#L195) instead of using `customLabel` for that purpose. If you set `label` to `database_name` and `customLabel` to the React component, you'll get the desired behavior.

justinpark:
> justinpark if you look at the default [DEFAULT_SORT_COMPARATOR](https://github.com/apache/superset/blob/07b2449bd777cf04a266f1de3d72af6874b9c305/superset-frontend/src/components/Select/constants.ts#L30C14-L30C37) implementation, you can see that it will sort items alphabetically if `label` is of type `string`. In this case, it's not sorting alphabetically because `label` is being set to a React component [here](https://github.com/apache/superset/blob/07b2449bd777cf04a266f1de3d72af6874b9c305/superset-frontend/src/components/DatabaseSelector/index.tsx#L195) instead of using `customLabel` for that purpose. If you set `label` to `database_name` and `customLabel` to the React component, you'll get the desired behavior.

You are right, but I think that if the API that includes the actual sorting items is implemented, we would want the results to be sorted according to the API. Also, relying on client-side sorting can lead to inconsistencies, especially when handling paginated results, so I believe it is better to follow the API's results.

Additionally, I have optimized the code to use the suggested order field.

michael-s-molina (approved):

michael-s-molina:
@justinpark Could you also update the PR description to say that the objective is to sort by the API order?

PR #28755, opened 2024-05-29
fix: Workaround for Pandas.DataFrame.to_csv bug

john-bodley:
### SUMMARY

Per https://github.com/pandas-dev/pandas/issues/47871, in Python 3.10+ (which is now the minimal supported Python version in Superset) there's a Pandas bug where `DataFrame.to_csv` unnecessarily requires one to specify an `escapechar` even though—per [here](https://github.com/apache/superset/blob/6575cacc5d1cc59c7cacd9e186a1d05e76259183/superset/utils/csv.py#L42-L64)—special values have already been escaped.

### TESTING INSTRUCTIONS

Updated unit tests.

Also tested locally, i.e., I was able to successfully export a CSV file containing escaped special characters in Python 3.9 yet it throws in Python 3.10 with the following error, 

```
_csv.Error: need to escape, but no escapechar set
```

unless an escape character is specified.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/csv.py
+++ b/superset/utils/csv.py
@@ -78,7 +78,7 @@ def df_to_escaped_csv(df: pd.DataFrame, **kwargs: Any) -> Any:
                 if isinstance(value, str):
                     df.at[idx, name] = escape_value(value)
 
-    return df.to_csv(**kwargs)
+    return df.to_csv(escapechar="\\", **kwargs)
 
 
 def get_chart_csv_data(
--- a/tests/unit_tests/utils/csv_tests.py
+++ b/tests/unit_tests/utils/csv_tests.py
@@ -14,7 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-import io
 
 import pandas as pd
 import pyarrow as pa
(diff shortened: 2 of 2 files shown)


superset/utils/csv.py
```diff
@@ -49,10 +49,6 @@ def escape_value(value: str) -> str:
     is_negative_number = negative_number_re.match(value) is not None
 
     if needs_escaping and not is_negative_number:
-        # Escape pipe to be extra safe as this
```
john-bodley, line 52: This is now handled via the `escapechar` character in the `pandas.DataFrame.to_csv` method.

tests/integration_tests/utils/csv_tests.py
```diff
@@ -59,24 +58,39 @@ def test_escape_value():
...
     assert escaped_csv_rows == [
-        ["col_a", "'=func()"],
-        ["-10", r"'=cmd\|' /C calc'!A0"],
-        ["a", "'=b"],  # pandas seems to be removing the leading ""
-        ["' =a", "b"],
+        ["a"],
+        ["col_a"],
+        ["'=func()"],
+        ["-10"],
+        [r"'=cmd\\|' /C calc'!A0"],
```
john-bodley, line 90: There's now an extra `\`. This is due to how the Pandas `DataFrame` is composed as opposed to to inclusion of an escape character.

michael-s-molina (approved):

michael-s-molina (approved):

PR #26654, opened 2024-01-17
chore: add unique constraint to tagged_objects

mistercrunch:
### SUMMARY

I ran into this issue where I had duplicates in tagged_object that prevented the deletion of a dashboard. This PR adds a unique constraint to that table, preventing this issue from happening again. This may create another error on update or insert of a dashboard object, hoping a unit test would catch that if there's such a raise condition.

The error I observed was:
```
sqlalchemy.orm.exc.StaleDataError: DELETE statement on table 'tagged_object' expected to delete 2 row(s); Only 3 were matched.
```

Looking into my database and researching the issue, I found dups in my tagged_object table:
```
mysql> select * from tagged_object where object_type ='dashboard' and object_id=11 LIMIT 10;
+---------------------+---------------------+-----+--------+-----------+-------------+---------------+---------------+
| created_on          | changed_on          | id  | tag_id | object_id | object_type | created_by_fk | changed_by_fk |
+---------------------+---------------------+-----+--------+-----------+-------------+---------------+---------------+
| 2024-01-16 15:21:13 | 2024-01-16 15:21:13 | 133 |      3 |        11 | dashboard   |             1 |             1 |
| 2024-01-16 16:19:29 | 2024-01-16 16:19:29 | 180 |      4 |        11 | dashboard   |             1 |             1 |
| 2024-01-16 16:19:29 | 2024-01-16 16:19:29 | 181 |      4 |        11 | dashboard   |             1 |             1 |
+---------------------+---------------------+-----+--------+-----------+-------------+---------------+---------------+
```

Handling it on the cascade delete seemed harder than preventing it, and prevention is also a great thing. Note that this includes a migration that does delete eventual dups too.

# phase 2 - fixing the root cause
After setting up the unique constraint, I hit the root cause, something around auto-owner management where we maintain special owner tags for objects. The approach was to delete-and-recreate them all systematically, and I think the logic + transaction handling was the root cause for the dups. I changed the logic there to go id-by-id and remove and only remove or add individual ids if needed.

I also cleaned up the API to upsert tags it gets instead of raising issue. So if you call the API now to add a tag that's already set, it won't return an error, just add if needed. 

# reuse code in alembic migration
I cracked a solution to share a module across alembic migrations. Somehow that's way more tricky than it seems, I failed at least a few times before at finding a solution for this. We now have `superset/migrations/migration_utils.py` for that purpose

PR #28769, opened 2024-05-29
fix: address tag-related deletion issues raised in #26654

mistercrunch:
I wasn't able to reproduce the issues in #26654, but wanted to submit this diff here as something to try. I'm not likely to push this through but thought a PR would be a good way to suggest a fix and open centralize the discussion


mistercrunch:
Ok, was told this didn't fix the issue... closing.

It'd be great to have a way to reproduce this...

PR #28770, opened 2024-05-29
feat: impersonate with email prefix

betodealmeida:
### SUMMARY

This PR adds a new feature flag, `IMPERSONATE_WITH_EMAIL_PREFIX`. When the flag is on (default is off) the user impersonation functionality will use the email prefix as the username, instead of the actual username.

This is useful for deployments using OAuth2 for login, where the username might be something like:

```
> SELECT username, email FROM ab_user;
google-oauth2|633391720998176487074, beto@example.org
```

With the feature flag on the impersonation would use `beto` as the username passed to the database.

N/A

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/RESOURCES/FEATURE_FLAGS.md
+++ b/RESOURCES/FEATURE_FLAGS.md
@@ -50,6 +50,7 @@ These features are **finished** but currently being tested. They are usable, but
 - ESTIMATE_QUERY_COST
 - GLOBAL_ASYNC_QUERIES [(docs)](https://github.com/apache/superset/blob/master/CONTRIBUTING.md#async-chart-queries)
 - HORIZONTAL_FILTER_BAR
+- IMPERSONATE_WITH_EMAIL_PREFIX
 - PLAYWRIGHT_REPORTS_AND_THUMBNAILS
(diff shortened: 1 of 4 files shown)


john-bodley:
@betodealmeida given this seems to be associated with OAuth are there existing OAuth related feature flags/configurations which could be leveraged?

eschutho (approved):
LGTM, but pending after @john-bodley's question is answered.

betodealmeida:
> @betodealmeida given this seems to be associated with OAuth are there existing OAuth related feature flags/configurations which could be leveraged?

To clarify, the OAuth I mention here is for authentication, not authorization (https://github.com/apache/superset/issues/20300).

We could do something like:

```python
from flask_appbuilder.security.manager import AUTH_REMOTE_USER

if config.AUTH_TYPE == AUTH_REMOTE_USER:
    # use email prefix instead of username
```

But I'm not familiar with all types of OAuth login, so I'd rather not assume that's always true.

PR #28771, opened 2024-05-29
fix(Mixed Chart Filter Control): Allow delete condition for `adhoc_filters_b`

rtexelm:
### SUMMARY

This is meant to fix a bug that would prevent the deletion of filters from `Query B` in the control panel of a mixed chart.
The condition to run the `canDelete` function for a filter previously only allowed for `adhoc_filters` as a valid value but the mixed chart uses `adhoc_filters_b` for the `Query B` group. This change should allow any string starting with `adhoc_filters` to be checked for deletion conditions.

B:
filters_old-ezgif com-optimize
A:
filters_new-ezgif com-optimize

### TESTING INSTRUCTIONS

1. Create a Mixed Chart
2. Add a filter to Query B
3. Delete filter
4. Delete last remaining temporal filter to test warning if is displayed

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/ControlPanelsContainer.tsx
+++ b/superset-frontend/src/explore/components/ControlPanelsContainer.tsx
@@ -485,7 +485,7 @@ export const ControlPanelsContainer = (props: ControlPanelsContainerProps) => {
         ? baseDescription(exploreState, controls[name], chart)
         : baseDescription;
 
-    if (name.startsWith('adhoc_filters')) {
+    if (name.includes('adhoc_filters')) {
       restProps.canDelete = (
         valueToBeDeleted: Record<string, any>,
         values: Record<string, any>[],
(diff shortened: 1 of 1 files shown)


superset-frontend/src/explore/components/ControlPanelsContainer.tsx
```diff
@@ -485,7 +485,7 @@ export const ControlPanelsContainer = (props: ControlPanelsContainerProps) => {
         ? baseDescription(exploreState, controls[name], chart)
         : baseDescription;
 
-    if (name === 'adhoc_filters') {
+    if (name.startsWith('adhoc_filters')) {
```
Antonio-RiveroMartnez, line 488: is it guaranteed that the filter starts with this? why not using `includes` for example?

Antonio-RiveroMartnez (approved):

eschutho:
@rtexelm can you add a small test for this, too? Thanks!

PR #28773, opened 2024-05-29
chore: make docker-compose use less memory

mistercrunch:
Playing a few tricks to make docker-compose use less memory:

- not starting the `superset-tests-worker` service on a normal `docker-compose up` by using `profile`. From my understanding this service is used only by `scripts/tests/run.sh`, and it starts/stops it on its own - doesn't assume its already running. This saves ~900MB, while maintaining 8 workers to run tests by default, but makes it confiurable if you wanted more you could more easily change it now
- starting only 2 workers by default for `superset-worker` as opposed to the default of 8, which is not necessary for dev workflows. This saves maybe ~500MB

### before
Screenshot 2024-05-29 at 4 16 17 PM

### after
Screenshot 2024-05-29 at 4 59 49 PM

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -172,6 +172,8 @@ services:
         required: true
       - path: docker/.env-local # optional override
         required: false
+    environment:
+      CELERYD_CONCURRENCY: 2
     restart: unless-stopped
     depends_on: *superset-depends-on
     user: *superset-user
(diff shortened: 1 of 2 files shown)


rusackas (approved):
Added a teensy suggestion to clarify the comment a bit. Not sure if it's worth adding a comment in the docker-compose.yml file as well for clarity. This may or may not be worth a mention on the docs site, too, if you think it might be something people fiddle with much. Aside from those nits, LGTM!

PR #25304, opened 2023-09-14
feat: Adds CLI commands to execute viz migrations

michael-s-molina:
### SUMMARY
Adds a new CLI command called `viz-migrations` to allow users to migrate charts of a specific type. This command is particularly helpful to migrate visualizations to the latest version and at the same time disable their legacy versions with the [VIZ_TYPE_DENYLIST](https://github.com/apache/superset/blob/ec6191023263ac5f8292dd37f037805c3196e029/superset/config.py#L829) configuration.

```
Usage: superset viz-migrations [OPTIONS] COMMAND [ARGS]...

  Migrates a viz from one type to another.

Commands:
  downgrade  Downgrades a viz to the previous version.
  upgrade    Upgrade a viz to the latest version.

```

### TESTING INSTRUCTIONS
1 - Execute an upgrade of a particular viz type.
2 - Downgrade the previously upgraded viz type.
3 - Check the viz versions between these steps.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #28780, opened 2024-05-30
feat: Adds Histogram chart migration logic

michael-s-molina:
### SUMMARY
This PR adds the Histogram chart migration logic (legacy ➡️ ECharts). Users can execute this migration using the [CLI command](https://github.com/apache/superset/pull/25304) and disable the legacy version with the [VIZ_TYPE_DENYLIST](https://github.com/apache/superset/blob/ec6191023263ac5f8292dd37f037805c3196e029/superset/config.py#L829) configuration.

This PR also improves the Histogram chart to automatically handle `NULL` values.

Screenshot 2024-06-05 at 10 27 14
Screenshot 2024-06-05 at 10 29 12

### TESTING INSTRUCTIONS
1 - Upgrade a Histogram (legacy) chart using the CLI command
2 - Check the new chart
3 - Downgrade a Histogram (legacy) chart using the CLI command
4 - Check the legacy chart

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Histogram/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Histogram/buildQuery.ts
@@ -25,8 +25,10 @@ export default function buildQuery(formData: HistogramFormData) {
   return buildQueryContext(formData, baseQueryObject => [
     {
       ...baseQueryObject,
+      extras: { where: `${column} IS NOT NULL` },
       columns: [...groupby, column],
       post_processing: [histogramOperator(formData, baseQueryObject)],
+      metrics: undefined,
     },
   ]);
 }
--- a/superset/utils/pandas_postprocessing/histogram.py
+++ b/superset/utils/pandas_postprocessing/histogram.py
@@ -53,7 +53,7 @@ def histogram(
         raise ValueError(f"The column '{column}' must be numeric.")
 
     # calculate the histogram bin edges
-    bin_edges = np.histogram_bin_edges(df[column], bins=bins)
+    bin_edges = np.histogram_bin_edges(df[column].dropna(), bins=bins)
 
     # convert the bin edges to strings
     bin_edges_str = [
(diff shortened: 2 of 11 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Histogram/buildQuery.ts
```diff
@@ -25,8 +25,10 @@ export default function buildQuery(formData: HistogramFormData) {
   return buildQueryContext(formData, baseQueryObject => [
     {
       ...baseQueryObject,
+      extras: { where: `${column} IS NOT NULL` },
```
michael-s-molina, line 28: Given that `column` can only contain numeric values, we automatically exclude `NULL` values from the result. This also aligns better with the limit control.

superset/utils/pandas_postprocessing/histogram.py
```diff
@@ -53,7 +53,7 @@ def histogram(
         raise ValueError(f"The column '{column}' must be numeric.")
 
     # calculate the histogram bin edges
-    bin_edges = np.histogram_bin_edges(df[column], bins=bins)
+    bin_edges = np.histogram_bin_edges(df[column].dropna(), bins=bins)
```
michael-s-molina, line 56: I added the `dropna()` to deal with `NULL` values independently of the query results.

superset-frontend/plugins/plugin-chart-echarts/src/Histogram/buildQuery.ts
```diff
@@ -25,8 +25,10 @@ export default function buildQuery(formData: HistogramFormData) {
...
       columns: [...groupby, column],
       post_processing: [histogramOperator(formData, baseQueryObject)],
+      metrics: undefined,
```
michael-s-molina, line 31: `metrics` by default is `[]` but we need it to be `undefined` to [avoid a GROUP BY](https://github.com/apache/superset/blob/a9cebe81ea8cd1ae56df05d5807baa8f05398cc5/superset/models/helpers.py#L1525).

justinpark (approved):
LGTM

PR #28781, opened 2024-05-30
feat (datasets): allowing control over certification datasets

SkinnyPigeon:
### SUMMARY
We are interested in controlling who can certify datasets. We have a central analytics engineering team who are the single-source of truth regarding datasets. We would like them to be the only team that can certify their datasets.

This introduces the ability to create a new type of role ('certifier') that we can grant to members of that team. There is an additional feature flag that would enable this feature. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -352,6 +352,14 @@ AUTH_TYPE = AUTH_DB
 PUBLIC_ROLE_LIKE: str | None = None
 
 # ---------------------------------------------------
+# Certification config
+# ---------------------------------------------------
+# Allows the limiting of users who can certify datasets based on the role.
+# Requires the creation of a role within Superset with the matching name
+# and the setting of the LIMIT_CERTIFICATION feature flag to True.
+# ROLE_CERTIFIER: str = "certifier"
+
+# ---------------------------------------------------
 # Babel config for translations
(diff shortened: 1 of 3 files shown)


superset/config.py
```diff
@@ -496,6 +504,8 @@ class D3Format(TypedDict, total=False):
     "PLAYWRIGHT_REPORTS_AND_THUMBNAILS": False,
     # Set to True to enable experimental chart plugins
     "CHART_PLUGINS_EXPERIMENTAL": False,
+    # Set to True to limit the number of users who can certify datasets
+    "LIMIT_CERTIFICATION": False,
```
john-bodley, line 508: Couldn't this be controlled via Superset's/FAB's existing permission/view model? This feels more consistent/coherent than creating more custom logic in the security manager.

SkinnyPigeon: Thanks. I'll give that another look. I couldn't see within that section how to target individual fields. We want all users to be able to edit their own datasets to add things like metrics etc. but limit who can add this field.

SkinnyPigeon:
Will try a different approach

PR #28784, opened 2024-05-31
feat: skip `npm run build` in docker-compose

mistercrunch:
When using docker-compose, we mount the local drive and fire up a `npm run dev` which is effectively webpack in --watch mode, so that it looks for changes in frontend files and compiles-as-you-change interactively.

`npm run build` is in the Dockerfile, specifically in the superset-node layer which is used to compile and later COPY the static assets into a python-based layer, `lean`.

But in the context of docker-compose, we don't need all of it, and that `npm run build` has gotten expensive in terms of memory usage and takes a while to build. In recent build memory monitoring I've seen the docker container peak at like 13GB memory during this phase.

So in this PR I'm:
- making `npm run build` optional with a dockerfile ARG (default is current behavior, but docker-compose skips it)
- adding a `superset-node-base` layer for that superset-node service in docker-compose that runs `npm run dev`, ensuring the same base for both processes, but stopping/diverging prior to `npm ci`
- making `npm install` run only if node_module is not present in the superset-node service. The user can simply `npm i` locally if/when they alter package.json. That speeds up firing up `docker-compose up` as unless you nuke node_modules/ it'll skip that step

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,8 @@ x-common-build: &common-build
   target: dev
   cache_from:
     - apache/superset-cache:3.10-slim-bookworm
+  args:
+    INCLUDE_NODE: false
 
 services:
   nginx:
@@ -147,7 +149,11 @@ services:
       disable: true
 
   superset-node:
-    image: node:18
+    build:
+      context: .
+      target: superset-node-base
+      cache_from:
+        - type=registry,ref=apache/superset-cache:3.10-slim-bookworm
     environment:
       # set this to false if you have perf issues running the npm i; npm run dev in-docker
       # if you do so, you have to run this manually on the host, which should perform better!
--- a/superset-frontend/webpack.config.js
+++ b/superset-frontend/webpack.config.js
@@ -496,7 +496,7 @@ const config = {
     'react/lib/ReactContext': true,
   },
   plugins,
-  devtool: 'source-map',
+  devtool: isDevMode ? 'source-map' : false,
 };
 
 // find all the symlinked plugins and use their source code for imports
--- a/superset-frontend/webpack.config.js
+++ b/superset-frontend/webpack.config.js
@@ -496,7 +496,7 @@ const config = {
     'react/lib/ReactContext': true,
   },
   plugins,
-  devtool: isDevMode ? 'source-map' : false,
+  devtool: 'source-map',
 };
 
 // find all the symlinked plugins and use their source code for imports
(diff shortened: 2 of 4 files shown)


superset-frontend/webpack.config.js
```diff
@@ -496,7 +496,7 @@ const config = {
     'react/lib/ReactContext': true,
   },
   plugins,
-  devtool: 'source-map',
+  devtool: isDevMode ? 'source-map' : false,
```
mistercrunch, line 499: read somewhere that turning off source maps in prod may help memory usage, but also thinking it's better to not have source maps in prod as it's lighter weight / more secure

docker-compose.yml
```diff
@@ -147,7 +149,11 @@ services:
       disable: true
 
   superset-node:
-    image: node:18
+    build:
```
mistercrunch, line 152: This makes things a bit more DRY and in sync with the Dockerfile, main thing is the base is set in docker for this, and there's a `frontend-mem-nag` script that checks some assumption and raises message if docker conditions aren't set properly.

mistercrunch:
@dpgaspar @craig-rueda curious what you think here

docker-compose.yml
rusackas: That memory nag seems to be set to 3GB. Seems like we might as well raise it, no? Not sure what a reasonable minimum is at this point, but something significant - maybe 10GB?

superset-frontend/webpack.config.js
rusackas: The source maps are SUPER helpful in prod, actually... it allows us to help pinpoint runtime errors in the codebase rather than see garbage minimized JS and stacktraces. I would leave these on.

rusackas (changes requested):
I think we should leave source-maps on (they're quite useful), these the "changes requested" button here. Also wonder if we should bump that minimum memory spec to... more memory. Otherwise this looks great!

mistercrunch:
Source maps are useful, but are they out of place in production? Though clearly that's a total bycatch and probably doesn't belong in this PR

docker-compose.yml
mistercrunch: ok let's try 10, unclear what exactly is needed and different contexts, but 10 seems much more accurate than 3

mistercrunch: if people have ideas as to how to contain this docker beast it'd be great

mistercrunch:
Actually I think I may have been misled by GPT and thought docker was smarter than it is, as an immutable system it's hard to do anything conditional in docker. I think we'd have to remodel things quite a bit and have multiple docker files to stay sane while doing this.

PR #24539, opened 2023-06-28
feat(sqllab): non-blocking persistence mode

justinpark:
### SUMMARY
Following up [SIP-93](https://github.com/apache/superset/issues/21385), this commit implements the first step of non-blocking persistence mode for the current working sqllab editor activities.

As described in [SIP-93](https://github.com/apache/superset/issues/21385), the existing PERSISTENCE logic causes a user interaction delay due to the network (or server side) lagging. For example, user had to wait the server response in order to update the tab name or query limit count. (see the before screenshot)

Therefore, this commit aims to improve current tab activities by replacing the following blocking logics with a non-blocking way by cumulating the changes and syncing them with the latest updates asynchronously: 

```
- querySuccess
- queryFailed
- toggleLeftBar
- queryEditorSetDb
- queryEditorSetSchema
- queryEditorSetAutorun
- queryEditorSetTitle
- queryEditorSetSql
- queryEditorSetQueryLimit
- queryEditorSetTemplateParams
```

#### Mechanism for syncing unsaved changes

debounced

There can be unsaved changes since it is syncing in a non-blocking way and asynchronously. To address this issue, unsaved changes will be stored in local storage (using localStorage in persistence mode) until they are successfully saved on the server. This hybrid approach ensures that user data is not lost and that any unsaved changes are retrieved, even in the event of an unexpected interruption.

#### Resolve conflicts

When a user opens multiple windows, they may be overwritten by the inactive editor tab state. To prevent this unexpected overwrite, the latest update time will be compared with the local cache state, and the out-of-updated state will be discarded. This ensures that users can work on multiple windows without losing their progress or having their work overwritten.

#### Following up

Creating new tabs and removing tabs are still blocking way in this commit but will be migrated into the non-blocking way in future PRs. Additionally, the existing migrateQueryEditorFromLocalStorage logic will be consolidated or replaced by the AutoSync with mutation APIs in the upcoming PRs. These changes will further improve the user experience by making the entire process non-blocking and asynchronous.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:

https://github.com/apache/superset/assets/1392866/7efa5a29-b7e0-4698-bb39-f4295693588d

After:

https://github.com/apache/superset/assets/1392866/3d5b2817-add5-41c8-9071-890e82e81615

### TESTING INSTRUCTIONS
Set SQLLAB_BACKEND_PERSISTENCE true and play with the SQL Lab editor and keep hard refresh to verify the persisted states.

[...]

PR #28795, opened 2024-05-31
refactor(sqllab): nonblocking new query editor

justinpark:
### SUMMARY
Following up #24539
This commit modifies the `addQueryEditor` function to asynchronously handle the process of creating a queryEditor in localStorage in the same way as in non-persistent mode and periodically saving the queryEditor state to persistence via AutoSync's `migrateQueryEditorFromLocalStorage`. 

This commit made the following changes:
- Migrate migrateQueryEditorFromLocalStorage from the App's componentDidMount to EditorAutoSync.
- EditorAutoSync will execute migrateQueryEditorFromLocalStorage periodically.
- MIGRATE_QUERY_EDITOR will replace the old queryEditor at the same location instead of pushing it to the end.
- Remove the logic of syncing localStorage tabHistory since it is already handled in the middleware.

In persistence mode, new tabs can now open and access data directly from localStorage without the delays caused by network or database latency.

* with network latency

Before:

https://github.com/apache/superset/assets/1392866/f07ea4e1-efcb-47b9-89f6-dfaa1726dcea

After:

https://github.com/apache/superset/assets/1392866/9f16fb8e-2a13-43dd-9e44-84631d5ac7a5

### TESTING INSTRUCTIONS

Set SQLLAB_PERSISTENCE_MODE on and then open a multiple tabs
please reload the page and then check all tabs are persisted

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -467,7 +467,16 @@ function migrateQuery(queryId, queryEditorId, dispatch) {
     );
 }
 
-export function migrateQueryEditorFromLocalStorage(queryEditor) {
+/**
+ * Persist QueryEditor from local storage to backend tab state.
+ * This ensures that when new tabs are created, query editors are
+ * asynchronously stored in local storage and periodically synchronized
+ * with the backend.
+ * When switching to persistence mode, the QueryEditors previously
+ * stored in local storage will also be synchronized to the backend
+ * through syncQueryEditor.
+ */
+export function syncQueryEditor(queryEditor) {
   return function (dispatch, getState) {
     const { tables, queries } = getState().sqlLab;
     const localStorageTables = tables.filter(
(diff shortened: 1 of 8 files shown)


superset-frontend/src/SqlLab/actions/sqlLab.js
```diff
@@ -467,12 +467,15 @@ function migrateQuery(queryId, queryEditorId, dispatch) {
     );
 }
 
-export function migrateQueryEditorFromLocalStorage(
-  queryEditor,
-  tables,
-  queries,
-) {
-  return function (dispatch) {
+export function migrateQueryEditorFromLocalStorage(queryEditor) {
```
michael-s-molina, line 470: Is this a migration or you're syncing the query editor state with the backend?

superset-frontend/src/SqlLab/actions/sqlLab.js
```diff
@@ -467,7 +467,16 @@ function migrateQuery(queryId, queryEditorId, dispatch) {
     );
 }
 
-export function migrateQueryEditorFromLocalStorage(queryEditor) {
+/**
```
michael-s-molina, line 470: Nice!

michael-s-molina (approved):

PR #28783, opened 2024-05-30
fix: use upload v2 for slack

eschutho:
### SUMMARY

After https://github.com/apache/superset/pull/28423 I found one more place where we are still calling the old slack api method. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #28797, opened 2024-06-01
fix: use channel id with new slack api for file uploads

eschutho:
### SUMMARY

Fixing changes from https://github.com/apache/superset/pull/28783 where we needed to update the slack sdk call to the new files_upload_v2 method. The api changes a few things that were missed in the original PR: the `channels` arg is now the channel id instead of name. Also it is recommended to use `channel` instead of `channels`. I also found that the slack util file wasn't being used and cleaned up a few other affected parts. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/reports/notifications/slack.py
```diff
@@ -60,16 +61,24 @@ class SlackNotification(BaseNotification):  # pylint: disable=too-few-public-met
...
-        :returns: The comma separated list of channel(s)
+        :returns: A list of channel ids: "EID676L"
         """
         recipient_str = json.loads(self._recipient.recipient_config_json)["target"]
 
-        return ",".join(get_email_address_list(recipient_str))
+        channel_recipients = get_email_address_list(recipient_str)
+
+        conversations_list_response = client.conversations_list(
+            types="public_channel,private_channel"
```
eschutho, line 74: We now need to map the list of names to a list of ids. This is the slack sdk method that fetches information that has both the name and id.

sadpandajoe: Will this work if the application doesn't have the right scope? I tried testing this [here](https://api.slack.com/methods/conversations.list/test) with my token and got this response:

```
{
    "ok": false,
    "error": "missing_scope",
    "needed": "channels:read,groups:read,mpim:read,im:read",
    "provided": "commands,chat:write,chat:write.public,channels:history,im:history,mpim:history,groups:history,files:write"
}
```

eschutho: oh, that's a good point. @Vitor-Avila had a great solution which is to try the deprecated method first and then try the new one, since only new apps are affected right now. I'm also nervous about introducing a breaking change here.

superset/reports/notifications/slack.py
```diff
@@ -60,16 +62,25 @@ class SlackNotification(BaseNotification):  # pylint: disable=too-few-public-met
 
     type = ReportRecipientType.SLACK
 
-    def _get_channel(self) -> str:
+    def _get_channels(self, client: WebClient) -> List[str]:
```
betodealmeida, line 65: Nit: you can use`list[str]` now, given the lowest Python version we support.

superset/reports/notifications/slack.py
betodealmeida: I assume right now we store the names of the channels, which is why we're mapping back to IDs? In the future it would be nice to store the IDs directly, not only it would eliminate the need for this mapping but also reports would keep working when channels are renamed.

superset/reports/notifications/slack.py
```diff
@@ -162,29 +177,40 @@ def _get_body(self) -> str:
...
-        title = self._content.name
-        channel = self._get_channel()
-        body = self._get_body()
-        global_logs_context = getattr(g, "logs_context", {}) or {}
-        try:
-            client = get_slack_client()
-            # files_upload returns SlackResponse as we run it in sync mode.
-            if files:
+    @deprecated(deprecated_in="4.1")
+    def _deprecated_upload_files(
```
betodealmeida, line 190: Nice!

superset/reports/notifications/slack.py
```diff
@@ -193,6 +219,46 @@ def send(self) -> None:
...
+        try:
+            client = get_slack_client()
+            title = self._content.name
+            body = self._get_body()
+
+            try:
+                channels = self._get_channels(client)
+            except SlackApiError:
+                logger.warning(
+                    "Slack scope missing. Using deprecated API to get channels. Please update your Slack app to use the new API.",
```
betodealmeida, line 236: Do we know what scope is missing here? Can we add it to the error message?

tests/integration_tests/fixtures/world_bank_dashboard.py
```diff
@@ -143,6 +145,21 @@ def _cleanup(dash_id: int, slices_ids: list[int]) -> None:
...
 
+def _cleanup_reports(dash_id: int, slices_ids: list[int]) -> None:
+    reports_with_dash = (
+        db.session.query(ReportSchedule).filter_by(dashboard_id=dash_id).all()
+    )
+    reports_with_slices = (
+        db.session.query(ReportSchedule)
+        .filter(ReportSchedule.chart_id.in_(slices_ids))
+        .all()
+    )
```
betodealmeida, line 156: You can probably combine this into something like (untested):

```python
reports = (
    db.session.query(ReportSchedule)
    .filter(
        or_(
            ReportSchedule.dashboard_id == dash_id),
            ReportSchedule.chart_id.in_(slices_ids),
        )
    ).all()
)
```

betodealmeida (approved):
Looks good! Left a few comments but nothing blocking.

superset/reports/notifications/slack.py
eschutho: I assume it will be sent by slack to the error message.

eschutho:
I'll address @betodealmeida's comments in a fast-follow.

superset/reports/notifications/slack.py
eschutho: Yeah, correct, since that is what is in the report form. If we stored the ids, then we would need to map then back to the name when someone tries to view/edit the form, but that's prob still better.

superset/reports/notifications/slack.py
eschutho: We expect `channels:read` as per the difference between v1 and v2 apis, but it could be anything if they're not set up correctly.

superset/reports/notifications/slack.py
eschutho: I was thinking about this more @betodealmeida and it would be a breaking change because the existing applications would need the `channels:read` slack scope for this, but maybe for the next major release, it would a nice improvement to have a multiselect ui for the channel and then also store the id. ❤️

PR #28838, opened 2024-06-03
fix: Update downgrade path for migration to remove sl_tables

sadpandajoe:
### SUMMARY
In https://github.com/apache/superset/pull/28704 we removed `sl_tables` however, in the downgrade path there are dependencies on other `sl_tables` that cause the downgrade to fail. 

`sl_datasets` referenced by `sl_dataset_users` and `sl_dataset_tables`
`sl_tables` referenced by `sl_dataset_tables` and `sl_table_columns`
`sl_columns` referenced by `sl_table_columns` and `sl_dataset_columns`

so the new creation order is
`sl_datasets`
`sl_tables`
`sl_dataset_columns`
`sl_dataset_users`
`sl_dataset_tables`
`sl_table_columns`
`sl_dataset_columns`

Also normalized cases when creating columns since `sa.DATETIME()` was creating an issue but `sa.DateTime()` would create the tables without any errors.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Antonio-RiveroMartnez (approved):
lgtm

mistercrunch (approved):
Oops my bad, thanks for fixing this. I'm pretty sure I tested the downgrade at some point, but maybe I changed something after (?).

I'm thinking it'd be a good thing to add a CI step that tests downgrades.

PR #29077, opened 2024-06-04
chore: only use cypress.io when triggered manually

mistercrunch:
Preset is picking up the bill for Cypress.io and it's getting pricey for the usage we have for it. In this PR, I'm moving normal pull_request and push events to run in local GHA mode without the `--record` flag. This means we can't use the `--parallel` feature, but I wrote a little python script that segments the jobs and make them run in parallel.

I'm also providing a `workflow_dispatch` option where you can trigger a cypress.io run in dashboard mode, so you can still go use their stuff on a given branch if/where needed.

### DIFF
--- a/.github/workflows/superset-e2e.yml
+++ b/.github/workflows/superset-e2e.yml
@@ -13,6 +13,14 @@ on:
         description: 'Use Cypress Dashboard (true/false) [paid service - trigger manually when needed]'
         required: false
         default: 'false'
+      ref:
+        description: 'The branch or tag to checkout'
+        required: false
+        default: ''
+      pr_id:
+        description: 'The pull request ID to checkout'
+        required: false
+        default: ''
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
@@ -31,7 +39,7 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        parallel_id: [0, 1, 2, 3, 4]
+        parallel_id: [0, 1, 2, 3, 4, 5]
         browser: ["chrome"]
     env:
       SUPERSET_ENV: development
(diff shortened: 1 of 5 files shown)


mistercrunch:
Screenshot 2024-06-04 at 11 24 27 AM

ok so --parallel is a paid feature, seems we'll have to find some other way to segment/parallelize

rusackas:
Doing it on `master` only might make sense too.

mistercrunch:
I think I'm finding a solution. Turns out cypress doesn't do much of the compute/work here, and the dashboard is not super useful, so I'm moving to trigger the dashboard manually using workflow_dispatch, only when useful. 

The other thing is the parallelization is also part of the service, and it's "intelligent" meaning it know about recent runs and tries to balance / segment the jobs properly. I've been workign with GPT at segmenting the files based on a hash of the filename, so as long as we don't have some files much bigger than other it should distribute all write. If not we can just break down the slower files into smaller chunks.

scripts/cypress_run.py
```diff
@@ -0,0 +1,146 @@
...
+    os.environ["TERM"] = "xterm"
+    os.environ["ELECTRON_DISABLE_GPU"] = "true"
+    build_id = generate_build_id()
+    browser = os.getenv("CYPRESS_BROWSER", "chrome")
+
+    if use_dashboard:
+        # Run using cypress.io service
+        cypress_record_key = (
+            subprocess.check_output(
+                ["echo", os.getenv("CYPRESS_KEY") or "DUMMY", "|", "base64", "--decode"]
```
rusackas, line 60: Whenever there's an action that runs a script, and the script uses a KEY, it raises my security flag now :) CC @dpgaspar to make sure this is OK :)

.github/workflows/superset-e2e.yml
```diff
@@ -20,12 +34,12 @@ jobs:
...
     strategy:
       # when one test fails, DO NOT cancel the other
-      # containers, because this will kill Cypress processes
+      # parallel_id, because this will kill Cypress processes
       # leaving the Dashboard hanging ...
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        containers: [1, 2, 3]
+        parallel_id: [0, 1, 2, 3, 4, 5]
```
rusackas, line 42: FWIW, since the setup for these runners takes a while, I was keeping the number of workers low enough to see a higher ROI on parallelization. This might shave off a minute or two, but will burn twice the coal.

mistercrunch: right, there's like 7 minutes of setup and in aggregate maybe 45 minutes of parallelizable stuff. At 6 workers it's about 50/50 of the time on overhead VS work. I thought bumping this up made some sense as it has now become the bottleneck as the longest CI workflow.

mistercrunch: One thing that could be nice would be to have a single worker prepare some stuff for the matrix as a whole, so that overhead could remain 7 minutes, but not be done on N workers... Not sure what GHA offers around that.

scripts/cypress_run.py
dpgaspar: this is triggered by `pull_request` GHA so should be ok, and I'm assuming `CYPRESS_KEY` is not a secret right?
yes script can be "abused" by a PR but will it have privileges to do anything destructive?

rusackas: I think not, I didn't realize the key is public.

rusackas (approved):
❤️

.github/workflows/superset-e2e.yml
```diff
@@ -47,19 +62,30 @@ jobs:
...
+        if: github.event_name == 'workflow_dispatch' && github.event.inputs.ref != ''
+        uses: actions/checkout@v2
         with:
-          ref: "refs/pull/${{ github.event.number }}/merge"
           persist-credentials: false
+          ref: ${{ github.event.inputs.ref }}
           submodules: recursive
+      - name: Checkout using PR ID (workflow_dispatch)
+        if: github.event_name == 'workflow_dispatch' && github.event.inputs.pr_id != ''
+        uses: actions/checkout@v2
```
rusackas, line 83: Might wanna bump this?

rusackas (approved):

rusackas (approved):

PR #29096, opened 2024-06-05
feat(alerts): enable tab selection for dashboard alerts/reports

fisjac:
### SUMMARY

Implements a new feature enabling a tab to be selected as a view for an alert/report. When a dashboard is selected as the content for a report, an antd TreeSelect component is rendered to load and select tab options. When selected, the tab selection is added to the report payload as an anchor.

On execution, the feature takes advantage of existing logic for the report.extra.dashboard.anchor_tab attribute of the ReportSchedule SQLA model. When extra metadata are attached to the report schedule, a permalink is generated with the metadata included, and this permalink is then used to generate screenshots of the selected dashboard.

https://www.loom.com/share/e075b22bd5574e30bcf3bbec5984884f?sid=aeefd476-bf0a-4289-b13f-6b5fe2854156

### TESTING INSTRUCTIONS

Create a tab or tabs within an existing dashboard
Navigate to the alerts and reports listview from the settings menu
Create a new report
Under the content section select the dashboard
A selection field for tabs should appear with the tabs available within the select options
select a tab and save the report with the other required fields
at the selected interval, a report should be sent out with the selected tab in focus

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: ALERT_REPORTS, ALERT_REPORT_TABS
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/AlertReportModal.test.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.test.tsx
@@ -106,11 +106,18 @@ const ownersEndpoint = 'glob:*/api/v1/alert/related/owners?*';
 const databaseEndpoint = 'glob:*/api/v1/alert/related/database?*';
 const dashboardEndpoint = 'glob:*/api/v1/alert/related/dashboard?*';
 const chartEndpoint = 'glob:*/api/v1/alert/related/chart?*';
+const tabsEndpoint = 'glob:*/api/v1/dashboard/1/tabs';
 
 fetchMock.get(ownersEndpoint, { result: [] });
 fetchMock.get(databaseEndpoint, { result: [] });
 fetchMock.get(dashboardEndpoint, { result: [] });
 fetchMock.get(chartEndpoint, { result: [{ text: 'table chart', value: 1 }] });
+fetchMock.get(tabsEndpoint, {
+  result: {
+    all_tabs: {},
+    tab_tree: [],
+  },
+});
 
 // Create a valid alert with all required fields entered for validation check
 
@@ -411,6 +418,21 @@ test('renders screenshot options when dashboard is selected', async () => {
       name: /ignore cache when generating report/i,
     }),
   ).toBeInTheDocument();
+});
+
+test('renders tab selection when Dashboard is selected', async () => {
+  render(<AlertReportModal {...generateMockedProps(false, true, true)} />, {
+    useRedux: true,
+  });
+  userEvent.click(screen.getByTestId('contents-panel'));
+  await screen.findByText(/test dashboard/i);
+  expect(
+    screen.getByRole('combobox', { name: /select content type/i }),
+  ).toBeInTheDocument();
+  expect(
+    screen.getByRole('combobox', { name: /dashboard/i }),
+  ).toBeInTheDocument();
+  expect(screen.getByText(/select tab/i)).toBeInTheDocument();
 });
 
 test('changes to content options when chart is selected', async () => {
--- a/superset-frontend/src/features/alerts/AlertReportModal.tsx
+++ b/superset-frontend/src/features/alerts/AlertReportModal.tsx
@@ -46,7 +46,7 @@ import TimezoneSelector from 'src/components/TimezoneSelector';
 import { propertyComparator } from 'src/components/Select/utils';
 import withToasts from 'src/components/MessageToasts/withToasts';
 import Owner from 'src/types/Owner';
-import { AntdCheckbox, AsyncSelect, Select } from 'src/components';
+import { AntdCheckbox, AsyncSelect, Select, TreeSelect } from 'src/components';
 import TextAreaControl from 'src/explore/components/controls/TextAreaControl';
 import { useCommonConf } from 'src/features/databases/state';
 import { InfoTooltipWithTrigger } from '@superset-ui/chart-controls';
@@ -57,12 +57,15 @@ import {
   ChartObject,
   DashboardObject,
   DatabaseObject,
+  Extra,
   MetaObject,
   Operator,
   Recipient,
   AlertsReportsConfig,
   ValidationObject,
   Sections,
+  TabNode,
+  SelectValue,
 } from 'src/features/alerts/types';
 import { useSelector } from 'react-redux';
 import { UserWithPermissionsAndRoles } from 'src/types/bootstrapTypes';
@@ -79,11 +82,6 @@ const TEXT_BASED_VISUALIZATION_TYPES = [
   'table',
   'paired_ttest',
 ];
-
-type SelectValue = {
-  value: string;
-  label: string;
-};
 
 export interface AlertReportModalProps {
   addSuccessToast: (msg: string) => void;
@@ -103,6 +101,12 @@ const DEFAULT_NOTIFICATION_METHODS: NotificationMethodOption[] = [
   NotificationMethodOption.Email,
 ];
 const DEFAULT_NOTIFICATION_FORMAT = 'PNG';
+const DEFAULT_EXTRA_DASHBOARD_OPTIONS: Extra = {
+  dashboard: {
+    anchor: '',
+  },
+};
+
 const CONDITIONS = [
   {
     label: t('< (Smaller than)'),
@@ -215,6 +219,10 @@ const StyledModal = styled(Modal)`
       flex: 1 1 auto;
     }
   }
+`;
+
+const StyledTreeSelect = styled(TreeSelect)`
+  width: 100%;
 `;
 
 const StyledSwitchContainer = styled.div`
@@ -439,6 +447,8 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
   const [sourceOptions, setSourceOptions] = useState<MetaObject[]>([]);
   const [dashboardOptions, setDashboardOptions] = useState<MetaObject[]>([]);
   const [chartOptions, setChartOptions] = useState<MetaObject[]>([]);
+  const [tabOptions, setTabOptions] = useState<TabNode[]>([]);
+
   // Validation
   const [validationStatus, setValidationStatus] = useState<ValidationObject>({
     [Sections.General]: {
@@ -489,6 +499,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
   const isEditMode = alert !== null;
   const formatOptionEnabled =
     isFeatureEnabled(FeatureFlag.AlertsAttachReports) || isReport;
+  const tabsEnabled = isFeatureEnabled(FeatureFlag.AlertReportTabs);
 
   const [notificationAddState, setNotificationAddState] =
     useState<NotificationAddStatus>('active');
@@ -545,6 +556,7 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     active: true,
     creation_method: 'alerts_reports',
     crontab: ALERT_REPORTS_DEFAULT_CRON_VALUE,
+    extra: DEFAULT_EXTRA_DASHBOARD_OPTIONS,
     log_retention: ALERT_REPORTS_DEFAULT_RETENTION,
     working_timeout: ALERT_REPORTS_DEFAULT_WORKING_TIMEOUT,
     name: '',
@@ -591,6 +603,22 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     settings.splice(index, 1);
     setNotificationSettings(settings);
     setNotificationAddState('active');
+  };
+
+  const updateAnchorState = (value: any) => {
+    setCurrentAlert(currentAlertData => {
+      const dashboardState = currentAlertData?.extra?.dashboard;
+      const extra = {
+        dashboard: {
+          ...dashboardState,
+          anchor: value,
+        },
+      };
+      return {
+        ...currentAlertData,
+        extra,
+      };
+    });
   };
 
   // Alert fetch logic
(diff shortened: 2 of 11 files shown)


superset-frontend/src/dashboard/containers/DashboardPage.tsx
```diff
@@ -147,6 +147,7 @@ export const DashboardPage: FC<PageProps> = ({ idOrSlug }: PageProps) => {
         if (!isDashboardHydrated.current) {
           isDashboardHydrated.current = true;
         }
+        console.log('active tabs:', activeTabs);
```
eschutho, line 150:
```suggestion
```

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -774,6 +798,28 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+  const dashboard = currentAlert?.dashboard;
+  useEffect(() => {
+    if (!tabsEnabled) return;
+
+    if (dashboard?.value) {
+      SupersetClient.get({
+        endpoint: `/api/v1/dashboard/${dashboard.value}/tabs`,
+      }).then(response => {
+        const tabTree = response.json.result.tab_tree;
+        const allTabs = response.json.result.all_tabs;
```
eschutho, line 810: I think you should be able to destructure here to get the variables?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -774,6 +798,28 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+
+        setTabOptions(tabTree);
+        const currentAnchor = currentAlert?.extra?.dashboard?.anchor;
+        if (currentAnchor) {
+          if (!(currentAnchor in allTabs)) {
```
eschutho, line 815: can you combine these two lines into one if statement?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -1602,6 +1652,21 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
           </StyledInputContainer>
+          {tabsEnabled && contentType === 'dashboard' && (
+            <StyledInputContainer>
+              <>
+                <div className="control-label">{t('Select tab')}</div>
+                <TreeSelect
+                  disabled={tabOptions?.length === 0}
+                  treeData={tabOptions}
+                  value={currentAlert?.extra?.dashboard?.anchor}
+                  onSelect={value => updateAnchorState(value)}
```
eschutho, line 1663:
```suggestion
                  onSelect={updateAnchorState}
```

superset-frontend/src/dashboard/containers/DashboardPage.tsx
geido, line 150:
```suggestion
```

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -774,6 +798,25 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+        const { tab_tree: tabTree, all_tabs: allTabs } = response.json.result;
+        setTabOptions(tabTree);
+        if (currentAlert?.extra?.dashboard?.anchor) {
+          if (!(currentAlert?.extra?.dashboard?.anchor in allTabs)) {
+            updateAnchorState('');
+          }
+        }
+      });
+    }
+  }, [dashboard, tabsEnabled, currentAlert?.extra]);
```
geido, line 818:
```suggestion
  }, [dashboard?.value, tabsEnabled, currentAlert?.extra]);
```

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -1602,6 +1649,21 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+          {tabsEnabled && contentType === 'dashboard' && (
+            <StyledInputContainer>
+              <>
+                <div className="control-label">{t('Select tab')}</div>
+                <TreeSelect
+                  disabled={tabOptions?.length === 0}
+                  treeData={tabOptions}
+                  value={currentAlert?.extra?.dashboard?.anchor}
+                  onSelect={updateAnchorState}
+                  style={{ width: '100%' }}
```
geido, line 1661: Can we use styled or css from emotion for this?

superset/commands/report/execute.py
```diff
@@ -168,15 +168,15 @@ def _get_url(
                 force=force,
                 **kwargs,
             )
-
         # If we need to render dashboard in a specific state, use stateful permalink
-        if dashboard_state := self._report_schedule.extra.get("dashboard"):
+        if (
+            dashboard_state := self._report_schedule.extra.get("dashboard")
+        ) and feature_flag_manager.is_feature_enabled("ALERT_REPORT_TABS"):
```
geido, line 174: Is this supposed to be an `and`?

superset-frontend/src/features/alerts/AlertReportModal.test.tsx
```diff
@@ -413,6 +420,21 @@ test('renders screenshot options when dashboard is selected', async () => {
...
+  });
+  userEvent.click(screen.getByTestId('contents-panel'));
+  await screen.findByText(/test dashboard/i);
+  expect(
+    screen.getByRole('combobox', { name: /select content type/i }),
+  ).toBeInTheDocument();
+  expect(
+    screen.getByRole('combobox', { name: /dashboard/i }),
+  ).toBeInTheDocument();
+  expect(screen.getByText(/select tab/i)).toBeInTheDocument();
```
geido, line 435: Should we verify that the TreeSelect renders and has some options?

superset/commands/report/execute.py
fisjac: yes, the logic here is saying that if there is dashboard metadata and the feature flag is enabled, open in a stateful permalink, else open in the normal url without permalinks.

superset-frontend/src/features/alerts/AlertReportModal.test.tsx
fisjac: None of the other select options are being rendered for owners, dashboards, charts, or tabs. I haven't been able to get options to show for antd select components within the RTL framework. Is it worth trying another testing framework for this?

geido: I think it's ok for now

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -776,6 +804,25 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     [],
   );
 
+  const dashboard = currentAlert?.dashboard;
+  useEffect(() => {
+    if (!tabsEnabled) return;
+
+    if (dashboard?.value) {
+      SupersetClient.get({
```
geido, line 812: What happens if this fails? Should we catch it and display an error message?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -776,6 +804,25 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+        endpoint: `/api/v1/dashboard/${dashboard.value}/tabs`,
+      }).then(response => {
+        const { tab_tree: tabTree, all_tabs: allTabs } = response.json.result;
+        setTabOptions(tabTree);
+        if (currentAlert?.extra?.dashboard?.anchor) {
+          if (!(currentAlert?.extra?.dashboard?.anchor in allTabs)) {
```
geido, line 818: You are already checking above if this exists. Maybe it would help readability to move it inside a const.

```
const anchor = currentAlert?.extra?.dashboard?.anchor
if (anchor && !anchor in allTabs)...
```

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -887,6 +934,28 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
       endpoint: `/api/v1/chart/${chart.value}`,
     }).then(response => setChartVizType(response.json.result.viz_type));
 
+  const updateEmailSubject = () => {
+    if (contentType === 'chart') {
+      if (currentAlert?.name || currentAlert?.chart?.label) {
+        setEmailSubject(
+          `${currentAlert?.name}: ${currentAlert?.chart?.label || ''}`,
```
geido, line 941: I am not sure if the condition here is correct but it seems that the intention was to check if `currentAlert?.name` exists? What if name is `undefined`, is that an option?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -887,6 +934,28 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+      } else {
+        setEmailSubject('');
+      }
+    } else if (contentType === 'dashboard') {
+      if (currentAlert?.name || currentAlert?.dashboard?.label) {
+        setEmailSubject(
+          `${currentAlert?.name}: ${currentAlert?.dashboard?.label || ''}`,
+        );
+      } else {
+        setEmailSubject('');
```
geido, line 952: Can we use returns here to avoid duplicate code or maybe a switch?

superset/commands/report/execute.py
geido: What I am saying is that now this is only applied if the `ALERT_REPORT_TABS` is enabled but shouldn't it be applied anyway of there is an `extra` parameter? Is there a reason to tie this specifically to the `ALERT_REPORT_TABS` FF?

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -646,14 +674,14 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
       ),
       recipients,
       report_format: reportFormat || DEFAULT_NOTIFICATION_FORMAT,
+      extra: contentType === 'dashboard' ? currentAlert?.extra : null,
```
geido, line 677: Can we create a type for this `contentType` instead of relying on strings?

kasiazjc:
Functionality wise - looks good! Thanks @fisjac

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -593,6 +605,22 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
     setNotificationAddState('active');
   };
 
+  const updateAnchorState = (value: any) => {
```
kgabryje, line 608: Can we use more specific type?

yousoph:
/testenv up FEATURE_ALERT_REPORTS=True FEATURE_ALERT_REPORT_TABS = True

superset/commands/report/execute.py
fisjac: My rationale for hiding that functionality behind the `ALERT_REPORT_TABS` FF was that if a report was saved with an `extra` parameter, then the FF was turned off, there would no longer be an interface for viewing/updating the currently selected tab. It seemed more intuitive to have the behavior default to ignoring the `extra` parameter altogether while the FF was disabled.

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: This wasn't a piece of the code I was working in. I just moved it up a few lines to avoid the `updateEmailSubject` being called before it's defined warnings.

I can look more into it.

yousoph:
/testenv up FEATURE_ALERT_REPORTS=True FEATURE_ALERT_REPORT_TABS=True

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: Ideally, I'd like for this to be typed as  `string | undefined`, but I'm running into issues from antd's typing for the `onSelect` prop which is typed as `'RawValueType | LabelValueType'` and throws an error on compile when typed as such. I've used any as a workaround.

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: I've added a `ContentType` Enum in the latest commit to address this.

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: Made the code more readable in the latest commit.

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: implemented switch

superset-frontend/src/features/alerts/AlertReportModal.tsx
fisjac: Added a `.catch()` with `dangerToast`

superset/commands/report/execute.py
geido: If we are confident that extra is only used for this use case, then it's good.

geido (approved):

rusackas:
Any reason not to merge this? Itchy trigger finger kicking in ;)

PraveshEverstage:
Hi @fisjac , @geido  does this feature support multi-tab selection? If not, could you share why it hasn’t been implemented yet? Were there any specific challenges or constraints? It would be really helpful to understand the context. Thanks!

PR #29098, opened 2024-06-05
fix: Cypress CI process while opening PR from a fork

mistercrunch:
As observed in https://github.com/apache/superset/pull/28774 (thanks for catching this @rusackas), I broke the PR-from-fork in my recently merged PR here https://github.com/apache/superset/pull/29077 .

This PR removes special handling of `ref:` on github_event in (pull_request, push)

### DIFF
--- a/superset/__init__.py
+++ b/superset/__init__.py
@@ -14,8 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-"""Package's main module!"""
-
 from flask import current_app, Flask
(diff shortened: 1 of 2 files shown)


superset/__init__.py
```diff
@@ -14,8 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-"""Package's main module!"""
```
mistercrunch, line 17: I removed this silly string to trigger CI as it only runs certain steps based on which folders/files have changed

eschutho (approved):

PR #29109, opened 2024-06-06
feat: OAuth2 client initial work

betodealmeida:
### SUMMARY

We're planning to introduce OAuth2 client management to Superset, so the clients can be configured on the app, and assigned to multiple databases.

This PR is a first step in that direction, allowing client configuration to be stored in a the database `encrypted_extra` parameter. This allows us to add support for OAuth2 to more DB engine specs before starting the work on a full UI solution.

In addition to reading the client configuration from `encrypted_extra` this PR also refactors the logic for triggering OAuth2, introducing a new method called `needs_oauth2` to the DB engine spec. This is needed to give more flexibility to account for different OAuth2 providers.

The PR also introduces a new trigger for OAuth2, when the `raw_connection` fails due to needing a personal user token.

N/A

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -130,9 +130,7 @@ builtin_time_grains: dict[str | None, str] = {
 }
 
 
-class TimestampExpression(
-    ColumnClause
-):  # pylint: disable=abstract-method, too-many-ancestors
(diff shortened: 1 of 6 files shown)


eschutho (approved):
LGTM

PR #29119, opened 2024-06-08
fix(mixed-timeseries-plugin): Second query stacks stacked on top of first query series

kgabryje:
### SUMMARY
In Mixed Series chart, when user sets "Stack" option to true for both queries, the series of the second query are stacked on top of the series of the first query. The reason for that is that both queries could have the same stack ids.
This PR introduces `stackIdSuffix`, which adds `_a` and `_b` suffixes to stack ids of queries A and B to ensure that they're unique.

In the screenshots - the bars represent the second query. Before, they were stacked on top of the series from 1st query.

Before:

image

After:

image

### TESTING INSTRUCTIONS
1. Create Mixed Series chart
2. Set Stack to true for both queries
3. Make sure that the series of both queries are stacked separately

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
@@ -151,6 +151,7 @@ export function transformSeries(
     areaOpacity?: number;
     seriesType?: EchartsTimeseriesSeriesType;
     stack?: StackType;
+    stackIdSuffix?: string;
     yAxisIndex?: number;
     showValue?: boolean;
     onlyTotal?: boolean;
@@ -179,6 +180,7 @@ export function transformSeries(
     areaOpacity = 1,
     seriesType,
     stack,
+    stackIdSuffix,
     yAxisIndex = 0,
     showValue,
     onlyTotal,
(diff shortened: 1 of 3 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
```diff
@@ -225,6 +227,9 @@ export function transformSeries(
   } else if (stack && isTrend) {
     stackId = getTimeCompareStackId(forecastSeries.type, timeCompare, name);
   }
+  if (stackId && stackIdSuffix) {
+    stackId += stackIdSuffix;
```
villebro, line 231: maybe a nit, but a cybersec person once said that you should always use `\n` as the separator in concatenated keys in cases where it's unlikely to appear naturally in the original keys. I'm not sure there's lots of collision risk here, but just in case I'd probably use `\n` instead of `_`.

villebro (approved):
LGTM with a general comment about composite keys.

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
kgabryje: done 🙂

PR #29163, opened 2024-06-10
refactor(sqllab): migrate share queries via kv by permalink

justinpark:
### SUMMARY
This PR changes the existing `SHARE_QUERIES_VIA_KV_STORE` feature to store data using the permalink API instead of the KV store. This is one of the prerequisites of [Remove old KV store endpoint and its associated asset](https://github.com/orgs/apache/projects/345/views/1?pane=issue&itemId=59020789).

https://github.com/user-attachments/assets/03b39a58-87b2-450a-81d0-422d9a5e2eec

### TESTING INSTRUCTIONS
Go to SQL Lab and then click "Copy Link"

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/SqlLab/components/ShareSqlLabQuery/index.tsx
```diff
@@ -92,7 +96,7 @@ const ShareSqlLabQuery = ({
   };
   const getCopyUrl = (callback: Function) => {
     if (isFeatureEnabled(FeatureFlag.ShareQueriesViaKvStore)) {
-      return getCopyUrlForKvStore(callback);
+      return getCopyUrlForPermalink(callback);
```
villebro, line 99: Shouldn't this feature be enabled by default? Now the permalink feature is only available if the `SHARE_QUERIES_VIA_KV_STORE` feature flag is enabled:

image

I feel we could now remove the feature flag, and make this the default behavior.

In addition, I think it would be a good idea to create the permalink based on the same query snippet that gets executed when running the query - Currently the entire editor gets persisted into the permalink, even if only a specific portion of the editor is selected.

villebro (commented):
Great improvement, this is long overdue! I suggest removing the `SHARE_QUERIES_VIA_KV_STORE` feature flag, and making the permalink only store the selected portion of the editor. Other than that this looks great!

superset-frontend/src/SqlLab/components/ShareSqlLabQuery/index.tsx
justinpark: > I feel we could now remove the feature flag, and make this the default behavior.

@villebro as @michael-s-molina mentioned, we'll remove the feature flag during the 5.0 proposal work.

> In addition, I think it would be a good idea to create the permalink based on the same query snippet that gets executed when running the query - Currently the entire editor gets persisted into the permalink, even if only a specific portion of the editor is selected.

Clipping only specific executed sections into a permalink might cause confusion for some users. Instead, I'm thinking about a method where, if an area is selected, it automatically highlights the selection in the editor.

justinpark:
@villebro @michael-s-molina could you take a look?

michael-s-molina:
Can you add support for old saved query URLs?

justinpark:
> Can you add support for old saved query URLs?

Done. @michael-s-molina PTAL

superset-frontend/src/SqlLab/components/ShareSqlLabQuery/index.tsx
justinpark: > I feel we could now remove the feature flag, and make this the default behavior.

@villebro Done. permalink sets to default now.

RESOURCES/FEATURE_FLAGS.md
```diff
@@ -101,3 +101,4 @@ These features flags currently default to True and **will be removed in a future
 - DRILL_TO_DETAIL
 - ENABLE_JAVASCRIPT_CONTROLS
 - KV_STORE
+- SHARE_QUERIES_VIA_KV_STORE
```
michael-s-molina, line 104:
```suggestion
```

UPDATING.md
```diff
@@ -31,6 +31,7 @@ assists people when migrating to a new version.
 - [31262](https://github.com/apache/superset/pull/31262) NOTE: deprecated `pylint` in favor of `ruff` as our only python linter. Only affect development workflows positively (not the release itself). It should cover most important rules, be much faster, but some things linting rules that were enforced before may not be enforce in the exact same way as before.
 - [31173](https://github.com/apache/superset/pull/31173) Modified `fetch_csrf_token` to align with HTTP standards, particularly regarding how cookies are handled. If you encounter any issues related to CSRF functionality, please report them as a new issue and reference this PR for context.
 - [31385](https://github.com/apache/superset/pull/31385) Significant docker refactor, reducing access levels for the `superset` user, streamlining layer building, ...
+- [29163](https://github.com/apache/superset/pull/29163) Deprecated the `SHARE_QUERIES_VIA_KV_STORE` feature flag
```
michael-s-molina, line 34: I think this needs to be updated to reflect the last state. It should also say that we will stop supporting the old links in 6.0.

superset-frontend/src/types/bootstrapTypes.ts
```diff
@@ -164,6 +164,7 @@ export interface BootstrapData {
     dashboard_id: string;
   };
   requested_query?: JsonObject;
+  permalink?: string;
```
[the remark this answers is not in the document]
justinpark: Replace the bootstrap logic by parsing the url path

UPDATING.md
```diff
@@ -32,7 +33,8 @@ assists people when migrating to a new version.
...
 - [31173](https://github.com/apache/superset/pull/31173) Modified `fetch_csrf_token` to align with HTTP standards, particularly regarding how cookies are handled. If you encounter any issues related to CSRF functionality, please report them as a new issue and reference this PR for context.
 - [31385](https://github.com/apache/superset/pull/31385) Significant docker refactor, reducing access levels for the `superset` user, streamlining layer building, ...
-- [31503](https://github.com/apache/superset/pull/31503) Deprecating python 3.9.x support, 3.11 is  now the recommended version and 3.10 is still supported over the Superset 5.0 lifecycle.
+- [31503](https://github.com/apache/superset/pull/31503) Deprecating python 3.9.x support, 3.11 is now the recommended version and 3.10 is still supported over the Superset 5.0 lifecycle.
+- [29163](https://github.com/apache/superset/pull/29163) The /kv/ API used in the SQL Lab share link has been deprecated and replaced with a permalink. The legacy share link (i.e., /sqllab/?id=xxxx) will be temporarily available in 5.0 but will no longer be supported in 6.0
```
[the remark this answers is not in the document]
justinpark: Done

michael-s-molina (approved):
LGTM. Thanks for addressing the comments @justinpark!

LevisNgigi:
Hey, @justinpark I just fetched this to local and when I try to copy the link in Sqllab i get the error, "Request URL:
http://localhost:9000/api/v1/sqllab/permalink
Request Method:
POST
Status Code:
403 FORBIDDEN" could I be missing something? I'm logged in as an admin user.

PR #29166, opened 2024-06-11
chore: enable ruff lint rule TRY201 and B904 to improve `raise` stack traces

mistercrunch:
When intercepting and re-raising an error in python, it's almost always better to simply `raise` over raising the exception alias ie: `raise ex`. The main reason is that the stack trace often (always?) gets truncated in the process, losing track of where the error occurred.

Similarly, when intercepting and re-raising another exception, it's better to `raise AnotherException() from ex` (using  `from`), so I'm fixing and enforcing this as well.

I found this while debugging and dealing with incomplete stack traces, and thought it be a good idea to fix and enforce using ruff's TRY201 and B904

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -435,7 +435,14 @@ target-version = "py310"
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-select = ["E4", "E7", "E9", "F"]
+select = [
+    "B904",
+    "E4",
+    "E7",
+    "E9",
+    "F",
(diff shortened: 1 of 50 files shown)


mistercrunch:
Summoning some of our backend / python folks - @betodealmeida @dpgaspar @john-bodley , what do you think?

giftig (approved):

giftig:
I think there might be cases where you legitimately don't want to use `from` because you don't care about the original exception, but it's probably not very often. But a bare `raise` is always better than reraising the original exception you just caught imo, for the reasons you mentioned.

villebro (approved):
TIL

john-bodley (approved):

mistercrunch:
> TIL

yeah I didn't know that until I was like "where's the rest of my stacktrace!?!?! how did we get here?"

PR #29229, opened 2024-06-12
fix: Improving handling for tag relationship when deleting assets v2

Vitor-Avila:
### SUMMARY
This is a follow up for https://github.com/apache/superset/pull/29117, since it was confirmed on [this comment](https://github.com/apache/superset/pull/26654#issuecomment-2162050230) that the issue still persists.

I'm updating these relationships to `viewonly=True`, since there's already an `after_delete` hook in place to delete tag association after an asset is deleted - [reference](https://github.com/apache/superset/blob/master/superset/tags/models.py#L276-L290). With that in mind, handling the cascade deletion in the relationship is redundant.

Long-term, we could delete this hook and fix the relationships for a proper cascading deletion -- I'll look into that in a following PR (need to make sure that removing this hook won't impact any other logic in the application).

No UI changes.

### TESTING INSTRUCTIONS
1. Make sure you have a dashboard and a chart with the same ID.
2. Assign a tag for both assets. Also add owners to both assets.
3. Validate that you're able to delete one of the assets.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/dashboard.py
+++ b/superset/models/dashboard.py
@@ -155,7 +155,7 @@ class Dashboard(AuditMixinNullable, ImportExportMixin, Model):
         primaryjoin="and_(Dashboard.id == TaggedObject.object_id, "
         "TaggedObject.object_type == 'dashboard')",
         secondaryjoin="TaggedObject.tag_id == Tag.id",
-        viewonly=True,
+        viewonly=True,  # cascading deletion already handled by superset.tags.models.ObjectUpdater.after_delete
     )
     published = Column(Boolean, default=False)
     is_managed_externally = Column(Boolean, nullable=False, default=False)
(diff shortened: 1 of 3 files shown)


Vitor-Avila:
@mistercrunch @john-bodley let me know your thoughts with this approach

mistercrunch (changes requested):
It'd be great to capture this as comment, something like
```
viewonly=True,  # cascading handled by tag.models.ObjectUpdater.after_delete
```
for all 3 intances plz. As someone reading the code, it'd be a head scratcher, and I have to say that while I worked on this bug I never knew this ObjectUpdater existed...

mistercrunch (approved):

PR #29231, opened 2024-06-12
feat: add new SQLLAB_FORCE_RUN_ASYNC feature flag

mistercrunch:
When setting up a celery backend to run asynchronous workloads, the administrator can set any given database connection to use (or not use) that backend. This PR introduced a new feature flag SQLLAB_FORCE_RUN_ASYNC (False by default) that disregards the database configuration, and forces SQL Lab to use the async backend.

NOTE: superset/sqllab/sqllab_execution_context.py didn't have unit tests so I wrote started a new module to run basic tests around it.

### DIFF
--- a/superset-frontend/src/components/InfoTooltip/index.tsx
+++ b/superset-frontend/src/components/InfoTooltip/index.tsx
@@ -20,9 +20,10 @@
 import { styled, useTheme } from '@superset-ui/core';
 import { Tooltip } from 'src/components/Tooltip';
 import Icons from 'src/components/Icons';
+import React from 'react';
 
 export interface InfoTooltipProps {
-  iconStyle?: string;
+  iconStyle?: React.CSSProperties;
   tooltip: string;
   placement?:
     | 'bottom'
--- a/superset-frontend/src/components/InfoTooltip/index.tsx
+++ b/superset-frontend/src/components/InfoTooltip/index.tsx
@@ -73,8 +73,9 @@ export default function InfoTooltip({
   viewBox = '0 -1 24 24',
 }: InfoTooltipProps) {
   const theme = useTheme();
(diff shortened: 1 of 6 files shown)


john-bodley:
@mistercrunch could you provide more details with regards to the motivation for this change? It likely introduces a degraded UX as admins likely could be misinformed that a given database is running in sync mode whereas in actuality (when the feature flag is enabled) the database is using async.

mistercrunch:
The requirement is we were considering enforcing this for some of our customers at Preset, but I get your point and I agree that I should probably disable the UI element in the database connection form.

mistercrunch:
@john-bodley I added frontend cues ->
Screenshot 2024-06-18 at 12 09 38 PM

superset-frontend/src/components/InfoTooltip/index.tsx
```diff
@@ -68,12 +65,18 @@ const defaultColor = 'rgba(0,0,0,0.9)';
...
   trigger = 'hover',
   overlayStyle = defaultOverlayStyle,
   bgColor = defaultColor,
   viewBox = '0 -1 24 24',
 }: InfoTooltipProps) {
+  const theme = useTheme();
+  const alteredIconStyle = { ...iconStyle };
+  if (alteredIconStyle.color === undefined) {
+    alteredIconStyle.color = theme.colors.grayscale.base;
+  }
```
geido, line 79:
```suggestion
  const alteredIconStyle = { 
    ...iconStyle,
    color:  iconStyle.color || theme.colors.grayscale.base,
  };
```

geido (approved):
Sorry my suggestion wasn't properly indented. However, this looks good!

mistercrunch:
No worries, a quick --fix took care of it

PR #29233, opened 2024-06-12
refactor(sqllab): nonblocking delete query editor

justinpark:
### SUMMARY
Following up #24539 
This commit modifies the `removeQueryEditor` function to asynchronously handle the process of deleting a queryEditor in localStorage in the same way as in non-persistent mode and periodically post deleted items to persistence via AutoSync's `deleteSqlEditor`.

This commit made the following changes:
- In persistence mode, deleting a tab will no longer block the removal until it is fetched. Instead, the tab will be immediately discarded and stored the deleted ids in the local storage. (`destroyedQueryEditors`)
- `removeQueryEditor` has been modified to always update to localStorage in the same way as in non-persistence mode
- EditorAutoSync will execute `deleteSqlEditor` periodically and then remove the completed item from `destroyedQueryEditors` on success
- Add the logic of storing localStorage `destroyedQueryEditors` for persistence mode 

Before:

https://github.com/apache/superset/assets/1392866/a0c393dd-7d1c-474f-a7fb-8e88a61b3aea

After:

https://github.com/apache/superset/assets/1392866/0855d553-1c5f-4718-bdc0-372ca412a38c

### TESTING INSTRUCTIONS

set SQLLAB_BACKEND_PERSISTENCE on
go to sqllab and then close each tab quickly

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/EditorAutoSync/index.tsx
+++ b/superset-frontend/src/SqlLab/components/EditorAutoSync/index.tsx
@@ -29,12 +29,14 @@ import {
 import {
   useUpdateCurrentSqlEditorTabMutation,
   useUpdateSqlEditorTabMutation,
+  useDeleteSqlEditorTabMutation,
 } from 'src/hooks/apiResources/sqlEditorTabs';
 import { useDebounceValue } from 'src/hooks/useDebounceValue';
 import {
   syncQueryEditor,
   setEditorTabLastUpdate,
   setLastUpdatedActiveTab,
+  clearDestoryedQueryEditor,
 } from 'src/SqlLab/actions/sqlLab';
 import useEffectEvent from 'src/hooks/useEffectEvent';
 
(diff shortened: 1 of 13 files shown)


superset-frontend/src/SqlLab/components/EditorAutoSync/index.tsx
```diff
@@ -119,6 +126,22 @@ const EditorAutoSync: FC = () => {
     }
   });
 
+  const syncDeletedQueryEditor = useEffectEvent(() => {
```
michael-s-molina, line 129: Maybe use `async/await` here? I think it's more readable so feel free to apply the same pattern everywhere.

michael-s-molina (approved):

superset-frontend/src/SqlLab/components/EditorAutoSync/index.tsx
justinpark: This structure sends delete requests simultaneously, so if implemented with await, additional `Promise.all` handling might be required. Therefore, I intend to keep the current structure.

PR #24969, opened 2023-08-11
chore(dao/command): Add transaction decorator to try to enforce "unit of work"

john-bodley:
### SUMMARY

This is a PR I've had on the back-burner for many months, but have struggled with on numerous occasions—often in part due to the flakey/delicate tests (and their associated frameworks). The initial desire was to fulfill the approach outlined in [[SIP-99B] Proposal for (re)defining a "unit of work"](https://github.com/apache/superset/issues/25108), but alas I failed, in part due to the challenges trying to untangle Superset logic which inherently is not overly conducive to adhering to the construct that a command should serve as a "unit of work".

Why is that? It's complicated, but asynchronous logic does not help given that a Celery task running within the confines of another command needs to read a previously committed state given the `READ COMMITTED` isolation level. Issues like this could likely be overcome by having two commands—prepare and execute—as opposed to a single execute command. 

The TL;DR is this PR should likely be interpreted as the first phase of SIP-99B. The general framework holds, i.e., DAOs no longer commit and a `transaction` decorator is used to wrap any command which perform either an `INSERT`, `UPDATE`, or `DELETE`.

Finally, I apologize for the size of the PR. I struggled to downside the footprint, but once you start enforcing that DAOs should not commit, then the files which touched begins to snowball.

Regrettably my time (for now) working on Apache Superset is likely drawing to a close, so for completeness I thought there was merit in [sharing the incremental diff](https://github.com/john-bodley/superset/compare/john-bodley--dao-nested-session...john-bodley:superset:john-bodley--dao-nested-session-phase-2?expand=1) for what I was hoping to achieve in case @michael-s-molina @villebro et al. wanted to carry the baton on.

### TESTING INSTRUCTIONS

CI.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/25108
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29235, opened 2024-06-12
chore: Remove the need for explicit bubble up of certain exceptions

john-bodley:
### SUMMARY

There are instances where we bubble up exceptions, however this is akin to simply not catching it, since any uncaught exception is propagated up the call stack.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -442,10 +442,15 @@ class ChartDataRestApi(ChartRestApi):
     def _create_query_context_from_form(
         self, form_data: dict[str, Any]
     ) -> QueryContext:
+        """
+        Create the query context from the form data.
+
+        :param form_data: The chart form data
+        :returns: The query context
+        :raises ValidationError: If the request is incorrect
+        """
+
         try:
             return ChartDataQueryContextSchema().load(form_data)
         except KeyError as ex:
             raise ValidationError("Request is incorrect") from ex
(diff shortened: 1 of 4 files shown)


mistercrunch:
Right. I considered doing the same during https://github.com/apache/superset/pull/29166, but thought there might have some reason for the code to point out the "likely exception" in the code, maybe to allow future handling or something. I decided to leave them as they were since my goal was enabling the lint rule and minimizing the code changes.

mistercrunch (approved):
Approving as those lines effectively do nothing - except maybe for calling out semantically in the code the exceptions that are likely to happen during the try block.

john-bodley:
@mistercrunch the catalyst for the change is I was rebasing https://github.com/apache/superset/pull/24969 (a longstanding PR I've been wrangling with failing tests for a while).

michael-s-molina:
> thought there might have some reason for the code to point out the "likely exception" in the code, maybe to allow future handling or something.

I think for this case, the best approach is to document the exceptions in docstrings:

```
"""
    Perform some operation.

    Raises:
        ValueError: If the input is invalid.
        FileNotFoundError: If a required file is not found.
        CustomException: If a specific condition is not met.
    """
```

michael-s-molina (approved):

john-bodley:
@michael-s-molina I addressed your comment.

PR #29241, opened 2024-06-13
feat: Support a dynamic minimum interval for alerts and reports

Vitor-Avila:
### SUMMARY
This PR adds the ability to control the `ALERT_MINIMUM_INTERVAL` and/or `REPORT_MINIMUM_INTERVAL` configs dynamically (rather than having it statically defined during runtime). This is useful in case the value should be dynamically changed/evaluated on demand (you might want to control it externally).

No UI changes.

### TESTING INSTRUCTIONS
Added unit tests to make sure the config supports both an `integer` and a `callable`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


betodealmeida (approved):
Nice!

One comment I had, I wonder if we should have some parameters passed to the function, even if your examples don't need them. This way we don't need to change the signature in the future. Eg, we could pass the database, so that databases where query are more expensive would have a higher minimum interval.

Vitor-Avila:
@betodealmeida that makes perfect sense! I added `**kwargs` to the example method so that if in the future additional work is done to pass parameters, this shouldn't break any existing setup.

PR #29242, opened 2024-06-13
feat: Improves the Drill By feature

michael-s-molina:
### SUMMARY
This PR improves the drill by feature by:
- Automatically focusing the menu search field
- Adding virtualization to the options list to handle many items
- Diminishing the request payload by selecting only required data
- Adding an extension point to allow custom option loaders that might filter the options list depending on the chart's form data.

https://github.com/apache/superset/assets/70410625/a7fc350d-b047-4c13-be22-47b5e8c1932d

### TESTING INSTRUCTIONS
- Check that the menu can handle many items efficiently
- Check that the loader can be overridden
- Check the improved request

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
+++ b/superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
@@ -323,7 +323,7 @@ export const DrillByMenuItems = ({
             <List
               width="100%"
               height={SUBMENU_HEIGHT}
-              itemSize={40}
+              itemSize={35}
               itemCount={filteredColumns.length}
               itemData={{ columns: filteredColumns, ...rest }}
               overscanCount={20}
--- a/superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
+++ b/superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
@@ -327,7 +327,7 @@ export const DrillByMenuItems = ({
               `}
             >
               <AutoSizer>
-                {({ height, width }) => (
+                {({ height, width }: { height: number; width: number }) => (
                   <List
                     width={width}
                     height={height}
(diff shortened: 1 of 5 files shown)


superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
```diff
@@ -119,51 +157,59 @@ export const DrillByMenuItems = ({
   const verboseMap = useVerboseMap(dataset);
 
   useEffect(() => {
+    async function loadOptions() {
+      const datasetId = +formData.datasource.split('__')[0];
```
justinpark, line 161: While this method seems convenient, it might cause misunderstandings for those not familiar with JavaScript. Therefore, it would be better to apply a more explicit number conversion function.

```suggestion
      const datasetId = Number(formData.datasource.split('__')[0]);
```

superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
```diff
@@ -253,20 +323,23 @@ export const DrillByMenuItems = ({
...
-                  onClick={e => handleSelection(e, column)}
-                >
-                  {column.verbose_name || column.column_name}
-                </MenuItemWithTruncation>
-              ))}
+              <AutoSizer>
+                {({ height, width }: { height: number; width: number }) => (
+                  <List
+                    width={width}
+                    height={height}
```
justinpark, line 333: Since the height is already fixed, what do you think about locking the height value into the same variable to improve resize performance?
Additionally I would also like to avoid using the autosizer if we can set the width to a specific size, since there are numerous callback functions linked to resize events within the charts on the current dashboard.

```suggestion
                {({ width }: { width: number }) => (
                  <List
                    width={width}
                    height={MAX_SUBMENU_HEIGHT}
```

michael-s-molina: That makes total sense. Great suggestion.

justinpark (approved):
LGTM!

PR #29248, opened 2024-06-14
fix(sqllab): Do not strip comments when executing SQL statements

john-bodley:
### SUMMARY

Man, the SQL Lab execution/parsing logic is a real rabbit warren—so many methods and diversions to execute SQL.

https://github.com/apache/superset/pull/27725 _attempted_ to remove stripping comments when executing queries in SQL Lab, however it likely wasn't suffice. https://github.com/apache/superset/pull/28363 (which also preventing stripping of comments in Explore) remedied this, however it was later reverted in https://github.com/apache/superset/pull/28567.

The original issue with https://github.com/apache/superset/pull/27725 is it seems like the SQL is pre-processed/rendered in the command prior to the typical SQL Lab execution logic via Celery et al. This PR simply ensures that comments are never stripped within the context of SQL Lab. This change shouldn't cause the issue surfaced in https://github.com/apache/superset/pull/28567.

### TESTING INSTRUCTIONS

CI. The relevant tests were updated using logic from https://github.com/apache/superset/pull/28363.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sqllab/query_render.py
+++ b/superset/sqllab/query_render.py
@@ -60,7 +60,6 @@ class SqlQueryRenderImpl(SqlQueryRender):
 
             parsed_query = ParsedQuery(
                 query_model.sql,
-                strip_comments=True,
                 engine=query_model.database.db_engine_spec.engine,
             )
             rendered_query = sql_template_processor.process_template(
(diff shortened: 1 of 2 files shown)


rusackas:
> a real rabbit warren
... I think this is the first time I've ever heard someone use the correct nomenclature with this expression.

michael-s-molina (approved):

PR #29257, opened 2024-06-14
fix: Catalog with restricted permissions produces an error during database connection

geido:
### SUMMARY
When connecting to a Postgres database, a generic database error occurred. When fetching schemas via catalogs, restricted schemas triggered an error, causing the error to propagate to the UI, even though the database connection was correctly stored.

This PR implements a soft fail for the above scenario, preventing the error from propagating to the UI.

N.A.

### TESTING INSTRUCTIONS
1. Create a Postgres database connection using. a SQLAlchemy URI
2. The connection should be created successfully with no errors

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


betodealmeida (approved):
I think this makes sense. The problem here is that the `get_catalog_names` method from the DB engine spec currently returns **all** catalog names, regardless of if they can be accessed with the current set of credentials. When I wrote it I assumed that the database would filter out catalog names that the user didn't have access to, but apparently that's not the case.

We probably want to return only catalogs that are accessible, but the problem is that will probably require connecting to the database for each one, which will take a long time. Not sure if there's a clean way to fix this.

john-bodley:
@geido would you mind adding some tests (if possible)? I gather the motivation for this PR was a recent regression which likely was missed due lack of tests.

geido:
> @geido would you mind adding some tests (if possible)? I gather the motivation for this PR was a recent regression which likely was missed due lack of tests.

Sorry, I just noticed this now. There will be a follow-up PR to fix the source of the problem by checking the catalog permissions which will include all necessary test cases.

PR #26875, opened 2024-01-30
chore: make TS enums strictly PascalCase

villebro:
### SUMMARY
TypeScript enum names and keys should always be PascalCase, which is the commonly accepted best practice (see the official docs: https://www.typescriptlang.org/docs/handbook/enums.html). This PR does the following:
- Adds ESLint rules to enforce both PascalCase for the enum names and their respective keys.
- Fix the whole FS codebase, including `superset-ui` packages/plugins. The enum values have been left in tact so as to not cause a breaking change.

Note, that I've made a few exceptions, where deviating from pascal case makes more sense, especially the locales and currencies, where it felt dangerous and less intuitive to move to pascal case.

### TESTING INSTRUCTIONS
1. Pull the codebase, try to rename a TS enum to snake case, and notice how the linter complains.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29260, opened 2024-06-14
fix: Custom SQL filter control

michael-s-molina:
### SUMMARY
This PR fixes a number of problems with the Custom SQL filter control:
- There was a layout flick when changing the clause type
- The algorithm that detects if something changes was not considering the filter clause. This was blocking users from changing only the filter clause (not the SQL) and saving the filter.
- There were invalid values for the clause select which was leading to wrong behavior. The `Where` clause was actually being applied as `HAVING`. This bug was introduced in https://github.com/apache/superset/pull/26875.

This PR also reimplements the tests using RTL and creates a helper function to interact with the `Select` component in tests.

https://github.com/apache/superset/assets/70410625/f7c06f66-7182-4331-ba32-966a2e4b5c99

https://github.com/apache/superset/assets/70410625/b1792abc-a048-4c9e-83c9-b7bc7e3ce61c

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/explore/components/controls/FilterControl/AdhocFilterEditPopoverSqlTabContent/index.jsx
```diff
@@ -95,11 +90,6 @@ export default class AdhocFilterEditPopoverSqlTabContent extends Component {
   render() {
     const { adhocFilter, height, options } = this.props;
 
-    const clauseSelectProps = {
```
michael-s-molina, line 98: Moved to the component declaration.

justinpark (approved):
LGTM! Thanks for the RTL migration too

PR #29274, opened 2024-06-17
chore: trigger CI jobs on all release-related branches

mistercrunch:
### SUMMARY

Currently, we trigger the bulk of our CI on all PRs, and on push to the `master` and release branches.

Release branches currently are defined as `[0-9].[0-9]` meaning it will work for `3.5` and `10.10`, but not for `3.1.1`, `3.0-beta2` or whatever other release branch that don't fit the "minor" release pattern.

Not I'm extending this to `[0-9].[0-9]*`. As a glob that mean it'll run for practically anything that loosely looks like a release branch.

The goal is to make it easier for people working in forks, whether private or public to trigger CI while doing their own release management. Currently if you want to trigger CI, you pretty much to either open a dummy PR, or push to a branch that fits the current pattern, which doesn't allow for flexibility around branch naming, which is key for release management.

This PR should not affect the repo or release process on this repo, it's just about making it easier to trigger CI on forks.

### DIFF
--- a/.github/workflows/check_db_migration_confict.yml
+++ b/.github/workflows/check_db_migration_confict.yml
@@ -5,7 +5,7 @@ on:
       - "superset/migrations/**"
     branches:
       - "master"
-      - "[0-9].[0-9]"
+      - "[0-9].[0-9]*"
   pull_request:
     paths:
       - "superset/migrations/**"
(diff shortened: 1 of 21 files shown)


sadpandajoe (approved):
Looks good to me

michael-s-molina (approved):

rusackas (approved):

PR #29279, opened 2024-06-17
fix(sqllab): excessive API calls for schemas

justinpark:
### SUMMARY
Currently, the refresh logic for schemas is being triggered excessively due to unnecessary dependencies, causing it to be called every time a tab is switched.

This commit addresses the issue by modifying the logic to execute only when a trigger is needed.

Before:

https://github.com/apache/superset/assets/1392866/b0ba7b17-1c7a-4d0e-aa9b-1e2fcaa5604b

After:

https://github.com/apache/superset/assets/1392866/a785f3fe-3d04-4c1b-a574-34783d6c496c

### TESTING INSTRUCTIONS

Switch the tabs between the same db selected tabs
Check the network tab and then only schemas called once

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/hooks/apiResources/catalogs.ts
+++ b/superset-frontend/src/hooks/apiResources/catalogs.ts
@@ -77,47 +77,28 @@ export function useCatalogs(options: Params) {
     },
   );
 
-  const handleOnSuccess = useEffectEvent(
-    (data: CatalogOption[], isRefetched: boolean) => {
-      onSuccess?.(data, isRefetched);
+  const fetchData = useEffectEvent(
+    (dbId: FetchCatalogsQueryParams['dbId'], forceRefresh = false) => {
+      if (dbId && (!result.currentData || forceRefresh)) {
+        trigger({ dbId, forceRefresh }).then(({ isSuccess, isError, data }) => {
+          if (isSuccess) {
+            onSuccess?.(data || EMPTY_CATALOGS, forceRefresh);
+          }
+          if (isError) {
+            onError?.();
+          }
+        });
+      }
     },
   );
 
-  const handleOnError = useEffectEvent(() => {
-    onError?.();
-  });
-
-  const resolver = useEffectEvent(() =>
-    result.currentData ? undefined : trigger({ dbId, forceRefresh: false }),
-  );
-
   const refetch = useCallback(() => {
-    if (dbId) {
-      trigger({ dbId, forceRefresh: true }).then(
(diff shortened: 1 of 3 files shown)


superset-frontend/src/hooks/apiResources/catalogs.ts
```diff
@@ -88,6 +87,10 @@ export function useCatalogs(options: Params) {
     onError?.();
   });
 
+  const resolver = useEffectEvent(() =>
```
michael-s-molina, line 90: There's a lot of duplicated code. Maybe something like this?

```
export function useCatalogs(options: Params) {
  const { dbId, onSuccess, onError } = options || {};
  const [trigger] = useLazyCatalogsQuery();
  const result = useCatalogsQuery(
    { dbId, forceRefresh: false },
    {
      skip: !dbId,
    },
  );

  const fetchData = useCallback(
    (forceRefresh: boolean) => {
      trigger({ dbId, forceRefresh }).then(({ isSuccess, isError, data }) => {
        if (isSuccess) {
          onSuccess?.(data || EMPTY_CATALOGS, forceRefresh);
        }
        if (isError) {
          onError?.();
        }
      });
    },
    [dbId, onError, onSuccess, trigger],
  );

  const refetch = useCallback(() => {
    if (dbId) {
      fetchData(true);
    }
  }, [dbId, fetchData]);

  useEffect(() => {
    if (dbId && !result.currentData) {
      fetchData(false);
    }
  }, [dbId, fetchData, result.currentData]);

  return {
    ...result,
    refetch,
  };
}
```

superset-frontend/src/hooks/apiResources/schemas.ts
```diff
@@ -92,20 +91,24 @@ export function useSchemas(options: Params) {
     onError?.();
   });
 
+  const resolver = useEffectEvent(() =>
```
michael-s-molina, line 94: Same as before.

michael-s-molina (commented):
Thanks for the PR @justinpark!

superset-frontend/src/hooks/apiResources/catalogs.ts
justinpark: That's a good suggestion. One thing that including result.currentData in the dependencies array of useEffect can cause unintended fetchData calls. I'll change it to handle this within the fetchData function.

michael-s-molina (approved):

PR #29286, opened 2024-06-18
fix(key-value): use flush instead of commit

villebro:
### SUMMARY
In the key value commands, we are often returning the keys from newly generated ORM objects, which are persisted in the `key_value` table. As we're currently using `commit()` to persist the objects prior to reading the keys, the session is reset, requiring a new session for fetching the keys. This can cause issues when distributing metastore reads to read replicas, as the data may sometimes not be available, causing an `ObjectDeletedError` exception during `CreateKeyValueCommand.run()`:

```
sqlalchemy.orm.exc.ObjectDeletedError: Instance '<KeyValueEntry at 0x7fe508293280>' has been deleted, or its row is otherwise not present.
```

A simple fix is to use `flush()`, which will ensure the same session is used, as stated in [SIP-99A](#25107):

> It is worth noting that objects within the session, in either pending* or persisted state, can be queried albeit not having been committed to the database. We have a tendency to over commit (be that in the code or tests) resulting in fractured/partial atomic units. Typically working with objects in a persisted state is sufficient.

As we're now proposing to replace `commit()` with `flush()` in the key value commands, the session will need to be explicitly committed once the full unit of work is done. For example, for the create/upsert commands, we will now only commit after the last task finishes, ensuring atomicity:
1. delete expired entries
2. insert new entry
3. commit

However, there are instances where we explicitly need commits to happen during a single request lifecycle. For instance, the Key Value Distributed Lock needs to commit the lock to the metastore so other workers can observe it for the duration of the lock event. The same applies to the Metastore cache, where every set/add/delete operation should commit to the cache. For this reason we're adding explicit `commit()` calls both in the Metastore Cache and Lock context manager to ensure the values are persisted in the db during execution flow.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/dashboard/permalink/create.py
+++ b/superset/commands/dashboard/permalink/create.py
@@ -18,6 +18,7 @@ import logging
 
 from sqlalchemy.exc import SQLAlchemyError
 
+from superset import db
 from superset.commands.dashboard.permalink.base import BaseDashboardPermalinkCommand
 from superset.commands.key_value.upsert import UpsertKeyValueCommand
 from superset.daos.dashboard import DashboardDAO
@@ -62,6 +63,7 @@ class CreateDashboardPermalinkCommand(BaseDashboardPermalinkCommand):
                 codec=self.codec,
             ).run()
             assert key.id  # for type checks
+            db.session.commit()
             return encode_permalink_key(key=key.id, salt=self.salt)
         except KeyValueCodecEncodeException as ex:
             raise DashboardPermalinkCreateFailedError(str(ex)) from ex
--- a/superset/utils/lock.py
+++ b/superset/utils/lock.py
@@ -24,9 +24,10 @@ from contextlib import contextmanager
 from datetime import datetime, timedelta
 from typing import Any, cast, TypeVar, Union
 
+from superset import db
 from superset.exceptions import CreateKeyValueDistributedLockFailedException
 from superset.key_value.exceptions import KeyValueCreateFailedError
-from superset.key_value.types import KeyValueResource, PickleKeyValueCodec
+from superset.key_value.types import KeyValueResource, JsonKeyValueCodec
 from superset.utils import json
 
 LOCK_EXPIRATION = timedelta(seconds=30)
@@ -53,6 +54,10 @@ def serialize(params: dict[str, Any]) -> str:
     return json.dumps(params)
 
 
+def get_key(namespace: str, **kwargs: Any) -> uuid.UUID:
+    return uuid.uuid5(uuid.uuid5(uuid.NAMESPACE_DNS, namespace), serialize(kwargs))
+
+
 @contextmanager
 def KeyValueDistributedLock(  # pylint: disable=invalid-name
     namespace: str,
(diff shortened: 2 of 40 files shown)


superset/utils/lock.py
```diff
@@ -77,21 +82,23 @@ def KeyValueDistributedLock(  # pylint: disable=invalid-name
...
 
-    key = uuid.uuid5(uuid.uuid5(uuid.NAMESPACE_DNS, namespace), serialize(kwargs))
+    key = get_key(namespace, **kwargs)
     logger.debug("Acquiring lock on namespace %s for key %s", namespace, key)
     try:
         DeleteExpiredKeyValueCommand(resource=KeyValueResource.LOCK).run()
         CreateKeyValueCommand(
             resource=KeyValueResource.LOCK,
-            codec=PickleKeyValueCodec(),
+            codec=JsonKeyValueCodec(),
```
villebro, line 91: This is fixed in a separate PR (I had to pull in the change here as the test needed to be updated): #29285

superset/commands/dashboard/permalink/create.py
```diff
@@ -62,6 +63,7 @@ def run(self) -> str:
                 codec=self.codec,
             ).run()
             assert key.id  # for type checks
+            db.session.commit()
```
john-bodley, line 66: Does the `UpsertKeyValueCommand` not commit? Note that _hopefully_ (if merged) that https://github.com/apache/superset/pull/24969 should remove the need to have to commit in various places given it violates the "unit of work" philosophy.

villebro: Here I'm proposing to not commit in these commands, as we have cases where we want to chain multiple commands together. But if we consider each Key Value command a unit of work, then we should naturally commit there.

nytai: i think there's a case to be made for both. Maybe we can have default behavior be to commit but provide an option to skip for use cases where there's a command chaining multiple sub-commands? It would be inconsistent for some commands to commit while others requiring the caller to commit.

villebro: While this has previously been considered an antipattern (=having `commit: bool` flag in these types of methods or similar), I'm personally also kind of leaning in that direction. This would make it possible to use all existing commands both as full units of work, or as part of a bigger chain, with minimal code duplication. Thoughts @john-bodley ? Also pinging @michael-s-molina as you've worked on these components in the past.

john-bodley: SIP-99B will handle the chaining of commands (if necessary) via nested sessions where only the outermost session commits.

I think this logic is fine for now, though @villebro some of it maybe updated if/when I get my SIP-99B through.

john-bodley (approved):

superset/commands/dashboard/permalink/create.py
michael-s-molina: > This would make it possible to use all existing commands both as full units of work, or as part of a bigger chain, with minimal code duplication

We'll achieve that with nested transactions using [being_nested](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.begin_nested). Check https://github.com/apache/superset/pull/24969 for reference.

villebro: > We'll achieve that with nested transactions using [being_nested](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.begin_nested). Check https://github.com/apache/superset/pull/24969 for reference.

@michael-s-molina I read up on the docs I could find, and I agree, this should be an elegant solution to this dual use case.

@john-bodley it would be great if you could add a description to #24969 so we can start reviewing/testing it. I'm keen on getting this important refactor in, as it will have a profound impact on the general quality and perormance of the backend.

Issue #21234, opened 2022-08-29
Native filters doesn't update their cached values after Refresh Dashboard


muralikrishnankh:
The Native filters doesn't update their cached values after Refresh Dashboard even though the data is updated in the Database. The Native filters seems to be getting refreshed but the data doesn't get updated.

#### How to reproduce the bug

1. Install any version of Superset
2. Add some data to the connected Database. I have connected to Clickhouse DB.
3. Open Superset.
4. The Native filters will have the added values
5. Add some more data to the DB
6. Click Refresh Dashboard on the Dashboard

### Expected results

The Native Filters data should get updated as per latest data.

### Actual results

The Native filter data doesn't get refreshed

#### Screenshots

Clicking the Refresh Dashboard:
![image](https://user-images.githubusercontent.com/88179228/187179583-cef49be5-eeef-41ce-9250-58123edef928.png)

GIF indicating the Native filters getting refreshed after clicking Refresh Dashboard
![image](https://user-images.githubusercontent.com/88179228/187182839-c843faa0-2cd0-43a0-8904-2d8c2bc9fc39.png)

### Environment

(please complete the following information):

- browser type and version: Microsoft Edge Version 104.0.1293.63 (Official build) (64-bit)
- superset version: Tried in both 1.4.0 and 2.0.0
- any feature flags active:

    FEATURE_FLAGS = {
        "ENABLE_TEMPLATE_PROCESSING": True,
        "ALERT_REPORTS": True
    }
- Cache setting:
   
  Tried with default cache setting and redis cache. For redis, the following settings were used:
      CACHE_CONFIG = {
        'CACHE_TYPE': 'RedisCache',
        'CACHE_REDIS_URL': f"redis://{env('REDIS_HOST')}:{env('REDIS_PORT')}/0",
        'CACHE_DEFAULT_TIMEOUT': 10,

        'CACHE_KEY_PREFIX': 'superset_filter_'
    }
    FILTER_STATE_CACHE_CONFIG = CACHE_CONFIG

michalmalus:
Is there any update to this issue?

Shazad-khan:
This issue has been fixed already please check this issue and close.

billcrook:
@Shazad-khan in which version is this fixed?

unnyns-307:
Hi, I'm experiencing this issue in 2.1.1 too. @Shazad-khan  is this fixed in V.3.0.0?

rusackas:
[...]

PR #29291, opened 2024-06-18
fix: filters not updating with force update when caching is enabled

### LINKED ISSUE #25893: bug: Filters are not updating with force update if cache is enabled for some time
When the dashboard has a filter and responses are cached for let say 24 hours, even after force refreshing the entire dashboard, list of values in the filter do not update (considering there are new values in filter)

#### How to reproduce the bug

1. Go to dashboard that has dropdown filter
2. Force refresh the entire dashboard
3. If you expect new value in the filter, it will not be there.

### Expected results

Updated filter value should be there.

### Actual results

Updated list of values.

### Environment
Caching the data for 24 hours. Using superset 3.0.1

ka-weihe:
### SUMMARY
Currently, when you have caching enabled and you click "refresh dashboard" it only updates the charts but not the filters. This is despite the filters having a loading spinner next to them.

### TESTING INSTRUCTIONS
1. Setup Superset with caching
2. Create a dataset with a simple table and create any chart and add it to a dashboard. Add a filter to the dashboard - could be the values of one of the columns. Insert a new row to the table with a new value for the column you have created a filter for.
3. Force refresh the dashboard using "refresh dashboard" in the upper right hand corner. 

You will see your chart updating, but not the filter. 

### ADDITIONAL INFORMATION

- [x] Has associated issue:
https://github.com/apache/superset/issues/25893
https://github.com/apache/superset/issues/21234

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
@@ -175,7 +175,7 @@ const FilterValue: FC<FilterControlProps> = ({
       setIsRefreshing(true);
       getChartDataRequest({
         formData: newFormData,
-        force: false,
+        force: shouldRefresh,
         ownState: filterOwnState,
       })
         .then(({ response, json }) => {


michael-s-molina (approved):
Thanks for the fix @ka-weihe!

villebro (approved):
Nice, really great improvement!

Vitor-Avila:
This is awesome! thank you 🙏

PR #29302, opened 2024-06-19
chore(distributed-lock): refactor tests

villebro:
### SUMMARY
This PR refactors the distributed lock tests in the following way:
1. Replace assertions of mocked internals with assertions of the functional aspects of the lock
2. Assert that the lock state is committed to the metastore during the locking process

This will make it safer to refactor SQLAlchemy session logic in accordance with [SIP-99A](#25107), as the test will ensure that commits happen at the critical moments (lock is taken, lock is released). In addition, it makes the test immune to refactoring of internals of the lock.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/key_value/get.py
+++ b/superset/commands/key_value/get.py
@@ -16,7 +16,6 @@
 # under the License.
 
 import logging
-from datetime import datetime
 from typing import Any, Optional, Union
 from uuid import UUID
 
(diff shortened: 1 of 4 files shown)


superset/commands/key_value/get.py
```diff
@@ -67,6 +66,6 @@ def validate(self) -> None:
     def get(self) -> Optional[Any]:
         filter_ = get_filter(self.resource, self.key)
         entry = db.session.query(KeyValueEntry).filter_by(**filter_).first()
-        if entry and (entry.expires_on is None or entry.expires_on > datetime.now()):
+        if entry and not entry.is_expired():
```
michael-s-molina, line 69: Nice!

michael-s-molina (approved):

PR #29311, opened 2024-06-20
fix(sqllab): access issue:#28596

### LINKED ISSUE #28596: Can't access database in SQL lab
### Bug description

I have authorized the permissions of sample.main to this user, but this account still cannot be queried in sql_lab

### How to reproduce the bug

1. Create a role `testRole`, and grant `schema access on [examples].[main]`
2. Create an account and give roles `Gamma`, `sql_lab`, `testRole`
3. Use the new account to query sql `SELECT * FROM main.bart_lines`
4. An error occurred
image

### Screenshots/recordings

image

image

### Superset version

3.1.3

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

leoguo1024:
fix sqllab access issue. Fixes #28596

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -1622,7 +1622,7 @@ def extract_tables_from_jinja_sql(sql: str, database: Database) -> set[Table]:
     return (
         tables
         | ParsedQuery(
-            sql_statement=processor.process_template(template),
+            sql_statement=sql,
             engine=database.db_engine_spec.engine,
         ).tables
     )


villebro:
@leoguo1024 please provide a more verbose description, explaining the rationale behind the changes. In particular, the fact that the test `extract_tables_from_jinja_sql` is passing both before and after this change make feel we need to revisit the tests introduced in #27470. Tagging @john-bodley as he introduced the new function and tests.

superset/sql_parse.py
```diff
@@ -1622,7 +1622,7 @@ def extract_tables_from_jinja_sql(sql: str, database: Database) -> set[Table]:
     return (
         tables
         | ParsedQuery(
-            sql_statement=processor.process_template(template),
+            sql_statement=sql,
```
john-bodley, line 1625: @leoguo1024 we purposefully don't wont to render the Jinja template here as it might reference the `[hive|presto|trino].latest_partition` macro (or similar), i.e., the failing test a good check to verify the existing logic.

Hen0k:
Looking forward to this getting fixed. All my users are unable to access data through the SQL lab. The latest version that doesn't have this issue is 3.1.1

Hen0k:
Hey @rusackas  @john-bodley got any suggestions for merging this?

Hen0k:
Hey @leoguo1024 

I am still stuck with this issue. got any recommendations?

villebro:
Echoing @john-bodley's comment here - we really do need to do template processing here, in case there's jinja logic in the query. @leoguo1024 @Hen0k can you check if this issue is still relevant on master branch and/or the latest 4.1 release candidate? If so I can help figure out what's wrong here..

Hen0k:
Hey @villebro @rukshn This is now solved in rc2 for 4.1

After using the latest image, I am able to run queries like before. 

Just to recap the issue was with users that only had access to some superset dataset not being able to run any query on the sql_lab. But this works when I provide the user permission to the entire database. But that is not ideal for my situation where access needed to be limited. 

Now, the users with the sql_lab role are able to make queries with the datasets they have access to (based on other roles) like before.

vicky-sai:
Can we merge this as this is very important for all of the users here ?

giftig:
@vicky-sai in its current state it has important outstanding comments as above -- it's broken the build and an important test is failing, and it needs a better explanation of what it's attempting to do (which might also clarify how it can be properly fixed).

However it looks like @Hen0k has verified that this issue has already been resolved in 4.1rc2. I'd suggest seeing if that build works for you too. If it's fixed in the release candidate we can just close this PR.

rusackas:
This still has outstanding comments, and needs a rebase. I'd love to see this merged, but it needs to be brought up to speed so it's mergeable. Any help would be appreciated on this front, as I'd love to close the related issue.

giftig:
Given this PR is a one-liner which seems to cause more problems than it fixes and is now very stale, I think we can just close this one; it would be easier to raise a fresh PR than build on this one imo.

PR #29314, opened 2024-06-20
fix: Remove recursive repr call

jessie-ross:
### SUMMARY

This repr call is recursive:
```
  File \"/opt/superset/lib/python3.11/site-packages/superset/commands/exceptions.py\", line 31, in __repr__
    return repr(self)
           ^^^^^^^^^^
  File \"/opt/superset/lib/python3.11/site-packages/superset/commands/exceptions.py\", line 31, in __repr__
    return repr(self)
           ^^^^^^^^^^
  File \"/opt/superset/lib/python3.11/site-packages/superset/commands/exceptions.py\", line 31, in __repr__
    return repr(self)
           ^^^^^^^^^^
  [Previous line repeated 321 more times]"}
```

Causing the error not to be printed.

### TESTING INSTRUCTIONS

Demonstration:
```
>>> class Parent():
...     def __init__(self, x):
...         self.x = x
...     def __repr__(self):
...         return 'Parent: ' + self.x
>>> Parent('0')
Parent: 0

>>> class Child(Parent):
...     def __repr__(self):
...         return repr(self)
>>> Child('1')
  File "<stdin>", line 3, in __repr__
  File "<stdin>", line 3, in __repr__
  File "<stdin>", line 3, in __repr__
  [Previous line repeated 2495 more times]
RecursionError: maximum recursion depth exceeded


>>> class Child(Parent):
...     def __repr__(self):
...         return super().__repr__()
>>> Child('2')
Parent: 2
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/exceptions.py
+++ b/superset/commands/exceptions.py
@@ -28,7 +28,7 @@ class CommandException(SupersetException):
     def __repr__(self) -> str:
         if self._exception:
             return repr(self._exception)
-        return repr(self)
+        return repr(super())
 
 
 class ObjectNotFoundError(CommandException):
--- a/superset/commands/exceptions.py
+++ b/superset/commands/exceptions.py
@@ -28,7 +28,7 @@ class CommandException(SupersetException):
     def __repr__(self) -> str:
         if self._exception:
             return repr(self._exception)
-        return repr(super())
+        return super().__repr__()
 
 
 class ObjectNotFoundError(CommandException):


superset/commands/exceptions.py
```diff
@@ -28,7 +28,7 @@ class CommandException(SupersetException):
     def __repr__(self) -> str:
         if self._exception:
             return repr(self._exception)
```
john-bodley, line 30: Should this line also be updated?

jessie-ross: This isn't recursive, as it is calling `repr` on a separate object

villebro (approved):

villebro:
@jessie-ross thanks for the illustrative example in the description! ❤️ LGTM

PR #29315, opened 2024-06-20
docs: update docker container and celery worker chrome installation

Malcolm:
### SUMMARY
The existing documentation has the user download the latest release of Chrome and use ChromeDriver v102.  This combination doesn't function, providing the user with an error saying the Chrome version is unsupported from ChromeDriver during alert/report generation.

This update pulls both the version of Chrome and ChromeDriver from the Chrome Labs repository which is centrally maintained and allows for both the Chrome and ChromeDriver version to stay in lock step.

Additionally the Superset base image version and Chrome versions can be set as build-args during image creation.

### TESTING INSTRUCTIONS
1. Build the container as per `docs/docs/configuration/alerts-reports.mdx` with a customised tag
1. Update `x-superset-image` in `docker-compose.yml` to point to your customised tag
1. Start the stack, configure SMTP, create then fire a report

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sfirke:
@Malcolm Thanks for this PR and sorry it took so long for you to get a review!  It looks great to me but I haven't set up Chrome for my reporting and this isn't something I know much about.  I've put this to a few people for review and hopefully someone can take a look.

And if no one is able to weigh in, I'd say let's err on the side of merging this - I'll put a calendar reminder for myself to circle back here.

docs/docs/configuration/alerts-reports.mdx
```diff
@@ -246,25 +246,28 @@ USER superset
...
+    && apt-get install -y --no-install-recommends \
+        curl unzip jq \
+        libglib2.0-0 libnss3 libxcb1 libatk1.0-0 libatk-bridge2.0-0 \
+        libcups2 libdrm2 libxkbcommon0 libxcomposite1 libxdamage1 \
+        libxfixes3 libxrandr2 libgbm1 libpango-1.0-0 libcairo2 libasound2 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+RUN curl -s -O -O \
+    $(curl -s \
+        https://googlechromelabs.github.io/chrome-for-testing/known-good-versions-with-downloads.json \
```
mistercrunch, line 266: Looks like this will lead to non-deterministic builds (re-running the same docker build can lead to different binaries) which is not ideal. I'd like to keep our build as deterministic as possible.

mistercrunch (dismissed):
Can we switch to a solution that is deterministic?

mistercrunch:
Disregard my previous review/comment, is there not an easier/cleaner way to install this!? I didn't like the way we did it before, and this seems like a good iteration based on this approach, but would there be a better way?

mistercrunch:
Hey, looking at `playwright`, it seems we may be able to streamline things quite a bit by consistently using it to install headless browsers, both chromium and firefox. Wondering if we should target that instead. Also wondering if we could make sure we install versions while using playwright, something like `playwright install --with-deps chromium@1060.0.5414.74 firefox@99.0` seems like it may work and simplify both our dockerfile and documentation

hoalongnatsu:
Maybe we don't need to run:

```
apt-get install -y libglib2.0-0 libnss3 libxcb1 libatk1.0-0 libatk-bridge2.0-0 \
        libcups2 libdrm2 libxkbcommon0 libxcomposite1 libxdamage1 \
        libxfixes3 libxrandr2 libgbm1 libpango-1.0-0 libcairo2 libasound2
```

This command will install these packages:

```
apt-get install -y --no-install-recommends ./google-chrome-*.deb
```

Additionally, we can obtain a version of Chrome using a command, so we don't need to fix one:

```
export CHROMEDRIVER_VERSION=$(curl --silent https://googlechromelabs.github.io/chrome-for-testing/LATEST_RELEASE_116)
```

mistercrunch:
Can we look into using the `playwright` approach instead? Seems it would simplify all this greatly - both for Firefox and Chromium

rusackas:
Converting this to draft whilst it awaits feedback to the questions on the thread, since it's been silent for so long. It may be closed due to inactivity eventually, but please let us know if/when it's ready for re-review!

rusackas:
Closing due to ongoing inactivity, but we appreciate the contribution and are more than happy to reopen this if you want to pick it back up.

PR #29318, opened 2024-06-20
fix(actions): adds checkout step to Change Detector so it can be run independently

rusackas:
### SUMMARY

When manually triggering the E2E action on Github, it throws a bunch of errors, like so:
https://github.com/apache/superset/actions/runs/9604059326

I think/hope that this checkout step should resolve that.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
n/m... this isn't the right way to do this...

PR #29321, opened 2024-06-21
fix(Query): Parse html string error responses to avoid displaying raw HTML as error message

rtexelm:
### SUMMARY

Adds a function to check if a string is HTML and not capable of JSON parsing to ensure the client does not show raw HTML where there should be an error message.

### TESTING INSTRUCTIONS

1. Change the return of the `ChartDataRestApi` response for a `POST` method to be an error with the `message` `kwarg ` containing raw html in string form. e.g. `self.response_500(message='<html><div>Error</div></html>')`
2. Load a dashboard, and click `See more` on the unexpected error alert produced by charts.
3. The message displayed should be `Server error` or, in the case of a detected 404, `Not found`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
@@ -53,13 +53,24 @@ type ErrorType =
 
 type ErrorTextSource = 'dashboard' | 'chart' | 'query' | 'dataset' | 'database';
 
+const ERROR_CODE_LOOKUP = {
+  400: t('Bad request'),
+  401: t('Unauthorized'),
+  403: t('Forbidden'),
+  404: t('Not found'),
+  500: t('Server error'),
+  502: t('Bad gateway'),
+  503: t('Service unavailable'),
+  599: t('Network error'),
+};
+
 export function parseErrorString(response: string): string {
   if (!isJsonString(response) && isProbablyHTML(response)) {
     if (/500|server error/i.test(response)) {
-      return t('Server error');
+      return ERROR_CODE_LOOKUP[500];
     }
     if (/404|not found/i.test(response)) {
-      return t('Not found');
+      return ERROR_CODE_LOOKUP[404];
     }
     return t('Internal error');
   }
@@ -181,6 +192,17 @@ export function getClientErrorObject(
           });
         });
       return;
+    }
+    // check response http status code for generic error return
+    if ((response as any).status) {
+      const { status } = response as any;
+      const error = ERROR_CODE_LOOKUP[status];
+      if (error) {
+        resolve({
+          error,
+        });
+        return;
+      }
     }
 
     // fall back to Response.statusText or generic error of we cannot read the response
--- a/superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
@@ -27,8 +27,8 @@ import {
   isJsonString,
 } from '@superset-ui/core';
 
-// The response always contains an error attribute, can contain anything from the
-// SupersetClientResponse object, and can contain a spread JSON blob
+// The response always contains an error attribute, can contain anything from
+// the SupersetClientResponse object, and can contain a spread JSON blob
 export type ClientErrorObject = {
   error: string;
   errors?: SupersetError[];
@@ -110,17 +110,17 @@ export function getErrorFromStatusCode(status: number): string | null {
 
 export function retrieveErrorMessage(
   str: string,
-  response: Response | JsonObject,
+  errorObject: JsonObject,
 ): string {
   const statusError =
-    'status' in response ? getErrorFromStatusCode(response.status) : null;
+    'status' in errorObject ? getErrorFromStatusCode(errorObject.status) : null;
 
-  // Prefer status code message over the response text
+  // Prefer status code message over the response or HTML text
   return statusError || parseStringResponse(str);
 }
 
-export function parseErrorJson(responseObject: JsonObject): ClientErrorObject {
-  let error = { ...responseObject };
+export function parseErrorJson(responseJson: JsonObject): ClientErrorObject {
+  let error = { ...responseJson };
   // Backwards compatibility for old error renderers with the new error object
   if (error.errors && error.errors.length > 0) {
     error.error = error.description = error.errors[0].message;
@@ -231,13 +231,20 @@ export function getClientErrorObject(
         .clone()
         .json()
         .then(errorJson => {
-          const error = { ...responseObject, ...errorJson };
+          // Destructuring instead of spreading to avoid loss of sibling properties to the body
+          const { url, status, statusText, redirected, type } = responseObject;
+          const responseSummary = { url, status, statusText, redirected, type };
+          const error = {
+            ...errorJson,
+            ...responseSummary,
+          };
           resolve(parseErrorJson(error));
         })
         .catch(() => {
           // fall back to reading as text
           responseObject.text().then((errorText: any) => {
             resolve({
+              // Destructuring not necessary here
               ...responseObject,
               error: t(retrieveErrorMessage(errorText, responseObject)),
             });
@@ -255,7 +262,7 @@ export function getClientErrorObject(
     }
     resolve({
       ...responseObject,
-      error,
+      error: t(parseStringResponse(error)),
     });
   });
 }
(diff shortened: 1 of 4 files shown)


superset-frontend/packages/superset-ui-core/src/utils/html.tsx
```diff
@@ -44,10 +44,26 @@ export function sanitizeHtml(htmlString: string) {
   return xssFilter.process(htmlString);
 }
 
+export function hasHtmlTagPattern(str: string): boolean {
```
michael-s-molina, line 47: Do we have a package or a function inside our current dependencies that provide this feature?

superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
```diff
@@ -171,9 +197,14 @@ export function getClientErrorObject(
     // fall back to Response.statusText or generic error of we cannot read the response
     let error = (response as any).statusText || (response as any).message;
     if (!error) {
-      // eslint-disable-next-line no-console
-      console.error('non-standard error:', response);
-      error = t('An error occurred');
+      // check response http status code and return a generic error message
+      const { status } = response as any;
```
mistercrunch, line 201: So it seems this only gets triggered only if the response object has no `statusText` and no `message`(?) Yet I think we've established that the "message" exist but also may be junk or complex html we don't want to surface as is. 

I'm not sure if that's best. I'd say if 1) there is proper response object, and 2) that response object has an error code and [maybe] 3) we haven't identified a proper/custom/specific error message, we should roll with the http error code as the way to define the message.

Thinking about precedence here:

1. ideally we have a json blob with a clear error message. This is best case and we want to surface that error message that came in the payload. Best case.
2. we have a proper http response with an error code. Here I'd say if `parseErrorString` returns a clear hit then use that, and if not fall back on `ERROR_CODE_LOOKUP` (?). Here I could also go with a pure ERROR_CODE_LOOKUP over `parseErrorString`
3. if we received a string (not a response object) we just have a string and no response object, here maybe we do the `parseErrorString`

I don't like that polymorphic function that is set to receive one of 3 types. Not sure if the refactor belongs in this PR, but I'd like to have consistency on what this function receive. Why can't it always get a response object and never a string? Would it make sense to see all the calls and make them more uniform?

superset-frontend/packages/superset-ui-core/src/utils/html.tsx
rtexelm: it doesn't seem so after quickly running through the package.json and checking with GPT, not sure if it would be as specific to the tags as this regex is

superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
rtexelm: Tried to refactor to suit the order of ops better. If the JSON message is HTML it looks for the status code first then goes through the HTML code to try to pull the error, otherwise surfaces the good message, else it treats the body as text and then looks for the status code or tries the html parsing, last resort it shows errorText.

superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
```diff
@@ -51,6 +53,72 @@ type ErrorType =
...
+  409: 'Conflict',
+  410: 'Gone',
+  411: 'Length required',
+  412: 'Precondition failed',
+  413: 'Payload too large',
+  414: 'URI too long',
+  415: 'Unsupported media type',
+  416: 'Range not satisfiable',
+  417: 'Expectation failed',
+  418: "I'm a teapot",
```
mistercrunch, line 75: 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖 🫖

mistercrunch: TIL https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418

rtexelm: Haha, yeah i had to research this one

superset-frontend/packages/superset-ui-core/src/utils/html.tsx
mistercrunch: seems https://cheerio.js.org/ would be the best fit, and is somewhere in our deptree, but at the same time I'd rather not bring yet another dependency for this relatively small/contained function...

Just for reference ->
```bash
$ npm ls cheerio
superset@0.0.0-dev /Users/max/code/superset/superset-frontend
└─┬ enzyme@3.11.0
  └── cheerio@1.0.0-rc.12
```

I think Enzyme is getting deprecated, so I wouldn't rely on this bycatch...

mistercrunch (approved):
Given everything, this overall LGTM. If the html-handling module tends to grow in the future I'd advocate to refactor using cheerio, but for now I'd rather maintain the logic here than adding a dependency.

rtexelm:
In the latest commit I've tried preserving the `status` and other properties from the response object so that it can be parsed inside the `parseErrorJson` by the `retrieveErrorMessage` function. In the past the properties of the complex response object were being ignored in the passage to `parseErrorJason` by the spread operator. By explicitly destructuring them it should preserve the values for parsing and also allow for passage to a logging function in the future to deduce the root of the unreadable errors.

I've also added a final check to the error message if the response is incapable of being read as a `Response` object.

superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
```diff
@@ -255,7 +262,7 @@ export function getClientErrorObject(
     }
     resolve({
       ...responseObject,
-      error,
+      error: t(parseStringResponse(error)),
```
mistercrunch, line 265: from my understanding `t()` only works for hard-coded strings. I think behind the scene it's using static analysis while running `babel`. It's probably not a problem as is (no-op), but not useful either

superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
```diff
@@ -110,17 +110,17 @@ export function getErrorFromStatusCode(status: number): string | null {
...
+    'status' in errorObject ? getErrorFromStatusCode(errorObject.status) : null;
 
-  // Prefer status code message over the response text
+  // Prefer status code message over the response or HTML text
   return statusError || parseStringResponse(str);
 }
 
-export function parseErrorJson(responseObject: JsonObject): ClientErrorObject {
-  let error = { ...responseObject };
+export function parseErrorJson(responseJson: JsonObject): ClientErrorObject {
```
mistercrunch, line 122: I had to lookup `ClientErrorObject` and think it could be renamed `SupersetClientErrorResponse` for clarity as it appears to be pretty much a `SupersetClientResponse` object with a set of error-related keys. It's probably outside the scope of this PR though. Just a note, feel free to address or leave for another PR

mistercrunch: Similarly the name `parseErrorJson` isn't great. What are we doing in this function really? Looks like we're massaging the `SupersetClientResponse` into fitting `SupersetClientErrorResponse`, kinda. `processErrorResponse` (?) Maybe GPT would have good naming suggestion, it's not great at everything but it's good at that.

mistercrunch (changes requested):
Let's address the `t()` comment, but otherwise LGTM, feel free to address the naming-related comments or save it for another PR

rtexelm:
I'll save the name change for a new PR just so it becomes the focus in case I miss any usage cases across the codebase when replacing the calls to it.

superset-frontend/packages/superset-ui-core/src/query/getClientErrorObject.ts
rtexelm: Just to note it, using `t()` on the string values inside the `ERROR_CODE_LOOKUP` object for some reason caused a failure in the `logging.test` file in `ui-core/test/utils` by causing `console` to evaluate to `undefined`. Maybe some kind of timeout? Whatever the cause I thought it was curious

superset-frontend/packages/superset-ui-core/src/utils/html.tsx
rtexelm: Yeah that seems a little fishy. If it is a problem for complexity then we can just delete it. It was mostly a sort of triple check to expose any HTML so not the most useful of regex calls but I assume it will rarely be conjured.

mistercrunch (approved):

mistercrunch:
standing

PR #29327, opened 2024-06-21
chore(e2e): simplify Cypress record key usage

rusackas:
### SUMMARY

The Cypress Record Key on the repo secrets is no longer base-64 encoded, so there's no need for this decoding. We also might as well just _call it_ the CYPRESS_RECORD_KEY everywhere to match the name of the repo secret, for sanity's sake.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/scripts/cypress_run.py
+++ b/scripts/cypress_run.py
@@ -55,7 +55,6 @@ def get_cypress_cmd(
 
     if use_dashboard:
         # Run using cypress.io service
-        os.environ["CYPRESS_RECORD_KEY"] = os.getenv("CYPRESS_RECORD_KEY", "")
         spec: str = "*/**/*"
         cmd = (
             f"{XVFB_PRE_CMD} "
--- a/scripts/cypress_run.py
+++ b/scripts/cypress_run.py
@@ -66,7 +66,7 @@ def get_cypress_cmd(
         )
     else:
         # Run local, but split the execution
-        os.environ.pop("CYPRESS_KEY", None)
+        os.environ.pop("CYPRESS_RECORD_KEY", None)
         spec_list_str = ",".join(sorted(spec_list))
         if _filter:
             spec_list_str = ",".join(sorted([s for s in spec_list if _filter in s]))
--- a/scripts/cypress_run.py
+++ b/scripts/cypress_run.py
@@ -24,7 +24,7 @@ from datetime import datetime
 XVFB_PRE_CMD = "xvfb-run --auto-servernum --server-args='-screen 0, 1024x768x24' "
 REPO = os.getenv("GITHUB_REPOSITORY") or "apache/superset"
 GITHUB_EVENT_NAME = os.getenv("GITHUB_REPOSITORY") or "push"
-CYPRESS_RECORD_KEY = os.getenv("CYPRESS_RECORD_KEY") or "push"
+CYPRESS_RECORD_KEY = os.getenv("CYPRESS_RECORD_KEY") or ""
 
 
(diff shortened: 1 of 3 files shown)


scripts/cypress_run.py
```diff
@@ -55,12 +55,7 @@ def get_cypress_cmd(
 
     if use_dashboard:
         # Run using cypress.io service
-        cypress_key = os.getenv("CYPRESS_KEY")
-        command = f"echo {cypress_key} | base64 --decode"
-        cypress_record_key = (
-            subprocess.check_output(command, shell=True).decode("utf-8").strip()
-        )
-        os.environ["CYPRESS_RECORD_KEY"] = cypress_record_key
+        os.environ["CYPRESS_RECORD_KEY"] = os.getenv("CYPRESS_RECORD_KEY", "")
```
mistercrunch, line 58: isn't this kind of a no-op?

rusackas: I had the same thought... let's scrap it!

scripts/cypress_run.py
```diff
@@ -70,7 +64,6 @@ def get_cypress_cmd(
         )
     else:
         # Run local, but split the execution
-        os.environ.pop("CYPRESS_KEY", None)
```
rusackas, line 73: This seemed unnecessary as well.

michael-s-molina (approved):

scripts/cypress_run.py
```diff
@@ -65,7 +64,6 @@ def get_cypress_cmd(
         )
     else:
         # Run local, but split the execution
-        os.environ.pop("CYPRESS_RECORD_KEY", None)
```
mistercrunch: Do we not want that pop? In the case of running off the repo where the env var is always present, it may always disregard the `use_dashboard` param effectively

rusackas: ¯\_(ツ)_/¯ I wasn't sure... I'll add it back in... it certainly doesn't hurt.

mistercrunch (approved):

michael-s-molina:
@rusackas I'm not sure if the problem is related to this PR or other Cypress-related one but I'm getting errors when pushing to the release branches:

```
Opening Cypress...
  Your Record Key b9c81...2394a is not valid with this projectId: ukwxzo
  
  It may have been recently revoked by you or another user.
  
  Please log into the Dashboard to see the valid Record Keys.
  
  https://on.cypress.io/dashboard/projects/ukwxzo
```

This is problematic because it's essential that the whole suite of tests run in release branch pushes.

michael-s-molina:
@rusackas I don't know if this PR is the source for the problem but we can execute Cypress anymore for release branches such as [4.0](https://github.com/apache/superset/actions/runs/9681795234/job/26713216396). This problematic because we need the whole suite of tests when pushing to release branches in order to check if the release is good to go.

```
Opening Cypress...
  Your Record Key b9c81...2394a is not valid with this projectId: ud5x2f
  
  It may have been recently revoked by you or another user.
  
  Please log into the Dashboard to see the valid Record Keys.
  
  https://on.cypress.io/dashboard/projects/ud5x2f
```

rusackas:
It should run the tests on those branches, just not with the cypress dashboard enabled. I rotated the key, and changed the action to use the repos secret, so you may have to pull this in as a cherry?

michael-s-molina:
It was a little tricky, because 4.0 structure is completely different than master, but I was able to cherry-pick the necessary changes to make Cypress work again.

PR #29330, opened 2024-06-21
fix: refactor view error handling into a separate module

mistercrunch:
### SUMMARY

Starting over after a change of plan. Changing title and summary for the PR.

In this PR:
- centralizing views/app error handling in `superset/views/error_handling.py`
- moving app referencing outside of module scope an into a function that `superset/initialization` can call
- altering imports 
- merging `json_errors_response` into `json_error_response` with a bit of polymorphy so it can receive different types as its first param
- minimal refactoring outside of this, mostly just centralizing the views error handling functions into a separate module

Future / potential follow ups:
- refactor further, cleaning up code, DRYing further 
- making sure all errors are covered
- look back at error-handling-related SIPs https://github.com/apache/superset/issues/9298 and SIP-40 for next steps

~~I've recently encountered some errors on the dashboard page, within Chart widgets where html is displayed as a result of calling the `api/chart/data` endpoint, and presumably while hitting some 502 / 504s-type error. The error is hard to reproduce, but I'm thinking the approach bellow should improve / prevent that endpoint from ever returning html.~~

~~Two changes here: sprinkling the @handle_api_error decorator in critical related views improving the 500 flask error handler to be a bit smarter, hoping it'll catch some of the problems and return json when expecting it~~

## Screenshot

Screenshot 2024-06-12 at 11 19 34 AM

~~Also note that this is a backend counterpart to https://github.com/apache/superset/pull/29321, where that PR goes broad and also handles proxies/load-balancer/nginx or something else returning an unexpected response.~~

### DIFF
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -52,12 +52,7 @@ from superset.utils.core import (
     get_user_id,
 )
 from superset.utils.decorators import logs_context
-from superset.views.base import (
-    CsvResponse,
-    generate_download_headers,
-    handle_api_exception,
-    XlsxResponse,
-)
+from superset.views.base import CsvResponse, generate_download_headers, XlsxResponse
 from superset.views.base_api import statsd_metrics
 
 if TYPE_CHECKING:
@@ -76,7 +71,6 @@ class ChartDataRestApi(ChartRestApi):
         action=lambda self, *args, **kwargs: f"{self.__class__.__name__}.data",
         log_to_statsd=False,
     )
-    @handle_api_exception
     def get_data(self, pk: int) -> Response:
         """
         Take a chart ID and uses the query context stored when the chart was saved
@@ -190,7 +184,6 @@ class ChartDataRestApi(ChartRestApi):
         action=lambda self, *args, **kwargs: f"{self.__class__.__name__}.data",
         log_to_statsd=False,
     )
-    @handle_api_exception
     def data(self) -> Response:
         """
         Take a query context constructed in the client and return payload
@@ -231,7 +224,6 @@ class ChartDataRestApi(ChartRestApi):
               $ref: '#/components/responses/500'
         """
         json_body = None
-
         if request.is_json:
             json_body = request.json
         elif request.form.get("form_data"):
@@ -277,7 +269,6 @@ class ChartDataRestApi(ChartRestApi):
         f".data_from_cache",
         log_to_statsd=False,
     )
-    @handle_api_exception
     def data_from_cache(self, cache_key: str) -> Response:
         """
         Take a query context cache key and return payload
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -52,7 +52,12 @@ from superset.utils.core import (
     get_user_id,
 )
 from superset.utils.decorators import logs_context
-from superset.views.base import CsvResponse, generate_download_headers, XlsxResponse
+from superset.views.base import (
+    CsvResponse,
+    generate_download_headers,
+    handle_api_exception,
+    XlsxResponse,
+)
 from superset.views.base_api import statsd_metrics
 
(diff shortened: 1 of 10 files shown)


superset/views/core.py
```diff
@@ -886,13 +885,6 @@ def fetch_datasource_metadata(self) -> FlaskResponse:
         datasource.raise_for_access()
         return json_success(json.dumps(sanitize_datasource_data(datasource.data)))
 
-    @app.errorhandler(500)
```
mistercrunch, line 889: This errorhandler is lonely here in core.py, all them other ondes are in base.py. Also ideally should have logic to return json depending on mimetype

superset/views/base.py
```diff
@@ -522,6 +522,7 @@ def show_command_errors(ex: CommandException) -> FlaskResponse:
 
 # Catch-all, to ensure all errors from the backend conform to SIP-40
 @superset_app.errorhandler(Exception)
+@superset_app.errorhandler(500)
```
mistercrunch, line 525: It seems this existing logic fits what we need to handle `500`s properly

mistercrunch:
@dpgaspar let me know what you think!

superset/charts/data/api.py
```diff
@@ -71,6 +76,7 @@ class ChartDataRestApi(ChartRestApi):
         action=lambda self, *args, **kwargs: f"{self.__class__.__name__}.data",
         log_to_statsd=False,
     )
+    @handle_api_exception
```
john-bodley, line 79: Either @betodealmeida or @dpgaspar would probably be best to review this.

dpgaspar: Unsure about these, since the endpoints only handle 404 and 400 by themselves so every other possibility either succeeds or raises and exception (assuming `command.validate()` doesn't return flask responses).
So all exceptions are already handled by `@superset_app.errorhandler`

yet, `SupersetSecurityException`, `IntegrityError`, `DatabaseError`, `DataError` specially handled by `@handle_api_exception` seem to be missing from our flask errorhandlers.

mistercrunch: I'm now thinking about the tradeoffs between generic app-level handlers VS doing it through decorators. Maybe a better approach would be to factor out the error-handlers into a nice module, and make sure all cases are handled: API vs HTML and all potential errors and good catch-alls. @betodealmeida pointed out this solid SIP https://github.com/apache/superset/issues/9298 that @john-bodley wrote in 2020. Maybe now is the time to revive it and make it a reality.

mistercrunch:
@john-bodley @dpgaspar I'm feeling that improving/refactoring the app-level error handlers might be the way forward here. I'm also realizing now that the html screenshot above might have some from our API-gateway as opposed to Flask/Superset here, pointing to a solution in that area and/or in the frontend.

Given this, I'd probably just close this PR, and open another that would bring all app-level error-handlers in a module (sitting in something like `superset/initialization/app_error_handlers.py`) and make sure they all handle both api/json VS html/templates properly.

mistercrunch:
@dpgaspar @john-bodley I'd love a review. Note this is mostly moving functions into a module (didn't touch logic much), but I'm looking for direction as where to where to take this next in context of https://github.com/apache/superset/issues/9298 . I'll try and revive this SIP too

betodealmeida (approved):
Nice! This is lovely!

superset/utils/core.py
```diff
@@ -432,7 +432,7 @@ def error_msg_from_exception(ex: Exception) -> str:
             msg = ex.message.get("message")  # type: ignore
         elif ex.message:
             msg = ex.message
-    return msg or str(ex)
+    return str(msg) or str(ex)
```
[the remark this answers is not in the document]
mistercrunch: I'm a bit afraid to touch it, I'll merge this and do a follow up to keep improving error handling.

PR #29338, opened 2024-06-24
fix(tags): tags api change tag_get_objects method to be aligned with api documentation

Bierbarbar:
### SUMMARY
I just came across an issue where the implemented logic was not aligned with the API documentation on OpenAPI. As a result, the method was not usable. I modified the method to align with the documentation, although I saw a benefit in the initial implementation of a better filter mechanism and bulk requests. I think a breaking change should be discussed first; therefore, I removed the non-functional code here, so that we have a fixed api soon.

### TESTING INSTRUCTIONS

Start up superset and use the `tags/get_objects/<tag_id>` method it now should return the charts that belong to a tag.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/tags/api.py
+++ b/superset/tags/api.py
@@ -560,9 +560,8 @@ class TagRestApi(BaseSupersetModelRestApi):
             name: types
             schema:
               type: array
-              schema:
-                items:
-                  type: string
+              items:
+                type: string
 
           responses:
             200:
--- a/superset/tags/api.py
+++ b/superset/tags/api.py
@@ -542,27 +542,29 @@ class TagRestApi(BaseSupersetModelRestApi):
         """Get all objects associated with a tag.
         ---
         get:
-          summary: Get all objects associated with a tag. If tagIds is set tags will be ignored.
+          summary: Get all objects associated with a tag
+          description: >-
+            Get all objects associated with a tag.
+            If tagIds is set, tags will be ignored.
           parameters:
-          - in: path
+          - in: query
             name: tagIds
             schema:
               type: array
               items:
                 type: integer
-          - in: path
+          - in: query
             name: tags
             schema:
               type: array
               items:
                 type: string
-          - in: path
+          - in: query
             name: types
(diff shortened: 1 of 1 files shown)


rusackas:
The existing endpoint should work, you just need to pass it a single Tag ID in the list, rather than a longer list of tag IDs.

This would be a breaking change, as you're removing support for supplying the API with multiple Tag IDs.

The real change we need to make is to the documentation. The docs say the API expects `tag_id` (int) but in fact, it's expecting `tagIds` (array). We should probably just update the docs since the API supports all of the above use cases.

Bierbarbar:
@rusackas, thanks for your feedback. I think somehow you have a point here. I thought it was the correct way to change the behavior as the OpenAPI specification is a public interface and expects a totally different variable type here: int vs array, different name, and different type: path vs parameter. But for me, it is also fine to consider that the bug is that the API is not properly documented and to fix this. I will hand something in for it.

Bierbarbar:
I adapted the documentation as discussed above:
@rusackas @dpgaspar please have a look :)

rusackas:
Closing/reopening to get CI running again. Hopefully it'll pass so we can review/merge. Sorry this has been stuck so long!

rusackas:
Looks like CI still isn't happy... at least `pre-commit` for starters. 

```
pre-commit install
```
```
pre-commit run --all-files
```

rusackas:
@Bierbarbar this still looks like a clean fix and it's mergeable. I'll see if I can help rebase this.

rusackas (approved):
Thanks for pivoting this to a docs fix (and allowing some touch-ups). The OpenAPI now matches the actual `tagIds`/`tags`/`types` query params instead of the bogus `tag_id` path int, no breaking change. LGTM, merging soon, hopefully...

rusackas:
Thank you!!!

PR #29344, opened 2024-06-24
chore(key-value): convert command to dao

villebro:
### SUMMARY
This PR converts the Key Value commands to a DAO, which gives the caller much more control over when commits/rollbacks happen, and what to do about expired entries. Other changes:
- Update all code that uses the KV commands to use the new DAO, making sure flushes/commits happen outside the DAO. In particular, this covers the following major components:
  - Explore permalinks
  - Dashboard permalinks
  - Distributed lock. During the refactor, the distributed lock was moved under a module of its own + break out the stateful logic into commands that use the new DAO.
  - Metastore cache. I considered breaking out the logic into commands, but it felt more sensible to keep the code as-is under a single python file, as it's under `extensions`. In addition, this doesn't use the new `transaction` decorator, as we need to adhere to the `Flask-Caching` API.
- The old KV command integration tests are converted into functionally equivalent unit tests

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/key_value/types.py
```diff
@@ -32,12 +31,6 @@
 from superset.utils.backports import StrEnum
 
 
-@dataclass
-class Key:
-    id: int | None
-    uuid: UUID | None
-
-
```
villebro, line 40: This is no longer needed, as the DAO always returns the full entry

superset/daos/key_value.py
```diff
@@ -0,0 +1,195 @@
...
+)
+from superset.key_value.models import KeyValueEntry
+from superset.key_value.types import KeyValueCodec, KeyValueResource
+from superset.key_value.utils import get_filter
+from superset.utils.core import get_user_id
+
+logger = logging.getLogger(__name__)
+
+
+class KeyValueDAO(BaseDAO[KeyValueEntry]):
```
john-bodley, line 40: I think the current behavior (though wrong) is for DAOs to commit. Should you mimic that same logic here? I can update the logic in https://github.com/apache/superset/pull/24969 (if it ever lands).

superset/daos/key_value.py
```diff
@@ -0,0 +1,195 @@
...
+    def create_entry(
+        resource: KeyValueResource,
+        value: Any,
+        codec: KeyValueCodec,
+        key: int | UUID | None = None,
+        expires_on: datetime | None = None,
+    ) -> KeyValueEntry:
+        """
+
+        :param resource:
```
john-bodley, line 113: If you're not going to define these I would just remove the docstring.

superset/daos/key_value.py
villebro: Sounds good, I'll update accordingly.

superset/daos/key_value.py
villebro: Oops, good catch, I meant to add docstrings, but later they felt redundant. I'll remove them.

superset/daos/key_value.py
villebro: @john-bodley Uhh, I started doing this, and updating the code with the old intermediate commit logic is feeling super wrong now that I've gotten used to the new way of doing it 🤮  I'm wondering if we should just keep this logic as-is, and accept some disconnect in the updated DAOs? Or do you think that's dangerous? The Key Value DAO is well tested currently, so it should be pretty immune to regressions.

villebro: As #24969 is now merged, this comment is no longer valid

tests/unit_tests/fixtures/common.py
```diff
@@ -73,3 +76,18 @@ def create_columnar_file(
...
+    user = User(
+        first_name="Alice",
+        last_name="Doe",
+        email="adoe@example.org",
+        username="alice_admin",
+        roles=[role],
+    )
+    db.session.add(user)
+    db.session.flush()
+    yield user
```
villebro, line 93: I tried to use the `admin` user that gets created during test startup, and that worked when running the `key_value_test.py` in isolation, but when running the full test suite the user disappears. So it appears there's a test that deletes it. I didn't have time to try to find the culprit, but at any rate I feel it may be a better idea to not rely on any pre-existing fixtures, hence I suggest using this fixture going forward for new tests needing an admin user.

superset/daos/key_value.py
```diff
@@ -0,0 +1,145 @@
...
+from superset.key_value.types import Key, KeyValueCodec, KeyValueResource
+from superset.key_value.utils import get_filter
+from superset.utils.core import get_user_id
+
+logger = logging.getLogger(__name__)
+
+
+class KeyValueDAO(BaseDAO[KeyValueEntry]):
+    @staticmethod
+    def get_entry(
```
villebro, line 42: To reviewers who think the `_entry` suffix seems redundant: `BaseDAO` already has `create`, `update`, `delete`, which expects passing an instance of the model. This doesn't fit this use case very well, as the interface here is slightly different from the typical ORM object:
- `key` can be an `int` or `UUID`
- The Key Value model encodes/decodes values using a codec that we want to enforce in our method signatures

For this reason I needed to name these slightly differently to not collide with the base methods.

superset/daos/key_value.py
```diff
@@ -0,0 +1,145 @@
...
+        resource: KeyValueResource,
+        key: Key,
+    ) -> KeyValueEntry | None:
+        filter_ = get_filter(resource, key)
+        return db.session.query(KeyValueEntry).filter_by(**filter_).first()
+
+    @classmethod
+    def get_value(
```
villebro, line 50: This is a convenience method for when one only wants the value, not the full entry

superset/distributed_lock/__init__.py
```diff
@@ -75,28 +53,25 @@ def KeyValueDistributedLock(  # pylint: disable=invalid-name
...
+    from superset.commands.distributed_lock.create import CreateDistributedLock
+    from superset.commands.distributed_lock.delete import DeleteDistributedLock
+    from superset.commands.distributed_lock.get import GetDistributedLock
 
     key = get_key(namespace, **kwargs)
+    value = GetDistributedLock(namespace=namespace, params=kwargs).run()
+    if value:
+        logger.debug("Lock on namespace %s for key %s already taken", namespace, key)
+        raise CreateKeyValueDistributedLockFailedException("Lock already taken")
+
```
villebro, line 67: I added a get before trying to create the lock, as I noticed that creating locks when it was already taken was producing ugly error logs on Postgres. While this adds one more roundtrip to the metastore, the locking operations are fairly rare, hance it felt cleaner to do it this way. Note, that we're still handling the rare case that the lock got created after the get, so that will still be handled gracefully.

tests/integration_tests/explore/permalink/commands_tests.py
```diff
@@ -149,13 +149,14 @@ def test_get_permalink_command_with_old_dataset_key(
...
-        datasource_string = f"{dataset.id}__{DatasourceType.TABLE}"
+        datasource_string = f"{dataset.id}__{DatasourceType.TABLE.value}"
 
         decode_id_mock.return_value = "123456"
-        get_kv_command_mock.return_value = {
+        kv_get_value_mock.return_value = {
             "chartId": slice.id,
             "datasetId": dataset.id,
             "datasource": datasource_string,
+            "datasourceType": DatasourceType.TABLE.value,
```
villebro, line 159: bycatch: the return value was missing a required field

superset/commands/explore/permalink/create.py
```diff
@@ -65,15 +66,13 @@ def run(self) -> str:
...
-        )
-        key = command.run()
-        if key.id is None:
+        entry = KeyValueDAO.create_entry(self.resource, value, self.codec)
+        db.session.flush()
+        key = entry.id
+        if key is None:
             raise ExplorePermalinkCreateFailedError("Unexpected missing key id")
-        return encode_permalink_key(key=key.id, salt=self.salt)
+        db.session.commit()  # pylint: disable=consider-using-transaction
```
michael-s-molina, line 74: Why do you need `commit` here given that the method is annotated with `@transaction`?

superset/extensions/metastore_cache.py
```diff
@@ -130,7 +116,6 @@ def has(self, key: str) -> bool:
         return False
 
     def delete(self, key: str) -> Any:
-        # pylint: disable=import-outside-toplevel
-        from superset.commands.key_value.delete import DeleteKeyValueCommand
-
-        return DeleteKeyValueCommand(resource=RESOURCE, key=self.get_key(key)).run()
+        ret = KeyValueDAO.delete_entry(RESOURCE, self.get_key(key))
+        db.session.commit()  # pylint: disable=consider-using-transaction
```
michael-s-molina, line 120: Isn't `@transaction` possible here?

tests/unit_tests/dao/key_value_test.py
```diff
@@ -0,0 +1,385 @@
...
+            key=ID_KEY,
+            value=NEW_VALUE,
+            codec=JSON_CODEC,
+        )
+        found_entry = KeyValueDAO.get_entry(resource=RESOURCE, key=ID_KEY)
+        assert found_entry is not None
+        assert JSON_CODEC.decode(found_entry.value) == NEW_VALUE
+        assert entry.changed_by_fk == admin_user.id
+
+    db.session.rollback()
```
michael-s-molina, line 312: Can you extract `db.session.rollback()` to an after each fixture?

superset/commands/explore/permalink/create.py
villebro: Whoops, that wasn't meant to be there

superset/extensions/metastore_cache.py
villebro: Good catch, updating

michael-s-molina (approved):
LGTM. Thanks for the improvements and for addressing all comments @villebro!

rtexelm:
I can't pinpoint why but something in this PR has caused local dev builds to fail with a flask `app_context` error. Also observed by Joe Li at Preset.

```zsh
2024-07-01 17:47:46,317:ERROR:superset.app:Failed to create app
  File "/Users/user/superset/app.py", line 40, in create_app
    app_initializer.init_app()
  File "/Users/user/superset/initialization/__init__.py", line 479, in init_app
    self.configure_cache()
  File "/Users/user/superset/initialization/__init__.py", line 511, in configure_cache
    cache_manager.init_app(self.superset_app)
  File "/Users/user/superset/utils/cache_manager.py", line 93, in init_app
    self._init_cache(
  File "/Users/user/superset/utils/cache_manager.py", line 87, in _init_cache
    cache.init_app(app, cache_config)
  File "/Users/user/site-packages/flask_caching/__init__.py", line 145, in init_app
    self._set_cache(app, config)
  File "/Users/user/site-packages/flask_caching/__init__.py", line 155, in _set_cache
    cache_factory = import_string(import_me)
  File "/Users/user/site-packages/werkzeug/utils.py", line 596, in import_string
    __import__(import_name)
  File "/Users/user/superset/extensions/metastore_cache.py", line 27, in <module>
    from superset.daos.key_value import KeyValueDAO
  File "/Users/user/superset/daos/key_value.py", line 32, in <module>
    from superset.key_value.models import KeyValueEntry
  File "/Users/user/superset/key_value/models.py", line 24, in <module>
    from superset.models.helpers import AuditMixinNullable, ImportExportMixin
  File "/Users/user/superset/models/__init__.py", line 17, in <module>
    from . import core, dynamic_plugins, sql_lab, user_attributes  # noqa: F401
  File "/Users/user/superset/models/core.py", line 74, in <module>
    from superset.models.helpers import AuditMixinNullable, ImportExportMixin
  File "/Users/user/superset/models/helpers.py", line 107, in <module>
    config = app.config
  File "/Users/user/site-packages/werkzeug/local.py", line 318, in __get__
    obj = instance._get_current_object()
  File "/Users/user/site-packages/werkzeug/local.py", line 519, in _get_current_object
    raise RuntimeError(unbound_message) from None
RuntimeError: Working outside of application context.

This typically means that you attempted to use functionality that needed
the current application. To solve this, set up an application context
with app.app_context(). See the documentation for more information.
```

Followed by this

```zsh
  File "/Users/user/bin/superset", line 8, in <module>
    sys.exit(superset())
  File "/Users/user/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/user/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/user/lib/python3.10/site-packages/click/core.py", line 1685, in invoke
    super().invoke(ctx)
  File "/Users/user/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/user/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/user/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/user/lib/python3.10/site-packages/flask/cli.py", line 355, in decorator
    app = __ctx.ensure_object(ScriptInfo).load_app()
  File "/Users/user/lib/python3.10/site-packages/flask/cli.py", line 309, in load_app
    app = locate_app(import_name, name)
  File "/Users/user/lib/python3.10/site-packages/flask/cli.py", line 238, in locate_app
    return find_app_by_string(module, app_name)
  File "/Users/user/lib/python3.10/site-packages/flask/cli.py", line 166, in find_app_by_string
    app = attr(*args, **kwargs)
  File "/Users/user/superset/app.py", line 40, in create_app
    app_initializer.init_app()
  File "/Users/user/superset/initialization/__init__.py", line 479, in init_app
    self.configure_cache()
  File "/Users/user/superset/initialization/__init__.py", line 511, in configure_cache
    cache_manager.init_app(self.superset_app)
  File "/Users/user/superset/utils/cache_manager.py", line 93, in init_app
    self._init_cache(
  File "/Users/user/superset/utils/cache_manager.py", line 87, in _init_cache
    cache.init_app(app, cache_config)
  File "/Users/user/lib/python3.10/site-packages/flask_caching/__init__.py", line 145, in init_app
    self._set_cache(app, config)
  File "/Users/user/lib/python3.10/site-packages/flask_caching/__init__.py", line 155, in _set_cache
    cache_factory = import_string(import_me)
  File "/Users/user/lib/python3.10/site-packages/werkzeug/utils.py", line 596, in import_string
    __import__(import_name)
  File "/Users/user/superset/extensions/metastore_cache.py", line 27, in <module>
    from superset.daos.key_value import KeyValueDAO
  File "/Users/user/superset/daos/key_value.py", line 32, in <module>
    from superset.key_value.models import KeyValueEntry
  File "/Users/user/superset/key_value/models.py", line 24, in <module>
    from superset.models.helpers import AuditMixinNullable, ImportExportMixin
  File "/Users/user/superset/models/__init__.py", line 17, in <module>
    from . import core, dynamic_plugins, sql_lab, user_attributes  # noqa: F401
  File "/Users/user/superset/models/core.py", line 74, in <module>
    from superset.models.helpers import AuditMixinNullable, ImportExportMixin
  File "/Users/user/superset/models/helpers.py", line 107, in <module>
    config = app.config
  File "/Users/user/lib/python3.10/site-packages/werkzeug/local.py", line 318, in __get__
    obj = instance._get_current_object()
  File "/Users/user/lib/python3.10/site-packages/werkzeug/local.py", line 519, in _get_current_object
    raise RuntimeError(unbound_message) from None
RuntimeError: Working outside of application context.

This typically means that you attempted to use functionality that needed
the current application. To solve this, set up an application context
with app.app_context(). See the documentation for more information.
```

villebro:
@rtexelm I was able to reproduce, it was due to a last second cleanup where I moved imports to the top level from the methods. I'll open up a fix shortly after doing some more testing.

PR #29361, opened 2024-06-25
feat: Adds chart IDs option to migrate-viz

michael-s-molina:
### SUMMARY
Adds the optional `chart_id` parameter to the `migrate-viz` command to allow upgrading/downgrading a specific chart.

```
Usage: superset migrate-viz upgrade [OPTIONS]

  Upgrade a viz to the latest version.

Options:
  -t, --viz_type TEXT  The viz type to upgrade: area, bar, bubble, dist_bar,
                       dual_line, heatmap, histogram, line, pivot_table,
                       sunburst, treemap  [required]
  --id INTEGER         The chart ID to upgrade. It can set set multiple times.
  --help               Show this message and exit.
```

```
Usage: superset migrate-viz downgrade [OPTIONS]

  Downgrade a viz to the previous version.

Options:
  -t, --viz_type TEXT  The viz type to upgrade: area, bar, bubble, dist_bar,
                       dual_line, heatmap, histogram, line, pivot_table,
                       sunburst, treemap  [required]
  --id INTEGER         The chart ID to downgrade. It can set set multiple times.
  --help               Show this message and exit.
```

### TESTING INSTRUCTIONS
Check that you can upgrade/downgrade a specific chart using the `-id` option.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/cli/viz_migrations.py
+++ b/superset/cli/viz_migrations.py
@@ -14,10 +14,11 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from __future__ import annotations
+
 from enum import Enum
 
 import click
-from click_option_group import optgroup, RequiredMutuallyExclusiveOptionGroup
 from flask.cli import with_appcontext
 
 from superset import db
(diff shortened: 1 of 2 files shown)


giftig:
Two small functional suggestions:

1. Looks like it's currently either one chart or all charts of a type, maybe it'd make sense to accept a list of chart IDs instead of just a single one. You could either use a comma separated list or a click option to let you specify the parameter multiple times, to do that.
2. Should visualisation type be optional if you're specifying id(s)? Currently you'd have to specify it and it'd have to match the vis type of the chart ID.

superset/cli/viz_migrations.py
```diff
@@ -46,37 +47,43 @@ def migrate_viz() -> None:
...
+    help=f"The viz type to upgrade: {', '.join(list(VizType))}",
+    required=True,
+)
+@click.option(
+    "--chart_id",
+    help="The chart ID to upgrade",
+    type=int,
 )
-def upgrade(viz_type: str) -> None:
+def upgrade(viz_type: str, chart_id: int | None = None) -> None:
```
john-bodley, line 61: Should this be something which can be specified [multiple times](https://click.palletsprojects.com/en/8.1.x/options/#multiple-options)? Also I wonder if `--chart` would be suffice, i.e., it's somewhat implicit that you would be specifying the chart ID.

giftig: Actually given the context (migrating a visualisation) maybe `--id` would make more sense and be more in line with `-t` for viz type.

michael-s-molina: I updated the description with:

```
--ids TEXT           A comma separated list of chart IDs to downgrade
```

michael-s-molina:
> Two small functional suggestions:

> Looks like it's currently either one chart or all charts of a type, maybe it'd make sense to accept a list of chart IDs instead of just a single one. You could either use a comma separated list or a click option to let you specify the parameter multiple times, to do that.

> Should visualisation type be optional if you're specifying id(s)? Currently you'd have to specify it and it'd have to match the vis type of the chart ID.

Great suggestions. I did both.

giftig (approved):
LGTM

superset/cli/viz_migrations.py
```diff
@@ -47,66 +87,81 @@ def migrate_viz() -> None:
...
 @optgroup.option(
     "--viz_type",
     "-t",
-    help=f"The viz type to migrate: {', '.join(list(VizType))}",
+    help=f"The viz type to upgrade: {', '.join(list(VizType))}",
+    type=str,
 )
-def upgrade(viz_type: str) -> None:
+@optgroup.option(
+    "-ids",
```
john-bodley, line 99: Shouldn't this have two dashes rather than one? Also why not use the multiple options `--id 1 --id 2 ...` so that way you don't need to split a string and the types will be correct, i.e., the following will correctly fail during the validation phase:

```
... --id 1 --id foo 
```

michael-s-molina: Change it to use double dashes 👍🏼

michael-s-molina: Change it to use a multi-value parameter `--id`

PR #29371, opened 2024-06-26
chore(ci): Start Celery worker as a background process

john-bodley:
### SUMMARY

I've been fighting with inconsistencies between running CI locally and remote. Whilst working on a specific PR the Celery tests would always succeed locally whereas they would fail in a remote setting. The errors were always suppressed, but my hunch was that maybe the Celery worker(s) weren't responsive as queries remained in the pending state.

If I mangled the `python_tests.sh` script to only run the `tests/integration_tests/celery_tests.py` tests everything seemed to be fine which indicated that maybe there was a temporal facet to the flakiness of the tests. Reading through the GitHub actions/workflows it seems like we manually background the task (via the `&`) yet one can actually run Celery in a detached state, which seems more robust.

### TESTING INSTRUCTIONS

CI. Also verified that the tests on my other PR successfully ran.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/bashlib.sh
+++ b/.github/workflows/bashlib.sh
@@ -117,6 +117,20 @@ testdata() {
   say "::endgroup::"
 }
 
+celery-worker() {
+  cd "$GITHUB_WORKSPACE"
+  say "::group::Start Celery worker"
+  # must specify PYTHONPATH to make `tests.superset_test_config` importable
+  export PYTHONPATH="$GITHUB_WORKSPACE"
+  celery \
+    --app=superset.tasks.celery_app:app \
+    worker \
+      --concurrency=2 \
+      --detach \
+      --optimization=fair
+  say "::endgroup::"
+}
+
 cypress-install() {
   cd "$GITHUB_WORKSPACE/superset-frontend/cypress-base"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/bashlib.sh
```diff
@@ -117,6 +117,20 @@ testdata() {
   say "::endgroup::"
 }
 
+celery-worker() {
```
john-bodley, line 120: DRY.

.github/workflows/bashlib.sh
```diff
@@ -117,6 +117,20 @@ testdata() {
...
+  cd "$GITHUB_WORKSPACE"
+  say "::group::Start Celery worker"
+  # must specify PYTHONPATH to make `tests.superset_test_config` importable
+  export PYTHONPATH="$GITHUB_WORKSPACE"
+  celery \
+    --app=superset.tasks.celery_app:app \
+    worker \
+      --concurrency=2 \
```
john-bodley, line 128: This seems clearer than `-c 2`.

.github/workflows/bashlib.sh
```diff
@@ -117,6 +117,20 @@ testdata() {
...
+      --detach \
+      --optimization=fair
```
john-bodley, line 130: This seems clearer than `-Ofair`.

dpgaspar (approved):

PR #29381, opened 2024-06-26
chore(security): Clean up session/commit logic

john-bodley:
### SUMMARY

A few housekeeping tasks with regards to the security manager:

1. The Flask-AppBuilder `get_session` method is a property and thus `get_session()` actually unnecessarily instantiates a new session.
2. There's a pattern of over committing. Note that the Flask-AppBuilder [`add_permission_view_menu`](https://github.com/dpgaspar/Flask-AppBuilder/blob/bbf2adb1312ac1f994ef04e5d5e581d1447cb732/flask_appbuilder/security/sqla/manager.py#L610-L635) function commits (which is called by `merge_pv`) and thus there's no need to re-commit.

Note many of the existing commits could likely be also removed as we have a tendency of over commiting. I'm hoping to tackle that as part of SIP-99B by way of transactions.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -2458,9 +2458,8 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             )
             .filter(RLSFilterRoles.c.role_id.in_(user_roles))
         )
-        filter_tables = (
-            self.get_session.query(RLSFilterTables.c.rls_filter_id)
-            .filter(RLSFilterTables.c.table_id == table.id)
+        filter_tables = self.get_session.query(RLSFilterTables.c.rls_filter_id).filter(
+            RLSFilterTables.c.table_id == table.id
         )
         query = (
             self.get_session.query(
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -1017,9 +1017,9 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
                 == None,  # noqa: E711
             )
         )
-        self.get_session.commit()
         if deleted_count := pvms.delete():
             logger.info("Deleted %i faulty permissions", deleted_count)
+        self.get_session.commit()
 
     def sync_role_definitions(self) -> None:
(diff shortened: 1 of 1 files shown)


superset/security/manager.py
```diff
@@ -1017,9 +1017,9 @@ def clean_perms(self) -> None:
                 == None,  # noqa: E711
             )
         )
-        self.get_session.commit()
         if deleted_count := pvms.delete():
             logger.info("Deleted %i faulty permissions", deleted_count)
+        self.get_session.commit()
```
john-bodley, line 1022: I'm not really sure what's going on here. The commit was after a `SELECT` as opposed to the `DELETE`.

superset/security/manager.py
```diff
@@ -1047,9 +1047,6 @@ def sync_role_definitions(self) -> None:
             )
 
         self.create_missing_perms()
-
-        # commit role and view menu updates
-        self.get_session.commit()
```
john-bodley, line 1052: The `create_missing_perms` logic already commits.

superset/security/manager.py
```diff
@@ -2673,12 +2666,9 @@ def raise_for_ownership(self, resource: Model) -> None:
         :raises SupersetSecurityException: If the current user is not an owner
         """
 
-        # pylint: disable=import-outside-toplevel
-        from superset import db
-
         if self.is_admin():
             return
-        orig_resource = db.session.query(resource.__class__).get(resource.id)
+        orig_resource = self.get_session.query(resource.__class__).get(resource.id)
```
john-bodley, line 2671: We should use the Flask-AppBuilder session everywhere for consistency. Note that `db.session` and `self.get_session` are the same Flask-SQLAlchemy session so the logic remains unchanged.

michael-s-molina (approved):

PR #29391, opened 2024-06-27
chore(Table): Add aria-label to Table page size selector

geido:
### SUMMARY
Makes the page size selector for the table accessible to screen readers

### BEFORE/AFTER

image

### TESTING INSTRUCTIONS
1. Interact with a page size selector of a table via screen reader
2. The screen reader should be able to read the size selector

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
@@ -181,10 +181,10 @@ function SearchInput({ count, value, onChange }: SearchInputProps) {
     <span className="dt-global-filter">
       {t('Search')}{' '}
       <input
+        aria-label={t('Search %s records', count)}
         className="form-control input-sm"
         placeholder={tn('search.num_records', count)}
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -200,6 +200,9 @@ function SelectPageSize({
     <span className="dt-select-page-size form-inline">
       {t('page_size.show')}{' '}
       <select
+        aria-label={`${t('page_size.show')} ${current} ${t(
```
john-bodley, line 203: Could you provide an example of what this label resolves to?

geido: Thanks. Added a screenshot in the PR description and using the macro as suggested.

kgabryje (approved):
lgtm

PR #29395, opened 2024-06-28
feat(dashboard): update tab drag and drop reordering with positional placement and indicators for UI

rtexelm:
### SUMMARY

Change reordering and UI specifically for `Tab` components within `Tabs` parent components across the codebase.

B:

https://github.com/user-attachments/assets/c9029ae0-8db5-4f9b-88ed-264da4cda784

A:

https://github.com/user-attachments/assets/3a4767a3-23ba-485f-91eb-8325dc6cc4bf

### TESTING INSTRUCTIONS

- Create or navigate to a dashboard with tabs
- Drag tabs over other tabs to reorder

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx
@@ -105,12 +105,6 @@ const StyledTabsContainer = styled.div`
     }
   }
 
-  .ant-tabs-tab {
-    &:has(.dragdroppable--dragging) {
-      display: none;
-    }
-  }
-
   div .ant-tabs-tab-btn {
     text-transform: none;
   }
@@ -127,7 +121,7 @@ const RightDropIndicator = styled.div`
   position: absolute;
   top: 0;
   right: -4px;
-  border-radius: 4px;
+  border-radius: 2px;
 `;
 const LeftDropIndicator = styled.div`
   border: 2px solid ${({ theme }) => theme.colors.primary.base};
@@ -136,7 +130,7 @@ const LeftDropIndicator = styled.div`
   position: absolute;
   top: 0;
   left: -4px;
-  border-radius: 4px;
+  border-radius: 2px;
 `;
 
 const CloseIconWithDropIndicator = props => (
@@ -329,17 +323,8 @@ export class Tabs extends PureComponent {
 
   handleGetDropPosition(dragObject) {
     const { dropIndicator, isDraggingOver, index } = dragObject;
-    const { dragOverTabIndex } = this.state;
 
     if (isDraggingOver) {
-      console.log(
-        'dropIndicator',
-        dropIndicator,
-        'dragOverTabIndex',
-        dragOverTabIndex,
-        'index',
-        index,
-      );
       this.setState(() => ({
         dropPosition: dropIndicator,
         dragOverTabIndex: index,
@@ -445,7 +430,6 @@ export class Tabs extends PureComponent {
               type={editMode ? 'editable-card' : 'card'}
             >
               {tabIds.map((tabId, tabIndex) => (
-                // Add droppable component to each tab
                 <LineEditableTabs.TabPane
                   key={tabId}
                   tab={
--- a/superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx
@@ -157,6 +157,7 @@ export class Tabs extends PureComponent {
     this.handleDropOnTab = this.handleDropOnTab.bind(this);
     this.handleDrop = this.handleDrop.bind(this);
     this.handleGetDropPosition = this.handleGetDropPosition.bind(this);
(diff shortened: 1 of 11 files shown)


rtexelm:
Still need to add/edit the tests and refine some things but this is the general idea.

geido:
Let's include a video of before and after, please. Thank you

geido:
A couple of visual feedback:

- When dragging the "x" that closes the tab stays in the original position of the tab, shouldn't that be hidden when dragging or dragged across with the tab?
- When trying to drop on the right side of a tab, it does not catch the drop indicator until you hover the title of the tab, which is inconvenient
- The title of the tab is being highlighted in addition to the drop indicator, is that necessary?

I know some of these things were harder to implement but I feel these are still quite relevant. Wdyt @yousoph?

geido:
Also, not sure if this worked like that before, but when you try to drag a selected tab, it will auto-select the next available tab on the right.

Another issue is that as you drag and target a new tab, it will make the target tab selected.

These create a lot of unnecessary movement of the tabs and the content which can be resource-intensive when you have a bunch of charts on top of being undesirable in terms of UX.

rtexelm:
@geido @yousoph  Removing the tab respective X (delete) icon from the from the tabs bar while dragging posed a disproportionately complex problem due to the antd structure of the `closeIcon` prop. Using the `isDragging` state in it's current form to change the styling of the tab being dragged will not update to false, so the internal state of the tabs component keeps the tab un-rendered until a different tab is dragged. Because the DragDroppable is only able to wrap the tab title I added the CSS to remove that component but it can be set back so the tab title is at 20% opacity instead.

The necessity to hover over the tab title is another result of only wrapping that component in the DragDroppable.

Tab title highlighting during drag over can be changed, it was just the default behavior.

All of the tab activation behavior, i.e. the active tab changes to that which is hovered over, is the unchanged previous behavior. Wasn't sure if it was in scope to alter that as it is also the behavior of dragging charts into tabs.

yousoph:
I think it would be good to remove the title highlighting if possible! 

The active tab behavior looked ok to me - I think it's fine to keep it as it was.

geido:
> I think it would be good to remove the title highlighting if possible!
> 
> The active tab behavior looked ok to me - I think it's fine to keep it as it was.

I still think that the "x" staying at the original place is weird and should probably not happen.

rtexelm:
@geido @yousoph I tried a different approach here that I was reluctant to in past efforts. I had assumed the dnd configs would break for other dashboard elements if changed them overtly but the changes here have been pretty innocuous. The tabs should vanish during drag now and the title dropindicator should only render for anything besides tab components.

superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx
```diff
@@ -108,6 +110,38 @@ const StyledTabsContainer = styled.div`
...
+const RightDropIndicator = styled.div`
+  border: 2px solid ${({ theme }) => theme.colors.primary.base};
+  width: 5px;
+  height: 100%;
+  position: absolute;
+  top: 0;
+  right: -4px;
+  border-radius: 2px;
+`;
+const LeftDropIndicator = styled.div`
```
justinpark, line 126: The two Indicator components appear to be identical. Do we need two identical components?

If so, how about doing it this way?
```suggestion
const LeftDropIndicator = RightDropIndactor;
```

rtexelm: Much dryer, thank you for pointing that out and also supplying the emotion conducive code. I only changed your suggestion because the file is not yet in typescript.

geido (approved):
There might be some clean up possible for the way state is handled, but I think this looks good. Looking to get some feedback from @kgabryje too

kgabryje:
It look kinda weird to me that the dragged tab disappears and other tabs become active when I drag over them. Also it's working kinda slowly - it's not smooth at all on my m2pro mac, so it would be really bad on older computers.

CC @yousoph 


https://github.com/user-attachments/assets/434a858f-62d2-40a5-b43f-139155fa009f

geido (changes requested):
Due to @kgabryje's comment. I still continue to suspect that state management might be the issue here.

yousoph:
re: the active tab changing, I think that is already happening on the current version of tab dragging (before this PR)

kgabryje:
> re: the active tab changing, I think that is already happening on the current version of tab dragging (before this PR)

Oh you're right! I haven't noticed that before. Still though, I like that this PR adds the indicators of where the dragged tab is going to land, but I'm not sure about the dragged tab disappearing from its original spot. Now when we drag a tab, there's a "ghost" (low opacity) left. But it's not a blocker if you think it looks better now

geido:
@kgabryje I am not experiencing the slowness issue. If this is a go from @yousoph I think we might let this go.

geido (approved):

rtexelm:
Just wanted to drop a thanks for moving this forward while I've been away on hiatus. Let me know if there's any complications I can help with itf

PR #24060, opened 2023-05-15
fix: always allow tags to be returned via the API

hughhhh:
### SUMMARY

Going to soften the `TAGGING_SYSTEM`, and always allow tags to be returned regardless of the feature flag.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29412, opened 2024-06-29
fix(Tags filter): Filter assets by tag ID

Vitor-Avila:
### SUMMARY
https://github.com/apache/superset/pull/28570 updated `SelectTagsValue` and `tagToSelectOption`. These were also used by the Tags filter in the CRUD views (dashboards, charts and saved queries), so these filters stopped working (as they're now configured to use the `tag.id` as the value (rather than the `tag.name` which is expected by the API filter).

This PR introduces a new filter operator for these assets to be able to filter tags by ID. I decided not to update the existing filter (that filters by name) and instead create a new one, to avoid a breaking change in the API.

I also removed the `TAGGING_SYSTEM` FF check in the `SavedQueries` API, for consistency with the `Dashboard` (https://github.com/apache/superset/pull/23909) and `Chart` (https://github.com/apache/superset/pull/24060) APIs.

**Before**

https://github.com/apache/superset/assets/96086495/3c4c4694-210d-4fe5-852f-f2ed8f7e56e3

**After**

https://github.com/apache/superset/assets/96086495/87ff01e3-5abe-42cc-8761-f813be73e174

### TESTING INSTRUCTIONS
Added integration tests for both API filters.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [X] Required feature flags: `TAGGING_SYSTEM`
- [ ] Changes UI

### DIFF
--- a/superset/views/base_api.py
+++ b/superset/views/base_api.py
@@ -180,8 +180,19 @@ class BaseTagFilter(BaseFilter):  # pylint: disable=too-few-public-methods
     """ The Tag class_name to user """
     model: type[Dashboard | Slice | SqllabQuery | SqlaTable] = Dashboard
     """ The SQLAlchemy model """
+    id_based_filter = False
 
     def apply(self, query: Query, value: Any) -> Query:
+        # ID based filter
+        if self.id_based_filter:
+            tags_query = (
+                db.session.query(self.model.id)
+                .join(self.model.tags)
+                .filter(Tag.id == value)
+            )
+            return query.filter(self.model.id.in_(tags_query))
+
+        # Name based filter
         ilike_value = f"%{value}%"
         tags_query = (
             db.session.query(self.model.id)
--- a/superset/views/base_api.py
+++ b/superset/views/base_api.py
@@ -31,7 +31,6 @@ from marshmallow import fields, Schema
 from sqlalchemy import and_, distinct, func
 from sqlalchemy.orm.query import Query
 
-from superset.connectors.sqla.models import SqlaTable
 from superset.exceptions import InvalidPayloadFormatError
 from superset.extensions import db, event_logger, security_manager, stats_logger_manager
 from superset.models.core import FavStar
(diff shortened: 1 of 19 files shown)


superset/views/base_api.py
```diff
@@ -180,8 +180,19 @@ class BaseTagFilter(BaseFilter):  # pylint: disable=too-few-public-methods
     """ The Tag class_name to user """
     model: type[Dashboard | Slice | SqllabQuery | SqlaTable] = Dashboard
     """ The SQLAlchemy model """
+    id_based_filter = False
 
     def apply(self, query: Query, value: Any) -> Query:
+        # ID based filter
+        if self.id_based_filter:
```
hughhhh, line 187: Can we write a small unit test to verify the query is properly grabbing the write rows?

Vitor-Avila: Hey @hughhhh -- just added unit tests for the `BaseTagFilter` -- the `ChartTagNameFilter`, `ChartTagIdFilter`, `DashboardTagNameFilter`, `DashboardTagIdFilter`, `SavedQueryTagNameFilter` and `SavedQueryTagIdFilter` should have integration tests already 🙌 

Let me know if you have any other concerns or feedback 🙏

betodealmeida (approved):
Nice! I left a recommendation about splitting the `BaseTagFilter` class into two, if you think it's not a lot of additional work — it might make the code easier to follow.

superset/views/base_api.py
```diff
@@ -180,8 +180,19 @@ class BaseTagFilter(BaseFilter):  # pylint: disable=too-few-public-methods
     """ The Tag class_name to user """
     model: type[Dashboard | Slice | SqllabQuery | SqlaTable] = Dashboard
     """ The SQLAlchemy model """
+    id_based_filter = False
```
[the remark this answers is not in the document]
Vitor-Avila: that makes sense! I just pushed a new commit with this change. I also ended up moving these two filters to `superset.tag.filters` as I believe it makes more sense -- let me know your thoughts 🙏

PR #29414, opened 2024-06-30
chore(build): sync Jest version across plugins

hainenber:
chore(build): sync Jest version across plugins

### SUMMARY

I noticed a newly added plugin still use old Jest version v26 versus current one of v29. This PR is to rectify that and to ensure generated plugin in future will have Jest v29 as dev deps.

### TESTING INSTRUCTIONS

All CI checks pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/packages/generator-superset/generators/plugin-chart/templates/package-lock.erb
```diff
@@ -12,7 +12,7 @@
         "@airbnb/config-babel": "^2.0.1",
         "@babel/cli": "^7.16.0",
         "@types/jest": "^26.0.4",
-        "jest": "^26.6.3",
+        "jest": "^29.7.0",
```
michael-s-molina, line 15: Maybe this is a good opportunity to make Jest a peer dependency and also use `*` as the version because we want future bumps of Jest to also be applicable to plugins.

hainenber: I'm not sure if specifying Jest as peerDeps would be correct since the library is used for development only and for the plugin's operations. If there's anything akin to `peerDevDependencies`, it'd be more suitable :D

I mean, yeah, it'll be convenience for a single future Jest/Storybook/Babel bump to have an effect across plugins but I feel it's a little bit trading off correction for convenience. I'm not sure how to articulate this point, tbh

michael-s-molina: > If there's anything akin to peerDevDependencies, it'd be more suitable :D

Yeah, that would be the ideal solution. I believe that with current limitations, correction beats convenience so you can ignore my suggestion.

hainenber:
hi there folks, is this PR merge-able? Anything I need to do to proceed forward? 👀

rusackas (approved):

PR #29417, opened 2024-06-30
fix(verbose map): Correct raw metrics handling in verbose map

mcdogg17:
### SUMMARY
Previously, if metric and column names coincided, the names from the columns would overwrite the values from the metrics in the verbose map, leading to potential data discrepancies and loss of important metric verbose names. This fix corrects this behavior.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mcdogg17:
@supersetbot orglabel

mistercrunch:
Run `pre-commit install`  it'll do a series of common checks on the files you're touching in your PR pre-commit to prevent CI catching issues/failing

mcdogg17:
@mistercrunch 
Hi! I'm sorry for the long answer. I installed a pre-commit and pushed the changes. Everything should be fine now.

superset/connectors/sqla/models.py
```diff
@@ -360,10 +360,21 @@ def order_by_choices(self) -> list[tuple[str, str]]:
     def verbose_map(self) -> dict[str, str]:
         verb_map = {"__timestamp": "Time"}
         verb_map.update(
-            {o.metric_name: o.verbose_name or o.metric_name for o in self.metrics}
+            {
+                o.metric_name: verb_map.setdefault(
```
mistercrunch, line 364: not a big fan for `setdefault`, somehow it is super cryptic to me, had to read the docs and it's "get the value but set it if it doesn't exist". Ambiguous trying to be a getter and a setter at the same time. Can we go more verbose/procedural here using the normal getter/setter-type primitives?

mistercrunch (approved):

mistercrunch:
Thanks for the contribution! ❤️

PR #29420, opened 2024-06-30
fix: SQL label missing for non-group-by queries

ghost:
### SUMMARY

**Issue:** Currently, the column label is not working if the query is not an aggregate query.

**Steps to Reproduce:**

1. Create a chart with type table.
2. Set QUERY MODE to Raw records.
3. Rename any column.

image

After:
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


villebro (approved):
LGTM, thanks for the fix!

PR #29421, opened 2024-07-01
feat(echarts): add data zoom options (start/end/lock)

ghost:
### SUMMARY

Add 3 Customization Options for echarts data zoom:
- start: The end percentage of the window out of the data extent, in the range of 0 ~ 100.
- end: The end percentage of the window out of the data extent, in the range of 0 ~ 100.
- zoomLock: Specify whether to lock the size of window (selected area).

Currently, when Data Zoom is enabled in ECharts, the default start and end values are set to 0 and 100, respectively, which means all data is displayed in a single chart by default. This may not be ideal for high-density data scenarios.

By introducing these customization options, users can set the **default data range** for data zooming. This can be particularly beneficial for users with high-density data, allowing them to control the default size of the sliding window more effectively. 

For example, users can set start to 0 and end to 10, which will display the top 10% of the data by default. 
Users can still navigate through the rest of the data range by dragging the data zoom bar.

Before:
image

After:
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -32,6 +32,8 @@ import {
 import { DEFAULT_FORM_DATA } from './types';
 import { EchartsTimeseriesSeriesType } from '../Timeseries/types';
 import {
+  dataZoom,
+  dataZoomSliderOptions,
   legendSection,
   minorTicks,
   richTooltipSection,
@@ -54,7 +56,6 @@ const {
   stack,
   truncateYAxis,
   yAxisBounds,
-  zoomable,
   yAxisIndex,
 } = DEFAULT_FORM_DATA;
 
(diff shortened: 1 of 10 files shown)


rusackas:
Hmm... I've never really wished for this, I just scale down the zoom window until it's the density I want. But... I guess this seems reasonable. Two questions:
1) Do the zoom start/end settings apply to the default display of the chart? This might be a problem to some who wand to see ALL the data on the dashboard by default before they play with the slider.
2) Are the new controls dynamic (e.g. they only appear when you check the "Data Zoom" checkbox?

I'll spin up a test env to kick the tires, but just thought I'd express those questions while I wait for it to spin up :D

hexcafe:
Hi Evan, 
Thank you for your review!

This feature allows customization of the default zoom window for a specific chart in the design phase, while still permitting users to adjust it as needed in the view phase.

> 1. Do the zoom start/end settings apply to the default display of the chart? This might be a problem to some who wand to see ALL the data on the dashboard by default before they play with the slider.

No, these settings take effect only if users set them manually. Otherwise, the default values (0 for start and 100 for end) will be used. This ensures that all data is displayed by default and maintains the current behavior.

> 2. Are the new controls dynamic (e.g. they only appear when you check the "Data Zoom" checkbox?

Yes, they are fully dynamic and only appear when the "Data Zoom" checkbox is checked.

msyavuz:
Hey, @hexcafe Do you intend to bring this to the finish line? The activity around this feature seems minimal and i am tempted to close this if no one else picks it up.

rusackas:
This PR hasn't been updated in a long time. Testing instructions are missing from the PR description. We also noticed there may be overlap with #34421 which covers similar data zoom enhancements — it would be worth coordinating with the author of that PR to avoid duplicate work. We'll mark this as draft for now — please rebase against `master`, add testing instructions, clarify the relationship to #34421, and set it back to 'ready for review' when it's ready.

rusackas:
Looks like the MixedTimeseries control panel gets zoomStart/zoomEnd/zoomLock, but I don't see MixedTimeseries/transformProps.ts reading them, so those sliders would render and do nothing on Mixed charts. That's @sebint's question above too. Can you wire those through, or hide the controls where they don't apply? Still conflicting with master and the testing instructions are empty, so I can't really kick the tires here. I'm going to close this since it's been in draft a long time, but would be happy to reopen it if you want to pick this back up.

PR #29441, opened 2024-07-01
fix: ISO code of Île-de-France in france_regions.geojson

AurelienC:
### SUMMARY
Change the value of ISO for Île-de-France from "FR-IDF\t" to "FR-IDF"


rusackas:
Thank you for this, but if you don't make a change in the jupyter notebook, the geojson you've changed here will be overwritten next time someone runs the notebook. 

Hopefully this helps, but if not, I'd be happy to update it:
https://preset.io/blog/how-and-when-and-why-to-update-apache-supersets-country-maps/

mistercrunch (approved):

mistercrunch:
Oops disregard my approval I had not seen @rusackas ' comment

rusackas (changes requested):
Blocking here just to make sure the Jupyter notebook gets updated :D

AurelienC:
Thanks, I will have a look!

PR #29471, opened 2024-07-03
fix: Reading env file when running project locally without docker

hirensoni913:
### SUMMARY

Imported load_dotenv in config.py to read the .env file

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -63,7 +63,9 @@ from superset.utils.core import is_test, NO_TIME_RANGE, parse_boolean_string
 from superset.utils.encrypt import SQLAlchemyUtilsAdapter
 from superset.utils.log import DBEventLogger
 from superset.utils.logging_configurator import DefaultLoggingConfigurator
+from dotenv import load_dotenv
 
+load_dotenv()
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
--- a/superset/config.py
+++ b/superset/config.py
@@ -41,6 +41,7 @@ from typing import Any, Callable, Literal, TYPE_CHECKING, TypedDict
 import click
 import pkg_resources
 from celery.schedules import crontab
+from dotenv import load_dotenv
 from flask import Blueprint
 from flask_appbuilder.security.manager import AUTH_DB
 from flask_caching.backends.base import BaseCache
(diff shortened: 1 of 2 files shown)


rusackas (approved):
LGTM, but pinging @mistercrunch in case he knows of some intricacy I don't :) Thanks for opening this!

rusackas:
Looks like you'll need to run the [pre-commit hook](https://superset.apache.org/docs/contributing/development/#git-hooks-1) to make the linters happy.

superset/config.py
```diff
@@ -63,7 +63,9 @@
 from superset.utils.encrypt import SQLAlchemyUtilsAdapter
 from superset.utils.log import DBEventLogger
 from superset.utils.logging_configurator import DefaultLoggingConfigurator
+from dotenv import load_dotenv
```
rusackas, line 66: Looks like you only need to put this in alphabetical order to make the linter happy.

mistercrunch (changes requested):
Could this potentially be a breaking change in some environments?

I checked that traditionally-set env vars take precedence over `.env` files here, but there may be a small risk of envrionments where:
1. a [previously unused] `.env` file exists - maybe some service template at some companies lay a default `.env` file for all services
2. `.env` file happens to set something that's not desired / causes a change of behavior

Another potential issue would be around docker-compose and the `.env` that already exist in the `docker/` folder, and just want to make sure that this shouldn't be a problem there either. I can't think of anything but could see edge cases there.

Given issues are super unlikely, I'd like to push this forward, but would like to first:
- add a note in UPDATING.md giving a heads-up to people performing upgrades. Typically they would know whether this might create a change of behavior in their env
- add a note in our docs, letting users know that we support `.env` files

More generally it would be great for our config system to support more env vars, especially for simple / str parameters, maybe for an approved list of parameter and with a clear prefix, as in `SUPERSET__SQLALCHEMY_EXAMPLES_URI`. I think this should be a matter of creating a list of params that support this, and for-loop over it while checking for env vars fitting the pattern. + maybe a bit of smart casting to also support boolean, int, maybe even json if we were feeling fancy

superset/config.py
hirensoni913: Done. Moved the line up

rusackas:
Looks like this is in need of a rebase, and there's been no reply to the above comment. I'll convert it to draft for now, but it might be closed in time if it doesn't see updates.

PR #29476, opened 2024-07-03
chore: run babel_update.sh to update po files

mistercrunch:
## Summary
Periodically we need to run this script to rebuild the list of translatable strings from static analysis of the code base.

Also adding `--no-location` to `pybabel extract` fixing merge conflicts issues we've been having around these .po files. 

Also fixing the docs builds, they've been broken.

~~About the fact that merge conflicts are really tricky around translation work in general, I was thinking it may be doable to write a simple utility that would strip all comments from po files, as they are just comments referencing where the msgid exists, and operate at that level if/when needed. Like you hit a merge conflict, strip both files from comment, diff them, and commit without comments. I think the next pybabel script run would bring back the comments.~~

~~Maybe we even want to pass a flag to pybabel to compeltely avoid adding source references as comments. We'd loose the ability to easily find where the string lives in the codebase, but for that you can just `git grep`. ~~


rusackas (approved):
LGTM! And the codebase is getting not just more maintainable, but also smaller, which is always good in my book!

mistercrunch:
Oh wow look at that ->
Screenshot 2024-07-08 at 10 46 03 AM

PR #29498, opened 2024-07-05
feat: Enable customizing the docker admin password

c-w:
### SUMMARY

This PR modifies the docker-compose setup to enable customizing the password used for the admin user. This is useful in scenarios where we want to use the docker-compose setup to spin up a self-contained superset instance but when we don't want to risk someone being able to log in as an admin with the known default credentials.

### TESTING INSTRUCTIONS

1. Configure password override: `echo 'ADMIN_PASSWORD=supersecret' > docker/.env-local`
2. Run the stack: `TAG=4.0.2 docker compose -f docker-compose-image-tag.yml up`
3. Go to http://localhost:8088 and log in using username admin and the password from step 1

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration

### DIFF
--- a/docker/docker-init.sh
+++ b/docker/docker-init.sh
@@ -37,7 +37,7 @@ Init Step ${1}/${STEP_CNT} [${2}] -- ${3}
 
 EOF
 }
-ADMIN_PASSWORD="admin"
+ADMIN_PASSWORD="${ADMIN_PASSWORD:-admin}"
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):

mistercrunch:
Thanks for the contribution!

Issue #29410, opened 2024-06-29
Download to CSV shows a special characters in Arabic


Habeeb556:
### Bug description

Hello, I am encountering an issue with special characters appearing when exporting Arabic data as a CSV from SQL Lab. Note that version 2.1.3 was working fine for me with this parameter in `superset_config.py`, but this problem started after I upgraded to version 4.01.

```
# CSV Options: key/value pairs that will be passed as argument to DataFrame.to_csv
# method.
# note: index option should not be overridden
CSV_EXPORT = {"encoding": "'utf-8-sig'"}
```

### How to reproduce the bug

1- Run query.
2- Export the query results to a CSV file.

### Screenshots/recordings

![image](https://github.com/apache/superset/assets/35889071/db4bc286-f905-43da-9f38-93ccfc5bfccd)

### Superset version

4.0.1

### Python version

3.11

### Node version

I don't know

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

Habeeb556:
@dosu

Where I should add this part of the code?

`csv_data = df_to_escaped_csv(df, index=False, encoding='utf-8', **config["CSV_EXPORT"])`

Habeeb556:
@dosu

Where is this class file path?

Habeeb556:
This did not solve the problem. I am still facing the same issue with version 4.01, while version 2.1.3 did not have this problem. I am unsure if it is related to the Python version or something else.

Habeeb556:
The issue has been resolved by downgrading the package with the following command: `pip install Werkzeug==2.3.8`.

Issue #29224, opened 2024-06-12
Superset 4.0.1 - Exported CSV Contains Garbled Chinese Characters and Numbers


glenn-gj-liang:
### Bug description

I have installed the latest version of Superset (4.0.1). However, when exporting query results to a CSV file, the Chinese characters and numbers in the file are always garbled. I have tried changing the encoding to utf-8, utf-8-sig, and gbk, but the issue persists.

### How to reproduce the bug

Install Superset 4.0.1.
Run a query that includes Chinese characters and numbers.
Export the query results to a CSV file.
Open the CSV file and observe that the Chinese characters and numbers are garbled.

### Superset version

4.0.1

### Python version

3.11

### Node version

20.12.2

### Browser

Chrome

### Additional context

Expected Behavior:
The Chinese characters and numbers should be displayed correctly in the exported CSV file.

Actual Behavior:
The Chinese characters and numbers are garbled in the exported CSV file.

Environment:

Superset version: 4.0.1
Encoding attempted: utf-8, utf-8-sig, gbk
Additional Context:
Please provide any additional information that might help in diagnosing the 

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
Related issue linked above ^^^

Might even be considered a duplicate, really, but they seem _kind of_ unique.

TommyLeung-gj:
No ,It still doesn't work after I changed encoding = "utf-8"

TommyLeung-gj:
maybe there is something wrong in the process of async query ,  the celery or redis resultbackend cache

bionexit:
+1

Habeeb556:
+2

even I added the below configuration to `superset_config.py`

```
# CSV Configuration

# CSV Options: key/value pairs that will be passed as argument to DataFrame.to_csv
# method.
# note: index option should not be overridden
CSV_EXPORT = {"encoding": "'utf-8-sig'"}
```

Habeeb556:
+2

Having the same issue with `VERSION 4.01` even `VERSION 2.1.3` WORKING FINE WITH THESE PARAMETER CONFIG.

Habeeb556:
[...]

Issue #28623, opened 2024-05-21
superset 4.0.1 uses werkzeug 3.0.1 with CVE => should update to 3.0.3 or newer


nigzak:
### Bug description

the current 4.0.1 image has a high CVE finding with werkzeug
CVE-2024-34069⁠

There should be made a update to 3.0.3 (or newer)

### How to reproduce the bug

download 4.0.1 dockerhub image
scan image

### Superset version

4.0.1

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

michael-s-molina:
Hi @nigzak. Please follow the instructions on [Reporting Security Vulnerabilities](https://superset.apache.org/docs/security/#reporting-security-vulnerabilities). There's also this warning in the issue template:

🚨 Reporting a security issue: send an email to [security@superset.apache.org](mailto:security@superset.apache.org). DO NOT USE GITHUB ISSUES TO REPORT SECURITY PROBLEMS.

PR #29506, opened 2024-07-06
fix: Download to CSV shows a special Chinese characters

Habeeb556:
fix: Download to CSV shows a special Chinese characters.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #29224, #29410
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -378,7 +378,7 @@ vine==5.1.0
     #   kombu
 wcwidth==0.2.13
     # via prompt-toolkit
-werkzeug==3.0.3
+werkzeug==2.3.8
     # via
     #   -r requirements/base.in
     #   flask


mistercrunch:
Let's not downgrade every single installs because of a particular issue that's important to few. Is there a werkzeug-specific documented issue for this? I'd much rather push forward rather then stepping back.  

If we think this is important enough to prevent certain version, the proper way to do it would be to change the range in pyproject.toml and add a comment right above where we set the range explaining why the upper bound is in effect, ideally linking to a werkzeug-hosted issue.

Habeeb556:
> Let's not downgrade every single installs because of a particular issue that's important to a few, as it's fair to assume that many others may need the progress that exists since that version was published. Is there a werkzeug-specific documented issue for this? I'd much rather push forward rather then stepping back.
> 
> If we think this is important enough to prevent certain version, the proper way to do it would be to change the range in pyproject.toml and add a comment right above where we set the range explaining why the upper bound is in effect, ideally linking to a werkzeug-hosted issue.

I agree with you; I think it's related to non-English characters. I have opened a bug issue case with Werkzeug to check on this: https://github.com/pallets/werkzeug/issues/2922. This bug appeared starting from `version 3.x`.

mistercrunch:
Closing for now, happy to push Werkzeug dependency forward when the fix is in

Habeeb556:
Hi again, I have reproduced the issue in both Superset v4 and v5 and traced it to `Werkzeug 3.x` and v2. The problem happens when exporting CSVs with utf-8-sig (used for non-English characters (e.g., Arabic, Chinese, etc.). I have opened a https://github.com/pallets/werkzeug/issues/3057 after reviewing their 3.x changelog — looks like the enforced UTF-8 behavior (from PR https://github.com/pallets/werkzeug/issues/2602) is the cause.

While downgrading to `Werkzeug==2.3.8` resolves the issue, it's not ideal — that version was flagged in issue #28623 due to a CVE, so relying on it long-term is risky.

davidism:
The only thing the change in Werkzeug did was remove the ability to configure anything but utf-8 for automatic encoding. You can still encode text manually, passing the exact bytes you want to respond with: `Response(text.encode(...))`

Habeeb556:
> The only thing the change in Werkzeug did was remove the ability to configure anything but utf-8 for automatic encoding. You can still encode text manually, passing the exact bytes you want to respond with: `Response(text.encode(...))`

Thanks @davidism, I already do this manually encodes the CSV string before returning it in the response. Since **Werkzeug 3** removed support for automatic encoding configuration, I update `superset/commands/sql_lab/export.py`

```
# Manual encoding using the specified encoding (default to utf-8 if not set)
csv_string = csv.df_to_escaped_csv(df, index=False, **config["CSV_EXPORT"])
csv_data = csv_string.encode(config["CSV_EXPORT"].get("encoding", "utf-8"))
```

To replace:

`csv_data = csv.df_to_escaped_csv(df, index=False, **config["CSV_EXPORT"])`

And it passed successfully.

PR #29511, opened 2024-07-07
docs: cleanup markdown warnings

CodeWithEmad:
This will clean the markdown warnings in the docs:

- Lists should be surrounded by blank lines.
- Fenced code blocks should have a language specified.
- Multiple consecutive blank lines.
- No trailing punctuation in heading
- Headings should be surrounded by blank lines.
- Fenced code blocks should be surrounded by blank lines.

### DIFF
--- a/docs/docs/contributing/contributing.mdx
+++ b/docs/docs/contributing/contributing.mdx
@@ -18,9 +18,9 @@ which can be joined by anyone):
 - [Superset Community Calendar](https://superset.apache.org/community)
 
 More references:
+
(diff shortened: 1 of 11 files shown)


rusackas:
Love it! Running CI. 

I also wonder if it's possible to add eslint config for this, so that they're automatically kept clean from these sorts of things.

CodeWithEmad:
It might get a bit challenging since some warnings like inline HTML are still there and I didn't want to touch them (do we want to remove them all?) but it would be interesting.

mistercrunch (approved):
Yes!

PR #29517, opened 2024-07-08
fix(plugins): Fix dashboard filter for Table and Big Number with Time Comparison

Antonio-RiveroMartnez:
### SUMMARY
When a dashboard is using the Table viz or the BigNumber viz with Time comparison, any time filter used in the dashboard is ignored by the charts. This PR address that fix and prevent an index out of range too.

https://github.com/apache/superset/assets/38889534/fe2aebb7-d61c-4ee5-9313-76e583305ec6

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/common/query_context_processor.py
+++ b/superset/common/query_context_processor.py
@@ -592,8 +592,9 @@ class QueryContextProcessor:
 
             if time_grain:
                 # move the temporal column to the first column in df
-                col = df.pop(join_keys[0])
-                df.insert(0, col.name, col)
+                if join_keys:
+                    col = df.pop(join_keys[0])
+                    df.insert(0, col.name, col)
 
                 # removes columns created only for join purposes
                 df.drop(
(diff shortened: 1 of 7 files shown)


superset/common/query_context_processor.py
```diff
@@ -592,8 +592,9 @@ def join_offset_dfs(
 
             if time_grain:
                 # move the temporal column to the first column in df
-                col = df.pop(join_keys[0])
-                df.insert(0, col.name, col)
+                if join_keys:
+                    col = df.pop(join_keys[0])
+                    df.insert(0, col.name, col)
```
betodealmeida, line 597: Is there always going to be a single join key? Should we future-proof this by doing:

```suggestion
                for join_key in join_keys:
                    col = df.pop(join_key)
                    df.insert(0, col.name, col)
```

betodealmeida (approved):
Nice!

Issue #28446, opened 2024-05-13
4.0 - Imported charts shows error on embedding


ss-ravi:
### Bug description

「Guest user cannot modify chart payload」 Error is displayed for imported charts.
It's related to this issue, but with a specific reproduction flow. (https://github.com/apache/superset/issues/27177)

### How to reproduce the bug

1. Create 2 superset 4.0 instances
2. Create a dashboard with a few charts (tested with "big-number" chart)
3. Export the dashboard and import it into the other superset instance
4. Try to embed the dashboard (using the sdk) into a web app
5. 「Guest user cannot modify chart payload」Error is displayed for each chart

### Screenshots/recordings

![image](https://github.com/apache/superset/assets/91879999/b4863ea6-ad08-40bf-9dbd-e9c1c2ca727f)

### Superset version

4.0.0

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

It seems that the export/import functionality doesn't copy the `query_context` field. (It becomes `null` after import)
I believe the `query_context` became important after this PR: https://github.com/apache/superset/pull/27484/files

### Workaround
I noticed that just by opening the imported chart in the superset side updates the `query_context` and the embed starts working for that chart. We can't use this workaround as we have hundreds of charts for each of our clients.

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

ss-ravi:
I actually tried to call the update api to see if the `query_context` would be automatically created, but it wasn't. I would have to know the exact json to put there, which I don't. Even if it were doable, It still feels that the `query_context` should also be copied with the import functionality.

ss-ravi:
I also noticed a different related issue.
When you are embedding standalone charts, if the chart has no `query_context` it will try to update the chart from the iframe, and since the Public role doesn't have permission to update the chart, it ends up redirecting to the login page and you end up having iframes with login pages showing on them.

rusackas:
@betodealmeida am I taking crazy pills or does this sound like the issue you fixed in unblocking 4.0?

lindenh:
[...]

PR #29530, opened 2024-07-09
fix: prevent guest users from changing columns

betodealmeida:
### SUMMARY

An extension to https://github.com/apache/superset/pull/26749/, preventing guest users from modifying adhoc columns when requesting charts.

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -177,7 +177,7 @@ def query_context_modified(query_context: "QueryContext") -> bool:
     )
 
     # compare columns and metrics in form_data with stored values
-    for key in ["metrics", "columns"]:
+    for key in ["metrics", "columns", "groupby"]:
         requested_values = {freeze_value(value) for value in form_data.get(key) or []}
         stored_values = {
             freeze_value(value) for value in stored_chart.params_dict.get(key) or []
@@ -189,7 +189,7 @@ def query_context_modified(query_context: "QueryContext") -> bool:
         queries_values = {
             freeze_value(value)
             for query in query_context.queries
-            for value in getattr(query, key) or []
+            for value in getattr(query, key, []) or []
         }
         if stored_query_context:
             for query in stored_query_context.get("queries") or []:
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -67,7 +67,6 @@ from superset.security.guest_token import (
     GuestUser,
 )
 from superset.sql_parse import extract_tables_from_jinja_sql, Table
-from superset.superset_typing import Metric
 from superset.utils import json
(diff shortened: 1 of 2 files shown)


superset/security/manager.py
```diff
@@ -145,9 +145,9 @@ def __init__(self, **kwargs: Any) -> None:
 RoleModelView.related_views = []
 
 
-def freeze_metric(metric: Metric) -> str:
+def freeze_value(metric: Metric) -> str:
```
Vitor-Avila, line 148: Super nitpicky, but maybe we want to also change the parameter name from `metric` to something generic as well (such as `value`)?

superset/security/manager.py
```diff
@@ -170,32 +170,37 @@ def query_context_modified(query_context: "QueryContext") -> bool:
...
+    )
 
-    # compare queries in query_context
-    queries_metrics = {
-        freeze_metric(metric)
-        for query in query_context.queries
-        for metric in query.metrics or []
-    }
+    # compare columns and metrics in form_data with stored values
+    for key in ["metrics", "columns"]:
```
Vitor-Avila, line 180: I think some viz types might have the columns listed as `groupby` in the `form_data`:

image

Vitor-Avila: I tested via the browser (just inspecting the request the frontend sends to the `/api/v1/chart/data` endpoint, but it's possible this structure changes until it gets to this point, so I could be wrong.

Vitor-Avila: I do see it listed as `columns` in the `queries`, tho:

image

Vitor-Avila (commented):
I left some comments -- let me know your thoughts

superset/security/manager.py
betodealmeida: D'oh, good point! I missed that!

Vitor-Avila (approved):
Thanks for the quick fixes! 🙌 LGTM!

Not a blocking comment, but I would just also add a test with a payload containing `groupby` as opposed (or maybe both) to `columns`.

rusackas:
Question from the PR review session: If the last PR that this builds on caused a lot of Issues for embedded users, will there be any additional fallout from this PR, and will we be available to address these concerns?

rusackas:
Not sure if/how this intersects with [this issue](https://github.com/apache/superset/issues/28446).

betodealmeida:
@rusackas I'm not expecting as much trouble as the first PR, since this one just extends the first one for a few more attributes.

PR #29545, opened 2024-07-10
chore(AntD5): touchup on component imports/exports, theming ListViewCard

rusackas:
### SUMMARY

This component simplifies some component imports/exports, making it so that ONLY the Superset components use the base AntD components. We should be consolidating on Superset-wrapped AntD components whenever/wherever possible and relevant.

Also, this styles the ListViewCard to inherit the AntD theme "tokens" for the "Card" and adds a pattern to override them when necessary (rather than getting into granular CSS, which adds complexity and might cause future upgrade problems).

It also adds an AntD theme provider in Storybook, so all components will automatically be handed the theme.

### TESTING INSTRUCTIONS

Tested these components in Storybook and compared to [the public storybook on `master.`](https://superset-storybook.netlify.app/).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/.storybook/preview.jsx
+++ b/superset-frontend/.storybook/preview.jsx
@@ -24,6 +24,7 @@ import thunk from 'redux-thunk';
 import { Provider } from 'react-redux';
 import reducerIndex from 'spec/helpers/reducerIndex';
 import { GlobalStyles } from '../src/GlobalStyles';
+import { AntdThemeProvider } from '../src/components/AntdThemeProvider';
 
 import 'src/theme.ts';
 import './storybook.css';
--- a/superset-frontend/.storybook/preview.jsx
+++ b/superset-frontend/.storybook/preview.jsx
@@ -24,7 +24,6 @@ import thunk from 'redux-thunk';
 import { Provider } from 'react-redux';
 import reducerIndex from 'spec/helpers/reducerIndex';
 import { GlobalStyles } from '../src/GlobalStyles';
-import { AntdThemeProvider } from '../src/components/AntdThemeProvider';
 
 import 'src/theme.ts';
 import './storybook.css';
--- a/superset-frontend/src/components/Card/index.tsx
+++ b/superset-frontend/src/components/Card/index.tsx
@@ -35,4 +35,6 @@ const Card = ({ padded, ...props }: CardProps) => (
   />
 );
 
-export default Card;
+export default Object.assign(Card, {
+  Meta: AntdCard.Meta,
+});
--- a/superset-frontend/src/components/Card/index.tsx
+++ b/superset-frontend/src/components/Card/index.tsx
@@ -28,6 +28,8 @@ const Card = ({ padded, ...props }: CardProps) => (
   <AntdCard
     {...props}
     css={(theme: SupersetTheme) => ({
+      // 'border-radius': `${theme.gridUnit}px`,
+      border: `1px solid ${theme.colors.grayscale.light2}`,
       '.antd5-card-body': {
         padding: padded ? theme.gridUnit * 4 : theme.gridUnit,
       },
--- a/superset-frontend/src/components/ListViewCard/index.tsx
+++ b/superset-frontend/src/components/ListViewCard/index.tsx
@@ -191,10 +191,6 @@ function ListViewCard({
   const Link = url && linkComponent ? linkComponent : AnchorLink;
   const theme = useTheme();
   return (
-    // <AntdThemeProvider theme={Object.assign(AtndTheme, {
-    //   Card:
-    //   colorBgContainerbackground: theme.grayscale.light4
-    // }}>
     <ConfigProvider theme={listViewCardTheme}>
       <StyledCard
         data-test="styled-card"
--- a/superset-frontend/src/components/ListViewCard/index.tsx
+++ b/superset-frontend/src/components/ListViewCard/index.tsx
@@ -20,8 +20,10 @@ import { ReactNode, ComponentType, ReactElement, FC } from 'react';
 import { styled, useTheme } from '@superset-ui/core';
 import { Skeleton, Card } from 'src/components';
 import { Tooltip } from 'src/components/Tooltip';
+import { theme as supersetTheme } from 'src/preamble';
 import ImageLoader, { BackgroundPosition } from './ImageLoader';
 import CertifiedBadge from '../CertifiedBadge';
+import { ConfigProvider } from 'antd-v5';
 
 const ActionsWrapper = styled.div`
   width: 64px;
@@ -29,6 +31,16 @@ const ActionsWrapper = styled.div`
   justify-content: flex-end;
 `;
 
+// Styling part 1: Override Card tokens when possible
+const listViewCardTheme = {
+  components: {
+    Card: {
+      colorBgContainer: supersetTheme.colors.grayscale.light5,
+    },
+  },
+};
+
+// Styling part 2: Use CSS when necessary
 const StyledCard = styled(Card)`
   ${({ theme }) => `
     overflow: hidden;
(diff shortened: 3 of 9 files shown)


superset-frontend/.storybook/preview.jsx
```diff
@@ -23,6 +23,7 @@ import thunk from 'redux-thunk';
 import { Provider } from 'react-redux';
 import reducerIndex from 'spec/helpers/reducerIndex';
 import { GlobalStyles } from '../src/GlobalStyles';
+import { AntdThemeProvider } from '../src/components/AntdThemeProvider';
```
rusackas, line 26: @geido this seems to be the way to bring the AntdThemeProvider into all Storybook components... it doesn't seem to cause any issues.

superset-frontend/src/components/Button/index.tsx
```diff
@@ -26,7 +26,7 @@ import {
 
 import { mix } from 'polished';
 import cx from 'classnames';
-import { AntdButton } from 'src/components';
+import { Button as AntdButton } from 'antd';
```
rusackas, line 29: I think our base components should be using the AntD component directly, not the re-exported version. In fact we shouldn't re-export these AntD components unless they're really needed. We should be using our Superset component if/when it exists, for consistency.

superset-frontend/src/components/Card/index.tsx
```diff
@@ -28,11 +28,15 @@ const Card = ({ padded, ...props }: CardProps) => (
   <AntdCard
     {...props}
     css={(theme: SupersetTheme) => ({
+      // 'border-radius': `${theme.gridUnit}px`,
+      border: `1px solid ${theme.colors.grayscale.light2}`,
```
rusackas, line 32: @geido I think both of these lines of styling can be replaced by AntD "tokens" but this where my problem lies.... I can enable the border-radius here and make it `gridUnit * 10` and I see the difference. if I make the `token` be `borderRadius * 10` it seems to make no difference in the UI.

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -18,7 +18,7 @@
  */
 import { ReactNode, ComponentType, ReactElement, FC } from 'react';
 import { styled, useTheme } from '@superset-ui/core';
-import { Skeleton, AntdCard } from 'src/components';
+import { Skeleton, Card } from 'src/components';
```
rusackas, line 21: Using the Superset card now 🎉

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -29,19 +29,10 @@ const ActionsWrapper = styled.div`
   justify-content: flex-end;
 `;
 
-const StyledCard = styled(AntdCard)`
+const StyledCard = styled(Card)`
   ${({ theme }) => `
-    border: 1px solid ${theme.colors.grayscale.light2};
-    border-radius: ${theme.gridUnit}px;
```
rusackas, line 35: moving these two styles to the base component for consistency.

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -29,19 +29,10 @@ const ActionsWrapper = styled.div`
...
     overflow: hidden;
 
-    .ant-card-body {
-      padding: ${theme.gridUnit * 4}px
-        ${theme.gridUnit * 2}px;
-    }
-    .ant-card-meta-detail > div:not(:last-child) {
-      margin-bottom: 0;
-    }
```
rusackas, line 44: these styles don't appear to be used/necessary.

superset-frontend/src/components/index.ts
```diff
@@ -55,8 +57,7 @@ export {
  * or extending the components in src/components.
  */
 export {
-  Breadcrumb as AntdBreadcrumb,
-  Button as AntdButton,
```
rusackas, line 59: I don't think this is needed.

superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -138,7 +137,8 @@ const actionButtonsStyle = theme => css`
   }
 `;
 
-const StyledUndoRedoButton = styled(AntdButton)`
+const StyledUndoRedoButton = styled(Button)`
+  // TODO: check if we need this.
```
rusackas, line 141: Ugh... I don't know how much we need any of this. But that seems like an exploration for another PR. We should probably make the main Button component more robust (i.e. use props) rather than overriding styles here.

superset-frontend/src/theme/index.ts
```diff
@@ -31,7 +31,7 @@ const themes = {
 
 const baseConfig: ThemeConfig = {
   token: {
-    borderRadius: supersetTheme.borderRadius,
+    borderRadius: supersetTheme.borderRadius * 10,
```
rusackas, line 34: @geido this is the part that is baffling me. I can change this number (or add a specific one for the Card component, as I've done below) and I'm not seeing the changes in Storybook OR in Superset's UI.

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -18,30 +18,33 @@
...
+import { ConfigProvider } from 'antd-v5';
 
 const ActionsWrapper = styled.div`
   width: 64px;
   display: flex;
   justify-content: flex-end;
 `;
 
-const StyledCard = styled(AntdCard)`
+// Styling part 1: Override Card tokens when possible
```
rusackas, line 34: @geido this works!

superset-frontend/src/components/Card/index.tsx
geido, line 32:
```suggestion
```

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -188,98 +191,105 @@ function ListViewCard({
...
-                    width: Math.trunc(theme.gridUnit * 62.5),
-                  }}
-                />
-                <div className="card-actions">
-                  <Skeleton.Button active shape="circle" />{' '}
-                  <Skeleton.Button
+    // <AntdThemeProvider theme={Object.assign(AtndTheme, {
+    //   Card:
+    //   colorBgContainerbackground: theme.grayscale.light4
+    // }}>
```
geido, line 197:
```suggestion
```

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -188,98 +191,105 @@ function ListViewCard({
...
+    <ConfigProvider theme={listViewCardTheme}>
```
geido, line 198: This will always override the Card theme though. If we wanted a dark theme in the future, this would always be forced to `supersetTheme.colors.grayscale.light5`. So I guess we go back to considering the theme in Superset to be the controller, so that `supersetTheme.colors.grayscale.light5` will need to be changed to something that works on a dark theme, which begs the question "Will this work everywhere?". Hardly so, as there is no specification as of where light5 should be used in the design system I think.

geido: However, this is still better than custom styles as we will likely be targeting class names which should be more sensible to changes than theme configs.

geido (approved):

superset-frontend/src/components/Card/index.tsx
rusackas: Oh, actually, there's probably a token we should be using for this!

superset-frontend/src/components/ListViewCard/index.tsx
rusackas: Agreed... using this config provider here is specifically to append/override styles for _that instance_ of the Card component. Other Cards higher up in the React/DOM tree will be left alone.

I agree that `light4` is going to be problematic, as sometimes that would actually be a dark color needed. Part of the theming SIP is to add "surface" colors (if I'm remembering the name right) and foreground/text colors, which would be more context-based and less shade-based. We'll have to replace these in a million places in the codebase, but doing it on AntD tokens will be a whole lot easier than doing it in CSS style overrides.

rusackas:
@dosu-bot any idea why my tests are failing? I get this kind of error several times:

```
FAIL src/pages/DashboardList/DashboardList.test.jsx
  ● Test suite failed to run

    Domain `undefined` was not found.

      87 |
      [88](https://github.com/apache/superset/actions/runs/10097447336/job/27922326057?pr=29545#step:13:89) |   translate(input: string, ...args: unknown[]): string {
    > [89](https://github.com/apache/superset/actions/runs/10097447336/job/27922326057?pr=29545#step:13:90) |     return this.i18n.translate(input).fetch(...args);
         |                                       ^
      [90](https://github.com/apache/superset/actions/runs/10097447336/job/27922326057?pr=29545#step:13:91) |   }
      91 |
      92 |   translateWithNumber(key: string, ...args: unknown[]): string {

      at Jed.dcnpgettext (node_modules/jed/jed.js:249:15)
      at Chain.fetch (node_modules/jed/jed.js:148:20)
      at Translator.fetch [as translate] (packages/superset-ui-core/src/translation/Translator.ts:89:39)
      at translate (packages/superset-ui-core/src/translation/TranslatorSingleton.ts:67:24)
      at Object.<anonymous> (src/pages/DashboardList/index.tsx:73:35)
      at Object.require (src/pages/DashboardList/DashboardList.test.jsx:34:1)
```

superset-frontend/src/components/ListViewCard/index.tsx
```diff
@@ -188,98 +191,101 @@ function ListViewCard({
...
+          <Card.Meta
+            title={
+              <>
+                <TitleContainer>
+                  <Skeleton.Input
                     active
+                    size="small"
                     css={{
-                      width: theme.gridUnit * 10,
+                      width: Math.trunc(theme.gridUnit * 62.5),
```
kgabryje, line 232: Can we use full gridUnits and drop that Math.trunc?

PR #28416, opened 2024-05-09
feat: add support for catalogs

betodealmeida:
### SUMMARY

Add support for catalogs to:

- Presto
- Trino
- Snowflake
- BigQuery

We already have support for Postgres and Databricks.

@john-bodley I'm worried about the DB migration here and how it will affect AirBnB. Can you take a look? The migration needs to:

- Read the default catalog for each of these 4 DBs and update the `catalog` column in a few models (`SqlaTable`, `Query`, `SavedQuery`, `TabState`, and `TableSchema`).
- Update the `schema_perm` column for any `SqlaTable` and `Slice` models in these databases.
- Rename existing permissions of type `schema_perm`.

I'm happy to hold back on this PR until we know it's safe to merge to `master`.

### ADDITIONAL INFORMATION

- [X] Has associated issue: https://github.com/apache/superset/issues/22862
- [ ] Required feature flags:
- [ ] Changes UI
- [X] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [X] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #28394, opened 2024-05-09
feat: catalog support for Databricks native

betodealmeida:
### SUMMARY

Add support for catalogs in Databricks, and improve the migration that updates permissions and models with the catalog.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Screen Shot 2024-05-09 at 9 51 57 AM

### TESTING INSTRUCTIONS

Before running the upgrade, add a Databricks database, create a dataset, make a chart, run and save a query. They all should have no catalog information:

```sql
databricks_catalogs=# SELECT catalog FROM saved_query UNION ALL SELECT catalog FROM query UNION ALL SELECT catalog FROM tab_state UNION ALL SELECT catalog FROM table_schema UNION ALL SELECT catalog FROM tables;
 catalog
---------





(5 rows)

databricks_catalogs=# SELECT schema_perm FROM slices UNION ALL SELECT schema_perm FROM tables;      schema_perm
------------------------
 [Databricks].[default]
 [Databricks].[default]
(2 rows)

databricks_catalogs=# SELECT * FROM ab_view_menu JOIN ab_permission_view ON ab_view_menu.id = view_menu_id JOIN ab_permission ON ab_permission.id = permission_id WHERE ab_permission.name IN ('schema_access', 'catalog_access');
 id  |             name              | id  | permission_id | view_menu_id | id |      name
-----+-------------------------------+-----+---------------+--------------+----+----------------
  76 | [Databricks].[default]        | 167 |            74 |           76 | 74 | schema_access
(1 row)
```

Run the upgrade, and check that the catalog is present:

```sql
databricks_catalogs=# SELECT catalog FROM saved_query UNION ALL SELECT catalog FROM query UNION ALL SELECT catalog FROM tab_state UNION ALL SELECT catalog FROM table_schema UNION ALL SELECT catalog FROM tables;
    catalog
----------------
 hive_metastore
 hive_metastore
 hive_metastore
 hive_metastore
 hive_metastore
(5 rows)

databricks_catalogs=# SELECT schema_perm FROM slices UNION ALL SELECT schema_perm FROM tables;
               schema_perm
-----------------------------------------
 [Databricks].[hive_metastore].[default]
 [Databricks].[hive_metastore].[default]
(2 rows)

[...]

PR #29549, opened 2024-07-10
fix: make catalog migration lenient

betodealmeida:
### SUMMARY

The function `upgrade_catalog_perms` is used to add catalog-related permissions when support for catalogs is introduced in a DB engine spec:

- #28317 (Postgres)
- #28394 (Databricks)
- #28416 (BigQuery, Presto, Trino)

The migration needs to query the analytical databases in order to read the schemas, so that new permissions can be added. It might be possible that during the migration the analytical database is offline, resulting in an exception.

This PR wraps the call to `database.get_all_schema_names` in a try/except block, to prevent the migration from failing. If a given database is offline during the migration, the script will read the existing known schemas from `schema_access`, so that the migration can succeed.

N/A

### TESTING INSTRUCTIONS

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/shared/catalogs.py
+++ b/superset/migrations/shared/catalogs.py
@@ -87,6 +87,24 @@ class Slice(Base):
     schema_perm = sa.Column(sa.String(1000))
 
 
+def get_schemas(database_name: str) -> list[str]:
+    """
+    Read all known schemas from the schema permissions.
+    """
+    query = f"""
+SELECT
+    avm.name
+FROM ab_view_menu avm
+JOIN ab_permission_view apv ON avm.id = apv.view_menu_id
+JOIN ab_permission ap ON apv.permission_id = ap.id
+WHERE
+    avm.name LIKE '[{database_name}]%' AND
+    ap.name = 'schema_access';
+    """
+    # [PostgreSQL].[postgres].[public] => public
+    return sorted({row[0].split(".")[-1][1:-1] for row in op.execute(query)})
+
+
 def upgrade_catalog_perms(engines: set[str] | None = None) -> None:
     """
     Update models when catalogs are introduced in a DB engine spec.
@@ -117,16 +135,7 @@ def upgrade_catalog_perms(engines: set[str] | None = None) -> None:
         )
         add_pvms(session, {perm: ("catalog_access",)})
 
-        try:
-            upgrade_schema_perms(database, catalog, session)
-        except Exception:  # pylint: disable=broad-except
-            # Ignore exceptions, since the analytical database could be offline. It's
-            # always possible to update the schema perms manually after, by updating the
-            # database.
-            logger.info(
-                "Failed to update schema perms for database %s",
-                database.database_name,
-            )
+        upgrade_schema_perms(database, catalog, session)
 
         # update existing models
         models = [
@@ -163,11 +172,16 @@ def upgrade_schema_perms(database: Database, catalog: str, session: Session) -> 
     Rename existing schema permissions to include the catalog.
     """
     ssh_tunnel = DatabaseDAO.get_ssh_tunnel(database.id)
-    for schema in database.get_all_schema_names(
-        catalog=catalog,
-        cache=False,
-        ssh_tunnel=ssh_tunnel,
-    ):
+    try:
+        schemas = database.get_all_schema_names(
+            catalog=catalog,
+            cache=False,
+            ssh_tunnel=ssh_tunnel,
+        )
+    except Exception:  # pylint: disable=broad-except
+        schemas = get_schemas(database.database_name)
+
+    for schema in schemas:
         perm = security_manager.get_schema_perm(
             database.database_name,
             None,
@@ -199,16 +213,7 @@ def downgrade_catalog_perms(engines: set[str] | None = None) -> None:
         if catalog is None:
             continue
 
-        try:
-            downgrade_schema_perms(database, catalog, session)
-        except Exception:  # pylint: disable=broad-except
-            # Ignore exceptions, since the analytical database could be offline. It's
-            # always possible to update the schema perms manually after, by updating the
-            # database.
-            logger.info(
-                "Failed to update schema perms for database %s",
-                database.database_name,
-            )
+        downgrade_schema_perms(database, catalog, session)
 
         # update existing models
         models = [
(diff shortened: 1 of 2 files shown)


superset/migrations/shared/catalogs.py
```diff
@@ -183,25 +199,16 @@ def downgrade_catalog_perms(engines: set[str] | None = None) -> None:
...
-        ssh_tunnel = DatabaseDAO.get_ssh_tunnel(database.id)
-        for schema in database.get_all_schema_names(
-            catalog=catalog,
-            cache=False,
-            ssh_tunnel=ssh_tunnel,
-        ):
-            perm = security_manager.get_schema_perm(
+        try:
+            downgrade_schema_perms(database, catalog, session)
+        except Exception:  # pylint: disable=broad-except
```
sadpandajoe, line 204: What happens if we can't downgrade but the app still downgrades? I'm assuming the user won't experience any errors since catalog_perms may exist but nothing is calling it.

betodealmeida: Hmm, good point.

The biggest problem with that is not the catalog permissions, but the schema permissions. For example, imagine we have a single table `catalog1.schema1.table1`, and a dataset `dataset1` on it. On an upgrade, the migration will:

1. Create the perm `catalog access on catalog1`
2. Rename the perm `schema access on schema1` to `schema access on catalog1.schema1`
3. Rename `dataset1.schema_perms` from `[database1].[schema1]` to `[database1].[catalog1].[schema1]`.

If we do a downgrade and it fails connecting to the analytical database, we would still have the new perms (1 and 2), but 3 would be reverted.

Now the question is, "would a user with access to `schema1` before the downgrade still be able to access `table1`"? And the answer is NO, because we do a string comparison to see if the dataset permissions match any of the user's DAR permissions.

So it seems to me that we need to change the downgrade to rename existing perms even if it can't connect to the analytical database. I'll change the logic so that if we can't get the list of schemas we update all the known schemas (by reading them from current perms, both on upgrade and downgrade). This is a much better solution.

betodealmeida:
@sadpandajoe:

- [X] Updated description with new approach
- [X] Added unit test

Antonio-RiveroMartnez (approved):
lgtm

michael-s-molina:
@gooroodev please review

gooroodev:
## Please double-check what I found in the pull request:


| 🐞Mistake | 🤪Typo | 🚨Security | 🚀Performance | 💪Best Practices | 📖Readability | ❓Others |
|-----------|--------|------------|---------------|------------------|---------------|----------|
| 0         | 0      | 0          | 0             | 1                | 1             | 0        |

### Summary of Proposed Changes
* 🛠️ Added `Session` import from `sqlalchemy.orm`.
* 🛠️ Introduced `get_schemas` function to read all known schemas from schema permissions.
* 🛠️ Refactored `upgrade_catalog_perms` to use `upgrade_schema_perms` for schema permission updates.
* 🛠️ Added `upgrade_schema_perms` function to handle schema permission updates, including a fallback if the database connection fails.
* 🛠️ Refactored `downgrade_catalog_perms` to use `downgrade_schema_perms` for schema permission updates.
* 🛠️ Added `downgrade_schema_perms` function to handle schema permission downgrades.
* ➕ Added unit test `test_upgrade_catalog_perms_graceful` to ensure the upgrade process is lenient when the database connection fails.

### Identified Issues
| ID  | Type             | Details                                                                 | Severity | Confidence |
|-----|------------------|-------------------------------------------------------------------------|----------|------------|
| 1   | 💪Best Practices | Broad exception handling in `upgrade_schema_perms` and `downgrade_schema_perms` | 🟠Medium | 🟠Medium   |
| 2   | 📖Readability    | Inline SQL query in `get_schemas` could be more readable with better formatting | 🟡Low    | 🟡Low      |

#### Issue 1: Broad exception handling in `upgrade_schema_perms` and `downgrade_schema_perms`
**Explanation:**
Broad exception handling can mask other potential issues and make debugging difficult.

**File and Lines:**
- `superset/migrations/shared/catalogs.py`, lines 145-147 and 273-275

**Suggested Fix:**
```python
def upgrade_schema_perms(database: Database, catalog: str, session: Session) -> None:
    """
    Rename existing schema permissions to include the catalog.
    """
    ssh_tunnel = DatabaseDAO.get_ssh_tunnel(database.id)
    try:
        schemas = database.get_all_schema_names(
            catalog=catalog,
            cache=False,
            ssh_tunnel=ssh_tunnel,
        )
    except DatabaseConnectionError as e:  # Replace with specific exception
        logging.error(f"Database connection failed: {e}")
        schemas = get_schemas(database.database_name)

    for schema in schemas:
        perm = security_manager.get_schema_perm(
            database.database_name,
            None,
            schema,
        )
        existing_pvm = session.query(ViewMenu).filter_by(name=perm).one_or_none()
        if existing_pvm:
            existing_pvm.name = security_manager.get_schema_perm(
                database.database_name,
                catalog,
                schema,
            )

def downgrade_schema_perms(database: Database, catalog: str, session: Session) -> None:
    """
    Rename existing schema permissions to omit the catalog.
    """
    ssh_tunnel = DatabaseDAO.get_ssh_tunnel(database.id)
    try:
        schemas = database.get_all_schema_names(
            catalog=catalog,
            cache=False,
            ssh_tunnel=ssh_tunnel,
        )
    except DatabaseConnectionError as e:  # Replace with specific exception
        logging.error(f"Database connection failed: {e}")
        schemas = get_schemas(database.database_name)

    for schema in schemas:
        perm = security_manager.get_schema_perm(
            database.database_name,
            catalog,
            schema,
        )
        existing_pvm = session.query(ViewMenu).filter_by(name=perm).one_or_none()
        if existing_pvm:
            existing_pvm.name = security_manager.get_schema_perm(
                database.database_name,
                None,
                schema,
            )
```
**Explanation of Fix:**
Replace the broad `Exception` with a more specific exception type, such as `DatabaseConnectionError`, to handle only the expected error and log the error message for better debugging.

#### Issue 2: Inline SQL query in `get_schemas` could be more readable with better formatting
**Explanation:**
The inline SQL query in `get_schemas` can be formatted for better readability.

**File and Lines:**
- `superset/migrations/shared/catalogs.py`, lines 94-101

**Suggested Fix:**
```python
def get_schemas(database_name: str) -> list[str]:
    """
    Read all known schemas from the schema permissions.
    """
    query = f"""
    SELECT
        avm.name
    FROM
        ab_view_menu avm
    JOIN
        ab_permission_view apv ON avm.id = apv.view_menu_id
    JOIN
        ab_permission ap ON apv.permission_id = ap.id
    WHERE
        avm.name LIKE '[{database_name}]%' AND
        ap.name = 'schema_access';
    """
    # [PostgreSQL].[postgres].[public] => public
    return sorted({row[0].split(".")[-1][1:-1] for row in op.execute(query)})
```
**Explanation of Fix:**
Reformat the SQL query to improve readability by aligning the clauses and keywords.

### General Review of Code Quality and Style
The proposed changes generally improve the code by making the migration process more lenient and adding tests to ensure robustness. However, the broad exception handling should be replaced with more specific exceptions to avoid masking other potential issues. Additionally, minor improvements in SQL query formatting can enhance readability.

--
I **only** arrive when I am mentioned and asked to review the pull request.
Please reply or add a reaction to this review.

sadpandajoe (approved):

betodealmeida:
The broad except is needed because it could happen with any of the 50+ drivers we support.

PR #25669, opened 2023-10-16
fix: Obscure error message on dataset creation

justinpark:
### SUMMARY
When a SupersetSecurityException error is thrown while creating a dataset, it only shows "Fater Error" error message.
This commit adds the SupersetSecurityException case that displays the error message from the statement validation error.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:
Screenshot 2023-10-16 at 4 32 05 PM

After:
Screenshot 2023-10-16 at 4 41 52 PM

### TESTING INSTRUCTIONS
Go to SQL Lab and create an invalid select statement (`(SELECT 1)` with parentheses)
Create a dataset and see the error message

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29554, opened 2024-07-10
fix(dataset): Obscure error message on creation

justinpark:
### SUMMARY
Following up #25669 and #24969
This commit adds the SupersetSecurityException case that displays the error message from the statement validation error.
(When a SupersetSecurityException error is thrown while creating a dataset, it only shows "Fater Error" error message.)

Before:
Screenshot 2023-10-16 at 4 32 05 PM

After:
Screenshot 2023-10-16 at 4 41 52 PM

### TESTING INSTRUCTIONS
Go to SQL Lab and create an invalid select statement (`(SELECT 1)` with parentheses)
Create a dataset and see the error message

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/dataset/create.py
+++ b/superset/commands/dataset/create.py
@@ -20,6 +20,7 @@ from typing import Any, Optional
 
 from flask_appbuilder.models.sqla import Model
 from marshmallow import ValidationError
+from sqlalchemy.exc import SQLAlchemyError
 
 from superset.commands.base import BaseCommand, CreateMixin
 from superset.commands.dataset.exceptions import (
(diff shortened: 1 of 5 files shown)


rusackas:
@justinpark do you still intend to merge this one?

rusackas:
Hi @justinpark\! Just following up on this one too - still on your radar? Let us know if you'd like to move forward with it or if we should close it out.

rusackas:
@justinpark pinging one more time — this still looks like a worthwhile fix, but it's gone stale and has conflicts. I'll see if I can rebase it to keep things moving forward here...

rusackas:
Dug into the red CI here @justinpark — the `raise reraise(ex)` change in `on_error` (decorators.py) is the culprit. It now feeds the original exception into every reraise constructor, so `test_delete_dataset_column_fail` comes back with `'()'` instead of `Dataset column delete failed.`. Worth scoping that to just the `SupersetSecurityException` path rather than changing the contract for all transaction-wrapped errors. I'll see if I can help here...

rusackas (approved):
Pushed the scoping fix for `on_error` so it only forwards the source message on the `SupersetSecurityException` path. CI's green now, yay! LGTM, thanks for the tests. Might just need to assess/address any lingering bot review comments.

justinpark:
The error is no longer occurring due to improvements in sqlglot parsing. I will close this PR for now and open a new one if the issue occurs again.

PR #29566, opened 2024-07-11
fix: Trino `get_columns`

betodealmeida:
### SUMMARY

This PR is a workaround to a weird bug we've encountered in Trino when adding a physical dataset.

When adding the dataset, Superset calls `get_columns` on the inspector. The Trino driver then does a check to see if the table exists, using `has_table`:

https://github.com/trinodb/trino-python-client/blob/505989a59e93f6319e8b6be89a16f81034a06404/trino/sqlalchemy/dialect.py#L176

The `has_table` method runs this query:

```sql
            SELECT "table_name"
            FROM "information_schema"."tables"
            WHERE "table_schema" = :schema
              AND "table_name" = :table
```

And here's the weird part: even though the table DOES exist, the query can return nothing. Even weirder: running a `SHOW COLUMNS FROM :schema.:table` in that case returns data.

The Trino DB engine spec already has a custom `get_indexes` method with the warning:

> Trino dialect raises NoSuchTableError in get_indexes if table is empty.

https://github.com/apache/superset/blob/84a1cd245c8a9477a081e85cd97f6e757dfe9288/superset/db_engine_specs/trino.py#L442-L462

So I suspect it's the same problem.

To fix this, I modified the Trino DB engine spec to first try the `inspector.get_columns` method, which queries the `information_schema.table`. If this fails with `NoSuchTableError` then it falls back to the `SHOW COLUMNS FROM ...` approach.

The Presto DB engine spec already implements a `get_columns` that uses `SHOW COLUMNS FROM ...`, so I moved it to `PrestoBaseEngineSpec`, so it can be used both by `PrestoEngineSpec` and `TrinoEngineSpec`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -1618,7 +1618,7 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         ]
 
     @classmethod
-    def select_star(  # pylint: disable=too-many-arguments,too-many-locals
+    def select_star(  # pylint: disable=too-many-arguments
         cls,
         database: Database,
         table: Table,
@@ -1653,14 +1653,7 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         if show_cols:
             fields = cls._get_fields(cols)
 
-        quote = engine.dialect.identifier_preparer.quote
-        quote_schema = engine.dialect.identifier_preparer.quote_schema
-        full_table_name = (
-            quote_schema(table.schema) + "." + quote(table.table)
-            if table.schema
-            else quote(table.table)
-        )
-
+        full_table_name = cls.quote_table(table, engine.dialect)
         qry = select(fields).select_from(text(full_table_name))
 
         if limit and cls.allow_limit_clause:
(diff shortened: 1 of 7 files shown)


superset/db_engine_specs/presto.py
```diff
@@ -840,211 +1043,6 @@ def get_view_names(
             results = cursor.fetchall()
             return {row[0] for row in results}
 
-    @classmethod
-    def _create_column_info(
```
betodealmeida, line 844: These methods were moved to the base class, `PrestoBaseEngineSpec`.

superset/db_engine_specs/base.py
```diff
@@ -2224,6 +2217,23 @@ def denormalize_name(cls, dialect: Dialect, name: str) -> str:
 
         return name
 
+    @classmethod
+    def quote_table(cls, table: Table, dialect: Dialect) -> str:
```
michael-s-molina, line 2221: Nice!

superset/db_engine_specs/trino.py
```diff
@@ -433,7 +437,17 @@ def get_columns(
         "schema_options", expand the schema definition out to show all
         subfields of nested ROWs as their appropriate dotted paths.
         """
-        base_cols = super().get_columns(inspector, table, options)
+        # The Trino dialect raises `NoSuchTableError` on the inspection methods when the
```
michael-s-molina, line 440: I appreciate the comments 👍🏼

michael-s-molina (approved):

Issue #27194, opened 2024-02-21
[SIP-118] Ability to assign a unique asset id to a superset dashboard / visual


hondyman:
### Motivation
We need to provide information on various classified data used in reports to Regulatory bodies globally

Description of the problem to be solved.
In large regulated industries, we have to keep track of classified data for various global reporting such a PII for GDPR, financials metrics for Dodd/Frank and other regulatory bodies. We have to be able to say where / when and how certain data sets are used inside the org. To do this we assign an asset id (UUID) to each report, process and interface we report, xfer or export data from and we manage these through Data Governance.

### Proposed Change
Superset has metadata but its not editable, users should be able to add additional metadata to reports that allows for easier identification and governance. In our case we need to add an asset id to the metadata so we can track its usage.
I have provided a logical diagram of how we would see the process working in our org. The data catalog manages all our metadata so it will have an inventory of all dashboards power bi, superset and looker each is assigned an asset id and is actively managed by DG. 

Because superset uses integers inside an environment such as Dev, QA and prod the integers in each environment can be different. The asset id does not change between environments, its a consistent id thats callable from our portal.

![Screenshot 2024-02-21 1 20 16 PM](https://github.com/apache/superset/assets/54151500/a67202da-9f39-4103-9dbe-5b4bb34c5d22)

### New or Changed Public Interfaces

Typically our Data Governance tool assigns asset ids (UUID) and we push them into the various assets such as dashboards either manually or via api. If a new metadata area was exposed we will expect this to be available via Swagger API for insert, read and update.

### New dependencies

Not Applicable

### Migration Plan and Compatibility

Not Applicable

### Rejected Alternatives

Not Applicable

webobite:
Hi @hondyman 
I would like this use case in more detail.
You are asking about a feature support which should be able to assign asset id to superset dashboard. I am not getting clear picture what would be usage of that with with your Database Governance Tool ?
Can you pls explain that part if possible ?

Thanks & Regards

hondyman:
Hi Subham
The superset id right now is an integer and we develop reports in
development and bring to QA and production. Each report will have a
separate ID in each environment, as a regulated industry we need to keep
track of all reports and processes that have confidential and personal data
for global regulatory requirements. We do this by assigning an Asset ID
(UUID) that remains static in each environment and that is documented in
our Data Governance platform, even if we upgrade or change reports. What
I'm asking for is a JSON section where organizations can put metadata that
[...]

Issue #24743, opened 2023-07-19
Asset import/export functionality, role and cross-filter are still ID-based, not uuid-based.


enzo-dechaene:
I have two Superset, one for the testing environment where I have two datasets (**indicator-monthly [id:1]** and **indicator-daily [id:2]**).
I do an asset export from the testing Superset and import on the second environment. But it seems that the id of my datasets has changed to **indicator-daily [id:1]** and **indicator-monthly with [id:2**]

If I do an export and import in the first environment, the identifier doesn't change.
I don't know the process but it seems that some content are based on uuid but not this one.

#### How to reproduce the bug

1. Create the dataset **indicator-monthly** and then the dataset **indicator-daily** on the first instance of Superset
2. Then export it with the api **/api/v1/assets/export**
3. Import the previous dataset on the new Superset with the api **/api/v1/assets/import** 
4. Finally, go to **Settings > List roles > edit a role > check permission** with datasource access

### Expected results

what you expected to happen.

![image](https://github.com/apache/superset/assets/107003976/5d10dcee-ff7a-4770-91ce-b6440f407999)

### Actual results

what actually happens.

![image](https://github.com/apache/superset/assets/107003976/5c588e4c-f583-4be4-adb1-2d3bb90b0547)

### Environment

(please complete the following information):
- installation type: **helm chart 0.6.5**
- browser type and version: **Brave Version 1.52.130 Chromium: 114.0.5735.198 (Build officiel)  (64 bits)**
- superset version: **2.1.0**
- python version: **3.8.16**
- any feature flags active:
>         dashboard_feature: |
>          FEATURE_FLAGS = {
>            "EMBEDDED_SUPERSET": True,
>            "ENABLE_TEMPLATE_PROCESSING": True,
>            "DRILL_TO_DETAIL": True,
>            "HORIZONTAL_FILTER_BAR": True,
>            "DASHBOARD_CROSS_FILTERS": True
>          }

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [x] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [x] I have reproduced the issue with at least the latest released version of superset.
- [x] I have checked the issue tracker for the same issue and I haven't found one similar.

rusackas:
Hi there... sorry this hasn't seen much activity. Can you confirm if this is still an issue in Superset 3.x?

rusackas:
[...]

PR #13352, opened 2021-02-26
fix(dashboard): Get dashboard by slug

suddjian:
### SUMMARY

Need get dashboard by slug. Can't get dashboard by slug. Override default FAB endpoint to get dashboard by slug.

Also add field `datasources`, two birds one stone.

### TEST PLAN

Tests to be written.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Changes UI
- [ ] Requires DB Migration.
- [ ] Confirm DB Migration upgrade and downgrade tested.
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29573, opened 2024-07-12
feat(api): Added uuid filed support to dataset, chart, dashboard API

dankor:
### SUMMARY
Extend API to handle uuid.

### TESTING INSTRUCTIONS

- GET/PUT/POST http://localhost:8088/api/v1/chart/{uuid} works the same way GET/PUT/POST http://localhost:8088/api/v1/chart/{id}

### ADDITIONAL INFORMATION
- [ ] Has associated issue: Fixes #29879
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -45,6 +45,7 @@ from superset.charts.filters import (
 from superset.charts.schemas import (
     CHART_SCHEMAS,
     ChartCacheWarmUpRequestSchema,
+    ChartGetResponseSchema,
     ChartPostSchema,
     ChartPutSchema,
     get_delete_ids_schema,
@@ -131,34 +132,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
     }
     class_permission_name = "Chart"
     method_permission_name = MODEL_API_RW_METHOD_PERMISSION_MAP
-    show_columns = [
-        "cache_timeout",
-        "certified_by",
-        "certification_details",
-        "changed_on_delta_humanized",
-        "dashboards.dashboard_title",
-        "dashboards.id",
-        "dashboards.json_metadata",
-        "description",
-        "id",
-        "owners.first_name",
-        "owners.id",
-        "owners.last_name",
-        "dashboards.id",
-        "dashboards.dashboard_title",
-        "params",
-        "slice_name",
-        "thumbnail_url",
-        "url",
-        "viz_type",
-        "query_context",
-        "is_managed_externally",
-        "tags.id",
-        "tags.name",
-        "tags.type",
-    ]
 
-    show_select_columns = show_columns + ["table.id"]
     list_columns = [
         "is_managed_externally",
         "certified_by",
@@ -230,6 +204,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
         "datasource_type",
         "description",
         "id",
+        "uuid",
         "owners",
         "dashboards",
         "slice_name",
@@ -255,6 +230,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
 
     add_model_schema = ChartPostSchema()
     edit_model_schema = ChartPutSchema()
+    chart_get_response_schema = ChartGetResponseSchema()
 
     openapi_spec_tag = "Charts"
     """ Override the name set for this collection of endpoints """
@@ -286,6 +262,53 @@ class ChartRestApi(BaseSupersetModelRestApi):
     }
 
     allowed_rel_fields = {"owners", "created_by", "changed_by"}
+
+    @expose("/<id_or_uuid>", methods=["GET"])
+    @protect()
+    @safe
+    @statsd_metrics
+    @event_logger.log_this_with_context(
+        action=lambda self, *args, **kwargs: f"{self.__class__.__name__}.get",
+        log_to_statsd=False,
+    )
+    def get(self, id_or_uuid: str) -> Response:
+        """Gets a chart
+        ---
+        get:
+          description: >-
+            Get a chart
+          parameters:
+          - in: path
+            schema:
+              type: string
+            name: id_or_uuid
+            description: Either the id of the chart, or its uuid
+          responses:
+            200:
+              description: Chart
+              content:
+                application/json:
+                  schema:
+                    type: object
+                    properties:
+                      result:
+                        $ref: '#/components/schemas/ChartGetResponseSchema'
+            302:
+              description: Redirects to the current digest
+            400:
+              $ref: '#/components/responses/400'
+            401:
+              $ref: '#/components/responses/401'
+            404:
+              $ref: '#/components/responses/404'
+        """
+        # pylint: disable=arguments-differ
+        try:
+            dash = ChartDAO.get_by_id_or_uuid(id_or_uuid)
+            result = self.chart_get_response_schema.dump(dash)
+            return self.response(200, result=result)
+        except ChartNotFoundError:
+            return self.response_404()
 
     @expose("/", methods=("POST",))
     @protect()
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -117,8 +117,10 @@ class DatasetRestApi(BaseSupersetModelRestApi):
     }
     list_columns = [
         "id",
+        "uuid",
         "database.id",
         "database.database_name",
+        "database.uuid",
         "changed_by_name",
         "changed_by.first_name",
         "changed_by.last_name",
@@ -153,6 +155,7 @@ class DatasetRestApi(BaseSupersetModelRestApi):
         "id",
         "database.database_name",
         "database.id",
+        "database.uuid",
         "table_name",
         "sql",
         "filter_select_enabled",
@@ -222,6 +225,7 @@ class DatasetRestApi(BaseSupersetModelRestApi):
         "columns.advanced_data_type",
         "is_managed_externally",
         "uid",
+        "uuid",
         "datasource_name",
         "name",
         "column_formats",
@@ -273,6 +277,7 @@ class DatasetRestApi(BaseSupersetModelRestApi):
     }
     search_columns = [
         "id",
+        "uuid",
         "database",
         "owners",
         "catalog",
--- a/superset/models/slice.py
+++ b/superset/models/slice.py
@@ -37,6 +37,7 @@ from sqlalchemy import (
 from sqlalchemy.engine.base import Connection
 from sqlalchemy.orm import relationship
 from sqlalchemy.orm.mapper import Mapper
+from sqlalchemy.sql.elements import BinaryExpression
 
 from superset import db, is_feature_enabled, security_manager
 from superset.legacy import update_time_range
@@ -361,9 +362,15 @@ class Slice(  # pylint: disable=too-many-public-methods
         return self.query_context_factory
 
     @classmethod
-    def get(cls, id_: int) -> Slice:
-        qry = db.session.query(Slice).filter_by(id=id_)
+    def get(cls, id_or_uuid: str) -> Slice:
+        qry = db.session.query(Slice).filter_by(id_or_uuid_filter(id_or_uuid))
         return qry.one_or_none()
+
+
+def id_or_uuid_filter(id_or_uuid: str) -> BinaryExpression:
+    if id_or_uuid.isdigit():
+        return Slice.id == int(id_or_uuid)
+    return Slice.uuid == id_or_uuid
 
 
 def set_related_perm(_mapper: Mapper, _connection: Connection, target: Slice) -> None:
--- a/superset/charts/schemas.py
+++ b/superset/charts/schemas.py
@@ -27,6 +27,7 @@ from marshmallow.validate import Length, Range
 
 from superset.common.chart_data import ChartDataResultFormat, ChartDataResultType
 from superset.db_engine_specs.base import builtin_time_grains
+from superset.tags.models import TagType
 from superset.utils import pandas_postprocessing, schema as utils
 from superset.utils.core import (
     AnnotationType,
@@ -241,6 +242,7 @@ class ChartPostSchema(Schema):
     )
     is_managed_externally = fields.Boolean(allow_none=True, dump_default=False)
     external_url = fields.String(allow_none=True)
+    uuid = fields.UUID(allow_none=True)
 
 
 class ChartPutSchema(Schema):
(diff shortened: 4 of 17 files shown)


superset/datasets/api.py
```diff
@@ -123,6 +125,36 @@ class DatasetRestApi(BaseSupersetModelRestApi):
         "schema",
         "sql",
         "table_name",
+        "always_filter_main_dttm",
```
michael-s-molina, line 128: @dpgaspar won't the payload critically increase with this change?

betodealmeida:
Hi, @dankor! Welcome to the Superset GitHub!

Can you add some context on why these fields are needed? Thanks!

superset/datasets/api.py
michael-s-molina: We have datasets with thousands of metrics and columns.

dankor:
> Hi, @dankor! Welcome to the Superset GitHub!
> 
> Can you add some context on why these fields are needed? Thanks!

Hi @betodealmeida! Sure! I'm working on establishing a robust and granular deployment process for Superset. The core issue here is similar to why we expose these UUIDs in the import/export API: we need persistent identifiers across multiple Superset instances.

You might suggest extending the import/export API instead, but we’re encountering several challenges:
-  No support for tags, roles, etc.
-  Deprecated resources aren’t cleaned up.
-  Archives are much harder to debug compared to classic CRUD operations.

What I’m really aiming for is a declarative way to manage Superset resources (let’s call it a "report-as-code" approach). Imagine having a development environment where everyone can experiment freely. Once a report is ready, it can be committed to git, reviewed, and then deployed to higher environments.

Ideally, I envision mature version of [this](https://github.com/platacard/terraform-provider-superset), but supporting all types of resources. You can check out a proof-of-concept in this [video](https://www.youtube.com/watch?v=ZSG8XRPzgUo), though no code is available, just video.

My plan is to implement this feature incrementally, starting with minor changes and aligning as closely as possible with upstream. 

Hope this makes sense! Let me know if you're open to these kinds of contributions and what concerns the community might have.

dankor:
Hey all! I'm just wondering if there are any upcoming plans or next steps here. Could you please let me what to expect?

rusackas:
It looks like CI is stuck because of some tests failing (which I'm re-running), and pre-commit checks (i.e. linting) so you will need to run the Pre-commit hooks

```
pip3 install -r requirements/integration.txt
pre-commit install
```
You can then run pre-commit manually:
```
pre-commit run --all-files
```

michael-s-molina (changes requested):
`list_columns` is used by the frontend to load the datasets/charts/dashboards pages. Projecting all column/metric information will significantly impact performance for environments that have thousands of columns/metrics. If you want to query for that information you can request them using the `q` parameter.

dankor:
Hi @rusackas ! Thanks! I fixed the linting.

@michael-s-molina Fair point. Actually, I have added this fields in order to `q` them:
```json
{
  "columns": [
    "always_filter_main_dttm",
    "cache_timeout",
    "catalog",
    "columns.advanced_data_type",
    "columns.column_name",
    "columns.description",
    "columns.expression",
    "columns.extra",
    "columns.filterable",
    "columns.groupby",
    "columns.is_active",
    "columns.is_dttm",
    "columns.python_date_format",
    "columns.type",
    "columns.verbose_name",
    "default_endpoint",
    "description",
    "extra",
    "fetch_values_predicate",
    "filter_select_enabled",
    "is_managed_externally",
    "is_sqllab_view",
    "main_dttm_col",
    "metrics.d3format",
    "metrics.description",
    "metrics.expression",
    "metrics.extra",
    "metrics.metric_name",
    "metrics.metric_type",
    "metrics.verbose_name",
    "metrics.warning_text",
    "normalize_columns",
    "offset",
    "schema",
    "sql",
    "table_name",
    "template_params",
    "uuid",
    "database.uuid"
  ],
  "filters": [
    {
      "col": "uuid",
      "opr": "eq",
      "value": "7d2a123a-1952-46e1-b9d0-4d41e1e042c1"
    }
  ]
}
```

Otherwise this fields won't be listed with `q`. 

Do you suppose to keep `list_columns` unchanged and perform two requests instead?
- Run `list_columns` with uuid filter and get id.
- Run `show_select_columns` against obtained id with all required fields.

I'd tend to have one request if possible, but I assume the front-end changes may be too big, right?

michael-s-molina:
> I'd tend to have one request if possible, but I assume the front-end changes may be too big, right?

You would need to change the requests for all pages but I'm also worried that the default implementation of the endpoint will perform these joins. 

@dpgaspar Is there a way we can achieve this without separate requests?

dankor:
I'm wondering if we can come up with some action plan here. It currently seems quite frontend-oriented. I'd like to find a common ground in order to both fulfill my use-case and contribute it upstream, so community can benefit too. What's your vision here?

superset/datasets/api.py
dankor: If it's only blocker, I don't mind to revert that. Perhaps it makes more sense to introduce a parallel API (v2) and catch up the front-end later. Please let me know, no rush here.

rusackas:
I think the main thing is that we can't impact performance. @dpgaspar is currently out of office, but may be able to provide suggestions upon his return of how to handle this at the API layer. 

I'm also wondering if this might be a precursor to a (fairly common) request to change URLs from being incremental integer IDs to UUIDs (e.g. [here](https://github.com/apache/superset/discussions/28552))

It might also tie into [this issue](https://github.com/apache/superset/issues/24743). 

Maybe we can just solve the performance concern and let this through, but at some point we need to have a SIP proposed about how/where UUIDs should be used more systemically by Superset. I think that's the real path forward - broadening the discussion to formulate a plan to use UUIDs everywhere relevant.

superset/datasets/api.py
dpgaspar: Could potentially generate huge payloads like @michael-s-molina is saying

dpgaspar:
> > I'd tend to have one request if possible, but I assume the front-end changes may be too big, right?
> 
> You would need to change the requests for all pages but I'm also worried that the default implementation of the endpoint will perform these joins (columns and metrics belong to separate tables).
> 
> @dpgaspar Is there a way we can achieve this without multiple requests?

If we include columns and metrics on a dataset payload, these results won't be paginated. So I advise making multiple requests

dankor:
Hey @betodealmeida  @rusackas @michael-s-molina  @dpgaspar !

Could you please take another look? I've reworked this PR to eliminate the need for listing calls when looking up a single item by UUID. I was inspired by the [dashboard GET](https://github.com/apache/superset/pull/13352) approach, where the dashboard API supports three different PKs: ID, slug, and UUID. I've only implemented this for the chart so far, to get your feedback on the new approach. I hope it makes more sense now. Looking forward to hearing from you.

dankor:
Hey @betodealmeida @rusackas @michael-s-molina @dpgaspar 

Just a kind reminder.  It seems you lost the PR from your radar.

superset/datasets/api.py
rusackas: Looks like this was reverted...

rusackas:
This seems like it ties in to other [proposals](https://github.com/apache/superset/issues/27194) and [efforts](https://github.com/apache/superset/pull/30398) involving using UUIDs more systemically, so I'll ping @hondyman and @mistercrunch to take a look here too, as well as @dpgaspar to take a look at the revisited/revised API approach, and @betodealmeida (one of the experts on [managing Superset assets as code](https://preset.io/blog/version-control-superset-charts-dashboards-superset/)) for a second look  :)

rusackas:
Oh, and this will require a revisit from @michael-s-molina to un-block, if his performance concerns are indeed resolved :)

dankor:
Thanks @rusackas ! I appreciate your input!

superset/charts/api.py
```diff
@@ -125,34 +126,7 @@ def ensure_thumbnails_enabled(self) -> Optional[Response]:
     }
     class_permission_name = "Chart"
     method_permission_name = MODEL_API_RW_METHOD_PERMISSION_MAP
-    show_columns = [
```
michael-s-molina, line 128: I believe this removal was not intended as it's still used by list methods?

superset/datasets/api.py
```diff
@@ -204,6 +207,7 @@ class DatasetRestApi(BaseSupersetModelRestApi):
         "columns.advanced_data_type",
         "is_managed_externally",
         "uid",
+        "uuid",
```
michael-s-molina, line 210: What's the difference between the `uid` field and `uuid`?

michael-s-molina (commented):
Thanks for addressing the comments @dankor. I left additional comments and also have some suggestions:
- Could you update the PR description to reflect the latest changes with a clear description of why the change is needed?
- Could you add tests for the change?

superset/charts/api.py
```diff
@@ -279,6 +256,53 @@ def ensure_thumbnails_enabled(self) -> Optional[Response]:
 
     allowed_rel_fields = {"owners", "created_by", "changed_by"}
 
+    @expose("/<id_or_uuid>", methods=["GET"])
```
mistercrunch, line 259: haven't dug in, but isn't there an existing method that we can modify to now support uuid while preserving the API, essentially doing a non-breaking change that supports uuid as an input, and adds new uuid fields in the output?

superset/datasets/api.py
dankor: uid is something else: https://github.com/apache/superset/blob/master/superset/models/sql_lab.py#L287

I expected you to know. :)

dankor:
Hi @michael-s-molina ! Thanks for the review. I have answered all comments.

> Could you update the PR description to reflect the latest changes with a clear description of why the change is needed?

I have updated to support uuid along with id.

> Could you add tests for the change?

Sure, I didn't add them because I didn't find similar in the similar [PR](https://github.com/apache/superset/pull/13352). So I tried to keep my PR as close to the [similar one](https://github.com/apache/superset/pull/13352) as possible in order to get it merged too.

rusackas:
Sorry, this seems to have slipped under the collective radar. Could you give it a rebase and we'll try to take a fresh look?

dankor:
Sure, done.

superset/charts/api.py
mistercrunch: oh, I checked to make sure the related objects have the same fields/schema, LGTM

superset/models/slice.py
```diff
@@ -361,11 +362,17 @@ def get_query_context_factory(self) -> QueryContextFactory:
...
     @classmethod
-    def get(cls, id_: int) -> Slice:
-        qry = db.session.query(Slice).filter_by(id=id_)
+    def get(cls, id_or_uuid: str) -> Slice:
+        qry = db.session.query(Slice).filter_by(id_or_uuid_filter(id_or_uuid))
         return qry.one_or_none()
 
 
+def id_or_uuid_filter(id_or_uuid: str) -> BinaryExpression:
+    if id_or_uuid.isdigit():
```
mistercrunch, line 371: i think this is safe... do our uuids expect dashes `-` when represented as strings? I did a bit of nerdy research and the odds of a uuid being made out of digits only is 1/34M. Though this is not an issue if we expect dashes.

dankor: OMG, I didn't expect it works:
```python
from marshmallow.fields import UUID
uuid_field = UUID()
print(uuid_field.deserialize("55018400129141141716446655440000")) 
# returns '55018400-1291-4114-1716-446655440000'
```

Nonetheless, we operate deserialized values only, so we are indeed safe.

superset/charts/api.py
```diff
@@ -186,6 +160,7 @@ def ensure_thumbnails_enabled(self) -> Optional[Response]:
         "edit_url",
         "form_data",
         "id",
+        "uuid",
```
Vitor-Avila, line 163: Heads up that the `uuid` field was already added to `list_columns` in https://github.com/apache/superset/pull/32414, so this would be redundant now. Not sure if this would cause it to return twice in the response, but worth double checking.

This is applicable to the other asset types too (not only to chart).

dankor: Haven't seen it was coming. Removed, thanks!

rusackas:
Just FYI, I set this to auto-close https://github.com/apache/superset/issues/29879, since... I think it does!

rusackas:
Still hoping to get this one through if we can... thanks for your patience :D Looks like it needs a rebase... not sure what's up with the other test failures, but hopefully the rebase clears the deck for success!

dankor:
Resolved the conflict.

mistercrunch:
Approved/triggered the CI run. @michael-s-molina since you have the "requested change" blocker, mind taking another look? 

Overall I'd love to go move towards a uuid-first approach and this is a step in the right direction. I was thinking about moving most single-`GET` endpoints to also support uuid (if id >= 8 chars, assume it's a uuid), without creating/deprecating new endpoints. In theory there's collision errors in environments with 1M+ entities, could shift to 9 or 10 digits to prevent that. But using ids has mild security risk implications and would be nice to deprecate autonumber id-lookups eventually.

rusackas:
Looks like this could use a rebase, which might resolve some CI failures if we're lucky. Only @michael-s-molina can remove the block, though.

rusackas:
Looks like this could use a rebase, which might resolve some CI failures if we're lucky. Only @michael-s-molina can remove the block, though.

michael-s-molina (approved):
@mistercrunch @rusackas Removing the blocker as the original strategy was changed. I guess we need to resolve CI issues now.

rusackas:
@dankor if you can do that rebase to remove conflicts, I think we can get this one merged more expediently now. Thanks in advance!

dankor:
@rusackas it should be rebased now.

mistercrunch:
CI issues ...

dankor:
@mistercrunch I have aligned the tests. Not sure how should I address the rest.

mistercrunch:
Loving this PR as a great first step in the direction of a wider support of moving towards uuid as the primary way to address and reference object in Superset. One deeper question is how to wire this logic for reusability to be ported to all core entities.

Clearly it's probably mergeable as is (and maybe that's where we start), but the bigger question is how we factor this logic for portability to all objects that use the `UUIDMixing` (model mixin). Thinking there might be a DAO mixin too and similar constructs, injecting the simple methods easily for many DAOs.

@dankor , curious if you'd be interested in widening the target to support all uuid entities using reusable constructs and injecting them in all the right places (all UUID-mixed-in models) or whether you think we should merge this and using it as a blueprint to refactor into wider UUID support.

dankor:
@mistercrunch I prefer to merge things bit by bit. My next step is to support GET dataset/{uuid} in the same way. If you’re okay with it, let’s merge this one and then focus on the next.

As for the full switch, my original goal was to add **uuid** support to the dashboards/charts/datasets APIs. The next PR (datasets) will cover everything I personally need, but if you’d like to go further, we should come up with a plan. I’d be happy to be involved.

mistercrunch:
Amazing, thanks for the contribution!

PR #29581, opened 2024-07-12
fix: catalog permission check

betodealmeida:
### SUMMARY

Fix a couple issues with the catalog permission check when adding a dataset:

- When listing catalogs there is an off-by-one error when including catalogs in which the user has schema access. For example, if a database `db` has 2 catalogs, `A` and `B`, if the user has access to the schema `db.A.schema1` we should show catalog `A` in the response from `get_catalogs_accessible_by_user`. The off by one error makes the method return `schema1` instead, which is then discarded because it is not in `{A, B}`.
- When fetching tables, if a catalog is not specified (either because "Allow changing catalogs" is disabled, or because the DB doesn't support catalogs) then the schema permission still needs to be built with `catalog or default_catalog`, since we need the catalog name in the built schema permission even if "Allow changing catalogs" is disabled, otherwise the schema check will fail.

### TESTING INSTRUCTIONS

WIP

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -860,7 +860,7 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             if len(parts) == 2 and default_catalog:
                 accessible_catalogs.add(default_catalog)
             elif len(parts) == 3:
-                accessible_catalogs.add(parts[2])
+                accessible_catalogs.add(parts[1])
 
         # datasource_access
         if perms := self.user_view_menu_names("datasource_access"):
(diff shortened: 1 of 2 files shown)


superset/security/manager.py
```diff
@@ -860,7 +860,7 @@ def get_catalogs_accessible_by_user(
             if len(parts) == 2 and default_catalog:
                 accessible_catalogs.add(default_catalog)
             elif len(parts) == 3:
-                accessible_catalogs.add(parts[2])
+                accessible_catalogs.add(parts[1])
```
betodealmeida, line 863: > There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

eschutho (approved):

PR #18576, opened 2022-02-03
feat: Improve state key generation for dashboards and charts

michael-s-molina:
### SUMMARY
When opening a dashboard or a chart in Explore, the URL will update with a new state key every time the page is reloaded even when the state isn't updated. This may not be scalable in high-traffic Superset deployments where it can quickly fill up disk space or exhaust the number of cache keys too fast, due to memory restrictions, making it more likely that a link you shared with someone a while ago becomes invalid.

To address this problem we should consider the request context such as user ID, resource ID, etc, to generate new keys only when the context changes. This way we can reuse existing keys but at the same time maintain the uniqueness of storage spaces for different users/dashboards/datasets/charts.

We could use HMAC to generate the same key for the same context but that would require key rotation management.

We can’t base key generation on previous keys because we have POST scenarios where the key is not present. One example would be when the user is inside a dashboard and selects “View chart in Explore”.

We can’t change the structure of the cache key to use the user and resource because this information is not present in the GET/DELETE requests and it would also require keeping backward compatibility with previous versions.

To implement the desired behavior we’ll keep a secondary mapping of user and resources with the previously generated keys. We'll store this mapping together with the entry so they’ll share the same expiration time. We’ll also keep it in sync with DELETE events. That way we can reuse the keys and keep the solution backward compatible.

Hat tip to @zhaoyongjie and @ktmud for bringing this concern during the reviews 👏🏼 

Follow-up of https://github.com/apache/superset/pull/18181

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
https://user-images.githubusercontent.com/70410625/152404373-a9680644-402c-4e48-a155-b47881f13993.mov

### TESTING INSTRUCTIONS
1 - Check that `form_data_key` is reused when the same user is accessing the same resource (dashboard, dataset, or chart)
2 - Check that `form_data_key` changes when the user, dashboard, dataset, or chart changes.
3 - Check that the values associated with a given key are correctly retrieved from the server

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29584, opened 2024-07-14
feat(frontend/hooks): replace 3rd-party BroadcastChannel with native Web API equivalence

hainenber:
feat(frontend/hooks): replace 3rd-party BroadcastChannel usage with native Web API equivalence

### SUMMARY

Implementation of  2 years old TODO :D

Safari 15.4 was released 2 years ago ([source](https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes)) and so we can utilize its native Web API to replace a 3rd-party implementation.

Credit to Eric Wong's [SO post](https://stackoverflow.com/a/78491331) for the implementation. I'm just a copy-paster after all :D

N/A

### TESTING INSTRUCTIONS

All CI checks should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/hooks/useTabId.ts
```diff
@@ -18,17 +18,13 @@
...
-  type: 'REQUESTING_TAB_ID' | 'TAB_ID_DENIED';
-  tabId: string;
-}
-
-// TODO: We are using broadcast-channel to support Safari.
-// The native BroadcastChannel API will be supported in Safari version 15.4.
-// After that, we should remove this dependency and use the native API.
-const channel = new BroadcastChannel<TabIdChannelMessage>('tab_id_channel');
+const channel: StrictBroadcastChannel<TabIdChannelMessage> =
+  new BroadcastChannel('tab_id_channel');
```
justinpark, line 27: nit: How about making the tab_id_channel value constant?

rusackas:
Thanks for tackling this! This looks like a great improvement, and long overdue! 

Is there already (or should there be) a test making sure that messages are sent/received correctly?

hainenber:
Unfortunately there's none but I can produce the expected result similar to what described in the original [PR](https://github.com/apache/superset/pull/18576) :D 

State key is kept between reloads for a dashboard/chart opened in Explore tab. Would be sufficient?


https://github.com/user-attachments/assets/f02233ef-99d1-41e0-8a7e-22e8d96ef9c5

rusackas (approved):
It might be nice to have added test coverage, but this is an improvement over the current state of affairs for the product, and moves away from package complexities toward native browser, support, so I call it a win :)

PR #29600, opened 2024-07-16
chore: Updates CHANGELOG.md with 4.0.2 data

michael-s-molina:
### SUMMARY
Updates `CHANGELOG.md` with [4.0.2](https://github.com/apache/superset/blob/4.0.2) data. It also updates the versions selector of the issue template.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/CHANGELOG/4.0.2.md
+++ b/CHANGELOG/4.0.2.md
@@ -1,0 +1,78 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Change Log
+
+### 4.0.2 (Wed Jun 26 14:26:58 2024 -0300)
+
+**Fixes**
+
+- [#28639](https://github.com/apache/superset/pull/28639) fix: adds the ability to disallow SQL functions per engine (@dpgaspar)
+- [#29367](https://github.com/apache/superset/pull/29367) fix(explore): don't respect y-axis formatting (@justinpark)
+- [#29345](https://github.com/apache/superset/pull/29345) fix(revert 27883): Excess padding in horizontal Bar charts (@michael-s-molina)
+- [#29349](https://github.com/apache/superset/pull/29349) fix(explore): restored hidden field values has discarded (@justinpark)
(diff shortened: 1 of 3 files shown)


CHANGELOG/4.0.2.md
```diff
@@ -0,0 +1,78 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Change Log
+
+### 4.0.2 (Wed Jun 26 14:26:58 2024 -0300)
+
```
sfirke, line 23:
```suggestion
Highlights of this patch release:
- New alerts & reports can now be saved, fixing [a widely-reported bug](https://github.com/apache/superset/issues/28149) introduced in 4.0.0.
- As a side effect of the fix in #29345, the y-axis ordering is now reversed on horizontal bar charts.  Users who wish to retain the ordering they had in 4.0.1 can toggle the `Y-Axis Sort Ascending` box in a chart's editor view.
```

sfirke: The commit title "fix(ar-modal): updateNotificationSettings not updating state" doesn't describe for most people that the alert & report modal now works again.  I wanted to call that bugfix out specifically.  And then note the change of behavior introduced by the other PR.  These are the two things that felt to me like admins should know them, vs. other things that are more behind-the-scenes fixes.

michael-s-molina: @sfirke I believe every fix is a highlight of a particular importance depending on if you are being affected by the fix. I believe its the responsibility of the PR author to write a good PR description, so I suggest editing the respective PRs to include more context instead of changing the template of the `CHANGELOG`.

sfirke: I was thinking of a mini-version of release notes, like https://github.com/apache/superset/blob/master/RELEASING/release-notes-4-0/README.md for 4.0.0.  I don't mean to mess with the `CHANGELOG` template; is there a place such mini release notes could go?

Not to take away from the other PRs, but objectively there are many more people asking about the alert/report fix in Slack than any other PR in this release.  And I don't expect anyone to click and read through every PR, which is what would be necessary to know about the side effect of the #29345 change with bars reordering.  I had reported that as a bug b/c I was unaware of it being a known side effect; calling it out here might stop other people from being similarly confused.  

I think it will benefit the project to call out these two, as it may reduce unnecessary GitHub issues and Slack questions.  But I will admit I don't know where that should go.

michael-s-molina: > I don't expect anyone to click and read through every PR

That's what we assume people will do with our current `CHANGELOG` template. Our `CHANGELOG` is automatically generated using the PR's titles, which I think will never be sufficient to fully understand the changes. One alternative would be to change our template to actually add meaningful descriptions that wouldn't require people to go the PRs. The problem with this is the amount of work necessary to produce such `CHANGELOG` and the possible effort duplication with well written PRs.

Considering our current template, that assumes that people need to check the PRs, I would suggest editing the poorly described PRs.

This may be a good topic for the next Town Hall.

sfirke: Okay, sounds good.  Thanks for the explanation.

eschutho (approved):

PR #29604, opened 2024-07-16
fix: schemas for upload API

betodealmeida:
### SUMMARY

Make sure `get_schema_access_for_file_upload` returns a set, so it can be used by the API.

### TESTING INSTRUCTIONS

Added unit test.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -982,7 +982,7 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
 
     def get_schema_access_for_file_upload(  # pylint: disable=invalid-name
         self,
-    ) -> list[str]:
+    ) -> set[str]:
         allowed_databases = self.get_extra().get("schemas_allowed_for_file_upload", [])
 
         if isinstance(allowed_databases, str):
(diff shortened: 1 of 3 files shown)


Vitor-Avila (approved):
LGTM!

sadpandajoe (approved):
LGTM

superset/models/core.py
```diff
@@ -993,7 +993,7 @@ def get_schema_access_for_file_upload(  # pylint: disable=invalid-name
                 self, g.user
             )
             allowed_databases += extra_allowed_databases
-        return sorted(set(allowed_databases))
```
michael-s-molina, line 996: @betodealmeida Do you know why it was sorted before?

betodealmeida: It was sorted so it can be displayed nicely in the UI, but since that's a UI-only concern we've been moving the sorting logic to the frontend. I'm going to update it next, for now this PR fixes only this because it's broken by returning a list.

michael-s-molina: Should the frontend change also happen in this PR to make it atomic? This would make the PR not only fix the list issue but also preserve the current sorted behavior.

PR #29606, opened 2024-07-16
fix: alternate method to override Antd5 style tokens

rusackas:
### SUMMARY

In another PR, I added some custom AntD theme overrides IN the ListViewCard component. This is an approach where we can keep those overrides centralized, and import/apply them in the individual components. Maybe this'll be easier to maintain?

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
now targeting the correct branch. @geido let me know if you think this looks better (moving the 'token' settings for custom Superset components alongside the "stock" AntD ones, to centralize them).

PR #29608, opened 2024-07-16
fix: set default catalog

betodealmeida:
### SUMMARY

A few more small fixes for the catalog feature, ensuring that permissions work when multi-catalog is disabled and the API passes `catalog=None` to the backend.

This PR also fixes a **ton** of integration tests that rely on side-effects from other tests.

### TESTING INSTRUCTIONS

Added and updated tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/slice.py
+++ b/superset/models/slice.py
@@ -370,6 +370,7 @@ def set_related_perm(_mapper: Mapper, _connection: Connection, target: Slice) ->
         if ds:
             target.perm = ds.perm
             target.schema_perm = ds.schema_perm
+            target.catalog_perm = ds.catalog_perm
 
 
 def event_after_chart_changed(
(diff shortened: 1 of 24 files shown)


superset/models/slice.py
```diff
@@ -370,6 +370,7 @@ def set_related_perm(_mapper: Mapper, _connection: Connection, target: Slice) ->
         if ds:
             target.perm = ds.perm
             target.schema_perm = ds.schema_perm
+            target.catalog_perm = ds.catalog_perm
```
Vitor-Avila, line 373: TIL about this function!

Vitor-Avila (approved):
Thank you so much for these fixes!

superset/models/slice.py
betodealmeida: Same!

betodealmeida:
Superseded by https://github.com/apache/superset/pull/29840.

PR #29609, opened 2024-07-16
feat: add a script to check environment software versions

mistercrunch:
quick utility to check environment requirements. Written with GPT's help, versions pretty open-ended for now but could become more specific as we go.
```
$ ./scripts/check-env.py
Status Software                  Version Found             Ideal Range               Supported Range
==============================================================================================================
✅ python                    3.10.13                   3.10.0 - 3.10.999         3.9.0 - 3.11.999
✅ npm                       10.5.2                    10.0.0 - 999.999.999      10.0.0 - 999.999.999
✅ node                      v18.20.0                  18.0.0 - 18.999.999       18.0.0 - 18.999.999
✅ docker                    26.1.4                    20.10.0 - 999.999.999     19.0.0 - 999.999.999
✅ docker-compose            2.28.1                    2.28.0 - 999.999.999      1.29.0 - 999.999.999
✅ git                       2.44.0                    2.30.0 - 999.999.999      2.20.0 - 999.999.999
✅ Memory: 16.00 GB
```

Seems it'd be useful for maintainers to point to the script and tell people to make some assertion on their environment. There's a whole class of confusing errors that can be weeded with something like this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,10 +57,11 @@ RUN npm run build-translation
 RUN rm /app/superset/translations/*/LC_MESSAGES/*.po
 RUN rm /app/superset/translations/messages.pot
 
+FROM python:${PY_VER} AS python-base
 ######################################################################
 # Final lean image...
 ######################################################################
-FROM python:${PY_VER} AS lean
+FROM python-base AS lean
 
 WORKDIR /app
 ENV LANG=C.UTF-8 \
@@ -89,6 +90,7 @@ COPY --chown=superset:superset pyproject.toml setup.py MANIFEST.in README.md ./
 # setup.py uses the version information in package.json
 COPY --chown=superset:superset superset-frontend/package.json superset-frontend/
 COPY --chown=superset:superset requirements/base.txt requirements/
+COPY --chown=superset:superset scripts/check-env.py scripts/
 RUN --mount=type=cache,target=/root/.cache/pip \
     apt-get update -qq && apt-get install -yqq --no-install-recommends \
       build-essential \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ x-superset-user: &superset-user root
 x-superset-depends-on: &superset-depends-on
   - db
   - redis
+  - superset-checks
 x-superset-volumes: &superset-volumes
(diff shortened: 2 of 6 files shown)


rusackas:
Maybe this is the right time to add a check for sufficient system and Docker memory allocation?

mistercrunch:
Yes that sounds reasonable though might require a bit more thinking as to how we structure/section things. Seems we need both a:
-  "host requirements for docker-compose environment" where we check whether the version of docker, docker-compose and resource allocation look dandy
- "requirements for a local dev env" where we just look at the current env and whether its fine

I guess ultimately its more or less the same script and its about whether we run in on the host or in-docker

mistercrunch:
@rusackas I bumped node to 20

Screenshot 2024-08-22 at 4 48 03 PM

but generally should help support PRs like this one, it's really just a single/simple utility script that should prove useful

rusackas:
Can this be triggered in `docker-bootstrap.sh` or `docker-init.sh` or something like that?

rusackas:
> @rusackas I bumped node to 20

Awesome... the repo has already been bumped everywhere else I can find... I'll reference this change in [that PR](https://github.com/apache/superset/pull/29988) for the next go-round.

rusackas (commented):
Excited to merge this... and then trying to add it to the Dockerfile (or one of the aforementioned scripts, but it _seems like_ Dockerfile is the right place). 

Rather than hard-coding all the versions in here, shall we make them more dynamic? For example, the node version could be extracted from "engines" in `package.json`

mistercrunch:
I think it belongs in `docker/docker-init.sh`, but we need to add it to the `depends_on` so all other processes depend on it. Also need to either `COPY` scripts inside docker or mount it. I started a branch but ran into all sort of other docker/npm/node/disk-space, more roadblocks-than-a-road-can-fit issues

rusackas (approved):
Oh, I'd forgotten about this! I guess it could also check for certain OS dependencies, like that `zstd` lib that has been causing trouble for some folks.

mistercrunch:
Right for os-level deps, seems the right approach would be to do in-python, with dummy/minimal function that would raise if os-level-dep is missing. 

```python
def test_ zstd():
    import  zstd
    zstd.do_something_that_will_raise_os_dep_isnt_installed()
 ```

Issue #28713, opened 2024-05-27
Problem Exporting Charts as Image


CharlesNkdl:
### Bug description

Hi Superset Community,

I have a problem exporting Charts as Image, from multiple Dashboards and instance of Superset.

### How to reproduce the bug

1. Go to a Dashboards with Charts
2. Click the three dots on the top right of a charts
3. Download => Download as Image

### Screenshots/recordings

Here is the result for my charts : 
![valeur-moyennes-selon-surface-2024-05-27T14-20-12 529Z](https://github.com/apache/superset/assets/100453363/127e9e1e-e5b7-49dc-a538-f1401d9a9ed8)

I was wondering if maybe this was related to the fact that the query is quite complicated + a lot of filters in the dashboard so i tried to export another charts from the examples datasets.

![genders-2024-05-27T14-21-50 342Z](https://github.com/apache/superset/assets/100453363/a736255c-cdd2-4469-bae8-d7e95fcd02be)

Not quite right too...

### Superset version

4.0.1

### Python version

3.10

### Node version

Not applicable

### Browser

Firefox

### Additional context

I use Apache Superset 4.0.1 in Local using Windows and a WSL Backend, but this problem was also on my prod. env with Apache Superset 4.0.0 used with Cloudron which handles the deployment.

Even if template processing is ON, i can't download image, with or without Jinja Templating in them.

Here are my superset_config.py flags in local : 
```
SECRET_KEY` = ""

SQLALCHEMY_DATABASE_URI = ''

WTF_CSRF_ENABLED = True (tried with False too...)
WTF_CSRF_EXEMPT_LIST = []
# A CSRF token that expires in 1 year
WTF_CSRF_TIME_LIMIT = 60 * 60 * 24 * 365
                                                                                                                
MAPBOX_API_KEY = ''

LANGUAGES = {
        'en': {'flag': 'us', 'name': 'English'},
        'fr': {'flag': 'fr', 'name': 'French'}
}

ENABLE_JAVASCRIPT_CONTROLS = True

APP_NAME = "Superset Local"

WELCOME_MESSAGE = "secret"

PRIMARY_COLOR = "green"

ENABLE_TEMPLATE_PROCESSING = True

CONTENT_SECURITY_POLICY_WARNING = False

TALISMAN_ENABLED = False

DEFAULT_FEATURE_FLAGS = {
        "ENABLE_TEMPLATE_PROCESSING" : True,
}
```

Here is my python stack trace, minus the get and post request : 
````
[...]

PR #29272, opened 2024-06-17
refactor(Dashboard): Fetch dashboard screenshot via dedicated endpoint

geido:
### SUMMARY
A dashboard can have an image attached in the markdown component that is hosted on a server that does not have proper CORS settings. Due to limitations of html2canvas, both PDF and image download will fail.

For reference from html2canvas:
> All the images that the script uses need to reside under the [same origin](http://en.wikipedia.org/wiki/Same_origin_policy) for it to be able to read them without the assistance of a [proxy](https://html2canvas.hertzen.com/proxy/). Similarly, if you have other canvas elements on the page, which have been tainted with cross-origin content, they will become dirty and no longer readable by html2canvas. 

This PR introduces the necessary frontend changes to defer the generation of the screenshot to a newly built endpoint. See https://github.com/apache/superset/pull/29187

### TESTING INSTRUCTIONS
1. Select a Dashboard and a specific tab
2. Download the dashboard as an image or PDF via the dedicated menu items
3. Wait for the download to complete
4. The screenshot should be of the chosen type and depict the target tab

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29614, opened 2024-07-17
Problem Exporting Charts as Image

xiaoqufengdi:
It is wrong not to display a chart if it is not in the browser window display range. When we download the dashboard as a picture, we find that some charts are not displayed

https://github.com/apache/superset/issues/28713

fix(dashboard): load charts correctly

### DIFF
--- a/superset-frontend/src/dashboard/components/gridComponents/Row.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/Row.jsx
@@ -169,16 +169,16 @@ class Row extends PureComponent {
           rootMargin: '100% 0px',
         },
       );
-      this.observerDisabler = new IntersectionObserver(
-        ([entry]) => {
-          if (!entry.isIntersecting && this.state.isInView) {
-            this.setState({ isInView: false });
-          }
-        },
-        {
-          rootMargin: '400% 0px',
-        },
-      );
+      // this.observerDisabler = new IntersectionObserver(
+      //   ([entry]) => {
+      //     if (!entry.isIntersecting && this.state.isInView) {
+      //       this.setState({ isInView: false });
+      //     }
+      //   },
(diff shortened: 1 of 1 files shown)


rusackas:
This PR disables a feature that was implemented to improve performance. We don't render all charts on a dashboard, but rather just render them when they're in the viewport. Dashboards with hundreds of charts will have poor performance with this change.

@kgabryje do you know of any way to make this change dynamic, so screenshots/alerts/reports/thumbnails/etc will work as expected? There are several GitHub issues about this, I'll try to find/link them here.

xiaoqufengdi:
> This PR disables a feature that was implemented to improve performance. We don't render all charts on a dashboard, but rather just render them when they're in the viewport. Dashboards with hundreds of charts will have poor performance with this change.
> 
> @kgabryje do you know of any way to make this change dynamic, so screenshots/alerts/reports/thumbnails/etc will work as expected? There are several GitHub issues about this, I'll try to find/link them here.




I understand the optimization, but you shouldn't hide it after the chart has been shown once.

kgabryje:
@xiaoqufengdi Thanks for reporting! PR https://github.com/apache/superset/pull/29272 that was merged yesterday should fix the issue. Could you please verify that it works for you now?

geido:
I am closing this for now as the solution won't be merged as is. Please, feel free to verify that the new changes introduced in https://github.com/apache/superset/pull/29272 have fixed the issue or feel free to re-open the PR with an appropriate solution. Thanks for your help!

geido:
I am closing this for now as the solution won't be merged as is. Please, feel free to verify that the new changes introduced in https://github.com/apache/superset/pull/29272 have fixed the issue or feel free to re-open the PR with an appropriate solution. Thanks for your help!

Issue #29546, opened 2024-07-10
[SIP-142] Improving Database Migration Management


mistercrunch:
## Motivation

Managing database migrations in Superset currently poses challenges such as deadlocks and downtime during DDL changes. We aim to improve the process to ensure smooth blue/green deployments, reduce downtime, and handle complex migrations effectively.

## Proposed Change

1. **Impose Blue/Green-Safe Migrations as part of minor releases**:
   - blue/green migration assume that the previous version of the app can ALWAYS run against the newer version of the database. This mean there cannot be destructive operations as part of a PR that contains a database migration
   - "theoritical but highly unlikely lossiness" can be accepted but understood and mitigated. For example if there's a short period of time during the migration where the old app might be updating a field that's effectively getting deprecated as part of the migration, and as a result that update/transaction might be lost in the shuffle, we can assess the likelihood of such operation and its critically, and accept it

1. **Functional Migrations:** functional programming enforces concepts around idempotency of tasks that limit side effects. For Superset migrations, we want to enforce these ideas in a way that limits mutations, and allows for running the same mutation twice. For example, when adding a new column, we probably want to check whether the new column exists, and delete / recreate it if necessary. More intricately, if we want to change the `Dashboard.json_metadata` schema, we probably want to avoid mutating it in-place, and instead create a new `Dashboard.json_metadata_v2` and have a deterministic function that take the original, and transforms it as needed before landing it in this new field. In this case, rolling back is more of a matter of pointing back to the original field than applying a reverse migration to try to recreate the original state from the old one. Note that this implies race conditions around the flickering that blue/green implies and/or around rolling back. Philosophically, we prefer this set of tradeoffs to the ones that come with biderectional mutations. In cases where lossiness is unacceptable, the solution may be to maintain both versions of the state for a time, meaning dashboard mutations post migration would keep updating `Dashboard.json_metadata` AND `Dashboard.json_metadata_v2` for a time.

1. **Accumulated Cleanup Migrations**: Defer non-critical cleanup migrations (e.g., column deletions) to major version upgrades, in a file (`superset/migration/future.py`) imposing a certain structure that's compatible with alembic-based approach. 

[...]

PR #29625, opened 2024-07-18
fix: try to prevent deadlocks when running upgrade

sadpandajoe:
### SUMMARY
When running the migration upgrade path for https://github.com/apache/superset/pull/28704, we noticed that there could be times where we see a deadlock. This PR breaks down each drop into it's own migration file.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2024-05-24_11-31_02f4f7811799_remove_sl__tables.py
```diff
@@ -24,21 +24,89 @@
...
+    sorted_tables_to_drop = get_drop_order(tables_to_drop, insp)
+
+    if db_dialect == "postgresql":
+        for table in sorted_tables_to_drop:
+            fks = get_foreign_keys(table, insp)
+            for fk in fks:
+                constraint = fk["name"]
+                if constraint:
+                    op.execute(
+                        f"ALTER TABLE {table} RENAME CONSTRAINT {constraint} TO {constraint}_old"
```
eschutho, line 92: This may have helped in my earlier PR when we were changing the constraint, but after looking into this, I don't think it will help when dropping the constraint. Maybe one idea would be to drop each table in a separate transaction and commit after each drop.

dpgaspar: careful a commit may leave a migration partially applied

sadpandajoe: ok, I revised this to just drop the constraint first and then afterwards to drop the table. Hopefully this will reduce any deadlocks there may be.

superset/migrations/versions/2024-05-24_11-31_02f4f7811799_remove_sl__tables.py
```diff
@@ -24,21 +24,41 @@
...
+    "sl_dataset_columns",
+    "sl_table_columns",
+    "sl_dataset_tables",
+    "sl_columns",
+    "sl_tables",
+    "sl_dataset_users",
+    "sl_datasets"
+]
+
+def drop_constraints(table, insp):
```
mistercrunch, line 43: This function seems useful and could be moved into `superset/migrations/shared/utils.py` so it can be reused in future migrations. I'd make both `insp` and `conn` arguments so the caller can reuse theirs and control transations.

mistercrunch:
Just a note, but in light of https://github.com/apache/superset/issues/29546, thinking in the future this type of migration would have been held up until the next major. In the meantime, it's probably ok to mitigate existing migrations to minimize deadlock/downtime.

eschutho (approved):
I ran the upgrade and downgrade a few times and no issues.

PR #29649, opened 2024-07-19
fix: remove old database constraint on the Dataset model

betodealmeida:
### SUMMARY

The dataset model has a comment that says:

https://github.com/apache/superset/blob/9487d6c9d6b3231b1bf31752c76fc50066be0014/superset/connectors/sqla/models.py#L1135-L1144

But turns out there is an old migration from 2016 that adds the uniqueness constraint to the schema:

https://github.com/apache/superset/blob/9487d6c9d6b3231b1bf31752c76fc50066be0014/superset/migrations/versions/2016-04-15_08-31_b4456560d4f3_change_table_unique_constraint.py#L36-L38

With the introduction of catalogs in [SIP-95](https://github.com/apache/superset/issues/22862) this constraint needs to either be updated to include `catalog`, or removed to align with the `SqlaTable` model. In this PR I went with the latter.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/versions/2024-07-19_16-11_df3d7e2eb9a4_remove__customer_location_uc.py
+++ b/superset/migrations/versions/2024-07-19_16-11_df3d7e2eb9a4_remove__customer_location_uc.py
@@ -1,0 +1,55 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""
+Remove _customer_location_uc
+
+Revision ID: df3d7e2eb9a4
+Revises: 02f4f7811799
+Create Date: 2024-07-19 16:11:26.740368
+"""
+
+import logging
+
+from alembic import op
+from migration_utils import create_unique_constraint, drop_unique_constraint
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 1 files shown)


superset/migrations/versions/2024-07-19_16-11_df3d7e2eb9a4_remove__customer_location_uc.py
```diff
@@ -0,0 +1,55 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def upgrade():
+    try:
+        drop_unique_constraint(op, "_customer_location_uc", "tables")
+    except Exception:  # pylint: disable=broad-except
+        # Unfortunately the DB migration that creates this constraint has a
+        # try/except block, so that we can't know for sure if the constraint exists.
```
eschutho, line 42: Should we raise again if it fails so that it rolls back? What are the implications here if there is an error and we can't roll back the transaction because we are not raising again? Can you use `generic_find_uq_constraint_name` to see if the constraint exists?

betodealmeida: Yeah, good point, I think we can use `generic_find_constraint_name`. I'm a bit concerned that the method might not work for all databases, but it should work with the most common at least.

mistercrunch:
Hey, thinking about applying some of the ideas in https://github.com/apache/superset/issues/29546, which would point towards either accumulating this PR and holding it until next release, or merging this in to a no-op place that we would bring into an actual migration later.

Curious as to whether you all think it's better to hold the PR, or to merge into a temporary place.

mistercrunch:
Easiest option here seems like it'd be to add a `hold:next-major:migration` label (or similar) and defer

betodealmeida:
Tagged this for 5.0.

betodealmeida:
Ping @mistercrunch

mistercrunch:
It would be nice to have a way to "pile-up" these no-op/cleanup migrations for scheduled downtime. I'm always scared of deadlocks and issues like that for something that's not required. 

Technically this one (and others like this one) could be run as best effort in a `try` block with a timeout one minute timeout, and run out-of-order of the alembic "rigorous", in-order chain.

It's not super important, and not required for this PR, but nice-to-have.

mistercrunch (approved):
Approving given the breaking window and the likely "downtime-required" for next major.

PR #29654, opened 2024-07-22
feat(ui): Disable user selection of data when user does not have export permissions

MialLewis:
### SUMMARY
Disables the user selection of data from the sqllab `resultTable`, `plugin-chart-pivot-table` and `plugin-chart-table` if the user does not have the `can_export_csv` / `can_csv` role. This is based off: https://github.com/apache/superset/pull/28429.

Relevant discussion: https://github.com/apache/superset/discussions/29078

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
+++ b/superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
@@ -105,6 +105,8 @@ export interface ChartPropsConfig {
   inputRef?: RefObject<any>;
   /** Theme object */
   theme: SupersetTheme;
+  /** Set chart data selectable by user **/
+  dataSelectionMode?: string;
 }
 
 const DEFAULT_WIDTH = 800;
@@ -155,6 +157,8 @@ export default class ChartProps<FormData extends RawFormData = RawFormData> {
 
   theme: SupersetTheme;
 
+  dataSelectionMode?: string;
+
   constructor(config: ChartPropsConfig & { formData?: FormData } = {}) {
     const {
       annotationData = {},
@@ -176,6 +180,7 @@ export default class ChartProps<FormData extends RawFormData = RawFormData> {
       inContextMenu = false,
       emitCrossFilters = false,
       theme,
+      dataSelectionMode,
     } = config;
     this.width = width;
     this.height = height;
@@ -198,6 +203,7 @@ export default class ChartProps<FormData extends RawFormData = RawFormData> {
     this.inContextMenu = inContextMenu;
     this.emitCrossFilters = emitCrossFilters;
     this.theme = theme;
+    this.dataSelectionMode = dataSelectionMode;
   }
 }
 
@@ -223,6 +229,7 @@ ChartProps.createSelector = function create(): ChartPropsSelector {
     input => input.inContextMenu,
     input => input.emitCrossFilters,
     input => input.theme,
+    input => input.dataSelectionMode,
     (
       annotationData,
       datasource,
@@ -243,6 +250,7 @@ ChartProps.createSelector = function create(): ChartPropsSelector {
       inContextMenu,
       emitCrossFilters,
       theme,
+      dataSelectionMode,
     ) =>
       new ChartProps({
         annotationData,
@@ -264,6 +272,7 @@ ChartProps.createSelector = function create(): ChartPropsSelector {
         inContextMenu,
         emitCrossFilters,
         theme,
+	dataSelectionMode,
       }),
   );
 };
--- a/superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
+++ b/superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
@@ -105,7 +105,7 @@ export interface ChartPropsConfig {
   inputRef?: RefObject<any>;
   /** Theme object */
   theme: SupersetTheme;
-  /** Set chart data selectable by user **/
+  /** Set chart data selectable by user */
   dataSelectionMode?: string;
 }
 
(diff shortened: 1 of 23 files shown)


MialLewis:
@rusackas I'm pinging you as you merged a related PR https://github.com/apache/superset/pull/28429.

I'm working to disable user select for data, which is required for our sensitive data. The more blockers we can put in the place of data export the better, even if they can be circumvented.

We are willing to fork `superset` to do this, but getting it merged would obviously be better! Do you think there would be any appetite for this change, willing to put behind a feature flag if necessary.

Also, any suggestions/examples of how best to use the permissions model inside the frontend plugins?

Thanks!

MialLewis:
This is ready for someone to take a look. I'm very open to feedback as this is the first time I've used react. I do have very limited time to work on this though, it's already serving our purposes as it.

Thanks!

superset-frontend/src/components/FilterableTable/index.tsx
```diff
@@ -249,11 +257,16 @@ const FilterableTable = ({
       }),
   }));
 
+  const canExportData = useSelector((state: RootState) =>
+    findPermission('can_export_csv', 'SQLLab', state.user?.roles),
+  );
+
```
justinpark, line 263: The FilterableTable is a common component, so accessing the Redux user state to check permissions is not an appropriate approach. (Additionally, if FilterableTable is used outside of SQLLab, this permission logic would be incorrect.)
Therefore, it would be more appropriate to add a prop to FilterableTable (e.g., disableTextSelection?: boolean) and include this logic in SqlLab/components/ResultSet, passing it to FilterableTable.

superset-frontend/src/components/Chart/Chart.jsx
```diff
@@ -156,6 +158,11 @@ class Chart extends PureComponent {
     super(props);
     this.handleRenderContainerFailure =
       this.handleRenderContainerFailure.bind(this);
+    this.canExportData = findPermission(
+      'can_csv',
+      'Superset',
+      this.props.user?.roles,
+    );
```
justinpark, line 165: For similar reasons [below](https://github.com/apache/superset/pull/29654/files#r1746286537), adding a user state to ChartContainer and ChartRenderer and including this kind of logic would not be flexible in terms of component reusability. Therefore, it would be more appropriate to handle this at the dashboard or explore component level. (Additionally, in the case of the findPermission logic, if there are many charts in the dashboard, repeating the same n+1 search logic for each chart in the user permission list would lead to inefficiency.)

In other words, this permission logic should be included in DashboardBuilder and passed as a property value to DashboardComponent, allowing access through `this.props.canExportData`.
Similarly, if it's included in ExploreChartPanel and passed as a property to ChartContainer, it can be accessed in the same way.

superset-frontend/packages/superset-ui-core/src/chart/models/ChartProps.ts
```diff
@@ -105,6 +105,8 @@ export interface ChartPropsConfig {
   inputRef?: RefObject<any>;
   /** Theme object */
   theme: SupersetTheme;
+  /** Set chart data selectable by user */
+  dataSelectionMode?: string;
```
justinpark, line 109: It seems more appropriate to pass a more shareable value like canExportData rather than sending specific data selection values, and then set it according to each component's needs.

```suggestion
  canExportData: boolean;
```

superset-frontend/plugins/plugin-chart-pivot-table/src/plugin/transformProps.ts
```diff
@@ -81,6 +81,7 @@ export default function transformProps(chartProps: ChartProps<QueryFormData>) {
     filterState,
     datasource: { verboseMap = {}, columnFormats = {}, currencyFormats = {} },
     emitCrossFilters,
+    dataSelectionMode,
```
justinpark, line 84:
```suggestion
    canExportData,
```

superset-frontend/plugins/plugin-chart-pivot-table/src/plugin/transformProps.ts
```diff
@@ -176,5 +177,6 @@ export default function transformProps(chartProps: ChartProps<QueryFormData>) {
     onContextMenu,
     timeGrainSqla,
     allowRenderHtml,
+    dataSelectionMode,
```
justinpark, line 180: Here is the place to pass the appropriate data selection mode value for the given viz plugin.
```suggestion
    dataSelectionMode: canExportData ? 'auto' : 'none'
```

rusackas:
Running CI. Ping us if it gets stuck!

superset-frontend/src/components/Chart/Chart.jsx
MialLewis: With regards to dashboards, there is already a property `supersetCanCSV` passed through to `dashboard/components/gridCompnents/Chart.jsx`, so we can pass that to `ChartContainer`: `canExportData={supersetCanCSV}`

MialLewis:
Thanks for the review @justinpark, I've made the changes you suggested.

@rusackas would you mind running the CI again, it wasn't quite ready last time. I've run prettier locally so hopefully that will stop the failures.

MialLewis:
@rusackas could you re-run the CI please?

I've built the js packages locally and see no test failures now. I haven't been able to run the cypress tests however as I see a `message": "400 Bad Request: The CSRF token is missing` when cypress attempts to login.

rusackas:
Running CI 🤞

rusackas:
Hi there... sorry this thing seems to have fallen under our radar. We're trying to thin the backlog of unmergeable things, so we can keep a better handle on the situation and have this happen _less_ often. I'm not sure how hard this is to rebase, but if you want to give it a go, we can try to get it merged (finally). I'll convert it to a draft in the meantime, and nowadays when you mark it ready for review, it'll get our attention again.

rusackas:
This draft hasn't been updated in >90 days, so it will be closed. Please feel free to reopen it at any time, or ping us on this thread for further help.

PR #29672, opened 2024-07-23
fix(explore): missing column autocomplete in custom SQL

justinpark:
### SUMMARY
The `keywords` field was missing in the Custom SQL editor of the column item, causing the autocomplete feature in the metrics field to be unavailable.
This commit fills in the missing keywords field in the Column component and adds styling to the tooltip section of the autocomplete feature.

Before:

https://github.com/user-attachments/assets/96afae67-0cd1-4c4f-946a-9933276ae2f5

After:

https://github.com/user-attachments/assets/c591b64d-0993-4971-8f11-9ee08ceecb2b

### TESTING INSTRUCTIONS
Open a column selector in explore
Click Custom SQL and type a column name

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/AsyncAceEditor/index.tsx
+++ b/superset-frontend/src/components/AsyncAceEditor/index.tsx
@@ -32,6 +32,10 @@ import AsyncEsmComponent, {
 } from 'src/components/AsyncEsmComponent';
 import useEffectEvent from 'src/hooks/useEffectEvent';
 import cssWorkerUrl from 'ace-builds/src-noconflict/worker-css';
+import { useTheme, css } from '@superset-ui/core';
+import { Global } from '@emotion/react';
+
+export { getTooltipHTML } from './Tooltip';
 
 config.setModuleUrl('ace/mode/css_worker', cssWorkerUrl);
 
@@ -135,6 +139,7 @@ export default function AsyncAceEditor(
         },
         ref,
       ) {
+        const supersetTheme = useTheme();
         const langTools = acequire('ace/ext/language_tools');
         const setCompleters = useEffectEvent(
           (keywords: AceCompleterKeyword[]) => {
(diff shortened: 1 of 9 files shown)


superset-frontend/src/components/AsyncAceEditor/index.tsx
```diff
@@ -167,15 +172,66 @@ export default function AsyncAceEditor(
...
-            theme={theme}
-            tabSize={tabSize}
-            defaultValue={defaultValue}
-            setOptions={{ fontFamily }}
-            {...props}
-          />
+          <>
+            <Global
+              styles={css`
+                .ace_tooltip {
```
michael-s-molina, line 178: @justinpark I noticed that `src/SqlLab/components/AceEditorWrapper/index.tsx` has similar logic as this code. Is it possible to avoid code duplication in this case and also ensure similar behavior for both editors by using a single component?

justinpark: As @michael-s-molina and I discussed offline, the SqlLab/AceEditorWrapper component has styles associated with the SqlLab Editor, and this section applies the global ace-editor styles.

michael-s-molina (approved):

sadpandajoe:
@supersetbot label 4.1

sadpandajoe:
@supersetbot unlabel 4.1

justinpark:
> supersetbot unlabel 4.1

@sadpandajoe Is there a reason why this was excluded from version 4.1? Could you include this commit in 4.1?

sadpandajoe:
> > supersetbot unlabel 4.1
> 
> @sadpandajoe Is there a reason why this was excluded from version 4.1? Could you include this commit in 4.1?

Yeah i can grab it

PR #24736, opened 2023-07-18
chore: make control panel sub sections look better

mistercrunch:
### SUMMARY
Improving the control panel section to make the hierarchy around the subsection look more clear. Currently it's unclear what is a section/sub-section/header. Bolding and adding an `<hr/>`  as suggested by @kasiazjc helps quite a bit

#### before
Screen Shot 2023-07-18 at 2 57 48 PM

#### after
Screen Shot 2023-07-18 at 2 54 50 PM

PR #29673, opened 2024-07-23
style: remove uppercase from labels, buttons, tabs to align with design system

mistercrunch:
Back in https://github.com/apache/superset/pull/24736, I tried to make form labels(headers) not use `text-transform: uppercase;`, but somehow it got reverted (presumably by mistake after looking at `git blame`). Here I'm pushing further by removing all css `text-transform: uppercase;` from the app, this includes buttons, tabs and labels.

# Tabs

## after
Screenshot 2024-07-23 at 2 31 57 PM

## before
Screenshot 2024-07-23 at 2 57 24 PM

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/controlPanel.tsx
@@ -279,7 +279,7 @@ const config: ControlPanelConfig = {
       label: t('Number format'),
     },
     x_axis: {
-      label: t('TEMPORAL X-AXIS'),
+      label: t('Temporal X-Axis'),
       ...temporalColumnMixin,
     },
   },
--- a/superset-frontend/src/GlobalStyles.tsx
+++ b/superset-frontend/src/GlobalStyles.tsx
@@ -40,7 +40,6 @@ export const GlobalStyles = () => (
         border-radius: ${theme.borderRadius}px;
         background: ${theme.colors.primary.base};
         border: none;
-        text-transform: uppercase;
         color: ${theme.colors.grayscale.light5};
         line-height: 1.5715;
         font-size: ${theme.typography.sizes.s}px;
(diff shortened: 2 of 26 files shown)


rusackas:
Should `Calculated columns` be `Calculated Columns`?  It's unclear from [the guidelines](https://github.com/apache/superset/wiki/Capitalization-Guidelines) if this should be considered a proper noun, or what really does qualify for title case treatment.

mistercrunch:
@kasiazjc curious on your take on clarifying "proper nouns" here. Happy to tackle as part of this PR, but I also think we may not need to figure out every single label as part of this PR

kasiazjc:
> @kasiazjc curious on your take on clarifying "proper nouns" here. Happy to tackle as part of this PR, but I also think we may not need to figure out every single label as part of this PR

I think we were looking to get rid of title case in most cases and transform them into sentence case, but I wouldn't focus on it right now - for now let's just get rid of uppercase and look for anything let's say, outrageous like "removing uppercase broke tabs here". 

I'm all in for iterative approach as it brings us closer to the better version of Superset, as in general in ux/ui design uppercase is not really being used (it comes across as shout-y and it's hard to define hierarchy with some being upper/some sentence case)

mistercrunch:
@yousoph @kasiazjc think this is mergeable as is? I did some clicking around locally and on the ephemeral envrionment there's nothing that looks outrageous. Happy to tweak how affected sections of the UI look in future PRs if necessary

kasiazjc:
> @kasiazjc curious on your take on clarifying "proper nouns" here. Happy to tackle as part of this PR, but I also think we may not need to figure out every single label as part of this PR

I think in general it looks fine - found one button which is still uppercase though. Do you maybe know why? Is it a different component? 
image

Also noting for later:
- left tab is not centered 
image
- if you have time @mistercrunch we can open another issue to change font size and weight of buttons, and also color of the primary button (as currently it's a little bit too light) 

I have to say I love how the fields look with sentence case labels in control panel etc!

mistercrunch:
@supersetbot label test

mistercrunch:
@supersetbot label test

mistercrunch:
@supersetbot unlabel test

mistercrunch:
Ok I think this is ready for review/merge. Couldn't find that `+ CHART` button easily by grepping for the code, but I'm planning on following up with more of these "look and feel" PRs

kgabryje (approved):
LGTM.
If you want to fix that uppercase "+ CHART" on homepage, the label is taken from `WelcomeTable` enum in `superset-frontend/src/features/home/types.ts`

kasiazjc (approved):

rusackas (approved):

sadpandajoe:
@supersetbot label 4.1

PR #29687, opened 2024-07-24
fix: Gamma users shouldn't be able to create roles

hughhhh:
### SUMMARY

Currently users with gamma roles are still able to create roles, This PR is made to fix this as well as add test. Reason why this is happening is due to the instantiation of the `ModelRestAPI` classes happening here https://github.com/dpgaspar/Flask-AppBuilder/blob/2d527aae09eed50fd52c6bf886a2970adb225428/flask_appbuilder/security/sqla/manager.py#L97

To fix it we'll just be adding the new security views names to the `ADMIN_ONLY_VIEW_MENUS` List in the manager class

```bash
# logged in as GAMMA
$ curl "http://127.0.0.1:9000/api/v1/security/roles/" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "name": "new_role"
}'
>> 201

$ curl "http://127.0.0.1:9000/api/v1/security/roles/" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "name": "new_role"
}'
>> 403
```

### TESTING INSTRUCTIONS

1. Configure SuperSet to use DB authentication and set `FAB_ADD_SECURITY_API = True`
2. Initialize SuperSet; be sure to add an administrator
3. Launch SuperSet
4. Browse to SuperSet
5. Log in as the administrator
6. Add a new user and grant them only the Gamma role
7. Log out
8. Log in as the new user
9. Browse to the swagger endpoint (/swagger/v1)
10. Find Security Roles
11. Expand POST /api/v1/security/roles
12. Click Try It Out
13. In the Model schema area provide a name for a new role
14. Click Execute

### ADDITIONAL INFORMATION

- [X] Has associated issue: SEC-70
- [X] Required feature flags:
- [ ] Changes UI


hughhhh:
@dpgaspar let me know if this pattern make sense or do you have another suggestion of either overriding the view registration, or adding security checks on the indy endpoints

dpgaspar (commented):
instead of overriding `register_views` can we leverage this instead https://github.com/apache/superset/blob/master/superset/security/manager.py#L226
add the FAB security views

tests/integration_tests/security/api_tests.py
```diff
@@ -135,3 +136,125 @@ def test_post_guest_token_bad_resources(self):
...
+    #     """
+    #     Security API: <TODO>
+    #     """
+    #     self.login(GAMMA_USERNAME)
+    #     response = self.client.delete(
+    #         f"{self.uri}1",
+    #         data=json.dumps({"name": "new_role"}),
+    #         content_type="application/json",
+    #     )
+    #     self.assert403(response)
```
eschutho, line 260: Yey for all the new tests! What should we do with these last two?

hughhhh: Going to finishes these up today, mainly wanted to get the adjustment @dpgaspar in and will close these up after

superset/config.py
```diff
@@ -320,6 +320,8 @@ def _try_json_readsha(filepath: str, length: int) -> str | None:
 # ex: http://localhost:8080/swagger/v1
 FAB_API_SWAGGER_UI = True
 
+FAB_ADD_SECURITY_API = True
```
hughhhh, line 323: @dpgaspar did we make and update in flask app builder to make this default True? When I recently pulled these were getting `KeyError: FAB_ADD_SECURITY_API not found` Also I saw that flask app builder has this default True so to allow these test to run I just made this value consistent with what I saw upstream.

Can you verify that this is what's happening?

dpgaspar: Default is false on FAB: https://github.com/dpgaspar/Flask-AppBuilder/blob/2d527aae09eed50fd52c6bf886a2970adb225428/flask_appbuilder/security/sqla/manager.py#L97

The link your mentioning is an example app

tests/integration_tests/security/api_tests.py
```diff
@@ -135,3 +136,64 @@ def test_post_guest_token_bad_resources(self):
...
+        Security API: Admin should be able to create roles
+        """
+        self.login(ADMIN_USERNAME)
+        response = self.client.get(self.uri)
+        self.assert200(response)
+
+    @with_config({"FAB_ADD_SECURITY_API": True})
+    def test_get_security_roles_gamma(self):
+        """
+        Security API: Admin should be able to create roles
```
dpgaspar, line 156: nit: comment is not correct

tests/integration_tests/security/api_tests.py
```diff
@@ -135,3 +136,64 @@ def test_post_guest_token_bad_resources(self):
...
         self.assert400(response)
+
+
+class TestSecurityRolesApi(SupersetTestCase):
+    uri = "api/v1/security/roles/"  # noqa: F541
+
+    @with_config({"FAB_ADD_SECURITY_API": True})
+    def test_get_security_roles_admin(self):
+        """
+        Security API: Admin should be able to create roles
```
dpgaspar, line 147: nit: `Security API: Admin should be able to get roles`

dpgaspar (approved):
Looks good, just a couple of non blocking comments

PR #29688, opened 2024-07-24
fix(ci): release process for labeling PRs

mistercrunch:
It appears the CI script that runs on-release to label PRs isn't working
as expected. Here I'm taking the `supersetbot` command out of the matrix
and running it in parallel with the docker image publication.

Tested with:
```
gh workflow run tag-release.yml --ref fix_release_tags --field release="4.0.2" --field git-ref="4.0.2" --field force-latest="true"
```

Test run:
```
https://github.com/apache/superset/actions/runs/10118610898
```

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -54,7 +54,6 @@ jobs:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v4
         with:
-          persist-credentials: false
           fetch-depth: 0
 
       - name: Use Node.js 20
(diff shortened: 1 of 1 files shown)


sadpandajoe (approved):

.github/workflows/tag-release.yml
```diff
@@ -54,10 +54,13 @@ jobs:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v4
         with:
-          persist-credentials: false
-          tags: true
           fetch-depth: 0
 
+      - name: Use Node.js 20
```
rusackas, line 59: Oooh, fancy. Time to upgrade the repo to 18... maybe it'll all be on the same version one day :D

mistercrunch: yeah `supersetbot` requires 20 as I'm using some of the newer ecmascript features and didn't want to set up babel. If I remember well I think that was just the from `import ... from ...` synthax that force me up to 20. No way I was going to deal with `require()`

rusackas (approved):

sadpandajoe:
@supersetbot label 4.1

sadpandajoe:
@supersetbot unlabel 4.1

PR #29721, opened 2024-07-26
fix: pass slack recipients correctly

eschutho:
### SUMMARY
There was a bug where the slack recipients weren't getting updated correctly when they changed from v1 name to v2 ids during execution. This also adds slack channels to header logs so that if there is an error of channel_not_found for example, we can check the logs to see which channels were passed in. And lastly I found that the integration tests weren't loading user data, so I fixed that as well. 

### TESTING INSTRUCTIONS

Run a test with slack and check that logs include this new information.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/scripts/tests/run.sh
+++ b/scripts/tests/run.sh
@@ -53,6 +53,9 @@ function test_init() {
   echo Superset init
   echo --------------------
   superset init
+  echo Load test users
+  echo --------------------
+  superset load-test-users
 }
 
 #
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -230,6 +230,7 @@ module = "tests.*"
 check_untyped_defs = false
 disallow_untyped_calls = false
 disallow_untyped_defs = false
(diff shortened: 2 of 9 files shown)


sadpandajoe (approved):

sadpandajoe (approved):

sadpandajoe:
@supersetbot label 4.1

scripts/tests/run.sh
```diff
@@ -53,6 +53,9 @@ function test_init() {
   echo Superset init
   echo --------------------
   superset init
+  echo Load test users
+  echo --------------------
+  superset load-test-users
```
eschutho, line 58: small fix after an earlier refactor where the test script stopped loading test users.

sadpandajoe:
Nice! I see some new tests in there too.

betodealmeida (approved):

PR #28789, opened 2024-05-31
feat: Adds the Featured Charts dashboard

michael-s-molina:
### SUMMARY
Creates a new example dashboard called Featured Charts with the objective of demonstrating our most active supported charts. Another objective is to provide a dashboard where we can test our plugin features using these chart types.

Screenshot 2024-05-31 at 09 27 47
Screenshot 2024-05-31 at 09 28 10
Screenshot 2024-05-31 at 09 28 32
Screenshot 2024-05-31 at 09 28 57

<br>This PR also changes the viz gallery to include missing featured charts and sort the charts alphabetically to help users find the charts they are looking for. Another small tweak was to add some margin to the Create New Chart button.

Screenshot 2024-05-31 at 10 36 49

### TESTING INSTRUCTIONS
Check the Featured Charts dashboard and the viz gallery changes.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29766, opened 2024-07-29
fix: Revert "feat: Adds the Featured Charts dashboard (#28789)"

sadpandajoe:
This reverts commit 95706d9be2b5414ed496ad762ba1996041429e01.

### SUMMARY
Noticed that ephemerals don't have any example data. This reverts the PR that could have caused this in case if we don't find a fix forward before we finish with 4.1

### TESTING INSTRUCTIONS

In the PR, bring up an ephemeral and verify that examples now show up.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Taking this out of draft state to see if it helps CI pass. If it does (or at least if it gets far enough) I'll spin up an ephemeral to make sure it works as expected :)

rusackas:
Hmm... the epehemeral still doesn't have examples. Maybe this PR isn't the root cause?

sadpandajoe:
> Hmm... the epehemeral still doesn't have examples. Maybe this PR isn't the root cause?

hmm weird. @eschutho were there any other changes from that and the 4.1 branch?

eschutho:
> hmm weird. @eschutho were there any other changes from that and the 4.1 branch?
Oh strange. That was the one that I reverted from 4.1, but let me see what else might have changed.

PR #29768, opened 2024-07-29
docs(contributing): fix broken link to translations sub-section

sfirke:
### SUMMARY
Fix broken link to sub-section header.  You can see the header is "Contributing Translations" so the link needs to reflect that:

image

### DIFF
--- a/docs/docs/contributing/contributing.mdx
+++ b/docs/docs/contributing/contributing.mdx
@@ -86,7 +86,7 @@ text strings from Superset's UI. You can jump into the existing
 language dictionaries at
 `superset/translations/<language_code>/LC_MESSAGES/messages.po`, or
 even create a dictionary for a new language altogether.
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):

mistercrunch:
@rusackas tried to install a markdown link-checker before, not sure why it was held up (?)

rusackas:
> @rusackas tried to install a markdown link-checker before, not sure why it was held up (?)

...because it works, and there are a slew of links that need to be cleaned up for it to pass its own CI :P 

https://github.com/apache/superset/pull/28424

rusackas:
Thanks for this, and thanks for merging. FWIW, I'm about to merge #28424, but that doesn't seem to catch links like this that are relative, within docusaurus. I'll keep working on that problem. It did find a ton of other broken stuff though, so it's still worth keeping/merging that action, I reckon!

PR #29783, opened 2024-07-30
feat(sqllab): Highlight and make actionable table name in the editor

justinpark:
### SUMMARY
Following https://github.com/apache/superset/discussions/26096

To make the editor smart, this commit highlights actionable items (such as table names) among the tokens hovered over by the mouse. When cmd + click is pressed, additional data linked to the relevant metadata is displayed in a popup.

The first item implemented in this commit is table names. If the SQL content includes a table name with a schema (e.g., main.bart_lines) or if the text matches a table name from the list of tables in the currently selected schema, it is marked as actionable. When clicked, it displays table metadata (including columns) and a preview query in a popup, similar to the left panel.

https://github.com/user-attachments/assets/26bd4cab-bda8-4dce-93db-688d96d498e3

### TESTING INSTRUCTIONS
type a sql including a table name and then mouse over the table name string

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/style/index.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/style/index.tsx
@@ -155,8 +155,10 @@ const defaultTheme = {
     },
   },
   zIndex: {
+    aboveEditorActiveLine: 3,
     aboveDashboardCharts: 10,
     dropdown: 11,
+    belowModal: 900,
     max: 3000,
   },
   transitionTiming: 0.3,
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -237,7 +237,7 @@ export function clearInactiveQueries(interval) {
   return { type: CLEAR_INACTIVE_QUERIES, interval };
 }
 
-export function startQuery(query) {
+export function startQuery(query, runPreviewOnly) {
   Object.assign(query, {
     id: query.id ? query.id : nanoid(11),
     progress: 0,
@@ -245,7 +245,7 @@ export function startQuery(query) {
     state: query.runAsync ? 'pending' : 'running',
     cached: false,
   });
-  return { type: START_QUERY, query };
+  return { type: START_QUERY, query, runPreviewOnly };
 }
 
 export function querySuccess(query, results) {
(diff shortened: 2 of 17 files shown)


kasiazjc:
@justinpark that looks nice! 

I think I have only one comment about preview tab which is - can copy + search (instead of filter) match the pattern in samples in explore? 
image
image

kasiazjc:
> > can copy + search (instead of filter) match the pattern in samples in explore?
> 
> @kasiazjc This preview result is output in the same way by reusing the existing preview result component. Screenshot 2024-07-31 at 11 20 20 AM Unfortunately, it is not compatible with the data format of the explore sample table, so it cannot be reused. Instead, how about modifying the existing SQL Lab preview result control to have a UI layout similar to the explore sample table?

honestly, that would be amazing as I think this would declutter SQL Lab a little bit! Is it something you would be able to work on?

I am also wondering about discoverability of this feature, because it can be hard to find if you dont know what you're looking for. Maybe for example in the empty empty state for SQL we could have some kind of "default doc placeholder sql comment" (sorry for the name lol), that would list shortcuts and features like the one you are implementing? What do you think?

justinpark:
> honestly, that would be amazing as I think this would declutter SQL Lab a little bit! Is it something you would be able to work on?

Sure. Since it's separate issue, I'll push a new PR for the change.

> I am also wondering about discoverability of this feature, because it can be hard to find if you dont know what you're looking for.

Of course, I have been considering the discoverability. As shown in the screenshot below, similar to existing editors, we have highlighted certain clickable keywords to provide feedback. When the cursor hovers over them, a tooltip is provided, making it easier to find the functionality.

https://github.com/user-attachments/assets/e14eb576-e201-4339-b4d9-87a059147e43

kasiazjc:
> > honestly, that would be amazing as I think this would declutter SQL Lab a little bit! Is it something you would be able to work on?
> 
> Sure. Since it's separate issue, I'll push a new PR for the change.
> 
> > I am also wondering about discoverability of this feature, because it can be hard to find if you dont know what you're looking for.
> 
> Of course, I have been considering the discoverability. As shown in the screenshot below, similar to existing editors, we have highlighted certain clickable keywords to provide feedback. When the cursor hovers over them, a tooltip is provided, making it easier to find the functionality.
> 
>  mouse-over-token-tooltip.mov

Makes sense, thank you!

rusackas:
Marking this as draft while it awaits rebase and/or other touchups.

PR #29786, opened 2024-07-31
refactor(Slider): Upgrade Slider to Antd 5

geido:
### SUMMARY
Migration of Slider component from Ant Design v4 -> v5.

### BEFORE
before_340247618-76360560-c8f6-4597-91e4-dd1774611b79

### AFTER
after_340247494-9c545bfb-8fde-49f8-aa69-e68e021fa110

### TESTING INSTRUCTIONS
1. Check the usage of the Slider
2. No visual changes should be introduced

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/Slider/Slider.stories.tsx
```diff
@@ -16,33 +16,97 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import Slider, { SliderSingleProps } from '.';
+import { AntdThemeProvider } from 'src/components/AntdThemeProvider';
```
rusackas, line 19: No longer needed... theme is available to all Stories now.

superset-frontend/src/components/Slider/index.tsx
```diff
@@ -16,13 +16,19 @@
...
 
 export type { SliderSingleProps, SliderRangeProps };
 
 export default function Slider(props: SliderSingleProps | SliderRangeProps) {
-  return <AntdSlider {...props} css={{ marginLeft: 0, marginRight: 0 }} />;
+  return (
+    <AntdSlider
+      {...props}
+      css={(theme: SupersetTheme) => css`
+        margin: ${theme.gridUnit * 2 + 2}px 0 0 ${theme.gridUnit * 2 + 2}px;
```
kgabryje, line 30: Can we only use full grid units?

superset-frontend/src/theme/index.ts
```diff
@@ -69,6 +69,12 @@ const baseConfig: ThemeConfig = {
       colorText: supersetTheme.colors.text.label,
       remainingColor: supersetTheme.colors.grayscale.light4,
     },
+    Slider: {
+      trackBgDisabled: supersetTheme.colors.grayscale.light1,
+      colorBgElevated: supersetTheme.colors.grayscale.light5,
+      handleSizeHover: supersetTheme.gridUnit * 2.5,
+      handleLineWidthHover: supersetTheme.gridUnit / 2,
```
kgabryje, line 76: I wonder if we should tie the handle size and line width do grid units. It sounds like something that's independent of grid size

superset-frontend/src/components/Slider/index.tsx
geido: This was done to mimic the previous behavior of the component. If we don't force it like this there will be a non-negligible difference in margins.

kgabryje: Yeah I know and... I guess it's fine? Unless those differences actually break something somewhere

superset-frontend/src/theme/index.ts
geido: Agreed

Antonio-RiveroMartnez (approved):
comments addressed, lgtm

superset-frontend/src/theme/index.ts
```diff
@@ -72,8 +72,8 @@ const baseConfig: ThemeConfig = {
     Slider: {
       trackBgDisabled: supersetTheme.colors.grayscale.light1,
       colorBgElevated: supersetTheme.colors.grayscale.light5,
-      handleSizeHover: supersetTheme.gridUnit * 2.5,
-      handleLineWidthHover: supersetTheme.gridUnit / 2,
+      handleSizeHover: 20,
```
kgabryje, line 75: Shouldn't that be `10`? 😄

superset-frontend/src/components/Slider/index.tsx
geido: I think, if we are good with visual differences we can omit the custom style entirely and rely on what Antdesign provides out of the box.

kgabryje (approved):
LGTM

superset-frontend/src/theme/index.ts
```diff
@@ -40,7 +40,7 @@ const baseConfig: ThemeConfig = {
     colorSuccess: supersetTheme.colors.success.base,
     colorTextBase: supersetTheme.colors.grayscale.dark2,
     colorWarning: supersetTheme.colors.warning.base,
-    controlHeight: supersetTheme.gridUnit * 32,
+    controlHeight: 32,
```
rtexelm, line 43: seems bizarre for this to have a previous value of 128!

rtexelm (commented):
LGTM

PR #29787, opened 2024-07-31
perf: Antd icons tree shaking

kgabryje:
### SUMMARY
Previously we were importing the whole @ant-design/icons library, which contained over 800 icons. We only use 40 of those.
Implementing tree-shakable import allowed to shrink the chunk size by over 5MB.

Before:

Screenshot 2024-07-31 at 15 32 27

After:

Screenshot 2024-07-31 at 15 34 58

### TESTING INSTRUCTIONS
Everything should work like before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


geido:
While I understand the need for this, the idea here was that you could use any Antd icon without having to specify it again. Not sure if there is a way around that to keep that functionality available.

michael-s-molina:
Agree with @kgabryje 👍🏼

superset-frontend/src/components/Icons/AntdEnhanced.tsx
```diff
@@ -17,10 +17,94 @@
  * under the License.
  */
 
-import * as AntdIcons from '@ant-design/icons/lib/icons';
+import {
```
mistercrunch, line 20: would be good to add a comment
```
// Targeted import (as opposed to `import * as antdIcons`) is important here for proper tree-shaking
```

superset-frontend/src/components/Icons/AntdEnhanced.tsx
```diff
@@ -17,10 +17,94 @@
...
+  SettingOutlined,
+  StopOutlined,
+  TagsOutlined,
+  UpOutlined,
+  UserOutlined,
+} from '@ant-design/icons';
 import { StyledIcon } from './Icon';
 import IconType from './IconType';
 
+const AntdIcons = {
```
mistercrunch, line 65: I looked for ways to keep things DRY here and not have to repeat icon names twice, but didn't find any way. Too bad ECMAScript doesn't support something like `import {a, b, c} as container from ....`

mistercrunch (approved):
LGTM - though would be nice to add this comment as someone reading the code might think "why did they not just `import *`!?"

rusackas (approved):
LG2M2!

superset-frontend/src/components/Icons/AntdEnhanced.tsx
mistercrunch: ok I added a commit with the comment - first time using that GH feature...

rusackas: > first time using that GH feature...

I'm in love with it... also great for fixing little linting issues on PRs and such.

kgabryje: oh nice! never used it myself

PR #29779, opened 2024-07-30
fix: remove --no-optional from docker-compose build

mistercrunch:
Remove unusual artifact from local docker-compose setup, using vanilla `npm install`

PR #29788, opened 2024-07-31
perf: Remove antd-with-locales import

kgabryje:
### SUMMARY
We were importing a huge package `antd-with-locales` (over 4MB) just to pick 1 locale for DatePicker from it.
This PR removes that import and lazy loads just the required locale.

Before:

Screenshot 2024-07-31 at 15 48 00

After:

Screenshot 2024-07-31 at 15 43 44

### TESTING INSTRUCTIONS
1. Add a time filter to a dashboard
2. Switch to "Custom frame"
3. Open a date picker
4. Should work like before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Loving these! Here are some suggestions (take 'em or leave 'em) that I came up with _totally on my own with no help_.

Potential concerns:

- There might be a brief moment where the component doesn't render while the locale is being loaded
- Error handling for failed locale imports could be improved

Suggestions for improvement:

- Consider adding a loading state instead of returning null when datePickerLocale is null
- Add error handling for the locale import promise to gracefully handle failed imports
- Consider memoizing the effect dependency array to prevent unnecessary re-runs of the effect

rusackas (approved):
Potential enhancements on the thread, but I love it. LGTM.

mistercrunch (approved):
LGTM! Thinking of ways to prevent this, but I can totally see how whoever brought this npm library having no indication of the impact on the bundle. 

Let's think about what CI automation we'd like to have in place to alert author/reviewers about bundle size changes. Let me start a discussion thread about it.

mistercrunch:
Started a GH discussion about preventing bundle size creep using CI here -> https://github.com/apache/superset/discussions/29803

kgabryje:
Had to bump `@testing-library/react` - our version had a bug which caused `waitForElementToBeRemoved` to be flaky

hainenber:
@kgabryje I think your PR's lockfile has slipped out the required `@nx` dependencies 👀 

This is in the `master` branch (6 search results with dependencies in `node_modules`)
image


This is in your PR's branch (only 2 search results without the ones in `node_modules`)
image

kgabryje:
> @kgabryje I think your PR's lockfile has slipped out the required `@nx` dependencies 👀

Oh that's interesting... I only bumped a tests library. I'll try to regenerate package-lock and see if it helps. Thanks!

kgabryje:
Looks like it's related to [this bug](https://github.com/npm/cli/issues/4828). I generated the package-lock file on my M2 macbook, and only arm64 references were generated. I fixed it by removing package-lock.json AND node_modules before rerunning `npm-install`

hainenber:
> Looks like it's related to https://github.com/npm/cli/issues/4828. I generated the package-lock file on my M2 macbook, and only arm64 references were generated. I fixed it by removing package-lock.json AND node_modules before rerunning npm-install

Thanks @kgabryje for going through and find the underlying RC! This seems precarious and I couldn't help but wonder if migrating to `yarn` could help alleviating this if not going with Lerna downgrade approach. Maybe a little too "overkill" though. Wdyt?

kgabryje:
> Edit: according to this [comment](https://github.com/dprint/dprint/issues/896#issuecomment-2254530242) in the linked issue, upgrading to currently latest `npm` version, v10.8.2, could help with the mitigation 🤔


I did all that with the latest npm version 🙁 A quick (and a bit dirty) fix might be to `nx-linux-x64-gnu` as Superset dependency - I noticed that @eschutho had a similar problem in another PR some time ago, so I'm not the only one

eschutho:
> I did all that with the latest npm version 🙁 A quick (and a bit dirty) fix might be to `nx-linux-x64-gnu` as Superset dependency - I noticed that @eschutho had a similar problem in another PR some time ago, so I'm not the only one

That should be fixed in @mistercrunch's [PR](https://github.com/apache/superset/pull/29779), but give us a shout if it didn't.

PR #17411, opened 2021-11-11
feat: Upgrade documentation V2

hughhhh:
### SUMMARY

Upgrading documentation to use docusaurus instead of gatsbyJS. This will allow us to have better control on the docs provide with different versions of superset as we continue to develop.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Screen Shot 2022-01-05 at 2 30 58 PM

### Dark mode shot with darkmode logo
Screen Shot 2022-01-05 at 2 31 54 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29790, opened 2024-07-31
chore(docs): remove mention of MariaDB in dev environment setup

sfirke:
MariaDB is not a supported backend for Superset, so it is misleading to suggest it for a development environment.  The commit that named MariaDB here is quite old -- I traced it back to [a docs overall from 3 years ago](https://github.com/apache/superset/pull/17411), it is likely older than that (the PR was upgrading the docs system) but I stopped there.

There are two other things that could be addressed here:
- Does, in fact, a DB need to be installed at this step?  I didn't see it referenced immediately.  If no, remove this whole sentence.  If yes, provide an example command.
- Would Postgres be an acceptable alternative?  If so, say that.

If someone knows the answer please feel free to suggest the complete fix.  But if no one knows, removing the MariaDB mention is an adequate improvement for now.

### DIFF
--- a/.github/workflows/superset-docs-verify.yml
+++ b/.github/workflows/superset-docs-verify.yml
@@ -13,6 +13,7 @@ concurrency:
 
 jobs:
   linkinator:
+    # See docs here: https://github.com/marketplace/actions/linkinator
     name: Link Checking
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


rusackas (approved):

rusackas:
Oops... my little documentation link checker is finding issues. I'll push a commit here to (hopefully) resolve that.

Issue #29797, opened 2024-07-31
Reporting / Alert scheduling doesn't properly convert timezones


danielli-ziprecruiter:
### Bug description

Operating System: Ubuntu 22.04
Version: 4.0.2

Create an alert that will trigger.

### How to reproduce the bug

1. Create an alert that will trigger.
2. Schedule the alert to run soon with a timezone other than UTC.
3. Verify that alert doesn't run when it should.

### Superset version

4.0.2

### Python version

3.10

### Node version

16

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
Fixed by **https://github.com/apache/superset/pull/29798**

PR #29798, opened 2024-07-31
fix: set default timezone to UTC for cron timezone conversions

danielliu-eon:
### SUMMARY

This fix addresses the issue mentioned in #29797. The datetime object passed into the cron scheduling function may not have a timezone associated with it. Therefore, when trying to convert the datetime to the specified timezone, it just adds timezone to the datetime. This is an issue because the function attempts to convert the datetime back into UTC, which causes an incorrect timezone conversion in this case. The fix is to add UTC to the timezone info for the datetime object before converting it to the specified timezone.

### TESTING INSTRUCTIONS

Attempt to reproduce the bug as described in #29797. The alert should trigger correctly.

### ADDITIONAL INFORMATION

- [ x] Has associated issue: #29797
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/tasks/scheduler.py
+++ b/superset/tasks/scheduler.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 import logging
-from datetime import datetime
+from datetime import datetime, timezone
 
 from celery import Celery
 from celery.exceptions import SoftTimeLimitExceeded
@@ -50,7 +50,7 @@ def scheduler() -> None:
         datetime.fromisoformat(scheduler.request.expires)
         - app.config["CELERY_BEAT_SCHEDULER_EXPIRES"]
         if scheduler.request.expires
-        else datetime.utcnow()
+        else datetime.now(tz=timezone.utc)
     )
     for active_schedule in active_schedules:
         for schedule in cron_schedule_window(


mistercrunch:
I haven't dug in the code, but will this cause a change in behavior (meaning the actual time at which certain alerts will trigger) in some environments for some alerts? If that's the case we should either prevent that with some flag and/or adding a comment in `UPDATING.md` to let administrator know what might be affected and how to tweak things to preserve the current timing/behavior.

eschutho:
Thanks @danielli-ziprecruiter for the PR! In addition to @mistercrunch's comments, can you add some tests to this PR?

danielli-ziprecruiter:
Yes, this would cause a change in behavior, but the previous behavior didn't make sense since the alert / reporting run would not occur at the configured time and timezone and instead occur at time offset by the configured timezone's relation to UTC.

mistercrunch:
> he alert / reporting run would not occur at the configured time and timezone and instead occur at time offset by the configured timezone's 

Makes sense, though wondering its possible that people might have set their alerts/reports while relying on this quirky behavior. 

Can we clarify the scope of alerts that would be affected (say "only the ones that set a TZ")? Is this an edge case or does it affect most alerts out there? We have to think of the thousands of organizations running hundreds of thousands or alerts/reports every day...

If it's a small portion of alerts/reports, maybe a note in UPDATING.md is sufficiant.

If it's a a large portion of alerts/reports, we have different options
- a db migration that makes sure that effective run time of alerts are not affected, altering the existing alerts
- wait for the next major release to ship as a breaking change
- some sort of flag for each alert (new_scheduling_logic = True/False) where we apply the old/new logic conditionally

danielli-ziprecruiter:
I think I've found where this issue was actually introduced. This [commit](https://github.com/apache/superset/commit/f672d5da5cb9390b83176bb12c27ce7eeea3e8ae) changes `datetime.now(tz=dt_timezone.utc)` to `datetime.utcnow()`. `datetime.utcnow()` is not timezone aware, hence the issue described earlier. I've updated the PR accordingly. Would I still need to add additional tests or comments to `UPDATING.md`?

mistercrunch:
I think a message in https://github.com/apache/superset/blob/master/UPDATING.md seems important to let people know about any "change in behavior" for when they are upgrading. In theory whoever conducts upgrades should read through that file and ack the notices, notify their users, or take action if needed. Ideally the message is clear about the scope of the change of behavior (which types of alerts/reports will be affected, maybe it's only the alerts/reports that specify a TZ, idk) and what's the old/new behavior.

michael-s-molina (approved):

mistercrunch:
@michael-s-molina @rusackas did we not think a comment in UPDATING.md was helpful/necessary?

michael-s-molina:
I'm sorry @mistercrunch, I think I missed your comment. I agree that a comment in `UPDATING.md` is necessary. @danielli-ziprecruiter could you open a follow-up please?

mistercrunch:
I just created a new `hold:updating.md` label we can use that should prevent merging without having to do a "requires changes" review

nicholaslimck:
@mistercrunch should the hold:updating.md label be removed since @danielli-ziprecruiter added the comment to UPDATING.md in #29863? Hoping that this fix can make it into the next release of Superset.

PR #29802, opened 2024-07-31
fix: create permissions on DB import

betodealmeida:
### SUMMARY

A few bugs related to catalogs and permissions:

1. When a database is imported, catalog and schema permissions are not created. Ideally we should have the import functionality call the corresponding command (`CreateDatabaseCommand` in this case) to prevent duplicate logic (the import/export flow was created before the introduction of commands). For now this PR just adds the logic manually.
2. When a database is updated, datasets and charts are not having their `catalog_perms` updated.
3. Databricks has a weird behavior where `get_default_catalog() ⊈ get_catalog_names()`, so I fixed it.

N/A

### TESTING INSTRUCTIONS

Added and updated tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/databricks.py
+++ b/superset/db_engine_specs/databricks.py
@@ -434,7 +434,25 @@ class DatabricksNativeEngineSpec(DatabricksDynamicBaseEngineSpec):
         cls,
         database: Database,
     ) -> str | None:
+        """
+        Return the default catalog.
+
+        The default behavior for Databricks is confusing. When Unity Catalog is not
+        enabled we have (the DB engine spec hasn't been tested with it enabled):
+
+            > SHOW CATALOGS;
+            spark_catalog
+            > SELECT current_catalog();
+            hive_metastore
+
+        To handle permissions correctly we use the result of `SHOW CATALOGS` when a
+        single catalog is returned.
+        """
         with database.get_sqla_engine() as engine:
+            catalogs = {catalog for (catalog,) in engine.execute("SHOW CATALOGS")}
+            if len(catalogs) == 1:
+                return catalogs.pop()
+
             return engine.execute("SELECT current_catalog()").scalar()
 
     @classmethod
--- a/superset/commands/database/create.py
+++ b/superset/commands/database/create.py
@@ -41,6 +41,7 @@ from superset.commands.database.ssh_tunnel.exceptions import (
 from superset.commands.database.test_connection import TestConnectionDatabaseCommand
 from superset.daos.database import DatabaseDAO
 from superset.databases.ssh_tunnel.models import SSHTunnel
+from superset.db_engine_specs.base import GenericDBException
 from superset.exceptions import SupersetErrorsException
 from superset.extensions import event_logger, security_manager
 from superset.models.core import Database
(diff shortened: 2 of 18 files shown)


pyproject.toml
```diff
@@ -236,7 +236,7 @@ legacy_tox_ini = """
 # Remember to start celery workers to run celery tests, e.g.
 # celery --app=superset.tasks.celery_app:app worker -Ofair -c 2
 [testenv]
-basepython = python3.10
+basepython = python3.11
```
Vitor-Avila, line 239: Just to double check: if this version change were to affect any other tests or logic, it should fail the GH Actions for this PR as well, right?

superset/db_engine_specs/databricks.py
```diff
@@ -434,7 +434,25 @@ def get_default_catalog(
...
+            > SELECT current_catalog();
+            hive_metastore
+
+        To handle permissions correctly we use the result of `SHOW CATALOGS` when a
+        single catalog is returned.
+        """
         with database.get_sqla_engine() as engine:
+            catalogs = {catalog for (catalog,) in engine.execute("SHOW CATALOGS")}
+            if len(catalogs) == 1:
+                return catalogs.pop()
```
Vitor-Avila, line 454: if `SHOW CATALOGS` returns only `spark_catalog`, it seems we're returning it. Do we want to return `spark_catalog`, or do we need to return `hive_metastore` instead (when there's only `spark_catalog`)? I'm asking it because I think there are issues with other metadata calls if we try to use `spark_catalog` when the catalog selected in the form is `hive_metastore` (or if UC is not enabled).

Also, if `multi_catalog` is disabled, do we want to rely on `SHOW CATALOGS`? Should we validate if `multi_catalog` is disabled first, and if so return the value of `current_catalog()` directly (which should be the catalog from the form)?

pyproject.toml
betodealmeida: Ah, I need to revert this, it was for testing locally. Sorry.

superset/db_engine_specs/databricks.py
betodealmeida: The problem is that we need to create the permissions regardless of the state of `multi_catalog`. For Databricks, the migration process will create permissions for:

1. Whatever `get_default_catalog()` returns.
2. All the catalogs from `get_catalog_names` different from the default catalog from (1).

For Databricks, this means we're creating 2 permissions when Unity Catalog is not enabled:

1. `[Databricks].[hive_metastore]`
2. `[Databricks].[spark_catalog]`

This in itself is already a problem for the admin — which would should they assign roles to? Either one is problematic, depending on the status of `multi_catalog`:

With `multi_catalog` disabled the catalog sent by API calls is `null`, so it gets replaced with the default catalog, `hive_metastore`. If the admin assign roles to `[Databricks].[spark_catalog]` people won't be able to access datasets, because of the name mismatch.

With `multi_catalog` enabled the only option in the dropdown is `spark_catalog`. Which means if the admin assigned roles to `[Databricks].[hive_metastore]` people won't be able to access datasets.

With the change in this PR, everything is consistent, because the default catalog is part of the list of catalogs.

sadpandajoe:
@supersetbot label 4.1

pyproject.toml
Vitor-Avila: Hey @betodealmeida I believe this should be good to be approved as soon as you revert this change 🙌

Vitor-Avila (approved):

PR #29832, opened 2024-08-01
fix: slack update for v2

eschutho:
### SUMMARY

There was a bug where the slack recipients weren't getting updated correctly when they changed from v1 name to v2 ids during execution. This also adds slack channels to header logs so that if there is an error of channel_not_found for example, we can check the logs to see which channels were passed in. And lastly I found that the integration tests weren't loading user data, so I fixed that as well.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -228,6 +228,7 @@ ignore_errors = true
 [[tool.mypy.overrides]]
 module = "tests.*"
 check_untyped_defs = false
+disable_error_code = annotation-unchecked
 disallow_untyped_calls = false
 disallow_untyped_defs = false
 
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -227,7 +227,7 @@ ignore_errors = true
 
 [[tool.mypy.overrides]]
 module = "tests.*"
-check_untyped_defs = false
+disable_error_code = "annotation-unchecked"
```
eschutho, line 230: it looks like this is the way to suppress the untyped function definitions.

PR #29840, opened 2024-08-02
fix: handle empty catalog when DB supports them

betodealmeida:
### SUMMARY

Handle edge cases when catalog is passed as `null` but the database supports catalog. This happens when "Allow changing catalogs" is disabled in the database. In these cases we need to use the default catalog, so that permissions checks work as expected.

N/A

### TESTING INSTRUCTIONS

1. Add a database with catalogs (eg, Postgres)
  - [X] Catalog permissions are created (eg, `[Postgres].[superset]`)
  - [X] Schema permissions are created (eg, `[Postgres].[superset].[public]`)

2. Create a user A with catalog-level permissions to the default catalog
  - [X] User can see schemas in SQL Lab
  - [X] User can choose schema in SQL Lab
  - [X] User can choose table in SQL Lab
  - [X] User can run queries
  - [X] User can create a new dataset
  - [X] User can see charts built from the dataset

3. Create a user B with schema-level permissions to the default catalog
  - [X] User can see only 1 schema in SQL Lab
  - [X] User can choose table in SQL Lab
  - [X] User can run queries
  - [X] User can create a new dataset
  - [X] User can see charts built from the dataset

4. Create a user C with catalog-level permissions to a non-default catalog
  - [X] User can't see schemas/tables in SQL Lab
  - [X] User can't run queries in SQL Lab

5. Create a user D with schema-level permissions to a non-default catalog
  - [X] User can't see schemas/tables in SQL Lab
  - [X] User can't run queries in SQL Lab

6. Enable multi-catalog
  - [X] Admin should see all catalogs
  - [X] A, B, C, D should only see the catalogs they have access to
  - [X] Repeat steps from (2)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/scripts/change_detector.py
+++ b/scripts/change_detector.py
@@ -52,7 +52,7 @@ GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
 def fetch_files_github_api(url: str):  # type: ignore
     """Fetches data using GitHub API."""
     req = Request(url)
-    req.add_header("Authorization", f"token {GITHUB_TOKEN}")
+    req.add_header("Authorization", f"Bearer {GITHUB_TOKEN}")
     req.add_header("Accept", "application/vnd.github.v3+json")
 
     print(f"Fetching from {url}")
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -461,9 +461,11 @@ class BaseDatasource(AuditMixinNullable, ImportExportMixin):  # pylint: disable=
                 )
             else:
                 _columns = [
-                    utils.get_column_name(column_)
-                    if utils.is_adhoc_column(column_)
-                    else column_
+                    (
+                        utils.get_column_name(column_)
+                        if utils.is_adhoc_column(column_)
+                        else column_
+                    )
                     for column_param in COLUMN_FORM_DATA_PARAMS
                     for column_ in utils.as_list(form_data.get(column_param) or [])
                 ]
@@ -1963,7 +1965,7 @@ class SqlaTable(
         if self.has_extra_cache_key_calls(query_obj):
             sqla_query = self.get_sqla_query(**query_obj)
             extra_cache_keys += sqla_query.extra_cache_keys
-        return extra_cache_keys
+        return list(set(extra_cache_keys))
 
     @property
     def quote_identifier(self) -> Callable[[str], str]:
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -1159,7 +1159,7 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
         self.incr_stats("init", self.select_star.__name__)
         try:
             result = database.select_star(
-                Table(table_name, schema_name),
+                Table(table_name, schema_name, database.get_default_catalog()),
                 latest_partition=True,
             )
         except NoSuchTableError:
--- a/tests/integration_tests/databases/api_tests.py
+++ b/tests/integration_tests/databases/api_tests.py
@@ -1563,34 +1563,6 @@ class TestDatabaseApi(SupersetTestCase):
         rv = self.client.get(uri)
         self.assertEqual(rv.status_code, 404)
 
-    def test_get_select_star_datasource_access(self):
-        """
-        Database API: Test get select star with datasource access
-        """
-        table = SqlaTable(
-            schema="main", table_name="ab_permission", database=get_main_database()
-        )
-        db.session.add(table)
-        db.session.commit()
-
-        tmp_table_perm = security_manager.find_permission_view_menu(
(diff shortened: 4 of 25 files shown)


rusackas:
I just put this in draft mode whilst it doesn't have a title/description :) Best of luck with whatever it is! :D

scripts/change_detector.py
```diff
@@ -52,7 +52,7 @@
 def fetch_files_github_api(url: str):  # type: ignore
     """Fetches data using GitHub API."""
     req = Request(url)
-    req.add_header("Authorization", f"token {GITHUB_TOKEN}")
+    req.add_header("Authorization", f"Bearer {GITHUB_TOKEN}")
```
betodealmeida, line 55: > Note: In most cases, you can use Authorization: Bearer or Authorization: token to pass a token. However, if you are passing a JSON web token (JWT), you must use Authorization: Bearer. [[source](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28#about-authentication)]

sadpandajoe:
@supersetbot label 4.1

scripts/change_detector.py
Vitor-Avila: Nice!

Vitor-Avila (approved):
LGTM!

tests/integration_tests/databases/api_tests.py
```diff
@@ -1563,34 +1563,6 @@ def test_get_select_star_not_allowed(self):
         rv = self.client.get(uri)
         self.assertEqual(rv.status_code, 404)
 
-    def test_get_select_star_datasource_access(self):
```
Vitor-Avila, line 1566: Non-blocking comment: would be good if you could mention why these tests were removed.

superset/databases/api.py
```diff
@@ -1159,7 +1159,7 @@ def select_star(
         self.incr_stats("init", self.select_star.__name__)
         try:
             result = database.select_star(
-                Table(table_name, schema_name),
+                Table(table_name, schema_name, database.get_default_catalog()),
```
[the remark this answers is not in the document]
betodealmeida: Yeah, but I I couldn't find any API calls to this endpoint — it seems like we return this in the database request now, instead of calling `/select_star/`, which is why I didn't change the code to pass a parameter. I think we should keep this endpoint until the next major version and then get rid of it.

tests/integration_tests/databases/api_tests.py
betodealmeida: These tests were poorly written and relied on weird side effects.

superset/databases/api.py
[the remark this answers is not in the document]
Vitor-Avila: gotcha! thank you

tests/integration_tests/databases/api_tests.py
Vitor-Avila: thank you!

superset/connectors/sqla/models.py
```diff
@@ -1963,7 +1965,7 @@ class and any keys added via `ExtraCache`.
         if self.has_extra_cache_key_calls(query_obj):
             sqla_query = self.get_sqla_query(**query_obj)
             extra_cache_keys += sqla_query.extra_cache_keys
-        return extra_cache_keys
+        return list(set(extra_cache_keys))
```
ERGO1995, line 1968: Why we replace this line? Actually not doing great job in our case related to https://github.com/apache/superset/issues/31975

betodealmeida: This is just a way of deduping the keys, what problems are you having with it?

Airliquide76: in our case it break the chart key retreiving ! 
we've got some logs saying cache retreiving failed ! 
Query works on sql-lab but crash in chart or dashboard ( works if force refresh as it does not go to cache ) 
We are using a lot of url_params reading at the dataset execution.

betodealmeida: But how does it break? Is there a stacktrace?i

All this is doing is deduping a list. If it's causing errors, the reason must be upstream of this code.

Airliquide76: Main error was a 422 
with "Error loading data from cache"uwsgi_superset.log:2025-01-23 
And a warning like 15:10:33,731:WARNING:superset.common.utils.query_cache_manager:force_cached (QueryContext): value not found for key 8bda314040760dc2b33b861fb943fc52
Flipping this line make verything works. 
Maybe the key for cache creation is not deduplicate and we may have some duplicate on the key witch fail retriving

PR #29846, opened 2024-08-03
fix: load slack channels earlier

eschutho:
### SUMMARY

Loading slack channels earlier to ensure that the method dropdowns have the correct value when they are loaded. Instead of loading slack channels when the slack option is selected, I am now loading the options when the notification method is rendered. That way, we know if we need to use slack v1 or v2 by the time the notification method options are rendered (since we filter out one of the options, so that only one slack option displays)

no visual changes

### TESTING INSTRUCTIONS

If the feature flag is on, but the scopes are missing, a v1 (textarea input field) should display for slack. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/components/NotificationMethod.tsx
+++ b/superset-frontend/src/features/alerts/components/NotificationMethod.tsx
@@ -211,6 +211,8 @@ export const NotificationMethod: FunctionComponent<NotificationMethodProps> = ({
   const [ccValue, setCcValue] = useState<string>(cc || '');
   const [bccValue, setBccValue] = useState<string>(bcc || '');
   const theme = useTheme();
+  const [methodOptionsLoading, setMethodOptionsLoading] =
+    useState<boolean>(true);
   const [slackOptions, setSlackOptions] = useState<SlackOptionsType>([
     {
       label: '',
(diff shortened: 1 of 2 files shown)


sadpandajoe:
@supersetbot label 4.1

superset-frontend/src/features/alerts/components/NotificationMethod.tsx
```diff
@@ -257,51 +259,47 @@ export const NotificationMethod: FunctionComponent<NotificationMethodProps> = ({
...
-                ],
-              }),
-            );
-            if (method === NotificationMethodOption.Slack) {
-              onMethodChange({
-                label: NotificationMethodOption.Slack,
-                value: NotificationMethodOption.SlackV2,
-              });
+            if (
+              method &&
```
betodealmeida, line 275: This line shouldn't be necessary, right?

betodealmeida (approved):

superset-frontend/src/features/alerts/components/NotificationMethod.tsx
eschutho: It's to help typescript assert that the value exists, rather than runtime execution. e.g., for `NotificationMethodOption` as it doesn't include `undefined`. I could coerce the type, but I think that sometimes looks like a shortcut, too. wdyt?

eschutho: I'm going to merge, but if you suggest another way to approach this, I'll put it up in a new PR.

PR #29864, opened 2024-08-05
fix: mypy issue on py3.9 + prevent similar issues

mistercrunch:
Someone reported that they had this mypy-related issue on py3.9, so I decided to fix it and to run pre-commit checks against more python versions to prevent this kind of issue from slipping into the repo

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -16,6 +16,9 @@ concurrency:
 jobs:
   pre-commit:
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python-version: ["current", "next", "previous"]
     steps:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


rusackas:
Not sure why the pre-commit job canceled here.

.github/workflows/pre-commit.yml
```diff
@@ -16,6 +16,9 @@ concurrency:
 jobs:
   pre-commit:
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python-version: ["current", "next", "previous"]
```
rusackas, line 21: Should we make this `python-version: ["3.9", "3.10", "3.11"]` to be a little more explicit in what we support?

mistercrunch: It's like this for very good reason - has to do with `required_status_checks` and matrix-management. There's all sorts of issues around renaming checks/matrix members and required checks. When we'll bump python version we'll be able to do it centrally in `.github/actions/setup-backend/action.yml` (current/next/future is defined by us in there), and we won't have issues around having to rename check names in master prior to being able to merge the branch

mistercrunch: This recent PR shows the kind of trouble we have to go through in order to rename checks... https://github.com/apache/superset/pull/29866

superset/commands/distributed_lock/base.py
```diff
@@ -17,7 +17,7 @@
 
 import logging
 import uuid
-from typing import Any
+from typing import Any, Union
```
mistercrunch, line 20: python 3.9 doesn't support the pipe operator in type hints...

rusackas (approved):

PR #29868, opened 2024-08-05
chore: remove useless GitHub action

mistercrunch:
I noticed this GHA is a no-op, with one check doing effectively nothing, and the other being a duplicate.


.github/workflows/superset-python-misc.yml
```diff
@@ -1,53 +0,0 @@
...
-          persist-credentials: false
-          submodules: recursive
-      - name: Check for file changes
-        id: check
-        uses: ./.github/actions/change-detector/
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Setup Python
-        uses: ./.github/actions/setup-backend/
-        if: steps.check.outputs.python
```
mistercrunch, line 33: `python-lint` doesn't actually lint, it does nothing. All the linting is in pre-commit now, see this section here -> https://github.com/apache/superset/blob/master/.pre-commit-config.yaml#L74-L93

.github/workflows/superset-python-misc.yml
```diff
@@ -1,53 +0,0 @@
...
-        id: check
-        uses: ./.github/actions/change-detector/
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Setup Python
-        if: steps.check.outputs.python
-        uses: ./.github/actions/setup-backend/
-      - name: Test babel extraction
-        if: steps.check.outputs.python
-        run: scripts/translations/babel_update.sh
```
mistercrunch, line 53: same as this https://github.com/apache/superset/blob/master/.github/workflows/superset-translations.yml#L65-L67

eschutho (approved):

PR #29875, opened 2024-08-06
feat(build): webpack visualizer

rusackas:
### SUMMARY

Some updates to the bundle analyzer in honor of @kgabryje's recent efforts. This PR does several things:

- Adds a new `npm run bundle-stats` script to run package bundle analysis
- Makes webpack's production build process (a.k.a. `npm run build`) exits when complete.
- Removes the earlier `webpack-bundle-analyzer` flag in favor of the new npm script
- Sets `webpack-bundle-analyzer` to `static` mode rather than `server` mode.
- Adds a new bundle analyzer with a sunburst chart, which makes some aspects easier
- Has BOTH bundle analyzers open at the end of the build process in your browser.

Here's what the new bundle analyzer looks like:
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


kgabryje:
Nice! I wonder if we shoukd run the plugin only if a flag is passed, like we do with bundle analyzer

rusackas:
@kgabryje check out the updated PR description, it all works pretty smoothly now.

superset-frontend/package.json
```diff
@@ -38,11 +38,12 @@
   ],
   "scripts": {
     "_prettier": "prettier './({src,spec,cypress-base,plugins,packages,.storybook}/**/*{.js,.jsx,.ts,.tsx,.css,.less,.scss,.sass}|package.json)'",
-    "build": "cross-env NODE_OPTIONS=--max_old_space_size=8192 NODE_ENV=production BABEL_ENV=\"${BABEL_ENV:=production}\" webpack --mode=production --color",
+    "build": "cross-env NODE_OPTIONS=--max_old_space_size=8192 NODE_ENV=production BABEL_ENV=\"${BABEL_ENV:=production}\" webpack --color --mode production",
     "build-dev": "cross-env NODE_OPTIONS=--max_old_space_size=8192 NODE_ENV=development webpack --mode=development --color",
     "build-instrumented": "cross-env NODE_ENV=production BABEL_ENV=instrumented webpack --mode=production --color",
     "build-storybook": "storybook build",
     "build-translation": "scripts/po2json.sh",
+    "bundle-stats": "BUNDLE_ANALYZER=true npm run build && npx open-cli ../superset/static/stats/statistics.html",
```
kgabryje, line 46: should we start with `cross-env`?

rusackas, line 46:
```suggestion
    "bundle-stats": "cross-env BUNDLE_ANALYZER=true npm run build && npx open-cli ../superset/static/stats/statistics.html",
```

rusackas: Done!

kgabryje (approved):
LGTM!

michael-s-molina:
Given that we removed `--analyzeBundle` parameter, should we add something to `UPDATING.md`?

rusackas:
> Given that we removed `--analyzeBundle` parameter, should we add something to `UPDATING.md`?

Happy to if you want. I only skipped it since neither Kamil nor I could get the --analyzeBundle parameter to actually WORK. We both just tweak the code to run the analyzer, it turns out. I figured this basically takes a broken tool and makes it work, and it's a dev tool rather than user-facing, so I wasn't sure how much value there is.

What might cover this would actually be a Wiki page explaining what all our scripts do, since there are a TON of them, and there's not really any real documentation on the whole lot.

PR #22853, opened 2023-01-25
chore: Refactor ExploreMixin to power both Datasets (SqlaTable) and Query models

hughhhh:
### SUMMARY

Cleaning up tech debt from intial SIP-68 work to move towards having newer models (Table, Dataset) to run in explore. Since we are deprioritizing that work, refactoring the ExploreMixin to power `SqlaTable` and `Query` which are legacy models. Main reason is to have one codepath on executing queries in explore vs. having divergent paths at the moment.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29884, opened 2024-08-07
fix: Security manager incorrect calls

michael-s-molina:
### SUMMARY
While testing 4.1 migrations, I noticed that the return types of the security manager calls were not being correctly validated by the IDE. To fix this, I added a type hint to our proxy definition and was able to find and fix problems in the code.

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -2274,6 +2274,6 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
         # otherwise the database should have been filtered out
         # in CsvToDatabaseForm
         schemas_allowed_processed = security_manager.get_schemas_accessible_by_user(
-            database, schemas_allowed, True
+            database, None, schemas_allowed, True
         )
         return self.response(200, schemas=schemas_allowed_processed)
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -2274,6 +2274,6 @@ class DatabaseRestApi(BaseSupersetModelRestApi):
         # otherwise the database should have been filtered out
         # in CsvToDatabaseForm
         schemas_allowed_processed = security_manager.get_schemas_accessible_by_user(
-            database, None, schemas_allowed, True
+            database, database.get_default_catalog(), schemas_allowed, True
         )
         return self.response(200, schemas=schemas_allowed_processed)
--- a/superset/extensions/__init__.py
+++ b/superset/extensions/__init__.py
@@ -31,6 +31,7 @@ from superset.async_events.async_query_manager import AsyncQueryManager
 from superset.async_events.async_query_manager_factory import AsyncQueryManagerFactory
 from superset.extensions.ssh import SSHManagerFactory
 from superset.extensions.stats_logger import BaseStatsLoggerManager
+from superset.security.manager import SupersetSecurityManager
 from superset.utils.cache_manager import CacheManager
 from superset.utils.encrypt import EncryptedFieldFactory
 from superset.utils.feature_flag_manager import FeatureFlagManager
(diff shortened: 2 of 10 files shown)


superset/databases/api.py
```diff
@@ -2274,6 +2274,6 @@ def schemas_access_for_file_upload(self, pk: int) -> Response:
         # otherwise the database should have been filtered out
         # in CsvToDatabaseForm
         schemas_allowed_processed = security_manager.get_schemas_accessible_by_user(
-            database, schemas_allowed, True
+            database, None, schemas_allowed, True
```
michael-s-molina, line 2277: This was a problem introduced with the catalog feature. @betodealmeida let me know if `None` is correct in this context or how we should get the catalog.

michael-s-molina:
@betodealmeida @villebro @hughhhh Could you help me fix the following errors?

```
superset/models/helpers.py:824: error: Argument 1 to "get_rls_filters" of "SupersetSecurityManager" 
has incompatible type "ExploreMixin"; expected "BaseDatasource"  [arg-type]

superset/models/helpers.py:834: error: Argument 1 to "get_guest_rls_filters" of "SupersetSecurityManager" 
has incompatible type "ExploreMixin"; expected "BaseDatasource"  [arg-type]
```

This the signature of the methods:

```
def get_rls_filters(self, table: "BaseDatasource") -> list[SqlaQuery]:

def get_guest_rls_filters(self, dataset: "BaseDatasource") -> list[GuestTokenRlsRule]:
```

The calls to the security manager methods were introduced by https://github.com/apache/superset/pull/22853 and it seems that we already needed to pass `table` and `dataset` at that time but we never did. Do you know where to get them from?

villebro:
Changes look good to me, but @betodealmeida and @hughhhh are probably best positioned to comment on the open questions here.

superset/databases/api.py
betodealmeida: Ah, yeah, we need to implement support for catalogs in the CSV upload feature.

Every time you need a catalog in a function and one is not explicitly passed you should use `database.get_default_catalog()` — it will return `None` for databases that don't support catalogs, and the default catalog for those that do.

In this case, we need to pass `database.get_default_catalog()` to `get_schemas_accessible_by_user`, otherwise the permissions check won't match if the database supports catalogs.

betodealmeida:
For the type error, I think the current `get_sqla_row_level_filters` method should be moved out of `ExploreMixin` and into `BaseDatasource`, since only datasets can have RLS filters associarted with them (while a `Query` can't). We might need to leave a no-op method in the `ExploreMixin`, since I assume when building a chart from a query the method will get called.

@hughhhh does this seem right to you?

superset/security/manager.py
```diff
@@ -2645,7 +2645,7 @@ def has_guest_access(self, dashboard: "Dashboard") -> bool:
         dashboards = [
             r
             for r in user.resources
-            if r["type"] == GuestTokenResourceType.DASHBOARD.value
+            if r["type"] == GuestTokenResourceType.DASHBOARD
```
michael-s-molina, line 2648: @betodealmeida @sadpandajoe This have the potential to break current security checks but it's the correct comparison for the defined type. Let me know if it's ok or if we should change the type to `str` instead.

michael-s-molina:
@betodealmeida I think I addressed all your comments.

superset/extensions/__init__.py
```diff
@@ -132,7 +133,7 @@ def init_app(self, app: Flask) -> None:
 migrate = Migrate()
 profiling = ProfilingExtension()
 results_backend_manager = ResultsBackendManager()
-security_manager = LocalProxy(lambda: appbuilder.sm)
+security_manager: SupersetSecurityManager = LocalProxy(lambda: appbuilder.sm)
```
dpgaspar, line 136: nice this will help the IDE a lot!

dpgaspar (approved):
thank you for doing this

PR #29885, opened 2024-08-08
fix: add mutator to get_columns_description

eschutho:
### SUMMARY

Aiming to have the mutator run on every analytical db request, I'm adding it to the get_columns_description which is used when creating a virtual dataset and looking up which columns should be added to the dataset. 

update the `SQL_QUERY_MUTATOR`. As a test, I edited mine to look like this:
```
def mutator(sql: str, database: Any, **kwargs: Any) -> str:

    key = "foo"
    output = f"-- {key}\n"

    return output + f"select color from bart_lines\n-- {key}"


SQL_QUERY_MUTATOR = mutator
```

with the mutator:
Screenshot 2024-08-08 at 3 35 18 PM
Screenshot 2024-08-08 at 3 35 12 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/datasource_tests.py
+++ b/tests/integration_tests/datasource_tests.py
@@ -68,6 +68,24 @@ def create_test_table_context(database: Database):
         engine.execute(f"DROP TABLE {full_table_name}")
 
 
+@contextmanager
+def create_and_cleanup_table(table=None):
+    if table is None:
+        table = SqlaTable(
+            table_name="dummy_sql_table",
+            database=get_example_database(),
+            schema=get_example_default_schema(),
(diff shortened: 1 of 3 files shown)


tests/integration_tests/datasource_tests.py
```diff
@@ -68,6 +68,24 @@ def create_test_table_context(database: Database):
         engine.execute(f"DROP TABLE {full_table_name}")
 
 
+@contextmanager
+def create_and_cleanup_table(table=None):
```
betodealmeida, line 72: ❤️

betodealmeida (approved):

PR #29893, opened 2024-08-08
chore: Logs the duration of migrations execution

michael-s-molina:
### SUMMARY
Depending on the amount of data an organization has, the migration scripts can take a long time to run. This PR adds the duration of the complete migration process to help administrators to estimate downtime for upgrades/downgrades. Duration is only logged in case all migrations successfully complete.

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/env.py
+++ b/superset/migrations/env.py
@@ -55,9 +55,10 @@ target_metadata = Base.metadata  # pylint: disable=no-member
 
 
 def print_duration(start_time: float) -> None:
-    duration = time.time() - start_time
-    formatted_time = time.strftime("%H:%M:%S", time.gmtime(duration))
-    logger.info(f"Migration scripts completed. Duration: {formatted_time}")
+    logger.info(
+        "Migration scripts completed. Duration: %s",
+        time.strftime("%H:%M:%S", time.gmtime(time.time() - start_time)),
+    )
 
 
 def run_migrations_offline() -> None:
@@ -80,7 +81,7 @@ def run_migrations_offline() -> None:
 
     with context.begin_transaction():
         context.run_migrations()
-        print_duration(start_time)
+    print_duration(start_time)
 
 
 def run_migrations_online() -> None:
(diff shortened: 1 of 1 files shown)


superset/migrations/env.py
```diff
@@ -54,6 +54,12 @@
 # ... etc.
 
 
+def print_duration(start_time: float) -> None:
+    duration = time.time() - start_time
+    formatted_time = time.strftime("%H:%M:%S", time.gmtime(duration))
+    logger.info(f"Migration scripts completed. Duration: {formatted_time}")
```
giftig, line 60: Nitpick but usually you should let the logger format the message for you rather than doing it yourself, since the logger will do it lazily and avoid having to do the formatting if INFO leve llogs are off, for example.

superset/migrations/env.py
```diff
@@ -117,6 +130,7 @@ def process_revision_directives(  # pylint: disable=redefined-outer-name, unused
     try:
         with context.begin_transaction():
             context.run_migrations()
+            print_duration(start_time)
```
giftig, line 133: Should this be outside the transaction context? I assume the context manager here is going to commit the transaction on exit, so it's not really done until that happens.

giftig (commented):
A couple of minor nitpicks

superset/migrations/env.py
michael-s-molina: Good point

superset/migrations/env.py
michael-s-molina: Done

giftig (approved):

mistercrunch (approved):
[optional] a context decorator could be nice and simple for this:
```python
with print_duration(label=f"Migration: {migration_id}"):
   do_stuff()
```
Wondering if just using `event_logger.log_context(action='migration')` would just work (in CI, the logging should be `utils.log.StdOutEventLogger`, though it doesn't seem to be the default)

michael-s-molina:
> [optional] a context decorator could be nice and simple for this:

That would be nice. We can think about this when creating the helper functions for the migrations as part of [[SIP-142] Improving Database Migration Management](https://github.com/apache/superset/issues/29546)

PR #29776, opened 2024-07-30
fix(Dashboard): Copying a Dashboard does not commit the transaction

geido:
### SUMMARY
We noticed that when copying a Dashboard the transaction wasn't being committed. This PR introduces a command for the dashboard copy operation and uses the transaction decorator to commit the transaction.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
N.A.

### TESTING INSTRUCTIONS
1. Copy a Dashboard
2. The Dashboard should be copied successfully

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29894, opened 2024-08-08
fix(Embedded): Deleting Embedded Dashboards does not commit the transaction

geido:
### SUMMARY
We noticed that when deleting an Embedded Dashboard the transaction wasn't being committed. This PR introduces a command for the delete embedded dashboard operation and uses the transaction decorator to commit the transaction.

Related https://github.com/apache/superset/pull/29776

N.A.

### TESTING INSTRUCTIONS
1. Deleting an Embedded Dashboard should persist

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/dashboard/delete.py
+++ b/superset/commands/dashboard/delete.py
@@ -23,18 +23,32 @@ from flask_babel import lazy_gettext as _
 from superset import security_manager
 from superset.commands.base import BaseCommand
 from superset.commands.dashboard.exceptions import (
+    DashboardDeleteEmbeddedFailedError,
     DashboardDeleteFailedError,
     DashboardDeleteFailedReportsExistError,
     DashboardForbiddenError,
     DashboardNotFoundError,
 )
-from superset.daos.dashboard import DashboardDAO
+from superset.daos.dashboard import DashboardDAO, EmbeddedDashboardDAO
 from superset.daos.report import ReportScheduleDAO
 from superset.exceptions import SupersetSecurityException
 from superset.models.dashboard import Dashboard
 from superset.utils.decorators import on_error, transaction
 
 logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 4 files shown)


superset/commands/dashboard/delete.py
```diff
@@ -37,6 +38,19 @@
 logger = logging.getLogger(__name__)
 
 
+class DeleteEmbeddedDashboardCommand(BaseCommand):
```
Antonio-RiveroMartnez, line 41: Are we missing the usage of this command in some other place besides the test file?

superset/commands/dashboard/delete.py
```diff
@@ -37,6 +38,19 @@
...
+    def __init__(self, dashboard: Dashboard):
+        self._dashboard = dashboard
+
+    @transaction(on_error=partial(on_error, reraise=DashboardDeleteEmbeddedFailedError))
+    def run(self) -> None:
+        self.validate()
+        return EmbeddedDashboardDAO.delete(self._dashboard.embedded)
+
+    def validate(self) -> None:
+        pass
```
Antonio-RiveroMartnez, line 51: Should we include some minimal validations like: the model exist, etc?

geido: I am not sure what are the implications here. This had no checks originally and I could not find documentation about the expected behavior. So I would not want to touch it unless we have a clear definition of the expected behavior.

superset/commands/dashboard/delete.py
geido: Oh wow, the commit did not include it. Thanks for this!

sadpandajoe:
@supersetbot label 4.1

michael-s-molina (approved):

Antonio-RiveroMartnez (approved):

Issue #26041, opened 2023-11-20
Bar chart with time delay displayed with the same color


martin-raymond:
Before the 3.0.1 : 
![image](https://github.com/apache/superset/assets/10060967/fe32c18d-dbaa-418a-9347-07f5ea6fdc2f)

After the 3.0.1 : 
![image](https://github.com/apache/superset/assets/10060967/ca7b8053-972e-439a-962f-1a0ca828647b)

#### How to reproduce the bug

1. Create a bar chart
2. Add a time delay to visualize a comparison between the bar (a year, a month, it does not matter)
3. Before the 3.0.1, each couple of bar had two colors, now they are all in the same color

### Environment

- superset version: `3.0.1`

sfirke:
When you say "before 3.0.1" do you mean 3.0.0 did not have this problem?  What's the latest official release you've found that has the prior, desired behavior?

martin-raymond:
we were using the 2.0.1 version, and now we have upgraded to 3.0.1 @sfirke

michael-s-molina:
This was a side-effect of another [fix](https://github.com/apache/superset/pull/24048). They should have the same color but we need to add a differentiator between normal and offset bars similar to what the line chart does by using dashed lines.

CairoLee:
I also have the same question. For line charts, using dashed lines to differentiate offset lines for the same indicator is possible. But for bar charts, I think maintaining the previous logic might seem more reasonable.

jack-chao-htg:
In line charts, it makes the time-delay lines dotted instead of solid but in the same color. This is pretty nice, however, if you have multiple time-delay lines, they will all look identical.

michael-s-molina:
Thank you all for the discussion.

> I prefer to have them display different colors, especially when we are only drawing a single indicator and want to compare the data from a week ago. Showing different colors would be very convenient for observation.

My concerns here would be:
- Consistency on how we display offsets between charts. It will be hard for users to understand how to view offsets if we use different color strategies for every chart.
- Multiple series and their offsets. If a bar chart has multiple series, it's hard to identify the offsets if the colors don't match. 

> In line charts, it makes the time-delay lines dotted instead of solid but in the same color. This is pretty nice, however, if you have multiple time-delay lines, they will all look identical.

You're correct. We would be need have a number of patterns (dashed, solid, dotted, etc) equal to the number of supported offsets to differentiate the metrics.

Tagging @kasiazjc to bring the designer's perspective to the discussion.

[...]

PR #24048, opened 2023-05-12
feat: Make time shifted series colors match the original series

michael-s-molina:
### SUMMARY
This PR upgrades Timeseries ECharts to display time shifted series as dashed lines and using the same color as the original series.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
https://github.com/apache/superset/assets/70410625/05bfa3ee-b387-4d24-b5fc-e8ac697cbd48

### TESTING INSTRUCTIONS
Check the video for instructions.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29897, opened 2024-08-08
feat(explore): Add time shift color control to ECharts

rtexelm:
### SUMMARY

Adds a new checkbox control to timeseries capable charts allowing the user to choose whether or not to duplicate the original series colors for the shifts.

https://github.com/user-attachments/assets/0df125af-3917-4a0e-b5c8-927357bbf3a4

### TESTING INSTRUCTIONS

1. Create a new chart using the Sales dataset.
2. Use `COUNT(*)` as your metric.
3. Set `order_date` as the X-Axis and a Month time grain.
4. Apply any time range filter that returns data.
5. Expand the time comparison and set 1 month ago.
6. Click on `CREATE CHART`.
7. Navigate to `Customize` tab
8. Click the `Time Shift color` checkbox

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/sharedControls.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/sharedControls.tsx
@@ -340,12 +340,16 @@ const color_scheme: SharedControlConfig<'ColorSchemeControl'> = {
 
 const time_shift_color: SharedControlConfig<'CheckboxControl'> = {
   type: 'CheckboxControl',
-  label: t('Time Shift Color'),
+  label: t('Match time shift color with original series'),
   default: true,
   renderTrigger: true,
   description: t(
     'When unchecked, colors from the selected color scheme will be used for time shifted series',
   ),
+  visibility: ({ controls }) =>
+    Boolean(
+      controls?.time_compare?.value && !isEmpty(controls?.time_compare?.value),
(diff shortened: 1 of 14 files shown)


michael-s-molina:
The only concern that I have with this approach is about charts consumption at the dashboard layer. It might be confusing for users that time shifts are represented differently among charts which might harm UX. I'm not sure if we should discuss more about this as you can see in https://github.com/apache/superset/issues/26041

michael-s-molina:
Another option is to assume that all time shifts will be of varied colors and remove the need for this checkbox. Essentially, not make the correlation between the series and its time shift using colors, only the legend.

yousoph:
I'm not too concerned about the dashboard level - I'm guessing that there will be consistency at the org or creator level with this control based on preference so there aren't likely to be too many discrepancies at the dashboard level. 

In terms of defaulting to use varied colors, I'm not opposed to that but I thought the original fix in https://github.com/apache/superset/pull/24048 was to retain the same color for the time shift as the original control. Since that was the fix that was implemented, we thought adding a control would be the best way to give the flexibility to use either option.

Antonio-RiveroMartnez (approved):
LGTM, Thanks!

rtexelm:
@fpassantino sorry for the delay.

Issue #27645, opened 2024-03-25
[SIP-123] Proposal replacement of data table components with ag-grid


justinpark:
### Motivation

The data table, similar to a query result in SQL LAB, has been extensively utilized for data exploration. While the current ant-table has been a satisfactory solution, it has limitations when it comes to supporting advanced features.
Users have expressed their desire for more powerful functionalities such as "resizing a column width," "pinning/unpinning a column," "hiding/unhiding a column," and "multiple column sorting." Although there is a demo version available in ant-table that supports some of these features, it is still premature and buggy, and no longer actively maintained.
Fortunately, there are several alternative open-source solutions that already support these features, making replacement a suitable approach in the current situation.

### Proposed Change

After thorough benchmarking of various open-source solutions, ag-grid has emerged as the best choice due to its extensive APIs and support for the advanced features we require (as mentioned above). Additionally, it is fully MIT licensed and has a separate enterprise version, making it a reliable option for future maintenance.
At Airbnb, we have successfully built a data table using the community version of ag-Grid and have utilized its open APIs to enable all the desired advanced features. It has been working well for us since this year. Today, I am excited to share this solution with the open-source community.

![Image](https://github.com/apache/superset/assets/1392866/82185b00-29ea-406b-b07b-fc0d58fd4255)

### New or Changed Public Interfaces

There are no interface changes. We are simply remapping the existing properties to the new ag-Grid component.

### New Dependencies

ag-grid-community / ag-grid-react

### Migration Plan and Compatibility

The migration process starts with the Table component in FilterableTable, which is used in the SQL Lab query result. Following that, we will migrate the TableCollection used in explore/chart result preview, and then the Table components(i.e. Pivot Table, DataTable) in chart-plugins.

As ag-theme-quartz comes [recommended](https://ag-grid.com/javascript-data-grid/themes/#provided-themes), we will be customizing the style based on the quartz theme.

Since Ag-grid conveniently provides [CSS variables](https://ag-grid.com/javascript-data-grid/global-style-customisation-variables/) for custom styling, we can easily override these variables to achieve the desired style. To apply the quartz theme, all ag-grid tables will be enclosed within the `.ag-theme-quartz` class name. This will allow us to easily customize the style to match the existing theme style. For example, let's take a look at how we can match the font-family as an example.

```
 <Global
    styles={({ theme }) => css`
        .ag-theme-quartz {
          --ag-font-family: ${theme.typography.families.sansSerif};
        }
    `}
 />
```

[...]

PR #29900, opened 2024-08-09
feat(sqllab): Replace FilterableTable by AgGrid Table

justinpark:
### SUMMARY
Following up #27645, This commit includes the migration of the component used for SQLLab results from the ant-table component to the ag-grid component.
During this process, the existing logic(`getWidthsForColumns`) that measured column size to adjust the width was removed because it was not only performance-intensive but also unnecessary, as the resizable column option in ag-grid allows users to adjust the width as needed.
Additionally, other property options were modified to ensure compatibility with ag-grid. Moreover, a column menu feature was added, enabling users to pin specific columns and apply filters.

Before:

Screenshot 2024-08-09 at 7 33 03 AM

After:

Screenshot 2024-08-20 at 11 46 53 AM

Header menu (Actions):

Image

### TESTING INSTRUCTIONS

Go to SQL Lab and run a query to explore the data

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


kasiazjc:
@justinpark this looks great! I think I might have asked this question before, so apologies if I did, but for how long custom table settings are kept - just for session and just for this specific query?

justinpark:
> how long custom table settings are kept - just for session and just for this specific query?

This will only apply to the current query in the current session. I considered saving it, but I thought it might be better not to, as revisiting the query later when the memory has faded could cause confusion with different items being displayed in the query results.

superset-frontend/package.json
```diff
@@ -126,6 +126,8 @@
     "@visx/xychart": "^3.5.1",
     "abortcontroller-polyfill": "^1.7.5",
     "ace-builds": "^1.35.4",
+    "ag-grid-community": "31.0.3",
```
kgabryje, line 129: Can we use the latest 32.2 version?

justinpark: done

Hokwang:
when is the release target date  and release version ?

eschutho:
> when is the release target date and release version ?

@Hokwang this will most likely be released in 5.0 around April 2025.

Hokwang:
@eschutho Thanks for your comment, but I can't find [5.0 project](https://github.com/orgs/apache/projects/345). Please add this item.

michael-s-molina:
> @eschutho Thanks for your comment, but I can't find at [5.0 project](https://github.com/orgs/apache/projects/345). Please add this item.

@Hokwang The 5.0 project board contain only breaking changes that need consensus which is not the case for this PR. That does not mean that only what's in the board will be released. On February 3rd, we'll cut from master branch and all PRs merged by then will be included in the release. Check our [Release Process](https://github.com/apache/superset/wiki/Release-Process) for more details.

Hokwang:
@michael-s-molina Oh, I see. Thanks for more detail information.

superset/config.py
```diff
@@ -1592,7 +1592,7 @@ def EMAIL_HEADER_MUTATOR(  # pylint: disable=invalid-name,unused-argument  # noq
 TALISMAN_CONFIG = {
     "content_security_policy": {
         "base-uri": ["'self'"],
-        "default-src": ["'self'"],
+        "default-src": ["'self'", "data:"],
```
michael-s-molina, line 1595: When you include `data:` in your CSP's `default-src` or any other directive, it allows any data URI to be loaded, regardless of its source. This means that any data URL can be executed or rendered by the browser, which can pose security risks such as exposing your application to cross-site scripting (XSS) attacks.

To restrict sources to trusted domains, you can specify the domains explicitly in the `default-src` or other specific directives (like `script-src`, `style-src`, etc.)

```
Content-Security-Policy: default-src 'self' https://trusted.com; 
```

justinpark: update to font-src

michael-s-molina: @justinpark Can you restrict `font-src` to trusted domains?

superset-frontend/src/components/FilterableTable/index.tsx
```diff
@@ -188,86 +109,52 @@ const FilterableTable = ({
...
   );
 
-  // exclude the height of the horizontal scroll bar from the height of the table
-  // and the height of the table container if the content overflows
-  const totalTableHeight =
-    container.current && totalTableWidth.current > container.current.clientWidth
-      ? height - SCROLL_BAR_HEIGHT
-      : height;
+  const keyword = useRef<string | undefined>(filterText);
+  keyword.current = filterText;
```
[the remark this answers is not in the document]
michael-s-molina: @justinpark I'm not sure if you missed this comment?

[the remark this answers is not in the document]
justinpark: This is the process of updating the ref object to ensure that the keywordFilter function references the latest updated keyword value. If the keyword is added to the dependencies of the keywordFilter, grid table will rerender every time the keyword changes.
I change it to reference ref.current in order to prevent it.

michael-s-molina (approved):

PR #29911, opened 2024-08-10
fix: Rename database from 'couchbasedb' to 'couchbase' in documentation and db_engine_specs

ayush-couchbase:
### SUMMARY
Changing Name from couchbasedb to couchbase. Makes it more clear and logical about underlying database.
No major change done.
### TESTING INSTRUCTIONS
mannual testing

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/RESOURCES/INTHEWILD.md
+++ b/RESOURCES/INTHEWILD.md
@@ -80,7 +80,6 @@ Join our growing community!
 - [Caizin](https://caizin.com/) [@tejaskatariya]
 - [Careem](https://www.careem.com/) [@SamraHanifCareem]
 - [Cloudsmith](https://cloudsmith.io) [@alancarson]
-- [CnOvit](http://www.cnovit.com/) [@xieshaohu]
 - [Cyberhaven](https://www.cyberhaven.com/) [@toliver-ch]
 - [Deepomatic](https://deepomatic.com/) [@Zanoellia]
 - [Dial Once](https://www.dial-once.com/)
--- a/RESOURCES/INTHEWILD.md
+++ b/RESOURCES/INTHEWILD.md
@@ -80,6 +80,7 @@ Join our growing community!
 - [Caizin](https://caizin.com/) [@tejaskatariya]
 - [Careem](https://www.careem.com/) [@SamraHanifCareem]
 - [Cloudsmith](https://cloudsmith.io) [@alancarson]
+- [CnOvit](https://www.cnovit.com/) [@xieshaohu]
 - [Cyberhaven](https://www.cyberhaven.com/) [@toliver-ch]
 - [Deepomatic](https://deepomatic.com/) [@Zanoellia]
 - [Dial Once](https://www.dial-once.com/)
--- a/superset/db_engine_specs/couchbase.py
+++ b/superset/db_engine_specs/couchbase.py
@@ -74,14 +74,14 @@ class CouchbaseParametersSchema(Schema):
     )
 
 
-class CouchbaseDbEngineSpec(BasicParametersMixin, BaseEngineSpec):
-    engine = "couchbasedb"
+class CouchbaseEngineSpec(BasicParametersMixin, BaseEngineSpec):
+    engine = "couchbase"
     engine_name = "Couchbase"
-    default_driver = "couchbasedb"
+    default_driver = "couchbase"
     allows_joins = False
     allows_subqueries = False
     sqlalchemy_uri_placeholder = (
-        "couchbasedb://user:password@host[:port]?truststorepath=value?ssl=value"
+        "couchbase://user:password@host[:port]?truststorepath=value?ssl=value"
     )
     parameters_schema = CouchbaseParametersSchema()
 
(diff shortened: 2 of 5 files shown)


superset/sql_parse.py
```diff
@@ -102,7 +102,7 @@
     "clickhouse": Dialects.CLICKHOUSE,
     "clickhousedb": Dialects.CLICKHOUSE,
     "cockroachdb": Dialects.POSTGRES,
-    "couchbasedb": Dialects.MYSQL,
+    "couchbase": Dialects.MYSQL,
```
villebro, line 105: This would entail a breaking change, and would break every Couchbase connection created by existing users. At minimum we would need to create an alias for the old dialect name to ensure old connections still work.

villebro (changes requested):
There are some issues reported by CI. In addition, the breaking change is problematic, and would need to at least be addressed in `UPDATING.md` and/or by adding an alias to the old name.

superset/sql_parse.py
ayush-couchbase: I have corrected this in new patch. I am adding alias to old name also. I am seeing all checks passed too.

villebro: > I am adding alias to old name also.

I don't see the alias here - can you check how the alias is implemented in the Postgres spec?

villebro (commented):
Second pass comment

superset/sql_parse.py
ayush-couchbase: My mistake I thought alias means updating sql_parse , I saw oceanbase and postgres db_engine_specs and similar to that I have added `engine_aliases = {"couchbase", "couchbasedb"}` to db_engine_spec/couchbase.py .

superset/db_engine_specs/couchbase.py
```diff
@@ -74,14 +74,15 @@ class CouchbaseParametersSchema(Schema):
     )
 
 
-class CouchbaseDbEngineSpec(BasicParametersMixin, BaseEngineSpec):
-    engine = "couchbasedb"
+class CouchbaseEngineSpec(BasicParametersMixin, BaseEngineSpec):
+    engine = "couchbase"
+    engine_aliases = {"couchbase", "couchbasedb"}
```
villebro, line 79: You donät need the main name in the alias set, this should be enough:
```suggestion
    engine_aliases = {"couchbasedb"}
```

villebro (commented):
One last nit, after that this LGTM 👍

superset/db_engine_specs/couchbase.py
ayush-couchbase: Updated

ayush-couchbase: There are 2 failed checks, but they were not failing earlier, are they because of my patch ?

villebro: The failed MySQL test is flaky - we can restart the test. The docs one seems unrelated - It seems `resources/INTHEWILD.md` has a link to `/www.cnovit.com` that isn't working. I think we may need to remove the link if the homepage isn't up anymore. Do you want to do it in this PR? I can also open a PR to fix it if needed.

villebro:
@michael-s-molina this will not be a breaking change as per the last changes. So I think it's safe to remove that label.

superset/db_engine_specs/couchbase.py
ayush-couchbase: Fixed

RESOURCES/INTHEWILD.md
```diff
@@ -80,7 +80,6 @@ Join our growing community!
 - [Caizin](https://caizin.com/) [@tejaskatariya]
 - [Careem](https://www.careem.com/) [@SamraHanifCareem]
 - [Cloudsmith](https://cloudsmith.io) [@alancarson]
-- [CnOvit](http://www.cnovit.com/) [@xieshaohu]
```
rusackas, line 83: Is this removal intentional? Maybe you meant to add Couchbase?

ayush-couchbase: This link is not working. So tests were failing. @villebro is discussing about it in previous conversation.

villebro: @ayush-couchbase it seems the link was failing due to being `http` and not `https` - could you update that? The `https` one works for me at least. After that I think we're good to go.

ayush-couchbase: Sure , Updated in new patch

villebro (approved):
LGTM, thanks for all the patience and iterations!

RESOURCES/INTHEWILD.md
rusackas: FYI, the link checker does not block merging... it just gives us little to-do items as maintainers. I'll open a PR today to sweep up whatever's left at the moment.

rusackas:
FYI @sadpandajoe the author would like to get this into the next release opportunity. I tagged it as 4.1 in case there's an RC4, but let me know if there's anything else I should do to get it into 4.1.1 if RC3 passes the vote.

PR #29916, opened 2024-08-12
fix: duplicate `truncateXAxis` option in `BarChart`

dmitriyVasilievich1986:
### SUMMARY
In control panel for echart/BarChart there is a factory-function that supposed to build panels for x/y axises. Unfortunately `truncateXAxis` and `xAxisBounds` do not affected by axis name, therefore those options appears in both panels('X Axis', 'Y Axis'). They are complete duplicate and share the same state. Solution is to move those options from factory-function directly to `controlSetRows`.

Screenshot 2024-08-12 at 11 35 38
Screenshot 2024-08-12 at 11 57 19

### TESTING INSTRUCTIONS
The problem is available in the chart editor in the customization panel. Since both options have a same state, the value changes in both.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
@@ -213,8 +213,6 @@ function createAxisControl(axis: 'x' | 'y'): ControlSetRow[] {
         },
       },
     ],
-    [truncateXAxis],
-    [xAxisBounds],
     [
       {
         name: 'truncateYAxis',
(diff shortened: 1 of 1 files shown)


kgabryje:
@dmitriyVasilievich1986 Thanks for the fix! I wonder if we should tie `truncateXAxis` position to the bars' orientation. In the default "vertical" mode, `truncate x axis` truncates the horizontal axis (X) as expected. But if you change the orientation to horizontal, then truncate x axis actually truncates the vertical Y axis.
So maybe we should override `truncateXAxis` control's visibility and label (X if vertical, Y if horizontal) based on the orientation. WDYT?

dmitriyVasilievich1986:
Good day. I looked at what could be done to solve this problem. Unfortunately I think that in order to use generic `truncateXAxis` and `xAxisBounds` panels you will have to make too many changes. The easiest option is to delete those generic ones and instead make a copy of the abstract ones for the y-axis. Replacing the name and visibility conditions.
I tested this option and switching the chart orientation worked without problems. I can make a commit with this option. Or am I missing something else?

kgabryje (approved):
Got it, thanks for the context @dmitriyVasilievich1986. LGTM

Issue #22979, opened 2023-02-03
Dependent filter selects first value independently of the filter it depends on, if it is set to "Select first filter value as default"


e-geist:
Hi,

we experienced that using required filters in Dashboards that cascade/depend on each other with "first value" as default lead to invalid filter combinations upon first dashboard load.

If a filter "events" exists whose values are defined as dependent on a filter "year" and both filters are
- required
- set to "Select first value by default"

upon loading a dashboard, the first value in the filter "year" will be picked (as supposed) and the first value in the filter "events" will be picked - even if the event doesn't exist for the picked year.
The Charts are then queried with this invalid filter combination and lead to "No data" - until the user fixes the picked filters. 

#### How to reproduce the bug

1. Create a dashboard with two filters and some charts these filters are applied to
2. Make both filters required and "Select first filter value as default"
3. Make one of the filters dependent on the other
4. Make sure the first value of the dependent filter is not available with the first value of the other filter
5. Load dashboard in UI
6. First value of both filters is chosen automatically -> invalid combination of filters is chosen -> "No data" in charts

### Expected results

The dependent filters' value is chosen dependent on the other filters' value - so:
First value of first filter is selected -> second filters' values are filtered according to first filter's value -> second filters' filtered list is used to choose first value

### Actual results

Both filters' values are chosen indepedently from each other.

### Environment
- browser type and version: Firefox
- superset version: 2.0.1
- python version: Python 3.8.12
- any feature flags active: nothing too special - we use embedding dashboards, but this also occurs for normal dashboards

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [x] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [x] I have reproduced the issue with at least the latest released version of superset.
- [x] I have checked the issue tracker for the same issue and I haven't found one similar.
- [x] I have asked other people in Superset Slack - it seems to be reproducible also for others (see here: https://apache-superset.slack.com/archives/C016B3LG5B4/p1675352073631509)

If you have any further questions or anything is unclear, please let me know.

sfirke:
I also experience this and would just add: once the user makes a selection for the 2nd variable, they are correctly restricted from choosing an invalid combination.  I.e., the filter 2 choices correctly depend on filter 1.  I hope that means this would be fixable.

fmannhardt:
I have encountered the same issue. Unfortunately, anyone using the "required value" setting and a cascading filter is forced to select a default value leading to this bug. 

[...]

PR #29932, opened 2024-08-13
Fixed issue raised in ticket #22979

anantaoutlook:
Fixed issue raised in ticket https://github.com/apache/superset/issues/22979

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "master"
+    ]
+}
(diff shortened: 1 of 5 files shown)


sadpandajoe:
@anantaoutlook i see parts where you commented out the code, can we clean those up?

anantaoutlook:
> @anantaoutlook i see parts where you commented out the code, can we clean those up?

> Definitely

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
kgabryje, line 1: This change isn't probably related to the issue?

PR #29935, opened 2024-08-13
fix: Fix delete_fake_db

stamplevskiyd:
Function delete_fake_db misses db.session.commit(), which may result in test fake database not being deleted, leading to failing subsequent tests.

delete_fake_db_for_macros, which performs a similar function, includes this line:

```python
def delete_fake_db(self):
    database = (
        db.session.query(Database)
        .filter(Database.database_name == FAKE_DB_NAME)
        .scalar()
    )
    if database:
        db.session.delete(database)
        db.session.commit()  # Added
```

```python
@staticmethod
def delete_fake_db_for_macros():
    database = (
        db.session.query(Database)
        .filter(Database.database_name == "db_for_macros_testing")
        .scalar()
    )
    if database:
        db.session.delete(database)
        db.session.commit()
```

Also consider making delete_fake_db a staticmethod:
```python
@staticmethod
def delete_fake_db():
    database = (
        db.session.query(Database)
        .filter(Database.database_name == FAKE_DB_NAME)
        .scalar()
    )
    if database:
        db.session.delete(database)
        db.session.commit()  # Added
```

### DIFF
--- a/tests/integration_tests/base_tests.py
+++ b/tests/integration_tests/base_tests.py
@@ -443,6 +443,7 @@ class SupersetTestCase(TestCase):
         )
         if database:
             db.session.delete(database)
+            db.session.commit()
 
     def create_fake_db_for_macros(self):
         database_name = "db_for_macros_testing"


michael-s-molina (approved):

stamplevskiyd:
```python
    def test_export_datasource_ui_cli(self):
        # TODO(bkyryliuk): find fake db is leaking from
        self.delete_fake_db()
 ```
 
 fake_db_leaking TODO (tests/integration_tests/dict_import_export_tests.py) may be also fixed by this PR

PR #21248, opened 2022-08-30
feat: filter parameters from DB API

betodealmeida:
### SUMMARY

Filter out some of the parameters from the database connection in the API.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #29943, opened 2024-08-14
chore: improve mask/unmask encrypted_extra

betodealmeida:
### SUMMARY

When the user edits a database, the password in the SQLAlchemy URI is masked (replaced with `XXXXXXXXXX`), so that the sensitive information is not transmitted down the wire unneedlessly. If the user saves the database edit, the mask is replaced with the original password, unless it has been changed.

The `encrypted_extra` field in a database may also contain sensitive fields. Instead of masking the whole field, in https://github.com/apache/superset/pull/21248 we introduced logic to mask only selected fields, so we can provide a better editing experience for users. For BigQuery, eg, the credentials look like this:

```json
{
  "type": "service_account",
  "project_id": "your project id",
  "private_key_id": "your private key id",
  "private_key": "-----BEGIN PRIVATE KEY-----...-----END PRIVATE KEY-----\n",
  ...
}
```

When editing, it will show in database modal as:

```json
{
  "type": "service_account",
  "project_id": "your project id",
  "private_key_id": "your private key id",
  "private_key": "XXXXXXXXXX",
  ...
}
```

The user can then selectively change some of the fields, and when the database is edited the `private_key` will be replaced with the old value if it's equal to `XXXXXXXXXX`.

Currently this functionality is only implemented for BigQuery and GSheets, but it should be implemented in all databases that store sensitive information under `encrypted_extra`. To simplify the process and remove duplicate code, in this PR I updated the methods in the base class to work with JSON paths. For BigQuery, for example, all that is needed now is:

```python
encrypted_extra_sensitive_fields = ["$.credentials_info.private_key"]
```

This PR also adds the logic for Snowflake.

N/A

### TESTING INSTRUCTIONS

Confirmed existing tests work, added a few more tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -45,6 +45,7 @@ from deprecation import deprecated
 from flask import current_app, g, url_for
 from flask_appbuilder.security.sqla.models import User
 from flask_babel import gettext as __, lazy_gettext as _
+from jsonpath_ng import parse
 from marshmallow import fields, Schema
 from marshmallow.validate import Range
 from sqlalchemy import column, select, types
@@ -59,7 +60,7 @@ from sqlalchemy.types import TypeEngine
 from sqlparse.tokens import CTE
 
 from superset import sql_parse
-from superset.constants import TimeGrain as TimeGrainConstants
+from superset.constants import PASSWORD_MASK, TimeGrain as TimeGrainConstants
 from superset.databases.utils import get_table_metadata, make_url_safe
 from superset.errors import ErrorLevel, SupersetError, SupersetErrorType
 from superset.exceptions import DisallowedSQLFunction, OAuth2Error, OAuth2RedirectError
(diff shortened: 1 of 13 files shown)


superset/db_engine_specs/base.py
```diff
@@ -2160,29 +2161,63 @@ def get_impersonation_key(cls, user: User | None) -> Any:
         """
         return user.username if user else None
 
+    # list of JSON path to fields in `encrypted_extra` that should be masked when the
+    # database is edited
+    # pylint: disable=invalid-name
+    encrypted_extra_sensitive_fields: list[str] = []
```
mistercrunch, line 2167: [optional] usually we put class-level attributes at the top above all the methods, but I can see how it's nice to have it there right by the method that uses this one... Personally I'd expect this being at the top

mistercrunch: [optional] another approach here would be to assume all keys in `encrypted_extra_sensitive_fields` are sensitive and allow-list the ones that we know are not sensitive. Seems it would be more cautious overall if it's not more work (?)

superset/db_engine_specs/base.py
```diff
@@ -45,6 +45,7 @@
 from flask import current_app, g, url_for
 from flask_appbuilder.security.sqla.models import User
 from flask_babel import gettext as __, lazy_gettext as _
+from jsonpath_ng import parse
```
mistercrunch, line 48: [optional] `jsonpath_ng.parse` could be wrapped in `utils/json.py`, maybe something like `def redact_sensitive(json, sensitive_paths)`

mistercrunch (approved):
Added a few optional optional comments for consideration, otherwise LGTM

superset/db_engine_specs/base.py
betodealmeida: Ah, good point about the default behavior. We could have the attribute in the base class default to:

```python
["$"]
```

This way everything would be masked.

superset/db_engine_specs/base.py
betodealmeida: Another good point! I've had issues with JSONPath libraries in the past, wrapping them in a helper method will help in the future.

mistercrunch:
mp

mistercrunch:
... may cause seizures ...

PR #29944, opened 2024-08-14
fix: only show dataset name in list

eschutho:
### SUMMARY

Shows just the dataset name, as with the schema it tends to be too long. I moved the full schema/dataset name into the tooltip. The link remains the same. 

https://github.com/user-attachments/assets/16698417-4dfd-4397-a845-3ef2bba22410

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/pages/ChartList/index.tsx
+++ b/superset-frontend/src/pages/ChartList/index.tsx
@@ -378,7 +378,11 @@ function ChartList(props: ChartListProps) {
               datasource_url: dsUrl,
             },
           },
-        }: any) => <GenericLink to={dsUrl}>{dsNameTxt}</GenericLink>,
+        }: any) => (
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM. Doesn't seem like a breaking change since the info is still available.

michael-s-molina:
It may be misleading to not show the schema alongside the dataset name given that many datasets have the same name. I understand that users can still view this information in the tooltip, but I don't think it's sufficient. It also kind of force the user to hover the name to see if it's actually the dataset from a particular schema. Maybe add a column for schema?

michael-s-molina:
There's no schema filter either. Maybe it would be good to add one, specially if we have a schema column.

PR #29959, opened 2024-08-16
feat(sqllab): Add timeout on fetching query results

justinpark:
### SUMMARY
This commit adds the timeout on fetching query results.
Additionally, a 'Retry without timeout' option is provided so that when a timeout issue occurs, users can attempt to proceed by ignoring the error.

https://github.com/user-attachments/assets/af5dfc12-4072-484a-94d8-57b3258b3163

### TESTING INSTRUCTIONS
SET SQLLAB_QUERY_RESULT_TIMEOUT with a small timeout
Run a query async
Check the error message and then click "Retry" to get the result without timeout errors

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
@justinpark we currently have

```
# Timeout duration for SQL Lab synchronous queries
SQLLAB_TIMEOUT = int(timedelta(seconds=30).total_seconds())

# The MAX duration a query can run for before being killed by celery.
SQLLAB_ASYNC_TIME_LIMIT_SEC = int(timedelta(hours=6).total_seconds())
```

Can't we reuse one of these for this? I'm thinking that we could expand the scope of `SQLLAB_TIMEOUT` (changing the comment) to be valid for both sync and async queries.

michael-s-molina:
Another question is about retry without time limit. Isn't this dangerous and probably inaccurate? I'm assuming all database engines will impose timeouts and this setting will not bypass this? I'm also assuming that a timeout was set for a reason, so giving users the ability to override this seems dangerous. If you're thinking only hanging frontend requests, wouldn't https://github.com/apache/superset/pull/29956 be sufficient to handle these cases?

justinpark:
> Can't we reuse one of these for this? I'm thinking that we could expand the scope of `SQLLAB_TIMEOUT` (changing the comment) to be valid for both sync and async queries.

@michael-s-molina QUERY_RESULT_TIMEOUT is different from the existing SQLLAB_TIMEOUT and SQLLAB_ASYNC_TIME_LIMIT_SEC. This refers to a request that occurs in an ASYNC QUERY, which is different from the actual query execution time. It is a request to retrieve the completed query result from the cache. In formula form, it can be expressed as follows:
```
SQLLAB_TIMEOUT = SQLLAB_ASYNC_TIME_LIMIT_SEC + SQLLAB_QUERY_RESULT_TIMEOUT
```

The reason for delays related to the timeout is due to locking caused by updates to the query table, which leads to delays in querying the information linked to the resultsKey of the query ID. To address this issue, a separate timeout is configured.

> Another question is about retry without time limit. Isn't this dangerous and probably inaccurate? I'm assuming all database engines will impose timeouts and this setting will not bypass this? I'm also assuming that a timeout was set for a reason, so giving users the ability to override this seems dangerous. If you're thinking about zombie frontend requests, wouldn't https://github.com/apache/superset/pull/29956 be sufficient to handle these cases?

The locking delay in synchronizing query status was resolved with PR 29956. This patch also addresses the intermittent issue of the fetching state freezing (when the timer stops and results are retrieved) that occasionally occurred after the query status was confirmed as successful.

Making a request without a timeout applies the max request time related on the HTTP server, which is the same as the existing result retrieval process. The new addition is simply an update to the UX, where pending states are cut off slightly earlier than the HTTP server’s max request timeout. If a request requires more time than the configured timeout, users can re-initiate the request using the original method, depending on their preference.

justinpark:
Screenshot 2024-09-11 at 12 13 32 PM
Updated the button label to "Retry fetching results" which will run as-is (which means no overrides of timeout)

michael-s-molina (approved):

justinpark:
@sadpandajoe Could you add this fix on 4.1?

PR #29966, opened 2024-08-18
fix(#29965): add assets prefix on all manifest assets

### LINKED ISSUE #29965: Missing prefix in URLs of scripts and CSS files when STATIC_ASSETS_PREFIX config is set
### Bug description

Per document on STATIC_ASSETS_PREFIX in config, it should be the prefix of all static assets URLs. However, it's not true for most of JS and CSS files. I believe it's a bug in `superset/extensions/__init__.py`. The `UIManifestProcessor.get_manifest` method should add STATIC_ASSETS_PREFIX to all assets listed in manifest.

### How to reproduce the bug

1. Set STATIC_ASSETS_PREFIX to a custom host and restart Superset server.
2. Reload Superset website and monitor network traffic in DevTools. We can see that most of JS and CSS files are not downloaded from the host we just set in step 1.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

huandu:
### SUMMARY

Per #29965, this PR fixes the issue that static assets prefix is not added to static assets listed in manifest.json.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #29965
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -59,6 +59,7 @@ assists people when migrating to a new version.
   as part of your bundling to expose translation packages, it's probably not needed anymore.
 - [29264](https://github.com/apache/superset/pull/29264) Slack has updated its file upload api, and we are now supporting this new api in Superset, although the Slack api is not backward compatible. The original Slack integration is deprecated and we will require a new Slack scope `channels:read` to be added to Slack workspaces in order to use this new api. In an upcoming release, we will make this new Slack scope mandatory and remove the old Slack functionality.
 - [29798](https://github.com/apache/superset/pull/29798) Since 3.1.0, the intial schedule for an alert or report was mistakenly offset by the specified timezone's relation to UTC. The initial schedule should now begin at the correct time.
+- [29966](https://github.com/apache/superset/pull/29966) If the config `STATIC_ASSETS_PREFIX` is set, bundles will now be fetched from that location.
 
 ### Potential Downtime
 
--- a/superset/extensions/__init__.py
+++ b/superset/extensions/__init__.py
@@ -16,7 +16,7 @@
 # under the License.
 import json
 import os
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Union
 
 import celery
 from flask import Flask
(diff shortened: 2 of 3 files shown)


rusackas (approved):

rusackas:
Thanks for the contribution! You need to get the pre-commit check to pass. Superset uses Git pre-commit hooks courtesy of [pre-commit](https://pre-commit.com/). To install run the following:
```
pre-commit install
```
```
pre-commit run --all-files
```

mistercrunch:
> You need to get the pre-commit check to pass. 

We could add a message in the logs https://github.com/apache/superset/blob/master/.github/workflows/pre-commit.yml#L43-L47, let me try to write a PR that adds a note so new contributors can find the error/solution as they click the failing check

mistercrunch:
@rusackas @kgabryje , wondering how this work on our infra at preset.io, how did we know to fetch the bundles from CDN? Or did do we fetch bundles from the webserver? Could this break our build?

mistercrunch (changes requested):
Actually let's make sure we won't break things for people, or if we do we should add a line item to `UDPATING.md` as in "if you use the `STATIC_ASSETS_PREFIX` config, bundles will now be fetched from that location"

mistercrunch:
Here, this should help: https://github.com/apache/superset/pull/29970

huandu:
@rusackas Thanks for your guidance. I've updated the PR and passed all pre-commit checks.

huandu:
> Actually let's make sure we won't break things for people, or if we do we should add a line item to `UDPATING.md` as in "if you use the `STATIC_ASSETS_PREFIX` config, bundles will now be fetched from that location"

I can add this notice in `UPDATING.md`. I'm not sure which section it should be added to, `4.0.0` or `Next` or others?

BTW, from my perspective, when I read the comment for `STATIC_ASSETS_PREFIX`, my understanding is that this config can affect all static assets, including all JS and CSS files used by Superset website. So I think this PR is a bug fix rather than a feature change. What do you think?

https://github.com/apache/superset/blob/cf083bf827e477987d6534f57b6f9dbd4b8ab22a/superset/config.py#L1631-L1633

eschutho:
@huandu, this could probably be better documented, but can you try to set an environment variable `ASSET_BASE_URL` to match your `STATIC_ASSETS_PREFIX` config and see if your js assets build with the correct filepath. I believe this is the way that they are currently being built. For reference: https://github.com/apache/superset/pull/15105

eschutho (changes requested):
let's try out the existing`ASSET_BASE_URL` instead. I think this will be a breaking change.

rusackas:
This looks like it's in need of a rebase, and sadly the conversation got stalled. Is there still interest in getting this mergeable/merged?

huandu:
> Is there still interest in getting this mergeable/merged?

@rusackas I'm OK to close it since it doesn't seem to be an issue concerned by many users. For me, I patched the code in docker image and worked fine for month in my production environment.

adamfarmer0:
This error still exists in v5.0.0. I had to patch the code with this part:
```def get_manifest(self) -> dict[str, Union[Callable[[str], list[str]], str]]:
        loaded_chunks = set()
        assets_prefix = self.app.config["STATIC_ASSETS_PREFIX"] if self.app else ""

        def get_files(bundle: str, asset_type: str = "js") -> list[str]:
            files = self.get_manifest_files(bundle, asset_type)
            filtered_files = [f for f in files if f not in loaded_chunks]
            for f in filtered_files:
                loaded_chunks.add(f)
            if assets_prefix:
                return [f"{assets_prefix}{f}" for f in filtered_files]
            return filtered_files

        return {
            "js_manifest": lambda bundle: get_files(bundle, "js"),
            "css_manifest": lambda bundle: get_files(bundle, "css"),
            "assets_prefix": assets_prefix,
        }
```
to make it work.

PR #29969, opened 2024-08-19
chore(antd): straightening out button import paths

rusackas:
### SUMMARY

Just straightening out some wiring. Now the Superset "Button" imports directly from AntD, and the other Button instances in the code import from the Superset component instead of the AntD base component.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/Header/index.jsx
+++ b/superset-frontend/src/dashboard/components/Header/index.jsx
@@ -35,8 +35,7 @@ import {
   LOG_ACTIONS_TOGGLE_EDIT_DASHBOARD,
 } from 'src/logger/LogUtils';
 import Icons from 'src/components/Icons';
-import Button from 'src/components/Button';
-import { AntdButton } from 'src/components/';
+import { Button } from 'src/components/';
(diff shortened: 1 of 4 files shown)


superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -138,7 +137,8 @@ const actionButtonsStyle = theme => css`
   }
 `;
 
-const StyledUndoRedoButton = styled(AntdButton)`
+const StyledUndoRedoButton = styled(Button)`
+  // TODO: check if we need this.
```
mistercrunch, line 141: I think it's used here ->
Screenshot 2024-08-19 at 4 20 58 PM

mistercrunch (approved):

superset-frontend/src/dashboard/components/Header/index.jsx
rusackas: oh, I meant the CSS below it ;) It might not need those overrides... but I'll circle back to that.

PR #29970, opened 2024-08-20
docs: improve pre-commit docs and discoverability when CI fails

mistercrunch:
Many first-time contributors don't set up pre-commit for some reason (they're probably using a docker-compose backed dev env setup), so it's a common point of confusion when CI fails for linting or minor glitches. This PR:
- adds a clear message about what is failing and why
- give a pointer to pre-commit
- points to the [now improved] docs for more details

Fancy emojis and all -> 
Screenshot 2024-08-20 at 9 13 06 AM

### DIFF
--- a/docs/docs/contributing/development.mdx
+++ b/docs/docs/contributing/development.mdx
@@ -92,7 +92,50 @@ To install run the following:
 pre-commit install
 ```
 
-A series of checks will now run when you make a git commit.
+This will install the hooks in your local repository. From now on, a series of checks will
+automatically run whenever you make a Git commit.
+
+#### Running Pre-commit Manually
+
+You can also run the pre-commit checks manually in various ways:
+
+- **Run pre-commit on all files (same as CI):**
+
+  To run the pre-commit checks across all files in your repository, use the following command:
+
+  ```bash
+  pre-commit run --all-files
+  ```
+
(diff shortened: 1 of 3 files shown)


docs/docs/contributing/development.mdx
```diff
@@ -92,7 +92,50 @@ To install run the following:
...
+
+  Or for a specific file:
+
+  ```bash
+  pre-commit run <hook_id> --files path/to/your/file.py
+  ```
+
+  Replace `<hook_id>` with the ID of the specific hook you want to run. You can find the list
+  of available hooks in the `.pre-commit-config.yaml` file.
+
```
rusackas, line 138: This is already covered here: https://superset.apache.org/docs/contributing/development/#git-hooks-1

What you have here seems to be more in-depth, however, so maybe just add a link from that other page to this one? It would just be better to maintain this info in one place.

mistercrunch: Oh this is the actual section of the docs you are pointing to :) . I'm just adding on to it. Got GPT to do it and it went a bit verbose...

rusackas: LOL... I should have realized that. 🤦

rusackas (approved):

PR #29974, opened 2024-08-20
feat(sqllab): Adds refresh button to table metadata in SQL Lab

Usiel:
### SUMMARY

When table metadata changes (e.g. due to queries launched in Superset or other external changes) the user has no convenient way to force a refresh of the data for their `TabState`. The metadata is persisted even across sessions and will never update if the user stays on the current tab.

This feature is a first step to make this experience better by adding a refresh button for each table that allows a manual refresh.

**Future improvements:**
- Automatically refresh when query is _likely_ to change the metadata
- Refresh metadata from time to time
- Do not persist the metadata for a SQL Lab tab forever: Maybe remove `TableSchema` and instead use a shared cache with well defined expiry

Refresh Button in Action

### TESTING INSTRUCTIONS

1. Load SQL Lab and expand a table
2. Change the table (e.g. add column)
3. Click refresh button
4. **Expected**: New column appears
5. Refresh the whole page
6. **Expected**: Column is still there (asserts the server-side `TableSchema` was properly updated)

### ADDITIONAL INFORMATION

- [ ] Has associated issue
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/TableElement/index.tsx
+++ b/superset-frontend/src/SqlLab/components/TableElement/index.tsx
@@ -32,6 +32,7 @@ import {
   syncTable,
 } from 'src/SqlLab/actions/sqlLab';
 import {
+  tableApiUtil,
   useTableExtendedMetadataQuery,
   useTableMetadataQuery,
 } from 'src/hooks/apiResources';
@@ -107,7 +108,7 @@ const TableElement = ({ table, ...props }: TableElementProps) => {
   const {
     currentData: tableMetadata,
     isSuccess: isMetadataSuccess,
-    isLoading: isMetadataLoading,
+    isFetching: isMetadataLoading,
     isError: hasMetadataError,
   } = useTableMetadataQuery(
     {
@@ -175,6 +176,13 @@ const TableElement = ({ table, ...props }: TableElementProps) => {
 
   const toggleSortColumns = () => {
     setSortColumns(prevState => !prevState);
+  };
+
+  const refreshTableMetadata = () => {
+    dispatch(
+      tableApiUtil.invalidateTags([{ type: 'TableMetadatas', id: name }]),
+    );
+    dispatch(syncTable(table, tableData));
   };
 
   const renderWell = () => {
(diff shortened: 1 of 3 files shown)


Usiel:
Rebased on `master` (thought I branched off the latest but apparently that was not the case)

superset-frontend/src/hooks/apiResources/tables.ts
```diff
@@ -117,6 +117,13 @@ const tableApi = api.injectEndpoints({
       }),
     }),
     tableMetadata: builder.query<TableMetaData, FetchTableMetadataQueryParams>({
+      providesTags: result =>
+        result
+          ? [
+              { type: 'TableMetadatas', id: result.name },
```
Usiel, line 123: Having a guaranteed unique id would be better. From what I can tell we would need to return it from the API though, so I just went with the name for now.

mistercrunch: what is `result.name` in this context? Is it a fully-qualified table name? Regardless I think at this time changing anything in the table selector flushes the metadata section, so even just a table name should be safe for now. If we were to support tables from multiple schemas to remain accessible in this metadata section, this could in theory create some collision.

Usiel: It's just the table name, no extra qualifier like schema. And I see it the same way regarding collisions, so I'm good leaving it as is it.

superset-frontend/src/SqlLab/components/TableElement/index.tsx
```diff
@@ -107,7 +108,7 @@ const TableElement = ({ table, ...props }: TableElementProps) => {
   const {
     currentData: tableMetadata,
     isSuccess: isMetadataSuccess,
-    isLoading: isMetadataLoading,
+    isFetching: isMetadataLoading,
```
mistercrunch, line 111: haven't dug deep, but I'd assume switching the name here we'd fetch some other reference to it somewhere else?

Usiel: We are destructuring `isFetching` (which is already part of the return of `useTableMetadataQuery`) into `isMetadataLoading`.
(`isLoading` is only true on the first load while `isFetching` is also true on subsequent fetches.)

I renamed the local variable `isMetadataLoading` to `isMetadataFetching` to be in sync with the naming convention.

mistercrunch (approved):

PR #29978, opened 2024-08-21
fix(sqllab): flaky json explore modal due to shallow equality checks for extra data

justinpark:
### SUMMARY
In the async query, the `extra` field of the pulling data is in object form, and when the result is completed, it fails to verify equality due to shallow equality checks.
This caused unnecessary over-rendering of the result panel, so this commit resolved the issue by comparing the extra field separately.

Before:

https://github.com/user-attachments/assets/cb329bf1-2b8f-4f3e-8ac6-10b846794d3d

After:

https://github.com/user-attachments/assets/c77006f2-2f09-496d-804a-07af50ee57a2

### TESTING INSTRUCTIONS
Run a large query in a new tab, then switch to another tab.
Run a quick query that includes a JSON value, then click the value to verify that the JSON modal is shown steadily.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/reducers/sqlLab.js
+++ b/superset-frontend/src/SqlLab/reducers/sqlLab.js
@@ -17,7 +17,7 @@
  * under the License.
  */
 import { normalizeTimestamp, QueryState, t } from '@superset-ui/core';
-import { omit } from 'lodash';
+import { isEqual, omit } from 'lodash';
 import { shallowEqual } from 'react-redux';
 import * as actions from '../actions/sqlLab';
 import { now } from '../../utils/dates';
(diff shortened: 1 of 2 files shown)


michael-s-molina (approved):

superset-frontend/src/SqlLab/reducers/sqlLab.js
```diff
@@ -696,7 +698,19 @@ export default function sqlLabReducer(state = {}, action) {
                 ? prevState
                 : currentState,
           };
-          change = true;
+          const newExtra = JSON.stringify(newQueries[id].extra);
```
michael-s-molina, line 701: You can use Lodash's [isEquals](https://lodash.com/docs/4.17.15#isEqual) here.

PR #29981, opened 2024-08-21
fix: trino thread app missing full context

dpgaspar:
### SUMMARY

Fixes the issue where Trino SQL execution was missing Flask's g values in the context.
Flask's requests and g are local to the request thread or greenlet, by creating a new thread we loose this context.

This PR passes a copy of g into the thread, g is then recreated inside the thread, this g will be local to the thread but will contain all the necessary context of the current flask request

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/trino.py
+++ b/superset/db_engine_specs/trino.py
@@ -27,7 +27,7 @@ from typing import Any, TYPE_CHECKING
 import numpy as np
 import pandas as pd
 import pyarrow as pa
-from flask import current_app, Flask, g
+from flask import ctx, current_app, Flask, g
 from sqlalchemy import text
 from sqlalchemy.engine.reflection import Inspector
 from sqlalchemy.engine.url import URL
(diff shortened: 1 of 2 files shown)


superset/db_engine_specs/trino.py
```diff
@@ -227,12 +227,22 @@ def execute_with_cursor(
...
         ) -> None:
             logger.debug("Query %d: Running query: %s", query_id, sql)
 
             try:
+                # Flask contexts are local to the thread that handles the request.
+                # When you spawn a new thread, it does not inherit the contexts
+                # from the parent thread,
+                # meaning the g object and other context-bound variables are not
+                # accessible
                 with app.app_context():
```
michael-s-molina, line 243: @dpgaspar I did a quick GPT search.

```
from flask import current_app
import threading

def my_thread():
    with current_app.app_context():
        # Access Flask context within the thread
        app = current_app._get_current_object()
        # Do something with the app

# Create and start the thread
thread = threading.Thread(target=my_thread)
thread.start()
```

Would something like this work for our use cases?

dpgaspar: seems basically similar, except I'm passing the current_app into the thread, also we need to recreate g inside the thread

michael-s-molina: Can't we use [copy_current_request_context](https://tedboy.github.io/flask/generated/flask.copy_current_request_context.html)?

```
@copy_current_request_context
def worker():
    # Access Flask's g object here
    # Your code goes here
```

dpgaspar: only the request context is copied according to: https://github.com/pallets/flask/issues/3306

michael-s-molina (approved):

dpgaspar:
Thank you for the review @michael-s-molina

Vitor-Avila (approved):

PR #30000, opened 2024-08-23
chore: use playwright to install browsers

mistercrunch:
Seems we use playwright to install chromium in some places, but not consistently for chromium and firefox. This makes for confusing/fragile stuff in the Dockerfile. Here I'm just using playwright consistently and in
the docs. Note that this bloats the dev/ci images with BOTH browsers, but people should really just build on top of lean and add their browsers and database drivers from there. We may want to have a single primary browse
r we support and just keep that one in our Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,10 +34,6 @@ ARG BUILD_TRANSLATIONS="false"
 # in dev we mount the repo and build the frontend inside docker
 ARG DEV_MODE="false"
 
-# Include headless browsers? Allows for alerts, reports & thumbnails, but bloats the images
-ARG INCLUDE_CHROMIUM="true"
-ARG INCLUDE_FIREFOX="false"
-
 # Somehow we need python3 + build-essential on this side of the house to install node-gyp
 RUN apt-get update -qq \
     && apt-get install \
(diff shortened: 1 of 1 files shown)


rusackas (approved):
OMG, I love the simplification, and look forward to NOT updating these versions or dealing with GH Issues about them.

Dockerfile
```diff
@@ -140,25 +140,15 @@ RUN apt-get update -qq \
         libxtst6 \
         git \
         pkg-config \
+        wget \
+        bzip2 \
         && rm -rf /var/lib/apt/lists/*
 
+# Instlalling headless browsers
```
villebro, line 147:
```suggestion
# Installing headless browsers
```

Dockerfile
```diff
@@ -140,25 +140,15 @@ RUN apt-get update -qq \
...
-RUN apt-get update -qq \
-    && apt-get install -yqq --no-install-recommends wget bzip2 \
-    && wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz -O - | tar xfz - -C /usr/local/bin \
-    # Install Firefox
-    && wget -q https://download-installer.cdn.mozilla.net/pub/firefox/releases/${FIREFOX_VERSION}/linux-x86_64/en-US/firefox-${FIREFOX_VERSION}.tar.bz2 -O - | tar xfj - -C /opt \
-    && ln -s /opt/firefox/firefox /usr/local/bin/firefox \
-    && apt-get autoremove -yqq --purge wget bzip2 && rm -rf /var/[log,tmp]/* /tmp/* /var/lib/apt/lists/*
-# Cache everything for dev purposes...
+RUN playwright install chromium --with-deps
+RUN playwright install firefox --with-deps
```
villebro, line 151: I'm curious, does with work with multiarch builds? I'm assuming the chromium and firefox binaries are different on `x86_64` and `arm64`. It seems previously we were only installing the `x86_64` versions.

villebro (commented):
LGTM, but I haven't been able to test this, so deferring approval to someone who can confirm that part.

mistercrunch:
I meant to write that I haven't tested this and was looking to comment ask for help or instructions as to how to test this.

Dockerfile
kgabryje: @mistercrunch According to the [docs](https://playwright.dev/docs/browsers#managing-browser-binaries), playwright installs the browsers in some local directory, not in `usr/local/bin`. We can change that with `PLAYWRIGHT_BROWSERS_PATH` env variable

mistercrunch: @villebro I think `playwright` is smart and figures out what platform its own and downloads the right bins. 

@kgabryje is targeting `usr/local/bin` a requirement? seems on this build cypress e2e is working as expected (?) Not sure how to more directly test it than just trusting cypress.

kgabryje: I think cypress installs its own chromium binary. A proper way to test it would be to run a report

PR #30001, opened 2024-08-23
fix: Reports are not sent when selecting to send as PNG, CSV or text

eschutho:
In AlertReportModal.tsx, it looks like extra cannot be null which is causing 500s when trying to send a report as PNG, CSV or text.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas (approved):

sadpandajoe:
breaking code isn't currently in 4.1 so removing the label

PR #30002, opened 2024-08-23
feat(time_comparison): Support all date formats when computing custom and inherit offsets

Antonio-RiveroMartnez:
### SUMMARY
The time comparison feature that's being used in Table and Big Number charts does not support Human Readable dates and it's very limited on what date formats it  supports. 

Now the API supports sending `inherit` or a `YYYY-MM-DD` date as offset value so we can rely on the existing API to parse any date before computing the custom or inherit offset.

Before
error

Now
test

### TESTING INSTRUCTIONS
1. Set up a Table Chart with a date filter with value: `Aug 2024` until `tomorrow``
2. Select inherit time shift
3. The chart must display the correct data and the comparison label

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/ComparisonRangeLabel.tsx
+++ b/superset-frontend/src/explore/components/controls/ComparisonRangeLabel.tsx
@@ -122,7 +122,25 @@ export const ComparisonRangeLabel = ({
             ensureIsArray(newShifts),
           );
         }
-        return Promise.resolve({ value: '' });
+        return fetchTimeRange(filter.comparator, filter.subject).then(res => {
+          const datePattern = /\d{4}-\d{2}-\d{2}/g;
+          const dates = res?.value?.match(datePattern);
+          const [startDate, endDate] = dates ?? [];
+          const postProcessedShifts = getTimeOffset({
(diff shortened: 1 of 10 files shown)


superset-frontend/src/explore/components/controls/ComparisonRangeLabel.tsx
```diff
@@ -108,20 +108,61 @@ export const ComparisonRangeLabel = ({
...
           );
         }
+        // Need to compute custom or inherit shifts first and then mix with the non custom or inherit shifts
+        if (
+          (ensureIsArray(customOrInheritShifts).includes('custom') &&
+            startDate) ||
+          ensureIsArray(customOrInheritShifts).includes('inherit')
+        ) {
+          return fetchTimeRange(filter.comparator, filter.subject).then(res => {
+            const datePattern = /\d{4}-\d{2}-\d{2}/g;
```
kgabryje, line 135: Can we move it to a const, since we’re using it in many files?

kgabryje (approved):
LGTM

PR #30016, opened 2024-08-26
feat: Displaying details to Dataset/Database deletion modals

rusackas:
### SUMMARY
This PR lists all dashboards and charts affected by the (potential) deletion of a dataset/database.

The links to these assets all open in a _single_ (named) new window.

It also does a little cleanup along the way:
* modal is now always centered (so it scrolls sensibly)
* the "delete" button is now the "Delete" button
* Highlights the name of the dataset/database that's getting deleted... this seems important.
* Updates the .pot file since some strings are getting tweaked in there.

Before (Dataset deletion):
image

After (Dataset deletion):
image

Before (Database deletion)
Screenshot 2024-08-26 at 10 11 52 AM

After (Database deletion):
image
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina (changes requested):
How is this going to work when there are thousands of dashboards and charts linked to a database?

rusackas:
> How is this going to work when there are thousands of dashboards and charts linked to a database?

It'll scroll. 

I'm open to other ideas of course, which might include:
* Limited height scroll areas for each asset type
* Showing the first X instances and then saying _"...and Y more"_

This is why I pinged @kasiazjc for review. Even if it's annoying to _see__ thousands of dashboards, it's far less dangerous than _breaking_ those thousands of dashboards.

I think I prefer option 2, just showing 10 or so as a sampling with "and X others" but I'm flexible.

michael-s-molina:
Some considerations if we actually want to list all the names:
- We would probably need a search given that it will be hard to find any chart among thousands
- Users should not scroll to fill the input or click on Cancel or Delete
- What about other assets types such as Alerts, Saved Queries that will also be impacted?

rusackas:
> Some considerations if we actually want to list all the names:
> 
> * We would probably need a search given that it will be hard to find any chart among thousands

I'm merely displaying the list that's already data the component has access to. I don't particularly want to go crazy with functionality here. If we need to add full-text fuzzy searching, or a new API for this, I'm open to discussing an additional PR, but I'd say it's out of scope for the moment.

> * Users should not scroll to fill the input or click on Cancel or Delete

This _is_ possible for the button, as you can lock the modal's footer in place and scroll the contents. However, that footer would be disabled until the user scrolls down anyway to fill the input. We might be able to put the input IN the footer, but that's a weird new pattern to introduce.

> * What about other assets types such as Alerts, Saved Queries that will also be impacted?

Alerts are not in the scope of the current data AFAIK. I could list the Saved Queries on the database modal if interested, but that seemed of less utility (including the links to them). 

Is the scrolling the real problem here? I'll just limit the results to fit on a sensible screen. I was just looking for a quick improvement over a modal where you a user has _no idea_ what they're breaking. 

Another possibility was to just use tooltips where the numbers are, but with a long list, that'd be even more painful.

superset-frontend/src/pages/DatabaseList/DatabaseList.test.jsx
```diff
@@ -164,14 +164,7 @@ describe('Admin DatabaseList', () => {
...
 
-    expect(wrapper.find(DeleteModal).props().description)
-      .toMatchInlineSnapshot(`
-      <React.Fragment>
-        <p>
-          The database db 0 is linked to 0 charts that appear on 0 dashboards and users have 0 SQL Lab tabs using this database open. Are you sure you want to continue? Deleting the database will break those objects.
-        </p>
-      </React.Fragment>
-    `);
+    expect(wrapper.find(DeleteModal).props().description).toMatchSnapshot();
```
rusackas, line 167: InlineSnapshots are no longer supported by Jest. A few changed files in this PR allow us to use regular snapshots, which remain supported.

superset-frontend/src/components/DeleteModal/index.tsx
```diff
@@ -80,10 +80,11 @@ export default function DeleteModal({
       disablePrimaryButton={disableChange}
       onHide={hide}
       onHandledPrimaryAction={confirm}
-      primaryButtonName={t('delete')}
+      primaryButtonName={t('Delete')}
```
rusackas, line 83: Bycatch, another mis-capitalized button. Maybe @michael-s-molina has some regex-y grep magic to see if there are more of these :D

superset-frontend/src/pages/DatabaseList/DatabaseList.test.jsx
rusackas: Note that I also loosened constraints (prettier, license check) on those snapshot files, so they can be committed as-generated.

michael-s-molina:
> Also open to abusing some other components (Popovers with scrolling lists? Tabs? ¯\_(ツ)_/¯ ) but again was just looking to make a simple (hah!) safety improvement.

I understand your objective but simple might mean different things depending on the environment. I'm just trying to reach a simple solution that works for organizations with a lot of assets.

> I would also argue that if you don't even have time to scroll past the list of things you're about to break, you probably should not be allowed to click that button.
> Is the scrolling the real problem here? I'll just limit the results to fit on a sensible screen. I was just looking for a quick improvement over a modal where you a user has no idea what they're breaking.

I don't think the problem is just the scroll. Think about a scenario where a database has 40k charts. The admin will not review every single chart. They might review a sample of charts, maybe the "popular" ones to understand the impact of the change. This might be one way to resolve this, show the top 10 assets of each category. We would need to determine how to calculate the top. 

We also need to consider the payload size when returning the names for all results of each asset type.

> I'm merely displaying the list that's already data the component has access to.

This is really concerning because it means that this query returns thousands of names when we currently only display a count. If we go for the top 10 solution, this query should also be modified to just return the necessary data.

> Alerts are not in the scope of the current data AFAIK.

Alerts have a database connection associated to them. When you delete a database, you'll delete the alerts too because of cascading.

> I could list the Saved Queries on the database modal if interested, but that seemed of less utility (including the links to them).

Saved queries are really important at Airbnb as they might be really complex and take a long time to build. Rebuilding a dashboard or chart might be way faster than rebuilding a saved query.

In summary, I think the top 10 approach might work for giving admins an idea of what's being affected by a database deletion. If you really want to show all values, we could introduce a configuration to limit the max number of assets displayed and show a message indicating the limit.

mistercrunch (dismissed):
LGTM

mistercrunch:
Seems we can decouple the issues:
- primary purpose: showing the user what's about to cascade-delete and the fact it's long (`show more...` ?)
- no loading indicator: maybe can be addressed in this PR?
- slow query - I'd love to decouple this issue

About slow query, I might be able to help improve the query, I'd just need to figure out the backend call behind it, review the query and make sure it uses an index

rusackas:
As discussed with @michael-s-molina, I made one of several steps so that we might get this PR merged, and follow up with more.

The results of the displayed at-risk items are now limited to 10. This gives more context than we had before, without causing massive scrolling problems.
image

Next steps may include:
• Displaying/counting more at-risk assets (saved queries, alerts/reports, etc)
• Making the number if displayed items configurable
• Adding "view more" link to load/paginate more items
• Adding search/filters for the list
• Adjusting the API to _also_ limit to 10 items (or as configured) and thus reduce a pre-existing performance bottleneck (CC @mistercrunch who offered to chip in here)
• Add a spinner in case there _is_ a bottleneck.

... but for now, users merely have a bit more context on what's at risk, which was my original (simplistic) hope, even if the rabbit hole goes _a lot_ deeper from here.

superset-frontend/src/pages/DatabaseList/index.tsx
```diff
@@ -609,20 +608,160 @@ function DatabaseList({
...
+                  </ul>
+                </>
+              )}
               {DatabaseDeleteRelatedExtension && (
                 <DatabaseDeleteRelatedExtension
                   database={databaseCurrentlyDeleting}
                 />
               )}
             </>
+            /*
```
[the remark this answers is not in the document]
michael-s-molina: @rusackas I think you missed this comment 👆🏼

michael-s-molina (approved):

superset-frontend/src/pages/DatabaseList/index.tsx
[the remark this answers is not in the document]
rusackas: facepalm-duh

Fixed :)

PR #30017, opened 2024-08-26
fix: pass if table is already removed on upgrade

sadpandajoe:
### SUMMARY

Looks like if you try to run the migrations but the sl_ tables don't already exist the migrations will fail. This will just pass the upgrade check if the table doesn't already exist.

* Added a `has_table` function to return a boolean if the table exist or not
* Updated `drop_fks_for_table` to handle sqlite in the function

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/shared/utils.py
+++ b/superset/migrations/shared/utils.py
@@ -168,3 +168,18 @@ def try_load_json(data: Optional[str]) -> dict[str, Any]:
     except json.JSONDecodeError:
         print(f"Failed to parse: {data}")
         return {}
+
+
+def has_table(table_name: str) -> bool:
+    """
+    Check if a table exists in the database.
+
+    :param table_name: The table name
+    :returns: True if the table exists
+    """
+
+    insp = inspect(op.get_context().bind)
+    table_exists = insp.has_table(table_name)
+
+    print(f"Table {table_name} exist: {table_exists}")
+    return table_exists
--- a/superset/migrations/shared/utils.py
+++ b/superset/migrations/shared/utils.py
@@ -181,5 +181,4 @@ def has_table(table_name: str) -> bool:
     insp = inspect(op.get_context().bind)
     table_exists = insp.has_table(table_name)
 
-    print(f"Table {table_name} exist: {table_exists}")
     return table_exists
--- a/superset/migrations/versions/2024-08-13_15-17_39549add7bfc_remove_sl_table_columns_table.py
+++ b/superset/migrations/versions/2024-08-13_15-17_39549add7bfc_remove_sl_table_columns_table.py
@@ -34,9 +34,13 @@ down_revision = "02f4f7811799"
 
 def upgrade():
     connection = op.get_bind()
-    if connection.dialect.name != "sqlite":
-        drop_fks_for_table("sl_table_columns")
-    op.drop_table("sl_table_columns")
+    try:
+        if connection.dialect.name != "sqlite":
+            drop_fks_for_table("sl_table_columns")
+        op.drop_table("sl_table_columns")
+    except sa.exc.NoSuchTableError:
+        # Table doesn't exist, so we can safely ignore this error
+        pass
 
 
 def downgrade():
--- a/superset/migrations/versions/2024-08-13_15-17_39549add7bfc_remove_sl_table_columns_table.py
+++ b/superset/migrations/versions/2024-08-13_15-17_39549add7bfc_remove_sl_table_columns_table.py
@@ -32,10 +32,10 @@ from superset.migrations.shared.utils import has_table
 revision = "39549add7bfc"
 down_revision = "02f4f7811799"
 
+table_name = "sl_table_columns"
+
 
 def upgrade():
-    table_name = "sl_table_columns"
-
     if has_table(table_name):
         drop_fks_for_table(table_name)
         op.drop_table(table_name)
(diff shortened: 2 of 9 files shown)


superset/migrations/versions/2024-08-13_15-27_e53fd48cc078_remove_sl_dataset_users.py
```diff
@@ -34,9 +34,14 @@
...
     connection = op.get_bind()
-    if connection.dialect.name != "sqlite":
-        drop_fks_for_table("sl_dataset_users")
-    op.drop_table("sl_dataset_users")
+
+    try:
+        if connection.dialect.name != "sqlite":
+            drop_fks_for_table("sl_dataset_users")
+        op.drop_table("sl_dataset_users")
+    except sa.exc.NoSuchTableError:
```
michael-s-molina, line 42: @sadpandajoe Would you mind adding a method called `has_table` to https://github.com/apache/superset/blob/master/superset/migrations/shared/utils.py to be reused in these and future migrations?

```
def has_table(table_name: str) -> bool:
    """
    Check if a table exists in the database.

    :param table_name: The table name
    :returns: True iff the table exists
    """

    insp = inspect(op.get_context().bind)
    return insp.has_table(table_name)
```

mistercrunch: yes, and/or change `drop_fks_for_table` to receive a `fail_when_missing=False`

superset/migrations/versions/2024-08-13_15-29_a6b32d2d07b1_remove_sl_columns.py
```diff
@@ -34,9 +34,14 @@
 
 def upgrade():
     connection = op.get_bind()
-    if connection.dialect.name != "sqlite":
-        drop_fks_for_table("sl_columns")
-    op.drop_table("sl_columns")
+
+    try:
+        if connection.dialect.name != "sqlite":
```
mistercrunch, line 39: also seems like the special handling for sqlite could be handled within `drop_fks_for_table`. More generally would be great to have migrations that are simpler / self-healing / free-of-engine-specific-logic through better utils.

superset/migrations/versions/2024-08-13_15-27_e53fd48cc078_remove_sl_dataset_users.py
sadpandajoe: Even if we updated `drop_fks_for_table` we'd have to write something to drop table too, else it'll fail won't it?

mistercrunch: Whatever makes sense, maybe `drop_table(table_name, including_fks=[], only_if_it_exists=True)` or maybe it's for another PR, but a common issue is a migration that fails at step 2 and when you rerun it it fails at step one since step one can only run once (say a create table). Personally think we could use more of the "drop-table-but-only-if-it-exists" and "create-table-but-drop-it-first-if-it-exists" type utils to avoid getting people tangled up. We've had many people open issue with the message of "step 1 one of the migration fails, here's the stacktrace", only to realize that the issue was with a subsequent step.

superset/migrations/shared/utils.py
```diff
@@ -168,3 +168,18 @@ def try_load_json(data: Optional[str]) -> dict[str, Any]:
...
+    Check if a table exists in the database.
+
+    :param table_name: The table name
+    :returns: True if the table exists
+    """
+
+    insp = inspect(op.get_context().bind)
+    table_exists = insp.has_table(table_name)
+
+    print(f"Table {table_name} exist: {table_exists}")
```
michael-s-molina, line 184: Maybe delete this print? If not, then:
```suggestion
    print(f"Table {table_name} exists: {table_exists}")
```

superset/migrations/versions/2024-08-13_15-17_39549add7bfc_remove_sl_table_columns_table.py
```diff
@@ -26,17 +26,19 @@
...
 revision = "39549add7bfc"
 down_revision = "02f4f7811799"
 
 
 def upgrade():
-    connection = op.get_bind()
-    if connection.dialect.name != "sqlite":
-        drop_fks_for_table("sl_table_columns")
-    op.drop_table("sl_table_columns")
+    table_name = "sl_table_columns"
```
michael-s-molina, line 37: Can you mode this outside the `upgrade` method (below `down_revision` variable) and reuse it on `downgrade`?

superset/migrations/versions/2024-08-13_15-29_a6b32d2d07b1_remove_sl_columns.py
```diff
@@ -26,17 +26,19 @@
...
 revision = "a6b32d2d07b1"
 down_revision = "e53fd48cc078"
 
 
 def upgrade():
-    connection = op.get_bind()
-    if connection.dialect.name != "sqlite":
-        drop_fks_for_table("sl_columns")
-    op.drop_table("sl_columns")
+    table_name = "sl_columns"
```
michael-s-molina, line 37: Same here.

superset/migrations/versions/2024-08-13_15-17_39549add7bfc_remove_sl_table_columns_table.py
michael-s-molina: Please do this for all changed migrations.

sadpandajoe: done

michael-s-molina (approved):

sadpandajoe:
Not a breaking change since this handles the case where if migration has already been run gracefully.

PR #30021, opened 2024-08-27
chore: accelerate docker compose by skipping frontend build

mistercrunch:
while waiting for a `docker-compose build`, I decided to go and skip the long steps around the frontend buildsthat are not required while in dev mode while using  `docker-compose` where we mount the local repo and build it inside docker.

Note that it's always tricky to have conditional logic in docker files as things aren't procedural. I took the approach of using and `ARG` and having conditions in `RUN` commands which accomplishes the goal of skipping the frontend builds in DEV_MODE=true. 

changes:
    - skip `npm ci` when DEV_MODE=true
    - skip `npm run build` when DEV_MODE=true
    - introduce an ARG for INCLUDE_FIREFOX, skip by default
    - introduce an ARG for INCLUDE_CHROMIUM, include in `dev` layer by default - added a note in `UPDATING.md` even though people shouldn't really use the `dev` target from our Dockerfile.

### DIFF
--- a/RESOURCES/INTHEWILD.md
+++ b/RESOURCES/INTHEWILD.md
@@ -112,7 +112,7 @@ Join our growing community!
 - [Showmax](https://showmax.com) [@bobek]
 - [TechAudit](https://www.techaudit.info) [@ETselikov]
 - [Tenable](https://www.tenable.com) [@dflionis]
-- [Tentacle](https://tentaclecmi.com) [@jdclarke5]
+- [Tentacle](https://www.linkedin.com/company/tentacle-cmi/) [@jdclarke5]
 - [timbr.ai](https://timbr.ai/) [@semantiDan]
 - [Tobii](http://www.tobii.com/) [@dwa]
 - [Tooploox](https://www.tooploox.com/) [@jakubczaplicki]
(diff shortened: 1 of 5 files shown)


rusackas:
Can we add some docs to the installing section that explains how/why one would change modes, and what does/doesn't happen in each mode? This is a fantastic optimization, but might lead to Github Issues if it's non-obvious.

One of these days, I'm gonna try switching us to Yarn. It'd be faster, and likely more stable, but would undoubtedly cause a lot of outcry/confusion :P

mistercrunch:
@rusackas added some docs and rebased

RESOURCES/INTHEWILD.md
```diff
@@ -112,7 +112,7 @@ Join our growing community!
 - [Showmax](https://showmax.com) [@bobek]
 - [TechAudit](https://www.techaudit.info) [@ETselikov]
 - [Tenable](https://www.tenable.com) [@dflionis]
-- [Tentacle](https://tentaclecmi.com) [@jdclarke5]
+- [Tentacle](https://www.linkedin.com/company/tentacle-cmi/) [@jdclarke5]
```
rusackas, line 115: Just FYI, the link checker is non-blocking. I'm wondering if these guys went out of business or something. The site's been down for about a week thus far, and I didn't get a response from them on Slack.

rusackas (approved):
Love it! Thanks for the docs!

PR #30023, opened 2024-08-27
fix(plugins): display correct tooltip (fixes #3342)

### LINKED ISSUE #3342: Chord popups contain wrong information
### Superset version
Superset 0.19.0

### Expected results
When clicked on Chord diagram bean, it shows the correct flow from source to target

### Actual results
It shows just opposite numbers - from target to source

### Steps to reproduce

1) create a chord diagram, set source and target
2) click on diagram and notice that source and target in popup are switched places

I attach the source data and chord screen to show
table1
hord1

jonaschn:
### SUMMARY
Fixes #3342
Display the correct tooltip information for chord charts (source and target values were switched)

### BEFORE
image

### AFTER
image

### TESTING INSTRUCTIONS
Example SQL Dataset:
```sql
SELECT 'Alice' AS source, 'Bob' AS target, 5 AS interactions UNION ALL
SELECT 'Alice', 'Charlie', 3 UNION ALL
SELECT 'Bob', 'Charlie', 2 UNION ALL
SELECT 'Charlie', 'David', 1  UNION ALL
SELECT 'Bob', 'David', 4
```

### ADDITIONAL INFORMATION

- [x] Has associated issue: #3342
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/legacy-plugin-chart-chord/src/Chord.js
+++ b/superset-frontend/plugins/legacy-plugin-chart-chord/src/Chord.js
@@ -130,9 +130,9 @@ function Chord(element, props) {
     .text(
       d =>
         `${nodes[d.source.index]} → ${nodes[d.target.index]}: ${f(
-          d.source.value,
-        )}\n${nodes[d.target.index]} → ${nodes[d.source.index]}: ${f(
           d.target.value,
+        )}\n${nodes[d.target.index]} → ${nodes[d.source.index]}: ${f(
+          d.source.value,
         )}`,
     );
 }


superset-frontend/plugins/legacy-plugin-chart-chord/src/Chord.js
```diff
@@ -130,9 +130,9 @@ function Chord(element, props) {
     .text(
       d =>
         `${nodes[d.source.index]} → ${nodes[d.target.index]}: ${f(
```
jonaschn, line 132: The diff looks a bit misleading.
I simply swapped `d.source.value` and `d.target.value` (line 133 and 135)

rusackas:
Agreed that it was wrong and that this fixes _the display_ of the chart, but I'm _so_ confused by the meaning of the diff. I mean, yes - swapping it swaps it, and that works. But now the code (modified by the diff) is
`${nodes[d.target.index]} → ${nodes[d.source.index]}` which displays the `source` column then `→` then the `target` column. How is this now _displaying_ correctly but with the items _named_ incorrectly? 🤯

rusackas (approved):

jonaschn:
Thanks for merging. When do you expect to release this change?
Will it be part of 4.1.0?
I would really appreciate it because our users were constantly confused.

PR #30025, opened 2024-08-27
fix: Dashboard list row height does not match other lists

michael-s-molina:
### SUMMARY
Fixes the dashboard list row height to match the height of other lists. This problem was introduced by the [Slack Avatar integration](https://github.com/apache/superset/pull/27849).

As we can see the row height is not the same between the dashboard and charts list which makes the avatars close to the row separator in the dashboard list.

Screenshot 2024-08-27 at 11 20 21
Screenshot 2024-08-27 at 11 20 35

<br>After the fix, the dashboard list matches the layout of charts/datasets lists and there's a space between the avatars and row separators.

Screenshot 2024-08-27 at 11 21 47

### TESTING INSTRUCTIONS
Check that the layout match between lists.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/pages/DashboardList/index.tsx
+++ b/superset-frontend/src/pages/DashboardList/index.tsx
@@ -383,9 +383,6 @@ function DashboardList(props: DashboardListProps) {
         Header: t('Owners'),
         accessor: 'owners',
         disableSortBy: true,
-        cellProps: {
-          style: { padding: '0px' },
-        },
         size: 'xl',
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):
The less custom CSS the better - we should probably enforce a [soft] rule to keep css only in the component library / reusable components

PR #30029, opened 2024-08-27
fix(migrations): Fix the time comparison migration

Antonio-RiveroMartnez:
### SUMMARY
After the initial version of the time comparison feature for Table and Big Number  we introduced a migration to upgrade old charts to the new comparison controls, however, during that upgrade a table that never had comparison in it was wrongly showing Time Comparison Data.

This PR is fixing the migration so we can properly upgrade such old tables if needed, also we extend our tests to not only consider charts that have the enabled_time_comparison flag set to True.

### BEFORE
Old version of Chart using Time Comparison once but then removing it from the controls:
initial
removed

The old Chart is upgraded with the previous version of the migration:
upgrade

The chart start showing comparison data:
wrong

### AFTER
Old version of Chart using Time Comparison once but then removing it from the controls:
v2_initial
v2_no_comp

The old Chart is upgraded with the new version of the migration:
v2_upgrade

The chart doesn't show comparison data:
correct

### TESTING INSTRUCTIONS
1. Create a table chart with a version of the time comparison where Enable Time Comparison checkbox was used
2. Run a comparison data (Mark the checkbox and select a time range)
3. Unmark the checkbox and run the chart data again so your chart has no longer comparison info
4. Use more updated version and run the migrations so your table is upgraded
5. You must not see time comparison data in the table.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/migrations/f84fde59123a_update_charts_with_old_time_comparison__test.py
+++ b/tests/integration_tests/migrations/f84fde59123a_update_charts_with_old_time_comparison__test.py
@@ -83,7 +83,7 @@ base_params: dict[str, Any] = {
 }
 
 # Specific parameter objects overriding only the differing properties
-params_v1_with_custom = {
+params_v1_with_custom: dict[str, Any] = {
     **base_params,
     "metric": {
         **base_params["metric"],
@@ -113,7 +113,7 @@ params_v1_with_custom = {
     ],
 }
 
-params_v1_other_than_custom = {
+params_v1_other_than_custom: dict[str, Any] = {
     **base_params,
     "metric": {
         **base_params["metric"],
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Thanks for fixing this!

tests/integration_tests/migrations/f84fde59123a_update_charts_with_old_time_comparison__test.py
```diff
@@ -161,6 +161,70 @@
         }
     ],
 }
+params_v1_other_than_custom_false: dict[str, Any] = {
```
michael-s-molina, line 164: Would you mind extracting the common parts between `params_v1_with_custom`, `params_v1_with_custom`, `params_v1_other_than_custom_false`, `params_v2_with_custom`, `params_v2_other_than_custom`, `params_v2_other_than_custom_false` and just override the necessary properties for each variable to avoid code duplication and increase readability?

michael-s-molina (approved):
Thanks for addressing the comments @Antonio-RiveroMartnez!

PR #30062, opened 2024-08-29
fix(time_range): None time range should equal No filter

villebro:
### SUMMARY
While working on a Jinja time filtering feature, I noticed a curious quirk in how `get_since_until_from_time_range` works:
- When `time_range="No filter"` it returns `(None, None)` for `since` and `until` respectively
- For `time_range=None` it returns `(None, datetime(current_date))`

I don't see any reason for this, and can only assume it's a quirk of how the complex function works for other more intricate time ranges. Therefore, I feel it prudent to change the behavior, so that `None` and `"No filter"` both returns the same value.

Let's see if this change breaks some tests - if not, I can only assume this isn't intended, but rather an artifact of the current logic.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/date_parser.py
+++ b/superset/utils/date_parser.py
@@ -180,7 +180,7 @@ def get_since_until(  # pylint: disable=too-many-arguments,too-many-locals,too-m
     _relative_start = relative_start if relative_start else "today"
     _relative_end = relative_end if relative_end else "today"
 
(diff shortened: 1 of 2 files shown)


villebro:
Closing, as it's probably best to let sleeping dogs lie 🐶

Issue #25048, opened 2023-08-22
[SIP-99] Proposal for correctly handling business logic


john-bodley:
## [SIP-99] Proposal for correctly handling business logic

Whilst spelunking through the code as part of [[SIP-92] Proposal for restructuring the Python code base](https://github.com/apache/superset/issues/20630) it became apparent* that our grasp of SQLAlchemy sessions, atomic units of work, model/business validation, etc. was likely incorrect or misconstrued—resulting in overly complex or redundant Data Access Objects (DAOs), Commands, event handlers, tests, etc.

\* A benefit of refactoring the code into functional units highlighted the various discrepancies which exist.

This SIP serves as an umbrella for a series of lightweight backend SIPs related to defining best practices and guidelines for core/foundational constructs including:

- [[SIP-99A] Primer on managing SQLAlchemy sessions](https://github.com/apache/superset/issues/25107)
- [[SIP-99B] Proposal for (re)defining a “unit of work”](https://github.com/apache/superset/issues/25108)
- [[SIP-99C] Proposal for model and business validation](https://github.com/apache/superset/issues/25828)
- [[SIP-99D] Proposal for consolidating Command/DAO create, update, and upsert operations](https://github.com/apache/superset/issues/25830)

The hope is (if enacted) that these SIPs should:

- Simplify the code
- Simplify testing
- Simplify the mental model
- Improve the code quality
- Ensure the business layer represents a “unit of work”
- Ensure that the code adheres to the DRY principle
- Improve the developer experience
- Provide clear guidance on the subject and examples of accepted code standards

### Proposed Change

Please refer to the related SIPs for details.

### New or Changed Public Interfaces

Please refer to the related SIPs for details.

### New Dependencies

Please refer to the related SIPs for details.

### Migration Plan and Compatibility

Please refer to the related SIPs for details.

### Rejected Alternatives

Please refer to the related SIPs for details.

michael-s-molina:
Thank you for writing this set of SIPs @john-bodley! I look forward to the changes.

mistercrunch:
Overall the proposals seem very sound and virtuous. Curious on the assessment around how much work the refactor represents (seems fairly massive) and curious about a suggested approach in terms of rolling out the refactor (model-centric?) and whether to start with the most used ones, or the simplest ones (?)

john-bodley:
[...]

PR #30067, opened 2024-08-30
feat: allow certain exceptions to commit

betodealmeida:
### SUMMARY

[SIP-99](https://github.com/apache/superset/issues/25048) introduced in https://github.com/apache/superset/pull/24969 the amazing `@transaction` decorator, which transparently manages transactions in Superset in order to avoid eager commits.

[SIP-85](https://github.com/apache/superset/issues/20300) introduced a mechanism for authenticating to databases via OAuth2, using exceptions to signalize to the frontend from deep in the backend that the user needs to authenticate by following a link. These exceptions are benign, and carry in their `extra` payload (see [SIP-40](https://github.com/apache/superset/issues/9194)) the URL used for authentication, among others.

Because of this, when adding a database that has OAuth2 enabled we need to allow certain exceptions to still commit, while raising them. This allows the database to be created succesfully, while signalizing to the user that they still need to authenticate. This is needed because OAuth2 tokens are stored associated with a database ID, so it must exist _a priori_.

This PR changes the `@transaction` decorator to allow an optional group of exceptions that should trigger a commit before being re-raised. It's part of a bigger work to add support for OAuth2 to Snowflake, still in progress but already tested end-to-end.

N/A

### TESTING INSTRUCTIONS

Added unit tests for the `@transaction` decorator, covering the old use cases and the new one.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/decorators.py
+++ b/superset/utils/decorators.py
@@ -238,6 +238,7 @@ def on_error(
 
 def transaction(  # pylint: disable=redefined-outer-name
     on_error: Callable[..., Any] | None = on_error,
+    allowed: tuple[type[Exception], ...] = (),
 ) -> Callable[..., Any]:
     """
     Perform a "unit of work".
@@ -246,7 +247,12 @@ def transaction(  # pylint: disable=redefined-outer-name
     proved rather complicated, likely due to many architectural facets, and thus has
     been left for a follow up exercise.
 
+    In certain cases it might be desirable to commit even though an exception was
+    raised. For OAuth2, for example, we use exceptions as a way to signal the client to
+    redirect to the login page. In this case, we re-raise the exception and commit.
+
     :param on_error: Callback invoked when an exception is caught
+    :param allowed: Exception types to re-raise after committing
     :see: https://github.com/apache/superset/issues/25108
     """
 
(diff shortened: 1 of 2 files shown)


michael-s-molina:
@betodealmeida Not using exceptions to control flow is probably the top 10 best practice. Is it possible to change the code to signalize to the frontend that the user needs to authenticate using other means? I'm really worried about this change as it might generate even more cases where exceptions are used to control flow.

betodealmeida:
@betodealmeida while I recognize it's normally an anti-pattern, no rule is 100% correct ([an example](https://blog.jooq.org/rare-uses-of-a-controlflowexception/) where it can be useful).

For OAuth2 we're basically building a fancy 403 response that has information about how to proceed. We need to let the exception bubble up to the Flask error handler so that we can build a proper SIP-40 compliant error payload, which is what triggers OAuth2 in the frontend side. The benefit of this approach is that we can trigger OAuth2 without having to modify every API that might need it — instead all we need to do is attach a `ErrorMessageWithStackTrace` to the DOM.

PR #30071, opened 2024-08-30
feat: allow create/update OAuth2 DB

betodealmeida:
### SUMMARY

This PR is part of the work to add OAuth2 support for Snowflake. When creating the DB we provide only the OAuth2 client information (ID, secret, scope, auth/token URLs), which is not enough to test and validate the connection. We need to allow the DB to be created before the OAuth2 flow, so we can store the access tokens associated with a given DB.

<del>Depends on https://github.com/apache/superset/pull/30067.</del>

### TESTING INSTRUCTIONS

Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rtexelm (approved):
LGTM

PR #30072, opened 2024-08-30
chore(tests): skip extremely flaky gaq test

villebro:
### SUMMARY
The `test_chart_data_async` has been disabled, then re-enabled in the hopes that the new sleeps would fix it. This was not the case, and it's causing a lot of headache for getting CI green on PRs. I'm disabling it for now, and these will all likely be rewritten soon as part of SIP-134.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/charts/data/api_tests.py
+++ b/tests/integration_tests/charts/data/api_tests.py
@@ -718,6 +718,7 @@ class TestPostChartDataApi(BaseTestChartDataApi):
         if get_example_database().backend != "presto":
             assert "(\n      'boy' = 'boy'\n    )" in result
 
+    @unittest.skip("Extremely flaky test on MySQL")
     @with_feature_flags(GLOBAL_ASYNC_QUERIES=True)
     @pytest.mark.usefixtures("load_birth_names_dashboard_with_slices")
     def test_chart_data_async(self):


giftig (approved):
Adding sleeps aren't a good solution to getting a flaky test working so I agree let's just disable it until we have a real solution, it seems like it's causing more harm than good.

Given how much you're struggling with the cypress tests as well I'd disable that one while you're here too.

.github/workflows/superset-e2e.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        parallel_id: [0, 1, 2, 3, 4, 5]
+        parallel_id: [0, 1, 2, 0, 1, 2]
```
mistercrunch, line 42: I don't think you meant to get the changes in this file merged (?)

villebro: This was a quick test - I skimmed docs and realized parallelism needs to be adjusted in a different manner. I'm removing these changes and will attempt to fix the cypress issues in a separate PR.

mistercrunch:
Makes sense, converted to draft, feel free to reflag for review if you want to push this one through

PR #30082, opened 2024-08-31
feat: Oauth2 in DatabaseSelector

betodealmeida:
### SUMMARY

This PR modifies `onError` in the `useSchemas` and `useCatalogs` hooks so that the error payload is passed as a parameter. This allow us to prompt the user for OAuth2 when adding datasets or browsing tables in SQL Lab.

Depends on #30071 and #30067.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

After:

Screenshot 2024-08-31 at 19-21-36 Superset

Screenshot 2024-08-31 at 19-21-30 Superset

Before, generic error messages would flash.
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #30126, opened 2024-09-02
feat: OAuth2 database field

betodealmeida:
### SUMMARY

More work on Snowflake OAuth2 (this PR depends on #30082). This PR introduces a new field for the database configuration modal, allowing OAuth2 client information to be provided by the user:

Screenshot 2024-09-02 at 8 57 34 AM

The component combines the 5 different input fields, and then passes them inside the `masked_encrypted_extra` attribute on database CRUD:

```json
{
  "oauth2_client_info": {
    "id": "123456",
    "secret": "XXXXXXXXXXXX",
    "authorization_request_uri": "https://account.snowflakecomputing.com/oauth/authorize",
    "token_request_uri": "https://account.snowflakecomputing.com/oauth/token-request",
    "scope": "refresh_token session:role:USERADMIN"
  }
}
```

(Note that ideally we would pass these in the `parameters` field, and let the DB engine spec build the object in `encrypted_extra`. But the `build_sqlalchemy_uri` method — as the name suggests — only builds an URI, so we would have to modify it to return a tuple with `(uri, connect_args)` if we passed them client information inside `parameters`, and also figure out a way to inform that these fields should be stored in `encrypted_extra`, not `extra`, so for now the frontend does the work. I left a small TODO for me to fix this in the future.)

See above.

### TESTING INSTRUCTIONS

Adding tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/CommonParameters.tsx
```diff
@@ -161,23 +161,20 @@ export const httpPathField = ({
   getValidation,
   validationErrors,
   db,
-}: FieldPropTypes) => {
-  console.error(db);
```
betodealmeida, line 165: Removed this.

superset-frontend/src/features/databases/DatabaseModal/index.tsx
```diff
@@ -1656,6 +1666,16 @@ const DatabaseModal: FunctionComponent<DatabaseModalProps> = ({
             value: target.value,
           })
         }
+        onEncryptedExtraInputChange={({
+          target,
+        }: {
+          target: HTMLInputElement;
+        }) =>
+          onChange(ActionType.EncryptedExtraInputChange, {
```
rtexelm, line 1674: This is super helpful!

rtexelm (approved):

superset-frontend/src/features/databases/DatabaseModal/index.tsx
betodealmeida: Yeah, I was surprised it didn't exist yet!

PR #30128, opened 2024-09-02
chore(View): Remove unnecessary theme view and defer basic styles

geido:
### SUMMARY
Removes unnecessary theme view and defers styles in basic.html

### TESTING INSTRUCTIONS
1. UI should work as normal

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/templates/superset/basic.html
```diff
@@ -1,124 +1,130 @@
...
+      type="{{favicon.type if favicon.type else "image/png"}}" {% if favicon.sizes
+      %}sizes={{favicon.sizes}}{% endif %} href="{{ "" if
+      favicon.href.startswith("http") else assets_prefix }}{{favicon.href}}" > {%
+      endfor %}
+      <link
+        rel="stylesheet"
+        type="text/css"
+        href="{{ assets_prefix }}/static/appbuilder/css/flags/flags16.css"
+        media="print"
+        onload="this.media='all'"
```
kgabryje, line 36: I don't think this will work with prod CSP settings - violates no-eval rule

superset/templates/superset/basic.html
```diff
@@ -1,124 +1,130 @@
...
+      favicons %} <link rel="{{favicon.rel if favicon.rel else "icon"}}"
+      type="{{favicon.type if favicon.type else "image/png"}}" {% if favicon.sizes
+      %}sizes={{favicon.sizes}}{% endif %} href="{{ "" if
+      favicon.href.startswith("http") else assets_prefix }}{{favicon.href}}" > {%
+      endfor %}
+      <link
+        rel="stylesheet"
+        type="text/css"
+        href="{{ assets_prefix }}/static/appbuilder/css/flags/flags16.css"
+        media="print"
```
kgabryje, line 35: what does `media="print"` do?

geido: It's a performance optimization trick.  Initially, the stylesheet is loaded with media="print", so it’s fetched by the browser but not immediately applied.

superset/templates/superset/basic.html
geido: Alright, makes sense. Let me see if we should handle this with a separate script.

superset/templates/superset/basic.html
```diff
@@ -1,124 +1,139 @@
...
+      {% block tail_js %} {% if not standalone_mode %} {{ js_bundle('menu') }} {%
+      endif %} {% if entry %} {{ js_bundle(entry) }} {% endif %} {% include
+      "tail_js_custom_extra.html" %} {% endblock %}
+
+      <!-- Handle deferred stylesheets -->
+      <script>
+        document.addEventListener('DOMContentLoaded', function() {
+          var preloadLinks = document.querySelectorAll('link[rel="preload"][as="style"]');
+
+          preloadLinks.forEach(function(link) {
```
kgabryje, line 130: load event is triggered for the whole page, so we don't need a separate event listener for each link

Also, I wonder if we can just swap `rel` on DOMContentLoaded event

kgabryje:
It looks like the rels were not swapped

image

geido:
> http://35.87.80.58:8080

Weirdly enough they look swapped to me on the ephemeral

Screenshot 2024-09-03 at 18 13 09

superset/views/core.py
```diff
@@ -857,10 +857,6 @@ def dashboard_permalink(
     def log(self) -> FlaskResponse:
         return Response(status=200)
 
-    @expose("/theme/")
```
michael-s-molina, line 860: @geido Just to confirm, is this not needed for server-side rendered pages?

superset/templates/superset/basic.html
```diff
@@ -1,124 +1,136 @@
...
-    endif %} {% if entry %} {{ js_bundle(entry) }} {% endif %} {% include
-    "tail_js_custom_extra.html" %} {% endblock %}
-  </body>
-</html>
+      {% block tail_js %} {% if not standalone_mode %} {{ js_bundle('menu') }} {%
+      endif %} {% if entry %} {{ js_bundle(entry) }} {% endif %} {% include
+      "tail_js_custom_extra.html" %} {% endblock %}
+
+      <!-- Handle deferred stylesheets -->
+      <script>
```
rusackas, line 126: Should we be avoiding <script> tags in these templates, and move this to a `.js` file? ¯\\\_(ツ)_/¯

superset/views/core.py
geido: @michael-s-molina I think that was just an initial experiment to build an internal storybook, not really related to the theme.

rusackas (approved):
There's an unanswered question on the thread, but I don't see it as blocking. LG(enough)TM ❤️ 🚀

superset/templates/superset/basic.html
kgabryje: If we stick with this <script> tag, it needs a nonce to work with cap

PR #30132, opened 2024-09-02
feat(embedded): add hook to allow superset admins to validate guest token parameters

dmarkey:
### SUMMARY

This allows the Superset admin to tighten up Guest token configuration with a validator hook. Any aspect can be validated but most likely the RLS Clause.

### TESTING INSTRUCTIONS
Set ` GUEST_TOKEN_VALIDATOR_HOOK = lambda x: len(x["rls"]) == 1
            and "tenant_id=" in x["rls"][0]["clause"]` and the RLS clause in the guest token will need to contain "tenant_id="

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


nytai (approved):
LGTM

tests/integration_tests/security/api_tests.py
```diff
@@ -137,6 +137,149 @@ def test_post_guest_token_bad_resources(self):
...
+    @with_config({"GUEST_TOKEN_VALIDATOR_HOOK": None})
+    def test_guest_token_validator_hook_unconfigured(self):
+        """
+        Security API: Test the API works normally with GUEST_TOKEN_VALIDATOR_HOOK
+        configured to None.
+        """
+        self.dash = db.session.query(Dashboard).filter_by(slug="births").first()
+        self.embedded = EmbeddedDashboardDAO.upsert(self.dash, [])
+        self.login(ADMIN_USERNAME)
+        user = {"username": "bob", "first_name": "Bob", "last_name": "Also Bob"}
```
giftig, line 150: This could do with being a bit more DRY, looks like you're using the same user, resource and rls rule in almost all of these tests so I'd suggest just defining them once, or using pytest fixtures.

Basic user details and resource fixtures seem like things we must already have test fixtures for so you can probably reuse something existing.

dmarkey: Did some test consolidation there.

tests/integration_tests/security/api_tests.py
```diff
@@ -137,6 +138,79 @@ def test_post_guest_token_bad_resources(self):
...
+
+        self.assert400(response)
+
+        # Test True case from validator - should be 200
+        current_app.config["GUEST_TOKEN_VALIDATOR_HOOK"] = lambda x: True
+        response = self.client.post(
+            self.uri, data=json.dumps(params), content_type="application/json"
+        )
+
+        self.assert200(response)
```
giftig, line 169: I think these scenarios were better as they were before, as individual independent tests. I wasn't suggesting making one large test out of them, just reusing the common fixtures.

dmarkey: @nytai Do you have a preference? This *feels* better as it involves loading that heavy dashboard fixture only once.. the rest it just pushing different settings.

giftig: If the operation is intensive you could create a test class and create a fixture with `scope='class'`. Create a pytest fixture with `yield`: https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization

and give it `scope='class'` so that the setup/teardown is only run once for the class.

rusackas:
Thanks @giftig and @dmarkey for dialing this in. @nytai already approved it, which is giving me an itchy trigger finger. If any of @giftig 's change requests are blockers, feel free to hit the "request changes" button to block the merge. Just make sure you're around to _unblock_ it when the required adjustments are made ;) Thanks all :D

giftig:
My comments aren't really blocking, just a small improvement which would be nice imo. It's an easy +1 for me if the tests can be separated like they were before but without the repetition of test fixtures, but it's not a big enough issue to block the PR.

dmarkey:
Alright did another test refactor, broke out the tests again and moved them to their own class with the heavy dashboard fixture loaded in the class scope.

tests/integration_tests/security/api_tests.py
```diff
@@ -138,6 +138,85 @@ def test_post_guest_token_bad_resources(self):
         self.assert400(response)
 
 
+@pytest.mark.usefixtures("load_birth_names_dashboard_with_slices", scope="class")
+class TestSecurityGuestTokenApiTokenValidator(SupersetTestCase):
+    uri = "api/v1/security/guest_token/"  # noqa: F541
```
giftig, line 143: Is this noqa needed? Afaict F541 is "don't use unnecessary f-strings", which you haven't here. The linter didn't complain when I removed it just now.

giftig (approved):
Minor question about noqa but LGTM

dmarkey:
Not sure if I need to take action based on that failed cypress test?

villebro:
@dmarkey nothing that you've caused - we're investigating it currently, but it seems like a flaky test that may be caused by a perf regression that's been recently introduced. We're looking into it, but for now we'll just keep restarting the CI check until it passes..

dmarkey:
Thanks for merging! I guess it will be in 4.2.0?

rusackas:
No, thank YOU! 

It'll _definitely_ be in 5.0, and if we do a 4.2 (TBD) it'll definitely be in there, too. Right now, @sadpandajoe is building 4.1 RCs. I've labeled this PR to get cherries, but I'm not sure if that'll happen or not at this point.

Issue #26319, opened 2023-12-20
[SIP-112] Allow to deploy Superset under the prefix of an URL


vadym-serdiuk:
### Motivation

We faced a problem in the deployment of the Superset under the prefix like https://{domain}/superset. We can not use subdomains.
Configuring a gateway (like nginx) didn't work since there were problems with redirects. After a lot of efforts we decided to refactor the Superset. Furthermore, we saw many requests and questions about how to deploy Superset under the prefix without any working solutions.

### Proposed Change

I refactored a backend code by replacing simple string urls with the `url_for` function on the backend, or with the APPLICATION_ROOT variable in places with no app context. I added a function `withPrefix` on the frontend and use it for all the urls, generated on the frontend. 

### New or Changed Public Interfaces

There are no changes in the public interfaces

### New dependencies

No new dependencies

### Migration Plan and Compatibility

No migrations

### Rejected Alternatives

Configuration of gateways is not enough.

### Configuration

To use deployment with the prefix you just need to add the following variables into the Superset config

```python
ENABLE_PROXY_FIX = True
STATIC_ASSETS_PREFIX = "/superset"
APPLICATION_ROOT = "/superset"
```
and to configure the gateway with this minimal configuration (nginx syntax) 

```nginx
location /superset/ {
    proxy_pass         http://{superset_host}:{superset_port}/;
    proxy_set_header   X-Forwarded-Prefix /superset/;
}
```
Be noticed that the trailing slashes are important for the nginx configuration

sfirke:
People in the Slack chat often ask how to deploy like this, so I can confirm there is community demand for the feature to come out-of-the-box.  At the very least, this SIP should result in adding to the docs (though I hope it can result in implementing the feature).

sanasreenivasulu:
is the change complete, how to test

sfirke:
The linked pull request is a draft so no this has not been implemented. @vadym-serdiuk do you need any support/contributions from committers? There is now a recurring SIP call to discuss ideas like this and get them across the finish line

vadym-serdiuk:
@sfirke 
Hi,
[...]

PR #30134, opened 2024-09-03
feat: Allow superset to be deployed under a prefixed URL

martyngigg:
### SUMMARY

Introduces changes in the front and backend to allow Superset to be deployed under a prefixed-URL path rather than the root of the domain.

The backend `superset.app:create_app` has been augemented with a new `app_root` argument that accepts a root path for the application. This is passed through either a `SUPERSET_APP_ROOT` environment variable or by customizing the `FLASK_APP` environment variable or the `flask --app` cli argument, e.g. `FLASK_APP="superset:create_app(superset_app_root='/analytics')"`

When the path is not `/` a middleware is created that intercepts paths and updates them to strip the `app_root` and forward them to the rest of the application. This was inspired by [Application Dispatching](https://flask.palletsprojects.com/en/stable/patterns/appdispatch/).

The backend has also been refactored to avoid using hardcoded paths in redirects and instead uses `Flask.url_for` to construct all references. 

The frontend has been updated to include the prefix in any resource links along with two new helper utilities, `assetUrl` & `pathUtils`, to help construct the paths. The `SupersetClient` has been updated to remove the unused `baseUrl` field and rename it as `basePath` such that the object can be initialized with the base path and calling any endpoints through this class can use the same relative link as before. QUESTION: Have I broken an API here?

Documentation has been added to the "Configuring Superset" pages that demonstrate how to use the new argument.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.

martyngigg:
> Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.

Yes agreed I definitely need to add some documentation for how to use this. I wanted to wait until I was more confident that the general approach was acceptable before working on describing how to use it. In terms of a review as part of #26319 is there anything I need to do at the moment?

mistercrunch:
I think this may be quite an undertaking and might have to be tackled in phases. Would love to see this done :)

docker/docker-frontend.sh
```diff
@@ -17,19 +17,23 @@
 #
 set -e
 
+# zstd exe is required by simple-zstd package
+apt-get update
+apt-get install -y zstd
```
mistercrunch, line 22: there's a recent fix in master for this

ravikantkml:
> > Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.
> 
> Yes agreed I definitely need to add some documentation for how to use this. I wanted to wait until I was more confident that the general approach was acceptable before working on describing how to use it. In terms of a review as part of #26319 is there anything I need to do at the moment?

Thanks for these changes.
Can you please mention values for BASE_PATH and ASSET_BASE_URL. Just wanted to check these changes with superset and nginx.

kasper-rutten:
For what it's worth, we've been running this fork to serve superset on a path prefix in a Kubernetes(K3S) + traefik context without any problems so far. Though testing has been limited.

Thanks for the work so far

martyngigg:
> > > Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.
> > 
> > 
> > Yes agreed I definitely need to add some documentation for how to use this. I wanted to wait until I was more confident that the general approach was acceptable before working on describing how to use it. In terms of a review as part of #26319 is there anything I need to do at the moment?
> 
> Thanks for these changes. Can you please mention values for BASE_PATH and ASSET_BASE_URL. Just wanted to check these changes with superset and nginx.

Sorry for the delay in replying to this. I've had zero time to spend on this lately.

The latest commits rebase on the current master and I have also added some documentation to the configuring superset pages that describe the new variables with an example `nginx.conf`. Does that help?

rusackas:
If you think it's ready for review, I'd say you can go ahead and move it from Draft to "Ready for Review" - though it looks like CI might need a little more love still. 

As for the SIP, since @sfirke reopened the GitHub issue, I set the SIP itself back to "pre-discussion" state for its official consensus. Did you want to re-kindle that pre-existing SIP, or would it be easier to open a new one? I'm happy to help you carry that SIP through to fruition if you'd like. Feel free to DM me on [slack](http://bit.ly/join-superset-slack) if it's easier.

ascendlin:
When clicking on the user list or role list, and then clicking on datasets, a 404 error appears.

ravikantkml:
> > > > Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.
> > > 
> > > 
> > > Yes agreed I definitely need to add some documentation for how to use this. I wanted to wait until I was more confident that the general approach was acceptable before working on describing how to use it. In terms of a review as part of #26319 is there anything I need to do at the moment?
> > 
> > 
> > Thanks for these changes. Can you please mention values for BASE_PATH and ASSET_BASE_URL. Just wanted to check these changes with superset and nginx.
> 
> Sorry for the delay in replying to this. I've had zero time to spend on this lately.
> 
> The latest commits rebase on the current master and I have also added some documentation to the configuring superset pages that describe the new variables with an example `nginx.conf`. Does that help?

Thanks a lot for the response.
We tried the changes as recommended and observed that most of the things are working when we specify prefix. However there are few things where prefix is not applied correctly as shown in the image below (chart menu has prefix applied correctly)
image

Also dashboard menu items export to PDF/image not working (Dashboard -> Download -> Export to PDF/Download as Image)

ravikantkml:
> > > > > Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.
> > > > 
> > > > 
> > > > Yes agreed I definitely need to add some documentation for how to use this. I wanted to wait until I was more confident that the general approach was acceptable before working on describing how to use it. In terms of a review as part of #26319 is there anything I need to do at the moment?
> > > 
> > > 
> > > Thanks for these changes. Can you please mention values for BASE_PATH and ASSET_BASE_URL. Just wanted to check these changes with superset and nginx.
> > 
> > 
> > Sorry for the delay in replying to this. I've had zero time to spend on this lately.
> > The latest commits rebase on the current master and I have also added some documentation to the configuring superset pages that describe the new variables with an example `nginx.conf`. Does that help?
> 
> Thanks a lot for the response. We tried the changes as recommended and observed that most of the things are working when we specify prefix. However there are few things where prefix is not applied correctly as shown in the image below (chart menu has prefix applied correctly) image
> 
> Also dashboard menu items export to PDF/image not working (Dashboard -> Download -> Export to PDF/Download as Image)

I tried to fix above issue of missing prefixes in top-level menu. Please check this PR [2](https://github.com/martyngigg/superset/pull/2)

Dinesh-4320:
The solution really works. Solved a long term issue in our deployment behind proxies. Hoping to get merged soon with official code base

harri2012:
Very good, looking forward to merging the code soon

watercraft:
For whom it may concern, I'm testing with a port of this work to 3.1.1rc1 available from watercraft:superset branch prefix

jeanpommier:
Hi,
Good job there, thank you very much for tackling this. I have just discovered that Superset did not support that and was feeling disappointed.

I can see this PR is still flagged as DRAFT. What's left to do to get it "Ready for Review" ? 

@martyngigg do you need help in completing this PR ? I'm quite a superset newbie, but I'm willing to provide some help if necessary to get this merged.

martyngigg:
Dear all,

Thank you so much for taking the time to test out these changes and I apologise (again!) for the lack of response to any comments. @ravikantkml Thanks for the PR fixing the menu links - I think that looks okay and I'll merge it. 

I think the remaining issues are fixing the issues with screenshot generation. I had a brief look at this today and I'm having trouble getting them working even without any prefixed path set - I must have an issue in my development setup.

Today is my last day of work before Christmas break but I will have time when I return to pick this up properly. I'll aim to get it to "Ready for Review" as soon as possible after the holidays. @jeanpommier Thanks for the offer of help too - If you wanted to try and tackle the screenshot generation problem then that would be super helpful but equally I'm happy to look at it after the break.

Happy Holidays! 🎅🏻

landryb:
testing the PR here outside of the docker context (eg pip install from that github branch branch, rebuild the assets with BASE_PATH set), that work is much welcome.

some notes so far.. i think it should be mentioned that APP_ICON and FAVICON needs BASE_PATH prepended, otherwise it generates 404's here with the default value from `config.py`

looking a bit more at the code/PR, all assets below STATIC_ASSETS_PREFIX also need to default to BASE_PATH ? that fixes the FAVICON while here.

with this config the superset homepage loads fine with all its assets/images/favicon:
```
BASE_PATH = "/dashboards"                                                                                                                                                                                                                                                     
APP_ICON = BASE_PATH + "/static/assets/images/superset-logo-horiz.png"                                                                                                                                                                                                        
STATIC_ASSETS_PREFIX = BASE_PATH         
```

martyngigg:
Happy New Year!

I've investigated further and found the issue with "Export to PDF" & "Download image" not working on dashboards. The issue seems to be the interaction of the headless browser and `WEBDRIVER_BASEURL`. In the docker compose setup `superset_config.py` suggests setting `WEBDRIVER_BASEURL=http://superset_app:8088/`, so accessing the superset instance directly and skipping the nginx reverse proxy. This is fine when there is no url prefix set as the paths that both the reverse proxy and the superset app see are the same. When there is a url prefix set it is baked into the javascript files that webpack builds and when accessing the superset app directly the extra prefix part of the path raises an error in the superset app and the screenshot generation fails.

A fix/workaround (??) seems to be to set `WEBDRIVER_BASEURL = f"http://superset_nginx{base_path}/"` but this also requires removing this [with statement](https://github.com/apache/superset/blob/109e6c69ff25b82b603a4daff40207250788e6b1/superset/utils/urls.py#L35) that seems to set a test context that doesn't know about the prefix. I'm not sure of the knock on effects of this. @rusackas Would you know someone who could comment on this or possibly suggest an alternative approach?

jeanpommier:
Hello ! Happy new year !

> This is fine when there is no url prefix set as the paths that both the reverse proxy and the superset app see are the same. When there is a url prefix set it is baked into the javascript files that webpack builds and when accessing the superset app directly the extra prefix part of the path raises an error in the superset app and the screenshot generation fails.

Well, I might have a solution for this one. I didn't get the time to look at the image generation task you proposed, but I was rather investigating around the path prefix topic. My concern was to avoid relying on a reverse-proxy for the path-prefix thing. IMO, it makes things complicated. For instance the issue you met above: dev setup (without docker compo) and prod-like setup with a reverse proxy works differently, with a different frontend where the path is hard-coded.
Looking on the backend side of it, my first idea was that we maybe didn't use the Flask blueprints to their full extent, but it turns out this would not be that easy to adapt to our needs. And maybe not a clean solution.

***But***, I stumbled across [this blog post](https://dlukes.github.io/flask-wsgi-url-prefix.html) where I discovered that WSGI protocol supports a parameter that sets such a path prefix, and that is handled on the WSGI side of it (e.g. gunicorn). I tested it (on top of your PR since we need the url_for links) and it works like a charm.

**What does it change ?**

Supposing you choose `analytics` as the path prefix, the backend is accessed on http://superset_app:8088/analytics/. and the reverse-proxy has a much simpler job of just routing /analytics to http://superset_app:8088/analytics/. 
On your above-mentioned issue, this means that the paths will concord between the front and back end, so it should work without even needing the WEBDRIVER_BASEURL fix you propose.

I'm still quite uncomfortable at the necessity to build the frontend to support a path other than root. I haven't made experiments yet on this, but wouldn't it be conceivable that the frontend looks for a config instruction from the backend to get this information ?

martyngigg:
> Hello ! Happy new year !
> 
> > This is fine when there is no url prefix set as the paths that both the reverse proxy and the superset app see are the same. When there is a url prefix set it is baked into the javascript files that webpack builds and when accessing the superset app directly the extra prefix part of the path raises an error in the superset app and the screenshot generation fails.
> 
> Well, I might have a solution for this one. I didn't get the time to look at the image generation task you proposed, but I was rather investigating around the path prefix topic. My concern was to avoid relying on a reverse-proxy for the path-prefix thing. IMO, it makes things complicated. For instance the issue you met above: dev setup (without docker compo) and prod-like setup with a reverse proxy works differently, with a different frontend where the path is hard-coded. Looking on the backend side of it, my first idea was that we maybe didn't use the Flask blueprints to their full extent, but it turns out this would not be that easy to adapt to our needs. And maybe not a clean solution.
> 
> _**But**_, I stumbled across [this blog post](https://dlukes.github.io/flask-wsgi-url-prefix.html) where I discovered that WSGI protocol supports a parameter that sets such a path prefix, and that is handled on the WSGI side of it (e.g. gunicorn). I tested it (on top of your PR since we need the url_for links) and it works like a charm.
> 
> **What does it change ?**
> 
> Supposing you choose `analytics` as the path prefix, the backend is accessed on http://superset_app:8088/analytics/. and the reverse-proxy has a much simpler job of just routing /analytics to http://superset_app:8088/analytics/. On your above-mentioned issue, this means that the paths will concord between the front and back end, so it should work without even needing the WEBDRIVER_BASEURL fix you propose.

@jeanpommier Thanks for exploring and trying out the idea with having the backend also run on the prefix. I remember trying something like this a while ago and for some reason it didn't work properl. I think I may have been a bit hasty in dropping the idea and it was likely my lack of understanding of Superset that was the problem. I certainly agree relying on the proxy for the prefix handling is complicated. I'll try again since it certainly seems to just require an environment variable :)

> I'm still quite uncomfortable at the necessity to build the frontend to support a path other than root. I haven't made experiments yet on this, but wouldn't it be conceivable that the frontend looks for a config instruction from the backend to get this information ?

Yes this is also annoying. Now I'm a bit more familiar with the code base I wondered about including it in this https://github.com/apache/superset/blob/a986a61b5ff0c98d4e0704a1c2d7016b1e3ec721/superset/views/base.py#L333 config dictionary that is rendered into the templates? Anyone know if this is a good/bad idea?

wuqicyber:
> Happy New Year!  新年快乐！
> 
> I've investigated further and found the issue with "Export to PDF" & "Download image" not working on dashboards. The issue seems to be the interaction of the headless browser and `WEBDRIVER_BASEURL`. In the docker compose setup `superset_config.py` suggests setting `WEBDRIVER_BASEURL=http://superset_app:8088/`, so accessing the superset instance directly and skipping the nginx reverse proxy. This is fine when there is no url prefix set as the paths that both the reverse proxy and the superset app see are the same. When there is a url prefix set it is baked into the javascript files that webpack builds and when accessing the superset app directly the extra prefix part of the path raises an error in the superset app and the screenshot generation fails.我已经进一步调查并发现“导出为 PDF”和“下载图片”在仪表板上无法工作的问题。问题似乎是 headless 浏览器和 `WEBDRIVER_BASEURL` 之间的交互。在 docker compose 设置中， `superset_config.py` 建议设置 `WEBDRIVER_BASEURL=http://superset_app:8088/` ，因此直接访问 superset 实例并跳过 nginx 反向代理。如果没有设置 URL 前缀，这是可以的，因为反向代理和 superset 应用看到的路径是相同的。当设置了 URL 前缀时，它会被烘焙到 webpack 构建的 javascript 文件中，并且当直接访问 superset 应用时，路径的额外前缀部分会在 superset 应用中引发错误，截图生成失败。
> 
> A fix/workaround (??) seems to be to set `WEBDRIVER_BASEURL = f"http://superset_nginx{base_path}/"` but this also requires removing this [with statement](https://github.com/apache/superset/blob/109e6c69ff25b82b603a4daff40207250788e6b1/superset/utils/urls.py#L35) that seems to set a test context that doesn't know about the prefix. I'm not sure of the knock on effects of this. @rusackas Would you know someone who could comment on this or possibly suggest an alternative approach?一个修复/解决方案似乎是设置 `WEBDRIVER_BASEURL = f"http://superset_nginx{base_path}/"` ，但这也需要删除这个 with 语句，该语句似乎设置了一个不知道前缀的测试上下文。我不确定这会有什么副作用。您认识可以对此发表评论或可能提出替代方法的人吗？

@martyngigg try APPLICATION_ROOT, by setting it to perfix, i found it works

jeanpommier:
> I'll try again since it certainly seems to just require an environment variable :)
Hi @martyngigg 

The trick is, you still need to set the BASE_PATH env var for the frontend (needs for now to be built with the prefix), but not for  the superset backend, since this is the WSGI server that handles the prefix.

Here is a patch that should work, to run it with the gunicorn server. I've recently rebased your PR against master, so you might get a bit of noise in the docker-compo: [gunicorn_script_name.patch.gz](https://github.com/user-attachments/files/18331559/gunicorn_script_name.patch.gz)

And a similar patch to run the same principle with the flask dev server. I still haven't looked at a common way to get this to work, but it shouldn't be too complicated (I guess there is some flag somewhere indicating if we run the dev server):
[flask_devserver_script_name.patch.gz](https://github.com/user-attachments/files/18331605/flask_devserver_script_name.patch.gz)

jeanpommier:
> Yes this is also annoying. Now I'm a bit more familiar with the code base I wondered about including it in superset/views/base.py#L333 `bootstrap_data` config dictionary that is rendered into the templates? Anyone know if this is a good/bad idea?

Hi @martyngigg from me point of view, it looks like the perfect place, but I'm no Superset developer.

I've been playing around quite a bit on this topic, but am still struggling about dynamic configuration of the path prefix for the frontend (i.e. without dedicated build). Did you make progress on your side ?

martyngigg:
> > Yes this is also annoying. Now I'm a bit more familiar with the code base I wondered about including it in superset/views/base.py#L333 `bootstrap_data` config dictionary that is rendered into the templates? Anyone know if this is a good/bad idea?
> 
> Hi @martyngigg from me point of view, it looks like the perfect place, but I'm no Superset developer.
> 
> I've been playing around quite a bit on this topic, but am still struggling about dynamic configuration of the path prefix for the frontend (i.e. without dedicated build). Did you make progress on your side ?

Hi @jeanpommier,

I've managed to make a little progress with this. Firstly I've improved the situation with the backend such that its no longer the reverse-proxy that strips the prefix. I've added a middleware on the Flask app that does it so now the prefix is understood by the application so it works with the Flask dev server, gunicorn and Nginx. [Changeset](https://github.com/martyngigg/superset/commit/a84951a1e078ac8bf6bf4cdd90662dacce144cf6)

Secondly, I've got a work in progress [commit](https://github.com/martyngigg/superset/commit/328b0adc9be8416fe22524ba9f2bb33f090a7784) that attempts to push the `application_root` through the bootstrap data now that the app knows about it. It mostly works with the exception of issues with stylesheet imports through webpack that generate the wrong path. I need to explore further but the webpack [docs](https://github.com/webpack/docs/wiki/configuration#outputpublicpath) suggest `__webpack_public_path__` might be a way to go.



 and also improved the backed to remove the need for stripping. I've successfully implemented a small middleware class in the backend

docker/.env
```diff
@@ -54,8 +54,9 @@ REDIS_HOST=redis
 REDIS_PORT=6379
 
 FLASK_DEBUG=true
+SUPERSET_APP_ROOT="/"
 SUPERSET_ENV=development
-SUPERSET_LOAD_EXAMPLES=yes
+SUPERSET_LOAD_EXAMPLES=no
```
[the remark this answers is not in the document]
martyngigg: Nice catch. Thanks!

jeanpommier:
Not a review, but at first look, it looks really neat. And quite some work done here !

Dockerfile
```diff
@@ -62,9 +62,9 @@ RUN --mount=type=bind,source=./superset-frontend/package.json,target=./package.j
     --mount=type=cache,target=/root/.cache \
     --mount=type=cache,target=/root/.npm \
     if [ "$DEV_MODE" = "false" ]; then \
-        npm ci; \
+    npm ci; \
```
mistercrunch, line 65: why removing the indentation here?

superset-frontend/src/utils/pathUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ */
+import { applicationRoot } from 'src/utils/getBootstrapData';
+
+/**
+ * Takes a string path to a resource and prefixes it with the application root that is
+ * defined in the application configuration. The path is unsanitized.
+ * @param path A string path to a resource
+ */
+export function ensureAppRootUnsanitized(path: string): string {
```
mistercrunch, line 26: I'm not sure if I understand what a Sanitized or Unsanitized path means in this context (?)

superset-frontend/src/utils/getBootstrapData.ts
```diff
@@ -25,3 +26,17 @@ export default function getBootstrapData(): BootstrapData {
...
+const APPLICATION_ROOT_NO_TRAILING_SLASH =
+  getBootstrapData().common.application_root.replace(/\/$/, '');
+
+/**
+ * @param dom_sanitize If true, run the application root through dompurify before returning
+ * @returns The configured application root
+ */
+export function applicationRoot(dom_sanitize = false): string {
+  return dom_sanitize
+    ? DOMPurify.sanitize(APPLICATION_ROOT_NO_TRAILING_SLASH)
```
mistercrunch, line 40: I'm not sure if i understand why we need to sanitize the app route every time we read it(?) Seems it'd be a very static thing(?) Do we have trust issues with the bootstrap data? Also concerned about parsing/reparsing bootstrap data every time we read print a URL. Assuming just the basic menu has dozens of links, we'd be running `JSON.parse` for each one?

mistercrunch (dismissed):
I did a light pass on the PR and found a few things that need addressing. Stepping back a little, from a frontend perspective, given the need for more centralized management of routes/urls, it seems we may want to create a central object with all routes and bring it real close to `react-router`, which I believe has semantics for handling prefixes and such. That may be a bigger undertaking, but given the goal is "prefix all the url" and seems there's some stuff in the PR around dompurifying/sanitizing urls, seems like a good time to centralize all routes in one place. Might require a bit of proposal/SIP maybe. Open to phasing things, but we should have a plan.

About the requested changes:
- please don't de-indent Dockerfile as indents help readability
- let's avoid the aggressive reparsing of bootstrap data, seems this is a bit of a problem already in the codebase counting ~20 calls to `getBootstrapData()` where it seems we should at least store/cache the result to avoid multiple JSON.parse on what is becoming a fairly large object. I think putting that in the path of every path would make our pour browsers do a lot of redundant work

About centralizing routes, not sure how exactly how we should design the solution, but at least having a central object with all routes and path-related utilities would be great. Seems some routes are already defined as constants places, and be have so Rison utilities somewhere.

mistercrunch:
Loosely related -> https://github.com/apache/superset/pull/31996

Dockerfile
martyngigg: Sorry, this wasn't intentional. My IDE seems hell bent on doing this. I'll fix it.

superset-frontend/src/utils/getBootstrapData.ts
martyngigg: This was an attempt to address a security warning that CodeQL highlighted in multiple places when assigning the result of the new `ensureAppRootSanitized` function in `pathUtils` to a `href` element, for example:

image

Reflecting on this, yes that will be a lot of repeated work (and I'm not crazy about the new names `ensureAppRootSanitized` & `ensureAppRootUnsanitized` in `pathUtils` but I at least wanted to be clear they were doing something with the root itself).

Would an acceptable solution for this case be to cache it as a module level variable in `getBootstrapData`? while a wider solution to reparsing the bootstrap data and centralizing the routing code is created?

superset-frontend/src/utils/pathUtils.ts
martyngigg: I'm not wild on the names but was trying to emphasize that this does nothing to the app root itself before prefixing it (see the next comment below regarding a security warning I was trying to address).

superset-frontend/src/utils/getBootstrapData.ts
martyngigg: Looking at the changed files in #31996 it seems like there are existing check annotations mentioning the same security warning in several other places, e.g. [GenericLink](https://github.com/apache/superset/pull/31996/files#diff-652975a3a5f9c68f1ba3fd9ee7a49068a35f124dd52ef33d34c8c0d7498c01b8).

Would it be worth simplifying the code here to remove the sanitization and those issues be addressed elsewhere?

mistercrunch: Not a security expert but from my understanding CodeQL is worried about dom-injection-type attacks, which probably don't apply to our usage in most cases (?). Pretty sure it wouldn't apply to the APP_ROOT that's coming from a trusted place (served by the backend) That would only apply in some specific cases. I'd assume that React anchor tags and react-router would already take care of that.

jeanpommier:
Hi @martyngigg 
I believe you might have missed some places in the frontend:
This is kind of a hedge case, but if you edit a dashboard, go to "save as", select overwrite, it is redirecting to /superset/dashboard/yourdashslug after you save a dashboard edit (no prefix anymore)

I believe it might be due to https://github.com/martyngigg/superset/blob/allow_prefixed_paths/superset-frontend/src/dashboard/actions/dashboardState.js#L410 messing up the browser history also, at some point.

Reproduce it: 
- edit a dashboard
- save as
- overwrite
- you get redirected to the prefix-less path

martyngigg:
> Hi @martyngigg I believe you might have missed some places in the frontend: This is kind of a hedge case, but if you edit a dashboard, go to "save as", select overwrite, it is redirecting to /superset/dashboard/yourdashslug after you save a dashboard edit (no prefix anymore)
> 
> I believe it might be due to https://github.com/martyngigg/superset/blob/allow_prefixed_paths/superset-frontend/src/dashboard/actions/dashboardState.js#L410 messing up the browser history also, at some point.
> 
> Reproduce it:
> 
>     * edit a dashboard
> 
>     * save as
> 
>     * overwrite
> 
>     * you get redirected to the prefix-less path

Thanks! I think I've fixed this now.

surajsrivathsa:
Hello All,

I am trying to run the superset with these changes in my local with docker-compose.yml and ran into some problems. It would be helpful if you have any insights on this.

I tried with both SUPERSET_APP_ROOT with both default "/" and a non-root path "/analytics/reports.
But i get below login screen and welcome screen which is not real screen.

Should i set any other configs to get this working? Thanks in advance for your help

image
image

mistercrunch:
@surajsrivathsa , looks like your frontend build didn't work, can you fire up `docker-compose up --build` on `master` and see if that works? Another trick to rule out local setups/modifications is to start fresh in a new folder with a fresh `git clone`.

BrianJavier99:
Not sure if this is already reported, but:
image

+ > Data > Create Dataset button still doesn't reflect the prefixed URL

mistercrunch:
> Not sure if this is already reported

Yeah my main worry with this PR is that it doesn't put controls in place to make sure all urls use the prefix function on the backend and frontend. Wondering how we can best achieve this. Some ideas:
- write custom lint rules (`eslint` and `ruff`) looking for URLy strings and erroring
- run e2e tests with the prefix, now we might need this in a matrix of tests somehow because we want to confirm that tests pass with AND without a prefix set in config
- good old code review
- ?

BrianJavier99:
> > > > > > Thanks for tackling this... it's a popular request, no doubt about it. If you think it's relevant, we might also want to add an entry in the documentation somewhere.
> > > > > 
> > > > > 
> > > > > Yes agreed I definitely need to add some documentation for how to use this. I wanted to wait until I was more confident that the general approach was acceptable before working on describing how to use it. In terms of a review as part of #26319 is there anything I need to do at the moment?
> > > > 
> > > > 
> > > > Thanks for these changes. Can you please mention values for BASE_PATH and ASSET_BASE_URL. Just wanted to check these changes with superset and nginx.
> > > 
> > > 
> > > Sorry for the delay in replying to this. I've had zero time to spend on this lately.
> > > The latest commits rebase on the current master and I have also added some documentation to the configuring superset pages that describe the new variables with an example `nginx.conf`. Does that help?
> > 
> > 
> > Thanks a lot for the response. We tried the changes as recommended and observed that most of the things are working when we specify prefix. However there are few things where prefix is not applied correctly as shown in the image below (chart menu has prefix applied correctly) image
> > Also dashboard menu items export to PDF/image not working (Dashboard -> Download -> Export to PDF/Download as Image)
> 
> I tried to fix above issue of missing prefixes in top-level menu. Please check this PR [2](https://github.com/martyngigg/superset/pull/2)

Was this PR reverted? I think this was supposed to fix the issue that I have mentioned in my [previous comment](https://github.com/apache/superset/pull/30134#issuecomment-2720261157)

antonibote:
Hi, is there a docker image of this PR on docker hub?

mistercrunch:
@antonibote you can look at various CI step titled "Build & publish docker iages", I think they do push to a place where you may be able to pull from. But since CI isn't succeeding I'm not sure if I'd invest the time to test an image we know has issues.

mistercrunch:
Ok spoke with a few Preset folks and we'd like to help pushing this through, happy to help here. In order to merge this PR:
- resolve merge conflicts
- get through CI checks
- add note in `config.py` about this config flag being in BETA
- rule out the security issue raised by CodeQL, we generally need to be cautious when composing any user-provided things in URL watching for XSS. There's an opportunity for the wrapping function to sanitize urls centrally here...
- remove package-lock.json from this PR (shouldn't be affected)
- figure out how to jam this in the existing test matrixes - without doubling the number of unit tests. That means running existing matrix elements with/without APP_ROOT config. Say we we have mysql/postgres matrix, you make one instance be mysql-with-approot, postgress-without-approot

mistercrunch:
> "rule out the security issue raised by CodeQL"

Wondering if it'd make sense to generalize a `urlFor(url, sanitize=True)` (renaming the method to match the backend (?)) which would sanitize by default. Could expose other functionality like receiving an object of `params` and be able to url-encode to rison VS standard-querystring - not sure if it's in-scope for this PR or not...

> "figure out how to jam this in the existing test matrixes"
Someone can help with this, thinking we'll need the ability to set a `superset_config.py` in CI, maybe simply hooking up env vars. Say if `superset/config.py` reads from `SUPERSET__URL_PREFIX` (or similar), we can probably easily dynamically parameterize this in the matrix. I think we have an example somewhere of JSON-combination of matrix parameter that aren't a full cartesian product.

martyngigg:
> Not sure if this is already reported, but: image
> 
> `+ > Data > Create Dataset` button still doesn't reflect the prefixed URL

Sorry for the delay in reply.

The latest push should have fixed this.

martyngigg:
> 
> > "figure out how to jam this in the existing test matrixes"
> 
> Someone can help with this, maybe simply hooking this up env vars. Say if `superset/config.py` reads from `SUPERSET__URL_PREFIX` (or similar), we can probably easily dynamically parameterize this in the GHA matrixes. I think we have an example somewhere of JSON-combination of matrix parameter that aren't a full cartesian product.
> 
> Dug out the similar use case using a dynamicaly-parameterized matrix here -> https://github.com/apache/superset/blob/master/.github/workflows/docker.yml#L18-L27

@mistercrunch Regarding the testing, what do you think of something like https://github.com/apache/superset/pull/30134/commits/7ccb69c31a0f37f9ff1940f00c177410998f431a

mistercrunch:
> @mistercrunch Regarding the testing, what do you think of something like https://github.com/apache/superset/commit/7ccb69c31a0f37f9ff1940f00c177410998f431a

Looks like this doubles the size of the matrix, not sure if that's reasonable for a single feature... Would have to do something similar for backend tests (pytest) and normal unit tests (jest). For pytest, it's not sharded and could probably do say "mysql + app_root" and "postgres no-app-route", so it wouldn't affect the CI workload.

Exploring options here, maybe a nightly or only on `push` events on `master` would contain the workload quite a bit and allow us to catch-on-merge. 

Personal preference would be "only run the extra workload" on push to `master`, which is a fraction of the work from doing it on every PR push. Curious what other committers/PMC think though.

landryb:
testing this in a downstream project which got rebased on the last commits from this branch, superset being deployed under `/dashboards/`, when im on https://FQDN/dashboards/logmodelview/list/ the sqllab entry in the `SQL` menu points at `/sqllab/` instead of `/dashboards/sqllab/`

that's the same issue for the other entries in the `SQL`menu, and for the `/tablemodelview/list/` entry in the menu, they all lack the configured prefix. The two leftmost entries in the menu have the correct prefix in their link (so clicking on them lead to the right page).

i'm seeing the same issue when im on the other entries from the settings menu, eg if on `/dashboards/users/list/` the same sqllab/tablemodelview links lack the prefix.

kgabryje:
@landryb Thanks for reporting that, should be fixed now

kgabryje:
@martyngigg This broke only in FAB views - in those, the Menu is server side rendered

landryb:
> @landryb Thanks for reporting that, should be fixed now

thanks a lot, i've locally applied d463c0027aa and it fixes the broken links i had seen

eschutho (approved):

martyngigg:
Thank you to everyone for helping to get this merged! 🎉 😀

ryuuc:
For those who want to try deploying superset to k8s using the Helm chart before its official release, the following modifications are required:

1.  custom image tag, I found it from github ci https://github.com/apache/superset/actions/runs/14366622555/job/40281019265 , use the py311 version, since wsgiref.types does not exist on Python 3.10, there is a pr to fix https://github.com/apache/superset/pull/33072/files#diff-9689cb75a0ebe47604d59e0a54c904040cd60b45a283a87d0f8d77cac60751d8L21

```
image:
  repository: apachesuperset.docker.scarf.sh/apache/superset
  tag: 09b92e7-py311
  pullPolicy: IfNotPresent
``` 
2. install the required dependencies. refer to https://github.com/apache/superset/discussions/31431, but need to use `uv` instead of `pip`
```
bootstrapScript: |
  #!/bin/bash

  # Install system-level dependencies
  apt-get update && apt-get install -y \
    python3-dev \
    default-libmysqlclient-dev \
    build-essential \
    pkg-config 

  # Install required Python packages
  uv pip install --native-tls \
    authlib \
    psycopg2-binary \
    mysqlclient

  # Create bootstrap file if it doesn't exist
  if [ ! -f ~/bootstrap ]; then
    echo "Running Superset with uid {{ .Values.runAsUser }}" > ~/bootstrap
  fi
```
3.  Add `SUPERSET_APP_ROOT` env variables, and update health check paths
```
supersetNode:
  env:
    SUPERSET_APP_ROOT: /your-path
  livenessProbe:
    httpGet:
      path: /your-path/health
  readinessProbe:
    httpGet:
      path: /your-path/health
  startupProbe:
    httpGet:
      path: /your-path/health

```

superset/views/dashboard/views.py
```diff
@@ -86,7 +85,11 @@ def new(self) -> FlaskResponse:
         )
         db.session.add(new_dashboard)
         db.session.commit()  # pylint: disable=consider-using-transaction
-        return redirect(f"/superset/dashboard/{new_dashboard.id}/?edit=true")
+        return redirect(
+            url_for(
+                "Superset.dashboard", dashboard_id_or_slug=new_dashboard.id, edit=True
```
mistercrunch, line 90: Somehow this line broke some tests on `master` but I don't understand how it didn't get caught by CI (????).

Issues is with the capitalize `True` here... Maybe you IDE did this? Linter?

mistercrunch:
Debugging something funky that happening with this PR, somehow it failed when it merged on master and digging in the last commit's run here https://github.com/apache/superset/actions/runs/14382934496/job/40331066514, it appears that it skipped the python-related test, presumably because of an issue with our logic that checks which files have been touched by the PR and triggers/skips based on whether it sees the backend/frontend getting touched.

I'm guessing the issue is around the logic checking for files touched by the PR, and specific to [maybe] having a merge-based workflow instead of a rebase-type workflow.

Regardless, tests fail here and it's a bit tricky to sort this out, could use some help here to fix master

mistercrunch:
Related: https://github.com/apache/superset/pull/33092

mistercrunch:
@martyngigg  anyway you can fix the issues that landed through this PR on `master`? I tried and fix some but didn't have full context. Would prefer not reverting this PR

martyngigg:
It looks like the Python unit tests on HEAD of master seem to have a different failure:

```
/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/_pytest/config/__init__.py:331: PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.
Plugin: helpconfig, Hook: pytest_cmdline_parse
ConftestImportFailure: ModuleNotFoundError: No module named 'wsgiref.types' (from /home/runner/work/superset/superset/tests/conftest.py)
For more information see https://pluggy.readthedocs.io/en/stable/api_reference.html#pluggy.PluggyTeardownRaisedWarning
  config = pluginmanager.hook.pytest_cmdline_parse(
ImportError while loading conftest '/home/runner/work/superset/superset/tests/conftest.py'.
tests/conftest.py:41: in <module>
    from tests.integration_tests.test_app import app
tests/integration_tests/test_app.py:20: in <module>
    from superset.app import create_app
superset/__init__.py:20: in <module>
    from superset.app import create_app  # noqa: F401
superset/app.py:21: in <module>
    from wsgiref.types import StartResponse, WSGIApplication, WSGIEnvironment
E   ModuleNotFoundError: No module named 'wsgiref.types'
```
Is that expected?

kgabryje:
> It looks like the Python unit tests on HEAD of master seem to have a different failure:
> 
> ```
> /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/_pytest/config/__init__.py:331: PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.
> Plugin: helpconfig, Hook: pytest_cmdline_parse
> ConftestImportFailure: ModuleNotFoundError: No module named 'wsgiref.types' (from /home/runner/work/superset/superset/tests/conftest.py)
> For more information see https://pluggy.readthedocs.io/en/stable/api_reference.html#pluggy.PluggyTeardownRaisedWarning
>   config = pluginmanager.hook.pytest_cmdline_parse(
> ImportError while loading conftest '/home/runner/work/superset/superset/tests/conftest.py'.
> tests/conftest.py:41: in <module>
>     from tests.integration_tests.test_app import app
> tests/integration_tests/test_app.py:20: in <module>
>     from superset.app import create_app
> superset/__init__.py:20: in <module>
>     from superset.app import create_app  # noqa: F401
> superset/app.py:21: in <module>
>     from wsgiref.types import StartResponse, WSGIApplication, WSGIEnvironment
> E   ModuleNotFoundError: No module named 'wsgiref.types'
> ```
> 
> Is that expected?

I think that the problem is that wsgiref types is available since Python 3.11 and master branch runs on 3.10. But I can’t understand why those tests passed in this branch though…

martyngigg:
@kgabryje Thanks. #33902 contains fixes for the touched files logic that meant that the tests didn't run here. #33095 contains the fixes for those tests that I hadn't realised I'd broken.

mistercrunch:
Reporting that in `master`, simply clicking on "Datasets" in the navbar at the top leads to a 404. Same with `SQL Lab`, `Saved Queries` and many others:
Screenshot 2025-04-12 at 2 14 41 PM

kgabryje:
> Reporting that in `master`, simply clicking on "Datasets" in the navbar at the top leads to a 404. Same with `SQL Lab`, `Saved Queries` and many others: Screenshot 2025-04-12 at 2 14 41 PM

Opened a fix here https://github.com/apache/superset/pull/33114

wqbbs911:
> For those who want to try deploying superset to k8s using the Helm chart before its official release, the following modifications are required:
> 
> 1. custom image tag, I found it from github ci https://github.com/apache/superset/actions/runs/14366622555/job/40281019265 , use the py311 version, since wsgiref.types does not exist on Python 3.10, there is a pr to fix https://github.com/apache/superset/pull/33072/files#diff-9689cb75a0ebe47604d59e0a54c904040cd60b45a283a87d0f8d77cac60751d8L21
> 
> ```
> image:
>   repository: apachesuperset.docker.scarf.sh/apache/superset
>   tag: 09b92e7-py311
>   pullPolicy: IfNotPresent
> ```
> 
> 2. install the required dependencies. refer to [ModuleNotFoundError: No module named 'psycopg2' during k8 installation #31431](https://github.com/apache/superset/discussions/31431), but need to use `uv` instead of `pip`
> 
> ```
> bootstrapScript: |
>   #!/bin/bash
> 
>   # Install system-level dependencies
>   apt-get update && apt-get install -y \
>     python3-dev \
>     default-libmysqlclient-dev \
>     build-essential \
>     pkg-config 
> 
>   # Install required Python packages
>   uv pip install --native-tls \
>     authlib \
>     psycopg2-binary \
>     mysqlclient
> 
>   # Create bootstrap file if it doesn't exist
>   if [ ! -f ~/bootstrap ]; then
>     echo "Running Superset with uid {{ .Values.runAsUser }}" > ~/bootstrap
>   fi
> ```
> 
> 3. Add `SUPERSET_APP_ROOT` env variables, and update health check paths
> 
> ```
> supersetNode:
>   env:
>     SUPERSET_APP_ROOT: /your-path
>   livenessProbe:
>     httpGet:
>       path: /your-path/health
>   readinessProbe:
>     httpGet:
>       path: /your-path/health
>   startupProbe:
>     httpGet:
>       path: /your-path/health
> ```

seems it doesn't work in 5.0.0rc2

wqbbs911:
5.0.0 was released, seems still doesn't work

mistercrunch:
5.0 was cut in Feb ... took a while to stabilize and release the branch. 6.0 should be a fast-follow though, hoping around a month from now.

landryb:
this has been merged a while ago and is working fine, but i think the swagger link doesnt have the prefix, eg going to https://instance/path/swagger/v1 tries to fetch `/api/v1/_openapi` without the prefix (which should be coming from `application_root` afaict), which yields a 404. 

looking at the code but so far i fail to see where this should be configured/handled, eg in FAB or superset frontend itself or somewhere else...

**edit** bah i guess that's already been reported as #33304

mistercrunch:
labeled https://github.com/apache/superset/issues/33304 as "good first issue" for hopefully someone to pick up.

Ayoub-BOUCHACHIA:
I'm using Superset 4.1.2, but the solution isn't working for me. Can anyone explain why?

Here’s the relevant output from my logs:

```bash
[notice] A new release of pip is available: 25.0.1 -> 25.2
[notice] To update, run: pip install --upgrade pip
Starting web app (using development server)...
Usage: flask run [OPTIONS]
Try 'flask run --help' for help.

Error: The factory "create_app(superset_app_root='/archipel')" in module 'superset' could not be called with the specified arguments.
+ '[' 0 -eq 1 ']'
+ [[ 1 -gt 0 ]]
+ [[ dev == \d\e\v ]]
+ echo 'Skipping gunicorn, running app in dev mode'
+ bash /app/docker/docker-bootstrap.sh app
Skipping gunicorn, running app in dev mode
Installing local overrides at /app/docker/requirements-local.txt
Requirement already satisfied: psycopg2-binary==2.9.6 in /usr/local/lib/python3.10/site-packages (from -r /app/docker/requirements-local.txt (line 1)) (2.9.6)
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.

[notice] A new release of pip is available: 25.0.1 -> 25.2
[notice] To update, run: pip install --upgrade pip
Starting web app (using development server)...
Usage: flask run [OPTIONS]
Try 'flask run --help' for help.

Error: The factory "create_app(superset_app_root='/archipel')" in module 'superset' could not be called with the specified arguments.
```

jeanpommier:
Hi @Ayoub-BOUCHACHIA 
I don't think it's been ported to 4.1.2.
Actually, AFAIK, it's not part of any release at the current time: it's been merged on master branch, but not released yet.

PR #30142, opened 2024-09-03
feat(jinja): add advanced temporal filter functionality

villebro:
### SUMMARY
Currently the `from_dttm` and `to_dttm` Jinja variables make it possible to reference the time range endpoints from within the Jinja context. However, they have some shortcomings:
- `to_dttm` always points to the current day, even if there's no time range present in the query (this appears to be a quirk of how the `get_since_until` helper function works).
- They are always formatted using the `.isoformat()` method. This means that it's difficult to use them for database-specific formatting that's readily available in `BaseEngineSpec.convert_dttm`.
- It's impossible to access a specific time filter if there are multiple time filters present.
- In the case of virtual datasets, it's not possible to remove the time filter from the outer query (this is possible for regular filter types using the `filter_value` and `get_filters` macros).

This PR adds a new Jinja macro called `get_time_filter`, which returns a dataclass with the following properties:
- `from_expr`: The SQL expression representing the start of the time range, if available
- `to_expr`: The SQL expression representing the end of the time range, if available
- `time_range`: The (usually) human readable raw value of the time filter (e.g. "No filter", "Last week" etc)

Please refer to the added docs for how the feature works in practice:

image
image

### TESTING INSTRUCTIONS
- Run examples in added docs
- Added unit tests (100 % coverage for added logic)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/configuration/sql-templating.mdx
+++ b/docs/docs/configuration/sql-templating.mdx
@@ -413,7 +413,6 @@ FROM logs
 WHERE
   dttm >= {{ time_filter.from_expr }}
   AND dttm < {{ time_filter.to_expr }}
-{% endif %}
 ```
 
 **Datasets**
--- a/docs/docs/configuration/sql-templating.mdx
+++ b/docs/docs/configuration/sql-templating.mdx
@@ -348,35 +348,38 @@ Here's a concrete example:
 
 **Time Filter for a Specific Column**
 
-The `{{ get_time_filter() }}` macro returns the time filter applied to a specific column. The macro takes the following
-parameters:
+The `{{ get_time_filter() }}` macro returns the time filter applied to a specific column. This is useful if you want
+to handle time filters inside the virtual dataset, as by default the time filter is placed on the outer query. This can
+have considerable performance implications, as many databases and query engines are able to optimize the query better
+if the temporal filter is placed on the inner query, as opposed to the outer query.
+
+The macro takes the following parameters:
 - `column`: Name of the temporal column. Leave undefined to reference the time range from a Dashboard Native Time Range
   filter (when present).
 - `target_type`: The target temporal type as recognized by the target  database (e.g. `TIMESTAMP`, `DATE` or
   `DATETIME`). If `column` is defined, the format will default to the type of the column. This is used to produce
-  the format of the `from_dttm` and `to_dttm` properties of the returned `TimeFilter` object. Note, that omitting
-  `column` and `target_type` will render format the temporal values as ISO format.
+  the format of the `from_expr` and `to_expr` properties of the returned `TimeFilter` object.
 
 The return type has the following properties:
-- `since`: the start of the time filter (if any)
-- `until`: the end of the time filter (if any)
+- `from_expr`: the start of the time filter (if any)
+- `to_expr`: the end of the time filter (if any)
 - `time_range`: The time range as defined on the
 
 Here's a concrete example using the `logs` table from the Superset metastore:
 
 ```
 {% set time_filter = get_time_filter("dttm", remove_filter=True) %}
-{% set since = time_filter.since %}
-{% set until = time_filter.until %}
+{% set from_expr = time_filter.from_expr %}
+{% set to_expr = time_filter.to_expr %}
 {% set time_range = time_filter.time_range %}
 
 SELECT
   *,
   {% if time_range %}'{{ time_range }}'{% else %}''{% endif %} as time_range
 FROM logs
-{% if since or until %}WHERE 1 = 1
-{% if since %}AND dttm >= {{ since }}{% endif %}
-{% if until %}AND dttm < {{ until }}{% endif %}
+{% if from_expr or to_expr %}WHERE 1 = 1
+{% if from_expr %}AND dttm >= {{ from_expr }}{% endif %}
+{% if to_expr %}AND dttm < {{ to_expr }}{% endif %}
 {% endif %}
 ```
 
--- a/docs/docs/configuration/sql-templating.mdx
+++ b/docs/docs/configuration/sql-templating.mdx
@@ -346,7 +346,7 @@ Here's a concrete example:
     order by lineage, level
 ```
 
-**Time Filter for a Specific Column**
+**Time Filter**
 
 The `{{ get_time_filter() }}` macro returns the time filter applied to a specific column. This is useful if you want
 to handle time filters inside the virtual dataset, as by default the time filter is placed on the outer query. This can
@@ -356,9 +356,12 @@ if the temporal filter is placed on the inner query, as opposed to the outer que
 The macro takes the following parameters:
 - `column`: Name of the temporal column. Leave undefined to reference the time range from a Dashboard Native Time Range
   filter (when present).
+- `default`: The default value to fall back to if the time filter is not present, or has the value `No filter`
 - `target_type`: The target temporal type as recognized by the target  database (e.g. `TIMESTAMP`, `DATE` or
   `DATETIME`). If `column` is defined, the format will default to the type of the column. This is used to produce
   the format of the `from_expr` and `to_expr` properties of the returned `TimeFilter` object.
+- `remove_filter`: When set to true, mark the filter as processed, removing it from the outer query. Useful when a
+  filter should only apply to the inner query.
 
 The return type has the following properties:
 - `from_expr`: the start of the time filter (if any)
@@ -372,10 +375,9 @@ Here's a concrete example using the `logs` table from the Superset metastore:
 {% set from_expr = time_filter.from_expr %}
 {% set to_expr = time_filter.to_expr %}
 {% set time_range = time_filter.time_range %}
-
 SELECT
   *,
-  {% if time_range %}'{{ time_range }}'{% else %}''{% endif %} as time_range
+  '{{ time_range }}' as time_range
 FROM logs
 {% if from_expr or to_expr %}WHERE 1 = 1
 {% if from_expr %}AND dttm >= {{ from_expr }}{% endif %}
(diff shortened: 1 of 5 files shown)


nytai:
So something that comes up often in the templateized datasets using these filter values is having defaults so that users can run the "sync columns from source" action and have their query execute. The dataset modal does have "Template parameters" setting but that actually ends up overriding anything set at the dashboard or chart level which is kind of unexpected. 

I've been advising users to always use if/else to check if the filter values are defined otherwise fallback to a reasonable default (eg last week or day), so that their query is valid even if not run in a dashboard or chart context. It would be good to bake this default behavior into the jinja expression so that expressing this is a little less verbose. Something like: 

`get_time_filter(..., default_time_range="Last Week")` and maybe even `get_time_filter(..., default_ from_expr ="2024-09-04", default_to_expr="2024-09-05")`

villebro:
> So something that comes up often in the templateized datasets using these filter values is having defaults so that users can run the "sync columns from source" action and have their query execute. The dataset modal does have "Template parameters" setting but that actually ends up overriding anything set at the dashboard or chart level which is kind of unexpected.
> 
> I've been advising users to always use if/else to check if the filter values are defined otherwise fallback to a reasonable default (eg last week or day), so that their query is valid even if not run in a dashboard or chart context. It would be good to bake this default behavior into the jinja expression so that expressing this is a little less verbose. Something like:
> 
> `get_time_filter(..., default_time_range="Last Week")` and maybe even `get_time_filter(..., default_from_expr ="2024-09-04", default_to_expr="2024-09-05")`

That's a really good idea 👍 To keep things simple, I would prefer only having `default_time_range`, as it would be possible to express the explicit dates in the latter example with the time range syntax as follows: `"2024-09-04 : 2024-09-05"`. And maybe even simplifying the naming further to keep it in line with the existing `filter_values`, and just calling it `default`:

```python
def get_time_filter(
    self,
    column: str | None = None,
    default: str | None = None,
    target_type: str | None = None,
    remove_filter: bool = False,
) -> TimeFilter:
    ...
    # logic to populate time_range from filter value
    ...
    if time_range == NO_TIME_RANGE and default:
        time_range = default
    ...
```

villebro:
@nytai I've added the default option in the latest commit: https://github.com/apache/superset/pull/30142/commits/288e6af1b08bd25be6d72ebf6d27e69c3aa3a7ba

nytai (approved):
LGTM

docs/docs/configuration/sql-templating.mdx
```diff
@@ -17,8 +17,8 @@ made available in the Jinja context:
 
 - `columns`: columns which to group by in the query
 - `filter`: filters applied in the query
-- `from_dttm`: start `datetime` value from the selected time range (`None` if undefined)
-- `to_dttm`: end `datetime` value from the selected time range (`None` if undefined)
+- `from_dttm`: start `datetime` value from the selected time range (`None` if undefined) (deprecated, use `get_time_filter` instead)
```
sfirke, line 20: Maybe say "deprecated beginning in version 4.1.0" since we don't have versioned docs?

docs/docs/configuration/sql-templating.mdx
```diff
@@ -346,6 +346,75 @@ Here's a concrete example:
     order by lineage, level
 ```
 
+**Time Filter**
+
+The `{{ get_time_filter() }}` macro returns the time filter applied to a specific column. This is useful if you want
+to handle time filters inside the virtual dataset, as by default the time filter is placed on the outer query. This can
+have considerable performance implications, as many databases and query engines are able to optimize the query better
```
sfirke, line 353:
```suggestion
considerably improve performance, as many databases and query engines are able to optimize the query better
```

sfirke: clarifying that the performance implications are positive

docs/docs/configuration/sql-templating.mdx
```diff
@@ -346,6 +346,75 @@ Here's a concrete example:
...
+- `target_type`: The target temporal type as recognized by the target  database (e.g. `TIMESTAMP`, `DATE` or
+  `DATETIME`). If `column` is defined, the format will default to the type of the column. This is used to produce
+  the format of the `from_expr` and `to_expr` properties of the returned `TimeFilter` object.
+- `remove_filter`: When set to true, mark the filter as processed, removing it from the outer query. Useful when a
+  filter should only apply to the inner query.
+
+The return type has the following properties:
+- `from_expr`: the start of the time filter (if any)
+- `to_expr`: the end of the time filter (if any)
+- `time_range`: The time range as defined on the
```
sfirke, line 369: This line is incomplete -- defined on the what?

docs/docs/configuration/sql-templating.mdx
```diff
@@ -346,6 +346,75 @@ Here's a concrete example:
...
+FROM logs
+{% if from_expr or to_expr %}WHERE 1 = 1
+{% if from_expr %}AND dttm >= {{ from_expr }}{% endif %}
+{% if to_expr %}AND dttm < {{ to_expr }}{% endif %}
+{% endif %}
+```
+
+Assuming we are creating a table chart with a simple `COUNT(*)` as the metric with a time filter `Last week` on the
+`dttm` column, this would render the following query on Postgres (note the formatting of the temporal filters, and
+the missing filters on the outer query):
```
sfirke, line 390:
```suggestion
the absence of time filters on the outer query):
```

superset/jinja_context.py
```diff
@@ -354,6 +375,78 @@ def get_filters(self, column: str, remove_filter: bool = False) -> list[Filter]:
...
+        :param column: Name of the temporal column. Leave undefined to reference the
+            time range from a Dashboard Native Time Range filter (when present).
+        :param default: The default value to fall back to if the time filter is
+            not present, or has the value `No filter`
+        :param target_type: The target temporal type as recognized by the target
+            database (e.g. `TIMESTAMP`, `DATE` or `DATETIME`). If `column` is defined,
+            the format will default to the type of the column. This is used to produce
+            the format of the `from_expr` and `to_expr` properties of the returned
+            `TimeFilter` object. Note, that omitting `column` and `target_type` will
+            render format the temporal values as ISO format.
```
sfirke, line 398:
```suggestion
            render the temporal values as ISO format.
```

sfirke (commented):
Docs & comments: A few documentation language suggestions and I pointed out a place where a sentence stops halfway.

Code: I glanced at the code, it's mostly beyond me so I'll defer to others there.

Overall: seems like a nice improvement in functionality, I appreciate the complete documentation to go with it 🙏

docs/docs/configuration/sql-templating.mdx
villebro: Good call - I'll deprecate as of 5.0, as this will likely not be cherried into the 4.1 branch.

villebro:
Thanks for the review @sfirke ! All very good callouts, I've addressed the comments in https://github.com/apache/superset/pull/30142/commits/0212fa9192848d634f77f2ca139436eda5f9bd81

giftig (approved):
LGTM

villebro:
I added a self-assigned task for removing docs for `from_dttm` and `to_dttm` in favor of this macro in 5.0: https://github.com/orgs/apache/projects/345?pane=issue&itemId=78905534

Vitor-Avila:
hey @villebro, 

This is amazing! Thank you so much for working on this new macro. 🙏 It's super powerful -- can't wait to test it. 🙌 

One question: you mentioned that one of the limitations with the `from_dttm` and `to_dttm` macros is the inability to apply multiple temporal filters. Could you confirm if this is now possible with this macro? I could see it happening since it supports specifying a `column`, however in the dashboard temporal filter a column is not specified, so I assume it's still impossible to add two distinct/indepentent temporal filters at the dashboard level -- is that correct?

villebro:
> hey @villebro,
> 
> This is amazing! Thank you so much for working on this new macro. 🙏 It's super powerful -- can't wait to test it. 🙌

Thanks for the feedback, super happy if this is useful to someone else, too ❤️ 

> One question: you mentioned that one of the limitations with the `from_dttm` and `to_dttm` macros is the inability to apply multiple temporal filters. Could you confirm if this is now possible with this macro? I could see it happening since it supports specifying a `column`, however in the dashboard temporal filter a column is not specified, so I assume it's still impossible to add two distinct/indepentent temporal filters at the dashboard level -- is that correct?

With this it's indeed possible to extract the temporal value for multiple different columns. However, in the context of dashboard filtering, unfortunately the dashboard time range filter doesn't yet support applying to a specific temporal column (when the time range filter applies a value, I believe it attaches to all temporal filters currently 🙁). What I'd like to see is adding an optional dataset and column dropdown in the time range filter, and if that's selected, then it will apply a temporal filter only to that column. I've heard lots of people ask for that, and it shouldn't be that difficult to implement IMO.

Vitor-Avila:
that totally makes sense! 🙏 I also believe this new macro would already allow you to pull distinct temporal filters from the chart context to apply to the inner query, which is super cool.

I 100% support and agree with those changes to the temporal filter - Ideally we should support selecting which temporal column is getting filtered, to allow multiple distinct filters in a dashboard (similar to the filter value). I think the only additional consideration is that right now when you create a chart, it automatically adds the dataset's main dttm column to the filter as a "placeholder", so that dashboard time range filters are applied to it. If it becomes possible to control/differentiate the column in the filter configuration, we would just need to make sure that this filter doesn't get applied to the "placeholder columns" too.

jkryanchou:
I wish this feature should be released as soon as it possible.

PR #30030, opened 2024-08-27
chore(docs): docker instructions use `docker compose` instead of the deprecated `docker-compose`

rusackas:
### SUMMARY
`docker-compose` isn't included in the newest installations of Docker, and is on the path toward deprecation, apparently. 

This PR makes the relevant changes to move us into a post-`docker-compose` world, where we now type `docker compose`. 

I will now close all 49 open Issues that mention `docker-compose`. Kidding!!!

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #30144, opened 2024-09-04
docs: document how docker-compose-image-tag requires -dev suffixed images

mistercrunch:
Realized [here](https://github.com/apache/superset/discussions/29999#discussioncomment-10490167) that `docker-compose-image-tag.yml` requires a -dev suffixed image to work as expected. Updating default TAG to latest-dev and related docs.

Side mission: removing warning related to version key in `docker-compose*.*` that are not used/supported anymore


sfirke (approved):
I think this will clarify things for a lot of people!  I have a few suggestions for clarity and to move from "docker-compose" to "docker compose" as in #30030

PR #30156, opened 2024-09-04
fix: FacePile is requesting avatars when SLACK_ENABLE_AVATARS is false

michael-s-molina:
### SUMMARY
https://github.com/apache/superset/pull/27849 introduced the Slack Avatar integration which augments the `FacePile` component with avatar images. This PR fixes an issue where the avatar images were being requested even when `SLACK_ENABLE_AVATARS` is `False`, generating unnecessary requests.

### TESTING INSTRUCTIONS
Check that avatar images are not requested if `SLACK_ENABLE_AVATARS` is `False`.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


villebro (approved):
LGTM, thanks for the fix

mistercrunch (approved):
Thanks for the improvement, this is better than the solution I had -> the backend would disregard the call and return a 204 when the flag was disabled https://github.com/apache/superset/blob/master/superset/views/users/api.py#L147-L164

Better to not call at all in the first place though! I didn't know an easy way to get access to the FF from the component...

PR #26791, opened 2024-01-24
fix(sqllab): flaky json explore modal due to over-rendering

justinpark:
### SUMMARY
This commit reopens #26156, which resolves the issue detailed in https://github.com/apache/superset/issues/26276 by excluding the changes in 'QueryAutoRefresh/index.tsx' and 'reducers/sqlLab.js'.
This commit also reverses the changes introduced in #18934. Now that the issue of overrendering due to redux has been resolved, it also effectively addresses the problem outlined in issue #26755.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:

https://github.com/apache/superset/assets/1392866/8d81ba78-7142-438c-8c5c-b5049ef6ea94

After:

https://github.com/apache/superset/assets/1392866/c2023683-47f1-4a45-97c0-c14e4d7b6469

### TESTING INSTRUCTIONS

Go to SQL Lab and run a query contains a json blob
Select a table in order to execute a table preview query
Click `Results` tab before the table preview query ends and then click the JSON column to open the JSON modal
Check the JSON modal persists while the async query status is updating

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #30171, opened 2024-09-05
fix(sqllab): Skip AceEditor in inactive tabs

justinpark:
### SUMMARY
Since `destroyInactiveTabPane` is disabled for improving rendering perf on switching tabs in [#26791](https://github.com/apache/superset/pull/26791/files#diff-f62a410d38c860c568090a5f5c0f0a3e2a3f621162ec665fa35ce17977cc34b0L284), I have confirmed that the Ace Editor was monitoring and responding to events even on inactive tabs.
As a result, race conditions related to the cursor position were occurring when switching between tabs. This commit addresses these issues by modifying the behavior so that, similar to `destroyInactiveTabPane`, the Ace Editor component is excluded when a tab becomes inactive.

Before:

https://github.com/user-attachments/assets/57f06042-dfb4-467a-a305-f6f899c643b4

Active:

https://github.com/user-attachments/assets/98c1ead8-7531-452f-b0c8-26ebbde0751f

### TESTING INSTRUCTIONS
Specs are added

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
+++ b/superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
@@ -880,15 +880,17 @@ const SqlEditor: FC<Props> = ({
               startQuery={startQuery}
             />
           )}
-          <AceEditorWrapper
-            autocomplete={autocompleteEnabled}
-            onBlur={onSqlChanged}
-            onChange={onSqlChanged}
-            queryEditorId={queryEditor.id}
-            onCursorPositionChange={handleCursorPositionChange}
-            height={`${aceEditorHeight}px`}
-            hotkeys={hotkeys}
-          />
+          {isActive && (
+            <AceEditorWrapper
+              autocomplete={autocompleteEnabled}
+              onBlur={onSqlChanged}
+              onChange={onSqlChanged}
+              queryEditorId={queryEditor.id}
+              onCursorPositionChange={handleCursorPositionChange}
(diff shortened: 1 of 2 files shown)


superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
```diff
@@ -880,15 +880,17 @@ const SqlEditor: FC<Props> = ({
...
-          <AceEditorWrapper
-            autocomplete={autocompleteEnabled}
-            onBlur={onSqlChanged}
-            onChange={onSqlChanged}
-            queryEditorId={queryEditor.id}
-            onCursorPositionChange={handleCursorPositionChange}
-            height={`${aceEditorHeight}px`}
-            hotkeys={hotkeys}
-          />
+          {isActive && (
```
michael-s-molina, line 883: @justinpark Isn't possible to just make the ace editor stop listening to events when inactive instead of removing it from the DOM? That way you would retain the previous performance gains.

justinpark: @michael-s-molina That approach involves many aspects that need to be checked, such as regression, making it challenging to implement as a hotfix.
It would be better to revisit and resolve the issue related to the performance of the Ace Editor (if we feel downgrade) after addressing the current problem.

michael-s-molina (approved):

PR #30176, opened 2024-09-06
fix: Disable cross filtering on charts with no dimensions

### LINKED ISSUE #30056: Cross-filtering is broken for charts that don't have dimensions
### Bug description

If you apply a cross-filter to a chart that does not have a dimension, the filter-indication will appear but data is not filtered and there's no entry for the cross-filter in the left panel.

### How to reproduce the bug

1. Apply a cross-filter to a chart that does not have dimensions
2. Check the filter indicator in another chart
3. Check that the data in the other chart is not filtered
4. Check that there's no entry for the cross-filter in the left panel

### Screenshots/recordings

https://github.com/user-attachments/assets/d1f06b28-b89a-45c8-ae54-65f1287f12ec

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

Also present in 4.1.0 RC2.

### Checklist

kgabryje:
### SUMMARY
Cross filtering feature is designed to apply filters based on a dimension that user clicked in the source chart.
In some chart plugins, when the chart has no dimensions (only metric series), click on a metric would emit an empty cross filter, providing confusing user experience.
This PR disables cross filtering on charts that have no dimensions set.

Additionally, some charts were marked as compatible with cross filtering feature (`behaviors: [Behavior.InteractiveChart]`), but were in fact not implementing the feature. In those charts I removed the behavior and added a `TODO` comment to implement cross filtering.

### TESTING INSTRUCTIONS
1. Go to "Featured Charts" dashboard.
2. Verify that charts with dimensions (e.g. Area, Box Plot, Gauge, Pie) still emit cross filters as before
3. Verify that charts with no dimensions (e.g. Mixed) do not emit cross filters - nothing happens on click, cross filtering option in context menu is disabled
4. Add more charts with no dimensions to the dashboard, verify that they don't emit cross filters after clicking on a metric series.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/30056
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/index.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Bubble/index.ts
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { Behavior, ChartMetadata, ChartPlugin, t } from '@superset-ui/core';
+import { ChartMetadata, ChartPlugin, t } from '@superset-ui/core';
 import thumbnail from './images/thumbnail.png';
 import transformProps from './transformProps';
 import buildQuery from './buildQuery';
@@ -25,6 +25,7 @@ import example1 from './images/example1.png';
 import example2 from './images/example2.png';
 import { EchartsBubbleChartProps, EchartsBubbleFormData } from './types';
 
+// TODO: Implement cross filtering
 export default class EchartsBubbleChartPlugin extends ChartPlugin<
   EchartsBubbleFormData,
   EchartsBubbleChartProps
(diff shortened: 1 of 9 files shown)


michael-s-molina (approved):
@kgabryje @yousoph I have no problem with cutting the scope down for 4.1.0 and simply disabling cross filter for charts with no dimensions. Keep in mind though that this should not be the final solution as our x-axis supports categorical, temporal and numeric types which could be used in a cross filter. One simple example is when a user creates a bar chart and the x-axis is categorical. By the way, Drill By works for this scenario so I would expect cross filters to work too.

https://github.com/user-attachments/assets/fb4ec459-3184-4a77-900e-5fa41a031005

PR #30177, opened 2024-09-06
feat: `is_mutating` method

betodealmeida:
### SUMMARY

Part of the work on [SIP-117](https://github.com/apache/superset/issues/26786), this PR introduces a `has_mutation()` method to `SQLScript`, as well as a `is_mutating()` method to `SQLStatement`, in order to check if the SQL contains any DML.

This also fixes a bug where this SQL is flagged incorrectly as DML:

```sql
with source as (
  select 1 as one
)
select * from source
```

N/A

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -452,6 +452,14 @@ class BaseSQLStatement(Generic[InternalRepresentation]):
         """
         raise NotImplementedError()
 
+    def is_dml(self) -> bool:
+        """
+        Check if the statement is DML.
+
+        :return: True if the statement is DML
+        """
+        raise NotImplementedError()
+
     def __str__(self) -> str:
         return self.format()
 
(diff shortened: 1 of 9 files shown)


superset/sql_parse.py
```diff
@@ -522,6 +530,43 @@ def _extract_tables_from_statement(
...
+        """
+        for node in self._parsed.walk():
+            if isinstance(
+                node,
+                (
+                    exp.Insert,
+                    exp.Update,
+                    exp.Delete,
+                    exp.Merge,
+                    exp.Create,
```
mistercrunch, line 547: technically this is DDL not DML, wondering if `is_mutating` would be a better name?

mistercrunch (approved):
LGTM outside of the naming concern, pre-approving assuming the name change

superset/sql_parse.py
betodealmeida: @mistercrunch do you think we should also rename the DB modal? It currently says "Allow DML".

mistercrunch: yes I would. Here's what it says currently ->  Screenshot 2024-09-09 at 9 11 58 AM

`Allow DML and DDL` - could be good to unpack the acronyms in the tooltip too. "Allow the execution of DDL (Data Definition Language, ie CREATE, DROP, TRUNCATE, ...) and DML (Data Modification Language, ie INSERT, UPDATE, DELETE, ...) {...}"

PR #30180, opened 2024-09-06
fix: filters panel broken due to tabs scroll

### LINKED ISSUE #30058: Native filters panel break due to tabs scroll
### Bug description

When a dashboard has inner tabs that contain scroll, the calculated height of the native filters panel is breaking when changing between tabs with different heights.

### How to reproduce the bug

1. Create a dashboard with 2 inner tabs
2. In one tab, add a bunch of charts so that it needs scrolling
3. In the other tab, add a few charts so it does not need scrolling
4. Create a native filter
5. Scroll to the bottom on the first tab
6. Select the second tab
7. Check that the dashboard height is not calculated correctly and presents a scroll
8. Scroll to the bottom and see the native filters panel break

### Screenshots/recordings

https://github.com/user-attachments/assets/e491d09c-a81d-4576-bd8d-37ef9779932c

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

Also present in 4.1.0 RC2

### Checklist

justinpark:
### SUMMARY
Fixes #30058

The EChart tooltip is hidden using `visibility: hidden` with `opacity: 0` after it disappears, which causes tooltips generated from the previous tab to be recognized as scroll items when configuring the layout for the next tab, resulting in unnecessary scrollbars.

> display:none means that the tag in question will not appear on the page at all (although you can still interact with it through the dom). There will be no space allocated for it between the other tags.
>
> visibility:hidden means that unlike display:none, the tag is not visible, but space is allocated for it on the page. The tag is rendered, it just isn't seen on the page.

This commit addresses the issue by using a CSS hack to change the visibility hidden tooltip to `display: none`, effectively removing it and resolving the scrollbar problem.

Before:

https://github.com/user-attachments/assets/315ac940-47dc-4c0d-b55b-4b16e2237195

After:

https://github.com/user-attachments/assets/8a8e3d1f-e4bb-4022-8982-63229293576f

### TESTING INSTRUCTIONS

1. Create a dashboard with 2 inner tabs
2. In one tab, add a bunch of charts so that it needs scrolling
3. In the other tab, add a few charts so it does not need scrolling
4. Create a native filter
5. Scroll to the bottom on the first tab and **mouse over a chart to show a tooltip**
6. Select the second tab
7. Check that the dashboard height is not calculated correctly and presents a scroll
8. Scroll to the bottom and see the native filters panel break

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/GlobalStyles.tsx
+++ b/superset-frontend/src/GlobalStyles.tsx
@@ -34,6 +34,11 @@ export const GlobalStyles = () => (
       th {
         font-weight: ${theme.typography.weights.bold};
       }
+      // CSS hack to resolve the issue caused by the invisible echart tooltip on
+      // https://github.com/apache/superset/issues/30058
+      .echarts-tooltip[style*='visibility: hidden'] {
+        display: none !important;
+      }
       // TODO: Remove when on Ant Design 5.
       // Check src/components/Modal for more info.
       .modal-functions-ok-button {
--- a/superset-frontend/plugins/plugin-chart-echarts/src/utils/tooltip.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/utils/tooltip.ts
@@ -25,6 +25,8 @@ export function getDefaultTooltip(refs: Refs) {
   return {
     appendToBody: true,
     borderColor: 'transparent',
+    // CSS hack applied on this class to resolve https://github.com/apache/superset/issues/30058
+    className: 'echarts-tooltip',
     position: (
       canvasMousePos: [number, number],
       params: CallbackDataParams,


justinpark:
@michael-s-molina I also confirmed that this bug exists in 4.0 as well.

superset-frontend/src/GlobalStyles.tsx
```diff
@@ -34,6 +34,11 @@ export const GlobalStyles = () => (
       th {
         font-weight: ${theme.typography.weights.bold};
       }
+      // CSS hack to resolve the issue caused by the invisible echart tooltip on
+      // https://github.com/apache/superset/issues/30058
+      .echarts-tooltip[style*='visibility: hidden'] {
+        display: none !important;
+      }
```
rusackas, line 41: Couple of options if you want to optimize further:

1) we can probably narrow this from Superset global to being part of the ECharts plugin(s), perhaps using [Emotion Globals](https://emotion.sh/docs/globals).
2) This might be a good excuse to open a PR on ECharts and fix it at the root ;)

 Either way, I don't see this as a blocker for now. Things are lookin' good!

rusackas (approved):
LGTM... we can probably get fancy with how/where we do this, but this works well for today!

superset-frontend/src/GlobalStyles.tsx
justinpark: Approach 2 is, of course, the fundamental solution, but I decided to leave it as a long-term solution considering the time gap involved in merging the patch and testing Superset compatibility with the corresponding version. (I plan to propose a related PR to echart later.)
Initially, I considered Approach 1, but I had concerns that adding a dependency on @emotion/react to the plugins' npm package might open the possibility for other custom plugins to apply custom global CSS. Therefore, I opted for this approach.

PR #30196, opened 2024-09-09
build(packages): npm build/publish improvements. Making packages publishable again.

rusackas:
### SUMMARY

The plugin scripts were a little out of whack... they were doing lots of problematic things like running `npm run perge` before publishing, which strips the compiled module folders. Modules also couldn't be published with changelogs because the changelogs were too big, and lerna.json files weren't passing linting. This PR includes all the changes needed to get a module release out the door.

This does a bunch of random stuff:
• Builds plugins and packages in PROD mode rather than DEV mode
• Fixes up a bunch of typescript issues that cropped up in dev mode
• Fixes a bunch of dependency and import issues that were caught by TS and ESlint
• Updates build and publication scripts to work as expected. They should also have more obvious tag names going forward.
• Adds lerna-generated changelogs, which _should_ now include ASF license boilerplate going forward
• All kinds of little corrections/overrides/nudges to get things to build and pass correctly.

New versions of packages have been built mid-flight. When this is merged, I'll bump them again and dial in any automation as needed in a subsequent PR.

### TESTING INSTRUCTIONS

`npm run plugins:build` should work
`npm run build` should (still) work
all CI passes
various plugin versioning and publishing commands all work (tested!)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/lerna.json
```diff
@@ -19,4 +23,4 @@
     }
   },
   "$schema": "node_modules/lerna/schemas/lerna-schema.json"
-}
+}
```
villebro, line 26: The newline is missing here

villebro (approved):
Thanks for cleaning these up! Minor nit on one of the config files, but other than that looking good!

superset-frontend/lerna.json
rusackas: Right... these are generated by lerna. They were getting caught by the pre-commit, but then it mods the files, which means you can't do the release... it's an annoying catch-22. I'll try to find a way around it in this or a subsequent PR. 

Actually, now that I look at CI, there's a similar thing where I need to prepend the ASF license to CHANGELOG.md files... so I'll have to add a script for that at the very least to get the PR through.

rusackas, line 26:
```suggestion
}

```

mistercrunch (approved):
LGTM

PR #30201, opened 2024-09-09
chore: linkinator to NOT validate links in INTHEWILD.md

mistercrunch:
Linkinator makes sure that our docs points to valid URL. Lately I've seen 2-3 instances of CI failing because a link in INTHEWILD.md is temporarily or permanently broken. I'm hoping this doesn't affect CI as this is disruptive for daily work


rusackas (changes requested):
Call me stubborn, but I'd rather not disable this here.

The check is not blocking CI, even if it fails (you can still merge).

It's been useful in keeping the files in good condition, and the current failures are legit.

PR #30222, opened 2024-09-10
fix(uploads): respect db engine spec's supports_multivalues_insert value for file uploads & enable multi-insert for MSSQL

sfirke:
### SUMMARY
This PR makes two changes to enable the setting of `method = 'multi'` when the Pandas function `.to_sql` is called to write a file upload to a data warehouse.  This speeds up file uploads by 15x.  See discussion https://github.com/apache/superset/discussions/30152 for background and discussion.

**Changes:**
1. The base method `df_to_sql` does not currently access the `supports_multivalues_insert` attribute when it is set in a Superset EngineSpec.  Instead it relies on this attribute being set in the SQLAlchemy implementation for that dialect.  After this PR, Superset will check both of these places for the attribute.
2. Sets the attribute `supports_multivalues_insert` to `True` for Microsoft SQL Server (MSSQL), which has supported this behavior since SQL Server 2008.

Thanks to @nytai for advising.

### TESTING INSTRUCTIONS
Check that uploading a flat file to a data warehouse succeeds.  I tested it for MSSQL.  This PR was written to not reduce any existing functionality (there's an `or` with the previous condition) but someone could check on another data warehouse too.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -1283,7 +1283,9 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
             catalog=table.catalog,
             schema=table.schema,
         ) as engine:
-            if getattr(engine.dialect, 'supports_multivalues_insert', False) or getattr(cls, 'supports_multivalues_insert', False):
+            if getattr(engine.dialect, "supports_multivalues_insert", False) or getattr(
+                cls, "supports_multivalues_insert", False
+            ):
                 to_sql_kwargs["method"] = "multi"
             df.to_sql(con=engine, **to_sql_kwargs)
 
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -347,6 +347,7 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
     # Does database support join-free timeslot grouping
     time_groupby_inline = False
     limit_method = LimitMethod.FORCE_LIMIT
+    supports_multivalues_insert = False
     allows_joins = True
     allows_subqueries = True
     allows_alias_in_select = True
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -1284,9 +1284,7 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
             catalog=table.catalog,
             schema=table.schema,
         ) as engine:
-            if getattr(engine.dialect, "supports_multivalues_insert", False) or getattr(
-                cls, "supports_multivalues_insert", False
-            ):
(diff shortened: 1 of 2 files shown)


superset/db_engine_specs/mssql.py
```diff
@@ -53,6 +53,7 @@ class MssqlEngineSpec(BaseEngineSpec):
     max_column_name_length = 128
     allows_cte_in_subquery = False
     allow_limit_clause = False
+    supports_multivalues_insert = True
```
villebro, line 56: Please also add `supports_multivalues_insert` to `BaseEngineSpec`. I assume it will need to be disabled by default.

villebro (changes requested):
LGTM, but all db engine spec class properites need to be also added to `BaseEngineSpec` with conservative defaults.

superset/db_engine_specs/mssql.py
sfirke: Done in [217ad1f](https://github.com/apache/superset/pull/30222/commits/217ad1fb3378297b11ebe64862181d7424ea0677). Agreed with setting it to False by default: I expect many DB types support it, but since it's not currently defined in BaseEngineSpec at all we're not reducing any existing functionality with this, and setting it to True might break uploads for some DBs that don't support it.

superset/db_engine_specs/base.py
```diff
@@ -1283,9 +1284,10 @@ def df_to_sql(
             catalog=table.catalog,
             schema=table.schema,
         ) as engine:
-            if engine.dialect.supports_multivalues_insert:
+            if getattr(engine.dialect, "supports_multivalues_insert", False) or getattr(
+                cls, "supports_multivalues_insert", False
+            ):
```
villebro, line 1289: I think you can simplify more (as we've added `supports_multivalues_insert` to `BaseEngineSpec`, it'll always be defined, so no need to `getattr`):
```suggestion
            if getattr(engine.dialect, "supports_multivalues_insert", False) or cls.supports_multivalues_insert:
```
Also, something worth checking, do we really need to `getattr(engine.dialect, "supports_multivalues_insert", False)`, or would `engine.dialect.supports_multivalues_insert` be sufficient (I kinda assume it'll always be defined)? I can help checking this if needed.

villebro (commented):
Generally LGTM but could be simplified more

superset/db_engine_specs/base.py
sfirke: Agreed that we can go back to `engine.dialect.supports_multivalues_insert` -- that's what it was before this PR so I feel good simplifying it to that.

villebro (approved):
Nice, really stoked about this PR, LGTM! 🚀

PR #30246, opened 2024-09-11
fix: additional lower case text changes

sadpandajoe:
### SUMMARY

Additional missed text from: https://github.com/apache/superset/pull/29673. A few edits here:
* In the dataset editor, changing `Certified By` to `Certified by` in the `Columns` and `Calculated columns` screens
* In the SQL menu, changing from `SQL History` to `SQL history`
* In the SQL menu, changing from `Saved Queries` to `Saved queries`
* In the main menu, changing from `Annotation Layers` to `Annotation layers`
* On a blank tags page, change from `Create a new Tag` to `Create a new tag`

#### Before
Screenshot 2024-09-12 at 9 38 46 AM
Screenshot 2024-09-12 at 9 38 30 AM
Screenshot 2024-09-12 at 8 59 47 AM

#### After
Screenshot 2024-09-12 at 8 59 24 AM
Screenshot 2024-09-12 at 8 59 33 AM
Screenshot 2024-09-12 at 9 36 49 AM
Screenshot 2024-09-12 at 11 19 06 AM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/sections/annotationsAndLayers.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/sections/annotationsAndLayers.tsx
@@ -33,7 +33,7 @@ export const annotationsAndLayersControls: ControlPanelSectionConfig = {
           type: 'AnnotationLayerControl',
           label: '',
           default: annotationLayers,
-          description: t('Annotation Layers'),
+          description: t('Annotation layers'),
           renderTrigger: false,
         },
       },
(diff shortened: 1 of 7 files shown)


yousoph (approved):

rusackas:
Heya! I'm tempted to block on this, but I'll just post a comment instead. 

When changing capitalizations like this, it'll break translations for all of these strings. Can you do a global find/replace that captures these in the .po/.pot files?

If we don't make that change, all the translations will disappear the next time strings are extracted to the .pot files, and .po files get updated.

I'm also adding this to the Town Hall agenda... we need to find a way to capture/enforce this at the PR level or in CI.

rusackas:
Also, that's weird: 
image
¯\\\_(ツ)_/¯

rusackas (changes requested):
Might as well fix the `.po` files containing these strings (and any obvious translated strings next to them), so that we don't regress translations.

Issue #30061, opened 2024-08-29
Cross-filtering is reloading unaffected charts


michael-s-molina:
### Bug description

When a user applies a cross-filter, charts that belong to different datasets and are not affected by the cross filter are being reloaded. Each reloaded chart fires a server-side request for data, decreasing the dashboard's performance and wasting resources.

### How to reproduce the bug

1 - Create a dashboard with charts from multiple datasets
2 - Apply a cross filter on a chart
3 - Check that charts that belong to different datasets are reloaded

### Screenshots/recordings

https://github.com/user-attachments/assets/12ec3fac-177c-45c5-97e3-6d0b453a6d51

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

Also present in 4.1.0 RC2

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

PR #30252, opened 2024-09-12
fix(CrossFilters): Do not reload unrelated filters in global scope

geido:
### SUMMARY
This PR partially accounts for the issue reported on #30061. However, more work is required so that when a scope is applied the frontend can check columns / calculated columns across different datasets to see whether the query should be sent over the backend at all. Working on a follow-up PRs that should solve the issue for both native and cross filters.

### BEFORE

https://github.com/user-attachments/assets/7b91a4a8-2c14-4bfb-84f4-94d82d02ec84

### AFTER

https://github.com/user-attachments/assets/6a85fc52-2c13-4cf4-977f-f9623c0a55f3

### TESTING INSTRUCTIONS
1. In global scope a cross filter should only affect charts of the same dataset

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/util/crossFilters.ts
+++ b/superset-frontend/src/dashboard/util/crossFilters.ts
@@ -52,6 +52,20 @@ export const getCrossFiltersConfiguration = (
     return undefined;
   }
 
+  const chartsByDataSource: Record<string, number[]> = Object.values(
+    charts,
+  ).reduce((acc: Record<string, number[]>, chart) => {
+    if (!chart.form_data) {
+      return acc;
+    }
+    const { datasource } = chart.form_data;
+    if (!acc[datasource]) {
+      acc[datasource] = [];
+    }
+    acc[datasource].push(chart.id);
+    return acc;
+  }, {});
+
   const globalChartConfiguration = metadata.global_chart_configuration?.scope
     ? {
         scope: metadata.global_chart_configuration.scope,
@@ -97,10 +111,13 @@ export const getCrossFiltersConfiguration = (
           },
         };
       }
+      const chartDataSource = charts[chartId].form_data.datasource;
       chartConfiguration[chartId].crossFilters.chartsInScope =
         isCrossFilterScopeGlobal(chartConfiguration[chartId].crossFilters.scope)
           ? globalChartConfiguration.chartsInScope.filter(
-              id => id !== Number(chartId),
+              id =>
+                id !== Number(chartId) &&
+                chartsByDataSource[chartDataSource].includes(id),
             )
           : getChartIdsInFilterScope(
               chartConfiguration[chartId].crossFilters.scope,
--- a/superset-frontend/src/dashboard/util/crossFilters.ts
+++ b/superset-frontend/src/dashboard/util/crossFilters.ts
@@ -117,7 +117,7 @@ export const getCrossFiltersConfiguration = (
           ? globalChartConfiguration.chartsInScope.filter(
               id =>
                 id !== Number(chartId) &&
-                chartsByDataSource[chartDataSource].includes(id),
+                chartsByDataSource[chartDataSource]?.has(id),
             )
           : getChartIdsInFilterScope(
               chartConfiguration[chartId].crossFilters.scope,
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
👍🏼

superset-frontend/src/dashboard/util/crossFilters.ts
```diff
@@ -52,6 +52,20 @@ export const getCrossFiltersConfiguration = (
...
+    charts,
+  ).reduce((acc: Record<string, number[]>, chart) => {
+    if (!chart.form_data) {
+      return acc;
+    }
+    const { datasource } = chart.form_data;
+    if (!acc[datasource]) {
+      acc[datasource] = [];
+    }
+    acc[datasource].push(chart.id);
```
justinpark, line 65: Using a set would likely make the algorithm more efficient.
```suggestion
    if (!acc[datasource]) {
      acc[datasource] = new Set();
    }
    acc[datasource].add(chart.id);
```

superset-frontend/src/dashboard/util/crossFilters.ts
```diff
@@ -97,10 +111,13 @@ export const getCrossFiltersConfiguration = (
...
         };
       }
+      const chartDataSource = charts[chartId].form_data.datasource;
       chartConfiguration[chartId].crossFilters.chartsInScope =
         isCrossFilterScopeGlobal(chartConfiguration[chartId].crossFilters.scope)
           ? globalChartConfiguration.chartsInScope.filter(
-              id => id !== Number(chartId),
+              id =>
+                id !== Number(chartId) &&
+                chartsByDataSource[chartDataSource].includes(id),
```
justinpark, line 120: Including optional chaining would likely help prevent undefined errors.

```suggestion
                chartsByDataSource[chartDataSource]?.has(id),
```

michael-s-molina:
> This PR partially accounts for the issue reported on https://github.com/apache/superset/issues/30061. However, more work is required so that when a scope is applied the frontend can check columns / calculated columns across different datasets to see whether the query should be sent over the backed at all. Working on a follow-up PRs that should solve the issue for both native and cross filters.

@geido Do you have a link for the follow-up that will fix the cross filter issue?

geido:
> > This PR partially accounts for the issue reported on https://github.com/apache/superset/issues/30061. However, more work is required so that when a scope is applied the frontend can check columns / calculated columns across different datasets to see whether the query should be sent over the backed at all. Working on a follow-up PRs that should solve the issue for both native and cross filters.
> 
> @geido Do you have a link for the follow-up that will fix the cross filter issue?

Still working on it

fmannhardt:
@geido I am testing the 4.1 branch and realized that our cross filters are not working anymore. Do I read this PR right in that it would entirely disable cross filtering if the data source is not the same, even if shared columns exist? And that this is fixed, also for columns not actually used in the chart, with: https://github.com/apache/superset/pull/30438

geido:
> @geido I am testing the 4.1 branch and realized that our cross filters are not working anymore. Do I read this PR right in that it would entirely disable cross filtering if the data source is not the same, even if shared columns exist? And that this is fixed, also for columns not actually used in the chart, with: #30438

That is a good point @fmannhardt . I updated https://github.com/apache/superset/pull/30438 to make sure we are targeting all possible common columns in all datasources. Let me know if that makes sense to you now.

fmannhardt:
When testing including https://github.com/apache/superset/pull/30438 I still do not see the chart being refreshed. The query is correct and also when looking at the result table it is correct but the chart visualization itself does not show up refreshed.

geido:
> When testing including #30438 I still do not see the chart being refreshed. The query is correct and also when looking at the result table it is correct but the chart visualization itself does not show up refreshed.

@fmannhardt would you mind providing a test case to reproduce this issue on my side, please? That would help.

fmannhardt:
Ok. I will try to provide it. Maybe it had to do with using JINJA templating in a filtering statement.

fmannhardt:
I have tried to reproduce with the latest changes on the other PR and the issue is gone. Thanks.

PR #30253, opened 2024-09-12
chore(build): remove extraneous prettier step in superset-frontend CI

hainenber:
chore(build): remove extraneous prettier step in superset-frontend CI

### SUMMARY

I noticed there are duplicate CI steps in `superset-frontend` and `pre-commit` jobs. Since `pre-commit` job is ran regardless of all changes, I think it's better to remove the duplication in `superset-frontend`.

Unless there are missing historical contexts I'm not aware of ...

N/A

### TESTING INSTRUCTIONS

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/superset-frontend.yml
+++ b/.github/workflows/superset-frontend.yml
@@ -49,11 +49,6 @@ jobs:
         working-directory: ./superset-frontend
         run: |
           npm run type
-      - name: prettier
-        if: steps.check.outputs.frontend
-        working-directory: ./superset-frontend
-        run: |
-          npm run prettier-check
       - name: Build plugins packages
         if: steps.check.outputs.frontend
         working-directory: ./superset-frontend


mistercrunch (approved):
Thanks for the contribution!

mistercrunch:
Looks like we may be able to drop the task in `.github/workflows/superset-websocket.yml` as well since I think it's covered by the [pre-commit](https://github.com/apache/superset/blob/master/.pre-commit-config.yaml#L58-L65) checks (?)

hainenber:
Prettier task in `supserset-websocket.yml` isn't duplicate of the one removed and the pre-commit check though. They're different in their working directories 🤓 

-  Prettier in `pre-commit` and `superset-frontend` CI
image
image


- Prettier in `superset-websocket.yml`
image

mistercrunch:
Yes I notice the setup isn't identical but I think it's effectively the same. Let's merge this as a net positive and can revisit this other potential duplication later one.

Issue #30238, opened 2024-09-11
Bar Chart sorting does not appear to work


HJelinkova:
### Bug description

It is imposible to change the layout of the bar chart by using the "Sort by" function. 
The ASC/DESC checkbox changes the query but the chart stays unaffected. 

Same issues previously reported: #15260, #15318

### How to reproduce the bug

1. Create a bar chart.
2. In the "Data" tab find the "Sort by" section.
3. Choose the column and the aggregate by which you want to sort the data by.
4. Tick/Untick the "Sort descending" checkbox.

### Screenshots/recordings

![Bar chart bug (1)](https://github.com/user-attachments/assets/b9943f63-e1aa-460c-9e78-94385cb645c3)
![Bar chart bug (2) ](https://github.com/user-attachments/assets/ad42d47b-7bd6-4ac7-b942-eba2073ce2b3)
![Bar chart bug - query (1)](https://github.com/user-attachments/assets/f5dc5e75-5e2f-4364-aaca-b2376f4d1d4b)
![Bar chart bug - query (2)](https://github.com/user-attachments/assets/62227fe1-939c-467f-8d5c-a2fe0588349f)

### Superset version

4.0.2

### Python version

Python 3.10.14

### Node version

NODE version 20

### Browser

Chrome

### Additional context

Note: Superset installed using docker. 

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [ ] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

sfirke:
If you read the mouseover tooltip for SORT BY you'll see that's for something else.  That should probably be labeled as "DATA LIMIT SORT BY".  What you need is X-AXIS SORT BY, which switches to Y-AXIS SORT BY when a bar chart is switched to horizontal orientation as it appears yours is.  I don't know why you don't see that option in your screenshot; are you positive you're running Superset 4.0.2?

Here's what it looks like for me.  With the default sorting:
![image](https://github.com/user-attachments/assets/1da3d002-dd28-4931-9f8c-bd27b35cf9e5)

Sorting by a specific metric changes the bars' order:
![image](https://github.com/user-attachments/assets/dfc42a08-2f16-4ea2-8a22-f4d48b2a5433)

HJelinkova:
[...]

PR #30254, opened 2024-09-12
style(explore): clarify ambiguously named "sort by" field

sfirke:
### SUMMARY
Users sometimes don't understand the field with the ambiguous name "Sort By".  For instance, in https://github.com/apache/superset/issues/30238 someone expected it would govern the order of the bars along the axis.  Other times they think it will sort the order of series on a dimension field.  These two kinds of sorting have more clearly-named fields; this "Sort By" sorts the underlying data which is relevant only in the case where a row limit or series limit is reached.

This PR renames that field from "Sort By" to "Data Limit Sort By".

**BEFORE**
image

**AFTER**
image


rusackas (approved):
LGTM. It's an improvement for sure, but I wonder if there are any other label choices... the wording feels awkward even if clearer :) I'm a little stumped ¯\\\_(ツ)_/¯

sfirke:
I hear you @rusackas, I went back and forth... of the Copilot suggestions, the other contenders I liked were "Pre-Render Sort By" and "Truncation Sort By" -- do you like one of those better?

michael-s-molina:
I'm also not sure about Data Limit Sort By... Maybe `Data Sort By` or `Query Sort By`? I'm ok with you don't find a better name for it.

sfirke:
I like `Query Sort By`, that makes it clear it's a backend thing.

rusackas:
> I like `Query Sort By`, that makes it clear it's a backend thing.

Sort Query By?

sfirke:
I updated the label text to `Sort query by`.

And I updated the tooltip.  This sorts the query result which affects data download + drilling so doesn't matter only in the case that a limit is reached -- another reason that your suggestions are better than the original idea.

sfirke:
This one Cypress test keeps failing with a vague error message. Is there a way to tell if there is a Cypress test that is failing because this label changed or if it's just flaky?

PR #30255, opened 2024-09-12
chore(scripts): purge node_modules folder on `npm prune`

rusackas:
The npm prune script wasn't purging node modules folders, which (AFAIK) should be nuked from time to time. `npm install` should be run when building/publishing packages, or building/installing superset.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/package.json
+++ b/superset-frontend/package.json
@@ -65,7 +65,7 @@
     "prettier": "npm run _prettier -- --write",
     "prettier-check": "npm run _prettier -- --check",
     "prod": "npm run build",
-    "prune": "rm -rf ./{packages,plugins}/*/{lib,esm,tsconfig.tsbuildinfo,package-lock.json}",
+    "prune": "rm -rf ./{packages,plugins}/*/{node_modules,lib,esm,tsconfig.tsbuildinfo,package-lock.json}",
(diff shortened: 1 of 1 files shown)


mistercrunch:
Just a thought but could be good to nuke `.temp_cache/` too, mine got to 3GB

```bash
$ du -sh .temp_cache/
2.9G    .temp_cache/
```
after nuking + a rebuild it's about

rusackas:
@mistercrunch done (finally) :)

mistercrunch:
Wondering if this `npm run prune` is ever viable given that it removes the `package-lock.json`, doesn't that leas to upgrading all packages/sub-packages in a way that's too-much-all-at-once?

rusackas:
> Wondering if this `npm run prune` is ever viable given that it removes the `package-lock.json`, doesn't that lead to upgrading all packages/sub-packages in a way that's too-much-all-at-once?

Nope... there shouldn't ever be a need to have package-lock files in packages/plugins. Running `npm install` from the root `superset-frontend` directory installs them all for Superset. The package lock files in these plugins/packages aren't needed and in fact cause scripts to fail when building/deploying.

mistercrunch (approved):

PR #30262, opened 2024-09-12
chore: bump cypress to v 11

eschutho:
### SUMMARY

Per the cypress docs, this experimental flag should help with some of the chromium memory problems we've been seeing, but it needs at least v12 in order to do anything: 
https://docs.cypress.io/guides/references/error-messages#The-Chromium-Renderer-process-just-crashed

Bumping this version to 11. We'll do 12 in a separate PR
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/cypress-base/cypress.config.ts
```diff
@@ -26,9 +26,9 @@ export default eyesPlugin(
     defaultCommandTimeout: 8000,
     numTestsKeptInMemory: 0,
     experimentalFetchPolyfill: true,
+    experimentalMemoryManagement: true,
     requestTimeout: 10000,
     video: false,
-    videoUploadOnPasses: false,
```
villebro, line 31: Was this removal intentional? I don't see it mentioned in the docs or the description..

villebro (commented):
Nice, let's hope this fixes that flaky test! One comment on the removed config.

superset-frontend/cypress-base/cypress.config.ts
eschutho: Yeah, it was flagged by typescript that it no longer exists.

eschutho:
I checked to see if we had tried this flag before, and sure enough @rusackas had a PR that enabled it, and then removed it because it wasn't supported in v10. (they added support in v12). So I'm trying my luck to see how easily I can bump to 12. 🤞

superset-frontend/cypress-base/cypress.config.ts
villebro: Oh ok, that explains why it wasn't mentioned in the docs 😄

superset-frontend/cypress-base/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "devDependencies": {
     "@types/querystringify": "^2.0.0",
-    "cypress": "^10.11.0",
+    "cypress": "^11.2.0",
```
sadpandajoe, line 32: This is a bump to cypress 11 and not 12, right?

eschutho: Yeah, I was running tests on 11 and then was going to to 12, but I think I'll just stick to 11 in this PR.

villebro (approved):
LGTM, thanks for working on this ❤️

superset-frontend/cypress-base/package.json
sadpandajoe: Ok this looks good then

sadpandajoe (approved):

geido (approved):

rusackas (approved):
I think there's a way to upgrade with a GUI using `npx` if you're not already aware. I've been curious how well that works, maybe it'll be helpful for the next bumps.

eschutho:
> I think there's a way to upgrade with a GUI using `npx` if you're not already aware. I've been curious how well that works, maybe it'll be helpful for the next bumps.

Oh nice. I'll try that. There's also a PR already where someone did the difficult work of bumping to v13. I was thinking of reviving that PR.

PR #30266, opened 2024-09-13
feat: allow configuring an engine context manager

betodealmeida:
### SUMMARY

This PR introduces a configurable context manager that wraps the `create_engine` call to analytical databases. The intended use case is to configure the Oracle driver with tenant-specific settings in a multi-tenant environment, but there are many other potential use cases for this:

1. Custom logging
2. Supressing noisy logs from 3rd party database drivers
3. `chroot`ing to prevent 3rd party databases drivers from accessing the filesystem (maybe using [pychroot](https://pypi.org/project/pychroot/))

N/A

### TESTING INSTRUCTIONS

Added a unit test.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -33,10 +33,11 @@ import os
 import re
 import sys
 from collections import OrderedDict
+from contextlib import contextmanager
 from datetime import timedelta
 from email.mime.multipart import MIMEMultipart
 from importlib.resources import files
-from typing import Any, Callable, Literal, TYPE_CHECKING, TypedDict
+from typing import Any, Callable, Iterator, Literal, TYPE_CHECKING, TypedDict
 
 import click
 import pkg_resources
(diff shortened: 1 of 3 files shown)


superset/config.py
```diff
@@ -1142,16 +1143,18 @@ def CSV_TO_HIVE_UPLOAD_DIRECTORY_FUNC(  # pylint: disable=invalid-name
 # uploading CSVs will be stored.
 UPLOADED_CSV_HIVE_NAMESPACE: str | None = None
 
+
 # Function that computes the allowed schemas for the CSV uploads.
 # Allowed schemas will be a union of schemas_allowed_for_file_upload
 # db configuration and a result of this function.
+def allowed_schemas_for_csv_upload(  # pylint: disable=unused-argument
```
betodealmeida, line 1150: I rewrote this as a function since it's easier to read and removes the lambda complaint from the linter.

mistercrunch (approved):
Nice. Did grepping to check if there were some rogue `create_engine` calls not going through this method and it looked squeaky cleaned.

dpgaspar (approved):
Nice!

betodealmeida:
> Nice. Did grepping to check if there were some rogue `create_engine` calls not going through this method and it looked squeaky cleaned.

@mistercrunch yeah, when we implemented SSH tunneling we had to consolidate all the logic in a single place.

PR #30274, opened 2024-09-13
chore: move SLACK_ENABLE_AVATARS from config to feature flag

mistercrunch:
Needed to make SLACK_ENABLE_AVATARS more dynamic @ Preset so it's easy for us to turn this feature on/off and allow us for a progressive rollout.


docker-compose.yml
```diff
@@ -149,7 +149,14 @@ services:
       disable: true
 
   superset-node:
-    image: node:20
+    build:
```
mistercrunch, line 152: bycatch - somehow superset-node wasn't using the proper target, which explains some issues with `zstd` we've seen before, which gets installed at the os-level here. There's also the "memory snag" script and the DEV_MODE logic to skip the actual webpack build since we run `npm run dev` on startup.

mistercrunch: I thought I had caught/fixed that in some other PR, but guessing it got lost in the shuffle.

rusackas (changes requested):
Please make sure it gets an entry in `FEATURE_FLAGS.md` as well :) 

Pinging @eschutho for review since she's been handling most of the Slack integration scope. 

Also, this might bring up the age-old debate of config vs. feature flags. I'm OK with Feature Flags for cases like this where platforms (Launch Darkly, Split, etc) can hotwire them as needed. Others, however, feel that Feature Flags should be temporary as a development pattern before being eventually deprecated/removed. Sounds like this one is effectively permanent.

docker-compose.yml
kgabryje: Nice, thanks for fixing that

superset/config.py
```diff
@@ -1427,11 +1431,6 @@ def EMAIL_HEADER_MUTATOR(  # pylint: disable=invalid-name,unused-argument
 SLACK_API_TOKEN: Callable[[], str] | str | None = None
 SLACK_PROXY = None
 
-# Whether Superset should use Slack avatars for users.
-# If on, you'll want to add "https://avatars.slack-edge.com" to the list of allowed
-# domains in your TALISMAN_CONFIG
-SLACK_ENABLE_AVATARS = False
-
```
eschutho, line 1434: Since this hasn't been released in an official version, I think this breaking change is ok.

mistercrunch: Right, and I added a note in UPDATING.md as a heads-up

rusackas (approved):
Thanks!!!!

PR #30042, opened 2024-08-28
fix: use StrEnum type for GuestTokenResourceType to fix token parsing

hao-zhuventures:
### SUMMARY
This change is to fix the "incorrect" check results from comparing the `guest_token` claims with the enum value. By changing the type to be `StrEnum`, the str based comparison will output the desired results.

Additional context: guest user jwt token claims are casted to the `GuestToken` type [here](https://github.com/apache/superset/blob/75c500c9a53ce503b8636761f17b5b63eb8ee8e2/superset/security/manager.py#L2614).

### TESTING INSTRUCTIONS
Manual check before / after
```
>>> 'dashboard' == GuestTokenResourceType.DASHBOARD
False
```
```
>>> 'dashboard' == GuestTokenResourceType.DASHBOARD
True
```

and CI checks.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #30318, opened 2024-09-18
feat: reading language from url + resource type check fix

nyohasstium:
### SUMMARY

fix(embedded dashboard): resource type was not checked appropriately against enum.
feat(UI): added currency and language to URL_PARAMS
feat(API):  using language based on query parameter if found and defaulting to session value 
### TESTING INSTRUCTIONS

- Embed a dashboard using the superset-embedded-sdk package.
- Set lang and currency in the urlParams options.
- Depending on the chosen language, the ui will be rendered with it.
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI

### DIFF
--- a/superset-frontend/src/constants.ts
+++ b/superset-frontend/src/constants.ts
@@ -111,6 +111,14 @@ export const URL_PARAMS = {
     name: 'focused_chart',
     type: 'number',
   },
+  language: {
+    name: 'lang',
+    type: 'string'
+  },
+  currency: {
+    name: 'currency',
+    type: 'string'
+  }
 } as const;
 
 export const RESERVED_CHART_URL_PARAMS: string[] = [
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -2652,7 +2652,7 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             return False
 
         dashboards = [
-            r for r in user.resources if r["type"] == GuestTokenResourceType.DASHBOARD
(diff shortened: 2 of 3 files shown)


superset/security/manager.py
```diff
@@ -2652,7 +2652,7 @@ def has_guest_access(self, dashboard: "Dashboard") -> bool:
             return False
 
         dashboards = [
-            r for r in user.resources if r["type"] == GuestTokenResourceType.DASHBOARD
+            r for r in user.resources if r["type"] == GuestTokenResourceType.DASHBOARD.value
```
villebro, line 2655: I believe this was already fixed here: #30042

superset-frontend/src/constants.ts
```diff
@@ -111,6 +111,14 @@ export const URL_PARAMS = {
     name: 'focused_chart',
     type: 'number',
   },
+  language: {
```
michael-s-molina, line 114: I believe the correct parameter here should be locale which is comprised of both the language and region. Another point is that we should be able to infer the currency from the locale.

rusackas:
Hi @nyohasstium - just curious your thoughts on the review comments. Let us know if we can help you carry this effort forward.

rusackas:
This PR hasn't been updated in a long time and is in need of a rebase. We'll mark it as draft, but please set it as 'ready for review' when it's been brought up to speed.

rusackas:
This bundles a couple of unrelated changes and the review comments never got a reply. Closing as inactive - ping me if you'd like to revive it (ideally split out the embedded-resource fix).

PR #10084, opened 2020-06-17
fix: Fixed bug for issue #9967 ( Hive date functions for group functions in charts )

Nj-kol:
### SUMMARY

The issue was first observed in https://github.com/apache/incubator-superset/issues/7270
However, in the subsequent releases, the variable **time_grain_functions** was renamed to **_time_grain_expressions**. The implementation was also left out for hive. So time functions were not renamed which were specific to Hive. This was causing the group by functions in the Charts to malfunction

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

**Before**
image

**After**
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes #9967
- [ ] Changes UI
- [ ] Requires DB Migration.
- [ ] Confirm DB Migration upgrade and downgrade tested.
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #7676, opened 2019-06-08
Break up db_engine_specs

villebro:
### CATEGORY

Choose one

- [x] Refactor

### SUMMARY
This PR changes `db_engine_specs.py` into a package with one module per engine, e.g. `db_engine_specs/bigquery.py`. All engines are automatically populated into `db_engine_specs.engines` like before (now in `db_engine_specs/__init__.py`), so only importing of specific engine specs will need to be changed:

Before: `from superset.db_engine_specs import BaseEngineSpec`
After: `from superset.db_engine_specs.base import BaseEngineSpec`

### TEST PLAN
Local testing + CI

### ADDITIONAL INFORMATION

- [x] Has associated issue: Closes #7412 

### REVIEWERS
@khtruong @john-bodley @betodealmeida @mistercrunch

PR #30325, opened 2024-09-18
fix(db_engine_specs): add a few missing time grains to Postgres spec

sfirke:
### SUMMARY
I added to the Postgres engine spec definitions for some of the time grains that are defined in the base spec but had not been added to Postgres yet.

image

### TESTING INSTRUCTIONS
In an ephemeral environment, confirm that the grains are there and that the generated queries are correct.

### DIFF
--- a/tests/unit_tests/db_engine_specs/test_postgres.py
+++ b/tests/unit_tests/db_engine_specs/test_postgres.py
@@ -20,10 +20,11 @@ from typing import Any, Optional
 
 import pytest
 from pytest_mock import MockerFixture
-from sqlalchemy import types
+from sqlalchemy import column, types
 from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION, ENUM, JSON
 from sqlalchemy.engine.url import make_url
 
+from superset.db_engine_specs.postgres import PostgresEngineSpec as spec
 from superset.exceptions import SupersetSecurityException
 from superset.utils.core import GenericDataType
 from tests.unit_tests.db_engine_specs.utils import (
@@ -53,8 +54,6 @@ def test_convert_dttm(
     expected_result: Optional[str],
     dttm: datetime,  # noqa: F811
 ) -> None:
-    from superset.db_engine_specs.postgres import PostgresEngineSpec as spec
-
     assert_convert_dttm(spec, target_type, expected_result, dttm)
 
 
@@ -91,8 +90,6 @@ def test_get_column_spec(
     generic_type: GenericDataType,
     is_dttm: bool,
 ) -> None:
-    from superset.db_engine_specs.postgres import PostgresEngineSpec as spec
-
     assert_column_spec(spec, native_type, sqla_type, attrs, generic_type, is_dttm)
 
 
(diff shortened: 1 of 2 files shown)


villebro (changes requested):
NIce! Would you be able to add some tests for the Postgres time grains to `tests/unit_tests/db_engine_specs/test_postgres.py`?  You can pretty much copy this test from https://github.com/apache/superset/blob/e54353c4da71c16d6ca1e3d3906d3bcdc8d7d7a1/tests/unit_tests/db_engine_specs/test_pinot.py#L23-L46

mistercrunch:
Manually tested the expressions by getting GPT to convert to this to this statement:
```sql
SELECT
    NOW() AS original_datetime,
    DATE_TRUNC('minute', NOW()) + INTERVAL '5 seconds' * FLOOR(EXTRACT(SECOND FROM NOW()) / 5) AS five_seconds,
    DATE_TRUNC('minute', NOW()) + INTERVAL '30 seconds' * FLOOR(EXTRACT(SECOND FROM NOW()) / 30) AS thirty_seconds,
    DATE_TRUNC('minute', NOW()) AS minute,
    DATE_TRUNC('hour', NOW()) + INTERVAL '5 minutes' * FLOOR(EXTRACT(MINUTE FROM NOW()) / 5) AS five_minutes,
    DATE_TRUNC('hour', NOW()) + INTERVAL '10 minutes' * FLOOR(EXTRACT(MINUTE FROM NOW()) / 10) AS ten_minutes,
    DATE_TRUNC('hour', NOW()) + INTERVAL '15 minutes' * FLOOR(EXTRACT(MINUTE FROM NOW()) / 15) AS fifteen_minutes,
    DATE_TRUNC('hour', NOW()) + INTERVAL '30 minutes' * FLOOR(EXTRACT(MINUTE FROM NOW()) / 30) AS thirty_minutes
;
```
LGTM

sfirke:
@villebro I can mimic those tests for Postgres.  Though I wonder about the rigor of comparing the code in my spec to pasting the same code in a test.  What if instead/also I had it test the transformation of a specific timestamp?  Like that rounding `... 00:15:01` to 1 minute intervals gives `... 00:15:00` etc.  Unless that's already happening here and I'm missing it.

villebro:
> @villebro I can mimic those tests for Postgres. Though I wonder about the rigor of comparing the code in my spec to pasting the same code in a test. What if instead/also I had it test the transformation of a specific timestamp? Like that rounding `... 00:15:01` to 1 minute intervals gives `... 00:15:00` etc. Unless that's already happening here and I'm missing it.

So the reason why these tests are valuable, despite looking like pure duplication, is because a change might happen that causes these time grain definitions to no longer be used. For example, I once did a big refactor of the db engine specs (#7676), and that refactor caused the time grains to be removed by mistake from the Hive engine spec (this was later fixed in #10084). If we would have had tests like this in place, we would have caught that error during the refactor. I know it may look funny to duplicate that functional logic to the unit tests, but having them there makes it less likely to introduce regressions, and it also makes it safer to refactor the implementation details.

Regarding actual functional verification that the expression actually rounds down the timestamp based on the time grain, that is something that we can't unfortunately test here, as that logic resides in Postgres. But correct me if I'm misunderstanding what you're saying.

sfirke:
Thank you for that thorough explanation!  I now see the point of these tests.  And whoops I missed that we can't actually test the output of the time grain because that happens in Postgres, good call.  Okay, I will add tests like these.

tests/unit_tests/db_engine_specs/test_postgres.py
```diff
@@ -20,10 +20,11 @@
 
 import pytest
 from pytest_mock import MockerFixture
-from sqlalchemy import types
+from sqlalchemy import column, types
 from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION, ENUM, JSON
 from sqlalchemy.engine.url import make_url
 
+from superset.db_engine_specs.postgres import PostgresEngineSpec as spec
```
villebro, line 27: Nice, apparently the circular import that's been there before is no longer an issue 👍

villebro (approved):
LGTM

Issue #30326, opened 2024-09-18
legacy-plugin-chart-country-map on the map of Russia Karelia is located on the territory of the Murmansk region


SBIN2010:
### Bug description

Karelia is not displayed correctly on the map of Russia. It should be located separately, and not as part of Murmansk

### How to reproduce the bug

Open legacy-plugin-chart-country-map map of Russia

### Screenshots/recordings

To ![image](https://github.com/user-attachments/assets/f238a47e-7c31-4ba0-b7d1-6bf15d785827)
after 
![image](https://github.com/user-attachments/assets/daf3294d-0313-436a-9b86-fe2c77b9b498)

### Superset version

4.0.2

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
If you're able/willing to open a PR to correct this, the right way to do so would be to make a correction in the Jypyter Notebook (the repo's one `.ipynb` file) and have THAT render out the geoJSON. Otherwise, any GeoJSON overrides will eventually be overwritten in the future. Also, the files we get are generated from NaturalEarth data... you may want to ping them on their repo about correcting it at the root. We're downstream, and just patch things with overrides as much as we're able to.

rusackas:
We'd love to see a contribution to fix this in the Jupyter notebook (which tweaks/generates the GEOJson files). Here's some info on how to do it: https://preset.io/blog/how-and-when-and-why-to-update-apache-supersets-country-maps/

rusackas:
This seems to (still) not be fixed in the upstream data source. We're still open to PRs!

rusackas:
Per @rusackas, the upstream map data source has not corrected the Karelia placement. If anyone can submit a corrected GeoJSON via the Jupyter notebook workflow, a PR updating the country map data would be welcome.

rusackas:
#44273 looks like it finally fixes this the right way, correcting the boundary through the notebook generator instead of hand-patching the GeoJSON. CI's green, just needs review.

PR #30328, opened 2024-09-18
fix:(legacy-plugin-chart-country-map)Fix the territories of Karelia and Murmansk(#30326)

SBIN2010:

 BEFORE
image

AFTER
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas (changes requested):
Please make any geojson changes via the Jupyter notebook (it's the `.ipynb` file in the repo). This notebook will overwrite any/all geojson files, so the changes as you've made them here are ephemeral. 

Also, our data is gathered via [NaturalEarth](https://www.naturalearthdata.com/), so I would advise filing an issue with them (on their repo or otherwise) to fix this upstream from Superset.

SBIN2010:
[rusackas](https://github.com/rusackas) 

I don't think it's a good idea to add such changes to the Jupyter notebook. Since I found this flaw in the NaturalEarth database, and have already created an issue to fix it. (https://github.com/nvkelso/natural-earth-vector/issues/929)
Therefore, when the corrections appear in NaturalEarth, then it will be correct in Superset.

PR #30335, opened 2024-09-19
fix(table): Use extras in queries

### LINKED ISSUE #30193: `show summary` does not correctly apply all filters
### Bug description

In the changes introduced in #28057 ([code](https://github.com/apache/superset/pull/28057/files#diff-0186dcc4a861c06db2a94660bfcd32fb3301eee43a0d9c16e8e3b592ad7f8fc1R248)), all parameters are discarded when passing extras. However, the custom SQL filters are set within the extras, leading to an issue where the show summary feature does not correctly apply these filters, as illustrated in the attached image.

### How to reproduce the bug

1. Set the filter using custom SQL.
2. Select the show summary option.
3. Observe that the summary of the query results does not apply the filter.

### Screenshots/recordings

image

### Superset version

master / latest-dev

### Python version

3.10

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

Antonio-RiveroMartnez:
### SUMMARY
`Table` chart was removing the `extras` from the summary query, assuming they were only used for time grains. This PR fixes that and put the extras back in the queryObject sent as payload, now, only excluding `time_grain_sqla` if no temporal column is included in your chart.

Before: Excluding extras from summary query and only adding it if no time comparison was used
before

Now: Sending extras in the total query regardless you're using time comparison
No temporal Column
now_no_temp
With temporal Column
now_temp

### TESTING INSTRUCTIONS
1. Create a `Table` Chart
2. Use no temporal column and a custom SQL filter
3. Mark the `Show summary` checkbox
4. The queries sent as payload must include the `where` in extras for both queries and not include a `time_grain_sqla`
5. Now add a temporal column to your chart and run the query again
6. The `time_grain_sqla` must be present in the extras alongside with the `where` clause

### ADDITIONAL INFORMATION

- [X] Has associated issue: https://github.com/apache/superset/issues/30193
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
@@ -198,10 +198,15 @@ const buildQuery: BuildQuery<TableChartFormData> = (
         (ownState.currentPage ?? 0) * (ownState.pageSize ?? 0);
     }
 
+    if (!temporalColum) {
+      // This query is using only textual columns, so it doesn't need time grain
+      extras.time_grain_sqla = undefined;
+    }
+
     let queryObject = {
       ...baseQueryObject,
       columns,
-      extras: !isEmpty(timeOffsets) && !temporalColum ? {} : extras,
+      extras,
       orderby,
       metrics,
       post_processing: postProcessing,
@@ -239,7 +244,6 @@ const buildQuery: BuildQuery<TableChartFormData> = (
         row_limit: 0,
         row_offset: 0,
         post_processing: [],
-        extras: undefined, // we don't need time grain here
         order_desc: undefined, // we don't need orderby stuff here,
         orderby: undefined, // because this query will be used for get total aggregation.
       });
--- a/superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
@@ -114,8 +114,8 @@ const buildQuery: BuildQuery<TableChartFormData> = (
       }
     }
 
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -198,10 +198,15 @@ const buildQuery: BuildQuery<TableChartFormData> = (
         (ownState.currentPage ?? 0) * (ownState.pageSize ?? 0);
     }
 
+    if (!temporalColum) {
+      // This query is using only textual columns, so it doesn't need time grain
```
giftig, line 202: Or numerical presumably. Probably most accurate just to say it doesn't have any temporal columns so there's no time grain.

superset-frontend/plugins/plugin-chart-table/test/buildQuery.test.ts
```diff
@@ -114,5 +114,66 @@ describe('plugin-chart-table', () => {
...
+            subject: null,
+            operator: null,
+            comparator: null,
+            isExtra: false,
+            isNew: false,
+            datasourceWarning: false,
+            filterOptionName: 'filter_v8m9t9oq5re_ndzk6g5am7',
+          } as any,
+        ],
+      });
```
giftig, line 170: There's a lot of repetition here, I see there's already `...basicFormData`, feels like the bulk of this could be similarly reused to stay DRY. It'd make it much easier to see exactly what's different between the two tests here as well.

giftig (approved):
Minor suggestions but LGTM

giftig (approved):

PR #30336, opened 2024-09-19
feat(Digest): Add RLS at digest generation for Charts and Dashboards

geido:
### SUMMARY
Digest generation for charts and dashboards wasn't considering RLS for users and guest users (via Embedded). This PR includes the RLS in the hash generation for the digest. The digest is included in cache keys and will be updated according to the RLS rules. Previously you could generate a screenshot and any RLS applied later won't be applied to follow-up screenshots as the same cache key was being reused.

N.A.

### TESTING INSTRUCTIONS
1. Generate a screenshot of a Dashboard
2. Add RLS rules for the current user
3. Generate the screenshot again
4. The screenshot should present the Dashboard with RLS rules applied

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -1046,14 +1046,14 @@ class DashboardRestApi(BaseSupersetModelRestApi):
             cache_dashboard_screenshot.delay(
                 username=get_current_user(),
                 guest_token=g.user.guest_token
-                if isinstance(g.user, GuestUser)
+                if hasattr(g, "user") and isinstance(g.user, GuestUser)
                 else None,
                 dashboard_id=dashboard.id,
                 dashboard_url=dashboard_url,
+                cache_key=cache_key,
                 force=True,
                 thumb_size=thumb_size,
                 window_size=window_size,
-                cache_key=cache_key,
             )
             return self.response(
                 202,
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -1040,14 +1040,13 @@ class DashboardRestApi(BaseSupersetModelRestApi):
         image_url = get_url_path(
             "DashboardRestApi.screenshot", pk=dashboard.id, digest=cache_key
         )
-        username = get_current_user()
 
         def trigger_celery() -> WerkzeugResponse:
             logger.info("Triggering screenshot ASYNC")
             cache_dashboard_screenshot.delay(
-                username=username,
+                username=get_current_user(),
                 guest_token=g.user.guest_token
-                if username and isinstance(g.user, GuestUser)
+                if get_current_user() and isinstance(g.user, GuestUser)
                 else None,
                 dashboard_id=dashboard.id,
                 dashboard_url=dashboard_url,
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -1040,13 +1040,14 @@ class DashboardRestApi(BaseSupersetModelRestApi):
         image_url = get_url_path(
             "DashboardRestApi.screenshot", pk=dashboard.id, digest=cache_key
         )
+        username = get_current_user()
 
         def trigger_celery() -> WerkzeugResponse:
             logger.info("Triggering screenshot ASYNC")
             cache_dashboard_screenshot.delay(
-                username=get_current_user(),
+                username=username,
                 guest_token=g.user.guest_token
-                if hasattr(g, "user") and isinstance(g.user, GuestUser)
+                if username and isinstance(g.user, GuestUser)
                 else None,
                 dashboard_id=dashboard.id,
                 dashboard_url=dashboard_url,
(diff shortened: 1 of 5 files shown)


Antonio-RiveroMartnez (approved):

superset/thumbnails/digest.py
```diff
@@ -18,15 +18,18 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING
```
villebro, line 21: For new code it'd be nice if we always use `from __future__ import annotations` and then use `| None` notation, rather than `Optional[]`

superset/thumbnails/digest.py
```diff
@@ -49,8 +52,42 @@ def _adjust_string_for_executor(
...
+def _adjust_string_with_rls(
+    unique_string: str,
+    datasources: list[Optional[SqlaTable]] | set[BaseDatasource],
+    executor_type: ExecutorType,
+    executor: str,
+) -> str:
+    """
+    Add the RLS filters to the unique string based on current executor.
+    """
+    if executor_type != ExecutorType.SELENIUM:
```
villebro, line 64: Is there a reason we're not applying RLS for the selenium user?

villebro (commented):
First pass comments

superset/dashboards/api.py
```diff
@@ -1046,14 +1046,14 @@ def trigger_celery() -> WerkzeugResponse:
             cache_dashboard_screenshot.delay(
                 username=get_current_user(),
                 guest_token=g.user.guest_token
-                if isinstance(g.user, GuestUser)
+                if hasattr(g, "user") and isinstance(g.user, GuestUser)
```
villebro, line 1049: Could we use `superset.tasks.utils.get_current_user` here?

superset/dashboards/api.py
```diff
@@ -1046,14 +1046,14 @@ def trigger_celery() -> WerkzeugResponse:
...
                 else None,
                 dashboard_id=dashboard.id,
                 dashboard_url=dashboard_url,
+                cache_key=cache_key,
                 force=True,
                 thumb_size=thumb_size,
                 window_size=window_size,
-                cache_key=cache_key,
```
villebro, line 1056: nit: Is there a reason we're reordering these?

superset/security/manager.py
```diff
@@ -2639,8 +2639,12 @@ def is_guest_user(user: Optional[Any] = None) -> bool:
 
         if not is_feature_enabled("EMBEDDED_SUPERSET"):
             return False
+
         if not user:
+            if not hasattr(g, "user"):
+                return False
```
villebro, line 2645: Again I would prefer to use `get_current_user` if that's possible

superset/thumbnails/digest.py
```diff
@@ -49,8 +52,40 @@ def _adjust_string_for_executor(
...
+                if (
+                    datasource
+                    and hasattr(datasource, "is_rls_supported")
+                    and datasource.is_rls_supported
+                ):
+                    rls_filters = datasource.get_sqla_row_level_filters()
+                    stringified_rls += "".join([str(f) for f in rls_filters])
+
+        if stringified_rls != "":
+            unique_string = f"{unique_string}\n{stringified_rls}"
```
villebro, line 81: nit:
```suggestion
        if stringified_rls:
            unique_string = f"{unique_string}\n{stringified_rls}"
```

tests/unit_tests/thumbnails/test_digest.py
```diff
@@ -131,45 +152,103 @@ def CUSTOM_CHART_FUNC(
...
             [ExecutorType.CURRENT_USER],
             False,
             False,
+            [],
             ExecutorNotFoundError(),
         ),
     ],
 )
-def test_dashboard_digest(
+def test_dashboard_digest_with_rls(
```
villebro, line 221: Not sure if we need to append `with_rls` to the test name, as we don't have a case where we want to ignore RLS in digest creation

superset/thumbnails/digest.py
```diff
@@ -49,8 +52,40 @@ def _adjust_string_for_executor(
...
+        stringified_rls = ""
+        with override_user(user):
+            for datasource in datasources:
+                if (
+                    datasource
+                    and hasattr(datasource, "is_rls_supported")
+                    and datasource.is_rls_supported
+                ):
+                    rls_filters = datasource.get_sqla_row_level_filters()
+                    stringified_rls += "".join([str(f) for f in rls_filters])
```
villebro, line 78: I think it may be a good idea to append the dataset id to the RLS string, just to make sure we don't introduce potential leaks. Something like `<id>\t<rls_filters>`. Also, I think it may be a good idea to use some other separator than an empty string(`\t` is probably a safe bet) and also finish it off with a `\n`, also for the same reason.

villebro (commented):
I forgot to say great work, thanks for taking this on, important improvement! Second pass comments.

superset/dashboards/api.py
geido: No not really, just readability

PR #30340, opened 2024-09-19
fix(CI): increase node JS heap size

rusackas:
### SUMMARY
This is a bit of a shot in the dark, but I saw that some Jest tests were failing due to Javascript running out of heap memory (see screenshot).
My _conjecture_ is that the Native Filter tests we're struggling with are the same issue we're running into here.

So... let's throw a dart at the problem. Apparently, the default size is 1.5GB, so let's go to 4! GPT thinks this'll work for both jest, and the chrome/chromium browser we use in E2E tests, so here goes nothing! Apparently the small GH action runners have 7GB ram, so this seems reasonable-ish.

Screenshot 2024-09-19 at 3 53 47 PM

### TESTING INSTRUCTIONS
If CI passes, we should be good. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


villebro:
No PR talking about dialing something up should go without a reference to Spinal Tap..

image

rusackas:
> No PR talking about dialing something up should go without a reference to Spinal Tap..
> 
> image

LOL, I'm literally trying to get a tattoo of that knob. Awaiting response from two artists.

villebro:
> LOL, I'm literally trying to get a tattoo of that knob. Awaiting response from two artists.

After that I feel you should be granted permission to self-approve any PR making these types of changes 🎛️

mistercrunch (approved):
TO ELEVEN!

dpgaspar (approved):

rusackas:
Hmmm... one of the OTHER tests is failing, but (thankfully) not the usual native filters tests. Maybe I only turned it up to like 9. Let's see what a couple more GB does...

villebro:
> Hmmm... one of the OTHER tests is failing, but (thankfully) not the usual native filters tests. Maybe I only turned it up to like 9. Let's see what a couple more GB does...

That's one of the less frequently flaking tests, and is probably unaffected by extra memory

villebro (approved):
LGTM, let's hope this makes the Cypress tests less flaky/quicker 👍

rusackas:
I'm actually optimistic about the quicker part... this should use less swap, at least.

PR #30354, opened 2024-09-21
chore: split cypress files for less memory

eschutho:
### SUMMARY

Testing to see if this will fix the cypress memory issues. 
Also added the chrome flag `disable-dev-shm-usage`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Matrix (1, chrome) failed so I restarted it manually. Wondering if a wider matrix would help. Also it seems we have 5 * chrome now, but could try running also against Firefox to see if the memory consuption is more or less of a problem on that browser.

rusackas:
> Matrix (1, chrome) failed so I restarted it manually. For the record `dashboard/nativeFilters.test.ts` is the one that failed, even after shrinking it down as part of this PR. I'm wondering if lowering the limit / cardinality of filters could help here too if the issue comes from too many filters with too high of a cardinality (assuming that this specific test is the hot spot).
> 
> Wondering if a wider matrix would help. Also it seems we have 5 * chrome now, but could try running also against Firefox to see if the memory consuption is more or less of a problem on that browser.

Yes, that is the hot spot, and it's always the one that fails. We've also seen GitHub issues about performance regressions when there are lots of filters, so this failure is likely reflecting a real problem. We can throw memory, split tests, reduce test scope, etc. to smooth out devEx, but I think we're just dancing around a real problem.

eschutho:
@rusackas this is passing CI.. can we try to merge it?

mistercrunch (approved):

rusackas (approved):

mistercrunch:
Looked at `master` hoping to see some green but no :/
Screenshot 2024-09-25 at 2 35 23 PM

mistercrunch:
Giving a shot at troubleshooting memory here -> https://github.com/apache/superset/pull/30397 , seems promising

PR #30364, opened 2024-09-23
feat(datasets): Allow swap dataset after deletion

Antonio-RiveroMartnez:
### SUMMARY
Currently, if you delete a dataset, the charts that were using it don't keep the controls config when swapping to a dataset that have similar structure.
This is because the state gets overwritten with form data that doesn't have the controls configs so the sync method is not able to fetch the correct values for them.

This PR keeps the values and other configs when the API response doesn't include dataset id (deleted) and fallback to read from the datasource when getting the controls values in case there's nothing in the state.

One thing to notice is that before, when you first loaded a chart that have no dataset because it was deleted, the explore data was a fallback and didn't trigger the API requests, now, because we're not using this fallback explore but keeping the form data info from the API response, the UI will show a `Not Found` error until you swap the dataset and update the chart.

Before:

https://github.com/user-attachments/assets/504820d1-de8b-469c-a471-31867649dacd

Now:

https://github.com/user-attachments/assets/f5495027-5270-42d4-9a9e-25fb9cbbe3a2

### TESTING INSTRUCTIONS
Create two datasets with the same schema (columns and metrics).
Create a chart using one of these and save it.
Delete this dataset.
Access the chart -- you'll see an option to swap the dataset.
Remap the chart to use the other dataset.
Your controls config must be kept

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/controlUtils/getControlValuesCompatibleWithDatasource.ts
+++ b/superset-frontend/src/explore/controlUtils/getControlValuesCompatibleWithDatasource.ts
@@ -54,37 +54,34 @@ const isControlValueCompatibleWithDatasource = (
       );
     }
   }
-  if (controlState.savedMetrics && isSavedMetric(value)) {
-    if (
-      controlState.savedMetrics.some(
-        (savedMetric: Metric) => savedMetric.metric_name === value,
-      ) ||
-      !isEmpty(datasource?.metrics)
-    ) {
-      return datasource.metrics.some(
-        (metric: Metric) => metric.metric_name === value,
-      );
-    }
+  if (
+    controlState.savedMetrics &&
+    isSavedMetric(value) &&
+    (controlState.savedMetrics.some(
+      (savedMetric: Metric) => savedMetric.metric_name === value,
+    ) ||
+      !isEmpty(datasource?.metrics))
+  ) {
+    return datasource.metrics.some(
+      (metric: Metric) => metric.metric_name === value,
+    );
   }
   if (
     controlState.columns &&
-    (isAdhocMetricSimple(value) || isSimpleAdhocFilter(value))
-  ) {
-    if (
-      (!isEmpty(controlState.columns) &&
-        controlState.columns.some(
-          (column: Column) =>
-            column.column_name === (value as AdhocMetric).column?.column_name ||
-            column.column_name === (value as SimpleAdhocFilter).subject,
-        )) ||
-      !isEmpty(datasource?.columns)
-    ) {
-      return datasource.columns.some(
+    (isAdhocMetricSimple(value) || isSimpleAdhocFilter(value)) &&
+    ((!isEmpty(controlState.columns) &&
+      controlState.columns.some(
         (column: Column) =>
           column.column_name === (value as AdhocMetric).column?.column_name ||
           column.column_name === (value as SimpleAdhocFilter).subject,
-      );
-    }
+      )) ||
+      !isEmpty(datasource?.columns))
+  ) {
+    return datasource.columns.some(
+      (column: Column) =>
+        column.column_name === (value as AdhocMetric).column?.column_name ||
+        column.column_name === (value as SimpleAdhocFilter).subject,
+    );
   }
   if (isAdhocMetricSQL(value)) {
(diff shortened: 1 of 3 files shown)


sfirke:
Wow, that is a nice quality of life improvement for users!  Huge potential time saver.

rusackas:
I think this PR is a fantastic step in the right direction ❤️ 

If there's interest/bandwidth to take things a step further, it's worth noting that deleting a dataset might break HUNDREDS of charts. While this PR allows you to restore them one at a time, it might be a great idea to (also) allow this swap/substitution in the Dataset deletion modal. You could give the user any/all of these choices:
* Replace {deleted dataset} with [Select dataset ˅]
* Delete dataset (charts will be broken)
* Delete dataset and affected charts

Of course, taking it even further over the horizon, we probably shouldn't be deleting _anything_ but rather doing "soft deletes" where rows are tagged as deleted, so that they can be restored by a DBA. (Note, this is not the same as "Archiving" which requires UI changes and un-archiving capabilities.

Love the PR, but just saying this is the entry of a longer tunnel ;)

superset-frontend/src/explore/controlUtils/getControlValuesCompatibleWithDatasource.ts
```diff
@@ -27,50 +27,64 @@ import {
...
-        column.column_name === (value as SimpleAdhocFilter).subject,
-    )
+    (isAdhocMetricSimple(value) || isSimpleAdhocFilter(value))
   ) {
-    return datasource.columns.some(
-      (column: Column) =>
-        column.column_name === (value as AdhocMetric).column?.column_name ||
-        column.column_name === (value as SimpleAdhocFilter).subject,
-    );
+    if (
```
kgabryje, line 73: same here

superset-frontend/src/pages/Chart/index.tsx
```diff
@@ -43,7 +43,10 @@ import { getItem, LocalStorageKeys } from 'src/utils/localStorageHelpers';
 import { getFormDataWithDashboardContext } from 'src/explore/controlUtils/getFormDataWithDashboardContext';
 
 const isValidResult = (rv: JsonObject): boolean =>
-  rv?.result?.form_data && isDefined(rv?.result?.dataset?.id);
+  rv?.result?.form_data && rv?.result?.dataset;
+
+const hasDataseId = (rv: JsonObject): boolean =>
```
kgabryje, line 48:
```suggestion
const hasDatasetId = (rv: JsonObject): boolean =>
```

superset-frontend/src/pages/Chart/index.tsx
```diff
@@ -52,7 +55,19 @@ const fetchExploreData = async (exploreUrlParams: URLSearchParams) => {
       endpoint: 'api/v1/explore/',
     })(exploreUrlParams);
     if (isValidResult(rv)) {
-      return rv;
+      if (hasDataseId(rv)) {
```
kgabryje, line 58:
```suggestion
      if (hasDatasetId(rv)) {
```

kgabryje (approved):
Looks great! Left some non-blocking comments.

Unrelated to this PR, but I'd love to see a more verbose error than "Not found" in the chart area

superset-frontend/src/explore/controlUtils/getControlValuesCompatibleWithDatasource.ts
```diff
@@ -27,50 +27,64 @@ import {
...
-    isSavedMetric(value) &&
-    controlState.savedMetrics.some(
-      (savedMetric: Metric) => savedMetric.metric_name === value,
-    )
-  ) {
-    return datasource.metrics.some(
-      (metric: Metric) => metric.metric_name === value,
-    );
+  if (controlState.savedMetrics && isSavedMetric(value)) {
+    if (
```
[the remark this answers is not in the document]
Antonio-RiveroMartnez: Trying to make them as the first one, but yeah, the new condition is not that crazy so I'll just use one.

geido (approved):

PR #30365, opened 2024-09-24
chore: add logging for dashboards/get warnings

eschutho:
### SUMMARY

We're currently not logging flask warnings to the logger as part of the "catch all" @safe or @handle_api_exception decorators. I think @handle_api_exception does a more thorough job of catching different errors and handling, logging, raising, etc. We're applying @handle_api_exception on most base restapi functions, but using @safe on the adhoc apis. Maybe we can look at combining these and only having one decorator. I also removed the exception handling in the api because they will be caught and handled in the decorator. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/exceptions.py
+++ b/superset/commands/database/exceptions.py
@@ -202,3 +202,15 @@ class DatabaseOfflineError(SupersetErrorException):
 
 class InvalidParametersError(SupersetErrorsException):
     status = 422
+
+
+class DatasetValidationError(CommandException):
+    status = 422
+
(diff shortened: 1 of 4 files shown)


betodealmeida (approved):
Looks good! I think we need to get rid of `@safe` at some point, since the schema of the error payload it produces are not [SIP-41 compliant](https://github.com/apache/superset/issues/9194).

Issue #30377, opened 2024-09-24
Dataset get_or_create API fail to create a dataset with the same table_name but different schema


luizcapu:
### Bug description

`get_or_create` Dataset API endpoint breaks with `500 - Internal Server` error if there is any existing dataset with the same `table_name`, regardless of the dataset `schema`.

This is not the behaviour we observe on other endpoints, such as a `POST /dataset` from where users are able to create datasets with the same table_name, as long as they belong to either a different `database_id` or a `different schema`.

### How to reproduce the bug

1. Go to the datasets page
2. Pick any existing dataset name and prepare a payload as follows (example using the `users` datasets)

```
{
  'table_name': 'users',
  'schema': 'other',
  'database_id': 1,
}
```

3. Submit this payload via a POST request to /api/v1/dataset/get_or_create
4. Note how the API will return with a `500 - Internal Server Error` response

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
Sorry this (and the PR) have been stuck for >200 days now! Must just be an edge case that isn't bugging most folks. Hopefully we can get the PR brought up to a mergable state and get it in, if you're still facing this issue in current versions (4.1.2 or newer)

rusackas:
Help welcomed in contributing to or adopting the linked PR. This might be closed as "stale" and/or "not planned" if nobody is interested in carrying it through.

rusackas:
Per @rusackas, if nobody picks up the linked PR or shows interest in fixing this, it will be closed as 'not planned'. If you'd like to contribute, please comment on the linked PR or here.

Abdulrehman-PIAIC80387:
Hi @luizcapu — are you still working on this / PR #30379? If not, I'd like to pick it up and open a PR shortly. Thanks!

PR #30379, opened 2024-09-24
fix(dataset-api): get_or_create creates a dataset for an existing table_name but different schema

luizcapu:
### SUMMARY

At Pinterest we were trying to use the `get_or_create` endpoint to automate the integration between our MetricsLayer and Superset. During our tests we've encountered the following issue: https://github.com/apache/superset/issues/30377

The issue happens because at the moment the `get_or_create` code does not account for the payload `schema` attribute, doing a search by `table_name` only.

This PR changes `get_or_create` to take the `schema` into account:

* A new `get_table_by_schema_and_name` was added to the `DatasetDAO` class
* `get_or_create` now checks if the dataset exists by calling `DatasetDAO.get_table_by_schema_and_name` (instead of `DatasetDAO.get_table_by_name`)

### TESTING INSTRUCTIONS

- Integration tests were added for the new behaviour
- Manual tests: reproducing the steps described on the issue

**Case 1 - False Positive**

1. Go to the datasets page
2. Pick any existing dataset name and prepare a payload as follows (example using the `users` datasets)

```
{
  'table_name': 'users',
  'schema': 'other',
  'database_id': 1,
}
```

3. Submit this payload via a POST request to /api/v1/dataset/get_or_create
4. A new dataset is created. The API returns with a `200` pointing to the new dataset. No false positives anymore.

**Case 2 - Internal Server Error**

1. Create 2 or more datasets with the same `table_name` and different `schemas` (either via UI or create dataset API)
2. Try to create a new dataset. Again, with same `table_name` but a different schema. Payload example:

```
{
  'table_name': 'users',
  'schema': 'any_new_schema_name',
  'database_id': 1,
}
```

3. Submit this payload via a POST request to /api/v1/dataset/get_or_create
4. A new dataset is created. The API returns with a `200` pointing to the new dataset. No 500 errors anymore.

**Backward Compatibility**

1. Go to the datasets page
2. Pick any existing dataset name and prepare a payload as follows (without passing the schema)

```
{
  'table_name': 'users',
  'database_id': 1,
}
```

3. Submit this payload via a POST request to /api/v1/dataset/get_or_create
4. No new dataset is created. The API returns with a `200` with the response body pointing to the existing dataset.

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/daos/dataset.py
+++ b/superset/daos/dataset.py
@@ -431,6 +431,16 @@ class DatasetDAO(BaseDAO[SqlaTable]):
         filterable.update(DATASET_CUSTOM_FIELDS)
         return filterable
 
+    @staticmethod
+    def get_table_by_schema_and_name(
+        database_id: int, schema: str, table_name: str
+    ) -> SqlaTable | None:
+        return (
+            db.session.query(SqlaTable)
+            .filter_by(database_id=database_id, schema=schema, table_name=table_name)
+            .one_or_none()
+        )
+
 
 class DatasetColumnDAO(BaseDAO[TableColumn]):
     pass
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -1094,9 +1094,13 @@ class DatasetRestApi(BaseSupersetModelRestApi):
             body = GetOrCreateDatasetSchema().load(request.json)
         except ValidationError as ex:
             return self.response(400, message=ex.messages)
+        schema_name = body.get("schema")
         table_name = body["table_name"]
         database_id = body["database_id"]
-        if table := DatasetDAO.get_table_by_name(database_id, table_name):
+
+        if table := DatasetDAO.get_table_by_schema_and_name(
+            database_id, schema_name, table_name
+        ):
             return self.response(200, result={"table_id": table.id})
 
         body["database"] = database_id
--- a/tests/integration_tests/datasets/api_tests.py
+++ b/tests/integration_tests/datasets/api_tests.py
@@ -2983,6 +2983,67 @@ class TestDatasetApi(SupersetTestCase):
         with examples_db.get_sqla_engine() as engine:
             engine.execute("DROP TABLE test_create_sqla_table_api")
 
+    def test_get_or_create_dataset_creates_table_for_different_schema(self):
+        """
+        Dataset API: Test get or create endpoint when table is created
+        """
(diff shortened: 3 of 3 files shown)


luizcapu:
Marking this PR as ready for review. [test-sqlite](https://github.com/apache/superset/actions/runs/11019081178/job/30655002034?pr=30379) is the only test failing.

However, it's failing with `sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: tables.table_name`, which is inherently opposed to the nature of this PR.

Furthermore, there are some signals that this constraint should be dropped:

- it doesn't align with other databases (MySQL/Postgres) or with the overall Superset functionalities (as it does allow duplicate table_names for different databases/schemas)
- The Github action logs says: `SQLite Database support for metadata databases will be removed in a future version of Superset.`
- There's [this migration](https://github.com/apache/superset/blob/master/superset/migrations/versions/2020-09-25_10-56_18532d70ab98_fix_table_unique_constraint_in_mysql.py) that is removing the uniqueness constraint on the table_name, but is doing it for `MySQLDialect` only

I could use some guidance to decide whether or not dropping this constraint is the right decision. And if so, how to do it.

Thank you in advance.

rusackas:
Sorry this seems to be stuck in limbo for so long... it looks like CI was failing all over. Perhaps after a rebase, you can help get tests to pass, and we can get this merged?

rusackas:
Resolved the conflict, but I'm not sure tests will pass... we shall see.

rusackas:
This PR hasn't been updated in a long time and has CI failures (pre-commit, test-sqlite) that likely stem from drift against `master`. The fix looks solid — we'll mark it as draft for now. Please rebase against `master`, fix any CI failures, and set it back to 'ready for review' when it's ready.

rusackas:
@luizcapu the schema-aware get_or_create looks right and is well-tested. I'll see about rebasing this and whether or not I can help get CI to pass. Thanks!

rusackas:
Upon closer inspection, it looks like this ended up getting handled by #30377, which landed a catalog-aware version of the same get_or_create fix (and returns a 400 instead of a 500 on the ambiguous case). Going to close this as superseded, but holler if there's something yours covers that #30377 missed.

PR #30389, opened 2024-09-25
fix: Pre-query normalization with custom SQL

michael-s-molina:
### SUMMARY
This PR fixes an issue with pre-query normalization when dealing with custom SQL. Previously, a custom SQL that had a label that does not match any column name would throw the following error:

```
superset/superset/connectors/sqla/models.py", line 1691, in _normalize_prequery_result_type
    column_ = columns_by_name[dimension]
KeyError: 'custom_sql'
```

Given function docs:

```
Convert a prequery result type to its equivalent Python type.

Some databases like Druid will return timestamps as strings, but do not perform
automatic casting when comparing these strings to a timestamp. For cases like
this we convert the value via the appropriate SQL transform.
```

We simply skip the conversion of custom SQL types as we cannot determine if they are temporal or not currently.

https://github.com/user-attachments/assets/e6191c06-2d78-43a9-8d08-334453ca78d2

### TESTING INSTRUCTIONS
Repro the steps shown in the video using a db engine spec with `allows_subqueries = False`

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -1688,10 +1688,10 @@ class SqlaTable(
         if isinstance(value, np.generic):
             value = value.item()
 
-        column_ = columns_by_name[dimension]
+        column_ = columns_by_name.get(dimension)
         db_extra: dict[str, Any] = self.database.get_extra()
 
-        if column_.type and column_.is_temporal and isinstance(value, str):
+        if column_ and column_.type and column_.is_temporal and isinstance(value, str):
             sql = self.db_engine_spec.convert_dttm(
                 column_.type, dateutil.parser.parse(value), db_extra=db_extra
             )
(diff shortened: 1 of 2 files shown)


villebro:
@michael-s-molina I'm trying to understand when this would happen. Could this have something to do with `BaseEngineSpec.make_label_compatible` changing that column alias?

michael-s-molina:
> @michael-s-molina I'm trying to understand when this would happen. Could this have something to do with `BaseEngineSpec.make_label_compatible` changing the column alias?

This happens when your engine spec has `allows_subqueries = False`

villebro (approved):
Fix LGTM, although I'm not convinced that method should be there in the first place. I'd almost prefer to follow Elon's "delete, delete, delete" mantra here and remove the whole method instead of fixing this bug, but in the interest of stabilizing the codebase maybe not right now. However, maybe we could start using some convention for annotating code that should be removed/cleaned up during breaking windows? If we don't start cleaning up stuff like this we'll never get out of the tech debt hole we're in..

PR #30393, opened 2024-09-25
fix: Allows X-Axis Sort By for custom SQL

michael-s-molina:
### SUMMARY
This PR fixes an issue where the X-Axis Sort By control was disappearing when a custom SQL was being used as an x-axis. Given that we currently have no way to identify the result type of a custom SQL, we make the control available and give the responsibility to the user to provide a sortable result. We implement the same behavior for the Time Grain control as you can see in the recordings.

https://github.com/user-attachments/assets/f40cbd8c-c957-45b7-83be-0df5a5f25ec3

https://github.com/user-attachments/assets/c4a48703-95c8-4780-a669-e101e214ac9d

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
@@ -23,7 +23,6 @@ import {
   GenericDataType,
   getColumnLabel,
   getMetricLabel,
-  isPhysicalColumn,
   QueryFormColumn,
   QueryFormMetric,
   t,
@@ -40,6 +39,7 @@ import {
   SORT_SERIES_CHOICES,
 } from '../constants';
 import { checkColumnType } from '../utils/checkColumnType';
+import { isSortable } from '../utils/isSortable';
 
 export const contributionModeControl = {
   name: 'contributionMode',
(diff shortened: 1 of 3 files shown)


superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
```diff
@@ -66,10 +67,16 @@ function isForcedCategorical(controls: ControlStateMapping): boolean {
...
+  // Given that we don't know the type of a custom SQL column,
+  // we treat it as sortable and give the responsibility to the
+  // user to provide a sortable result.
+  const isCustomSQL = !isPhysicalColumn(xAxisValue);
   return (
     isForcedCategorical(controls) ||
+    isCustomSQL ||
     checkColumnType(
-      getColumnLabel(controls?.x_axis?.value as QueryFormColumn),
+      getColumnLabel(xAxisValue),
```
villebro, line 79: I was wondering why this was passing our 100 % coverage check, but it turns out this file is `tsx` which is ignored in the check. I'm thinking at some point we should really move this type of critical non-React logic into `.ts` files and start adding tests for them. For this specific function, I'd maybe move it under `src/shared-controls/utils/isSortable.ts` and then add a test for it.

villebro (approved):
LGTM with an optional extra task to break out this into a dedicated util and add 100 % coverage for it (yay!)

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
michael-s-molina: So, I tried to add a test for our shared controls but quickly find out that the way they are structured makes the tests really hard. I hate this weird declarative structure that we have instead of working with real React composition. I think we should postpone writing tests for them right now because we should rewrite them completely.

michael-s-molina: @villebro I extracted the function and added tests.

villebro: Niiice! ❤️

villebro (approved):
Michael cleans up utils faster than this guy! 🚀 
image

PR #30215, opened 2024-09-10
fix(Fave): Charts and Dashboards fave/unfave do not commit transactions

geido:
### SUMMARY
We noticed that when faving/unfaving a Dashboard or a Chart the transaction wasn't being committed. This PR introduces commands and uses the transaction decorator to commit the transactions.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
N.A.

### TESTING INSTRUCTIONS
1. Fave/Unfave should save for both charts and dashboards

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #30394, opened 2024-09-25
feat: turn off autoflush on integration tests

betodealmeida:
### SUMMARY

https://github.com/apache/superset/pull/24969 introduced the `@transaction` decorator, to be used in commands to ensure that commits are explicit and happen only once in the lifetime of a request, preventing eagers commits. The PR failed to catch a few cases where legacy API endpoints were not using commands, resulting in successful requests that wouldn't be persisted to the database. These were fixed in https://github.com/apache/superset/pull/30215.

Why did our integration tests fail to catch these endpoints that were not committing? Tests failed because by default SQLAlchemy will autoflush before running a `SELECT`. In real world, favoriting a chart and verifying that it was favorited are two separate requests. If the first one fails to commit, the second one won't see a change. But our integration tests share the same global session (`db.session`), so that an insert followed by a select will result in flush between them, and the non-committed data will be seen even though we use `READ COMMITTED` as the isolation level.

To prevent this from happening again, this PR turns off the autoflush in the integration tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/conftest.py
+++ b/tests/integration_tests/conftest.py
@@ -43,6 +43,15 @@ CTAS_SCHEMA_NAME = "sqllab_test_db"
 ADMIN_SCHEMA_NAME = "admin_database"
 
 
+# When running tests it's important to disable autoflush to prevent dirty reads. If a
+# command modifies the database but doesn't explicitly commit the transaction, a
+# subsequent read won't see the data because we use READ COMMITTED isolation level. But
+# many integration tests write and read in the same session, and because by default
+# SQLAlchemy uses autoflush, the test would see data that in real world wouldn't be
+# there.
+db.session.configure(autoflush=False)
+
+
 @pytest.fixture
 def app_context():
     with app.app_context() as ctx:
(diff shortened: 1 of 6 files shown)


tests/integration_tests/conftest.py
```diff
@@ -43,6 +43,15 @@
 ADMIN_SCHEMA_NAME = "admin_database"
 
 
+# When running tests it's important to disable autoflush to prevent dirty reads. If a
+# command modifies the database but doesn't explicitly commit the transaction, a
+# subsequent read won't see the data because we use READ COMMITTED isolation level. But
+# many integration tests write and read in the same session, and because by default
+# SQLAlchemy uses autoflush, the test would see data that in real world wouldn't be
+# there.
+db.session.configure(autoflush=False)
```
mistercrunch, line 52: QQ: would this be needed/preferred on the `tests/unit_tests/` side of the house too? If so it could be great to factor out into `tests/common/` (?)

mistercrunch: just that one question, otherwise PR LGTM

geido (approved):

michael-s-molina:
@betodealmeida Another idea would be to keep autoflush in our tests and check the session for pending commits/rollbacks after our tests run. This can be done globally and would work for current and new code.

rusackas:
Are we still intending to get this merged? Converting to draft since it's been inactive for a while and is in need of a rebase.

PR #30395, opened 2024-09-25
chore: remove flaky test

eschutho:
### SUMMARY

Continuing to debug the flaky cypress issues. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/bashlib.sh
+++ b/.github/workflows/bashlib.sh
@@ -162,7 +162,17 @@ cypress-run-all() {
     USE_DASHBOARD_FLAG='--use-dashboard'
   fi
 
-  python ../../scripts/cypress_run.py --parallelism $PARALLELISM --parallelism-id $PARALLEL_ID $USE_DASHBOARD_FLAG
+  docker pull cypress/browsers:node-20.16.0-chrome-127.0.6533.119-1
+
+  docker run --rm \
+    -e CYPRESS_BASE_URL \
+    -e PARALLELISM \
+    -e PARALLELISM_ID \
+    -e USE_DASHBOARD=${USE_DASHBOARD} \
+    -v "$GITHUB_WORKSPACE/superset-frontend/cypress-base:/e2e" \
+    -w /e2e \
+    cypress/browsers:node-20.16.0-chrome-127.0.6533.119-1 \
+    npx cypress run --browser chrome $USE_DASHBOARD_FLAG
 
   # After job is done, print out Flask log for debugging
   echo "::group::Flask log for default run"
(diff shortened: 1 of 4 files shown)


superset-frontend/cypress-base/cypress/e2e/dashboard/nativeFilters.test.ts
```diff
@@ -312,15 +312,15 @@ describe('Native filters', () => {
   });
 
   describe('Nativefilters initial state not required', () => {
```
sadpandajoe, line 314: instead of skipping each test, i believe if you did a `describe.skip()` you'll skip all tests in the describe block

superset-frontend/cypress-base/cypress/e2e/dashboard/horizontalFilterBar.test.ts
```diff
@@ -81,7 +81,7 @@ describe('Horizontal FilterBar', () => {
     setFilterBarOrientation('vertical');
   });
 
-  it('should show all default actions in horizontal mode', () => {
+  it.skip('should show all default actions in horizontal mode', () => {
```
sadpandajoe, line 84: instead of doing all the skips here and below, you can just do `it.only` and it'll only run that test above. You can probably put a note after that test to list out why you're skipping all these other tests

superset-frontend/cypress-base/cypress/e2e/explore/chart.test.js
```diff
@@ -122,32 +122,32 @@ describe('Cross-referenced dashboards', () => {
...
-    };
-
-    cy.visitChartByParams(formData);
-    cy.wait('@getJson').its('response.statusCode').should('eq', 200);
-    cy.get('div.chart-container').contains(
-      'No results were returned for this query',
-    );
-  });
-});
+// describe('No Results', () => {
```
sadpandajoe, line 125: you can do a describe.skip here to skip all these tests

sadpandajoe:
one thing, if we're skipping tests, we should probably add a comment in the file to see why we are skipping all these tests.

.github/workflows/bashlib.sh
```diff
@@ -162,7 +162,17 @@ cypress-run-all() {
...
+
+  docker run --rm \
+    -e CYPRESS_BASE_URL \
+    -e PARALLELISM \
+    -e PARALLELISM_ID \
+    -e USE_DASHBOARD=${USE_DASHBOARD} \
+    -v "$GITHUB_WORKSPACE/superset-frontend/cypress-base:/e2e" \
+    -w /e2e \
+    cypress/browsers:node-20.16.0-chrome-127.0.6533.119-1 \
+    npx cypress run --browser chrome $USE_DASHBOARD_FLAG
```
mistercrunch, line 175: I think that mean means no parallelism, guessing you're not meaning to merge this change

PR #30405, opened 2024-09-26
fix: Incorrect hovered items in tooltips

### LINKED ISSUE #29565: [Dev build] Stacked chart tooltip highlights a different series than the one I'm hovering over
## Screenshot

image

## Description

On a stacked bar chart with data zoom enabled, when I hover over one of the series, the series that gets highlighted in the tooltip does not match the series my cursor is over.

To reproduce:
1. Create a stacked bar chart with multiple series, 
2. Sort stack by total
3. Make sure your dataset is one where the value for each series each day changes such that the series with the largest value is not always the same each day
4. Hover over one of the series, hover to the next series, etc. 
5. Observe which series is highlighted in the tooltip
6. You will see that as you go down the stacked bar, it just moves down the tooltip legend which is not

### LINKED ISSUE #30045: Chart tooltip and stacked bars are not in the same order
## Screenshot

image
(You can't see my mouse on the screenshot but I was hovering over the dark purple bar at the bottom.) 

_-- Sorry my dashboard is in French_ 

## Description

My issue is that the tooltip is not in the same order as the stacked bars. Looking at the bars, you would expect the legend to be in the following order: 
- Point that does not appear on the chart (Total) - Amount
- Light blue bar - Amount 
- Light purple bar - Amount 
- Dark purple bar - Amount 
Unfortunately the last three elements are inverted. 

**How did I setup this chart?**
I have built this chart with the `Mixed Chart` option. 
- Query A : Total amount, scatter plot with lowest opacity so that is doesn't show on the graph. 
- Query B : Total amount grouped by my dimension, bar chart.

**To resolve the issue I've tried to:** 
- Order the bars with the "Order By" option >> it doesn't affect my chart at all or the corresponding tooltip. 
 image
- Order the tooltip by metric in the Customize pane >> it works, the tooltip is sorted by the metric, but the bars are still in the same order   
image

Is there a way to fix this issue? 

Thanks!

michael-s-molina:
### SUMMARY
This PR fixes an issue with incorrect hovered items in tooltips. The problem was that the algorithm was not considering that some dimension/metric combinations might result in series with missing data points.

Fixes https://github.com/apache/superset/issues/29565
Fixes https://github.com/apache/superset/issues/30045

https://github.com/user-attachments/assets/cdc73641-4e2b-4d6d-8ca8-0d15a1bb00a9

https://github.com/user-attachments/assets/b3454e0e-d6f1-4605-ac70-99bc9bc4c6ca

### TESTING INSTRUCTIONS
Check the original issues for instructions.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -593,7 +593,7 @@ export default function transformProps(
           extractForecastValuesFromTooltipParams(forecastValue);
 
         const keys = Object.keys(forecastValues);
-        let focusedRow = -1;
+        let focusedRow;
         keys.forEach(key => {
           const value = forecastValues[key];
           // if there are no dimensions, key is a verbose name of a metric,
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -593,6 +593,7 @@ export default function transformProps(
           extractForecastValuesFromTooltipParams(forecastValue);
 
         const keys = Object.keys(forecastValues);
+        let focusedRow = -1;
         keys.forEach(key => {
           const value = forecastValues[key];
           // if there are no dimensions, key is a verbose name of a metric,
@@ -627,12 +628,11 @@ export default function transformProps(
               : tooltipFormatterSecondary,
           });
           rows.push(row);
+          if (key === focusedSeries) {
+            focusedRow = rows.length - 1;
+          }
         });
-        return tooltipHtml(
-          rows,
-          tooltipFormatter(xValue),
-          keys.findIndex(key => key === focusedSeries),
-        );
+        return tooltipHtml(rows, tooltipFormatter(xValue), focusedRow);
       },
     },
     legend: {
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -593,6 +593,7 @@ export default function transformProps(
           extractForecastValuesFromTooltipParams(forecastValue);
 
         const keys = Object.keys(forecastValues);
+        let focusedRow = -1;
```
villebro, line 596: nit: not a strong opinion, but as `focusedRow` already accepts `undefined`, I think it could be more explicit to set this to `undefined` here to clarify that it's in fact undefined.

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -569,6 +569,7 @@ export default function transformProps(
         const showTotal = Boolean(isMultiSeries) && richTooltip && !isForecast;
         const showPercentage = showTotal && !forcePercentFormatter;
         const keys = Object.keys(forecastValues);
+        let focusedRow = -1;
```
villebro, line 572: same here

villebro (approved):
LGTM with a non-blocking nit

justinpark (approved):

PR #30412, opened 2024-09-27
feat: cancel impala query on stop

wugeer:
feat: cancel sqllab impala query on stop
### SUMMARY

Currently, when executing an Impala query in SQL Lab, clicking the cancel button only cancels the query in Superset, but the query is not canceled on the Impala side, and the resources are not released.
This commit resolves the issue by ensuring that when an Impala query is canceled in SQL Lab, the query is also canceled on the Impala side, achieving the goal of releasing resources.

before
[impala_sql_lab_cancel_before.webm](https://github.com/user-attachments/assets/05750ebd-023b-4a2f-a6b6-652d2e082770)

after
[impala_cancel_query.webm](https://github.com/user-attachments/assets/80cdb661-1daa-47c7-9a00-aa4c8a43735e)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -443,7 +443,7 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
     # When this is changed to false in a DB engine spec it means the query id
     # is determined only after the specific query is executed and it will update
     # the `cancel_query` value in the `extra` field of the `query` object
-    is_query_id_associated_connect = True
+    has_query_id_before_execute = True
 
     @classmethod
     def is_oauth2_enabled(cls) -> bool:
@@ -1342,7 +1342,7 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         """
         logger.debug("Query %d: Running query: %s", query.id, sql)
         cls.execute(cursor, sql, query.database, async_=True)
-        if not cls.is_query_id_associated_connect:
+        if not cls.has_query_id_before_execute:
             cancel_query_id = query.database.db_engine_spec.get_cancel_query_id(
                 cursor, query
             )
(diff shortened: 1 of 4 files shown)


villebro (changes requested):

wugeer (commented):
Good! Done!

wugeer (commented):
Good! Done!

rusackas:
Only @villebro 's review can unblock this now... thanks to both of you for following up on this!

wugeer:
@villebro Could you please take some time to review my code? Thanks a lot!

superset/db_engine_specs/base.py
```diff
@@ -437,6 +437,14 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
     # Driver-specific exception that should be mapped to OAuth2RedirectError
     oauth2_exception = OAuth2RedirectError
 
+    # Does the query id related to the connection?
+    # The default value is True, which means that the query id is determined when
+    # the connection is created.
+    # When this is changed to false in a DB engine spec it means the query id
+    # is determined only after the specific query is executed and it will update
+    # the `cancel_query` value in the `extra` field of the `query` object
+    is_query_id_associated_connect = True
```
villebro, line 446: I like the comment, however, I find the property name difficult to understand. Could we rename this to something more descriptive, like `has_query_id_before_execute` or something similar.

villebro (commented):
Looking good, however, in the interest of legibility, I would prefer a slightly more explicit name for the new property.

wugeer:
> Looking good, however, in the interest of legibility, I would prefer a slightly more explicit name for the new property.

@villebro have Done! :)

villebro (approved):

PR #30417, opened 2024-09-27
chore: improve DML check

betodealmeida:
### SUMMARY

Small improvement: only check for DDL/DML in SQL Lab when the feature is disabled in the DB. This way we don't need to parse the query unnecessarily.

Also bump `sqlglot`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


dpgaspar (approved):

michael-s-molina:
@betodealmeida Can you add the `sqlglot` `CHANGELOG` to the PR description to help reviewers?

michael-s-molina:
@betodealmeida I noticed that there are many [breaking changes](https://github.com/tobymao/sqlglot/blob/main/CHANGELOG.md) between 23.0 and 25.24. Can any of these affect us?

betodealmeida:
> @betodealmeida I noticed that there are many [breaking changes](https://github.com/tobymao/sqlglot/blob/main/CHANGELOG.md) between 23.0 and 25.24. Can any of these affect us?

As far as I can tell, no. The breaking changes are either breaking internal APIs (which we don't rely heavily on), or breaking behavior that was previously incorrect.

PR #30425, opened 2024-09-28
fix(dashboard-export): Fixes  datasetId is not replaced with datasetUuid in Dashboard export in 4.1.x

### LINKED ISSUE #30424: Dashboard export does not export `datasetUuid` in native filter targets but `datasetId` in 4.1
### Bug description

When exporting any dashboard in 4.0.x, (tested with 4.0.2) the native filter configuration would be exported as:

```
[..]
    name: Column desc
    filterType: filter_select
    targets:
    - column:
        name: column_name
      datasetUuid: 7ac464dc-315a-4c9e-b9ec-048d7bbcf382
[..]
```

When exporting any dashboard in Superset 4.1.0RC2 it is exported as:

```
    name: Column desc
    filterType: filter_select
    targets:
    - datasetId: 17
      column:
        name: column_name
```

This works fine when exporting and just importing to the same instance of an empty Superset but leads to a corruption of the filter settings when datasets are already existing since the wrong dataset (based on `datasetId`) will used for a filter.

I did some research and the main (only relevant) change introduced is:
https://github.com/apache/superset/pull/26765
and it moved the code responsible for rewriting `datasetId` to `datasetUuid` away from the code that actually generates the content and in a path that is only responsible to find connected objects:
https://github.com/apache/superset/blame/2e4f6d3f38404b70f8d0324743c229a4917acaed/superset/commands/dashboard/export.py#L182-L194

I think this would need to be reintroduced in the `_file_content` method. I already prepared a patch internally and testing it now. If it works, I will try to provide a PR as soon as possible since this is a blocker for me to use 4.1.0RC2.

Note that there is already some duplication of code by separating generating the content from exporting the related objects. So, maybe this should be done differently than simply duplicating more code.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

Not applicable

### Node version

Not applicable

### Browser

Not applicable

### Additional context

Tested on 4.1.0RC2 

### Checklist

fmannhardt:
### SUMMARY

Fixes #30424 by duplicating the code that rewrites the `native_filter_configuration` to use `datasetUuid` instead of `datasetId` into the `_file_content` method of `ExportDashboardsCommand`. 

Before the `targets` attribute of `native_filter_configuration` items includes `datasetId`:

```
    targets:
    - datasetId: 17
      column:
        name: column_name
```

After it just has `datasetUuid`:

```
    targets:
    - column:
        name: column_name
      datasetUuid: 7ac464dc-315a-4c9e-b9ec-048d7bbcf382```
```

### TESTING INSTRUCTIONS

(1) Create a dashboard with any chart
(2) Add a single value filter for any column
(3) Export the dashboard
(4) Verify that `datasetUuid` is used to refer to the dataset and not `datasetId`

### ADDITIONAL INFORMATION

- [x] Has associated issue: 
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Could you write a unit test that would fail before and succeed after? That will help review the logic by stating the assumptions in a unit test.

fmannhardt:
Happy to do so. Since I have no clue of the testing approach of Superset, could you help me find the correct spot to test? Would the correct file be:
https://github.com/apache/superset/blob/cc9fd88c0d2a2406f202da7dcea2f0d14f1d017e/tests/unit_tests/commands/export_test.py#L42
or rather an integration test (since this also concerns code that also exports other resources:
https://github.com/apache/superset/blob/cc9fd88c0d2a2406f202da7dcea2f0d14f1d017e/tests/integration_tests/dashboards/commands_tests.py#L67

fmannhardt:
After some more searching, the example data used in the tests comes from here:
https://github.com/apache/superset/blob/master/superset/examples/world_bank.py
and are manually created via code right? Or is there somewhere examples that exists as YAML files using the import/export feature? 

Searching for `native` in the examples folder does not find anything:
https://github.com/search?q=repo%3Aapache%2Fsuperset+path%3A%2F%5Esuperset%5C%2Fexamples%5C%2F%2F+native&type=code

So, I would need to write code to create a dashboard with native filters from scratch? Or would there be a simpler way to write such test? I would need some guidance.

sadpandajoe:
@fmannhardt I created a [PR](https://github.com/apache/superset/pull/30495) to the covid vaccine example dashboard. I'm guessing you can probably add a test here: https://github.com/apache/superset/blob/master/tests/integration_tests/dashboards/commands_tests.py#L67 or update one of the tests to make sure when exporting native filter meta data is there.

Vitor-Avila:
Hey @fmannhardt, 

Thanks for working on this fix! 🙏 

Looking at the code changes, I noticed you have re-introduced that code block to `file_content`, but it also remains present in the `_export` method. Is it required to have this block in both? For instance I see the `dataset_id = target.pop("datasetId", None)` in both blocks.

If we need them in both places, should we consider to move this logic to a helper/etc method to make it DRYer?

fmannhardt:
@Vitor-Avila I think it is required in both places to also invoke the export of the related datasets via `yield from ExportDatasetsCommand([dataset_id]).run()`. I agree that a refactoring would make sense. However, then there are also other parts that are duplicated such as: 

``` 
        payload = model.export_to_dict(
            recursive=False,
            include_parent_ref=False,
            include_defaults=True,
            export_uuids=True,
``` 

It seems to me that a completely different approach to separate the export to YAML/JSON and walking the dependency graph of the to be exported dashboard would be better. Maybe, first collecting all the dependencies and then going through the list of dependencies and exporting them. However, that would exceed the capacity that I have for contributing right now (and I am also not aware of any design decision that led to the current approach).

I will try to add the test in the next few days.

sadpandajoe:
@fmannhardt @Vitor-Avila as this would be a regression for 4.1, let's I'd advise that we just add the tests and then if we need to refactor, refactor in a separate pr.

fmannhardt:
It seems that I also need to write a fixture for the COVID dashboard, right? Bare with me, still exploring the setup of Superset testing and locally many other tests fail probably due to missing dependencies.  I will try to look at this tomorrow.

villebro:
@fmannhardt thanks for working on this super critical regression. I will be reviewing this today and do my best to help unblock you.

villebro:
Note: this is a fix for a regression that was introduced by #26765

villebro:
> It seems that I also need to write a fixture for the COVID dashboard, right? Bare with me, still exploring the setup of Superset testing and locally many other tests fail probably due to missing dependencies. I will try to look at this tomorrow.

@fmannhardt you are correct - if there isn't a fixture that creates this dashboard + related assets, then you will need to add one. Let me know if you need help with this.

fmannhardt:
I may have time to look at this tomorrow (UTC+1). So, the existing fixtures hard code the dashboard and do not load from the examples? Just so I understand how to write it. It would be a copy of `world_bank_dashboard.py` but then the COVID dashboard code in there? 

In case, it is easy for you to add the fixture I am also happy to update the test I wrote to work against it.

villebro:
@fmannhardt I took a quick look at that particular fixture, and it seems to be mostly reusing existing logic. However, the data seems to be randomly generated for some reason. I'm not sure why, but maybe it's to avoid unnecessary web traffic. Please keep an open mind while working on the fixture, ad if you feel some aspect could be improved compared to the World Bank fixture, then please by all means feel free to implement accordingly. I'm in UTC-0700, so I can help/review tomorrow morning my time if needed, likely after you've had a chance to take a stab at it first.

fmannhardt:
@villebro I took another look but this would take me much more time to understand. I don't even need to data for the test. Only the have the COVID dashboard and its dataset loaded and available to be queries by UUID `f4065089-110a-41fa-8dd7-9ce98a65e250`. It is probably trivial but would take me several hours. So, it would take some more days. Please feel free to add this.

villebro:
@fmannhardt I totally understand. Feel free to disable the added test - I'll open a PR that adds the necessary fixtures and re-enables the test.

fmannhardt:
Test is disabled and current master branch is merged.

superset/commands/dashboard/export.py
```diff
@@ -190,6 +201,5 @@ def _export(
                 if dataset_id is not None:
                     dataset = DatasetDAO.find_by_id(dataset_id)
                     if dataset:
-                        target["datasetUuid"] = str(dataset.uuid)
```
villebro, line 193: At the risk of introducing yet more regressions, I feel it would be cleaner to do the following:
```python
if export_related:
    for target in native_filter...
```

The point being, no sense in iterating over all datasets, if `export_related` is `False`, as we're not doing anything else here (look at how charts are exported above).

villebro (approved):
LGTM, leaving open for another day so others can chime in. Note: I'll finalize the tests this week.

superset/commands/dashboard/export.py
fmannhardt: Done. Since this is now only to invoke the dataset export, it is fine to have the condition before iterating.

sadpandajoe (approved):
Spun up an ephemeral and verified import/export of native filters are working so going to approve

PR #30426, opened 2024-09-28
chore: run cypress headless in ci

eschutho:
### SUMMARY

Fixes CI for cypress by running it in headless mode.
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


scripts/cypress_run.py
```diff
@@ -61,7 +63,7 @@ def run_cypress_for_test_file(
         os.environ.pop("CYPRESS_RECORD_KEY", None)
         cmd = (
             f"{XVFB_PRE_CMD} "
-            f"{cypress_cmd} --browser {browser} "
+            f"{cypress_cmd} --browser {browser} --headless "
```
michael-s-molina, line 66: @eschutho @geido Would changing this setting to `--headless` break the screenshots used by Applitools and Cypress dashboard?

eschutho: I was testing to see if that was the case, but so far haven't seen any test failures related to it. I think it should be able to take screenshots headlessly, is my guess. This is actually the default flag, so this may already be running headlessly.

rusackas (approved):

scripts/cypress_run.py
```diff
@@ -44,7 +44,9 @@ def run_cypress_for_test_file(
     os.environ["TERM"] = "xterm"
     os.environ["ELECTRON_DISABLE_GPU"] = "true"
     build_id = generate_build_id()
-    browser = os.getenv("CYPRESS_BROWSER", "chrome")
+    browser = os.getenv("CYPRESS_BROWSER", "electron")
```
mistercrunch, line 47: a bit concerned to bring yet another engine in the mix, specifically for tests/automation. Supporting two is already a challenge, and thinking now that we should just pick one engine, either Chrome or Firefox, and use is consistently for thumbnail-generation, alert & reports and in e2e tests

mistercrunch: Also confused about what electron `--headless` means, isn't electron a "head" by definition?  Did a tiny bit of research just now, and Electron uses Blink (a fork of webkit), Chrome/Chromium use Blink as well, and FF uses Gecko. So chrome/chromium/electron in headless more should be pretty much the same AFAIK

eschutho:
Closing this as per the concerns about having a third engine in the mix.

PR #30438, opened 2024-09-30
fix(Filters): Apply native & cross filters on common columns

### LINKED ISSUE #30061: Cross-filtering is reloading unaffected charts

geido:
### SUMMARY
When in scope (global or specific) charts that are not applicable to the filter (do not share the same datasource or columns) were being applied anyway, causing unnecessary queries. This PR checks that the charts have columns in common or will exclude the charts.

### AFTER

https://github.com/user-attachments/assets/fe58ccbc-b689-4470-a85d-a8b494e9633c

https://github.com/user-attachments/assets/9a45c284-ac44-4f4e-960b-e66c0c5ed46d

### TESTING INSTRUCTIONS
1. Create a native filter that has no common columns with a chart
2. Add the chart in scope
3. Apply the filter
4. The chart should not reload and emit a query

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,108 @@
...
+): number[] {
+  const sourceSlice = slices[filterKey];
+  if (!sourceSlice) return [];
+
+  const sourceDatasource = datasources[sourceSlice.datasource];
+  if (!sourceDatasource) return [];
+
+  return Object.values(slices)
+    .filter(slice => {
+      if (slice.slice_id === Number[filterKey]) return false;
```
justinpark, line 63: Do you intend to `Number(filterKey)`?

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,108 @@
...
+            col.column_name === column?.name ||
+            col.column_name === column?.displayName,
+        )
+      );
+    })
+    .map(slice => slice.slice_id);
+}
+function getRelatedChartsForCrossFilter(
+  filterKey: string,
+  slices: Slice[],
```
justinpark, line 52: It appears slices is a Record not a list.

```suggestion
  slices: Record<string, Slice>,
```

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,108 @@
...
+          targetColumn => sourceColumn.column_name === targetColumn.column_name,
+        ),
+      );
+    })
+    .map(slice => slice.slice_id);
+}
+
+export function getRelatedCharts(
+  filters: Object,
+  slices: Slice[],
```
justinpark, line 79:
```suggestion
  slices: Record<string, Slice>,
```

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,108 @@
...
+function getRelatedChartsForSelectFilter(
+  filter: Filter,
+  slices: Slice[],
+  datasources: DatasourcesState,
+) {
+  return Object.values(slices)
+    .filter(({ datasource }) => {
+      const chartDatasource = datasources[datasource];
+      if (!chartDatasource) return false;
+      const { column, datasetId } = filter.targets[0];
```
justinpark, line 37: Let's make this safe in case of the empty case.
```suggestion
      const { column, datasetId } = filter.targets[0] ?? {};
```

superset-frontend/src/dashboard/util/activeAllDashboardFilters.ts
```diff
@@ -54,9 +54,13 @@ export const getAllActiveFilters = ({
       chartConfiguration?.[filterId]?.crossFilters?.chartsInScope ??
       allSliceIds ??
       [];
+    const filterType = nativeFilters?.[filterId]?.filterType ?? null;
```
justinpark, line 57: Do we have a handling for `null` case? Otherwise let's skip
```suggestion
    const filterType = nativeFilters?.[filterId]?.filterType;
```

geido:
Just realized that highlighting a filter that is not applicable should also be added as an enhancement for this PR. Working on that.

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,112 @@
+/**
```
michael-s-molina, line 1: Thanks for creating this file to keep the logic in one place.

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,112 @@
...
+  return Object.values(slices)
+    .filter(slice => {
+      if (slice.slice_id === Number(filterKey)) return false;
+      const targetDatasource = datasources[slice.datasource];
+      if (!targetDatasource) return false;
+      if (targetDatasource === sourceDatasource) return true;
+
+      return sourceDatasource.columns.some(sourceColumn =>
+        targetDatasource.columns.some(
+          targetColumn => sourceColumn.column_name === targetColumn.column_name,
```
michael-s-molina, line 70: We apply cross filters for charts from different datasets but with the same column name? What ensures that the values are the same? For example, I might have `product` defined in different datasets but the values are completely different.

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,112 @@
...
+export function getRelatedCharts(
+  filters: Object,
+  slices: Record<string, Slice>,
+  datasources: DatasourcesState,
+) {
+  return Object.entries(filters).reduce((acc, [filterKey, filter]) => {
+    if (filter.filterType === 'filter_select') {
+      return {
+        ...acc,
+        [filterKey]: getRelatedChartsForSelectFilter(
```
michael-s-molina, line 86: Is the scope not being considered then determining the related charts?

superset-frontend/src/dashboard/util/getRelatedCharts.ts
geido: The intent of these changes it to limit the number of unnecessary queries we send to the backend for processing. Right now we are sending any query as long as it is in scope. This is trying to prevent some of those. I believe making sure that filters are actually applicable by comparing the underlying data structures is still responsibility of the backend.

michael-s-molina: > The intent of these changes it to limit the number of unnecessary queries we send to the backend for processing.

I was understanding that we were going to achieve this by limiting the number of unnecessary requests to server. Given that each chart fires a request to server, if they don't belong to the same dataset or are out of scope, then that request would not be fired and the chart would not be loaded.

geido: @michael-s-molina I think we still want to allow that columns that belong to different datasets but share the same name might work? I think this was a functionality that also the UI seems to suggest and I don't think we should change it now without revising the functionality more holistically. I believe these changes will already limit a bunch of requests and be an improvement in that sense.

Screenshot 2024-10-03 at 17 29 19

Screenshot 2024-10-03 at 17 28 57

michael-s-molina: I didn't know this was already allowed. I agree it's important that these changes keep the compatibility.

superset-frontend/src/dashboard/util/getRelatedCharts.ts
geido: Good catch

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -80,12 +80,25 @@ export function getRelatedCharts(
   datasources: DatasourcesState,
 ) {
   return Object.entries(filters).reduce((acc, [filterKey, filter]) => {
+    const chartsInScope = filter.chartsInScope
```
kgabryje, line 83: `filter.chartsInScope?.filter(
           (chartId: number) => !filter.scope.excluded.includes(chartId),
         ) ?? filter.scope`

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -103,7 +103,7 @@ export function getRelatedCharts(
         ),
       };
     }
-    if (!filter.filterType) {
+    if (!filter.filterType && Object.keys(slices).includes(filterKey)) {
```
kgabryje, line 106: nit: we could probably simplify `Object.keys(slices).includes(filterKey)` to just `slices[filterKey]`

superset-frontend/src/dashboard/util/getRelatedCharts.ts
geido: Thanks. I think from looking at the unit tests there is an issue with this logic too. I am fixing it up.

kgabryje (approved):

geido:
Holding on this while testing more cases

michael-s-molina:
Some things I observed while testing

https://github.com/user-attachments/assets/5c8e861b-7f47-44a8-bb79-e291573d3315

- The first 2 charts should be affected by the filter but there's no filter indicator
- If you check the query, the filter was applied but the chart was not reloaded

michael-s-molina:
In this example, the last 2 charts are from different datasets but both have a `gender` column. Initially, the filter is correctly applied to both datasets but if I remove the last chart from the filter scope, it's still being affected.


https://github.com/user-attachments/assets/76dd859f-3adc-4c6b-b58d-347a2238c97f

michael-s-molina (changes requested):
I found some problems while testing with the ephemeral environment.

geido:
> I found some problems while testing with the ephemeral environment.

Thanks @michael-s-molina, I could not see many of these issues locally due to https://github.com/apache/superset/pull/30569 as I was under the false impression that I was testing the latest state of the code. I reverted that PR and now looking at these problems. Thanks for catching them!

michael-s-molina (approved):
I did my set of tests and was not able to find problems! Even datasets with calculated columns worked.

Nice work @geido. This PR should significantly reduce the number of requests to the server, reducing costs and improving performance. Thank you!

Given that this PR touches a critical part of Superset, it would be good to have more approvals before merging the PR.

sadpandajoe:
https://github.com/user-attachments/assets/7d08eec0-7d40-4df3-a124-87c60ae16096

I'm getting an unexpected server error on the FCC examples dashboard. Steps to repro:
1. go to the SQL lab
2. run a `select *` query on the fcc data
3. save it as a new dataset
4. create a new chart with the virtual dataset
5. add the chart to the fcc dashboard
6. go to the dashboard
7. click on a slice of it

superset-frontend/packages/superset-ui-core/src/query/types/Dashboard.ts
```diff
@@ -80,6 +80,24 @@ export type Filter = {
...
+export type AppliedCrossFilterType = {
+  filterType: undefined;
+  targets: number[];
+  scope: number[];
+} & AppliedFilter;
+
+export type AppliedNativeFilterType = {
+  filterType: 'filter_select';
+  scope?: number[] | NativeFilterScope; // TODO: inconsistent usage
+  targets: [Partial<NativeFilterTarget>];
```
justinpark, line 98: Do you intent to have this?

```suggestion
  targets: Partial<NativeFilterTarget>[];
```

justinpark (approved):
LGTM other than one minor type issue.

superset-frontend/src/dashboard/util/getRelatedCharts.ts
```diff
@@ -0,0 +1,200 @@
...
+  isAppliedCrossFilterType,
+  isAppliedNativeFilterType,
+  isNativeFilter,
+} from '@superset-ui/core';
+import { Slice } from 'src/types/Chart';
+import { DatasourcesState } from '../types';
+
+function checkForExpression(formData?: Record<string, any>) {
+  const groupby = ensureIsArray(formData?.groupby) ?? [];
+  const allColumns = ensureIsArray(formData?.all_columns) ?? [];
```
michael-s-molina, line 34:
```suggestion
  const groupby = ensureIsArray(formData?.groupby);
  const allColumns = ensureIsArray(formData?.all_columns);
```

michael-s-molina (approved):

sadpandajoe:
Just a quick question, should it affect all charts if it's from the same database table even if one is virtual? Example here: http://35.88.232.227:8080/superset/dashboard/11/?native_filters_key=vORa9YtGn5rLPgi2AjgirnIJB_30m_cpUE3kzlqWFgMGjPIw3CKoFrTKjNNAUkan

1. go to sql lab
2. run the query 
```
SELECT * 
FROM covid_vaccines 
WHERE country_name = "United States"
```
3. save as a new dataset
4. copy one of the charts that use covid_vaccines
5. switch the dataset to the copied one
6. put this on the covid vaccines dashboard
7. do a cross filter on the newly created dataset chart

right now following this step, all charts are affected by the virtual dataset.

sadpandajoe (approved):
Spoke with @yousoph and it sounds like my case is invalid so going to approve.

Mainer-g00t:
All our dashboards utilize cross-filters across various virtual datasets, and we use Jinja to capture and apply filters, even if the filtered columns in the filter are not present in the dataset.
For example: 
- dataset & chart 1: city, number of habitants
- dataset & chart 2: profession, number of proffesionals
We filter between both using the filter in the WHERE statement. The reason for having different datasets is that the cardinality in our tables is very high and a single dataset ruins performance.

I'm afraid that if this is applied automatically to every case could basically break all of our dashboards

fmannhardt:
Not sure if this should go here or to a new issue (and I won't manage to find time to create a reproducible example in the next week). However, I re-tested this feature together with cross filters and the following setup:
- A renamed column (colum_a AS "Colum A") is used as source for the cross filter from chart A (with data source A)
- A second chart B from a different data source B has the same underlying column (column_a) but this column is only used in a filter of that chart that is based on an JINJA expression (using  get_filters() as explained here: https://superset.apache.org/docs/configuration/sql-templating/)
- Result is that the cross filter on chart A does not affect chart B.

I looked through the code and it, in fact, does not consider searching for JINJA in the filter of other charts.

villebro:
I didn't review these changes, but it was my understanding that
1. by default, native and cross filters will only be scoped to charts using the same dataset
2. If one wants the filters to apply to charts that aren't using the same dataset, one simply adds them to the filter scope

Similar to the comments above, I know of many use cases where there's jinja logic that intercepts filters for columns that aren't present in the dataset, and use those to modify the chart query somehow. I'm not sure if this has been explicitly stated as a feature that Superset supports, but I think this use case has been well established for a longer time already, hence I feel removing it will be problematic.

Does this PR break said workflow? I'm happy to jump on a sync call to discuss this if needed.

villebro:
@Mainer-g00t @fmannhardt as you raised concerns here, did you find that this PR broke your use cases?

villebro:
> > @Mainer-g00t @fmannhardt as you raised concerns here, did you find that this PR broke your use cases?
> 
> Yes, it does. In our case, charts export filters that are picked up by jinja logic to filter dataset with data that is not part of the columns in the dataset. Specifically, our dataset looks like:
> 
> * customer (1k), products (1k), providers(1k), metrics...
> 
> This leads to a roughly 500M row table. We split it in 3 datasets, one for each dimension and use x-filters to select on the other datasets. ie, clicking on one customer on the first table will filter all products and providers for that customer and so on.
> 
> This allows us to display relationships in big datasets breaking it up in dimensions and using x-filters

Just to confirm, have you tested the changes here and can confirm that they do indeed break your use case? Because to me it seems like you should still be able to do the same cross filtering even after this change, as long as you've defined your filter scopes correctly.

Mainer-g00t:
> > > @Mainer-g00t @fmannhardt as you raised concerns here, did you find that this PR broke your use cases?
> > 
> > 
> > Yes, it does. In our case, charts export filters that are picked up by jinja logic to filter dataset with data that is not part of the columns in the dataset. Specifically, our dataset looks like:
> > 
> > * customer (1k), products (1k), providers(1k), metrics...
> > 
> > This leads to a roughly 500M row table. We split it in 3 datasets, one for each dimension and use x-filters to select on the other datasets. ie, clicking on one customer on the first table will filter all products and providers for that customer and so on.
> > This allows us to display relationships in big datasets breaking it up in dimensions and using x-filters
> 
> Just to confirm, have you tested the changes here and can confirm that they do indeed break your use case? Because to me it seems like you should still be able to do the same cross filtering even after this change, as long as you've defined your filter scopes correctly.

Yes, we just tested it on 4.1 and breaks cross filtering for us. It does not apply filters if the column is not in the dataset even when it is used in the WHERE statement with jinja

fmannhardt:
Thanks. Just to confirm it does break our use case in 4.1rc3 for two underlying reasons:

(1) The column filtering is renamed with a label (colx as X). So, cross filtering does consider it as X.

(2) We used JINJA to create a custom filter condition based on selected filters (on X) but since there there is no column X in the dataset, it won't do the filtering on the chart. Btw, it does filter correctly when clicking on show the data as table or as SQL statement. So, the filtering is inconsistent.

I think the code checking for expressions (quoted above) simply does not check there WHERE filters defined on the charts. It may also have issues with custom charts.
________________________________

michael-s-molina:
Hi @fmannhardt @Mainer-g00t @geido. May I suggest opening a specific GitHub issue for this so we can track it [here](https://github.com/orgs/apache/projects/418)?

Mainer-g00t:
> Hi @fmannhardt @Mainer-g00t @geido. May I suggest opening a specific GitHub issue for this so we can track it [here](https://github.com/orgs/apache/projects/418)?

Hello, sure, I'll try to do it today with an example of the behaviour before and after the change

fbgeobit:
Hi @michael-s-molina , 
We opened a new issue with use case commented by @Mainer-g00t : https://github.com/apache/superset/issues/30687

Thanks

michael-s-molina:
Thank you! I added the issue to the [4.1 board](https://github.com/orgs/apache/projects/418).

PR #30441, opened 2024-09-30
fix: battling cypress' dashboard feature

mistercrunch:
Following up on a series of PRs to try and fix cypress run, latest being this one https://github.com/apache/superset/pull/30430.

Here I'm working a bit in the dark since the logic on `master` in CI is slightly different from the logic in PRs.

On my latest try, I had an error related to a combination of retries in the context where the dashboard is used. Issue was around the group id's not being unique. Here I'm working around it by adding the attempt number to the group's id, which should fix it.

At this point I'm beyond annoyed with Cypress. Last try before I shut down using the dashboard feature (a paying service we don't really get much if any benefit from) at all.

Previous error for reference:
Screenshot 2024-09-30 at 9 26 33 AM


mistercrunch:
This ran as expected except all 5 retries ran and failed, trying to break down `nativeFilters.test.ts` into 2 files... hopefully the second test group doesn't depend on the first one...

rusackas:
> At this point I'm beyond annoyed with Cypress. Last try before I shut down using the dashboard feature (a paying service we don't really get much if any benefit from) at all.

angry-burn

mistercrunch:
OK THIS IS IT! The CYPRESS SAGA IS OVER. MERGE! MERGE! MERGE! MERGE!

betodealmeida (approved):
Go go go

mistercrunch:
hizzie-cat

PR #30443, opened 2024-09-30
fix(Jinja metric macro): Support Drill By and Excel/CSV download without a dataset ID

Vitor-Avila:
### SUMMARY
The Jinja `{{metric()}}` macro allows you to retrieve the corresponding SQL for a metric by using its `key`, and optionally the dataset ID. When the dataset ID is not specified, it should be automatically retrieved from the context (based on the current dataset). 

This PR improves the logic to retrieve this value from other places:
* It was not possible to use the Drill By option in a chart that had this macro without the dataset ID
* It was not possible to export a chart using this macro without a specified dataset ID to Excel/CSV

**Drill By**

**BEFORE**
image

**AFTER**
image

**CSV/Excel Export**

**BEFORE**
image

**AFTER**
User prompted to download the file

### TESTING INSTRUCTIONS
Tests updated accordingly. For manual testing:
1. Create a Pie Chart using this macro and a dimension. 
2. Save it to a dashboard.
3. Access the dashboard and Drill By on a dimension with any other column.
4. Then try exporting the chart to Excel.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -25,7 +25,7 @@ from functools import lru_cache, partial
 from typing import Any, Callable, cast, Optional, TYPE_CHECKING, TypedDict, Union
 
 import dateutil
-from flask import current_app, has_request_context, request
+from flask import current_app, g, has_request_context, request
 from flask_babel import gettext as _
 from jinja2 import DebugUndefined, Environment
 from jinja2.sandbox import SandboxedEnvironment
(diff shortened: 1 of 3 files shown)


adamcunnington-mlg:
Thanks for fixes!

superset/jinja_context.py
```diff
@@ -846,35 +846,45 @@ def dataset_macro(
...
+        if payload := request.get_json(cache=True) if request.is_json else None:
+            if dataset_id := payload.get("datasource", {}).get("id"):
+                return dataset_id
+            form_data.update(payload.get("form_data", {}))
+        request_form = loads_request_json(request.form.get("form_data"))
+        form_data.update(request_form)
+        request_args = loads_request_json(request.args.get("form_data"))
+        form_data.update(request_args)
+
+    if final_form_data := (form_data or getattr(g, "form_data", {})):
```
eschutho, line 874: It looks like we're updating form_data above but then reassigning it here?  Do we need another variable for form_data? It could cause confusions as to which one to use later.

Vitor-Avila: you're right, we don't need an additional variable here. Let me fix that. 🙏

eschutho (approved):

PR #30447, opened 2024-09-30
fix(explore): don't discard controls on deprecated

justinpark:
### SUMMARY
Currently, if a column or metric name changes in Explore, any previously selected control items disappear. This can lead to confusion during debugging, as the control items mentioned in the error message are not actually displayed.
This commit modifies explore to display a warning message in such cases, allowing the user to manually delete and modify the control items.

Before:

https://github.com/user-attachments/assets/b0debd40-7ae1-4986-ae67-f89675f5742b

After:

https://github.com/user-attachments/assets/210b4f3c-f3ec-4e71-9624-e5d01b8c6534

### TESTING INSTRUCTIONS
Create a chart with a dataset
Select a column / a metric in the control and then save
Edit the dataset and then change the name of the above selected column/metric
Revisit the chart and then see the error message

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/components/ColumnOption.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/components/ColumnOption.tsx
@@ -51,7 +51,8 @@ export function ColumnOption({
 }: ColumnOptionProps) {
   const { expression, column_name, type_generic } = column;
   const hasExpression = expression && expression !== column_name;
-  const warningMarkdown = column.warning_markdown || column.warning_text;
+  const warningMarkdown =
+    column.warning_markdown || column.warning_text || column.error_text;
   const type = hasExpression ? 'expression' : type_generic;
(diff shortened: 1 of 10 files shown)


mistercrunch:
Minor comment from watching the screen caps, but I'm thinking an error exclamation-point-in-a-red-circle would be better than the yellow warning you used. That yellow warning is use to raise attention to something that's not breaking like the other place we use the warning is when we want to show that "this filter is inherited from the dashboard context and won't be saved ..."

justinpark:
Updated the indicator with the error icon.

Screenshot 2024-10-01 at 2 00 31 PM

mistercrunch (approved):
LGTM

justinpark:
@sadpandajoe could you add this fix on 4.1?

PR #30490, opened 2024-10-01
fix(Explore): Apply RLS at column values

geido:
### SUMMARY
When applying a filter, column values would show all possible values without respecting RLS filters. This PR fixes that.

### AFTER

https://github.com/user-attachments/assets/da96b25c-b3ba-46bb-beec-df09bb6ed6ee

### TESTING INSTRUCTIONS
1. Apply RLS filters to an account
2. Create a new chart using the target dataset
3. Create a filter on a Table chart
4. The user should only see the values of the column with RLS applied

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila (approved):
LGTM! Thanks for this improvement!

superset/models/helpers.py
```diff
@@ -1350,6 +1350,10 @@ def values_for_column(
         if self.fetch_values_predicate:
             qry = qry.where(self.get_fetch_values_predicate(template_processor=tp))
 
+        rls_filters = self.get_sqla_row_level_filters(template_processor=tp)
```
geido, line 1353: This needs to check for the current column

superset/models/helpers.py
```diff
@@ -1350,6 +1377,9 @@ def values_for_column(
         if self.fetch_values_predicate:
             qry = qry.where(self.get_fetch_values_predicate(template_processor=tp))
 
+        if rls_filters := self.get_rls_filters_for_column(column_name, tp):
+            qry = qry.where(or_(*rls_filters))
```
mistercrunch, line 1381: are we sure the content of the `OR` clause will be wrapped in parenthesis?

superset/models/helpers.py
```diff
@@ -1314,7 +1315,33 @@ def get_time_filter(  # pylint: disable=too-many-arguments
             )
         return and_(*l)
 
-    def values_for_column(
+    def get_rls_filters_for_column(
```
mistercrunch, line 1318: Seems this logic must exist somewhere else already, are we duplicating logic?

superset/models/helpers.py
betodealmeida: I'm pretty sure SQLAlchemy does the right thing here, but a unit test would be good. Preferably one where one of the filters is `cond1 AND cond2`, ensuring the generated SQL is:

```sql
(cond1 AND cond2) OR cond3
```

And not:

```sql
cond1 AND cond2 OR cond3
```

superset/models/helpers.py
betodealmeida: Not using `sqlglot`. Ideally this logic would live in `superset/sql/parse.py`, since we're trying to centralize all the parsing and SQL manipulation logic there. But I'm in the process of converting the RLS logic implemented in `sqlparse` to use `sqlglot`, and I can do that.

superset/sql/parse.py
```diff
@@ -367,6 +367,27 @@ def get_settings(self) -> dict[str, str | bool]:
             for eq in set_item.find_all(exp.EQ)
         }
 
+    def get_predicates(self) -> list[exp.Predicate]:
```
betodealmeida, line 370: Can you add this to the base class as well? A dummy implementation raising `NotImplementedError`.

superset/sql/parse.py
```diff
@@ -367,6 +367,27 @@ def get_settings(self) -> dict[str, str | bool]:
...
+
+        if not where_clauses:
+            return []
+
+        for where_clause in where_clauses:
+            where_predicates = where_clause.find_all(exp.Predicate)
+            for pred in where_predicates:
+                predicates.append(pred)
+
+        return predicates
```
betodealmeida, line 389: You can simplify this to:

```python
return  [
    predicate
    for where in self._parsed.find_all(exp.Where)
    for predicate in where.find_all(exp.Predicate)
]
```

Since `find_all` always returns a list (empty on no match).

superset/models/helpers.py
```diff
@@ -1314,7 +1315,33 @@ def get_time_filter(  # pylint: disable=too-many-arguments
...
+        self, column_name: str, template_processor: BaseTemplateProcessor
+    ) -> list[TextClause]:
+        rls_filters_for_column = []
+
+        if rls_filters := self.get_sqla_row_level_filters(template_processor):
+            parsable_statement = "SELECT 1 WHERE"
+
+            for rls_filter in rls_filters:
+                parsable_statement += f" {str(rls_filter)} "
```
betodealmeida, line 1327: Here I think @mistercrunch's concern is valid: if an `rls_filter` is `cond1 OR cond2` then the generated SQL will look like:

```sql
SELECT 1 WHERE cond1 OR cond2 AND cond3
```

Where we want:

```sql
SELECT 1 WHERE (cond1 OR cond2) AND cond3
```

I think it's safer to write a helper function in `superset/sql/parse.py` that takes the `rls_filter` directly and parses that. Note that you can parse fragments with `parse_one`:

```python
>>> from sqlglot import parse_one
>>> sqlglot.parse_one('a=b')
EQ(
  this=Column(
    this=Identifier(this=a, quoted=False)),
  expression=Column(
    this=Identifier(this=b, quoted=False)))
```

betodealmeida (dismissed):
Left some comments on how to improve the logic.

Vitor-Avila:
I've just tested in the ephemeral and it worked properly for me both when combining multiple RLS conditions with the same group key, or a different group key. Thanks @geido!

Vitor-Avila (approved):

superset/models/helpers.py
```diff
@@ -1350,6 +1350,10 @@ def values_for_column(
         if self.fetch_values_predicate:
             qry = qry.where(self.get_fetch_values_predicate(template_processor=tp))
 
+        rls_filters = self.get_sqla_row_level_filters(template_processor=tp)
+        if rls_filters:
+            qry = qry.where(and_(*rls_filters))
```
betodealmeida, line 1355: I *think* that you don't need the `if rls_filters:` check here, since SQLAlchemy does the smart thing if `rls_filters` is an empty list:

```python
>>> str(select(text('t')).where(and_(*[])))
'SELECT t'
```

So you could simply do:

```suggestion
        qry = qry.where(and_(*rls_filters))
```

betodealmeida (approved):
Nice!

villebro:
Oof.. Nice catch + fix @geido!

PR #30491, opened 2024-10-01
fix(embedded): sankey charts

### LINKED ISSUE #30205: Superset 4.1.0RC2 - New Sankey charts cannot be embedded = "Guest user cannot modify chart payload"
### Bug description

When adding the new ECharts version of the Sankey to a dashboard that is embedded using the embedding SDK using a guest user token the chart will not render with error `Guest user cannot modify chart payload` 

### How to reproduce the bug

1. Create a new Sankey chart 
2. Create a new dashboard and add the Sankey chart to the new dash
3. Embed the dashboard into a web app using the embed functionality in Superset to get the embedding ID and add it using the embedding SDK. 
4. Create a guest token using the embedding SDK 
5. Load the page as a guest user (loading chart not logged in the superset UI works fine)

### Screenshots/recordings

image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

betodealmeida:
### SUMMARY

A saved Sankey chart query context has the `source` and `target` columns in the `groupby` field:

```python
{
    "queries": [
        {
            ...
            "columns": [],
            "metrics": ["count"],
            "groupby": ["bot_id", "channel_id"],  # here
         },
    ],
}
```

But the query context from the same chart, when embedded, looks like this:

```python
{
    "queries": [
        {
            ...
            "columns": ["bot_id", "channel_id"],
            "metrics": ["count"],
            # no "groupby" key
         },
    ],
}
```

This PR modifies the `query_context_modified` function to account for this case.

Note that this approach is a short term solution to unblock 4.1. I think two better (but more complext) solutions would be:

1. Make sure the query object from an unmodified chart in a dashboard always matches the stored query object.
2. Even better: dashboards should send the chart ID and any native filters, instead of sending the whole payload for the chart. This would prevent maliciously modifying requests over the wire.

Fixes https://github.com/apache/superset/issues/30205

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -177,7 +177,11 @@ def query_context_modified(query_context: "QueryContext") -> bool:
     )
 
     # compare columns and metrics in form_data with stored values
-    for key in ["metrics", "columns", "groupby"]:
+    for key, equivalent in [
+        ("metrics", ["metrics"]),
+        ("columns", ["columns", "groupby"]),
+        ("groupby", ["columns", "groupby"]),
+    ]:
         requested_values = {freeze_value(value) for value in form_data.get(key) or []}
         stored_values = {
             freeze_value(value) for value in stored_chart.params_dict.get(key) or []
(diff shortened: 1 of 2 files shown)


michael-s-molina (approved):

rusackas:
Yaaaaay! It's almost 4.1 go-time! 🚀

PR #30496, opened 2024-10-01
docs: fix broken links in CI

mistercrunch:
^^^ title, CI link validator was firing with those links

Seen on another PR ->
Screenshot 2024-10-01 at 4 32 11 PM


rusackas (approved):
Thanks!!!

rusackas:
I think that's the first time I've seen HTTP code 999 in the wild. That's pretty interesting. I'll look at the other failing links and open another PR soon to clean up some of this stuff. I think there's also more work to do that it's not parsing and checking relative links in the markdown files, but I/we can improve that soon.

PR #30500, opened 2024-10-02
fix(sqllab): Remove redundant scrolling

justinpark:
### SUMMARY
This commit fixes the redundant vertical scrolling in SQL Lab caused by the bottom resize handlers, which are not used for the sidebar resizer.

Before:

https://github.com/user-attachments/assets/bac25cbf-448d-4b3c-8021-06e3eb300c8f

After:

https://github.com/user-attachments/assets/8182ce66-a880-41d1-a9aa-bfa936af26f2

### TESTING INSTRUCTIONS
Go to SQL Lab and then check the scrollbar in the query panel.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
+++ b/superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
@@ -163,10 +163,14 @@ const StyledToolbar = styled.div`
   }
 `;
 
-const StyledSidebar = styled.div<{ width: number; hide: boolean | undefined }>`
+const StyledSidebar = styled.div<{
+  width: number;
+  hide: boolean | undefined;
+  padding: number;
+}>`
   flex: 0 0 ${({ width }) => width}px;
   width: ${({ width }) => width}px;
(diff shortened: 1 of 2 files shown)


superset-frontend/src/SqlLab/components/SqlEditor/index.tsx
```diff
@@ -940,12 +945,14 @@ const SqlEditor: FC<Props> = ({
           minWidth={SQL_EDITOR_LEFTBAR_WIDTH}
           initialWidth={SQL_EDITOR_LEFTBAR_WIDTH}
           enable={!hideLeftBar}
+          padding={leftBarPadding}
```
michael-s-molina, line 948: @justinpark Why does a left bar padding influences on a wrapper height? If it was a top/bottom padding, I would understand but a left one?

michael-s-molina: If this variable mean that a top padding is applied when the left bar is open, then it would be a good idea to rename the variable to make it more clear.

michael-s-molina (approved):

PR #30502, opened 2024-10-02
fix: Open control with Simple tab selected when there is no column selected

michael-s-molina:
### SUMMARY
This PR fixes the behavior when opening a column-related Explore control and no column was selected. Previously, if the dataset had any saved column, the Saved tab would be the default selected tab. This created a lot of confusion for our users as generally the number of saved columns is significantly smaller than the number of dataset columns. Also, the users are generally looking for the normal dataset columns and didn't notice that they had to switch to the `Simple` tab to find them. With this PR, the behavior will be:
- If there's no selected column, the Simple tab will be selected by default
- If there is a selected column, the tab that corresponds to its type will be selected by default

I also added tests for each use case.

https://github.com/user-attachments/assets/f7eddbb7-d11c-4ad8-90a1-958bc3dee9cb

https://github.com/user-attachments/assets/ac72ff6f-8865-4f27-8001-a193c8d099d4

### TESTING INSTRUCTIONS

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


justinpark (approved):
LGTM

Issue #30383, opened 2024-09-25
Dashboard Import error


JatinVenkatesh:
### Bug description

I was performing some exercise on the export and import of dashboards and databases. First, I exported the database and removed the database from the database connection. Then, when I try to import the database using the same zip file, I get the following error:.
![image](https://github.com/user-attachments/assets/e355a7e3-b271-4584-8b47-424cd5bf38a2)

Same goes for the dashboard as well:
![image](https://github.com/user-attachments/assets/6f27942b-efac-42a1-961a-8a88be568a13)

### How to reproduce the bug

1. Export a dashboard.
2. Delete the dashboard.
3. Try importing it.
4. Same for the database.

### Screenshots/recordings

Database error:
![image](https://github.com/user-attachments/assets/e355a7e3-b271-4584-8b47-424cd5bf38a2)

Dashboard error:
![image](https://github.com/user-attachments/assets/8f7e9af9-b4a6-4f95-ab88-ec7328c81420)

### Superset version

4.0.2

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [ ] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [ ] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

fisjac:
What type of database were you attempting to import?

rusackas:
Also curious if you see any more error logs/stacktraces ont the frontend or in the server logs.

JatinVenkatesh:
> What type of database were you attempting to import?

PostgreSQL

JatinVenkatesh:
> Also curious if you see any more error logs/stacktraces ont the frontend or in the server logs.

2024-09-30 10:11:53,610:INFO:werkzeug:127.0.0.1 - - [30/Sep/2024 10:11:53] "GET /health HTTP/1.1" 200 -
2024-09-30 15:42:00 Command failed validation
2024-09-30 15:42:00 2024-09-30 10:12:00,113:INFO:superset.commands.dashboard.importers.dispatcher:Command failed validation
2024-09-30 15:42:00 CommandException
2024-09-30 15:42:00 Traceback (most recent call last):
2024-09-30 15:42:00   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
[...]

Issue #19222, opened 2022-03-16
Import is not working: An error occurred while importing dashboard: Import dashboard failed for an unknown reason


Narendra678:
Hi Team,

I am trying to import dashboard getting below error.
An error occurred while importing dashboard: Import dashboard failed for an unknown reason

Note: when try to import i can see dataset added
when i try to import chart it is working.
I am able to import other reports

Regards,
Naren

AAfghahi:
Hello! Could you take some screenshots of what you are experiencing?

Narendra678:
![image](https://user-images.githubusercontent.com/88739186/158947761-8b49b19d-c2f4-40e9-b8fe-0ea8a47365eb.png)

Narendra678:
2022-03-18 07:29:47 default[1-3-1qa]  Error running import command
2022-03-18 07:29:47 default[1-3-1qa]  Traceback (most recent call last):    File "/workspace/superset/commands/importers/v1/__init__.py", line 67, in run      self._import(db.session, self._configs, self.overwrite)    File "/workspace/superset/dashboards/commands/importers/v1/__init__.py", line 126, in _import      config = update_id_refs(config, chart_ids, dataset_info)    File "/workspace/superset/dashboards/commands/importers/v1/utils.py", line 85, in update_id_refs      for old_id, columns in metadata["filter_scopes"].items()    File "/workspace/superset/dashboards/commands/importers/v1/utils.py", line 85, in <dictcomp>      for old_id, columns in metadata["filter_scopes"].items()  KeyError: 580
2022-03-18 07:29:47 default[1-3-1qa]
2022-03-18 07:29:47 default[1-3-1qa]  During handling of the above exception, another exception occurred:
2022-03-18 07:29:47 default[1-3-1qa]
2022-03-18 07:29:47 default[1-3-1qa]  Traceback (most recent call last):    File "/workspace/superset/dashboards/commands/importers/dispatcher.py", line 58, in run      command.run()    File "/workspace/superset/commands/importers/v1/__init__.py", line 71, in run      raise self.import_error()  superset.dashboards.commands.exceptions.DashboardImportError: Import dashboard failed for an unknown reason
2022-03-18 07:29:47 default[1-3-1qa]  2022-03-18 07:29:47,754:ERROR:superset.dashboards.commands.importers.dispatcher:Error running import command
[...]

PR #30503, opened 2024-10-03
fix(imports): Error when importing charts / dashboards with missing DB credentials

fisjac:
### SUMMARY

When importing charts, dashboards, and databases, certain databases (BigQuery) require credentials to be entered into the encrypted extra field for authorization. When importing this type of database, or any chart or dashboard that relies on this type of database, the db creation will fail generating an import failed error.

This fix implements a try except block to catch connection errors when attempting to populate db catalogs when importing a db. It prevents bigquery exceptions from being remapped to generic Exception types.

Before:
https://www.loom.com/share/ca5efd11ba004c19aea9c83484bdab47?sid=6be35dfc-42e1-45b2-81dd-0a8df1d41dd6

After:
https://www.loom.com/share/a9cc247ddb434ee198f81503514f6776?sid=93c7481f-9ace-4def-9862-6567ecf2f32b

### TESTING INSTRUCTIONS

Export a bigquery database or a chart/dashboard linked to a bigquery db.
Delete the underlying db
Attempt to import the db, chart, or dashboard.
The import should be successful, even though the db does not have auth creds

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/30383
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/importers/v1/utils.py
+++ b/superset/commands/database/importers/v1/utils.py
@@ -83,8 +83,6 @@ def import_database(
         add_permissions(database, ssh_tunnel)
     except Exception:
         logger.info("unable to connect to %s", database)
-    finally:
-        return database
 
     return database
 
--- a/superset/commands/database/importers/v1/utils.py
+++ b/superset/commands/database/importers/v1/utils.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import logging
 from typing import Any
 
 from superset import app, db, security_manager
@@ -25,6 +26,8 @@ from superset.exceptions import SupersetSecurityException
 from superset.models.core import Database
 from superset.security.analytics_db_safety import check_sqlalchemy_uri
 from superset.utils import json
+
+logger = logging.getLogger(__name__)
 
 
 def import_database(
@@ -64,7 +67,7 @@ def import_database(
     # Before it gets removed in import_from_dict
     ssh_tunnel_config = config.pop("ssh_tunnel", None)
 
-    database = Database.import_from_dict(config, recursive=False)
+    database: Database = Database.import_from_dict(config, recursive=False)
     if database.id is None:
         db.session.flush()
 
@@ -75,7 +78,13 @@ def import_database(
         ssh_tunnel = None
 
     # TODO (betodealmeida): we should use the `CreateDatabaseCommand` for imports
-    add_permissions(database, ssh_tunnel)
+
+    try:
+        add_permissions(database, ssh_tunnel)
+    except Exception:
+        logger.info("unable to connect to %s", database)
+    finally:
+        return database
 
     return database
 
@@ -89,6 +98,7 @@ def add_permissions(database: Database, ssh_tunnel: SSHTunnel) -> None:
             cache=False,
             ssh_tunnel=ssh_tunnel,
         )
+
         for catalog in catalogs:
             security_manager.add_permission_view_menu(
                 "catalog_access",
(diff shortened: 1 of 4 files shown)


superset/commands/database/importers/v1/utils.py
```diff
@@ -75,7 +79,11 @@ def import_database(
         ssh_tunnel = None
 
     # TODO (betodealmeida): we should use the `CreateDatabaseCommand` for imports
-    add_permissions(database, ssh_tunnel)
+
+    try:
+        add_permissions(database, ssh_tunnel)
+    except SupersetDBAPIConnectionError as ex:
```
Antonio-RiveroMartnez, line 85: This might occur by other connection errors and not only the credentials one, right? So we would be affecting other engines too and not just BigQuery

Antonio-RiveroMartnez: Could we instead use the [custom_errors in the db_engine_spec](https://github.com/apache/superset/blob/master/superset/db_engine_specs/bigquery.py#L183) and map the credentials one to something we could target individually?

fisjac: Correct, this would affect other DBs. While this bug emerged in the specific case of BigQuery, there are other DBs that may error as well when trying to set up catalogs on import. Regardless of the DB, the inability to add catalogs shouldn't necessarily import to fail since catalogs can be added later when updating the DB with valid creds.

superset/db_engine_specs/bigquery.py
```diff
@@ -738,7 +738,7 @@ def _get_fields(cls, cols: list[ResultSetColumnType]) -> list[Any]:
     @classmethod
     def parse_error_exception(cls, exception: Exception) -> Exception:
         try:
-            return Exception(str(exception).splitlines()[0].strip())
+            return type(exception)(str(exception).splitlines()[0].strip())
```
betodealmeida, line 741: This method `parse_error_exception` is very confusing... I think it's really unlikely that `str(exception).splitlines()[0].strip()` will raise an exception — it only happens if `str(exception)` is `''` or if there's some shennanigans in the exception class `__str__` method.

I think what the method is trying to do is limit the exception message to the first line, but there are better ways to do that. But it's OK, we can refactor later.

superset/commands/database/importers/v1/utils.py
betodealmeida: Ideally we'd only allow the exception if the database is configured for OAuth2. But even more ideally we'd do what the TODO says, and use `CreateDatabaseCommand` for the imports, since the command already handles edge cases like this. So I think this is OK for now.

betodealmeida (approved):
Looks good!

superset/db_engine_specs/bigquery.py
fisjac: The issue I was running into was when a mapped `SupersetDBAPIError` was being passed into this method, it would then be remapped to a base class `Exception` type with the message preserved. I was trying to preserve the Exception type as well as the message

devikaaaa33:
Hello @, I am also facing same issue. I am using version 4.1.1. I tried the steps below, but they did not work. Developer tools show the error as "{

{
"errors": [
{
"message": "Import dashboard failed for an unknown reason",
"error_type": "GENERIC_COMMAND_ERROR",
"level": "error",
"extra": {
"issue_codes": [
{
"code": 1010,
"message": "Issue 1010 - Superset encountered an error while running a command."
}
]
}
}
]
}"

Hi - I know this is probably really old, however, I made this work :- importing and export from v4.01 to 4.1rc2. Simply by moving all the files in the nested directory from the export zip to the root, then zipping and uploading to superset.. Hope this helps others. This attempted is a direct result from the helpful code above (Dosu).

https://github.com/apache/superset/issues/19222#issuecomment-2340944113
Can someone look into this and provide a solution?

PR #30506, opened 2024-10-03
fix(dashboard filters): add "previous quarter" as option

personofnorank:
### SUMMARY

The custom date range filter should include previous quarter along with previous calendar week, month or year
### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/26651
- [ ] Required feature flags:
- [x] Changes UI


personofnorank:
@sfirke as per Slack discussion, thank you!

sfirke:
Thank you! This repository requires PR titles to follow Conventional Commit format (I think it's called). For this one it could be: 

fix(dashboard filters): add "previous quarter" as option

That will cause CI failure until fixed. 

I'm off work today but maybe someone else will review, otherwise I'll look when I'm back.

superset-frontend/src/explore/components/controls/DateFilterControl/types.ts
```diff
@@ -80,10 +80,12 @@ export type CommonRangeType =
 
 export const PreviousCalendarWeek = 'previous calendar week';
 export const PreviousCalendarMonth = 'previous calendar month';
+export const PreviousQuarter = 'previous quarter';
```
michael-s-molina, line 83:
```suggestion
export const PreviousCalendarQuarter = 'previous calendar quarter';
```

superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
```diff
@@ -78,6 +78,10 @@ export const parseDttmToDate = (
         now.setMonth(now.getMonth() - 1);
       }
       return now;
+    case 'previous quarter':
```
michael-s-molina, line 81: Could you add test cases for this calculation? For example, the month of May should be quarter number 2.

personofnorank: I can add tests, but the purpose here isn't to return a quarter number. It's to add the option for "previous calendar quarter" to a date filter (and obviously calculate that correctly). Currently the "custom" options include previous week, month and year but not quarter.
image

personofnorank: Done :)

superset-frontend/packages/superset-ui-core/src/time-comparison/getTimeOffset.ts
```diff
@@ -78,6 +78,11 @@ export const parseDttmToDate = (
         now.setMonth(now.getMonth() - 1);
       }
       return now;
+    case 'previous calendar quarter': {
```
michael-s-molina, line 81: @personofnorank You can add tests to the` getTimeOffset.test.ts` file.

personofnorank: Done :)

michael-s-molina: Hi @personofnorank. There are problems with this logic as it can result in an invalid month value (e.g., -3, -2, -1) when the current quarter is the first quarter of the year.

I suggest adding some tests like these:

```
test('should return the first day of the previous calendar quarter when current date is in Q4', () => {
  const result = getTimeOffset('previous calendar quarter', false, now);
  expect(result).toEqual(new Date('2023-07-01T00:00:00Z'));
});

test('should return the first day of the previous calendar quarter when current date is in Q3', () => {
  now = new Date('2023-08-15T00:00:00Z');
  const result = getTimeOffset('previous calendar quarter', false, now);
  expect(result).toEqual(new Date('2023-04-01T00:00:00Z'));
});

test('should return the first day of the previous calendar quarter when current date is in Q2', () => {
  now = new Date('2023-05-15T00:00:00Z');
  const result = getTimeOffset('previous calendar quarter', false, now);
  expect(result).toEqual(new Date('2023-01-01T00:00:00Z'));
});

test('should return the first day of the previous calendar quarter when current date is in Q1', () => {
  now = new Date('2023-02-15T00:00:00Z');
  const result = getTimeOffset('previous calendar quarter', false, now);
  expect(result).toEqual(new Date('2022-10-01T00:00:00Z'));
});
```

personofnorank: It should work because Date allows a negative month.
Date(2016, -1) gives you December 2015.

sfirke:
Looks like there is a pre-commit linting check failing, it says to delete two spaces I think.  And then a test failing.  Can you investigate the CI failures when you get a chance, @personofnorank ?

rusackas:
Re-running CI :D

personofnorank:
@justinpark I think this is pending your review - please let me know if anything needed from my side, with apologies for any oversight as I'm new to the process.

sfirke:
Looks like there is a failing Python test: https://github.com/apache/superset/actions/runs/11278958608/job/31368926066?pr=30506

personofnorank:
Fixed a typo in the test script

sfirke:
Looks like tests are failing with a function getting too many arguments

personofnorank:
In the last run the code seemed to be working but test was expecting wrong thing... maybe this time :)

sfirke:
It's getting closer to passing tests 🙏

sfirke:
@personofnorank I forgot this is still open, ack! Can we try to get it done for 5.0.0?

personofnorank:
@sfirke sorry for slow response, just did a commit, is it too late for the release (assuming it works this time)?

personofnorank:
Seems to be failing because of "E501 Line too long" in superset/utils/date_parser.py, although that file is full of `# pylint: disable=line-too-long,useless-suppression`

sfirke:
Thanks for picking this back up! I just re-ran those failing tests in case some were just flaky and failing for unrelated reasons. I agree that adding the `pylint: disable` is appropriate, it matches what has been done above, and I don't see why it's not working.

sfirke:
This one at least looks like a real failing test:
```
 FAIL packages/superset-ui-core/test/time-comparison/getTimeOffset.test.ts (16.271 s)
  ● should handle custom range with previous calendar quarter

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Array [
    -   "5 days ago",
    +   "86 days after",
      ]

      48 |     includeFutureOffsets,
      49 |   });
    > 50 |   expect(result).toEqual(expected_result);
         |                  ^
      51 |   timezoneMock.unregister();
      52 | };
      53 |

      at toEqual (packages/superset-ui-core/test/time-comparison/getTimeOffset.test.ts:50:18)
      at Object.runTimezoneTest (packages/superset-ui-core/test/time-comparison/getTimeOffset.test.ts:537:3)

```

From https://github.com/apache/superset/actions/runs/13585311794/job/38000863290?pr=30506#step:4:22. Can you investigate?

personofnorank:
Yes, I had a syntax error (== instead of ===) which would have meant the code didn't execute causing any functional test to fail. I've fixed that now. If it fails again it won't be for that reason!

sfirke:
Good catch! Did you push that change? I'll approve re-running the tests but don't see the new commit

personofnorank:
Oops, thought I'd done a push - just done now.

sfirke:
For the line length, I wonder if adding `# noqa: E501` to the end of the line would help? Maybe that comment string needs to occur before the line length hits its limit and that's why it's not working right now? I'm looking at https://docs.astral.sh/ruff/rules/line-too-long/#error-suppression and https://stackoverflow.com/a/77396292

personofnorank:
Thanks, did a new push.

personofnorank:
I'm a bit stumped. It's failing on the tests in `superset-frontend/packages/superset-ui-core/test/time-comparison/getTimeOffset.test.ts` which I added for the "previous calendar quarter" use case as they exist for all the other use cases. 

I basically cloned the other tests but I'm not sure what exactly they're doing or whether it matters what value of `startDate` you set. Setting different values leads to different values in the error message, which implies I could hack it to get the right value but that doesn't feel right without understanding what it's doing.

LevisNgigi:
Hey, was wondering if this is a duplicate of this? https://github.com/apache/superset/issues/31870

sfirke:
> Hey, was wondering if this is a duplicate of this? #31870

Great catch @LevisNgigi, I think yes that issue and specifically the PR https://github.com/apache/superset/pull/31889 made this one obsolete.

I'm glad to see the functionality will be in Superset 5.0.0 but I'm sorry @personofnorank that your hard work did not make it in. A lesson to project committers to try to get PRs like this across the finish line faster so they don't get superseded.

personofnorank:
Well, an earlier issue can't be a duplicate of a later one :) but I'm glad this is getting fixed, and it was all good experience so thank you to everyone.

PR #30521, opened 2024-10-05
chore: enable lint PT009 'use regular assert over self.assert.*'

mistercrunch:
I just got bit by some old school self.assertEqual which is not good as the pytest-augmented, simple assert. Just leveraging the rule, running automated --fix for things it can fix and adding NOQA to the few it didn't auto-fix


villebro (approved):
Nice! that whole "we stole it from JUnit and couldn't bother to make it idiomatic python" assertion style always makes me cringe..

image

PR #30522, opened 2024-10-05
fix(SQL Lab): hang when result set size is too big

anamitraadhikari:
### SUMMARY
Users were encountering performance issues in SQL Lab, where large query result payloads caused the browser to hang. This issue was noticeable even when result sets contained only a few rows (e.g., 1-10 rows), but the payload was quite large. The large payloads overwhelmed the browser’s ability to render the data, resulting in crashes or freezing.

Before:
Screenshot 2024-10-06 at 3 05 03 AM

After:
Screenshot 2024-10-06 at 3 11 17 AM

### TESTING INSTRUCTIONS
Test by running a query that generates a large-sized result (e.g., > 100 MB), and verify that an appropriate error message is triggered if the result size becomes too large to handle efficiently. The size limit, after which the handling logic is triggered, can be defined in the Superset config file under SQL_LAB_PAYLOAD_MAX_MB.

### ADDITIONAL INFORMATION
NA


requirements/base.txt
```diff
@@ -408,3 +408,4 @@ zipp==3.19.0
     # via importlib-metadata
 zstandard==0.22.0
     # via flask-compress
+PyHive==0.7.0
```
villebro, line 411: Unrelated change?
```suggestion
PyHive==0.7.0
```

superset/sql_lab.py
```diff
@@ -214,7 +215,7 @@ def execute_sql_statement(  # pylint: disable=too-many-statements, too-many-loca
         insert_rls = (
             insert_rls_as_subquery
             if database.db_engine_spec.allows_subqueries
-            and database.db_engine_spec.allows_alias_in_select
+               and database.db_engine_spec.allows_alias_in_select
```
villebro, line 218: I assume this will be flagged by the linter
```suggestion
           and database.db_engine_spec.allows_alias_in_select
```

superset/sql_lab.py
```diff
@@ -601,6 +603,23 @@ def execute_sql_statements(
                 serialized_payload = _serialize_payload(
                     payload, cast(bool, results_backend_use_msgpack)
                 )
+
+                # Check the size of the serialized payload
+                if config.get("SQL_LAB_PAYLOAD_MAX_MB"):
+                    serialized_payload_size = sys.getsizeof(serialized_payload)
+                    sql_lab_payload_max_mb = config["SQL_LAB_PAYLOAD_MAX_MB"]
```
villebro, line 610: nit: walrus operator can be used to simplify:
```suggestion
                if sql_lab_payload_max_mb := config.get("SQL_LAB_PAYLOAD_MAX_MB"):
                    serialized_payload_size = sys.getsizeof(serialized_payload)
```

superset/sql_lab.py
```diff
@@ -601,6 +603,23 @@ def execute_sql_statements(
...
+                    max_bytes = sql_lab_payload_max_mb * 1024 * 1024
+
+                    if serialized_payload_size > max_bytes:
+                        logger.info("Result size exceeds the allowed limit.")
+                        raise SupersetErrorException(
+                            SupersetError(
+                                message=f"Result size ({serialized_payload_size / (1024 * 1024):.2f} MB) exceeds the allowed limit of {sql_lab_payload_max_mb} MB.",
```
villebro, line 617: nit: we're rewriting `1024 * 1024` here and in many places after this. To keep the codebase DRY, let's define it only once as a const.

tests/unit_tests/sql_lab_test.py
```diff
@@ -125,6 +130,90 @@ def test_execute_sql_statement_with_rls(
     SupersetResultSet.assert_called_with([(42,)], cursor.description, db_engine_spec)
 
 
+def test_execute_sql_statement_exceeds_payload_limit_log_check(mocker: MockerFixture, caplog) -> None:
+    """
+    Test for `execute_sql_statements` when the result payload size exceeds the limit,
+    and check if the correct log message is captured without raising the exception.
+    """
```
villebro, line 137: What is the purpose of this test? Shouldn't we always raise an exception if the payload size exceeds the threshold?

villebro (changes requested):
First pass comments

tests/unit_tests/sql_lab_test.py
anamitraadhikari: Updated the test to raise exception and assert on that.

tests/unit_tests/sql_lab_test.py
```diff
@@ -125,6 +127,128 @@ def test_execute_sql_statement_with_rls(
...
+            start_time=None,
+            expand_data=False,
+            log_params={},
+        )
+
+    # Check if the exception message is correct
+    assert (
+        "Result size" in str(excinfo.value)
+        and "exceeds the allowed limit" in str(excinfo.value)
+    ), f"Expected exception message about exceeding the limit not found. Actual message: {excinfo.value}"
```
villebro, line 189: To keep things simple, I'd just remove this assertion (I think asserting that the exception is raised is good enough)

tests/unit_tests/sql_lab_test.py
```diff
@@ -125,6 +127,128 @@ def test_execute_sql_statement_with_rls(
...
+    {"SQLLAB_PAYLOAD_MAX_MB": 50},  # Set the desired config value for testing
+)
+def test_execute_sql_statement_exceeds_payload_limit(mocker: MockerFixture) -> None:
+    """
+    Test for `execute_sql_statements` when the result payload size exceeds the limit,
+    and check if the correct log message is captured without raising the exception.
+    """
+
+    sql_statement = "SELECT 42 AS answer"
+    query_id = 1
```
villebro, line 141: nit: as these aren't reused anywhere, and are fairly simple, I'd almost just inline them directly on lines 176-177

tests/unit_tests/sql_lab_test.py
```diff
@@ -125,6 +127,128 @@ def test_execute_sql_statement_with_rls(
...
+    {"SQLLAB_PAYLOAD_MAX_MB": 50},  # Set the desired config value for testing
+)
+def test_execute_sql_statement_within_payload_limit(mocker: MockerFixture) -> None:
+    """
+    Test for `execute_sql_statements` when the result payload size is within the limit,
+    and check if the flow executes smoothly without raising any exceptions.
+    """
+
+    sql_statement = "SELECT 42 AS answer"
+    query_id = 1
```
villebro, line 203: nit: same here, let's just inline them

villebro (approved):
Looks great! A few optional nits, LMKWYT

Issue #27585, opened 2024-03-20
When trying to install Superset on python3.12 I get the error:


Teesmo:
When trying to install Superset on python3.12 I get the error:
`AttributeError: module 'pkgutil' has no attribute 'ImpImporter'.`

I believe this can be traced back to Numpy version < 1.26.0:
https://github.com/numpy/numpy/issues/23808

Currently apache-superset is pinned to 1.23.5 while the Numpy v1.26.0 was released this past September.  Is it possible to bump numpy?

_Originally posted by @ddxv in https://github.com/apache/superset/discussions/25963_

michael-s-molina:
Hi @Teesmo. Python 12 is not supported yet as you can see [here](https://github.com/apache/superset/blob/ad7bd09819d3fd6ac8f669c6f884551cbd66bcd6/setup.py#L216). Eventually, we'll support it and have a compatible Numpy version 😉

amotl:
Hi. We are also blocked on this, related to the very strict version pinning on NumPy. Maybe it could be relaxed just a bit?

> Currently apache-superset is pinned to 1.23.5 while the Numpy v1.26.0 was released this past September. Is it possible to bump numpy?

> [NumPy 1.26 is available for Python 3.12](https://pypi.org/project/numpy/1.26.4/#files), so maybe upgrading to that version from NumPy 1.23 might be feasible, because it's not too different?

-- https://github.com/crate/cratedb-examples/pull/336#issuecomment-2489696461

PR #30523, opened 2024-10-05
docs: Clarification on which command to use on which Ubuntu version.

kkovacs:
### SUMMARY
Clarification on which command to use on which Ubuntu version.

### ADDITIONAL INFORMATION
None.

### DIFF
--- a/docs/docs/installation/pypi.mdx
+++ b/docs/docs/installation/pypi.mdx
@@ -22,13 +22,13 @@ level dependencies.
 
 **Debian and Ubuntu**
 
-The following command will ensure that the required dependencies are installed:
+In Ubuntu **before 20.04** the following command will ensure that the required dependencies are installed:
 
 ```bash
 sudo apt-get install build-essential libssl-dev libffi-dev python-dev python-pip libsasl2-dev libldap2-dev default-libmysqlclient-dev
 ```
 
-In Ubuntu 20.04 the following command will ensure that the required dependencies are installed:
+In Ubuntu **20.04 and 22.04** the following command will ensure that the required dependencies are installed:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/docs/installation/pypi.mdx
```diff
@@ -22,13 +22,13 @@ level dependencies.
...
 
-The following command will ensure that the required dependencies are installed:
+In Ubuntu **before 20.04** the following command will ensure that the required dependencies are installed:
 
 ```bash
 sudo apt-get install build-essential libssl-dev libffi-dev python-dev python-pip libsasl2-dev libldap2-dev default-libmysqlclient-dev
 ```
 
-In Ubuntu 20.04 the following command will ensure that the required dependencies are installed:
+In Ubuntu **20.04 and 22.04** the following command will ensure that the required dependencies are installed:
```
michael-s-molina, line 31: What happens if it's a Ubuntu version that is not in the range? Maybe add this additional context.

sfirke (changes requested):
Minor change needed but thanks for clarifying this, it will be an improvement!

docs/docs/installation/pypi.mdx
kkovacs: Sadly, apparently superset can't be installed yet on Ubuntu 24.04. TLDR: the included `numpy` is too old.

Longer story:

- Ubuntu 24.04 has `python3.12`.
- `numpy` below 1.26.4 is incompatible with `python3.12`.
- Superset requires `numpy==1.23.5`.

Apparently, there was an issue about this earlier, but it was closed without resolution: https://github.com/apache/superset/issues/27585

sfirke (approved):

sfirke:
Thanks for this contribution @kkovacs !

PR #30527, opened 2024-10-05
docs: Docs update

Karan3304:
hello @bgreenlee ,I am new to this project and was just going through through the documentation and found some little mistakes and just tried to make them correct,i hope it will add a small bit of perfection in the documentation

### DIFF
--- a/docs/docs/configuration/alerts-reports.mdx
+++ b/docs/docs/configuration/alerts-reports.mdx
@@ -12,7 +12,7 @@ Users can configure automated alerts and reports to send dashboards or charts to
 - *Alerts* are sent when a SQL condition is reached
 - *Reports* are sent on a schedule
 
-Alerts and reports are disabled by default. To turn them on, you need to do some setup, described here.
+Alerts and reports are disabled by-default. To turn them on, you need to do some setup, described here.
 
 ## Requirements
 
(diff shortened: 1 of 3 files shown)


package-lock.json
sadpandajoe, line 1: why did the package-lock.json file get added?

docs/docs/configuration/alerts-reports.mdx
```diff
@@ -12,7 +12,7 @@ Users can configure automated alerts and reports to send dashboards or charts to
 - *Alerts* are sent when a SQL condition is reached
 - *Reports* are sent on a schedule
 
-Alerts and reports are disabled by default. To turn them on, you need to do some setup, described here.
+Alerts and reports are disabled by-default. To turn them on, you need to do some setup, described here.
```
sadpandajoe, line 15: I think the original is fine here, haven't seen it do `by-default`.

sfirke:
I appreciate your desire to contribute to Superset, @Karan3304 !  I'm going to close this pull request, though.  It doesn't make substantive improvements, and I think actually detracts from the Superset project because it consumes reviewer time that could be used for other PRs.

Please use Superset, get familiar with it, and then we would welcome improvements that do more than tinker with wording.  A good starting point might be to update screenshots on this page: https://superset.apache.org/docs/using-superset/creating-your-first-dashboard/  If you compare what's there now to Superset 4.0.2 or 4.1.0rc2 (the latter is probably the source you should use), you will differences.

Karan3304:
thank you @sfirke,for pointing me to the right direction

PR #23421, opened 2023-03-20
fix(#23176): adding URL decoding to SQLAlchemy URI

RekhaZemoso:
### SUMMARY

Using urllib we can overcome this issue, now it is working fine

### TESTING INSTRUCTIONS

set the superser config like
SQLALCHEMY_DATABASE_URI = "postgres://postgres:s3kr1t@localhost:5432/postgres?application_name=superset&options=--search_path%3D%22%24user%22,superset"

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes https://github.com/apache/superset/issues/23176

PR #30532, opened 2024-10-06
fix(migration): replace unquote with double percentages

### LINKED ISSUE #23176: superset db upgrade mishandles DATABASE_URL with % in it
Superset passes DATABASE_URL to Alembic by setting it as a value in a ConfigParser instance.

ConfigParser values are interpolated: https://docs.python.org/3/library/configparser.html#interpolation-of-values

This results in error message (or wrong behavior) if the DATABASE_URL contains %.

DATABASE_URL containing % is needed for some Postgres features, for example setting the schema where Superset metadata is stored: `postgres://postgres:s3kr1t@localhost:5432/postgres?application_name=superset&options=--search_path%3D%22%24user%22,superset`

#### How to reproduce the bug

1. Set in superset config:

```
SQLALCHEMY_DATABASE_URI = "postgres://postgres:s3kr1t@localhost:5432/postgres?application_name=superset&options=--search_path%3D%22%24user%22,superset"
```

2. Run `superset db upgrade` 

### Expected results

Superset medata tables created inside Postgres schema named `superset`.

### Actual results

```
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/app/superset/extensions/../migrations/env.py", line 45, in <module>
    config.set_main_option("sqlalchemy.url", DATABASE_URI)
  File "/usr/local/lib/python3.8/site-packages/alembic/config.py", line 231, in set_main_option
    self.set_section_option(self.config_ini_section, name, value)
  File "/usr/local/lib/python3.8/site-packages/alembic/config.py", line 258, in set_section_option
    self.file_config.set(section, name, value)
  File "/usr/local/lib/python3.8/configparser.py", line 1201, in set
    super().set(section, option, value)
  File "/usr/local/lib/python3.8/configparser.py", line 894, in set
    value = self._interpolation.before_set(self, section, option,
  File "/usr/local/lib/python3.8/configparser.py", line 402, in before_set
    raise ValueError("invalid interpolation syntax in %r at "
ValueError: invalid interpolation syntax in 'postgres://postgres:s3kr1t@localhost:5432/postgres?application_name=superset&options=--search_path%3D%22%24user%22,superset' at position 97
```

### Environment

- superset version: `0.0.0-dev` (docker image 3c95779d0029 from 2023-02-21)

### Checklist

### Additional context

How to fix it:

```
diff --git i/superset/migrations/env.py w/superset/migrations/env.py
index d0220e33f..274a4e24a 100755
--- i/superset/migrations/env.py
+++ w/superset/migrations/env.py
@@ -42,7 +42,11 @@ if "sqlite" in DATABASE_URI:
         "SQLite Database support for metadata databases will \
         be removed in a future version of Superset."
     )
-config.set_main_option("sqlalchemy.url", DATABASE_URI)
+config.set_main_option(
+    "sqlalchemy.url",
+    # Avoid ConfigParser interpolation.
+    DATABASE_URI.replace('%', '%%'),
+)
 target_metadata = Base.metadata  # pylint: disable=no-member

```

villebro:
### SUMMARY

#23421 introduced a regression, where passwords containing `@` characters would break migrations. To avoid having to decode the URL, we can simply escape the connection string for variable interpolation by doubling `%` chars, after which the original encoded connection string works as expected.

Note, that the change is functionally identical to the change proposed in the linked issue.

Here's one of many similar StackOverflow threads: https://stackoverflow.com/questions/39849641/in-flask-migrate-valueerror-invalid-interpolation-syntax-in-connection-string-a

### TESTING INSTRUCTIONS
This has been tested with a Postgres connection with a huge amount of special characters in the password, including `@` that previously broke the migration workflow.

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes #23176
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/env.py
+++ b/superset/migrations/env.py
@@ -16,7 +16,6 @@
 # under the License.
 import logging
 import time
-import urllib.parse
 from logging.config import fileConfig
 
 from alembic import context
@@ -43,8 +42,9 @@ if "sqlite" in DATABASE_URI:
         "SQLite Database support for metadata databases will \
         be removed in a future version of Superset."
     )
-decoded_uri = urllib.parse.unquote(DATABASE_URI)
-config.set_main_option("sqlalchemy.url", decoded_uri)
+# Escape % chars in the database URI to avoid interpolation errors in ConfigParser
+escaped_uri = DATABASE_URI.replace("%", "%%")
+config.set_main_option("sqlalchemy.url", escaped_uri)
 target_metadata = Base.metadata  # pylint: disable=no-member
 
 


giftig (approved):
It's important to note that the original problem here is the string being passed to a python-style interpolator accepting tokens like `%s`, and *not* anything to do with URLs. The original fix (mostly) worked by eliminating `%`s in the URL by decoding URL escape sequences like `%20`, which prevented the interpolator complaining about invalid interpolation sequences. However that leaves the URL improperly encoded, resulting in other issues popping up, and does not fix the problem where `%`s still appear in the string, e.g. when the password contains a literal `%`.

See the documentation link from @villebro 's linked stack overflow thread: https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config.set_main_option

Simply doubling the `%`s is the correct solution here so LGTM.

superset/migrations/env.py
```diff
@@ -43,8 +42,9 @@
         "SQLite Database support for metadata databases will \
         be removed in a future version of Superset."
     )
-decoded_uri = urllib.parse.unquote(DATABASE_URI)
-config.set_main_option("sqlalchemy.url", decoded_uri)
+# Escape % chars in the database URI to avoid interpolation errors in ConfigParser
+escaped_uri = DATABASE_URI.replace("%", "%%")
```
mistercrunch, line 46: Do we not need to `urllib.parse.unquote` anymore?

giftig: See my comment above; this was never needed, it was actually a bit of a dodgy hack.

mistercrunch (approved):

PR #30549, opened 2024-10-08
fix(Jinja): Extra cache keys to consider vars with set

geido:
### SUMMARY
The regex for `ExtraCache` that checks whether the statement has extra cache key calls via `has_extra_cache_key_calls` wasn't considering cases where the callable vars might not be in double curly brackets, such as:

```
{% set customer_id = current_user_email() %}
select '{{customer_id}}' as test;
```

Because of this issue, cached queries for charts would show previously cached data potentially exposing information from other users.

### TESTING INSTRUCTIONS
1. Create a dataset using 
```
{% set customer_id = current_user_email() %}
select '{{customer_id}}' as test;
```
2. Create a table chart based on the dataset
3. Add the chart to a Dashboard
4. Login with another user
5. You should see the current user email

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/jinja_context.py
```diff
@@ -104,13 +104,11 @@ class ExtraCache:
...
     regex = re.compile(
-        r"\{\{.*("
-        r"current_user_id\(.*\)|"
+        r"(current_user_id\(.*\)|"
         r"current_username\(.*\)|"
         r"current_user_email\(.*\)|"
         r"cache_key_wrapper\(.*\)|"
-        r"url_param\(.*\)"
-        r").*\}\}"
+        r"url_param\(.*\))"
```
dpgaspar, line 111: nit: we could be more strict and make sure we have:
{{ ... current_username() }} or {% current_username() %}

villebro: I agree with Daniel. We should check for both valid cases `{{` and `{%` to avoid false positives (although I admit they're going to be rare).

tests/integration_tests/sqla_models_tests.py
```diff
@@ -164,6 +164,24 @@ def test_extra_cache_keys(self, mock_user_email, mock_username, mock_user_id):
...
+              '{{ user_email }}' as email
+            """,
+            database=get_example_database(),
+        )
+
+        query_obj = dict(**base_query_obj, extras={})
+        extra_cache_keys = table1.get_extra_cache_keys(query_obj)
+        assert table1.has_extra_cache_key_calls(query_obj) is True
+        assert set(extra_cache_keys) == {1, "abc", "abc@test.com"}
+
```
villebro, line 184: bycatch - this test is huge, and could do with some heavy parametrization. Not necessarily something that has to be done here, but it would leave the codebase in a much tidier state.

villebro (changes requested):
Great catch! I don't agree with removing the check for the opening bracket - let's rather make it check for both.

villebro:
This comment looks mildly concerning.. Although to be fair, this probably isn't a problem in practice.
image

villebro (approved):
Beautiful test refactoring! ❤️ LGTM

neuromantik33:
Hello. This PR breaks are existing usage of RLS (thank heavens one for our campus users notified us). This is our input string:
```sql
campus_id IN UNNEST( {{ cache_key_wrapper(current_user_campus_ids()) }} )
```
which doesn't match using the pattern : `cache_key_wrapper\([^()]*\)`, 
but does using the old one (v4.0.2): `cache_key_wrapper\(.*\)`.

Is there somewhere in the docs on how we can manage this? The function we use looks as follows:

```python
def current_user_campus_ids() -> list[int]:
    """
    Get the list of campus ids associated with the current user
    """
    user: FortyTwoUser = g.user
    return user.campus_ids()
```
and is populated using the oauth2 jwt from keycloak. Any help would be much appreciated.

PR #30559, opened 2024-10-09
feat(formatting): Add memory units adaptive formatter to format bytes

mkopec87:
Add new memory units adaptive formatter to format bytes units into kB, MB, GB, etc.

### SUMMARY
Memory unit (bytes) is typically formatted into kB (kilobytes), MB (megabytes), etc.
Superset formatting options don't enable adaptive formatting with proper unit naming in case of bytes values.
There are two most commonly used multi-byte unit types (according to [wikipedia](https://en.wikipedia.org/wiki/Byte)), this PR implements them both.

image

### TESTING INSTRUCTIONS

1. Create a Chart, for example Table, with at least one numeric data column.
2. Edit the Chart, in the "Customize" tab select one of "Memory in bytes" options for formatting of your numeric column.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


villebro (commented):
This is a great feature, and one that many have been looking forward to seeing. In general this looks good, but would you be able to add some unit tests?

mkopec87:
@villebro , sure thing, let me give it a try :) I never wrote any Javascript before, therefore happy to get some feedback :)
CI setup is very useful here.

mkopec87:
@villebro I added unit tests and fixed some bugs on the way :) Let me know what you think!

villebro:
It seems there's a linting issue with the test:
image

superset-frontend/packages/superset-ui-core/test/number-format/factories/createMemoryFormatter.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+  it('formats bytes in human readable format with default options', () => {
+    const formatter = createMemoryFormatter();
+    expect(formatter(0)).toBe('0B');
+    expect(formatter(50)).toBe('50B');
+    expect(formatter(555)).toBe('555B');
+    expect(formatter(1000)).toBe('1kB');
+    expect(formatter(1111)).toBe('1.11kB');
+    expect(formatter(1024)).toBe('1.02kB');
+    expect(formatter(1337)).toBe('1.34kB');
```
villebro, line 35: lol

superset-frontend/packages/superset-ui-core/src/number-format/factories/createMemoryFormatter.ts
```diff
@@ -0,0 +1,55 @@
...
+) {
+  const { description, id, label, binary, decimals = 2 } = config;
+
+  return new NumberFormatter({
+    description,
+    formatFunc: value => {
+      if (value === 0) return '0B';
+
+      const sign = value > 0 ? '' : '-';
+      value = Math.abs(value);
```
villebro, line 39: Just create a new `const` and use that to get around the linting error:
```suggestion
      const absValue = Math.abs(value);
```

villebro (commented):
LGTM after the linting issue is fixed

superset-frontend/packages/superset-ui-core/test/number-format/factories/createMemoryFormatter.test.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+import { NumberFormatter, createMemoryFormatter } from '@superset-ui/core';
+
+describe('createMemoryFormatter()', () => {
+  it('creates an instance of MemoryFormatter', () => {
+    const formatter = createMemoryFormatter();
+    expect(formatter).toBeInstanceOf(NumberFormatter);
+  });
+  it('formats bytes in human readable format with default options', () => {
```
villebro, line 27: Oh, one more thing, we try to steer away from nested tests. So instead of doing `describe('foo)` with many `it('bar')` and `it('baz')`, just do individual `test('foo bar')` and `test('foo baz')`

superset-frontend/packages/superset-ui-core/src/number-format/factories/createMemoryFormatter.ts
mkopec87: changed as suggested

superset-frontend/packages/superset-ui-core/test/number-format/factories/createMemoryFormatter.test.ts
mkopec87: sure, done

villebro (approved):
Looks great, fingers crossed for CI to pass 🤞 Only comment is it'd be great to be able to customize the decimals, but that's outside the scope of this PR. I believe there's a discussion about this lying around somewhere, and would be a great improvement (the idea would be to be able to provide arbitrary config objects to the formatters), but that would require a SIP, so let's leave that for a future improvement.

mkopec87:
Thanks @villebro for your help merging this :)

WilliamGeuns1:
@villebro any idea when this will be in the release?

rusackas:
As far as I know it didn't make the cut opportunity for 4.1, so it'd most likely be in either a 4.2 (if that happens) or 5.0.

PR #30560, opened 2024-10-09
fix: Unable to parse escaped tables

### LINKED ISSUE #30551: Unable to parse escaped tables
### Bug description

The following query throws `This database does not allow for DDL/DML, and the query could not be parsed to confirm it is a read-only query. Please contact your administrator for more assistance.`

```
select *
from `birth_names`
```

The ability to escape table names is essential given that we might have names that conflict with reserved keywords.

This bug was introduced in https://github.com/apache/superset/pull/30417.

### Screenshots/recordings

Screenshot 2024-10-08 at 11 29 55

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

michael-s-molina:
### SUMMARY
https://github.com/apache/superset/pull/26476 introduced a `SQLGLOT_DIALECTS` mapping to be used when parsing SQL queries with SQLGlot. For companies that have custom database engines, with a name not present in the mapping, the default SQLGlot dialect would be used and could generate problems such as https://github.com/apache/superset/issues/30551. This PR adds a new configuration to allow extending registered SQLGlot dialects.

Fixes https://github.com/apache/superset/issues/30551

### TESTING INSTRUCTIONS
Check the original issue for instructions.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -48,7 +48,6 @@ from pandas import Series
 from pandas._libs.parsers import STR_NA_VALUES
 from sqlalchemy.engine.url import URL
 from sqlalchemy.orm.query import Query
-from sqlglot import Dialect, Dialects
 
 from superset.advanced_data_type.plugins.internet_address import internet_address
 from superset.advanced_data_type.plugins.internet_port import internet_port
@@ -69,6 +68,7 @@ logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from flask_appbuilder.security.sqla import models
+    from sqlglot import Dialect, Dialects
 
     from superset.connectors.sqla.models import SqlaTable
     from superset.models.core import Database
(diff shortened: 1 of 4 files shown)


superset/config.py
```diff
@@ -48,6 +48,7 @@
 from pandas._libs.parsers import STR_NA_VALUES
 from sqlalchemy.engine.url import URL
 from sqlalchemy.orm.query import Query
+from sqlglot import Dialect, Dialects
```
betodealmeida, line 51: Nit: move this to inside the `if TYPE_CHECKING:` block.

superset/sql_parse.py
```diff
@@ -88,61 +92,6 @@
 lex.set_SQL_REGEX(sqlparser_sql_regex)
 
 
-# mapping between DB engine specs and sqlglot dialects
```
betodealmeida, line 91: Thanks for removing the duplicate code here!

betodealmeida (approved):
Awesome!

PR #30588, opened 2024-10-12
fix(explore): column data type tooltip format

mistercrunch:
Minor visual tweak to add a column and a space inside the Explore -> LeftPanelDatasetView -> Column datatype tooltip.

Adding a colon and a space, and showing the data type as uppercase as it's pretty standard to expose it in that way elsewhere in Superset and as a general convention

### before
Screenshot 2024-10-11 at 5 47 05 PM

### after
Screenshot 2024-10-11 at 5 48 28 PM

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
@@ -55,8 +55,7 @@ const TooltipSection = ({
   text: ReactNode;
 }) => (
   <TooltipSectionWrapper>
-    <TooltipSectionLabel>{label}</TooltipSectionLabel>
-    <span>{text}</span>
+    <TooltipSectionLabel>{label}</TooltipSectionLabel>: <span>{text}</span>
   </TooltipSectionWrapper>
 );
 
@@ -71,12 +70,7 @@ export const getColumnTypeTooltipNode = (column: ColumnMeta): ReactNode => {
     return null;
   }
 
-  return (
-    <TooltipSection
-      label={t('Column datatype')}
-      text={column.type.toLowerCase()}
-    />
-  );
+  return <TooltipSection label={t('Column type')} text={column.type} />;
 };
 
 export const getColumnTooltipNode = (
--- a/superset-frontend/packages/superset-ui-chart-controls/test/components/labelUtils.test.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/test/components/labelUtils.test.tsx
@@ -90,7 +90,7 @@ test('should get column datatype rendered as tooltip when column has a type', ()
     </>,
   );
 
-  expect(screen.getByText('Column datatype')).toBeVisible();
+  expect(screen.getByText('Column type')).toBeVisible();
   expect(screen.getByText('text')).toBeVisible();
 });
 


superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
```diff
@@ -55,8 +55,7 @@ const TooltipSection = ({
   text: ReactNode;
 }) => (
   <TooltipSectionWrapper>
-    <TooltipSectionLabel>{label}</TooltipSectionLabel>
-    <span>{text}</span>
+    <TooltipSectionLabel>{label}</TooltipSectionLabel>: <span>{text}</span>
```
rusackas, line 58: might be a legit use case for a `&nbsp;`

rusackas: Also not sure the span has any real purpose here. ¯\\\_(ツ)_/¯

mistercrunch: I had a `{' '}` and I think prettier changed it to what is now there. Generally I prefer the `{' '}` to `&nbsp;`, and if prettier is configured to do what it did, let's roll with it. About the span, I think it's referenced in the CSS so that it's not bolded.

requirements/base.txt
```diff
@@ -394,7 +394,7 @@ werkzeug==3.0.3
     #   flask-login
 wrapt==1.16.0
     # via deprecated
-wtforms==3.1.2
+wtforms==3.2.1
```
rusackas, line 397: Should these bumps be done in a separate PR? They don't have a history of going super smoothly.

superset-frontend/package.json
```diff
@@ -244,10 +244,10 @@
     "@storybook/react-webpack5": "8.1.11",
     "@svgr/webpack": "^8.1.0",
     "@testing-library/dom": "^8.20.1",
-    "@testing-library/jest-dom": "^5.11.6",
+    "@testing-library/jest-dom": "^6.5.0",
     "@testing-library/react": "^12.1.5",
-    "@testing-library/react-hooks": "^5.1.3",
-    "@testing-library/user-event": "^12.7.0",
+    "@testing-library/react-hooks": "^8.0.1",
```
rusackas, line 249: Big jump! Changelog looking safe?

superset-frontend/package.json
```diff
@@ -340,7 +340,7 @@
     "react-resizable": "^3.0.5",
     "react-test-renderer": "^16.14.0",
     "redux-mock-store": "^1.5.4",
-    "sinon": "^18.0.1",
+    "sinon": "^18.0.0",
```
rusackas, line 343: Why the downgrade?

superset-frontend/packages/superset-ui-chart-controls/package.json
```diff
@@ -35,10 +35,10 @@
     "@emotion/react": "^11.4.1",
     "@superset-ui/core": "*",
     "@testing-library/dom": "^8.20.1",
-    "@testing-library/jest-dom": "^5.11.6",
+    "@testing-library/jest-dom": "*",
     "@testing-library/react": "^12.1.5",
-    "@testing-library/react-hooks": "^5.0.3",
-    "@testing-library/user-event": "^12.7.0",
+    "@testing-library/react-hooks": "*",
+    "@testing-library/user-event": "*",
```
rusackas, line 41: Part of me is worried about the root package file breaking not-so-well-tested plugins when we do this, but the other part of me wants to make them ALL wildcards :P

rusackas (approved):
I love how short this is now :)

PR #30593, opened 2024-10-14
feat(number-format): Add duration formatter with colon notation

gerbermichi:
### SUMMARY
Introduces a new formatter, DURATION_COL, for durations presented in colon notation (e.g., 0:10.5). 

image

### TESTING INSTRUCTIONS
1. Create a chart with a duration (in ms) column 
2. Change the number formatter to `Duration in ms (10500 => 0:10.5)`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/setup/setupFormatters.ts
+++ b/superset-frontend/src/setup/setupFormatters.ts
@@ -78,6 +78,10 @@ export default function setupFormatters(
       'DURATION_SUB',
       createDurationFormatter({ formatSubMilliseconds: true }),
     )
+    .registerValue(
+      'DURATION_COL',
+      createDurationFormatter({ colonNotation: true }),
+    )
     .registerValue('MEMORY_DECIMAL', createMemoryFormatter({ binary: false }))
     .registerValue('MEMORY_BINARY', createMemoryFormatter({ binary: true }));
 
--- a/superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
@@ -57,6 +57,7 @@ export const D3_FORMAT_OPTIONS: [string, string][] = [
   ...d3Formatted,
   ['DURATION', t('Duration in ms (66000 => 1m 6s)')],
   ['DURATION_SUB', t('Duration in ms (1.40008 => 1ms 400µs 80ns)')],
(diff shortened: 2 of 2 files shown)


superset-frontend/packages/superset-ui-core/src/number-format/factories/createTimeDurationFormatter.ts
```diff
@@ -0,0 +1,43 @@
...
+    formatFunc: value => {
+      const v = Math.abs(Math.round(value));
+      const h = Math.floor(v / 60);
+      const m = v % 60;
+      return `${h}:${String(m).padStart(2, '0')}`;
+    },
+    id: id ?? 'duration_format',
+    label: label ?? `Duration formatter`,
+  });
+}
```
villebro, line 43: We already have a dependency on [pretty-ms](https://www.npmjs.com/package/pretty-ms) that achieves this and (see `colonNotation` option).

superset-frontend/src/setup/setupFormatters.ts
```diff
@@ -78,6 +79,7 @@ export default function setupFormatters(
       'DURATION_SUB',
       createDurationFormatter({ formatSubMilliseconds: true }),
     )
+    .registerValue('TIME_DURATION', createTimeDurationFormatter())
```
villebro, line 82: Please check the `README.md` in the `number-format` directory for an example how to implement this. I believe this should do it:
```suggestion
    .registerValue('TIME_DURATION', createTimeDurationFormatter({ colonNotation: true }))
```

villebro (changes requested):
We don't need to create a new formatter for this, as the existing duration formatter already supports this functionality. Also, let's stick to milliseconds for now to keep this aligned with the existing duration formatters.

villebro:
FYI: I opened a PR to bump `pretty-ms` to pull in a critical fix for `colonNotation`, as that's what we should be using here: #30599

superset-frontend/src/setup/setupFormatters.ts
gerbermichi: Thank you. This does not the same, it will format it in hh:mm:ss and I would like to create a formatter for hh:mm. However, I can add both, DURATION_COL and DURATION_COL_SHORT. What are your thoughts?

villebro (approved):

PR #30613, opened 2024-10-15
chore: Enable suppressing default chart context menu

kgabryje:
### SUMMARY
Add a config option to chart metadata to allow suppressing the default chart context menu on the dashboard (the one with the default cross filter and drill to detail menu items)

### TESTING INSTRUCTIONS
1. Tests should pass
2. The dashboard context menu should work like before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/chart/models/ChartMetadata.ts
+++ b/superset-frontend/packages/superset-ui-core/src/chart/models/ChartMetadata.ts
@@ -50,6 +50,8 @@ export interface ChartMetadataConfig {
   labelExplanation?: string | null;
   queryObjectCount?: number;
   parseMethod?: ParseMethod;
+  // suppressContextMenu: true hides the default context menu for the chart.
+  // This is useful for viz plugins that define their own context menu.
   suppressContextMenu?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


michael-s-molina:
@kgabryje Can you explain a little bit more about the intention of this PR? Currently, disabled options are greyed out by default but you're trying to completely suppress the menu, which might lead to users asking why right click is not working. Could you explain what's the use case you're trying to solve?

kgabryje:
> @kgabryje Can you explain a little bit more about the intention of this PR? Currently, disabled options are greyed out by default but you're trying to completely suppress the menu, which might lead to users asking why right click is not working. Could you explain what's the use case you're trying to solve?

The intention is so that a viz plugin can define it's own, fully custom context menu. For example, the viz plugin could have a context menu with more actions, specific to this plugin.
This option is not intended to be used with any of our currently existing plugins

michael-s-molina:
> This option is not intended to be used with any of our currently existing plugins

I recommend adding a comment to the property explaining the use as it can be easily removed. This also raises the question if this shouldn't be an extension to be more explicit?

kgabryje:
@michael-s-molina Added a comment. I believe this change fits in Superset - for instance, it would be useful if in the future we wanted to improve the table viz plugin by adding more table-related actions to the context menu. Currently it's not possible as we are limited by the context menu defined in ChartRenderer which is the same for every plugin.

michael-s-molina (approved):

PR #30646, opened 2024-10-17
fix(Dashboard): Retain colors when color scheme not set

### LINKED ISSUE #23874: Cannot set color scheme for dashboard-associated chart even when dashboard color scheme is cleared
Once a chart is associated with a dashboard, it is no longer possible to change its color scheme. In the customize panel, the color scheme dropdown displays the warning:

> The color scheme is determined by the related dashboard. Edit the color scheme in the dashboard properties.

And disabled the dropdown via this code:
https://github.com/apache/superset/blob/594d3e0e0c4add06c1216490c63cdc6ac43e2f5a/superset-frontend/src/explore/components/controls/ColorSchemeControl/index.tsx#L111-L121

However, if the dashboard does not have a color scheme selected (or it's [been cleared](https://github.com/apache/superset/blob/594d3e0e0c4add06c1216490c63cdc6ac43e2f5a/superset-frontend/src/dashboard/components/ColorSchemeControlWrapper.jsx#L64)), the chart's color scheme is still respected (you can see this by selecting a color scheme for a chart before assigning it to a dashboard without a color scheme). In this situation, it does make sense to change the chart's color scheme, and so the dropdown shouldn't be disabled (the warning could still be shown though).

#### How to reproduce the bug

1. Create a chart not associated with any dashboard
2. Set its color scheme to something that's not the default.
3. Create a dashboard. Do not set its color scheme (or set it to `""` in the JSON config).
4. Add the chart to the dashboard. Notice that it still renders with the chosen color scheme.
5. Go to the chart's "Customize" tab. Notice that you cannot change its color scheme (nor is its current color scheme shown).

### Expected results

Since the dashboard has no color scheme set, and the chart's color scheme is thus used, I expect to be able to change it.

### Actual results

I am unable to change the chart's color scheme except by removing it from the dashboard, changing the color scheme, and adding it back.

### Environment

- browser type and version: Firefox 112.0.1.
- superset version: Hosted custom build, so I'm not 100% sure. I think it's close to master.
- python version: Hosted, so unknown.
- node.js version: Hosted, so unknown.
- any feature flags active: Hosted, so unknown.

### Checklist

### Additional context

I can't for the life of me find the exact point in the code where dashboard color scheme is chosen over the chart color scheme when the former is set, but I'm pretty sure that code is going to show that if the dashboard color scheme is unset (or `""`), then the chart color scheme will be used.

geido:
### SUMMARY
This PR does the following:
- It fixes an issue for which colors would sometimes not keep their consistency when leaving and revisiting the Dashboard.
- It fixes an issue for which charts original color schemes were not tracked making falling back to their origin color scheme impossible when removing the color scheme from the Dashboard.
- It attempts to fix and re-enable skipped E2E tests that are critical for the coverage of the color consistency implementation.
- Fixes https://github.com/apache/superset/issues/23874 with limitations:
  - The color scheme cannot be changed if the chart shares labels with multiple charts in the Dashboard. 
  - When setting custom label colors via the advanced settings of the Dashboard, the custom label colors will always be enforced also when a color scheme is not set for the Dashboard. 
  - The chart should consistently display colors that match those used in the Dashboard. 

### BEFORE

https://github.com/user-attachments/assets/d66e66ff-13ec-4bb1-a1bf-72f6afcd1bfc

### AFTER

https://github.com/user-attachments/assets/fd74e76e-ea20-4d89-8f05-49455bcb5fd0

### TESTING INSTRUCTIONS
1. Open a Dashboard with a color scheme and shared label colors
2. Edit the Dashboard properties and remove the color scheme
3. After applying the changes, verify that shared label colors are still respected
4. For charts that do not share labels, verify their original color scheme is respected
5. Save and exit the Dashboard by going to the Dashboard lists
6. Come back again to the same Dashboard you just saved
7. Make sure colors haven't changed from the previous state

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


hanslemm:
Upvote for this. We also have this problem.

rusackas:
OMG thank you!

superset-frontend/cypress-base/cypress/e2e/dashboard/editmode.test.ts
```diff
@@ -42,15 +42,29 @@ function openProperties() {
     cy.getBySel('header-actions-menu')
       .contains('Edit properties')
       .click({ force: true });
-    cy.wait(500);
+    cy.get('.ant-modal-body').should('be.visible');
```
rusackas, line 45: could add a `data-test` attribute to select if that's considered more stable than the antD class name.

rusackas (approved):
Thanks for fixing, LGTM!

geido:
I will try to tackle an issue for which after deleting a color scheme from a Dashboard and then editing the Chart from the Dashboard, Explore won't allow to customize the color.

fisjac (approved):
LGTM

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/sharedControls.tsx
```diff
@@ -333,9 +341,17 @@ const color_scheme: SharedControlConfig<'ColorSchemeControl'> = {
   choices: () => categoricalSchemeRegistry.keys().map(s => [s, s]),
   description: t('The color scheme for rendering chart'),
   schemes: () => categoricalSchemeRegistry.getMap(),
-  mapStateToProps: state => ({
-    dashboardId: state?.form_data?.dashboardId,
-  }),
+  mapStateToProps: state => {
```
kgabryje, line 344: We've got a lot of repeated code in color_scheme and linear_color_scheme. Could we DRY it a bit?

superset-frontend/packages/superset-ui-core/src/color/CategoricalColorNamespace.ts
```diff
@@ -37,10 +37,23 @@ export default class CategoricalColorNamespace {
     this.forcedItems = {};
   }
 
-  getScale(schemeId?: string) {
-    const id = schemeId ?? getCategoricalSchemeRegistry().getDefaultKey() ?? '';
+  /**
+   * A new CategoricalColorScale instance is created for each chart.
```
kgabryje, line 41: thanks for all those new docstrings!

kgabryje, line 41: thanks for all those new docstrings!

superset-frontend/packages/superset-ui-core/src/color/LabelsColorMapSingleton.ts
```diff
@@ -38,17 +42,38 @@ export class LabelsColorMap {
...
-      const currentColorScheme = colorScheme || originalChartColorScheme;
-      const colorScale = categoricalNamespace.getScale(currentColorScheme);
+      const { labels, ownScheme } = chartConfig;
+      const appliedColorScheme = colorScheme || ownScheme;
+      const colorScale = categoricalNamespace.getScale(appliedColorScheme);
 
       labels.forEach(label => {
-        const newColor = colorScale.getColor(label, sliceId);
-        newColorMap.set(label, newColor);
+        // if merge, apply the scheme only to new labels in the map
```
kgabryje, line 68: Why do the existing colors have priority over the new colors when merging?

superset-frontend/src/dashboard/components/DashboardBuilder/DashboardContainer.tsx
```diff
@@ -85,9 +86,17 @@ const DashboardContainer: FC<DashboardContainerProps> = ({ topLevelTabs }) => {
   const directPathToChild = useSelector<RootState, string[]>(
     state => state.dashboardState.directPathToChild,
   );
-  const chartIds = useSelector<RootState, number[]>(state =>
-    Object.values(state.charts).map(chart => chart.id),
+  const isEditMode = useSelector<RootState, boolean>(
+    state => state.dashboardState.editMode,
   );
+  const charts = useSelector<RootState, Chart[]>(state =>
```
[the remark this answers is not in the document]
kgabryje: It could probably be further optimised, so that we don't repeat some operations, but the general idea is to return values/objects from useSelector that change scarcely

superset-frontend/src/utils/colorScheme.ts
```diff
@@ -107,31 +140,67 @@ export const refreshLabelsColorMap = (
...
 
   // get the fresh map that was just updated or existing
   const labelsColorMapEntries = fresh
-    ? getLabelsColorMapEntries()
-    : labelsColorMap;
+    ? currentColorMapEntries
+    : merge
+      ? {
+          ...currentColorMapEntries,
+          ...labelsColorMap,
```
kgabryje, line 201: referring to my previous comment about `merge` - here it seems like the new colors override the existing ones, rather than the other way around. Is that intended?

superset-frontend/src/explore/components/controls/ColorSchemeControl/index.tsx
```diff
@@ -110,9 +138,20 @@ const ColorSchemeControl = ({
...
+    dashboardId &&
+    countSharedLabelsColor > 0 &&
+    chartLabels.some(label => sharedLabelsColors.includes(label))
+  );
+  const hasDashboardScheme = dashboardId && hasDashboardColorScheme;
   const theme = useTheme();
   const currentScheme = useMemo(() => {
-    if (dashboardId) {
+    if (hasDashboardScheme || hasSharedLabelsColor) {
       return 'dashboard';
```
kgabryje, line 155: nit: should this be a const? it's also used later in line 168

superset-frontend/src/explore/components/controls/ColorSchemeControl/index.tsx
```diff
@@ -121,13 +160,15 @@ const ColorSchemeControl = ({
...
     }
     return result;
-  }, [dashboardId, defaultScheme, schemes, value]);
+  }, [defaultScheme, hasDashboardScheme, hasSharedLabelsColor, schemes, value]);
 
   const options = useMemo(() => {
-    if (dashboardId) {
+    if (hasDashboardScheme || hasSharedLabelsColor) {
       return [
         <Option value="dashboard" label={t('dashboard')} key="dashboard">
```
kgabryje, line 168:
```suggestion
        <Option value="dashboard" label={t('Dashboard')} key="dashboard">
```

superset-frontend/src/utils/colorScheme.ts
geido: Yes this is intended. The idea is that if new labels appear they should get new colors. but existing labels should keep their original colors.

superset-frontend/src/dashboard/components/DashboardBuilder/DashboardContainer.tsx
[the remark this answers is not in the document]
kgabryje: I meant that we should get rid if this selector altogether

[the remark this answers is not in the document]
kgabryje: gotcha, lets merge and I'll do it when I work on perf

kgabryje (approved):

PR #30674, opened 2024-10-22
feat(oauth): adding necessary changes to support bigquery oauth

fisjac:
### SUMMARY

Adding needed changes to support future implementation of OAuth2 functionality for BigQuery.

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/OAuth2ClientField.tsx
+++ b/superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/OAuth2ClientField.tsx
@@ -22,7 +22,7 @@ import { useState } from 'react';
 import Collapse from 'src/components/Collapse';
 import { Input } from 'src/components/Input';
 import { FormItem } from 'src/components/Form';
-import { FieldPropTypes } from '../../types';
+import { FieldPropTypes, Engines } from '../../types';
 
 interface OAuth2ClientInfo {
   id: string;
@@ -30,6 +30,7 @@ interface OAuth2ClientInfo {
   authorization_request_uri: string;
   token_request_uri: string;
   scope: string;
+  project_id?: string;
 }
 
 export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
@@ -42,6 +43,7 @@ export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
     token_request_uri:
       encryptedExtra.oauth2_client_info?.token_request_uri || '',
     scope: encryptedExtra.oauth2_client_info?.scope || '',
+    project_id: encryptedExtra.oauth2_client_info?.project_id || '',
   });
 
   if (db?.engine_information?.supports_oauth2 !== true) {
@@ -106,6 +108,15 @@ export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
             onChange={handleChange('scope')}
           />
         </FormItem>
+        {db.engine === Engines.BigQuery && (
+          <FormItem label="Project ID">
+            <Input
+              data-test="project_id"
+              value={oauth2ClientInfo.project_id}
+              onChange={handleChange('project_id')}
+            />
+          </FormItem>
+        )}
       </Collapse.Panel>
     </Collapse>
   );
--- a/superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/OAuth2ClientField.tsx
+++ b/superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/OAuth2ClientField.tsx
@@ -24,6 +24,15 @@ import { Input } from 'src/components/Input';
 import { FormItem } from 'src/components/Form';
 import { FieldPropTypes, Engines } from '../../types';
 
+const LABELS = {
+  CLIENT_ID: 'Client ID',
+  SECRET: 'Client Secret',
+  AUTH_URI: 'Authorization Request URI',
+  TOKEN_URI: 'Token Request URI',
+  SCOPE: 'Scope',
+  PROJECT: 'Project ID',
+};
+
 interface OAuth2ClientInfo {
   id: string;
   secret: string;
@@ -70,14 +79,14 @@ export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
   return (
     <Collapse>
       <Collapse.Panel header="OAuth2 client information" key="1">
-        <FormItem label="Client ID">
+        <FormItem label={LABELS.CLIENT_ID}>
           <Input
             data-test="client-id"
             value={oauth2ClientInfo.id}
             onChange={handleChange('id')}
           />
         </FormItem>
-        <FormItem label="Client Secret">
+        <FormItem label={LABELS.SECRET}>
           <Input
             data-test="client-secret"
             type="password"
@@ -85,7 +94,7 @@ export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
             onChange={handleChange('secret')}
           />
         </FormItem>
-        <FormItem label="Authorization Request URI">
+        <FormItem label={LABELS.AUTH_URI}>
           <Input
             data-test="client-authorization-request-uri"
             placeholder="https://"
@@ -93,7 +102,7 @@ export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
             onChange={handleChange('authorization_request_uri')}
           />
         </FormItem>
-        <FormItem label="Token Request URI">
+        <FormItem label={LABELS.TOKEN_URI}>
           <Input
             data-test="client-token-request-uri"
             placeholder="https://"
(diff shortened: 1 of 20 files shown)


superset/commands/database/test_connection.py
```diff
@@ -225,6 +225,10 @@ def ping(engine: Engine) -> bool:
             # bubble up the exception to return proper status code
             raise
         except Exception as ex:
+            if database.is_oauth2_enabled() and database.db_engine_spec.needs_oauth2(
```
fisjac, line 228: BigQuery raises an error when running `create_engine()` necessitating another try/except block to look for the `needs_oauth2(ex)` exception. Most DB's allow for an engine to be created without valid creds, and instead raises the exception on `engine.connect()`

superset/db_engine_specs/base.py
```diff
@@ -1691,10 +1691,13 @@ def select_star(  # pylint: disable=too-many-arguments
         return sql
 
     @classmethod
-    def estimate_statement_cost(cls, statement: str, cursor: Any) -> dict[str, Any]:
+    def estimate_statement_cost(
```
fisjac, line 1694: Database is required to create the BigQuery Client when using OAuth2

superset/db_engine_specs/postgres.py
```diff
@@ -351,7 +351,16 @@ def get_allow_cost_estimate(cls, extra: dict[str, Any]) -> bool:
         return True
 
     @classmethod
-    def estimate_statement_cost(cls, statement: str, cursor: Any) -> dict[str, Any]:
+    def estimate_statement_cost(
```
fisjac, line 354: aligning other specs to have database, and adding missing docstring

superset/db_engine_specs/presto.py
```diff
@@ -365,9 +365,12 @@ def get_schema_from_engine_params(
         return parse.unquote(database.split("/")[1])
 
     @classmethod
-    def estimate_statement_cost(cls, statement: str, cursor: Any) -> dict[str, Any]:
+    def estimate_statement_cost(
```
fisjac, line 368: adding database param

superset/models/core.py
```diff
@@ -510,12 +511,22 @@ def _get_sqla_engine(  # pylint: disable=too-many-locals
         logger.debug("Database._get_sqla_engine(). Masked URL: %s", str(masked_url))
 
         if self.impersonate_user:
-            self.db_engine_spec.update_impersonation_config(
-                connect_args,
-                str(sqlalchemy_url),
-                effective_username,
-                access_token,
-            )
+            # Checking if the function signature can accept database as a param
```
fisjac, line 514: Again, here BigQuery needs the Database object to build the client. Checks if the function signature has `database` as a param, then passes it.

superset/utils/oauth2.py
```diff
@@ -192,3 +192,4 @@ class OAuth2ClientConfigSchema(Schema):
     )
     authorization_request_uri = fields.String(required=True)
     token_request_uri = fields.String(required=True)
+    project_id = fields.String(required=False)
```
fisjac, line 195: BigQuery takes `project_id` as a param in its `OAuth2` parameters

superset/models/core.py
```diff
@@ -510,12 +511,23 @@ def _get_sqla_engine(  # pylint: disable=too-many-locals
...
```
betodealmeida, line 514: Nice! A suggestion on the comment here on line 514, it's always better to have comments explain the *why* rather than the *what* (since the latter can be inferred from the code). I would rewrite this as something like "PR #30674 changed the signature of the method to include database. This ensures that the change is backwards compatible."

superset/models/core.py
```diff
@@ -510,12 +511,23 @@ def _get_sqla_engine(  # pylint: disable=too-many-locals
...
+                    effective_username,
+                    access_token,
+                )
+            else:
+                self.db_engine_spec.update_impersonation_config(  # type: ignore[call-arg]
+                    connect_args,
+                    str(sqlalchemy_url),  # type: ignore[arg-type]
+                    effective_username,  # type: ignore[arg-type]
+                    access_token,
+                )
```
betodealmeida, line 530: You could simplify this a bit to:

```python
args = [connect_args, str(sqlalchemy_url), effective_username, access_token]
if "database" in sig_parameters:
    args.insert(0, self)
self.db_engine_spec.update_impersonation_config(*args)
```

betodealmeida (commented):
Looks good, but wondering if we need to modify `_get_client` here to take a database object.

superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/OAuth2ClientField.tsx
```diff
@@ -106,6 +108,15 @@ export const OAuth2ClientField = ({ changeMethods, db }: FieldPropTypes) => {
             onChange={handleChange('scope')}
           />
         </FormItem>
+        {db.engine === Engines.BigQuery && (
+          <FormItem label="Project ID">
```
geido, line 112: Let's localize the label

superset/models/core.py
```diff
@@ -510,12 +511,13 @@ def _get_sqla_engine(  # pylint: disable=too-many-locals
...
         if self.impersonate_user:
-            self.db_engine_spec.update_impersonation_config(
-                connect_args,
-                str(sqlalchemy_url),
-                effective_username,
-                access_token,
-            )
+            # PR #30674 changed the signature of the method to include database.
+            # This ensures that the change is backwards compatible
+            sig = signature(self.db_engine_spec.update_impersonation_config)
```
geido, line 516: Do we have sufficient test coverage for this case?

superset/utils/oauth2.py
betodealmeida: I think ideally we'd want to store the project id in the URI, to make it compatible with non-oauth2 use cases. And then later when you need you can grab it from `database.sqlalchemy_uri.database`.

betodealmeida (changes requested):
Looks good, but I left a few comments on the handling of `project_id`. Happy to hop on a meeting to discuss it in more detail.

superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/OAuth2ClientField.tsx
```diff
@@ -22,14 +22,15 @@ import { useState } from 'react';
...
-import { FieldPropTypes } from '../../types';
+import { FieldPropTypes, Engines } from '../../types';
 
 interface OAuth2ClientInfo {
   id: string;
   secret: string;
   authorization_request_uri: string;
   token_request_uri: string;
   scope: string;
+  project_id?: string;
```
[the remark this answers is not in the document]
fisjac: The main reason I wanted to place this here is that, due to the flow for BigQuery, users will either add service creds, which contains the `project_id`, or they will add OAuth creds, which also contains the `project_id`. I agree in theory we should have a separate field, but that possibly gets even messier when you can have a `project_id` that can differ from the one used in the service creds.

superset/databases/api.py
```diff
@@ -1341,6 +1342,7 @@ def validate_sql(self, pk: int) -> FlaskResponse:
             return self.response_404()
 
     @expose("/oauth2/", methods=["GET"])
+    @transaction()
```
betodealmeida, line 1345: Nice!

Does this work even though the method is a `GET`? I remember you mentioning that it was not the case?

superset/databases/schemas.py
```diff
@@ -47,6 +47,11 @@
     SSHTunnelMissingCredentials,
 )
 from superset.constants import PASSWORD_MASK
+from superset.databases.types import (  # pylint:disable=unused-import
+    EncryptedDict,  # noqa: F401
+    EncryptedField,
+    EncryptedString,  # noqa: F401
+)
```
betodealmeida, line 54: If `EncryptedDict` and `EncryptedString` are not used, why are you importing them?

```suggestion
from superset.databases.types import EncryptedField
```

superset/models/core.py
```diff
@@ -543,6 +546,24 @@ def _get_sqla_engine(  # pylint: disable=too-many-locals
...
+        Examines a function signature looking for a database param.
+        If the signature requires a database, the function appends self in the
+        proper position.
+        """
+
+        # PR #30674 changed the signature of the method to include database.
+        # This ensures that the change is backwards compatible
+        sig = signature(func)
+        if "database" in (params := sig.parameters.keys()):
+            args.insert(list(params).index("database"), self)
```
betodealmeida, line 564: Nice! This is much more robust than my `insert(0, ...)` suggestion.

betodealmeida (changes requested):
This looks great, but I have to push back on adding project name/ID to the OAuth2 client information. Can we try adding a field for it?

superset/databases/schemas.py
fisjac: They are being imported from db_engine_specs from that location. Rather than change all of the references in every spec, I thought it would be cleaner to still point to the Schemas, which is where they really should be located were it not for app_context issues.

betodealmeida (approved):
Looks great!

PR #30688, opened 2024-10-23
fix: bump FAB to 4.5.2

dpgaspar:
### SUMMARY

WTForms 3.2.X breaks superset with the following error:

```
AttributeError: 'tuple' object has no attribute 'items'
```

More details on: https://github.com/dpgaspar/Flask-AppBuilder/pull/2279

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
     "cryptography>=42.0.4, <44.0.0",
     "deprecation>=2.1.0, <2.2.0",
     "flask>=2.2.5, <3.0.0",
-    "flask-appbuilder>=4.5.0, <5.0.0",
+    "flask-appbuilder==4.5.2rc1",
     "flask-caching>=2.1.0, <3",
     "flask-compress>=1.13, <2.0",
     "flask-talisman>=1.0.0, <2.0",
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Thanks!

michael-s-molina:
@dpgaspar Will we need to cherry-pick this into 4.1 which uses `wtforms==3.1.2`? I'm understanding that this fix is for 3.2.X which would make it not required for 4.1. Is that correct? @sadpandajoe

villebro (approved):
LGTM. However, should we add a lower bound here, too? https://github.com/apache/superset/blob/16981d63169507b39615ca9037f3b36baafba3c0/pyproject.toml#L48

dpgaspar:
> @dpgaspar Will we need to cherry-pick this into 4.1 which uses `wtforms==3.1.2`? I'm understanding that this fix is for 3.2.X which would make it not required for 4.1. Is that correct? @sadpandajoe

yes, I think thats safe to say, if our users install superset without using 4.1 specific dep requirements they will automatically upgrade WTForms and FAB

dpgaspar:
> LGTM. However, should we add a lower bound here, too?
> 
> https://github.com/apache/superset/blob/16981d63169507b39615ca9037f3b36baafba3c0/pyproject.toml#L48

No strong opinions about it

PR #30710, opened 2024-10-25
feat(helm-chart): Add extraLabels to all resources

maxforasteiro:
### SUMMARY
Some orgs require all Kubernetes resources to be labeled. Adding a non-breaking change that will allow our team to use Superset helm chart without having to copy/paste and modify it ourselves.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


helm/superset/values.yaml
```diff
@@ -27,6 +27,11 @@ nameOverride: ~
 # -- Provide a name to override the full names of resources
 fullnameOverride: ~
 
+# -- Global values for the Superset chart
+global:
+  # -- Labels to be added to all resources
+  extraLabels: {}
+
```
villebro, line 34: Based on the comment here, these would be added to `all` resources. However, it seems we're not adding these to the pod templates (see for instance `.Values.supersetWorker.podLabels` - is this intentional?

villebro (commented):
I like this idea. However, while I kind of like the `global` prefix, I'm wondering if it may be unnecessary. I believe we may already have some other global-like values that would seem non-global after this change. But I'm not super opinionated here.

Anyway, the main comment here applies to the labels not applying to the templates. Let's either apply them to those, too, or add a comment that they're not covered by this new property.

helm/superset/values.yaml
maxforasteiro: sorry, I just missed them. I'm adding it now

maxforasteiro:
> I like this idea. However, while I kind of like the `global` prefix, I'm wondering if it may be unnecessary. I believe we may already have some other global-like values that would seem non-global after this change. But I'm not super opinionated here.
> 
> Anyway, the main comment here applies to the labels not applying to the templates. Let's either apply them to those, too, or add a comment that they're not covered by this new property.

I agree with your comment, the global label can be misleading indeed.

villebro (approved):
LGTM, thanks for the improvement 🙇

villebro:
Oh, please bump the version!

maxforasteiro:
> Oh, please bump the version!

nice catch! completely forgot 🙈

villebro:
Seems there's a linting issue:
image
I didn't look closely at it, let me know if you need help resolving it.

maxforasteiro:
@villebro fixed. Sorry for the delay

villebro (approved):

PR #30715, opened 2024-10-25
fix(Jinja): Extra cache keys for Jinja columns

geido:
### SUMMARY
This PR fixes an issue where Jinja in columns was not considered during cache key checks, leading to cache spillover.

### TESTING INSTRUCTIONS

1. Create an aggregated table chart using the `Vehicle Sales` dataset from the `examples` connection.
2. Create below dimensions:
    a. `'{{current_user_id()}}'` as `id`
    a. `'{{current_username()}}'` as `username`
    a. `'{{current_user_email()}}'` as `email`
    a. `'{{url_param('test', 'default')}}'` as `url_param`
3. Set the row limit to 1.
4. Save the chart to a dashboard and access it.
5. Notice that the user-related macros will show your account data, and the `url_param` column will show `default` (which is set in the macro to be returned when the param is not defined).
6. Copy the dashboard URL, and append `&test=blah` to it (after the `native_filters_key` value) and access it.
7. The `url_param` column should now how the param value (`blah`).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -1980,6 +1980,12 @@ class SqlaTable(
             templatable_statements.append(extras["where"])
         if "having" in extras:
             templatable_statements.append(extras["having"])
+        if "columns" in query_obj:
+            templatable_statements += [
+                c["sqlExpression"]
+                for c in query_obj["columns"]
+                if c.get("sqlExpression")
+            ]
         if self.is_rls_supported:
             templatable_statements += [
                 f.clause for f in security_manager.get_rls_filters(self)
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -1984,7 +1984,7 @@ class SqlaTable(
             templatable_statements += [
                 c["sqlExpression"]
                 for c in query_obj["columns"]
-                if c.get("sqlExpression")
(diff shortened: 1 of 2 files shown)


superset/connectors/sqla/models.py
```diff
@@ -1980,6 +1980,12 @@ def has_extra_cache_key_calls(self, query_obj: QueryObjectDict) -> bool:
             templatable_statements.append(extras["where"])
         if "having" in extras:
             templatable_statements.append(extras["having"])
+        if "columns" in query_obj:
+            templatable_statements += [
+                c["sqlExpression"]
+                for c in query_obj["columns"]
+                if isinstance(c, dict) and c.get("sqlExpression")
```
villebro, line 1987: I believe we have a `TypeGuard` for this - if not let's add one.

villebro (commented):
LGTM but I suggest using a type guard for the check

villebro (approved):
LGTM, thanks for this critical fix!

kgabryje (approved):

michael-s-molina:
> LGTM, thanks for this critical fix!

@sadpandajoe I'm adding this PR to the release blockers based on this comment.

Vitor-Avila:
Thank you so much for working on this!!!

PR #30721, opened 2024-10-25
feat(dataset API): Add parameter to optionally render Jinja macros in API response

Vitor-Avila:
### SUMMARY
This PR adds support for optionally rendering Jinja macros in the API response when getting a dataset. The motivation for this feature is that external tools might be fetching dataset metadata via the API to define lineage/etc, and in case Jinja is used it's not possible to fully parse the SQL query to identify dependencies/etc. An example query:

``` sql
select * from {{dataset(1)}}
join cleaned_sales_data s on dataset_1.product_code = s.product_code
```

The query parameter `include_rendered_sql` (bool) can be used to add keys to the response:
* `rendered_sql` for the rendered virtual dataset SQL query.
* `rendered_expression` for the rendered expression from `metrics` and `columns`.  

**Before**
image

**After**
image

### TESTING INSTRUCTIONS
Testing added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -1097,8 +1097,10 @@ class DatasetRestApi(BaseSupersetModelRestApi):
                 schema:
                   $ref: '#/components/schemas/get_item_schema'
           - in: query
-            name: render
-            description: Should Jinja macros from sql, metrics and columns be rendered
+            name: include_rendered_sql
+            description: >-
+              Should Jinja macros from sql, metrics and columns be rendered
+              and included in the response
             schema:
               type: boolean
           responses:
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -18,11 +18,16 @@
 import logging
 from datetime import datetime
 from io import BytesIO
-from typing import Any
+from typing import Any, Optional
 from zipfile import is_zipfile, ZipFile
 
 from flask import request, Response, send_file
 from flask_appbuilder.api import expose, protect, rison, safe
+from flask_appbuilder.api.schemas import get_item_schema
+from flask_appbuilder.const import (
+    API_RESULT_RES_KEY,
+    API_SELECT_COLUMNS_RIS_KEY,
+)
 from flask_appbuilder.models.sqla.interface import SQLAInterface
 from flask_babel import ngettext
 from marshmallow import ValidationError
@@ -65,6 +70,7 @@ from superset.datasets.schemas import (
     GetOrCreateDatasetSchema,
     openapi_spec_methods_override,
 )
+from superset.jinja_context import BaseTemplateProcessor, get_template_processor
 from superset.utils import json
 from superset.utils.core import parse_boolean_string
 from superset.views.base import DatasourceFilter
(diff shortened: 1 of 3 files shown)


superset/datasets/api.py
```diff
@@ -1056,3 +1062,123 @@ def warm_up_cache(self) -> Response:
...
+            args,
+            **{API_SELECT_COLUMNS_RIS_KEY: pruned_select_cols},
+        )
+        if pruned_select_cols:
+            show_model_schema = self.model2schemaconverter.convert(pruned_select_cols)
+        else:
+            show_model_schema = self.show_model_schema
+
+        response["id"] = pk
+        response[API_RESULT_RES_KEY] = show_model_schema.dump(item, many=False)
```
Vitor-Avila, line 1145: This is FAB's default implementation for `get_headless` - [ref](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/api/__init__.py#L1453-L1485).

I decided to not simply do `self.get_headless()` and then modify the response for two reasons:
* `get_headless` returns a response object, so while we could just modify the `response.json` it feels a little hacky.
* Superset's `get_headless` implementation includes a call to `send_stast_metrics` ([ref](https://github.com/apache/superset/blob/master/superset/views/base_api.py#L450-L451)), which wouldn't include the full duration of the API call as the response would be further modified by this method before concluding.

I also thought about just overwriting FAB's `pre_get` method which is available in case the response needs to be modified as part of `get_headless`, but we need the `Database` associated with this dataset, and if the user chooses to not include the `database.id` in the response then `pre_get` wouldn't have this info.

superset/datasets/api.py
```diff
@@ -18,11 +18,16 @@
 import logging
 from datetime import datetime
 from io import BytesIO
-from typing import Any
+from typing import Any, Optional
```
betodealmeida, line 21: Hint: you can do `from __future__ import annotations` and then use `a | None` instead of `Optional[a]`.

betodealmeida (approved):
Nice!

villebro (approved):
LGTM, however, it's worth remembering that the rendered query can change, considerably even, if it has lots of conditional Jinja logic. It's pretty common to see additional where clauses added if certain filter parameters are set in the chart data request. Just something to keep in mind.

Vitor-Avila:
> LGTM, however, it's worth remembering that the rendered query can change, considerably even, if it has lots of conditional Jinja logic. It's pretty common to see additional where clauses added if certain filter parameters are set in the chart data request. Just something to keep in mind.

@villebro that's true. I actually decided to include both the regular `sql` and `expression` values and then add `rendered_sql` and `rendered_expression` if the param is present, so that users still get the original value. I think having both data is helpful for such cases.

superset/datasets/api.py
```diff
@@ -1179,13 +1185,22 @@ def render_item_list(item_list: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
 
-        if sql := data.get("sql"):
-            data["rendered_sql"] = processor.process_template(sql)
+            if sql := data.get("sql"):
+                item_type = "query"
+                data["rendered_sql"] = processor.process_template(sql)
+        except TemplateSyntaxError as ex:
+            raise SupersetTemplateException(
+                f"Unable to render expression from dataset {item_type}.",
+            ) from ex
```
betodealmeida, line 1204: Could possibly be simplified to:

```python
items = [
    ("query", "sql", "rendered_sql", processor.process_template),
    ("metric", "metrics", "metrics", render_item_list),
    ("column", "columns", "columns", render_item_list),
]
for item_type, key, new_key, func in items:
    if key not in data:  # or if not data.get(key), potentially
        continue

    try:
        data[new_key] = func(data[key])
    except TemplateSyntaxError as ex:
        raise SupersetTemplateException(
            f"Unable to render expression from dataset {item_type}.",
        ) from ex
```

But it's perfectly fine as is.

betodealmeida (approved):

superset/datasets/api.py
Vitor-Avila: That's cool! Let me implement this 🙏

PR #30724, opened 2024-10-28
fix(embedded): Hide "Edit chart" menu option as well as "Edit chart" button in "View as table" modal for embedded context

cezudas:
### SUMMARY

When a dashboard is loaded in embedded mode, no 'Edit chart' functionality makes sense regardless of whether the guest user has `supersetCanExplore` permission.
Why? Because the `/explore` page isn't available in the DashboardEmbed mode so if the role provided for `GUEST_ROLE_NAME` has `supersetCanExplore` permission the 'Edit chart' functionality would not work (guest user will see a blank page in the embed mode)

### TESTING INSTRUCTIONS

Screenshot 2024-10-28 at 09 06 14

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/hooks/useIsEmbedded.ts
+++ b/superset-frontend/src/hooks/useIsEmbedded.ts
@@ -1,0 +1,5 @@
+import { useSelector } from 'react-redux';
+import { RootState } from 'src/dashboard/types';
+
+export const useIsEmbedded = () =>
+  useSelector<RootState, boolean>(({ dashboardInfo }) => !dashboardInfo.userId);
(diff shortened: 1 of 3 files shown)


superset-frontend/src/hooks/useIsEmbedded.ts
```diff
@@ -0,0 +1,5 @@
+import { useSelector } from 'react-redux';
+import { RootState } from 'src/dashboard/types';
+
+export const useIsEmbedded = () =>
+  useSelector<RootState, boolean>(({ dashboardInfo }) => !dashboardInfo.userId);
```
geido, line 5: I am not sure that this is a reliable way to identify an embedded context

cezudas: Hi @geido @sadpandajoe PHAL :)

PR #30735, opened 2024-10-28
fix(Jinja): Extra cache keys for calculated columns and metrics using Jinja

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/30715. It adds logic to check Jinja in calculated columns and also metrics (both ad-hoc and from the dataset level).

No UI changes.

### TESTING INSTRUCTIONS
Added testing to this flow.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/sqla_models_tests.py
+++ b/tests/integration_tests/sqla_models_tests.py
@@ -15,11 +15,13 @@
 # specific language governing permissions and limitations
 # under the License.
 # isort:skip_file
+from __future__ import annotations
+
 import re
 from datetime import datetime
-from typing import Any, NamedTuple, Optional, Union
+from typing import Any, Literal, NamedTuple, Optional, Union
 from re import Pattern
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 import pytest
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tests/integration_tests/sqla_models_tests.py
```diff
@@ -913,54 +913,102 @@ def test_extra_cache_keys_in_sql_expression(
...
     expected_cache_keys,
     has_extra_cache_keys,
+    item_type,
 ):
     table = SqlaTable(
         table_name="test_has_no_extra_cache_keys_table",
         sql="SELECT 'abc' as user",
         database=get_example_database(),
     )
+    base_type = "metrics" if item_type == "columns" else "columns"
```
villebro, line 939: typo?
```suggestion
    base_type = "metrics" if item_type == "metrics" else "columns"
```

tests/integration_tests/sqla_models_tests.py
```diff
@@ -913,54 +913,102 @@ def test_extra_cache_keys_in_sql_expression(
...
-@patch("superset.jinja_context.get_user_email", return_value="abc@test.com")
-def test_extra_cache_keys_in_columns(
-    mock_user_email,
+def test_extra_cache_keys_in_adhoc_metrics_and_columns(
     mock_username,
     mock_user_id,
     sql_expression,
     expected_cache_keys,
     has_extra_cache_keys,
+    item_type,
```
villebro, line 932: Could we type these?
```suggestion
    item_type: "columns" | "metrics",
```

villebro (commented):
Thanks for the fix - a few comments regarding the tests

tests/integration_tests/sqla_models_tests.py
Vitor-Avila: this one is actually correct. When `item_type` is set to `columns`, we want to have an empty `metrics` list in the `base_query_obj`. When `item_type` is set to `metrics`, we want to have an empty `columns` list in it.

I agree this seems a little confusing too. Wondering if it would be better to just split this test in two.

villebro: In that case I'd maybe introduce a new property `overrides` that's a `dict` that you can merge into the `query_obj`. In this case you'd then define it as
```
{
    "metrics": [
        ...
    ],
}
```

or which ever set of properties you want to override.

Vitor-Avila: Ok found a way to make this simpler. Going to commit a change here.

tests/integration_tests/sqla_models_tests.py
Vitor-Avila: of course!

Vitor-Avila:
@villebro thanks for initial pass 🙏 I addressed your feedback + removed a redundant import. Let me know if you have any questions!

tests/integration_tests/sqla_models_tests.py
Vitor-Avila: yup, ended up doing something like that. I was having issues to do the override because the final dict was being constructed with `dict()` but removed that  and it worked.

villebro (approved):
LGTM

geido (approved):

PR #30750, opened 2024-10-30
feat: add uv to Makefile

hughhhh:
### SUMMARY
Adding uv to increase install times for dependencies. Read more here on why uv is better than just running `pip install <package>` [Click here to view why this is more valuable](https://astral.sh/blog/uv)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,14 +20,22 @@ PYTHON=`command -v python3.11 || command -v python3.10`
 
 .PHONY: install superset venv pre-commit
 
-install: superset pre-commit
+install: uv superset pre-commit
+
+uv:
+	# install uv
+	pip install uv
+
+	# create virtualenv
+	uv venv
 
 superset:
+
 	# Install external dependencies
-	pip install -r requirements/development.txt
+	uv pip install -r requirements/development.txt
 
 	# Install Superset in editable (development) mode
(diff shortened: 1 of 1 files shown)


sadpandajoe:
@hughhhh anything we need to do on the docker side for anyone that doesn't use the makefiles?

mistercrunch:
Would need to add uv to dev dependencies, but this raises a few relevant questions on my side:
- a more fully committed move to uv, to replace `pip-compile-multi` and ALL pip commands everywhere (in-docker, docs, everywhere), this may depend on a few things, related -> https://github.com/astral-sh/uv/issues/5487#issuecomment-2407871061
- general move away from Makefile and onto docker-compose for dev-environment setups, we could commit to only support docker-compose for dev workflows and get rid of the makefile altogether. Docker-compose has been working well lately to support a pretty solid dev env with all the bells and whistles (celery, websockets, postgres, ...) in a fairly uniform env. There's probably still use cases for local, out-of-docker development, but unclear how much we want to support this

Oh and if/when we do this, we'll need to add uv to `requirements/development.in`  and `requirements/development.txt`

PR #30753, opened 2024-10-30
docs: add Sarathi to users list

SaiSkandaTNI:
docs: add Sarathi to users list

Since Sarathi does not currently have a general webpage for information, I have linked an official Government Of India press release that references to Sarathi as a Tech Solution.

As and when we have a general webpage or maybe even a public version of a Superset dashboard, I will update this link. Hope this works.


rusackas (approved):
This is great, thanks! Looking forward to a link update when it's relevant. I would also love to see (somewhere) a list of links to public dashboards... maybe I should make a wiki page for anyone who wants to submit those.

PR #30755, opened 2024-10-30
fix(Dashboard): Sync/Async Dashboard Screenshot Generation

### LINKED ISSUE #30645: [4.1.0.rc3] Error to download PDF or Image
### Bug description

Go to one Dashboard
3 dots top right / Download / Export to PDF or Download as Image

One error message is showed. See screenshot

I've tested in environment http://35.87.42.94:8080/

### Screenshots/recordings

image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

geido:
### SUMMARY

PR https://github.com/apache/superset/pull/29187 introduced a new endpoint for generating dashboard screenshots and PR https://github.com/apache/superset/pull/29272 made this endpoint mandatory for the frontend. However, Celery should be an optional dependency in Superset and the endpoint is dependent on Celery to work. This PR makes **Celery optional** for generating dashboard screenshots by:

- **Reintroducing the previous download screenshot functionality**:  
  This update enables dashboards to be downloaded directly from the browser, without using the dedicated API endpoint. This approach has known CORS limitations but will work for most cases where Celery isn't configured.

- **Adding two new feature flags**:
  - **`ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS`**:  
    Controls whether the API endpoint that caches and retrieves dashboard screenshots via a webdriver is enabled, allowing this functionality to be decoupled from the UI. This flag is set to `False` by default.
  
  - **`ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT`**:  
    Controls whether the UI’s download functionality uses webdriver-based screenshot generation. This flag is also `False` by default and it is dependent on `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS`.

Fixes https://github.com/apache/superset/issues/30645

### TESTING INSTRUCTIONS
1. Enter a Dashboard
2. Download a PDF or Image screenshot via the Dashboard menu
3. The screenshot should download normally 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS` and `ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT`
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -54,7 +54,7 @@ from superset.advanced_data_type.plugins.internet_port import internet_port
 from superset.advanced_data_type.types import AdvancedDataType
 from superset.constants import CHANGE_ME_SECRET_KEY
 from superset.jinja_context import BaseTemplateProcessor
-from superset.key_value.types import JsonKeyValueCodec
+from superset.key_value.types import JsonKeyValueCodec, PickleKeyValueCodec
 from superset.stats_logger import DummyStatsLogger
 from superset.superset_typing import CacheConfig
 from superset.tasks.types import ExecutorType
@@ -478,6 +478,14 @@ DEFAULT_FEATURE_FLAGS: dict[str, bool] = {
     "PRESTO_EXPAND_DATA": False,
     # Exposes API endpoint to compute thumbnails
     "THUMBNAILS": False,
+    # Enable the endpoints to cache and retrieve dashboard screenshots via webdriver.
+    # Requires configuring Celery and a cache.
+    "ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS": False,
+    # Generate screenshots (PDF or JPG) of dashboards using the web driver.
+    # When disabled, screenshots are generated on the fly by the browser.
+    # This feature flag is used by the download feature in the dashboard view.
+    # It is dependent on ENABLE_DASHBOARD_SCREENSHOT_ENDPOINT being enabled.
+    "ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT": False,
     "SHARE_QUERIES_VIA_KV_STORE": False,
     "TAGGING_SYSTEM": False,
     "SQLLAB_BACKEND_PERSISTENCE": True,
@@ -789,6 +797,21 @@ EXPLORE_FORM_DATA_CACHE_CONFIG: CacheConfig = {
     # The following parameter only applies to `MetastoreCache`:
     # How should entries be serialized/deserialized?
     "CODEC": JsonKeyValueCodec(),
+}
+
+# Cache for thumbnail images. `CACHE_TYPE` defaults to `SupersetMetastoreCache`
+# that stores the values in the key-value table in the Superset metastore, as it's
+# required for Superset to operate correctly, but can be replaced by any
+# `Flask-Caching` backend.
+THUMBNAIL_CACHE_CONFIG = {
+    "CACHE_TYPE": "SupersetMetastoreCache",
+    "CACHE_DEFAULT_TIMEOUT": 300,
+    # Should the timeout be reset when retrieving a cached value?
+    "REFRESH_TIMEOUT_ON_RETRIEVAL": True,
+    # The following parameter only applies to `MetastoreCache`:
+    # How should entries be serialized/deserialized?
+    # Using PickleKeyValueCodec may be unsafe, use at your own risk.
+    "CODEC": PickleKeyValueCodec(),
 }
 
 # store cache keys by datasource UID (via CacheKey) for custom processing/invalidation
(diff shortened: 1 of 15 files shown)


superset/dashboards/api.py
```diff
@@ -1121,30 +1122,38 @@ def cache_dashboard_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse
...
-            )
-            return self.response(
-                202,
-                cache_key=cache_key,
-                dashboard_url=dashboard_url,
-                image_url=image_url,
-            )
+            cache_dashboard_screenshot.delay(**screenshot_kwargs)
+        else:
+            logger.info("Triggering screenshot SYNC")
```
michael-s-molina, line 1147: Why did you use a thread for the non-async version instead of calling `generate_screenshot` directly?

superset/config.py
```diff
@@ -791,6 +793,20 @@ class D3TimeFormat(TypedDict, total=False):
...
+# required for Superset to operate correctly, but can be replaced by any
+# `Flask-Caching` backend.
+THUMBNAIL_CACHE_CONFIG = {
+    "CACHE_TYPE": "SupersetMetastoreCache",
+    "CACHE_DEFAULT_TIMEOUT": 300,
+    # Should the timeout be reset when retrieving a cached value?
+    "REFRESH_TIMEOUT_ON_RETRIEVAL": True,
+    # The following parameter only applies to `MetastoreCache`:
+    # How should entries be serialized/deserialized?
+    "CODEC": PickleKeyValueCodec(),
```
michael-s-molina, line 807: I wonder if we should have a similar warning as this [one](https://github.com/apache/superset/blob/e79778ac6df9000080ec7ee5b24d59bc46e44423/superset/extensions/metastore_cache.py#L64).

superset/dashboards/api.py
geido: The idea here was to let the cache generation function to proceed without blocking the endpoint response. The subsequent call will start to check for the screenshot and will continue trying if a 404 is returned, keeping the same behavior as with Celery. Any particular concern?

michael-s-molina: My concern is that if the screenshot generation is always async then the feature flag name needs to be changed to something like `USE_CELERY_FOR_DASHBOARD_SCREENSHOTS`. Also, if simply spawning a thread is acceptable for asynchronously generating a screenshot, why use Celery?

michael-s-molina:
Thanks for the updated version of the PR @geido. Do you think we need `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS`? I'm ok with the endpoint being exposed all the time. We have the same characteristic for other endpoints where the feature might not be enabled client-side. Maybe only `ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT` is sufficient?

geido:
> Thanks for the updated version of the PR @geido. Do you think we need `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS`? I'm ok with the endpoint being exposed all the time. We have the same characteristic for other endpoints where the feature might not be enabled client-side. Maybe only `ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT` is sufficient?

This is to be consistent with the thumbnails endpoint that are also behind feature flag. In general, I think it is good to be intentional about enabling these endpoints as they expose performance-heavy functionality.

superset/config.py
```diff
@@ -791,6 +799,21 @@ class D3TimeFormat(TypedDict, total=False):
     "CODEC": JsonKeyValueCodec(),
 }
 
+# Cache for thumbnail images. `CACHE_TYPE` defaults to `SupersetMetastoreCache`
+# that stores the values in the key-value table in the Superset metastore, as it's
+# required for Superset to operate correctly, but can be replaced by any
+# `Flask-Caching` backend.
+THUMBNAIL_CACHE_CONFIG = {
```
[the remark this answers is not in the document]
michael-s-molina: Removing this default configuration will also eliminate the `PickleKeyValueCodec` problem.

michael-s-molina (approved):
Thank you so much for the fix @geido!

Vitor-Avila (approved):
LGTM!

michael-s-molina:
Ok, I think the [exception for the rgbcolor license](https://github.com/apache/superset/pull/30755/commits/efa3b3c592fbedee549463c84efaeef3301ffcb1) resolved CI. @villebro we need an owner approval 🙏🏼

rusackas (approved):
Approving license override

geido:
> Can we have a link to the PR that caused this regression? Just to make it easier to discover and understand how we can avoid these types of things in the future..

Description updated

PR #30760, opened 2024-10-31
feat: add latest partition support for BigQuery

### LINKED ISSUE #17299: Select * Limit is DANGEROUS in BigQuery
The SQL Lab (and unknown other places) currently submits a "select *" query with a limit when loading. This is potentially dangerous in BigQuery as it will query the entire table and every column regardless of the limit. BigQuery has other semantics for querying schema or previews of tables. This functionality should be disabled by default for BigQuery databases until a BigQuery "aware" version can be built.

mistercrunch:
Adding db_engine_spec-related features that enables SQL Lab to show the latest partition when using time partitioning in BigQuery as well as applying a WHERE clause on the latest partition by default when fetching the sample dataset. Turns out that `SELECT * FROM {{...}} LIMIT {n}` can be costly against large tables in BigQuery as it results in a full table scan.

closes https://github.com/apache/superset/issues/17299


tests/integration_tests/db_engine_specs/bigquery_tests.py
```diff
@@ -111,108 +121,45 @@ def values(self):
...
-            "clustering": {"cols": [["c_col1", "c_col2", "c_col3"]]},
         }
-        database.get_indexes = mock.MagicMock(return_value=index_metadata)
-        result = BigQueryEngineSpec.get_extra_table_metadata(
-            database,
-            Table("some_table", "some_schema"),
-        )
-        assert result == expected_result
-
-    def test_get_indexes(self):
```
mistercrunch, line 148: turns out BigQuery's implementation of get_indexes is a no-op -> https://github.com/googleapis/python-bigquery-sqlalchemy/blob/150143ce27c02f1cd711b8df4ffe587a98ada288/sqlalchemy_bigquery/base.py#L1266-L1268

docker/pythonpath_dev/superset_config.py
```diff
@@ -40,6 +40,7 @@
 EXAMPLES_HOST = os.getenv("EXAMPLES_HOST")
 EXAMPLES_PORT = os.getenv("EXAMPLES_PORT")
 EXAMPLES_DB = os.getenv("EXAMPLES_DB")
+SHOW_STACKTRACE = True
```
rusackas, line 43: Was this intended to be part of the PR, or just for local/temporary dev purposes?

mistercrunch: Let me remove that line, now thinking it may make sense to deprecate this flag as it's not evenly implemented.

superset/db_engine_specs/bigquery.py
```diff
@@ -284,66 +290,90 @@ def _truncate_label(cls, label: str) -> str:
...
-        """
-        Get the indexes associated with the specified schema/table.
+    ) -> Select | None:
+        sql = dedent(f"""\
+            SELECT
+                MAX(partition_id) AS max_partition_id
+            FROM `{table.schema}.INFORMATION_SCHEMA.PARTITIONS`
+            WHERE table_name = '{table.table}'
+        """)
+        df = database.get_df(sql)
```
betodealmeida, line 320:
```suggestion
        df = database.get_df(sql, catalog=table.catalog, schema=table.schema)
```

superset/db_engine_specs/bigquery.py
```diff
@@ -284,66 +290,90 @@ def _truncate_label(cls, label: str) -> str:
...
+    def get_time_partition_column(
+        cls,
+        database: Database,
+        table: Table,
+    ) -> str | None:
+        with cls.get_engine(
+            database, catalog=table.catalog, schema=table.schema
+        ) as engine:
+            client = cls._get_client(engine, database)
+            bq_table = client.get_table(f"{table.schema}.{table.table}")
```
betodealmeida, line 333: Here we don't need catalog because `engine` is already bound to it.

tests/integration_tests/db_engine_specs/bigquery_tests.py
betodealmeida: #TIL

betodealmeida (commented):
@mistercrunch can you test with "allow changing catalogs enabled"? I think we need to pass the catalog in a few places.

superset/db_engine_specs/bigquery.py
betodealmeida: I think we need a standard method in the DB engine spec to run SQL. Looks like some engine specs use `engine.execute`, some use `cursor.execute`, some `inspector.bind.execute`, and others `database.get_df`. Something where `catalog` and `schema` are not optional (but can be null) to prevent people from forgetting to pass them:

```
class BaseEngineSpec:
    def execute_sql(self, sql: str, catalog: str | None, schema: str | None):
        ...
```

mistercrunch:
> can you test with "allow changing catalogs enabled"?

Hit an unrelated UI bug trying to edit the json object, reported on internal slack

viyh:
@mistercrunch, any chance of getting this PR fixed/merged soon? This is a big deal for my team as we make heavy use of BigQuery and LIMIT is just not safe. Appreciate it, thanks!

mistercrunch:
@viyh I just updated the PR:
- rebased on latest master
- fixed the potential SQL injection issue using the sqla sql toolkit
- @betodealmeida I tested with and without the "allow catalog" checkbox

Should be go for another round of review and hopefully merging.

One thing I noticed - for another PR, is that we don't seem to get the indexes here anymore. The driver has changed and they made get_indexes a no-op now. Something to look at. This current PR is much more critical given the current state of doing a "full scan" on preview

superset/db_engine_specs/bigquery.py
```diff
@@ -289,66 +295,115 @@ def _truncate_label(cls, label: str) -> str:
...
-        :param table: The table instance to inspect
-        :returns: The indexes
-        """
+        # Build the query
+        query = select(
+            func.max(partitions_table.c.partition_id).label("max_partition_id")
+        ).where(partitions_table.c.table_name == table.table)
+
+        # Compile to BigQuery SQL
+        compiled_query = query.compile(
```
mistercrunch, line 341: there may be a better way to do this. Felt like `dialect` should be in-context here

viyh:
@mistercrunch 🙏 Thank you! Although looks like it failed some different checks now, naturally. :)

mistercrunch:
@viyh  I think I just fixed it, found a better way to retrieve the dialect with `Database.get_dialect()`

superset/db_engine_specs/bigquery.py
```diff
@@ -289,66 +294,115 @@ def _truncate_label(cls, label: str) -> str:
...
+            dialect=database.get_dialect(),
+            compile_kwargs={"literal_binds": True},
+        )
+
+        # Run the query and handle result
+        df = database.get_df(str(compiled_query))
+        if df.empty or df.iat[0, 0] is None:
+            return None
+
+        return df.iat[0, 0]
```
betodealmeida, line 350: Compiling and loading into a dataframe seems a bit overkill? Why not something like:

```python
          with database.get_raw_connection(
              catalog=table.catalog,
              schema=table.schema,
          ) as conn:
              cursor = conn.cursor()
              cursor.execute("""
                  SELECT MAX(partition_id)
                  FROM INFORMATION_SCHEMA.PARTITIONS
                  WHERE table_name=%s
              """, (table.table,))
              if row := cursor.fetchone():
                  return row[0]

              return None
```

We should standardize how DB engine specs run queries, BTW! 😄

betodealmeida (approved):
Looks good, just left a suggestion about using a cursor instead of `get_df(sql)`.

superset/db_engine_specs/bigquery.py
mistercrunch: > We should standardize how DB engine specs run queries, BTW! 

Totally, I just used `get_df` cause it's an easy way to run a query...

PR #30767, opened 2024-10-31
fix(empty dashboards): Allow downloading a screenshot of an empty dashboard

### LINKED ISSUE #30736: Dashboard Screenshot API Endpooint Gives and Error Message when a Dashboard is Empty or has no Charts
### Bug description

When a dashboard is empty OR has only text elements, when the user tries to send an API call to generate a screenshot, they get an error message instead of a screenshot 

### Repro steps: 
1. Create a dashboard with just a title and save it OR create a dashboard with some text elements only and save it 
2. Find the dashboard ID of the dashboard 
3. Use the API to try to generate a screenshot for the dashboard 

Use POST endpoint `/api/v1/dashboard/{{DashboardID}}/cache_dashboard_screenshot/` 
By default, an empty array can be used as the request body to get the dashboard screenshot. Alternatively, it's possible to include below fields in the body:

-     dataMask (array): Specifies the dataMask configuration (controls applied filters).
-     activeTabs (list): Specify the tab(s) that should be included/selected.
-     urlParams (list): List of URL params to be applied to the dashboard.

The request returns an image_url value, that can be used to download the dashboard. It's also possible to append ?download_format=pdf to download a PDF as opposed to an image.

### Expected results
A blank or text-only dashboard image/PDF will be generated.

### Actual results
{"message":"Not found"} error was shown.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

msyavuz:
### SUMMARY

Check if a dashboard contains charts first, if not don't wait for them to load etc. This fixes the bug where if the dashboard is empty screenshot fails.

### TESTING INSTRUCTIONS

1. Create an empty dashboard
2. Download png/pdf screenshot of the empty dashboard

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #30736
- [ ] Required feature flags:
- [ ] Changes UI


superset/utils/webdriver.py
```diff
@@ -379,11 +379,19 @@ def get_screenshot(self, url: str, element_name: str, user: User) -> bytes | Non
...
                 )
             except TimeoutException:
-                logger.exception(
-                    "Selenium timed out waiting for chart containers to draw at url %s",
-                    url,
-                )
-                raise
+                # Check if the page has grid-container class (no chart dashboard)
+                try:
+                    WebDriverWait(driver, self._screenshot_locate_wait).until(
```
geido, line 384:
```suggestion
                    WebDriverWait(driver, 0).until(
```

geido: I don't think you need wait here again

superset/utils/webdriver.py
```diff
@@ -379,11 +379,19 @@ def get_screenshot(self, url: str, element_name: str, user: User) -> bytes | Non
                     )
                 )
             except TimeoutException:
-                logger.exception(
-                    "Selenium timed out waiting for chart containers to draw at url %s",
-                    url,
-                )
-                raise
+                # Check if the page has grid-container class (no chart dashboard)
```
geido, line 382:
```suggestion
                # Fallback to allow a screenshot of an empty dashboard
```

geido (approved):
I did some text enhancements for comments and errors but the general approach looks fine to me. It would be nice to have a better way to differentiate between an empty dashboard (that will never load any charts) and a dashboard that is waiting for chart containers to draw, but I guess this should do fine for now. I tested it locally too. Thanks!

PR #30770, opened 2024-10-31
docs: make it more clear that GLOBAL_ASYNC_QUERIES is experimental/beta

mistercrunch:
Not sure if we're all aligned on this, but I've seen confusion around GLOBAL_ASYNC_QUERIES and the reality is we shouldn't recommend flipping this on until we iron out at least a few things. This comment I'm adding makes it a bit more clear you should steer away unless you know what you're doing

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -486,6 +486,7 @@ DEFAULT_FEATURE_FLAGS: dict[str, bool] = {
     "ESCAPE_MARKDOWN_HTML": False,
     "DASHBOARD_CROSS_FILTERS": True,  # deprecated
     "DASHBOARD_VIRTUALIZATION": True,
+    # This feature flag is stil in beta and is not recommended for production use.
     "GLOBAL_ASYNC_QUERIES": False,
     "EMBEDDED_SUPERSET": False,
     # Enables Alerts and reports new implementation


michael-s-molina:
Maybe it would be interesting to have link to [FEATURE_FLAGS.md](https://github.com/apache/superset/blob/master/RESOURCES/FEATURE_FLAGS.md) at the top of `FEATURE_FLAGS` section.

mistercrunch:
Agreed, though I hate having fragmented information about flags in config.py and in [FEATURE_FLAGS.md](https://github.com/apache/superset/blob/master/RESOURCES/FEATURE_FLAGS.md), we should migrate/merge all intel in one place and reference it from others. 

Maybe the larger task here is to merge what's in config.py onto FEATURE_FLAGS.md, clean it up and reference it from config.py?

mistercrunch:
Hey can we merge this as is, and defer to another task/issue of merging the content/comments in config.py into FEATURE_FLAGS.md?

sadpandajoe (approved):

sadpandajoe:
> Hey can we merge this as is, and defer to another task/issue of merging the content/comments in config.py into FEATURE_FLAGS.md?

@mistercrunch yeah, these would be follow-ups and not a blocker for this PR

PR #30818, opened 2024-11-01
chore(Accessibility): Fix accessibility for 'Show x entries' dropdown in tables

### LINKED ISSUE #30669: Accessibility: Missing Form Control for Show x entries on Tables
## Screenshot

image

## Description

To get to this state: 
1. Create a table chart
2. Enable server pagination 
3. Save the chart and add it to a dashboard 

Currently, the "Show x entries" doesn't have a label when using a screen reader

LevisNgigi:
### SUMMARY

This PR addresses an accessibility issue in the table charts component of Apache Superset. The 'Show x entries' dropdown lacked proper form control labeling, which is essential for screen reader users. The label has been updated to ensure that it provides a clear and descriptive indication of its purpose, improving the overall accessibility of the feature. 

Per page superset

### TESTING INSTRUCTIONS
1. Navigate to a table chart in Apache Superset.
2. Locate the 'Show x entries' dropdown.
3. Use a screen reader (e.g., NVDA or VoiceOver) to navigate to the dropdown.
4. Verify that the screen reader announces the label correctly as "Show [number] entries," where [number] is the selected value.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #30669
- [ ] Required feature flags:
- [ ] Changes UI


geido:
Hello @LevisNgigi you have a conflicting package-lock which I believe should not be there in the first place. Can you remove it?  

From re-reading the issue description, it also seems that the goal was to make also the number of entires accessible, so that the label should read "Show X entries per page". Can we fix that?

Thank you!

LevisNgigi:
> Hello @LevisNgigi you have a conflicting package-lock which I believe should not be there in the first place. Can you remove it?
> 
> From re-reading the issue description, it also seems that the goal was to make also the number of entires accessible, so that the label should read "Show X entries per page". Can we fix that?
> 
> Thank you!

Yes I can remove it. Not sure how it  appeared after fixing the lint issue. Yes I can fix that.

geido:
I still see the package-lock though

LevisNgigi:
> I still see the package-lock though. I think you can just get it from master and replace it locally so it matches.

Should be fixed now.

rusackas:
Running CI :)

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -197,15 +197,20 @@ function SelectPageSize({
...
+    >
+      {t('Show')}{' '}
       <select
         className="form-control input-sm"
         value={current}
         onBlur={() => {}}
         onChange={e => {
           onChange(Number((e.target as HTMLSelectElement).value));
         }}
+        aria-label={t('Show entries per page')}
```
geido, line 213: I think it should be enough to have just this aria-label here and set the number of entries here. Otherwise, the screen reader is reading "Show 10 entries per page, Show entries per page" which I think it is unnecessary. Also, on Firefox, when inspecting for accessibility I am getting this:

Screenshot 2024-11-13 at 16 09 29

I believe we need to have a label that labels the Select input. We can probably change the structure a bit so to keep the same layout but have the proper Label element and accessibility properties. What do you think?

geido (approved):

PR #30819, opened 2024-11-01
fix(plugin-chart-echarts): sort tooltip correctly

villebro:
### SUMMARY
The main ECharts series chart tooltips aren't sorting correctly. This is due to the tooltip keys being populated from `Object.keys` from an object that was originally an array, which of course doesn't have a guaranteed sort order. This issue appears to have been there for a very long time: I treid to hunt it down with Blame, and I stopped searching when I got to three years ago where the code had already changed considerably, but the same bug seemed to be there.

This PR fixes the bug, DRYs the code as the same logic is also needed in Mixed Timeseries chart, and adds simple tests for the new util. Naturally a full `transformProps` integration test would be preferable here, but it felt like too much added code for such a small fix. Also, the util test looks concise enough for people to understand what it's used for.

P.S. In the screenshots below please excuse my awful theme colors (I have those in my devenv to remind myself of the importance of theming work)

P.P.S. I suggest reviewing with hidden whitespace changes, as the changes are mostly that.

### AFTER
Notice how "Tooltip sort by metric" is working as expected (biggest metric values are on the bottom due to stacking):
image

Also sorting series by "Category name" works as expected (this can be seen comparing the tooltip colors vs the bars):
image

### BEFORE
Currently "Tooltip sort by metric" doesn't do anything (metric values are not monotonically growing):
image
When not sorting by metric, the series sort order isn't reflected in the tooltip order (the order is not the same in tooltip as on the bars):
image

### TESTING INSTRUCTIONS
1. Create a bar/line chart with a dimension to sort by
2. Try changing the series sort order to see if it's reflected in the tooltip
3. Try toggling "Tooltip sort by metric" control

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -65,6 +65,7 @@ import {
   extractDataTotalValues,
   extractSeries,
   extractShowValueIndexes,
+  extractTooltipKeys,
   getAxisType,
   getColtypesMapping,
   getLegendProps,
@@ -251,7 +252,7 @@ export default function transformProps(
     legendState,
   });
   const seriesContexts = extractForecastSeriesContexts(
-    Object.values(rawSeries).map(series => series.name as string),
+    rawSeries.map(series => series.name as string),
   );
   const isAreaExpand = stack === StackControlsValue.Expand;
   const xAxisDataType = dataTypes?.[xAxisLabel] ?? dataTypes?.[xAxisOrig];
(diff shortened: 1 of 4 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -642,3 +646,22 @@ export function getTimeCompareStackId(
...
+const TOOLTIP_SERIES_KEY = 'seriesId';
+export function extractTooltipKeys(
+  forecastValue: any[],
+  yIndex: number,
+  richTooltip?: boolean,
+  tooltipSortByMetric?: boolean,
+): string[] {
+  if (richTooltip && tooltipSortByMetric) {
+    return forecastValue
+      .slice()
```
villebro, line 659: Slicing to not mutate the input value (this would have caused issues in the new test)

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -251,7 +252,7 @@ export default function transformProps(
     legendState,
   });
   const seriesContexts = extractForecastSeriesContexts(
-    Object.values(rawSeries).map(series => series.name as string),
+    rawSeries.map(series => series.name as string),
```
villebro, line 255: I'm not sure why we were doing `Object.values` in an array, maybe a typo

eschutho (approved):

michael-s-molina:
> P.S. In the screenshots below please excuse my awful theme colors (I have those in my devenv to remind myself of the importance of theming work)

🤣 🤣 🤣 🤣 🤣 🤣

PR #30827, opened 2024-11-04
fix(dashboard): not filterable column now not emitting cross-filters in table charts

HaroldWrenFinch:
### SUMMARY

Only columns which are filterable emitting filters, which is very handyful when you are using `<a href`-s in a calculated column, and you don't want the click to emit filter.

### TESTING INSTRUCTIONS

1. In the datasets you can turn off "is filterable" column by editing the dataset and turning off the column checkbox. ex.: 
   1. in `video_game_sales` dataset you can turn off the `name` column "is filterable" checkbox
3. Then in the corresponding dashboard you can check it by clicking the column data in the table chart and it will not filtering the data. ex.:
   1. `video game sales` dashboard 
   2. `explore trends` tab 
   4. `games table` chart
   5. click any value in the name column

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
@@ -232,7 +232,9 @@ const processColumns = memoizeOne(function processColumns(
   const metricsSet = new Set(metrics);
   const percentMetricsSet = new Set(percentMetrics);
   const rawPercentMetricsSet = new Set(rawPercentMetrics);
-  const columnConfigs = props.datasource.columns;
+  const columnsByName = new Map(
+    (props.datasource.columns ?? []).map(col => [col.column_name, col]),
+  );
 
   const columns: DataColumnMeta[] = (colnames || [])
     .filter(
--- a/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
@@ -247,9 +247,7 @@ const processColumns = memoizeOne(function processColumns(
       const config = columnConfig[key] || {};
       // for the purpose of presentation, only numeric values are treated as metrics
       // because users can also add things like `MAX(str_col)` as a metric.
-      const isFilterable = columnConfigs.find(
-        c => c.column_name === key,
-      )?.filterable;
+      const isFilterable = columnsByName.get(key)?.filterable;
       const isMetric = metricsSet.has(key) && isNumeric(key, records);
       const isPercentMetric = percentMetricsSet.has(key);
       const label =
--- a/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
@@ -232,6 +232,7 @@ const processColumns = memoizeOne(function processColumns(
   const metricsSet = new Set(metrics);
   const percentMetricsSet = new Set(percentMetrics);
   const rawPercentMetricsSet = new Set(rawPercentMetrics);
+  const columnConfigs = props.datasource.columns;
 
   const columns: DataColumnMeta[] = (colnames || [])
     .filter(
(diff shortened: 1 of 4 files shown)


msyavuz:
This pr is becoming stale. It needs a rebase first but we can maybe push it through with @kgabryje. As i understand it this is intended to block cross filters when is filterable is off.

rusackas:
I was tempted to mark this as Draft and set it aside, but would love to see it through @HaroldWrenFinch - this bugs me too! Any hope of a rebase here?

rusackas:
@HaroldWrenFinch — this one has gone stale and now conflicts with master. The change is small (+8/-1) and CI was green before the conflict appeared. Could you rebase on `master` and resolve, and ideally add a quick test for the not-filterable cross-filter behavior? Once it's clean we can get a maintainer to take a fresh look. Let us know if you'd rather we close it.

HaroldWrenFinch:
> @HaroldWrenFinch — this one has gone stale and now conflicts with master. The change is small (+8/-1) and CI was green before the conflict appeared. Could you rebase on `master` and resolve, and ideally add a quick test for the not-filterable cross-filter behavior? Once it's clean we can get a maintainer to take a fresh look. Let us know if you'd rather we close it.

Hi @rusackas, thanks for the patience and the nudge!          
                                                                
  I've rebased onto the latest master and resolved the conflicts. I also added a unit test for the not-filterable cross-filter behavior in TableChart.test.tsx.
                                                             
  Ready for another look. 🙏

rusackas (approved):
@HaroldWrenFinch nice, thanks! Respecting the column's `filterable` flag looks good, and I like that you added a test. It's gone conflicting; Approving and rebasing, let's see how CI does...

rusackas:
Ahh, I had a go at rebasing this myself but GitHub won't let me push to it - the PR's coming from your fork's `master` branch, so maintainer edits are disabled on it. Could you rebase on current `master` and resolve the conflict on your end? Otherwise I'm happy to close this out for now... just let me know which you'd prefer. Still think the fix is worth landing :)

rusackas:
Setting this to draft whilst we await word/support from @HaroldWrenFinch. Please mark as ready for review when it is, or enable maintainer commits on your fork if you want me to chip in. Thanks!

HaroldWrenFinch:
> Ahh, I had a go at rebasing this myself but GitHub won't let me push to it - the PR's coming from your fork's `master` branch, so maintainer edits are disabled on it. Could you rebase on current `master` and resolve the conflict on your end? Otherwise I'm happy to close this out for now... just let me know which you'd prefer. Still think the fix is worth landing :)

Hi @rusackas, thanks for your patience!

I've rebased onto current master and resolved the conflict — the PR is now a single clean commit on top of master. 

Thanks again for keeping this moving! 🙏

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -232,6 +232,7 @@ const processColumns = memoizeOne(function processColumns(
   const metricsSet = new Set(metrics);
   const percentMetricsSet = new Set(percentMetrics);
   const rawPercentMetricsSet = new Set(rawPercentMetrics);
+  const columnConfigs = props.datasource.columns;
```
rusackas:
```suggestion
  const columnsByName = new Map(
    (props.datasource.columns ?? []).map(col => [col.column_name, col]),
  );
```
Build the lookup once here (pairs with the suggestion below).

superset-frontend/plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -244,6 +245,9 @@ const processColumns = memoizeOne(function processColumns(
       const config = columnConfig[key] || {};
       // for the purpose of presentation, only numeric values are treated as metrics
       // because users can also add things like `MAX(str_col)` as a metric.
+      const isFilterable = columnConfigs.find(
+        c => c.column_name === key,
+      )?.filterable;
```
rusackas:
```suggestion
      const isFilterable = columnsByName.get(key)?.filterable;
```
...then it's an O(1) read per column.

rusackas (commented):
I was worried about a breaking change here, potentially, but the behavior here looks right to me. `isFilterable !== false` means existing charts keep emitting cross-filters from every column, and it only backs off when someone explicitly unchecks "is filterable" (which is `True` by default server-side). So no default change to worry about. 

The bot's O(n²) note is fair but minor. I left a couple of suggestions to swap the per-column `.find` for a one-time lookup map, should be one-click to "apply changes" if you agree.

The "object injection" bot comment is a false positive (it's an array `.find` with `===`, not dynamic key access), so safe to ignore. And thanks for adding the test!

rusackas:
Thanks for the improvement! I can certainly use it in some of my own dashboards :D

rusackas (approved):
re-stamping here.

PR #30833, opened 2024-11-04
feat(tags): Export and Import Functionality for Superset Dashboards and Charts

asher-lab:
### SUMMARY

This PR is related to Discussion: #30629 by Spens

#### Motivation
We have adopted the tagging functionality and it is working quite well for us for dashboards and and charts. Our use case and process requires us to build dashboards on a development server, export the dashboards, commit the YAMLs to source control, and finally import the ZIPs to our production server. We would like the tags we assign on the development server to be captured in the export so they are replicated to our production server.

#### Proposed Change
UI elements are not affected by this change. All changes are related to updated logic/processing of the existing export and import functionality.

Only tags of type "custom" are exported. Other tags that exist but are not exported are "type" and "owner".

Export Logic
When an object (Dashboard/Chart) is exported, a new parameter is added to the exported YAML file. The parameter is "tags" and the value is an array of tag names. Here is an example:

```
slice_name: Vaccine Candidates per Approach & Stage
description: null
tags:
  - CovidTag
  - Items that start with the letter V
  - Superset Example Data
certified_by: null
```

...
Because tags have additional information (for example, a description), a new file is needed to hold this information. The file is called "tags.yaml" and lives in the top level of the zip file. For example:

```
dashboard_export_20241017T025146.zip
  charts (folder)
  dashboards (folder)
  databases (folder)
  datasets (folder)
  metadata.yaml (file)
  tags.yaml (file)
```

The format of the tags.yaml file is an array of tags

```
tags:
- tag_name: CovidTag
  description: Everything related to Covid
- tag_name: Items that start with the letter V
  description: Just for fun, tag Vanilla, Vikings, and anything that starts with V
- tag_name: Superset Example Data
  description: Native example data included with Superset
```

Import Logic
When an object is imported, the code checks to see if any tags included in the import file.
If yes, the tag is created if it does not already exist then the tag is assigned to the imported object. If the tag already exists, it is just assigned with no changes to the tag.

Details

1. A tag can only be created if the user doing the import has permission to create tags. Otherwise the tagging is ignored.
2. A created tag picks up the current time as the "created_on" and "changed_on" properties for the tag
3. A created tag picks up the current user as the "created_by_fk" and "changed_by_fk"
4. A created tag picks up the description from the tags.yaml file described above
5. If a tag already exists, the description is not updated (for consistency with other export/import behavior - for example, if you import a dashboard and the chart included in the zip already exists, Superset does not update that existing chart)
Existing Superset behavior: When importing objects that already exist, the user is prompted whether it is ok to overwrite. If no, the import process aborts. If yes, the old object is replaced with the new object.

To be consistent with this behavior, when importing an object and specifying "overwrite", old tags that are no longer specified with the object are removed, new tags that didn't originally exist are created, old tags that still remain in the new set of tags are retained (descriptions are not updated for example)
1. It is important to note that in this proposal, imported tags are not added to existing tags
This is intentional because for the use case of trying to mirror the settings and properties from one system to another, you must have the ability to remove tags from the target system. Otherwise, for example, if you rename a tag, the new name would get added and the old name would stick around which is not the desired behavior.

Thank you to anyone taking the time to read and comment!

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


docker/pythonpath_dev/superset_config.py
```diff
@@ -97,7 +97,10 @@ class CeleryConfig:
 
 CELERY_CONFIG = CeleryConfig
 
-FEATURE_FLAGS = {"ALERT_REPORTS": True}
+FEATURE_FLAGS =  {
```
michael-s-molina, line 100: Tags should be disabled by default.

superset/commands/chart/export.py
```diff
@@ -71,12 +72,26 @@ def _file_content(model: Slice) -> str:
         if model.table:
             payload["dataset_uuid"] = str(model.table.uuid)
 
+        # Fetch tags from the database
+        tags = (
```
michael-s-molina, line 76: You need to account for the scenario where Tags are disabled.

michael-s-molina:
Thank you for the PR @asher-lab. Could you add unit tests for the new feature with the feature flag on and off?

asher-lab:
Good day, @michael-s-molina. We've added logic to handle cases when the Tagging System is disabled in Superset. When disabled, the `tags.yaml` file and any associated tags within charts.yaml and dashboard.yaml will not be exported, and the same applies to the import operation.

Could you provide guidance on creating a unit test for this new feature? I have limited experience with writing unit tests, so any help would be appreciated.

Thank you!

michael-s-molina:
> Could you provide guidance on creating a unit test for this new feature? I have limited experience with writing unit tests, so any help would be appreciated.

Hi @asher-lab. You can check this [file](https://github.com/apache/superset/blob/master/tests/unit_tests/dashboards/commands/importers/v1/import_test.py) for examples.

asher-lab:
Hi @michael-s-molina . Added unit tests for the new feature with the feature flag TAGGING_SYSTEM on and off. Thank you.

asher-lab:
I seen a bunch of tests: 10 are failing and 23 are successful. Could you provide what tests that I need to look at?

superset/dashboards/schemas.py
```diff
@@ -453,7 +453,7 @@ class ImportV1DashboardSchema(Schema):
     certified_by = fields.String(allow_none=True)
     certification_details = fields.String(allow_none=True)
     published = fields.Boolean(allow_none=True)
-
+    tag = fields.List(fields.String(), allow_none=True)
```
Vitor-Avila, line 456: Could we have it named as `tags`, since it's a list?

superset/charts/schemas.py
```diff
@@ -1550,7 +1550,7 @@ class ImportV1ChartSchema(Schema):
     dataset_uuid = fields.UUID(required=True)
     is_managed_externally = fields.Boolean(allow_none=True, dump_default=False)
     external_url = fields.String(allow_none=True)
-
+    tag = fields.List(fields.String(), allow_none=True)
```
Vitor-Avila, line 1553: Could we have it named as `tags`, since it's a list?

superset/commands/chart/export.py
```diff
@@ -71,12 +73,23 @@ def _file_content(model: Slice) -> str:
         if model.table:
             payload["dataset_uuid"] = str(model.table.uuid)
 
+        # Fetch tags from the database if TAGGING_SYSTEM is enabled
+        if feature_flag_manager.is_feature_enabled("TAGGING_SYSTEM"):
+            tags = (
+                model.tags if hasattr(model, "tags") else []
+            )
+            # Filter out any tags that contain "type:" in their name
+            payload["tag"] = [tag.name for tag in tags if not any(x in tag.name for x in ["type:", "owner:"])]
```
Vitor-Avila, line 82: Since users can create tags with those strings, could we filter for type, instead? something like:
``` python
payload["tags"] = [tag.name for tag in tags if tag.type == TagType.custom]
```

superset/dashboards/schemas.py
Vitor-Avila, line 456: Same here

superset/commands/dashboard/export.py
```diff
@@ -159,6 +161,13 @@ def _file_content(model: Dashboard) -> str:
 
         payload["version"] = EXPORT_VERSION
 
+        # Check if the TAGGING_SYSTEM feature is enabled
+        if feature_flag_manager.is_feature_enabled("TAGGING_SYSTEM"):
+            tags = (
+                model.tags if hasattr(model, "tags") else []
+            )
+            payload["tag"] = [tag.name for tag in tags if not any(prefix in tag.name for prefix in ["type:", "owner:"])]
```
Vitor-Avila, line 169: Same here

Vitor-Avila:
Thank you so much @asher-lab! This is super cool and very useful 😍 🙌 I left a few comments, and I plan to give it another pass Tomorrow.

asher-lab:
Hi @Vitor-Avila , we have fix the integration test issue, updated it to use "tags" instead of "tag" and update filter using TagType. Thank you!

asher-lab:
Hi @Vitor-Avila and @michael-s-molina, if these are merged in, when is the expected date it be released?

rusackas:
Running CI 🤞 This would definitely be part of 5.0, which needs a new estimated date. It will have missed the boat for 4.1.0, but could make a 4.1.1 or a 4.2.0, though it's unclear if/when those would happen. Feel free to join us on [slack](http://bit.ly/join-superset-slack) in the release strategy or release announcement channels, if you want to get involved with the process.

asher-lab:
Currently fixing the tests that are failing. Will be making a commit in order to fix those. Thank you.

asher-lab:
Good morning team, @rusackas @Vitor-Avila and @michael-s-molina. Could I get another run for this one please. I did run it on my own fork and all tests have passed. 🤞

Vitor-Avila:
hey @asher-lab,
Sorry for the delay on my end. Approved the workflows and hoping to check it today

betodealmeida (commented):
Taking a look at this, but in the meantime a few comments:

1. I like the approach of having `tags.yml` in the ZIP, I think it aligns super well with the intent of the import/export flow.
2. Using `inspect` to figure out the caller makes me nervous, let's see if we can find a better solution.
3. Every time we change import/export we need to make sure that (1) old exports still work, and (2) new exports work in older version — that is, everything should be backwards and forwards compatible, ideally.

Vitor-Avila:
> Taking a look at this, but in the meantime a few comments:
> 
> 1. I like the approach of having `tags.yml` in the ZIP, I think it aligns super well with the intent of the import/export flow.
> 2. Using `inspect` to figure out the caller makes me nervous, let's see if we can find a better solution.
> 3. Every time we change import/export we need to make sure that (1) old exports still work, and (2) new exports work in older version — that is, everything should be backwards and forwards compatible, ideally.

Thanks @betodealmeida! I can help testing 3 if needed. I agree with 1 and 2 - I'm wondering if there's a better way to handle this flow. 

Another consideration is if we should adopt the similar `uuid` approach for tags, and also allow exporting tags via the UI (as an individual object type), and then the `uuid` would dictate if the tag gets updated or created. I don't think it's worth in the current scope, but if we end up expanding its features (for example, the ability to assign permissions to tags, etc) might be beneficial. This can also be re-evaluated if we ever get there

asher-lab:
Hello @betodealmeida , we have refrained from the use of inspect. Kindly let me know your comments on this. Looking forward for the next steps. cc @Vitor-Avila requesting to run the check again please 😊 Thank you.

superset/commands/chart/export.py
```diff
@@ -85,3 +102,12 @@ def _export(
...
             yield from ExportDatasetsCommand([model.table.id]).run()
+
+        # Check if the calling class is ExportDashboardCommands
+        if (
+            export_related
+            and ExportChartsCommand._include_tags
+            and feature_flag_manager.is_feature_enabled("TAGGING_SYSTEM")
+        ):
+            chart_id = model.id
+            yield from ExportTagsCommand().export(chart_ids=[chart_id])
```
betodealmeida, line 113: Why only include tags for dashboard exports? Why not always include tags here?

superset/commands/chart/importers/v1/__init__.py
```diff
@@ -46,7 +49,14 @@ class ImportChartsCommand(ImportModelsCommand):
     import_error = ChartImportError
 
     @staticmethod
-    def _import(configs: dict[str, Any], overwrite: bool = False) -> None:
+    def _import(
+        configs: dict[str, Any],
+        overwrite: bool = False,
+        contents: Optional[dict[str, Any]] = None,
```
betodealmeida, line 55: Nit, no need for `Optional`:

```suggestion
        contents: dict[str, Any] | None = None,
```

(You might need to include `from __future__ import annotations`.)

superset/commands/chart/importers/v1/__init__.py
```diff
@@ -15,21 +15,24 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import Any
+from typing import Any, Optional
```
betodealmeida, line 18:
```suggestion
from typing import Any
```

superset/commands/dashboard/importers/v1/__init__.py
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from typing import Any
+from typing import Any, Optional
```
betodealmeida, line 18: Same here.

superset/commands/export/assets.py
```diff
@@ -50,15 +51,17 @@ def run(self) -> Iterator[tuple[str, Callable[[], str]]]:
...
             ExportDashboardsCommand,
             ExportSavedQueriesCommand,
         ]
         for command in commands:
-            ids = [model.id for model in command.dao.find_all()]
-            for file_name, file_content in command(ids, export_related=False).run():
-                if file_name not in seen:
-                    yield file_name, file_content
-                    seen.add(file_name)
+            if hasattr(command, "dao"):
```
betodealmeida, line 59: I'm confused by this — does `ExportTagsCommand` have a DAO? If it has, why the `hasattr` check; if it doesn't, why include it in `commands`?

superset/commands/importers/v1/utils.py
```diff
@@ -214,3 +218,79 @@ def get_contents_from_bundle(bundle: ZipFile) -> dict[str, str]:
...
+    if "tags.yaml" in contents:
+        try:
+            tags_config = yaml.safe_load(contents["tags.yaml"])
+            for tag_info in tags_config.get("tags", []):
+                tag_name = tag_info.get("tag_name")
+                description = tag_info.get("description", None)
+                if tag_name:
+                    tag_descriptions[tag_name] = description
+        except yaml.YAMLError as err:  # Renamed to 'err' for clarity
+            logger.error("Error parsing tags.yaml: %s", err)  # Used lazy logging
```
betodealmeida, line 247: It's usually better to have small try/except blocks:

```suggestion
        try:
            tags_config = yaml.safe_load(contents["tags.yaml"])
        except yaml.YAMLError as err:  # Renamed to 'err' for clarity
            logger.error("Error parsing tags.yaml: %s", err)  # Used lazy logging
            tags_config = {}
            
        for tag_info in tags_config.get("tags", []):
            tag_name = tag_info.get("tag_name")
            description = tag_info.get("description", None)
            if tag_name:
                    tag_descriptions[tag_name] = description
```

betodealmeida (commented):
Looks great! I left a few questions and comments.

superset/commands/export/assets.py
asher-lab: My first thought was since it is export I "think" I should add it in the Assets commands, but no ExportTagsCommand  does not use dao at all. We have removed it in the `commands`. Thanks!!

asher-lab:
Hello @betodealmeida could I get another review and run please. Thank you! cc: @Vitor-Avila  and @rusackas

asher-lab:
Good morning team. cc: @betodealmeida  @

asher-lab:
Good morning @betodealmeida , resolve the merge conflicts one more run please. Thank you!

asher-lab:
Good morning team, could we get a follow up on this PR please. @betodealmeida  @Vitor-Avila @michael-s-molina . Thank you very much!

superset/commands/chart/export.py
Vitor-Avila: I think this logic is in place to prevent calling `ExportTagsCommand.export()` multiple times. It seems that if it's a dashboard export, then we don't export tags at the chart export level, and then at the dashboard level:
``` python
ExportTagsCommand.export( dashboard_ids=dashboard_ids, chart_ids=chart_ids )
```
When exporting charts only, then we export tags at the chart level.

superset/commands/chart/importers/v1/__init__.py
```diff
@@ -47,7 +51,13 @@ class ImportChartsCommand(ImportModelsCommand):
...
 
     @staticmethod
-    def _import(configs: dict[str, Any], overwrite: bool = False) -> None:
+    def _import(
+        configs: dict[str, Any],
+        overwrite: bool = False,
+        contents: dict[str, Any] | None = None,
+    ) -> None:
+        if contents is None:
+            contents = {}
```
Vitor-Avila, line 60:
``` python
contents = {} if contents is None else contents
```

superset/commands/dashboard/importers/v1/__init__.py
```diff
@@ -58,9 +62,15 @@ class ImportDashboardsCommand(ImportModelsCommand):
...
+    # pylint: disable=too-many-branches, too-many-locals, too-many-statements
     @staticmethod
-    def _import(configs: dict[str, Any], overwrite: bool = False) -> None:
+    def _import(
+        configs: dict[str, Any],
+        overwrite: bool = False,
+        contents: dict[str, Any] | None = None,
+    ) -> None:
+        if contents is None:
+            contents = {}
```
Vitor-Avila, line 73: Same here

superset/commands/database/importers/v1/__init__.py
```diff
@@ -42,7 +42,11 @@ class ImportDatabasesCommand(ImportModelsCommand):
     import_error = DatabaseImportError
 
     @staticmethod
-    def _import(configs: dict[str, Any], overwrite: bool = False) -> None:
+    def _import(
+        configs: dict[str, Any],
+        overwrite: bool = False,
+        contents: Optional[dict[str, Any]] = None,
```
Vitor-Avila, line 48: You can do `from __future__ import annotations` in this file as well so that you don't have to use `Optional`

asher-lab:
Noted on this Vitor, waiting for @betodealmeida feedback. Thank you! :)

superset/commands/importers/v1/utils.py
```diff
@@ -214,3 +218,80 @@ def get_contents_from_bundle(bundle: ZipFile) -> dict[str, str]:
...
+                tag_descriptions[tag_name] = description
+    existing_tags = (
+        db_session.query(TaggedObject)
+        .filter_by(object_id=object_id, object_type=object_type)
+        .all()
+    )
+
+    for tag_name in new_tag_names:
+        try:
+            tag = db_session.query(Tag).filter_by(name=tag_name).first()
```
betodealmeida, line 258: This is going to be potentially super slow, better to move the query outside of the loop. Something like:

```python
existing_tags = {tag.name: tag for tag in db_session.query(Tag).filter(tag.name.in_(new_tag_names))}
```

Then you can do:

```python
for tag_name in new_tag_names:
    tag = existing_tags.get(tag_name)
    if tag is None:
        tag = Tag(...)
```

superset/commands/tag/export.py
```diff
@@ -0,0 +1,131 @@
...
+        file_content = yaml.safe_dump(payload, sort_keys=False)
+        return file_content
+
+    @staticmethod
+    def export(
+        dashboard_ids: Optional[Union[int, List[Union[int, str]]]] = None,
+        chart_ids: Optional[Union[int, List[Union[int, str]]]] = None,
+    ) -> Iterator[tuple[str, Callable[[], str]]]:
+        if not feature_flag_manager.is_feature_enabled("TAGGING_SYSTEM"):
+            yield from iter([])
```
betodealmeida, line 126: I think we can just `return` here.

betodealmeida (commented):
Sorry for the delay reviewing this! It looks good, but I left a few comments.

asher-lab:
Thank you Beto, these are noted and have a happy holidays! Will be making these changes after the Holidays!!

asher-lab:
Good morning @betodealmeida and @Vitor-Avila , could I get another run for the pipeline please. Thank you!!

asher-lab:
Good morning @betodealmeida and @Vitor-Avila could I get another run please? I did fix the issue with one of the tests. Thank you so much.

asher-lab:
Thank you @betodealmeida  and @Vitor-Avila. All checks have passed. Are we ready for merging our superset changes? If there is something that needs change/update. Please let me know. Thank you!

superset/commands/dashboard/export.py
```diff
@@ -173,7 +183,14 @@ def _export(
 
         if export_related:
             chart_ids = [chart.id for chart in model.slices]
+            dashboard_ids = model.id
+            ExportChartsCommand.disable_tag_export()
             yield from ExportChartsCommand(chart_ids).run()
+            ExportChartsCommand.enable_tag_export()
```
betodealmeida, line 189: Setting the class attribute can be problematic if we have multiple exports happening at the same time in the same process, there could be a race condtion. It would be better to instantiate the command and set it there to prevent this:

```python
command = ExportChartsCommand(chart_ids)
command.disable_tag_export()
yield from command.run()
command.enable_tag_export()
```

But I think it's better and clearer if we just pass a parameter when we instantiate the command:

```suggestion
            yield from ExportChartsCommand(chart_ids, include_tags=True).run()
```

And then we update `ExportChartsCommand` accordingly:

```python
class ExportChartsCommand(ExportModelsCommand):
    def __init__(self, *args: Any, include_tags: bool = True, **kwargs: Any):
        self._include_tags = include_tags
        super(*args, **kwargs)
```

superset/commands/importers/v1/utils.py
```diff
@@ -214,3 +219,88 @@ def get_contents_from_bundle(bundle: ZipFile) -> dict[str, str]:
...
+    for tag_name in target_tag_names:
+        try:
+            tag = existing_tags.get(tag_name)
+
+            # If tag does not exist, create it
+            if tag is None:
+                description = tag_descriptions.get(tag_name, None)
+                tag = Tag(name=tag_name, description=description, type="custom")
+                db_session.add(tag)
+                db_session.commit()
```
betodealmeida, line 271: Let's not commit or rollback here, we should leave that to the command via the `@transaction` decorator. See https://github.com/apache/superset/issues/25108.

betodealmeida (changes requested):
I apologize for the delay in taking another look at this! This is looking great, I have just a couple comments:

- The `enable_tags()`/`disable_tags()` is not idiomatic, and can cause some race conditions. I left a suggestion of a better approach.
- We should leave all commits and rollbacks to the `@transaction` decorator, instead of doing it manually.

asher-lab:
Hello thank you @betodealmeida we will be making the appropriate changes as well so we can merge this as soonest hopefully. Thanks!

asher-lab:
Good morning @betodealmeida @Vitor-Avila could I get another run for this pipeline pls? Thank you! We have made the changes based on the feedback above. Thank you!

betodealmeida (approved):
Thank you so much for the work here, @asher-lab!

asher-lab:
Thank you @betodealmeida and @Vitor-Avila !!!

tireswing901:
@betodealmeida and @Vitor-Avila , 
Thanks for all of your help! When will this change be included in a release?

Vitor-Avila:
hey @michael-s-molina @sadpandajoe do you know if this is going to be included in `v5.0`?

sadpandajoe:
> hey @michael-s-molina @sadpandajoe do you know if this is going to be included in `v5.0`?

Unfortunately this is a feature that was merged post-5.0 cut. So it'll make it into the next release after 5.0 that is cut from master as we normally only cherry-pick fixes into our RCs after the initial cut.

PR #30848, opened 2024-11-06
fix(time_comparison): Allow deleting dates when using custom shift

Antonio-RiveroMartnez:
### SUMMARY
This PR fixes the date picker for custom start date in time comparison section so users can click the `x` to delete the date and select a new one. Currently it shows `Invalid date` and the whole date picker gets NaN values.

When the user clicks the `x` the control panel shows the error like we do for other fields and the chart cannot be created/updated because the button is disabled.

Before

https://github.com/user-attachments/assets/f6e1e513-9a18-4547-b801-35ab05bf2e9c

After
Creating

https://github.com/user-attachments/assets/ee05678e-32bf-4144-aefd-a99937499168

Updating

https://github.com/user-attachments/assets/3fcabe60-db4b-45b6-bca1-001a42add51c

### TESTING INSTRUCTIONS
Create a new Table with Time Comparison chart.
Expand the Time Comparison section.
Select Custom date on Time Shift.
Select a date on shift start date.
Click on the x next to it to remove it.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
@@ -114,7 +114,7 @@ export const timeComparisonControls: ({
                 controlState.value === 'Invalid date')
             ) {
               newState.externalValidationErrors = [
-                t('A date is required when using Custom date shift'),
+                t('A date is required when using custom date shift'),
               ];
             } else {
               newState.externalValidationErrors = [];
--- a/superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
@@ -18,7 +18,11 @@
  */
 import { t, ComparisonType } from '@superset-ui/core';
 
-import { ControlPanelSectionConfig } from '../types';
+import {
+  ControlPanelSectionConfig,
+  ControlPanelState,
+  ControlState,
+} from '../types';
 
 const fullChoices = [
   ['1 day ago', t('1 day ago')],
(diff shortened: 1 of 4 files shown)


superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
```diff
@@ -94,9 +98,29 @@ export const timeComparisonControls: ({
...
+            const { form_data } = state;
+            const { time_compare } = form_data;
+            const newState = { ...controlState };
+            if (
+              time_compare === 'custom' &&
+              (controlState.value === '' ||
+                controlState.value === 'Invalid date')
+            ) {
+              newState.externalValidationErrors = [
+                t('A date is required when using Custom date shift'),
```
geido, line 117: Is this "Custom" with capital C intended?

geido (approved):

superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
Antonio-RiveroMartnez: Oh! Good catch, thanks

superset-frontend/packages/superset-ui-chart-controls/src/sections/timeComparison.tsx
```diff
@@ -94,9 +98,29 @@ export const timeComparisonControls: ({
...
+            state: ControlPanelState,
+            controlState: ControlState,
+          ) => {
+            const { form_data } = state;
+            const { time_compare } = form_data;
+            const newState = { ...controlState };
+            if (
+              time_compare === 'custom' &&
+              (controlState.value === '' ||
+                controlState.value === 'Invalid date')
```
kgabryje, line 114: We use „Invalid date” here and in explore. Should we move it to a const in chart controls package? Also, should it be translated or does that string come from some external library?

Antonio-RiveroMartnez: Let me move that to a const, about translating, that's coming form moment indeed.

kgabryje: Can we write a small unit test for that „Invalid date” string to future proof it in case it ever changes in momentjs?

Antonio-RiveroMartnez: Sure thing

PR #30851, opened 2024-11-06
fix: Graph chart colors

michael-s-molina:
### SUMMARY
Fixes the Graph chart to obey to the color scheme when there are no color categories.

https://github.com/user-attachments/assets/3ba531af-b925-4573-aca6-5996be149f5b

https://github.com/user-attachments/assets/c9b60d2c-71df-49b1-a8cc-c55d423d6c71

### TESTING INSTRUCTIONS
1 - Check that the color scheme is applied when there are no color categories.
2 - Check that the color scheme is applied when there are color categories. 

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Graph/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Graph/transformProps.ts
@@ -198,6 +198,7 @@ export default function transformProps(
   const refs: Refs = {};
   const metricLabel = getMetricLabel(metric);
   const colorFn = CategoricalColorNamespace.getScale(colorScheme as string);
+  const firstColor = colorFn.range()[0];
   const nodes: { [name: string]: number } = {};
   const categories: Set<string> = new Set();
   const echartNodes: EChartGraphNode[] = [];
@@ -256,11 +257,10 @@ export default function transformProps(
       : undefined;
     const sourceNodeColor = sourceCategoryName
       ? colorFn(sourceCategoryName)
-      : colorFn('node');
+      : firstColor;
     const targetNodeColor = targetCategoryName
       ? colorFn(targetCategoryName)
-      : colorFn('node');
-    const linkColor = colorFn('link');
+      : firstColor;
 
     const sourceNode = getOrCreateNode(
       sourceName,
@@ -283,7 +283,7 @@ export default function transformProps(
       target: targetNode.id,
       value,
       lineStyle: {
-        color: linkColor,
+        color: firstColor,
       },
       emphasis: {},
       select: {},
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Graph/transformProps.ts
```diff
@@ -242,14 +248,39 @@ export default function transformProps(
...
+      );
+      sourceNodeColor = colorFn(sourceCategoryName);
+    }
+    if (targetCategory) {
+      targetCategoryName = getCategoryName(
+        targetCategory,
+        link[targetCategory],
+      );
+      targetNodeColor = colorFn(targetCategoryName);
+    }
```
justinpark, line 270: Can this let assignment consolidate the logic like this?
 
```suggestion
    const sourceCategoryName = sourceCategory ? getCategoryName(
        sourceCategory,
        link[sourceCategory],
      ) : undefined;
    const  targetCategoryName = targetCategory ? getCategoryName(
        targetCategory,
        link[targetCategory],
      ) : undefined;
    const sourceNodeColor = sourceCategory ? colorFn(sourceCategoryName) : colorFn('node');
    const targetNodeColor = targetCategory ? colorFn(targetCategoryName) : colorFn('node');
    const linkColor = colorFn('link');
```

superset-frontend/plugins/plugin-chart-echarts/src/Graph/transformProps.ts
```diff
@@ -248,8 +254,26 @@ export default function transformProps(
...
       : undefined;
-    const sourceNode = getOrCreateNode(sourceName, source, sourceCategoryName);
-    const targetNode = getOrCreateNode(targetName, target, targetCategoryName);
+    const sourceNodeColor = sourceCategoryName
+      ? colorFn(sourceCategoryName)
+      : colorFn('node');
+    const targetNodeColor = targetCategoryName
+      ? colorFn(targetCategoryName)
+      : colorFn('node');
+    const linkColor = colorFn('link');
```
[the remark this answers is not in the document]
michael-s-molina: Make sense. Done.

superset-frontend/plugins/plugin-chart-echarts/src/Graph/transformProps.ts
```diff
@@ -258,7 +282,9 @@ export default function transformProps(
       source: sourceNode.id,
       target: targetNode.id,
       value,
-      lineStyle: {},
+      lineStyle: {
+        color: firstColor,
```
villebro, line 286: Could we use `sourceNodeColor` here? I feel that helps distinguish the source/target better than just using `firstColor`

villebro (approved):
LGTM, thanks for the improvement and the multiple iterations!

PR #30857, opened 2024-11-06
fix(sqllab): Over-rendering on result table

justinpark:
### SUMMARY
The issue of the grid table being over-rendered every time the tab is changed is caused by the mapping of `orderedColumns`. To prevent this over-rendering, this commit modified the code to use `useMemo` and `React.memo`, ensuring that the same properties from the previous tab are passed to avoid unnecessary re-renders.

### TESTING INSTRUCTIONS
locally perf check

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina (approved):

rusackas:
This has needed a rebase for a long time. I'll mark it as draft, but it may be closed eventually if it doesn't get any activity.

rusackas:
Hi @justinpark\! Just following up - this PR still needs a rebase when you get a chance. Let us know if it's still on your radar\!

PR #30858, opened 2024-11-06
fix(chart data): removing query from /chart/data payload when accessing as guest user

fisjac:
### SUMMARY

When accessing a chart as a guest user, the `query` field is attached to the payload revealing the table names of db schema. This fix introduces a check for whether or not the user is a guest user, and removes the query field from the data payload.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -394,8 +394,13 @@ class ChartDataRestApi(ChartRestApi):
             )
 
         if result_format == ChartDataResultFormat.JSON:
+            queries = result["queries"]
+            if security_manager.is_guest_user():
+                for query in queries:
+                    with contextlib.suppress(KeyError):
+                        del query["query"]
             response_data = json.dumps(
-                {"result": result["queries"]},
+                {"result": queries},
                 default=json.json_int_dttm_ser,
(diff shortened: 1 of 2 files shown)


superset/charts/data/api.py
```diff
@@ -394,8 +394,13 @@ def _process_data(query_data: Any) -> Any:
             )
 
         if result_format == ChartDataResultFormat.JSON:
+            queries = result["queries"]
+            if security_manager.is_guest_user():
+                for query in queries:
+                    with contextlib.suppress(KeyError):
+                        del query["query"]
```
betodealmeida, line 401: This works too, `contextlib.supress` seems like using a canon to kill a fly. :)

```suggestion
                    query.pop("query", None)
```

dpgaspar, line 401: nit, I think you could do:

``` python
if security_manager.is_guest_user():
    for query in queries:
        query.pop("query", None)
```

dpgaspar (approved):

eschutho (approved):

Issue #30183, opened 2024-09-07
docker bug


vyouzhis:
### Bug description

website img
![image](https://github.com/user-attachments/assets/d2f97b21-6112-4d4e-aaf1-ebf47d2280fb)

docker log -f 

`
werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
2024-09-07 08:36:18,303:INFO:werkzeug:192.168.2.100 - - [07/Sep/2024 08:36:18] "GET /static/assets/images/superset-logo-horiz.png HTTP/1.1" 404 -
2024-09-07 08:36:18,364:INFO:werkzeug:192.168.2.100 - - [07/Sep/2024 08:36:18] "GET /static/appbuilder/css/webfonts/fa-solid-900.woff2 HTTP/1.1" 200 -
2024-09-07 08:36:18,429:WARNING:superset.views.base:HTTPException
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 713, in <lambda>
    view_func=lambda **kw: self_ref().send_static_file(**kw),  # type: ignore # noqa: B950
  File "/usr/local/lib/python3.10/site-packages/flask/scaffold.py", line 331, in send_static_file
    return send_from_directory(
  File "/usr/local/lib/python3.10/site-packages/flask/helpers.py", line 590, in send_from_directory
    return werkzeug.utils.send_from_directory(  # type: ignore[return-value]
  File "/usr/local/lib/python3.10/site-packages/werkzeug/utils.py", line 574, in send_from_directory
    raise NotFound()
werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
2024-09-07 08:36:18,430:INFO:werkzeug:192.168.2.100 - - [07/Sep/2024 08:36:18] "GET /static/assets/images/favicon.png HTTP/1.1" 404 -
[...]

PR #30864, opened 2024-11-06
docs: adapt docs to suggest 'docker compose up --build'

mistercrunch:
In https://github.com/apache/superset/issues/30183, it appeared to me
that the issue might be around running `docker compose up` without
re-running `docker compose build`. Here I'm recommending running
`docker compose up --build` in the docs which should be safer.

### DIFF
--- a/docs/docs/installation/docker-compose.mdx
+++ b/docs/docs/installation/docker-compose.mdx
@@ -76,7 +76,8 @@ on latest base images using `docker compose build --pull`. In most cases though,
 ### Option #1 - for an interactive development environment
 
 ```bash
-docker compose up
+# The --build argument insures all the layers are up-to-date
+docker compose up --build
 ```
 
 :::tip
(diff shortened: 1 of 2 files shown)


docs/docs/installation/docker-compose.mdx
```diff
@@ -227,3 +228,11 @@ may want to find the exact hostname you want to use, for that you can do `ifconf
 Docker for you. Alternately if you don't even see the `docker0` interface try (if needed with sudo)
 `docker network inspect bridge` and see if there is an entry for `"Gateway"` and note the IP
 address.
+
+## 4. To build or not to build
+
+When running `docker compose up`, docker will build what is required behind the scene, but
+may use the docker cache if assets already exist. Running `docker compose build` prior to
+`docker compose up` or the equivalent shortcut `docker compose up --build` ensures that your
+docker images matche the definition in the repository. This should only apply to the main
```
eschutho, line 237:
```suggestion
docker images matches the definition in the repository. This should only apply to the main
```

eschutho (approved):

PR #30870, opened 2024-11-07
feat: make ephemeral env use supersetbot + deprecate build_docker.py

mistercrunch:
~~Someone reported that ephemeral env builds are failing. Example action run here https://github.com/apache/superset/actions/runs/11721810688/job/32650049015#step:7:1219~~

The issue above seems to be a glitch/one-time-occurence, yet this PR is still relevant.

In this PR:
- allow the ephemeral-env workflow to be testable using `workflow_dispatch`
- aligns docker build to use supersetbot like the main docker-related workflows
- removes now obsolete python script build_docker, replaced by `supersetbot docker` now, along with references in the docs

test command -> 

```
gh workflow run ephemeral-env.yml --ref fix_ephemerals --field comment_body="/testenv up" --field issue_number=30870
```


mistercrunch:
Just a note that the test above operated from the main branch, since the changes are on the GHA workflow itself.

mistercrunch:
BTW wasn't able to reproduce the issue, but this PR makes it easier to debug and aligns with using `supersetbot` to generate the docker command

geido:
I think the issue is looking a bit different now. Previously, it did not respond at all. Right now, the bot responds but the ephemeral shows a 500 server error. This is happen consistently in every PR I tried.

rusackas:
How does this handle cleanup? Just making sure that `test env down` works, that ephemerals are cleaned up on close/merge, and/or there's some sort of TTL on the environments.

I'm also wondering if/how concurrency is handled, as we don't want multiple/impatient triggers duplicating effort/resources.

mistercrunch:
I don't think I have the keys to inspect what's happening inside AWS. Will relabel the PR title to reflect that this PR isn't solving for the current issue with ephemeral envs. Maybe we can get someone closer to infra to take a look at what's happening.

dpgaspar:
@mistercrunch @geido 

Seeing this error on the logs: ModuleNotFoundError: No module named 'trino'

for example on this PR: https://github.com/apache/superset/pull/30917

mistercrunch:
> ModuleNotFoundError: No module named 'trino'

I'm guessing the "ci" layer targeted by ephemeral environment is building upon lean, as opposed to the `dev` layer, I added a commit so the "ci" layer builds upon dev.

mistercrunch:
Manually triggered here -> https://github.com/apache/superset/actions/runs/11958529982

mistercrunch:
I did this thing in `supsersetbot` where I'm running extra docker commands to build some layers sequentially instead of in parallel, and it seems to be helping with the out-of-memory SIGKILLs we've experienced in CI. I know the superset-node layer is memory-heavy, so it's much better to sequence the build. Also for now I'm making the `ci` layer build on top of dev, this should fix the issue raised by @dpgaspar . I think we should merge this.

eschutho (approved):

PR #30875, opened 2024-11-08
fix: don't show metadata for embedded dashboards

### LINKED ISSUE #30188: Superset 4.1.0RC2 - Dashboard embedding - now includes editor and last modified timestamp
### Bug description

The latest 4.1.0rc2 has in the dashboard header a username and last modified timestamp. This doesn't make sense for the embedded use case and should be hidden from the dashboard title. 

This should be hidden just like the Published tag as this is metadata that doesnt make sense outside of the superset app. 

Introduced in #27857

### How to reproduce the bug

1. Create a dashboard
2. Embed the dashboard in a page 
3. View embedded page with very superset specific metadata. 

### Screenshots/recordings

image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

sadpandajoe:
### SUMMARY

As part of https://github.com/apache/superset/pull/27857 we started to show metadata even on embedded dashboards. This could have a potential of displaying data that embedded don't need to see. This PR will remove it if we are in embedded view.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/30188
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/Header/index.jsx
+++ b/superset-frontend/src/dashboard/components/Header/index.jsx
@@ -17,44 +17,44 @@
  * under the License.
  */
 /* eslint-env browser */
-import moment from 'moment';
-import { PureComponent } from 'react';
-import PropTypes from 'prop-types';
-import {
-  styled,
-  css,
-  isFeatureEnabled,
-  FeatureFlag,
-  t,
-  getExtensionsRegistry,
-} from '@superset-ui/core';
 import { Global } from '@emotion/react';
 import {
-  LOG_ACTIONS_PERIODIC_RENDER_DASHBOARD,
-  LOG_ACTIONS_FORCE_REFRESH_DASHBOARD,
-  LOG_ACTIONS_TOGGLE_EDIT_DASHBOARD,
-} from 'src/logger/LogUtils';
-import Icons from 'src/components/Icons';
+  css,
+  FeatureFlag,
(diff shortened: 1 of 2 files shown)


superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -496,6 +496,7 @@ class Header extends PureComponent {
     const refreshWarning =
       dashboardInfo.common?.conf
         ?.SUPERSET_DASHBOARD_PERIODICAL_REFRESH_WARNING_MESSAGE;
+    const isEmbedded = !dashboardInfo?.userId;
```
mistercrunch, line 499: [non blocking] curious whether there's a more explicit way to check whether we're in an embedded context?

rusackas: I don't think so, but this comes up often... we need to invent this kind of state/check for myriad reasons.

Vitor-Avila (approved):
LGTM!

superset-frontend/src/dashboard/components/Header/index.jsx
Vitor-Avila: I agree it's not ideal, but this is consistent with other places we assess if it's an embedded user or not.

PR #30887, opened 2024-11-10
fix(imports): import query_context for imports with charts

lindenh:
### SUMMARY
Previously, the query_context was removed on imports because of the necessary changes to datasource table ids on import to align with already-existing datasources even if the id itself was different (as it may be across environments). This became a problem in https://github.com/apache/superset/issues/28446 as the query context was now required for embedded charts. 

This normally isn't an issue, as the query context is generated when saving the chart on the frontend, but if charts are uploaded, this context is removed. This re-adds that imported context with the same changes made to the query_context as made to the imported charts themselves to ensure they point at the right data source.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/28446
- [ ] Required feature flags:
- [ ] Changes UI


lindenh:
Think you could give this a pat down @betodealmeida ?

betodealmeida:
Ah, let me take a look!

betodealmeida (approved):
This is awesome, thanks for fixing it!

PR #30694, opened 2024-10-23
feat: allow exporting all tabs to a single PDF in report

US579:
### SUMMARY

This update will give user an option to export all tabs if there are more then one in a dashboard, all tabs will be included in a single PDF document

ANIMATED GIF

output10

### BEFORE/AFTER SCREENSHOTS

Before user could only include one tab into a PDF report

Screenshot 2024-10-24 at 10 54 02 am

example output: 

[1_ FCC New Coder Survey 2018 (1).pdf](https://github.com/user-attachments/files/17499190/1_.FCC.New.Coder.Survey.2018.1.pdf)

After user could choose to include all tabs to their report

Screenshot 2024-10-24 at 10 53 04 am

example output: 
 [1_ FCC New Coder Survey 2018.pdf](https://github.com/user-attachments/files/17499180/1_.FCC.New.Coder.Survey.2018.pdf)

NOTE: This feature can also be modified to include specific tabs in the email, but it would require changes on the frontend (allow users to select multiple tabs).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: this need to enable feature flag: ALERT_REPORT_TABS
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #30888, opened 2024-11-10
fix: keep the tab order

US579:
### SUMMARY

When export all tabs to a PDF in report & alert, user noticed that the all tabs export doesn't seem to export the tabs in any specific order (our report with 3 tabs exported with tab 2, then 3, then 1 in the PDF)

reference: https://github.com/apache/superset/pull/30694#issuecomment-2455854080

### Context

The exported tabs are now ordered based on the sequence of anchors under the all_tabs field from the tabs endpoint response. This enhancement aligns the export order with the order displayed on the dashboard.

### BEFORE PDF

[before.pdf](https://github.com/user-attachments/files/17691280/before.pdf)

### AFTER PDF

[after.pdf](https://github.com/user-attachments/files/17691281/after.pdf)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [X] Required feature flags:
- [ ] Changes UI


US579:
FYI @yousoph

rusackas:
LGTM, but @eschutho might be better poised to actually _test_ this before approving/merging.

superset/dashboards/api.py
```diff
@@ -479,7 +480,10 @@ def get_tabs(self, id_or_slug: str) -> Response:
         try:
             tabs = DashboardDAO.get_tabs_for_dashboard(id_or_slug)
             result = self.tab_schema.dump(tabs)
-            return self.response(200, result=result)
+            response_data = json.dumps({"result": result})
+            return Response(
+                response_data, status=HTTPStatus.OK, mimetype="application/json"
+            )
```
eschutho, line 486: @US579 was the self.response not including the correct mimetype? I would think that all the responses should be json, but maybe @dpgaspar has more info on this.

US579: yep it does has mimetype, but it will reorder the dict

This is a test example

Screenshot 2025-01-08 at 9 49 44 pm


Screenshot 2025-01-08 at 9 48 59 pm

superset/models/dashboard.py
```diff
@@ -338,12 +338,16 @@ def build_tab_tree(
...
         queue: deque[tuple[dict[str, Any], list[dict[str, Any]]]] = deque()
         queue.append((root, tab_tree))
         while queue:
             node, children = queue.popleft()
             build_tab_tree(node, children)
-
+            if (childs := node.get("children")) and node["type"] == "TABS":
+                tabs_order.extend(childs)
+        if tabs_order:
+            all_tabs = {str(t): all_tabs[str(t)] for t in tabs_order}
```
eschutho, line 350: @US579 Since the api (in addition to the dictionary for all_tabs) returns a nested array structure of the tabs which would maintain their order, can you iterate through the tab_tree instead of relying on the all_tabs dict? 
Screenshot 2025-01-27 at 5 34 32 PM

US579: ah yep good found!

superset-frontend/src/features/alerts/AlertReportModal.tsx
```diff
@@ -826,11 +826,19 @@ const AlertReportModal: FunctionComponent<AlertReportModalProps> = ({
...
+          );
+
+          // Only show all tabs when there are more than one tab
+          if (allTabsWithOrder.length > 1) {
+            tabTree.push({
+              title: 'All Tabs',
+              // select tree only works with string value
+              value: JSON.stringify(allTabsWithOrder),
+            });
+          }
```
eschutho, line 840: @US579 will this change affect the pdf generation as well or just the dropdown ui in the modal?

US579: this will not affect pdf generation, just dropdown UI
when there is only one tab, the all tab option will not be showed up

US579: if we add all tab it will just be a list with one anchor
eg.
['anchor']

so it is same as setting one tab directly

eschutho (approved):
Thanks @US579!

US579:
@eschutho pipeline fixed, could you merge it?

PR #30903, opened 2024-11-12
fix(virtual dataset sync): Sync virtual dataset columns when changing the SQL query

fisjac:
### SUMMARY

When changing the SQL query for virtual datasets, the user first needs to save the virtual dataset, then needs to reopen the "edit dataset" modal and manually sync the columns. This PR will introduce changes to automatically sync the columns upon saving a virtual dataset with a new SQL query.

https://www.loom.com/share/b9a1121ec44b47438e2675334403ee79
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/explore/control.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/explore/control.test.ts
@@ -51,8 +51,8 @@ describe('Datasource control', () => {
     )
       .first()
       .focus();
-    cy.focused().clear();
-    cy.focused().type(`${newMetricName}{enter}`);
+    cy.focused().clear({ force: true });
+    cy.focused().type(`${newMetricName}{enter}`, { force: true });
 
     cy.get('[data-test="datasource-modal-save"]').click();
     cy.get('.antd5-modal-confirm-btns button').contains('OK').click();
--- a/superset-frontend/packages/superset-ui-chart-controls/src/types.ts
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/types.ts
@@ -84,6 +84,12 @@ export interface Dataset {
   filter_select?: boolean;
   filter_select_enabled?: boolean;
   column_names?: string[];
+  catalog?: string;
+  schema?: string;
+  table_name?: string;
+  database?: Record<string, unknown>;
+  normalize_columns?: boolean;
+  always_filter_main_dttm?: boolean;
 }
 
 export interface ControlPanelState {
--- a/superset-frontend/src/components/Datasource/DatasourceEditor.jsx
+++ b/superset-frontend/src/components/Datasource/DatasourceEditor.jsx
@@ -23,7 +23,6 @@ import { Radio } from 'src/components/Radio';
 import Card from 'src/components/Card';
 import Alert from 'src/components/Alert';
 import Badge from 'src/components/Badge';
-import { nanoid } from 'nanoid';
 import {
   css,
   isFeatureEnabled,
@@ -57,6 +56,7 @@ import CurrencyControl from 'src/explore/components/controls/CurrencyControl';
 import CollectionTable from './CollectionTable';
 import Fieldset from './Fieldset';
 import Field from './Field';
+import { fetchSyncedColumns, updateColumns } from './utils';
 
 const DatasourceContainer = styled.div`
   .change-warning {
@@ -139,6 +139,14 @@ const StyledButtonWrapper = styled.span`
     margin-left: ${theme.gridUnit * 3}px;
   `}
 `;
+
+const sqlTooltipOptions = {
+  placement: 'topRight',
+  title: t(
+    'If changes are made to your SQL query, ' +
+      'columns in your dataset will be synced when saving the dataset.',
+  ),
+};
 
 const checkboxGenerator = (d, onChange) => (
   <CheckboxControl value={d} onChange={onChange} />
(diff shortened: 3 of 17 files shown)


villebro:
This is a total pain, super happy to see this potentially get fixed soon! ❤️

superset-frontend/src/components/Datasource/DatasourceEditor.jsx
```diff
@@ -1146,6 +1057,12 @@ class DatasourceEditor extends PureComponent {
                         maxLines={Infinity}
                         readOnly={!this.state.isEditMode}
                         resize="both"
+                        tooltipOptions={{
+                          title:
+                            'If changes are made to your SQL query, ' +
```
geido, line 1062: Let's localize this

superset-frontend/src/components/Datasource/DatasourceModal.tsx
```diff
@@ -110,124 +122,139 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
...
+          );
+          currentDatasource.columns = columnChanges.finalColumns;
+          dispatch(syncDatasourceMetadata(currentDatasource));
+          dispatch(stopMetaDataLoading());
+          addSuccessToast(t('Metadata has been synced'));
+        } catch (error) {
+          dispatch(stopMetaDataLoading());
+          const { error: clientError, statusText } =
+            await getClientErrorObject(error);
+          addDangerToast(
```
geido, line 214: Do we need to print the actual error to the user? Can we control the error for a more user friendly output?

superset-frontend/packages/superset-ui-chart-controls/src/types.ts
```diff
@@ -84,6 +84,12 @@ export interface Dataset {
   filter_select?: boolean;
   filter_select_enabled?: boolean;
   column_names?: string[];
+  catalog?: string;
+  schema?: string;
+  table_name?: string;
+  database?: Record<string, unknown>;
```
kgabryje, line 90: Could we improve the typing here? We probably have a type for database defined somewhere

superset-frontend/src/components/Datasource/DatasourceModal.tsx
```diff
@@ -110,124 +122,139 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
...
+    offset: datasource.offset,
+    default_endpoint: datasource.default_endpoint,
+    cache_timeout:
+      datasource.cache_timeout === '' ? null : datasource.cache_timeout,
+    is_sqllab_view: datasource.is_sqllab_view,
+    template_params: datasource.template_params,
+    extra: datasource.extra,
+    is_managed_externally: datasource.is_managed_externally,
+    external_url: datasource.external_url,
+    metrics: datasource?.metrics?.map((metric: Record<string, unknown>) => {
```
kgabryje, line 148: can we cha
```suggestion
    metrics: datasource?.metrics?.map((metric: Metric) => {
```

superset-frontend/src/components/Datasource/DatasourceModal.tsx
```diff
@@ -110,124 +122,139 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
...
+        warning_text: metric.warning_text,
+        uuid: metric.uuid,
+        extra: buildExtraJsonObject(metric),
+      };
+      if (!Number.isNaN(Number(metric.id))) {
+        metricBody.id = metric.id;
+      }
+      return metricBody;
+    }),
+    columns: datasource?.columns?.map((column: Record<string, unknown>) => ({
```
kgabryje, line 168:
```suggestion
    columns: datasource?.columns?.map((column: Column) => ({
```
(or `ColumnMeta`)

kgabryje:
Small typing suggestions, other than that the code looks great 👍

superset-frontend/src/components/Datasource/DatasourceModal.tsx
fisjac: There are some issues with doing this due to a part of the code that checks for an `is_certified` property on somwhat ambiguous `Metric` `Column` and `Data` items. I agree that these types should be tightened up, but I don't have enough context to start tweaking these in this PR.

dkrat7:
Hi,
syncing columns without saving and reopening the editor is great improvement.

However, i have some concerns regarding autosyncing on save:

What will happen if I accidently make a  change to a query which will result in zero rows returned - all columns including metadata will be permanently removed?

Sometimes I include Jinja code in a query with condition that dataset returns zero rows if some filter is not selected (for example) - I won't be able to save that query without losing columns?

fisjac:
@dkrat7 to my knowledge, columns are not dropped based on whether row level data are available for that column, but based upon whether the column is selected as part of the query. In your case, the columns will be dropped until the query is edited to include the desired columns.

Point being, if your query would include (or exclude) a column, then the virtual dataset will be synced to reflect the state of that query upon save.

betodealmeida:
> What will happen if I accidently make a change to a query which will result in zero rows returned - all columns including metadata will be permanently removed?

It should be fine, since the cursor should still return a description of the columns even if the number of rows is 0. For example, for Postgres:

```python
import psycopg2

query = "SELECT * FROM your_table WHERE 1=0;"  # Query guaranteed to return zero rows

db_config = {
    "dbname": "DB",
    "user": "user",
    "password": "password",
    "host": "host",
    "port": 5432,
}

with psycopg2.connect(**db_config) as conn:
    with conn.cursor() as cursor:
        cursor.execute(query)

        print("Cursor description:")
        for col in cursor.description:
            print(
                f"Name: {col.name}, Type: {col.type_code}, Display Size: {col.display_size}"
            )
```

For a test table this prints:

```
Cursor description:
Name: show_id, Type: 25, Display Size: None
Name: type, Type: 25, Display Size: None
Name: title, Type: 25, Display Size: None
Name: director, Type: 25, Display Size: None
Name: country, Type: 25, Display Size: None
Name: date_added, Type: 25, Display Size: None
Name: release_year, Type: 23, Display Size: None
Name: rating, Type: 25, Display Size: None
Name: duration, Type: 25, Display Size: None
Name: listed_in, Type: 25, Display Size: None
Name: description, Type: 25, Display Size: None
Name: iso_date_added, Type: 25, Display Size: None
Name: iso_release_year, Type: 25, Display Size: None
Name: xata_version, Type: 23, Display Size: None
Name: xata_createdat, Type: 1184, Display Size: None
Name: xata_updatedat, Type: 1184, Display Size: None
Name: xata_id, Type: 25, Display Size: None
```

superset-frontend/cypress-base/cypress/e2e/explore/control.test.ts
```diff
@@ -1,276 +0,0 @@
...
- * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-// ***********************************************
-// Tests for setting controls in the UI
-// ***********************************************
-import { interceptChart } from 'cypress/utils';
```
geido, line 22: Is it intentional that we are deleting these E2E tests?

superset-frontend/packages/superset-ui-chart-controls/src/types.ts
geido: @fisjac +1 for this

superset-frontend/src/components/Datasource/DatasourceModal.tsx
geido: @fisjac should not be too hard to find out what the right type is? Maybe we need a partial type or to update the original type if a property appears to be missing. It defeats the purpose of using typescript if we opt for unknown types

superset-frontend/src/components/Datasource/utils.js
```diff
@@ -40,3 +43,86 @@ export function recurseReactClone(children, type, propExtender) {
...
+        type: col.type,
+        is_dttm: currentCol.is_dttm || col.is_dttm,
+      });
+      columnChanges.modified.push(col.column_name);
+    } else {
+      // unchanged
+      columnChanges.finalColumns.push(currentCol);
+    }
+  });
+  if (columnChanges.modified.length) {
```
geido, line 92: I am wondering if this is an opportunity to improve the user experience here and just show a generic message. From your sample video it looks really bad and possibly useless for the user to see such big success toasts with a bunch of columns in them. cc @kasiazjc

superset-frontend/src/explore/components/controls/TextAreaControl.jsx
```diff
@@ -94,31 +100,44 @@ class TextAreaControl extends Component {
       if (this.props.readOnly) {
         style.backgroundColor = '#f2f2f2';
       }
+      const codeEditor = (
+        <div>
```
geido, line 104: What is this `div` for?

superset-frontend/src/explore/components/controls/TextAreaControl.jsx
```diff
@@ -94,31 +100,44 @@ class TextAreaControl extends Component {
...
 
-      return (
-        <TextAreaEditor
-          mode={this.props.language}
-          style={style}
-          minLines={minLines}
-          maxLines={inModal ? 1000 : this.props.maxLines}
-          editorProps={{ $blockScrolling: true }}
+    const textArea = (
+      <div>
```
geido, line 127: Same for this `div`, what is this for?

geido (changes requested):
Thanks @fisjac. I am requesting changes as I am not sure we should delete the E2E tests. Also, some types need more love. Product feedback might also help to discuss the toasts. I feel, in general, this needs a bit more refinement.

superset-frontend/src/explore/components/controls/TextAreaControl.jsx
fisjac: When trying to wrap a tooltip around the `TextAreaEditor`  without a div, the tooltip doesn't land in the right place. Placing a div enables the tooltip to behave as intended.

superset-frontend/src/explore/components/controls/TextAreaControl.jsx
fisjac: same answer as above

superset-frontend/src/components/Datasource/utils.js
fisjac: agreed that showing each of the columns changed winds up being a bit of an overkill from a user standpoint. Should we just do away with the toasts altogether in favor of something like "columns updated"?

geido (approved):

michael-s-molina:
@fisjac @geido @kgabryje Virtual dataset queries are generally heavy and may take seconds to execute. Given that we don't check if columns/metrics were changed but just check if the SQL changed, there's a huge performance downgrade when users are editing the SQL to change something that does not involve changing metric/columns like adding comments, changing a where condition or join. These types of changes which were really fast in 4.1, now trigger the virtual dataset query execution and takes much more time to complete. We're talking about a 1 second to 45 seconds difference. Reverting this PR might be complicated due to the multiple refactors applied. I'm wondering if it would be possible to ask the user if they wish to sync when the SQL changes. If that's possible and the user chooses to not sync, would something break?

PR #30935, opened 2024-11-15
fix to avoid replacing charts with the Loading logo when charts are n…

### LINKED ISSUE #27532: Export to PDF leaves charts empty on first export
### Bug description

When using the new Export to PDF functionality, we get a mostly empty PDF on the first try. Second export usually works fine. If we refresh the page - same issues reappear: first export is empty, second is fine. 

The issue does not seem to impact Big Number and Table charts. But everything else: Pie Charts, Bar Charts, Area Charts, seem to be impacted.

### How to reproduce the bug

1. Open a dashboard which has a Pie Chart.
2. Click Export to PDF
3. Check PDF

### Screenshots/recordings

First click on Export to PDF:
Screenshot 2024-03-15 at 14 50 04  

Second click on Export to PDF:
Screenshot 2024-03-15 at 14 50 30

### Superset version

3.1.1

### Python version

3.10

### Node version

Not applicable

### Browser

Safari

### Additional context

_No response_

### Checklist

jpchev:
this is to fix issue 27532

### SUMMARY
by setting to True the following feature flag in config.py

`"DASHBOARD_DISABLE_LOADING": True`

one can avoid having charts replaced by the login icon when charts are not in the current view.
This is useful to export all the charts of a dashboard to a pdf file.

### TESTING INSTRUCTIONS
after having set the feature flag DASHBOARD_DISABLE_LOADING, see in the DOM that charts don't get replaced by the "loading" img upon scrolling the page.

Fixes #27532

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
+++ b/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
@@ -33,6 +33,7 @@ export enum FeatureFlag {
   /** @deprecated */
   DashboardCrossFilters = 'DASHBOARD_CROSS_FILTERS',
   DashboardVirtualization = 'DASHBOARD_VIRTUALIZATION',
+  DashboardDisableLoading = 'DASHBOARD_DISABLE_LOADING',
   DashboardRbac = 'DASHBOARD_RBAC',
   DatapanelClosedByDefault = 'DATAPANEL_CLOSED_BY_DEFAULT',
   DisableLegacyDatasourceEditor = 'DISABLE_LEGACY_DATASOURCE_EDITOR',
--- a/superset-frontend/src/components/Chart/Chart.tsx
+++ b/superset-frontend/src/components/Chart/Chart.tsx
@@ -301,7 +301,7 @@ class Chart extends PureComponent<ChartProps, {}> {
         {this.props.isInView ||
         !isFeatureEnabled(FeatureFlag.DashboardVirtualization) ||
         isCurrentUserBot() || 
-        isFeatureEnabled(FeatureFlag.DashboardDisableLoading)? (
+        isFeatureEnabled(FeatureFlag.DashboardDisableLoading) ? (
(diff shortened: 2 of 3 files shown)


jpchev:
@kgabryje ok, do you mean that setting ` ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT = True ` would be enough to have the pdf generated on the backend?

I thought that is only for scheduled dashboards

kgabryje:
> @kgabryje ok, do you mean that setting `ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT = True` would be enough to have the pdf generated on the backend?
> 
> I thought that is only for scheduled dashboards

No, that feature flag enables downloading dashboard as an image through Selenium, not client-side. I meant that we should do a similar thing for downloading as PDF to fix the problem with loading charts

jpchev:
I'm testing with `DASHBOARD_VIRTUALIZATION = false`, does't seem to help in embedded dashboards. Do you think it's available in version 4.0.1 ? 

BTW, for the generation on the backend, is it enough to set `ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT = True`
or does that go together with another setup?

Thank you

kgabryje:
TBH I don't really know the details of that feature, but perhaps @geido knows more?

geido:
> I'm testing with `DASHBOARD_VIRTUALIZATION = false`, does't seem to help in embedded dashboards. Do you think it's available in version 4.0.1 ?
> 
> BTW, for the generation on the backend, is it enough to set `ENABLE_DASHBOARD_DOWNLOAD_WEBDRIVER_SCREENSHOT = True` or does that go together with another setup?
> 
> Thank you

You also need `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS`

jpchev:
closing this PR as not relevant anymore

PR #30949, opened 2024-11-16
fix: adjust line type as well as weight for time series

eschutho:
### SUMMARY

The line thicknesses were with time adjustments were making the visualization difficult to read. Instead, I kept the existing thickness to match the legacy line chart and adjusted the dashed pattern instead. 

I also removed the line boldness on hover because on dashed lines the animation is a bit distracting. 

Before: 
image
image

After: 
Screenshot 2025-01-17 at 4 35 39 PM

Screenshot 2025-01-17 at 4 35 07 PM

**Hover animation:** 
Before:

https://github.com/user-attachments/assets/3eccc254-5c7f-406a-b8b7-e59c05c5e1f4

After:

https://github.com/user-attachments/assets/d7eea24c-9aed-48ef-bcc3-eada66612fe6

### TESTING INSTRUCTIONS

Create a line chart with a time comparison and add more than 5 time comparisons. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -281,7 +281,7 @@ export default function transformProps(
   const array = ensureIsArray(chartProps.rawFormData?.time_compare);
   const inverted = invert(verboseMap);
 
-  const offsetLineWidths: { [key: string]: number } = {};
+  const offsetLineWidths: string[] = [];
 
   rawSeries.forEach(entry => {
     const derivedSeries = isDerivedSeries(entry, chartProps.rawFormData);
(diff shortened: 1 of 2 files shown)


rusackas (approved):
LGTM!

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
```diff
@@ -314,15 +314,7 @@ export function transformSeries(
             opacity: opacity * areaOpacity,
           }
         : undefined,
-    emphasis: {
-      // bold on hover as required since 5.3.0 to retain backwards feature parity:
-      // https://apache.github.io/echarts-handbook/en/basics/release-note/5-3-0/#removing-the-default-bolding-emphasis-effect-in-the-line-chart
-      // TODO: should consider only adding emphasis to currently hovered series
-      lineStyle: {
-        width: 'bolder',
```
eschutho, line 322: @villebro is it ok to remove this?

eschutho:
cc @michael-s-molina

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
michael-s-molina: > I also removed the line boldness on hover because on dashed lines the animation is a bit distracting.

@eschutho I believe this was added to keep the hover behavior consistent with other chart types.

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -288,11 +288,20 @@ export default function transformProps(
         entry,
         ensureIsArray(chartProps.rawFormData?.time_compare),
       )!;
-      if (!offsetLineWidths[offset]) {
-        offsetLineWidths[offset] = Object.keys(offsetLineWidths).length + 1;
+      if (!offsetLineWidths.includes(offset)) {
```
michael-s-molina, line 291: The map approach is more efficient than looping through the elements with `include`.

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
eschutho: Is it ok to remove it? It looks like the echarts community didn't like the look of it on the line chart, so they removed it. It just seems more distracting than anything on the line chart imo. wdyt?

kgabryje (approved):

PR #31008, opened 2024-11-21
fix(explore): verified props is not updated

justinpark:
### SUMMARY
A bug occurred because the verifiedProps are modified by comparing the verify results with the existing otherProps, causing the verify results to be retained as the previous verify results when the original list resets to its original state.

This commit has been made to address this issue by ensuring that updates are made in comparison with the verified props.

Before:

Screenshot 2024-11-20 at 5 13 19 PM

After:

Screenshot 2024-11-20 at 5 13 43 PM

### TESTING INSTRUCTIONS
locally

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/withAsyncVerification.tsx
+++ b/superset-frontend/src/explore/components/controls/withAsyncVerification.tsx
@@ -142,12 +142,14 @@ export default function withAsyncVerification({
     } = props;
     const otherPropsRef = useRef(restProps);
     const [verifiedProps, setVerifiedProps] = useState({});
+    const verifiedPropsRef = useRef(verifiedProps);
     const [isLoading, setIsLoading] = useState<boolean>(initialIsLoading);
     const { addWarningToast } = restProps.actions;
 
     // memoize `restProps`, so that verification only triggers when material
     // props are actually updated.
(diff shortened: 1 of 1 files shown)


superset-frontend/src/explore/components/controls/withAsyncVerification.tsx
```diff
@@ -142,12 +142,14 @@ export default function withAsyncVerification({
     } = props;
     const otherPropsRef = useRef(restProps);
     const [verifiedProps, setVerifiedProps] = useState({});
+    const verifiedPropsRef = useRef(verifiedProps);
```
michael-s-molina, line 145: Can't we remove `verifiedProps` state and just use `verifiedPropsRef`?

justinpark: `verifiedProps` is necessary for rendering a new options list through `setVerifiedProps`.
I have an idea that this part can be handled using `useEffectEvent` without utilizing additional ref access.

michael-s-molina (approved):

PR #31029, opened 2024-11-22
chore: Refactor dashboard header to func component

kgabryje:
### SUMMARY
Refactor the dashboard Header component from class to function component. No functional changes, just a refactor

### TESTING INSTRUCTIONS
Everything should work like before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -170,209 +146,253 @@ const discardBtnStyle = theme => css`
...
+          t(
+            `This dashboard is currently auto refreshing; the next auto refresh will be in %s.`,
+            intervalMessage,
+          ),
+        );
+        if (
+          dashboardInfo.common?.conf?.DASHBOARD_AUTO_REFRESH_MODE === 'fetch'
+        ) {
+          // force-refresh while auto-refresh in dashboard
+          return boundActionCreators.fetchCharts(
```
Antonio-RiveroMartnez, line 287: nit: I know it was like this before this PR, but we could DRY it with a helper function so you can pass the boolean as arg and return the `fetchCharts` call with it here and the one below

superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -383,58 +403,79 @@ class Header extends PureComponent {
...
+      boundActionCreators.addDangerToast(
         t(
           'Your dashboard is too large. Please reduce its size before saving it.',
         ),
       );
     } else {
       if (positionJSONLength >= limit * 0.9) {
-        this.props.addWarningToast('Your dashboard is near the size limit.');
+        boundActionCreators.addWarningToast(
+          'Your dashboard is near the size limit.',
```
Antonio-RiveroMartnez, line 435: Should we translate this?

superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -170,209 +146,253 @@ const discardBtnStyle = theme => css`
...
+    [dispatch],
+  );
+
+  const startPeriodicRender = useCallback(
+    interval => {
+      let intervalMessage;
+
+      if (interval) {
+        const periodicRefreshOptions =
+          dashboardInfo.common?.conf?.DASHBOARD_AUTO_REFRESH_INTERVALS;
```
Antonio-RiveroMartnez, line 254: nit: Is this value dynamically changing somewhere or just a static config loaded once? If static, we could move this out the callback.

superset-frontend/src/dashboard/components/Header/index.jsx
kgabryje: nice catch!

superset-frontend/src/dashboard/components/Header/index.jsx
kgabryje: it doesn't change, but _technically_ it could since it's a part of the redux state object, so it should be included in the dependency array - it won't hurt and it won't trigger a lint warning

Antonio-RiveroMartnez (approved):
lgtm

michael-s-molina:
Maybe it would be nice to use `refactor` instead of `chore` for this type of PR.

kgabryje:
> Maybe it would be nice to use `refactor` instead of `chore` for this type of PR.

good point 🤦 will keep that in mind next time

PR #31031, opened 2024-11-22
fix(Dashboard): Ensure shared label colors are updated

geido:
### SUMMARY
Follow-up PR for [#30646](https://github.com/apache/superset/pull/30646) to ensure that when certain dimensions or metrics are removed or no longer present in the charts, the stored `shared_label_colors` can detect these changes and reflect the current state. This is achieved by comparing all labels across all charts to infer the shared ones once a Dashboard has rendered all its charts.

N.A.

### TESTING INSTRUCTIONS
1. Enter an existing Dashboard
2. Update a chart that has shared labels with other charts
3. Remove those dimensions/metrics from the chart
4. Ensure that when going back to the dashboard the dashboard is being updated with current shared label colors (by checking the network activity)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/actions/dashboardState.js
+++ b/superset-frontend/src/dashboard/actions/dashboardState.js
@@ -28,6 +28,7 @@ import {
   t,
   getClientErrorObject,
   getCategoricalSchemeRegistry,
+  promiseTimeout,
 } from '@superset-ui/core';
 import {
   addChart,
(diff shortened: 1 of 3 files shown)


michael-s-molina:
Isn't this a fix instead of a chore?

fisjac (approved):
Looks good to me

geido:
> Isn't this a fix instead of a chore?

This wasn't there before so it's not fixing something rather adding. But I agree that fix might be appropriate.

PR #31141, opened 2024-11-25
feat: add YDB as a new database engine

vgvoleg:
### SUMMARY

This PR provides [YDB](https://ydb.tech/) database support based on [YDB Dialect](https://github.com/ydb-platform/ydb-sqlalchemy)

### TESTING INSTRUCTIONS

Steps needed to install sqlalchemy dialect described on databases.mdx.

To test it you can start local YDB via docker-compose:
```
version: "3.3"
services:
  ydb:
    image: ydbplatform/local-ydb:trunk
    restart: always
    ports:
      - 2136:2136
    hostname: localhost
    environment:
      - YDB_USE_IN_MEMORY_PDISKS=true
      - YDB_ENABLE_COLUMN_TABLES=true
```

sqlalchemy uri will be like
```
ydb://localhost:2136/local
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


docs/docs/configuration/databases.mdx
```diff
@@ -1524,6 +1525,78 @@ Other parameters:
...
+```
+{
+    "protocol": "grpcs"
+}
+```
+
+Default is `grpc`.
+
+
+##### Authenticaions
```
giftig, line 1554: Typo here

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,94 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from __future__ import annotations
+import json
+import logging
+from typing import (
+    Any,
+    TYPE_CHECKING,
+)
```
giftig, line 23: Did you install and run the pre-commit hook? I think that'd stick these on one line.

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,94 @@
...
+
+    sqlalchemy_uri_placeholder = (
+        "ydb://{host}:{port}/{database_name}"
+    )
+
+    encrypted_extra_sensitive_fields = {"$.connect_args.credentials", "$.credentials"}
+
+    supports_file_upload = False
+
+    _time_grain_expressions = {
```
giftig, line 50: Did you find this was necessary to make it work? I would have thought the base spec's empty dict would be ok here if you're not configuring any custom time grain expressions, and several other specs aren't setting it. But I do see this `None: "{col}"` entry in other specs which do set them so maybe I'm missing something which needs this.

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,94 @@
...
+            connect_args = params.setdefault("connect_args", {})
+
+            if "protocol" in encrypted_extra:
+                connect_args["protocol"] = encrypted_extra.pop("protocol", "grpc")
+
+            if "credentials" in encrypted_extra:
+                credentials_info = encrypted_extra.pop("credentials", {})
+                credentials = None
+                if "username" in credentials_info:
+                    from ydb import StaticCredentials
```
giftig, line 73: A bit simpler just to collect these imports at the top of the method rather than place them individually in various if/else branches; you can't put them at the top of the file because of the driver being an optional dependency, but I don't think you need to go this extreme.

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,94 @@
...
+    ) -> None:
+        if not database.encrypted_extra:
+            return
+
+        try:
+            encrypted_extra = json.loads(database.encrypted_extra)
+            connect_args = params.setdefault("connect_args", {})
+
+            if "protocol" in encrypted_extra:
+                connect_args["protocol"] = encrypted_extra.pop("protocol", "grpc")
```
giftig, line 67: I'd suggest avoiding mutating `encrypted_extra` in case the mutated data gets saved back to the database accidentally at some point. Just use `.get()` instead.

Also that `"grpc"` string will never be used since you've checked the existence of the key first, you probably just meant to do

```python
connect_args["protocol"] = encryped_extra.pop("protocol", "grpc")
```
in all cases

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,94 @@
...
+                        credentials_info.pop("password", None),
+                    )
+                elif "token" in credentials_info:
+                    from ydb import AccessTokenCredentials
+                    credentials = AccessTokenCredentials(
+                        credentials_info.pop("token", None)
+                    )
+                elif "service_account_json" in credentials_info:
+                    from ydb.iam import ServiceAccountCredentials
+                    sa_json = credentials_info.pop("service_account_json", {})
```
giftig, line 85: Default args would never be used here or on L81 for same reason as above; you can just index it since you've checked.

giftig (changes requested):

giftig:
Also the logic in your spec should have some unit tests.

giftig:
Can you also provide some test instructions for how to verify this is working / describe what you tested, please? I'm assuming it's mostly just install the ydb driver, connect a ydb database, and try running some queries and building some dashboards, but maybe others will catch some edge case testing which may be useful with a new engine spec.

docs/docs/configuration/databases.mdx
giftig: Actually it ought to say "authentication methods" or something similar really since authentication itself can't be pluralised.

vgvoleg: fixed

superset/db_engine_specs/ydb.py
vgvoleg: fixed

superset/db_engine_specs/ydb.py
vgvoleg: It is needed for one test to pass (it checks that len of this dict is greater than 0)
I had an issue with generated query after filling this dict in correct for YDB way https://apache-superset.slack.com/archives/C014LS99C1K/p1732544528541239

TLDR we can't work with generated queries because YDB expecting something like

```
SELECT DateTime::StartOf(date, Interval('P1D')) AS date, count(client_id) AS `COUNT(client_id)` 
FROM (SELECT * from transactions_data
) AS virtual_table GROUP BY date ORDER BY `COUNT(client_id)` DESC
 LIMIT CAST(1000 AS UInt64);
```

instead of

```
SELECT DateTime::StartOf(date, Interval('P1D')) AS date, count(client_id) AS `COUNT(client_id)` 
FROM (SELECT * from transactions_data
) AS virtual_table GROUP BY DateTime::StartOf(date, Interval('P1D')) AS date ORDER BY `COUNT(client_id)` DESC
 LIMIT CAST(1000 AS UInt64);
```

Let me know if I can override something to make group by use aliases as well as order by.

superset/db_engine_specs/ydb.py
vgvoleg: ServiceAccountCredentials will need additional dependency (I've updated docs), so I can't just import it for every YDB usage. It feels weird to have 2 imports moved up, but 1 not, so I left every import under if's

superset/db_engine_specs/ydb.py
vgvoleg: fixed

superset/db_engine_specs/ydb.py
vgvoleg: fixed

superset/db_engine_specs/ydb.py
vgvoleg: nevermind :)

superset/db_engine_specs/ydb.py
vgvoleg: moved it out of this PR because it is already handled on DBAPI level

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,108 @@
...
+        TimeGrain.TEN_MINUTES: "DateTime::MakeDatetime(DateTime::StartOf({col}, Interval('PT10M')))",
+        TimeGrain.FIFTEEN_MINUTES: "DateTime::MakeDatetime(DateTime::StartOf({col}, Interval('PT15M')))",
+        TimeGrain.THIRTY_MINUTES: "DateTime::MakeDatetime(DateTime::StartOf({col}, Interval('PT30M')))",
+        TimeGrain.HOUR: "DateTime::MakeDatetime(DateTime::StartOf({col}, Interval('PT1H')))",
+        TimeGrain.DAY: "DateTime::MakeDatetime(DateTime::StartOf({col}, Interval('P1D')))",
+        TimeGrain.WEEK: "DateTime::MakeDatetime(DateTime::StartOfWeek({col}))",
+        TimeGrain.MONTH: "DateTime::MakeDatetime(DateTime::StartOfMonth({col}))",
+        TimeGrain.QUARTER: "DateTime::MakeDatetime(DateTime::StartOfQuarter({col}))",
+        TimeGrain.YEAR: "DateTime::MakeDatetime(DateTime::StartOfYear({col}))",
+    }
```
giftig, line 69: The pedant in me hates the fact that YDB apparently capitalises datetime two different ways in that function :(

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,108 @@
...
+
+        try:
+            encrypted_extra = json.loads(database.encrypted_extra)
+            connect_args = params.setdefault("connect_args", {})
+
+            if "protocol" in encrypted_extra:
+                connect_args["protocol"] = encrypted_extra["protocol"]
+
+            if "credentials" in encrypted_extra:
+                credentials_info = encrypted_extra["credentials"]
```
giftig, line 103: This argument seems unnecessary, could just be one line like L100

superset/db_engine_specs/ydb.py
```diff
@@ -0,0 +1,108 @@
...
+        params: dict[str, Any],
+    ) -> None:
+        if not database.encrypted_extra:
+            return
+
+        try:
+            encrypted_extra = json.loads(database.encrypted_extra)
+            connect_args = params.setdefault("connect_args", {})
+
+            if "protocol" in encrypted_extra:
```
giftig, line 99: Could you not just `connect_args["protocol"] = encrypted_extra.get("protocol")`? Or is it important the key be absent rather than `None`?

tests/unit_tests/db_engine_specs/test_ydb.py
```diff
@@ -0,0 +1,83 @@
...
+
+from superset.utils import json
+from tests.unit_tests.db_engine_specs.utils import assert_convert_dttm
+from tests.unit_tests.fixtures.common import dttm  # noqa: F401
+
+
+def test_epoch_to_dttm() -> None:
+    from superset.db_engine_specs.ydb import YDBEngineSpec
+
+    assert YDBEngineSpec.epoch_to_dttm() == "DateTime::MakeDatetime({col})"
```
giftig, line 34: I think we could probably live without this one since it's effectively just a constant string.

tests/unit_tests/db_engine_specs/test_ydb.py
```diff
@@ -0,0 +1,83 @@
...
+        ),
+        ("UnknownType", None),
+    ],
+)
+def test_convert_dttm(
+    target_type: str,
+    expected_result: Optional[str],
+    dttm: datetime,  # noqa: F811
+) -> None:
+    from superset.db_engine_specs.ydb import YDBEngineSpec as spec
```
giftig, line 53: Does this import need to be local?

giftig (commented):
Minor comments / questions but looks good to me.

I haven't had chance to try running your branch with YDB though, and we'll probably need to check if there's a process for including that `ydb.svg`, so we should wait for some more reviews.

giftig (approved):

tests/unit_tests/db_engine_specs/test_ydb.py
vgvoleg: I relied on other spec's tests.

tests/unit_tests/db_engine_specs/test_ydb.py
vgvoleg: I relied on other spec's tests :)

PR #31155, opened 2024-11-25
fix: helm chart deploy to open PRs to now-protected gh-pages branch

mistercrunch:
Cleaning up the mess I created by deleting the `gh-pages` branch by mistake while deleting hundreds of stale branches in the repo.

The changes here make it such that pushing a new helm chart version will now open a PR against the now-protected `gh-pages` branch, require a review and merge to be deployed.

Testing now with:
```
gh workflow run superset-helm-release.yml --ref helm_fix --field ref=master
```

Logs / iterations can be followed here -> https://github.com/apache/superset/actions/workflows/superset-helm-release.yml

The approach here is to:
- define a working branch `helm-publish-{SHA}`
- pull/push gh-pages onto that temp branch
- tell `chart-releaser-action` that the branch we're working with is NOT `gh-pages` (protected, can't push there), but instead is a temporary branch
- use a small script to open a PR from temp branch to gh-pages

### DIFF
--- a/.github/workflows/superset-helm-lint.yml
+++ b/.github/workflows/superset-helm-lint.yml
@@ -1,4 +1,4 @@
-name: Lint and Test Charts
+name: "Helm: lint and test charts"
 
 on:
   pull_request:
--- a/.github/workflows/superset-helm-release.yml
+++ b/.github/workflows/superset-helm-release.yml
@@ -56,6 +56,20 @@ jobs:
           git tag --list
           git show-ref --tags
           git tag -d superset-helm-chart-0.13.4 || true
+          lookup_latest_tag() {
+            git fetch --tags >/dev/null 2>&1
+            echo "Fetched all tags."
+
+            if git describe --tags --abbrev=0 HEAD~ 2>/dev/null; then
+              echo "Latest tag found by git describe:"
+              git describe --tags --abbrev=0 HEAD~
+            else
+              echo "git describe failed. Falling back to root commit."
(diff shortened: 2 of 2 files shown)


.github/actions/chart-releaser-action
rusackas, line 1: Huh... don't think I've seen it show this before. Neat!

.github/workflows/superset-helm-lint.yml
```diff
@@ -1,4 +1,4 @@
-name: Lint and Test Charts
+name: "Helm: lint and test charts"
```
rusackas, line 1: Nice improvement... I'll bet half the people that see this think it's eslint and dataviz charts :)

.github/workflows/superset-helm-release.yml
```diff
@@ -1,4 +1,8 @@
-name: Release Charts
+# This workflow automates the release process for Helm charts.
+# The workflow creates a new branch for the release and opens a pull request against the 'gh-pages' branch,
+# allowing the changes to be reviewed and merged manually.
```
rusackas, line 3: We should probably expand the code owners for these reviews... maybe you want to be one?

rusackas (approved):
LGTM. Don't see any risks here... like the approach, love the notes :D

.github/workflows/superset-helm-lint.yml
mistercrunch: It's been bugging me for a moment. Making the repo a better place, one line at a time.

PR #31163, opened 2024-11-26
fix(Dashboard): Backward compatible shared_label_colors field

geido:
### SUMMARY
A follow-up PR for https://github.com/apache/superset/pull/30646. This PR ensure that a dashboard that used the `shared_label_colors` field as a dict is compatible with the `shared_label_colors` being a `list` as per the latest changes and can be re-imported / copied without errors.

N.A.

### TESTING INSTRUCTIONS
1. Export a Dashboard before changes in https://github.com/apache/superset/pull/30646
2. Re-import the Dashboard after changes in https://github.com/apache/superset/pull/30646
3. Copy the dashboard and all of its charts utilizing "Save as" 
4. Copying the dashboard should work without errors

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/utils/colorScheme.ts
+++ b/superset-frontend/src/utils/colorScheme.ts
@@ -24,12 +24,24 @@ import {
 } from '@superset-ui/core';
 
 /**
- * Forces falsy namespace values to undefined to default to GLOBAL
+ * Force falsy namespace values to undefined to default to GLOBAL
  *
  * @param namespace
  * @returns - namespace or default undefined
  */
 export const getColorNamespace = (namespace?: string) => namespace || undefined;
+
+/**
+ *
+ * Field shared_label_colors used to be a dict of all colors for all labels.
+ * Force shared_label_colors field to be a list of actual shared labels.
+ *
+ * @param sharedLabelsColors - the shared label colors list
+ * @returns string[]
+ */
+export const enforceSharedLabelsColorsArray = (
+  sharedLabelsColors: string[] | Record<string, string> | undefined,
+) => (Array.isArray(sharedLabelsColors) ? sharedLabelsColors : []);
 
 /**
  * Get labels shared across all charts in a dashboard.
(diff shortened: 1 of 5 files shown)


superset/dashboards/schemas.py
```diff
@@ -116,6 +116,28 @@ def validate_json_metadata(value: Union[bytes, bytearray, str]) -> None:
...
+        attr: Union[str, None],
+        data: Union[Mapping[str, Any], None],
+        **kwargs: dict[str, Any],
+    ) -> list[str]:
+        if isinstance(value, list):
+            if all(isinstance(item, str) for item in value):
+                return value
+        elif isinstance(value, dict):
+            # Convert dict values to a list (for backward compatibility)
+            return []
```
Antonio-RiveroMartnez, line 136: Shouldn't we return the values?

superset-frontend/src/utils/colorScheme.ts
```diff
@@ -31,6 +31,10 @@ import {
  */
 export const getColorNamespace = (namespace?: string) => namespace || undefined;
 
+export const ensureSharedLabelsColorsArray = (
```
Antonio-RiveroMartnez, line 34: Probably because of the other "ensure" methods that transform  to a given type (Array for example), but this method seems to instead being doing a different thing with dicts by returning [], so we might want to consider a different name for it? or return a list with the dict values instead?

superset/dashboards/schemas.py
geido: Nope, it needs to be re-generated as the existing dict would be invalid

superset-frontend/src/utils/colorScheme.ts
Antonio-RiveroMartnez: Yeah, just that if I pass a dict {foo: "hello", bar: "world"} to something called ensure..Array I was expecting an array with the values and not an empty array? i.e, ["hello", "world"] instead of []. But probably a nit.

Antonio-RiveroMartnez (approved):

superset-frontend/src/utils/colorScheme.ts
geido: Ok, I see your point. I'll keep it like this for lack of a better name at the moment.

geido: Maybe enforce?

PR #31164, opened 2024-11-26
feat: purge OAuth2 tokens when DB changes

betodealmeida:
### SUMMARY

Add a new method to the `UpdateDatabaseCommand` so that, if the OAuth2 client information has changed, existing tokens are purged, since they are unlikely to still be valid. This is done if any of these keys change:

- Client ID (but not client secret)
- Scope (tokens are associated with a given scope)
- Endpoint URLs (likely a new OAuth2 provider, but not necessarily — we purge to be safe)

N/A

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/update.py
+++ b/superset/commands/database/update.py
@@ -78,6 +78,10 @@ class UpdateDatabaseCommand(BaseCommand):
         # since they're name based
         original_database_name = self._model.database_name
 
+        # Depending on the changes to the OAuth2 configuration we may need to purge
+        # existing personal tokens.
+        self._handle_oauth2()
+
         database = DatabaseDAO.update(self._model, self._properties)
         database.set_sqlalchemy_uri(database.sqlalchemy_uri)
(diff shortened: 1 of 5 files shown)


fisjac (commented):
Changes look good to me.

When OAuth2 client info is changed through `DATABASE_OAUTH2_CLIENTS` in the `config.py` file, will the purge trigger as well, or only when a `PUT` request is made to the db connection?

betodealmeida:
> Changes look good to me.
> 
> When OAuth2 client info is changed through `DATABASE_OAUTH2_CLIENTS` in the `config.py` file, will the purge trigger as well, or only when a `PUT` request is made to the db connection?

Ah, good point! I added a notice to the `config.py` about this, if people are changing the `config.py` file they should be able to handle the token deletion themselves.

fisjac (approved):

PR #31187, opened 2024-11-27
feat(sqllab): Popup notification when download data can exceed row count

justinpark:
### SUMMARY
When user clicks "download to CSV" we would like to add a warning that there may actually be more data, so you aren't downloading the full data. 

https://github.com/user-attachments/assets/5c4d2ccf-3409-4818-9b9d-1fac0ff02801

### TESTING INSTRUCTIONS

Set limit 10,000 and then run a query that may contain more than 10k.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/components/ResultSet/index.tsx
+++ b/superset-frontend/src/SqlLab/components/ResultSet/index.tsx
@@ -355,7 +355,7 @@ const ResultSet = ({
                     Modal.warning({
                       title: t('Download is on the way'),
                       content: t(
-                        'Downloading %(rows)s rows based on LIMIT setting! If you want the entire result set, change the LIMIT settings.',
+                        'Downloading %(rows)s rows based on the LIMIT configuration. If you want the entire result set, you need to adjust the LIMIT.',
                         { rows: rowsCount.toLocaleString() },
                       ),
                     });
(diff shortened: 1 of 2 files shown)


superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -342,7 +346,21 @@ const ResultSet = ({
...
+                onClick={() => {
+                  logAction(LOG_ACTIONS_SQLLAB_DOWNLOAD_CSV, {});
+                  if (
+                    limitingFactor === LimitingFactor.Dropdown &&
+                    limit === rowsCount
+                  ) {
+                    Modal.warning({
+                      title: t('Download is on the way'),
+                      content: t(
+                        'Downloading %(rows)s rows based on LIMIT setting! If you want the entire result set, change the LIMIT settings.',
```
michael-s-molina, line 358:
```suggestion
                        'Downloading %(rows)s rows based on the LIMIT configuration. If you want the entire result set, you need to adjust the LIMIT.',
```

michael-s-molina (approved):

PR #31199, opened 2024-11-29
fix(Databricks): Escape catalog and schema names in pre-queries

Vitor-Avila:
### SUMMARY
It's possible to create catalogs and schemas in Databricks with special characters, which requires to always "escape" them in queries. For example, if you use a hyphen in a catalog name, you must run:
``` sql
USE CATALOG `my-catalog`;
```

This PR escapes the catalog and schema names in the `get_prequeries` method for the Databricks Native Connector (if they were not escaped in the `connect_args` JSON).

No UI changes.

### TESTING INSTRUCTIONS
Unit tests added. For manual testing:
1. Create a catalog and a schema containing hyphens in Databricks.
2. Connect Superset to the instance using the Databricks Native connector.
3. Run a query in SQL Lab.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/databricks.py
+++ b/superset/db_engine_specs/databricks.py
@@ -464,8 +464,10 @@ class DatabricksNativeEngineSpec(DatabricksDynamicBaseEngineSpec):
     ) -> list[str]:
         prequeries = []
         if catalog:
+            catalog = f"`{catalog}`" if not catalog.startswith("`") else catalog
             prequeries.append(f"USE CATALOG {catalog}")
         if schema:
+            schema = f"`{schema}`" if not schema.startswith("`") else schema
             prequeries.append(f"USE SCHEMA {schema}")
         return prequeries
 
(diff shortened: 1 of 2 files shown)


superset/db_engine_specs/databricks.py
```diff
@@ -464,8 +464,10 @@ def get_prequeries(
     ) -> list[str]:
         prequeries = []
         if catalog:
+            catalog = f"`{catalog}`" if not catalog.startswith("`") else catalog
```
Vitor-Avila, line 467: I didn't use `database.quote_identifier()`, because the `Database` is no longer passed as an argument in `>=4.1.0`: https://github.com/apache/superset/blob/4.1.0/superset/db_engine_specs/base.py#L1412-L1431

Since this method is specifically for Databricks, I thought it was OK to hardcode the quote identifiers.

betodealmeida (approved):
tenor-1030972582

Issue #30307, opened 2024-09-17
Tracking Issue: Migrating from Mapbox to MapLibre [Approved SIP-130]


birkskyum:
This is tracking the implementation of the Approved SIP-130
- #28356

1. Step - Update legacy mapbox/deckgl plugins
The legacy Mapbox / DeckGL plugins use react-map-gl v6. In react-map-gl v7, this library has dual support for mapbox and maplibre, and updating will thus bring immediate improvements, and reduce the gap when transition to MapLibre:
- [ ] Update supercluster to v8 - #30305 
- [ ] Update react to v18
- [ ] Update react-map-gl to v7

Switching to the Plugin on the new Plugin:
- [ ]  New MapLibre plugin

StarkillerGDN:
Is it possible to have a configuration field in the UI to set a custom URL for a mapbox style ?

StarkillerGDN:
@birkskyum For example I have a Tileserver which serve mapbox styles. But to apply this new URI I must redeploy the superset instance with my hardcoded URI. Tell me if I'm wrong. It will be nice if we can provide a URI in a config file :)

birkskyum:
Absolutely, you be be able to use your own tile server like you do already - that shouldn't change.

datasc24:
Hello @birkskyum !
I agree with @StarkillerGDN, currently the tile server URL is hardcoded in the source code, which can make it difficult to customize for specific use cases. Indeed, to use our own tile server, we are obliged (if i understand correctly) to rebuild the entire project whereas we would like to use the Docker image directly instead. 

It would be great if this URL could be configured via the `superset_config.py ` file or as an alternative directly when building a chart (like in metabase).

Thank you for your work !

rusackas:
Heya! Are you still planning to tackle more of this effort, or is there no longer a need to keep this particular issue open? We can still consider the SIP in-progress separately (those just get closed when they're voted through).

saulfrr:
Hello

How is the work for the migration to MapLibre going? are there any news?

Thanks.

rolies106:
For anyone that want to move to Carto or MapLibre can try this hack while waiting for the official release, I'm using 4.1.2 tag and modify this script.

go to `superset-frontend/plugins/legacy-preset-chart-deckgl/src/utilities/Shared_DeckGL.jsx` and edit this section:

```javascript
export const mapboxStyle = {
  name: 'mapbox_style',
  config: {
    type: 'SelectControl',
    label: t('Map Style'),
[...]

PR #31201, opened 2024-12-01
chore(legacy-preset-deckgl): bump web-mercator in

birkskyum:
Part of https://github.com/apache/superset/issues/30307

Similar to:
- https://github.com/apache/superset/pull/30651

this pr updates the web-mercator deps in the deck preset to the same version as is used in the mapbox plugin


rusackas:
Hi! I'm not sure how long this has been in need of a rebase, but it looks like dependabot made the same upgrade at one point, so we should be all good! I'll close this one, and thank you for the contribution!

PR #31257, opened 2024-12-02
fix: add python-base layer to fix CI builds in 4.1

mistercrunch:


mistercrunch:
Fixed on the supersetbot side

PR #31268, opened 2024-12-03
refactor: Migrate AdhocFilterEditPopoverSqlTabContent to TypeScript

EnxDev:
### SUMMARY
Migrate AdhocFilterEditPopoverSqlTabContent to TypeScript

N.A.

### TESTING INSTRUCTIONS
All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
@EnxDev

superset-frontend/src/explore/components/controls/FilterControl/AdhocFilterEditPopoverSqlTabContent/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+      options.filter(
+        (option): option is ColumnMeta =>
+          typeof option === 'object' &&
+          option !== null &&
+          'column_name' in option &&
+          typeof option.column_name === 'string' &&
+          'type' in option,
+      ),
+    ),
+  );
```
justinpark, line 89: Please wrap this block by useMemo for the performance.

```suggestion
  const keywords = useMemo(() => sqlKeywords.concat(
    getColumnKeywords(
      options.filter(
        (option): option is ColumnMeta =>
          typeof option === 'object' &&
          option !== null &&
          'column_name' in option &&
          typeof option.column_name === 'string' &&
          'type' in option,
      ),
    ),
  ), [sqlKeywords]);
```

superset-frontend/src/explore/components/controls/FilterControl/AdhocFilterEditPopoverSqlTabContent/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const selectOptions = Object.values(Clauses).map(clause => ({
+    label: clause,
+    value: clause,
+  }));
```
justinpark, line 94: please wrap this block by useMemo for the performance.

```suggestion
  const selectOptions = useMemo(() => Object.values(Clauses).map(clause => ({
    label: clause,
    value: clause,
  })), [Clauses]);
```

superset-frontend/src/explore/components/controls/FilterControl/AdhocFilterEditPopoverSqlTabContent/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  onChange: (filter: AdhocFilter) => void;
+  options: OptionSortType[];
+  height: number;
+}) {
+  const aceEditorRef = useRef(null);
+  const theme = useTheme();
+
+  const handleAceEditorRef = useCallback(ref => {
+    if (ref) aceEditorRef.current = ref;
+  }, []);
```
justinpark, line 53: it's unnecessary. You can directly bind `aceEditorRef` like `ref={aceEditorRef}`

EnxDev: You're absolutely right. Binding aceEditorRef directly with ref={aceEditorRef} is sufficient, as it doesn't create new functions on every render. Using useCallback for this scenario is unnecessary and adds extra complexity without any practical advantage. Thank you for pointing that out! :)

geido (approved):

PR #31275, opened 2024-12-04
chore: relax greenlet requirements

sadpandajoe:
### SUMMARY

Currently we are pinning greenlet to be 3.0.3, however, there are some instances that greenlet needs to be in a higher version like 3.1.1. Loosening up the requirements here to include the upper bound as needed.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
I thought playwright has a hard requirement on 3.0.3 (jut cause they don't know better... submitted a PR to relax their req). The reason why I pinned it is because it throws off `supersetbot` - it tried to upgrade to 3.1.1 but playwright then complains...

sadpandajoe:
> I thought playwright has a hard requirement on 3.0.3 (jut cause they don't know better... submitted a PR to relax their req). The reason why I pinned it is because it throws off `supersetbot` - it tried to upgrade to 3.1.1 but playwright then complains...
> 
> I guess this only affects the `development.txt`, not the `base.txt`...

We currently use 3.1.1 in our system so when we try to do a build without this being relaxed it'll throw an error. So just relaxing it here but keeping the main .txt files untouched.

mistercrunch:
actually I have an idea, let me try something

mistercrunch (approved):
Moving the pin to `base.in`, let me know if that unblocks you. If not, we could move to `development.in`, but both environments would be out-of-sync in regards to their greenlet version, which isn't a problem.

PR #31282, opened 2024-12-04
Fix: this.props overwrite config in TextAreaControl

jasgok:
### SUMMARY
this.props overwrite config in TextAreaControl, in that case inModal state has the same style as not modal

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/TextAreaControl.jsx
+++ b/superset-frontend/src/explore/components/controls/TextAreaControl.jsx
@@ -97,6 +97,7 @@ class TextAreaControl extends Component {
 
       return (
         <TextAreaEditor
+          {...this.props}
           mode={this.props.language}
           style={style}
           minLines={minLines}
@@ -105,7 +106,6 @@ class TextAreaControl extends Component {
           defaultValue={this.props.initialValue}
           readOnly={this.props.readOnly}
           key={this.props.name}
-          {...this.props}
           onChange={this.onAreaEditorChange.bind(this)}
         />
       );


michael-s-molina:
@jasgok Could describe the problem you're facing in more detail? Maybe add before/after screenshots?

jasgok:
> @jasgok Could describe the problem you're facing in more detail? Maybe add before/after screenshots?

thank you for following up.

from chart "deck.gl", in avdvanced tab, where using "Edit javascript in modal"，
modal show only 10 lines but should show more as ` maxLines={inModal ? 1000 : this.props.maxLines}`  configured in TextAreaEditor.
As further inspection, this.props.maxLines=10 was found overwriting this config as `{...this.props}`  after it.

please refer to attached screenshots.
before
after

msyavuz:
@jasgok is this pr still needed? There were recent changes around modals and editors.

superset-frontend/src/explore/components/controls/TextAreaControl.jsx
```diff
@@ -97,6 +97,7 @@ class TextAreaControl extends Component {
 
       return (
         <TextAreaEditor
+          {...this.props}
```
mistercrunch, line 100: Seem to me both before/after this PR can have issue or create confusion, seems we should use the `...rest` pattern as in https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-core/src/components/Label/index.tsx#L37

rusackas:
Let us know if you can address the comments and resolve conflicts, @jasgok, we'd love to see this through :D Thanks in advance.

rusackas:
Feel free to touch things up and reopen/if when you'd like. Thanks!

PR #11391, opened 2020-10-22
feat: saved queries with execution info

dpgaspar:
### SUMMARY
Propagates the number of retrieved rows and timestamps to a "related" saved query when a SQLLab query executes.
This is a nice to have UI/UX information for a future home screen on Superset 

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Changes UI
- [ ] Requires DB Migration.
- [X] Confirm DB Migration upgrade and downgrade tested.
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #31294, opened 2024-12-04
fix(sqllab): Remove update_saved_query_exec_info to reduce lag

justinpark:
### SUMMARY
Currently, there is a significant bottleneck when executing queries in sqllab due to the delay in `update_saved_query_exec_info`. 

Superset

392578130-07d5b578-3beb-4797-a982-49ab524572bb_png__1197×268_

According to #11391, the information obtained during the `update_saved_query_exec_info` was added for display on the homepage, but it has been confirmed that the recently changed_date is actually being used.

To facilitate smoother execution of sqllab queries, this commit removes the `update_saved_query_exec_info` part. Additionally, it changed the displayed label (from `ran` to `modified`) on the homepage to reflect the actual data.

### TESTING INSTRUCTIONS

CI and specs

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/daos/query.py
+++ b/superset/daos/query.py
@@ -36,25 +36,6 @@ class QueryDAO(BaseDAO[Query]):
     base_filter = QueryFilter
 
     @staticmethod
-    def update_saved_query_exec_info(query_id: int) -> None:
-        """
-        Propagates query execution info back to saved query if applicable
-
-        :param query_id: The query id
-        :return:
-        """
-        query = db.session.query(Query).get(query_id)
-        related_saved_queries = (
-            db.session.query(SavedQuery)
-            .filter(SavedQuery.database == query.database)
-            .filter(SavedQuery.sql == query.sql)
-        ).all()
-        if related_saved_queries:
-            for saved_query in related_saved_queries:
-                saved_query.rows = query.rows
-                saved_query.last_run = datetime.now()
-
-    @staticmethod
     def save_metadata(query: Query, payload: dict[str, Any]) -> None:
         # pull relevant data from payload and store in extra_json
         columns = payload.get("columns", {})
--- a/superset-frontend/src/features/home/SavedQueries.tsx
+++ b/superset-frontend/src/features/home/SavedQueries.tsx
@@ -278,7 +278,7 @@ const SavedQueries = ({
                 url={`/sqllab?savedQueryId=${q.id}`}
                 title={q.label}
                 imgFallbackURL="/static/assets/images/empty-query.svg"
-                description={t('Ran %s', q.changed_on_delta_humanized)}
+                description={t('Modified %s', q.changed_on_delta_humanized)}
                 cover={
                   q?.sql?.length && showThumbnails && featureFlag ? (
                     <QueryContainer>
--- a/superset/sqllab/sql_json_executer.py
+++ b/superset/sqllab/sql_json_executer.py
@@ -95,7 +95,6 @@ class SynchronousSqlJsonExecutor(SqlJsonExecutorBase):
             data = self._get_sql_results_with_timeout(
                 execution_context, rendered_query, log_params
             )
-            self._query_dao.update_saved_query_exec_info(query_id)
             execution_context.set_execution_result(data)
         except SupersetTimeoutException:
             raise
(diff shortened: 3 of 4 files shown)


michael-s-molina:
I would like to add more context to this change to facilitate review:

> Currently, there is a significant bottleneck when executing queries in sqllab due to the delay in update_saved_query_exec_info.

`update_saved_query_exec_info` currently compares the executed SQL on SQL Lab with all saved queries to update when a saved query was last run. The cost of this operation can be really high depending on the SQL size and the number of saved queries that exist in the database. We also don't think the information of when a saved query was last run is worth that cost. In fact, the last changed information is way more important and it's what's displayed for the other entity types.

@rusackas @yousoph @villebro @dpgaspar @betodealmeida @mistercrunch

superset/daos/query.py
```diff
@@ -35,25 +35,6 @@
...
-        Propagates query execution info back to saved query if applicable
-
-        :param query_id: The query id
-        :return:
-        """
-        query = db.session.query(Query).get(query_id)
-        related_saved_queries = (
-            db.session.query(SavedQuery)
-            .filter(SavedQuery.database == query.database)
-            .filter(SavedQuery.sql == query.sql)
```
mistercrunch, line 50: Can't we just add `.filter(SavedQuery.id == query_id)` here? It should use the PK index and be super efficient. Also if we don't want to update if the query has been mutated, seems we should not only check `sql` and `database`, but schema/catalog too.

villebro: I was super confused by these filters, too. Is the intent here to update all entries where the same database + query combo are featured? Seems very strange to want to do that..?

mistercrunch: Yeah it says "check all saved queries that have the same SQL, and update their last-executed time". Clearly if it's not the same SavedQuery.id, we shouldn't update the last executed time.

mistercrunch: Adding a filter on SavedQuery.id should solve the long execution time in environments with lots of SavedQueries

villebro: But it seems to be very explicit about updating many rows, so I can only assume the intent here is to actually update many queries. Looking at #11391 it does indeed appear to be the intent. But this is clearly a perf problem, so I would do away with this feature, and just update the actual query that ran.

villebro: Ping @dpgaspar if you recall why this was needed.

justinpark: btw the condition `.filter(SavedQuery.id == query_id)` is not correct.
To find the saved query that matches the current query, it is appropriate to find the `saved_query_id` of the `tab_state` that matches the `latest_query_id` and `client_id` of the current query.

justinpark: If this information is truly necessary, it would be better to update the saved query by making an additional PUT request after the execute action is performed on the **frontend** (especially when the current query editor's savedQueryId exists.), rather than handling it during the `sqllab/execute process`.

mistercrunch: Gotcha, I though the `query_id` in context here was a `SavedQuery.id` but it's actually a `Query.id` ...

michael-s-molina: Even though we could do an additional PUT as @justinpark suggested, I still think showing the last time a saved query run is less important than showing the last time it was modified. That's what the other entities show and it's generally what users look for when they find a problem.

justinpark: I agree with @michael-s-molina  opinion. Unless there are significant reasons from @dpgaspar and others, I will prioritize merging this hotfix and will follow up with additional PRs if needed based on the remaining feedback.

michael-s-molina (approved):

mistercrunch:
>  we could do an additional PUT

it didn't seem to me like the last run time isn't displayed anywhere is it? It's not in the "Saved Query" CRUD at least, couldn't see it surfaced anywhere else... If that's the case, and knowing that it's not maintained now, should be just delete the column?

michael-s-molina:
> it didn't seem to me like the last run time isn't displayed anywhere is it? It's not in the "Saved Query" CRUD at least, couldn't see it surfaced anywhere else... If that's the case, and knowing that it's not maintained now, should be just delete the column?

Good point 👍🏼

PR #31303, opened 2024-12-05
feat: Adds helper functions for migrations

luizotavio32:
### SUMMARY

This PR aims to implement helper functions that will provide standard interfaces for the most used methods found in the migrations for the past 2 years, which are: create and drop tables, add and drop columns, create and drop indexes. Another objective to make logging standard among migrations by providing more human readable logs and displaying the same type of information.

These new functions will apply some validations before executing such as: 
- When adding/dropping a column, does this column already exist on the table? 
- Does the index to be created is already present on the table? 

These types of validations will make the migrations more resilient. The added helper functions and their logs are:

```
def create_table(table_name: str, *columns: SchemaItem) -> None:

INFO [alembic] Creating table temp_table…
INFO [alembic] Table temp_table created
```

```
def drop_table(table_name: str) -> None:

INFO  [alembic] Dropping table temp_table...
INFO  [alembic] Table temp_table dropped
```

```
def add_columns(table_name: str, *columns: Column) -> None:

INFO [alembic] Adding column new_id on table temp_table
INFO [alembic] Column id already present on table temp_table Skipping…
```

```
def drop_columns(table_name: str, *columns: str) -> None:

INFO  [alembic] Column address is not present on table temp_table Skipping...
INFO  [alembic] Dropping column new_id from table temp_table
```

```
def create_index(table_name: str, index_name: str, *columns: str) -> None:

INFO [alembic] Creating index my_index on table temp_table
INFO [alembic] Table temp_table already has index my_index Skipping…
```

```
def drop_index(table_name: str, index_name: str) -> None:

INFO  [alembic] Dropping index my_index from table temp_table
INFO  [alembic] Table temp_table doesn't have index other_index Skipping...
```

There’s also a new method for executing database operations in batch. This function will receive a `Callable`, the `count` and `batch_size` as the parameters. When executing, the `Callable` will receive the offset and limit for the current processed batch.

```
def batch_operation(callable: Callable[[int, int], None], count: int, batch_size: int) -> None:

INFO [alembic] Progress: 0/100,000 (0.00%)
INFO [alembic] Progress: 25,000/100,000 (25.00%)
INFO [alembic] Progress: 50,000/100,000 (50.00%)
INFO [alembic] Progress: 75.000/100,000 (75.00%)
INFO [alembic] Progress: 100,000/100,000 (100%)
INFO [alembic] End: callable_example batch operation successfully executed
```

### TESTING INSTRUCTIONS
Execute the migrations using `superset db upgrade` and superset db downgrade` commands and make sure everything works as expected.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/shared/utils.py
+++ b/superset/migrations/shared/utils.py
@@ -23,13 +23,22 @@ from uuid import uuid4
 
 import sqlalchemy as sa
 from alembic import op
-from sqlalchemy import inspect
+from sqlalchemy import Column, inspect
 from sqlalchemy.dialects.mysql.base import MySQLDialect
 from sqlalchemy.dialects.postgresql.base import PGDialect
+from sqlalchemy.dialects.sqlite.base import SQLiteDialect  # noqa: E402
+from sqlalchemy.engine.reflection import Inspector
 from sqlalchemy.exc import NoSuchTableError
 from sqlalchemy.orm import Query, Session
+from sqlalchemy.sql.schema import SchemaItem
 
 from superset.utils import json
+
+GREEN = "\033[32m"
+RESET = "\033[0m"
+YELLOW = "\033[33m"
+RED = "\033[31m"
+LRED = "\033[91m"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


sadpandajoe (approved):

mistercrunch:
Not directly related but I've been considering squashing old migration into a big large initial one, keeping say 2 years worth or migrations or since say 2.x

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,211 @@ def has_table(table_name: str) -> bool:
...
+        return  # sqlite doesn't like constraints
+
+    if has_table(table_name):
+        foreign_keys = inspector.get_foreign_keys(table_name)
+
+        for fk in foreign_keys:
+            op.drop_constraint(fk["name"], table_name, type_="foreignkey")
+
+
+logger = logging.getLogger("alembic")
```
michael-s-molina, line 216: This entry should replace the one at the beginning of the file.

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,211 @@ def has_table(table_name: str) -> bool:
...
+    connection = op.get_bind()
+    inspector = Inspector.from_engine(connection)
+
+    if isinstance(connection.dialect, SQLiteDialect):
+        return  # sqlite doesn't like constraints
+
+    if has_table(table_name):
+        foreign_keys = inspector.get_foreign_keys(table_name)
+
+        for fk in foreign_keys:
```
michael-s-molina, line 212: Could you add log messages here in the same way you do for the other methods? Something like:

```
logger.info(f"Dropping foreign key X for table Y...")
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,211 @@ def has_table(table_name: str) -> bool:
...
+    the start index and the end index of the current batch.
+    :param count: The total number of items to process.
+    :param batch_size: The number of items to process in each batch.
+    """
+    for offset in range(0, count, batch_size):
+        percentage = (offset / count) * 100 if count else 0
+        logger.info(f"Progress: {offset:,}/{count:,} ({percentage:.2f}%)")
+        callable(offset, min(offset + batch_size, count))
+
+    logger.info(f"Progress: {count:,}/{count:,} (100%)")
```
michael-s-molina, line 282: If count is zero you'll get `ZeroDivisionError`.

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,211 @@ def has_table(table_name: str) -> bool:
...
+def add_columns(table_name: str, *columns: Column) -> None:
     """
-    if not table_has_column(table_name, column.name):
-        print(f"Adding column '{column.name}' to table '{table_name}'.\n")
-        op.add_column(table_name, column)
-    else:
-        print(f"Column '{column.name}' already exists in table '{table_name}'.\n")
+    Adds new columns to an existing database table.
+
+    For each column on columns list variable, it checks if the column already exists in the table. If a column already exists, it logs an informational
```
michael-s-molina, line 292: Could you apply the same for similar function docs?
```suggestion
    If a column already exists, it logs an informational
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,211 @@ def has_table(table_name: str) -> bool:
...
+    message and skips the addition of that column. Otherwise, it proceeds to add the new column to the table.
+
+    The operation is performed within a batch operation context which allows a more performant approach
```
michael-s-molina, line 295: Could you apply the same for similar function docs?
```suggestion
    The operation is performed using Alembic's batch_alter_table.
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,211 @@ def has_table(table_name: str) -> bool:
...
+            f"Table {LRED}{table_name}{RESET} already has index {LRED}{index_name}{RESET} Skipping..."
+        )
+        return
+
+    logger.info(
+        f"Creating index {GREEN}{index_name}{RESET} on table {GREEN}{table_name}{RESET}"
+    )
+
+    with op.batch_alter_table(table_name) as batch_op:
+        batch_op.create_index(index_name=index_name, columns=columns)
```
michael-s-molina, line 374: Do you need to use a batch operation here given that it's a single operation? Same for `drop_index`.

michael-s-molina (commented):
Thank you for the valuable PR @luizotavio32. I left some comments.

superset/migrations/shared/utils.py
luizotavio32: Done 
image

superset/migrations/shared/utils.py
luizotavio32: Should I add a condition to skip this informational if count = 0 or print it: `Progress: 1/1 (100%)` ?

superset/migrations/shared/utils.py
luizotavio32: Updated on [commit](https://github.com/apache/superset/pull/31303/commits/5560992f36d18258dc3e394118ca50e8eed94464)

superset/migrations/shared/utils.py
luizotavio32: Updated on [commit](https://github.com/apache/superset/pull/31303/commits/5560992f36d18258dc3e394118ca50e8eed94464)

superset/migrations/shared/utils.py
luizotavio32: Updated on [commit](https://github.com/apache/superset/pull/31303/commits/5560992f36d18258dc3e394118ca50e8eed94464)

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+            )
+            op.drop_constraint(fk["name"], table_name, type_="foreignkey")
+
+
+def create_table(table_name: str, *columns: SchemaItem) -> None:
+    """
+    Creates a database table with the specified name and columns.
+
+    This function checks if a table with the given name already exists in the database.
+    If the table already exists, it logs an informational.
```
michael-s-molina, line 223:
```suggestion
    If the table already exists, it logs an informational message and skips the creation process.
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+    This function checks if a table with the given name exists in the database.
+    If the table does not exist, it logs an informational message and skips the dropping process.
+    If the table exists, it first attempts to drop all foreign key constraints associated with the table
+    (handled by `drop_fks_for_table`) and then proceeds to drop the table.
+
+    :param table_name: The name of the table to be dropped.
+    """
+
+    if not has_table(table_name=table_name):
+        logger.info(f"Table {GREEN}{table_name}{RESET} doesn't exist Skipping...")
```
michael-s-molina, line 252:
```suggestion
        logger.info(f"Table {GREEN}{table_name}{RESET} doesn't exist. Skipping...")
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+    This function checks if a table with the given name already exists in the database.
+    If the table already exists, it logs an informational.
+    Otherwise, it proceeds to create a new table using the provided name and schema columns.
+
+    :param table_name: The name of the table to be created.
+    :param columns: A variable number of arguments representing the schema just like when calling alembic's method create_table()
+    """
+
+    if has_table(table_name=table_name):
+        logger.info(f"Table {LRED}{table_name}{RESET} already exists Skipping...")
```
michael-s-molina, line 231:
```suggestion
        logger.info(f"Table {LRED}{table_name}{RESET} already exists. Skipping...")
```

superset/migrations/shared/utils.py
michael-s-molina: If count is zero you should log a different message to inform admins. Only log the progress one when count > 0.

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+    logger.info(
+        f"End: {GREEN}{callable.__name__}{RESET} batch operation {GREEN}succesfully{RESET} executed"
+    )
+
+
+def add_columns(table_name: str, *columns: Column) -> None:
+    """
+    Adds new columns to an existing database table.
+
+    If a column already exists, it logs an informational.
```
michael-s-molina, line 291:
```suggestion
    If a column already exists, it logs an informational message and skips the adding process.
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
     """
-    Adds a column to a table if it does not already exist.
 
-    :param table_name: Name of the table.
-    :param column: SQLAlchemy Column object.
+    cols_to_add = []
+    for col in columns:
+        if table_has_column(table_name=table_name, column_name=col.name):
+            logger.info(
+                f"Column {LRED}{col.name}{RESET} already present on table {LRED}{table_name}{RESET} Skipping..."
```
michael-s-molina, line 304:
```suggestion
                f"Column {LRED}{col.name}{RESET} already present on table {LRED}{table_name}{RESET}. Skipping..."
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+
+    :param table_name: The name of the table from which the columns will be removed.
+    :param columns: A list of column names to be dropped.
+    """
+
+    cols_to_drop = []
+    for col in columns:
+        if not table_has_column(table_name=table_name, column_name=col):
+            logger.info(
+                f"Column {LRED}{col}{RESET} is not present on table {LRED}{table_name}{RESET} Skipping..."
```
michael-s-molina, line 334:
```suggestion
                f"Column {LRED}{col}{RESET} is not present on table {LRED}{table_name}{RESET}. Skipping..."
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+                f"Dropping column {GREEN}{col}{RESET} from table {GREEN}{table_name}{RESET}"
+            )
+            batch_op.drop_column(col)
+
+
+def create_index(table_name: str, index_name: str, *columns: str) -> None:
+    """
+    Creates an index on specified columns of an existing database table.
+
+    If the index already exists, it logs an informational.
```
michael-s-molina, line 351:
```suggestion
    If the index already exists, it logs an informational message and skips the creation process.
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+    Otherwise, it proceeds to create a new index with the specified name on the given columns of the table.
+
+    :param table_name: The name of the table on which the index will be created.
+    :param index_name: The name of the index to be created.
+    :param columns: A list column names where the index will be created
+    """
+
+    if table_has_index(table=table_name, index=index_name):
+        logger.info(
+            f"Table {LRED}{table_name}{RESET} already has index {LRED}{index_name}{RESET} Skipping..."
```
michael-s-molina, line 361:
```suggestion
            f"Table {LRED}{table_name}{RESET} already has index {LRED}{index_name}{RESET}. Skipping..."
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+    )
+
+    op.create_index(table_name=table_name, index_name=index_name, columns=columns)
+
+
+def drop_index(table_name: str, index_name: str) -> None:
+    """
+    Drops an index from an existing database table.
+
+    If the index does not exists, it logs an informational.
```
michael-s-molina, line 376:
```suggestion
    If the index does not exists, it logs an informational message and skips the dropping process.
```

superset/migrations/shared/utils.py
```diff
@@ -185,15 +193,201 @@ def has_table(table_name: str) -> bool:
...
+    Otherwise, it proceeds with the removal operation.
+
+    :param table_name: The name of the table from which the index will be dropped.
+    :param index_name: The name of the index to be dropped.
+    """
+
+    if not table_has_index(table=table_name, index=index_name):
+        logger.info(
+            f"Table {LRED}{table_name}{RESET} doesn't have index {LRED}{index_name}{RESET} Skipping..."
```
michael-s-molina, line 385:
```suggestion
            f"Table {LRED}{table_name}{RESET} doesn't have index {LRED}{index_name}{RESET}. Skipping..."
```

michael-s-molina (approved):

PR #31316, opened 2024-12-06
feat(sqllab): giving the query history pane a facelift

mistercrunch:
Fixing some styling and UI components in the query history pane in SQL Lab

* using label with monospace for timestamps
* set mouse to pointer on clickable icons
* using consistent icon size 
* make the View clickable a button instead of label

### before
Screenshot 2024-12-05 at 7 04 25 PM

### after
Screenshot 2024-12-05 at 7 03 00 PM

### DIFF
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -421,7 +421,7 @@ export function postStopQuery(query) {
     })
       .then(() => dispatch(stopQuery(query)))
       .then(() => dispatch(addSuccessToast(t('Query was stopped.'))))
-      .catch(() => dispatch(addDangerToast(t('Failed at stopping query.'))));
+      .catch(() => dispatch(addDangerToast(t('Failed to stop query.'))));
   };
 }
 
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -384,7 +384,6 @@ export function runQueryFromSqlEditor(
   ctas,
   ctasMethod,
 ) {
-  console.log('YOYO', queryEditor);
   return function (dispatch, getState) {
     const qe = getUpToDateQuery(getState(), queryEditor, queryEditor.id);
     const query = {
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -384,6 +384,7 @@ export function runQueryFromSqlEditor(
   ctas,
   ctasMethod,
 ) {
+  console.log('YOYO', queryEditor);
   return function (dispatch, getState) {
     const qe = getUpToDateQuery(getState(), queryEditor, queryEditor.id);
     const query = {
@@ -421,9 +422,7 @@ export function postStopQuery(query) {
     })
       .then(() => dispatch(stopQuery(query)))
       .then(() => dispatch(addSuccessToast(t('Query was stopped.'))))
-      .catch(() =>
-        dispatch(addDangerToast(t('Failed at stopping query. %s', query.id))),
-      );
+      .catch(() => dispatch(addDangerToast(t('Failed at stopping query.'))));
   };
 }
 
--- a/superset-frontend/src/SqlLab/components/QueryTable/index.tsx
+++ b/superset-frontend/src/SqlLab/components/QueryTable/index.tsx
@@ -35,6 +35,7 @@ import TableView from 'src/components/TableView';
 import Button from 'src/components/Button';
 import { fDuration } from 'src/utils/dates';
 import Icons from 'src/components/Icons';
+import Label from 'src/components/Label';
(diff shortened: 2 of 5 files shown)


superset-frontend/src/SqlLab/actions/sqlLab.js
```diff
@@ -384,6 +384,7 @@ export function runQueryFromSqlEditor(
   ctas,
   ctasMethod,
 ) {
+  console.log('YOYO', queryEditor);
```
rusackas, line 387:
```suggestion
```

superset-frontend/src/SqlLab/actions/sqlLab.js
```diff
@@ -421,9 +422,7 @@ export function postStopQuery(query) {
     })
       .then(() => dispatch(stopQuery(query)))
       .then(() => dispatch(addSuccessToast(t('Query was stopped.'))))
-      .catch(() =>
-        dispatch(addDangerToast(t('Failed at stopping query. %s', query.id))),
-      );
+      .catch(() => dispatch(addDangerToast(t('Failed at stopping query.'))));
```
rusackas, line 425: "Failed to stop query"?

superset-frontend/src/SqlLab/components/QueryTable/index.tsx
```diff
@@ -125,49 +126,52 @@ const QueryTable = ({
...
         config: {
-          icon: <Icons.XSmall iconColor={theme.colors.error.base} />,
+          icon: <Icons.CloseOutlined iconColor={theme.colors.error.base} />,
           label: t('Failed'),
         },
       },
       running: {
         config: {
-          icon: <Icons.Running iconColor={theme.colors.primary.base} />,
+          icon: <Icons.LoadingOutlined iconColor={theme.colors.primary.base} />,
```
rusackas, line 148: Is this the animated one? AntD has a way to make it spin IIRC.

superset-frontend/src/SqlLab/components/QueryTable/index.tsx
```diff
@@ -187,16 +191,14 @@ const QueryTable = ({
         const status = statusAttributes[state] || statusAttributes.error;
 
         if (q.endDttm) {
-          q.duration = fDuration(q.startDttm, q.endDttm);
+          q.duration = (
+            <Label
+              style={{ 'font-family': theme.typography.families.monospace }}
+            >
```
rusackas, line 197: Since we're doing this twice here with style overrides, it might make sense to build it into the Label component as a "monospace" prop.

superset-frontend/src/SqlLab/components/QueryTable/index.tsx
mistercrunch: oh right, I should use that one. One issue I had is not all icons show up as the same size and the smaller ones don't seem to respond to the iconSize prop, we should fix that. It's apparent in the storybook

superset-frontend/src/SqlLab/components/QueryTable/index.tsx
mistercrunch: I agree, was on the edge of doing it there. Can refactor this in

kasiazjc:
Looks clean @mistercrunch! Two things from me me: 
- icons seems big on comparison to the rest of the items in the row - what size is it? 
- I like pill for duration, but I think using it for timestamp might be too much - we have so many different items that it can get overwhelming and in other places we just use text for timestamp 

Thoughts?

mistercrunch:
@kasiazjc -> screenshot based on latest tweaks
Screenshot 2024-12-09 at 10 28 53 AM

Thinking a bit more about it, I'd like to create a standard way to represent timestamps/durations through a reusable react component. 

GPT-designer suggested this with a little clock which I think is kind of cute, but will make things more busy ->
Screenshot 2024-12-09 at 10 30 23 AM

My issue with out current CRUD list/table views is that it looks very `raw` (I mean that as opposed to "rich"), and would love to make it looking richer if that makes sense. Adding clock icons might look more busy, but also richer in some capacity.

kasiazjc:
> @kasiazjc -> screenshot based on latest tweaks Screenshot 2024-12-09 at 10 28 53 AM
> 
> Thinking a bit more about it, I'd like to create a standard way to represent timestamps/durations through a reusable react component.
> 
> GPT-designer suggested this with a little clock which I think is kind of cute, but will make things more busy -> Screenshot 2024-12-09 at 10 30 23 AM
> 
> My issue with out current CRUD list/table views is that it looks very `raw` (I mean that as opposed to "rich"), and would love to make it looking richer if that makes sense. Adding clock icons might look more busy, but also richer in some capacity.

I think it looks more balanced with smaller icons and the pills for timestamp and duration look better too with it. Looks good for now and let's figure out crud enrichment/details next!

rusackas:
image
This progress bar misalignment is nagging at me... I'll see about a PR for that when I come up for air next :D

mistercrunch:
@rusackas addressed the comments, probably mergeable as is

rusackas (approved):

rusackas:
LGTM!

PR #31323, opened 2024-12-06
fix: Use clickhouse sqlglot dialect for YDB

### LINKED ISSUE #31322: Unable to use DB names with prefixes for YDB
### Bug description

Superset with YDB works fine with queries like
```
SELECT * FROM db_name
```
in SQL Lab.

But if database has a prefix, which should be defined like:

```
SELECT * from `prefix/db_name`
```

It fails with parsing error

### Screenshots/recordings

image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

vgvoleg:
### SUMMARY

Superset uses sqlglot to parse queries. It is not obvious in what cases it performs parsing and in what cases it doesnt, but we see an error where we are not able to use backticks and slashes in SQL Lab while creating datasets.
For example, query like

```
SELECT * FROM `prefix/some_name`
```

will fail because no sqlglot dialect is defined for YDB.

To have an ability to use db's with slashes we can use clickhouse's sqlglot dialect until we make our own.

Fixes https://github.com/apache/superset/issues/31322

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/31322
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql/parse.py
+++ b/superset/sql/parse.py
@@ -91,6 +91,7 @@ SQLGLOT_DIALECTS = {
     "teradatasql": Dialects.TERADATA,
     "trino": Dialects.TRINO,
     "vertica": Dialects.POSTGRES,
+    "yql": Dialects.CLICKHOUSE,
 }
 
 


michael-s-molina:
@vgvoleg There's also the [SQLGLOT_DIALECTS_EXTENSIONS](https://github.com/apache/superset/blob/48864ce8c7601c1c7d30114acbe7cd62e03b64a4/superset/config.py#L255) option if this is a temporary fix.

mistercrunch (approved):
LGTM, but would be nice for `ydb` to create is own dialect, even if it's just importing the clickhouse one and overriding nothing at first, and allowing for a place if/where it may diverge

vgvoleg:
> LGTM, but would be nice for `ydb` to create is own dialect, even if it's just importing the clickhouse one and overriding nothing at first, and allowing for a place if/where it may diverge

as I see it will be not easy to use new sqlglot version
https://github.com/apache/superset/blob/48864ce8c7601c1c7d30114acbe7cd62e03b64a4/pyproject.toml#L96

PR #31330, opened 2024-12-06
feat(filters): adding filter type icons in filter config modal.

rusackas:
### SUMMARY

Adds filter/divider icons in the Filter config modal, so that you can finally tell what these darn things _are_.

Pasted_Image_12_6_24__2_39 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FilterTitleContainer.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FilterTitleContainer.tsx
@@ -51,6 +51,18 @@ export const FilterTitle = styled.div`
   `}
 `;
 
+const StyledFilterIcon = styled(Icons.Filter)`
+  color: ${({ theme }) => theme.colorIcon};
+  margin-right: ${({ theme }) => theme.sizeUnit * 2}px;
+  line-height: normal;
+`;
+
+const StyledDividerIcon = styled(Icons.PicCenterOutlined)`
+  color: ${({ theme }) => theme.colorIcon};
+  margin-right: ${({ theme }) => theme.sizeUnit * 2}px;
+  line-height: normal;
+`;
+
 const StyledWarning = styled(Icons.ExclamationCircleOutlined)`
   color: ${({ theme }) => theme.colorErrorText};
   &.anticon {
@@ -74,6 +86,8 @@ interface Props {
   filters: string[];
   erroredFilters: string[];
 }
+
+const isFilterDivider = (id: string) => id.startsWith('NATIVE_FILTER_DIVIDER');
 
 const FilterTitleContainer = forwardRef<HTMLDivElement, Props>(
   (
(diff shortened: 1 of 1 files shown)


michael-s-molina:
@rusackas It looks like the icons are vertically misaligned in the screenshot.

justinpark:
I think having the divider designed similarly to the filter can cause confusion.
How about changing it to make the divider block easily recognizable, rather than using an icon like this?

USA_Births_Names

mistercrunch (approved):
Approving as net positive as it addresses the main issue. Can follow up with another PR if/when we get a design improvement

michael-s-molina (changes requested):
@rusackas Can you check this [comment](https://github.com/apache/superset/pull/31330#issuecomment-2528766132)?

rusackas:
> @rusackas Can you check this [comment](https://github.com/apache/superset/pull/31330#issuecomment-2528766132)?

This is because the (custom) icons themselves are goofy, with different SVG bounding boxes. I can add custom CSS overrides to bump things up or down a pixel, but that seems like the wrong fix. I'd rather go fiddle with the sizing/alignment of the icons themselves in a subsequent PR.

michael-s-molina:
>  I'd rather go fiddle with the sizing/alignment of the icons themselves in a subsequent PR.

I would argue that the whole point of this PR is to add icons and that their alignment seems to be essential for what you're doing. In other words, it looks weird that in one PR you add misaligned icons and in another PR you fix the alignment.

mistercrunch:
Just noting that there are no icon currently. You could remove the icons and that would make this PR focussed on the trigger.

In terms of design, one option could be to `border-bottom` the container div, looking like an underline going across the panel. I would look similar to what it looks like in the dashboard (text with a horizontal bar underneat)

michael-s-molina:
I'm ok with option 1 with a TODO to review the CSS after the Ant Design upgrade. From a user's perspective the icons will be aligned.

mistercrunch:
for reference -> Screenshot 2024-12-11 at 1 00 42 PM

rusackas:
Ok, seems I've stepped in it again... 

Pinging Dr. @kasiazjc for good measure. 

The options continue to proliferate...
1) Icons
2) No icons, and treat the divider differently (underline, or some other treatment)
3) Skip all this and consider a revamp, wherein:
* Filters in the filter bar get drag/drop reordering and "edit" buttons (on hover)
* "Add/edit filters" becomes "Add filter" and "Add divider" in the filter bar ⚙️ menu.
* Filter config modal now only needs to support a single filter... so just ditch its left panel entirely.

rusackas:
Rebased on latest master (was 2,110 commits behind). Updates include:

- Migrated to new AntD theme tokens (`theme.colorIcon`, `theme.sizeUnit`)
- Cleaned up the `isFilterDivider` helper function
- Squashed commits

Still need to manually verify the icon alignment issue raised by @michael-s-molina before taking out of draft.

PR #31332, opened 2024-12-06
fix: prevent multiple pvm errors on migration

eschutho:
### SUMMARY

We're seeing some errors when running the catalog migration because of some PVMs that already exist. In that case, I'm not going to update the PVM. 
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2024-05-01_10-52_58d051681a3b_add_catalog_perm_to_tables.py
```diff
@@ -45,10 +49,21 @@ def upgrade():
         "slices",
         sa.Column("catalog_perm", sa.String(length=1000), nullable=True),
     )
+
+    if not table_has_index(perm_table, perm_index_view_menu):
```
sadpandajoe, line 53: might be good to add these into a helper function call create_index that will do the check and add it like here https://github.com/apache/superset/pull/31303

eschutho: Nice, I was thinking of a helper function. I don't want to create a conflict with that PR. Perhaps we can use his helper function after it's merged.

superset/migrations/shared/catalogs.py
```diff
@@ -505,7 +505,15 @@ def upgrade_schema_perms(
...
             .one_or_none()
         ):
-            existing_pvm.name = new_perm
+            # check that new_perm does not exist
+            if session.query(ViewMenu).filter_by(name=new_perm).one_or_none():
+                logger.warning(
+                    "New permission %s already exists. Removing old one",
+                    new_perm,
+                )
+                session.delete(existing_pvm)
```
eschutho, line 514: @betodealmeida not sure if you would prefer to return pvms to be deleted instead of deleting them here

superset/migrations/versions/2024-05-01_10-52_58d051681a3b_add_catalog_perm_to_tables.py
```diff
@@ -45,10 +49,21 @@ def upgrade():
...
+
     upgrade_catalog_perms(engines={"postgresql"})
 
 
 def downgrade():
     downgrade_catalog_perms(engines={"postgresql"})
+    if table_has_index(perm_table, perm_index_permission_id):
+        op.drop_index(op.f(perm_index_permission_id), table_name=perm_table)
+    if table_has_index(perm_table, perm_index_view_menu):
+        op.drop_index(op.f(perm_index_view_menu), table_name=perm_table)
```
eschutho, line 67: @betodealmeida any thoughts whether you think this change should stay with the new migration or in a different PR? My thought is to keep them together so that tables don't get out of sync, but I also see the benefit of splitting them up.

superset/migrations/shared/catalogs.py
mistercrunch: Do we really need to delete anything here? Should we just rename the existing ones and create new ones for new catalogs? The old ones are attached to roles already and we probably want to maintain those associations.

superset/migrations/shared/security_converge.py
```diff
@@ -94,7 +95,10 @@ def __repr__(self) -> str:
 
 class PermissionView(Base):  # type: ignore
     __tablename__ = "ab_permission_view"
-    __table_args__ = (UniqueConstraint("permission_id", "view_menu_id"),)
+    __table_args__ = (
+        sqla.Index("idx_permission_view_menu_id", "view_menu_id"),
+        sqla.Index("idx_permission_permission_id", "permission_id"),
```
mistercrunch, line 100: haven't looked at the queries against that table, but could be that a compound index work well here. I saw some queries that predicate on `view_menu_id` and some on both `view_menu_id` and `permission_id`. If permission is never predicated by itself, the compound on `view_menu_id` and `permission_id` (in that order) would cover those

dpgaspar: Adding more info here, FAB already has a compound index since it defines a unique constraint on these FK: https://github.com/preset-io/Flask-AppBuilder/blob/master/flask_appbuilder/security/sqla/models.py#L73

Screenshot 2024-12-09 at 10 36 03

Checked the security converge code, and we have the following queries:
``` python
session.query(ViewMenu).filter(ViewMenu.name == view_name).one_or_none()
```
``` python
session.query(Permission).filter(Permission.name == permission_name).one_or_none()
```
``` python
session.query(PermissionView)
        .filter(
            PermissionView.view_menu_id == view_menu.id,
            PermissionView.permission_id == permission.id,
        )
```
``` python
        session.query(PermissionView)
        .join(Permission)
        .join(ViewMenu)
        .filter(ViewMenu.name == view_name, Permission.name == permission_name)
```
``` python
            session.query(PermissionView)
            .join(Permission)
            .filter(Permission.name == old_permission_name)
```
``` python
            session.query(PermissionView)
            .join(ViewMenu)
            .filter(ViewMenu.name == old_view_name)
```

We have indexes to efficiently query all of these:
- index on `ViewMenu.name`
- index on `Permission.name`
- Indexes on `Permission.id` and `ViewMenu.id`
 
Yet we can have more eficient joins by adding indexes on the FK for `ab_permission_view_role` and `ab_permission_view`
https://github.com/dpgaspar/Flask-AppBuilder/pull/2293

superset/migrations/shared/catalogs.py
betodealmeida: @mistercrunch the current code does that, we get the current perm (where the catalog name is `null`, see how we compute `current_perm` in line 492) and change it to point to the new perm name `new_perm`, which includes the default catalog name (line 497). Then we simply rename the existing PVM to have the new name, so that existing associations like roles are preserved (line 508 on the left).

In regards to this PR, I don't think we should delete existing PVMs. First, because they should not exist... the only reason I can imagine why it would exist is bad upgrade-downgrade that didn't rename the PVMs correctly. If that happens, I think it's better to just keep the existing PVM, since it already has the correct name, and might be associated with roles.

superset/migrations/shared/catalogs.py
```diff
@@ -683,7 +691,15 @@ def downgrade_schema_perms(
...
                 schema,
             )
-            pvms_to_rename.append((pvm, new_name))
+            # check to see if the new name already exists
+            if session.query(ViewMenu).filter_by(name=new_name).one_or_none():
+                logger.warning(
+                    "New permission %s already exists. Deleting old one",
+                    new_name,
+                )
+                pvms_to_delete.append(pvm)
```
betodealmeida, line 700: Same here, I would just leave the existing one, something like:

```python
            # check to see if the new name already exists
            if not session.query(ViewMenu).filter_by(name=new_name).one_or_none():
                # new perm doesn't exist, rename existing one to the new name
                pvms_to_rename.append((pvm, new_name))

mistercrunch:
Yes, deleting is yet more potential locks. I also think there are methods/CLI commands that will clean up perms that should not exist. I forgot whether it's a `fab` feature or not or what triggers it, but may even be on app startup (?)

superset/migrations/shared/security_converge.py
eschutho: Thanks @dpgaspar! I'll remove the changes here and we'll bring them in from FAB then.

mistercrunch (approved):
LGTM

PR #31333, opened 2024-12-07
fix: remove sort values on stacked totals

eschutho:
### SUMMARY

When a y axis sort is selected with a stacked bar chart with totals shown, the y axis column value was being added to the total. This pr checks that only the metrics are being used in calculating totals by passing in the list of metrics labels to the total calculation function, and only totaling the columns that are metrics, not the ones used in sorting.

Before:
image

After:
Screenshot 2024-12-06 at 4 54 10 PM

### TESTING INSTRUCTIONS

Create a chart with one metric, and a sort value. Turn on stacked and show values. The sort value column should not be added to the total of the metric. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/query/getMetricLabel.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/getMetricLabel.ts
@@ -23,7 +23,7 @@ export default function getMetricLabel(
   metric: QueryFormMetric,
   index?: number,
   queryFormMetrics?: QueryFormMetric[],
-  verboseMap: Record<string, string> = {},
+  verboseMap?: Record<string, string>,
 ): string {
   let label = '';
   if (isSavedMetric(metric)) {
@@ -37,5 +37,5 @@ export default function getMetricLabel(
   } else {
     label = metric.sqlExpression;
   }
-  return verboseMap[label] || label;
+  return verboseMap?.[label] || label;
 }
--- a/superset-frontend/packages/superset-ui-core/src/query/getMetricLabel.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/getMetricLabel.ts
@@ -19,17 +19,23 @@
 
 import { QueryFormMetric, isSavedMetric, isAdhocMetricSimple } from './types';
 
-export default function getMetricLabel(metric: QueryFormMetric): string {
+export default function getMetricLabel(
+  metric: QueryFormMetric,
+  index?: number,
+  queryFormMetrics?: QueryFormMetric[],
+  verboseMap: Record<string, string> = {},
+): string {
+  let label = '';
   if (isSavedMetric(metric)) {
-    return metric;
-  }
-  if (metric.label) {
-    return metric.label;
-  }
-  if (isAdhocMetricSimple(metric)) {
-    return `${metric.aggregate}(${
+    label = metric;
+  } else if (metric.label) {
+    ({ label } = metric);
+  } else if (isAdhocMetricSimple(metric)) {
+    label = `${metric.aggregate}(${
       metric.column.columnName || metric.column.column_name
     })`;
+  } else {
+    label = metric.sqlExpression;
   }
-  return metric.sqlExpression;
+  return verboseMap[label] || label;
 }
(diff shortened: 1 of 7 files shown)


superset-frontend/webpack.config.js
```diff
@@ -568,7 +568,7 @@ if (isDevMode) {
         warnings: false,
         runtimeErrors: error => !/ResizeObserver/.test(error.message),
       },
-      logging: 'error',
+      logging: 'verbose',
```
michael-s-molina, line 571: Is this intended?

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -290,12 +291,24 @@ export default function transformProps(
   const showValueIndexesB = extractShowValueIndexes(rawSeriesB, {
     stack,
   });
+
+  const metricMapFunc = (metric: string | AdhocMetric) => {
+    if (metric.hasOwnProperty('label')) {
+      return (metric as AdhocMetric).label;
```
kgabryje, line 297: I think we already have a util `getMetricLabel` for that

eschutho: Ah, yes. How should that function work with the verboseMap, though?

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -290,12 +292,20 @@ export default function transformProps(
   const showValueIndexesB = extractShowValueIndexes(rawSeriesB, {
     stack,
   });
+
+  const metricsLabels = metrics
+    .map(metric => getMetricLabel(metric, undefined, undefined, verboseMap))
+    .filter((label): label is string => label !== undefined);
+  const metricsLabelsB = metricsB.map((metric: QueryFormMetric) =>
+    getMetricLabel(metric, undefined, undefined, verboseMap),
```
eschutho, line 300: @kgabryje lmk what you think. I don't like having to pass a bunch of undefined values, but it felt safer here than changing the args to kwargs (a js object). A bunch of the map functions are passing values that were unused in getMetricLabel.

superset-frontend/packages/superset-ui-core/src/query/getMetricLabel.ts
```diff
@@ -19,17 +19,23 @@
 
 import { QueryFormMetric, isSavedMetric, isAdhocMetricSimple } from './types';
 
-export default function getMetricLabel(metric: QueryFormMetric): string {
+export default function getMetricLabel(
+  metric: QueryFormMetric,
+  index?: number,
+  queryFormMetrics?: QueryFormMetric[],
+  verboseMap: Record<string, string> = {},
```
geido, line 26: Nit. Does it need to be defaulting to an empty object or can make this also optional and check for both undefined and empty?

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -291,12 +292,20 @@ export default function transformProps(
   const showValueIndexesB = extractShowValueIndexes(rawSeriesB, {
     stack,
   });
+
+  const metricsLabels = metrics
+    .map(metric => getMetricLabel(metric, undefined, undefined, verboseMap))
```
geido, line 297: Can we move the `verboseMap` before the other two optional parameters in the function so we don't have to pass the `undefined`s here at all? I don't see many other places mapping these values, so it might be safe to just revise the params positions, but if you feel that might introduce issues, feel free to skip this comment

geido (approved):
A few nits

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
eschutho: Since the map function is always going to pass the index and array as second and third args, I can't move them around, but I was thinking of changing `getMetricLabel` to accept an object, but there are over 40 places in the codebase where this map function is used, and in some cases, it's passed in directly like `map(getMetricLabel)` instead of as an anonymous function. It just seemed less risky to do it this way, but I agree that it's not pretty.

PR #31340, opened 2024-12-08
fix(ci): change ephemeral env to use github labels instead of comments

dpgaspar:
### SUMMARY

Use labels instead of issue comment to spin up new environments, this is a safer approach, labels can only be added to PRs by users with write access to the repo. Also using labels it's possible to get the `pushed_at` to sanity check what we are building.

Use `testenv-up` labels to spin ephemeral envs

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


.github/workflows/ephemeral-env.yml
```diff
@@ -27,8 +27,28 @@ jobs:
...
+              pull_number: ${{ github.event.inputs.issue_number || github.event.issue.number }},
+          };
+          core.info(`Getting PR #${request.pull_number} from ${request.owner}/${request.repo}`);
+          const pr = await github.rest.pulls.get(request);
+          return pr.data;
+
+    - name: Debug
+      id: get-sha
+      run: |
+        echo "sha=${{ fromJSON(steps.get-pr-info.outputs.result).head.sha }}" >> $GITHUB_OUTPUT
```
RedYetiDev, line 50: You should exit early if the PR was updated after the comment. See https://github.com/apache/camel/blob/66fbdcd2c71a6588bacd7b3e0d2a03128c0cd069/.github/workflows/pr-comment.yml#L55-L57

```suggestion
    - name: Debug
      id: get-sha
      env:
      	COMMENT_AT: ${{ github.event.comment.created_at }}
      	PUSHED_AT: ${{ fromJSON(steps.get-pr-info.outputs.result).pushed_at }}
      run: |
        if [[ $(date -d "$PUSHED_AT" +%s) -gt $(date -d "$COMMENT_AT" +%s) ]]; then
            exit 1
        fi
        echo "sha=${{ fromJSON(steps.get-pr-info.outputs.result).head.sha }}" >> $GITHUB_OUTPUT
```

pwntester: Unfortunately, this is not useful as `pushed_at` field is deprecated and the value will be nil.

RedYetiDev: I don't see any evidence of that in the documentation. Can you point me to your source?

pwntester: If untrusted code needs to be run, I would move the IssueOps workflow to a Label gate (pull_request + label). There is no way to get a trustful commit date.

raboof: just to tie things together: `pushed_at` from https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request is slightly different from `Commit.pushedDate`, but I suppose the they're ultimately populated from the same source, so both are deprecated/going away? This wasn't obvious to me before but would make sense. I didn't see anything in https://docs.github.com/en/rest/about-the-rest-api/breaking-changes but I guess it just somewhat follows from https://docs.github.com/en/graphql/overview/breaking-changes .

dpgaspar: @pwntester @raboof sorry for the delay here. I've updated to workflow to use PR labels, would be great if you can review it again

.github/workflows/ephemeral-env.yml
```diff
@@ -1,145 +1,121 @@
...
+        id: get-sha
+        env:
+          COMMENT_AT: ${{ github.event.pull_request.updated_at }}
+          PUSHED_AT: ${{ fromJSON(steps.get-pr-info.outputs.result).pushed_at }}
+        run: |
+          if [[ $(date -d "$PUSHED_AT" +%s) -gt $(date -d "$COMMENT_AT" +%s) ]]; then
+           echo "Commit was pushed after the PR was updated."
+           exit 1
+          fi
+          echo "sha=${{ fromJSON(steps.get-pr-info.outputs.result).head.sha }}" >> $GITHUB_OUTPUT
```
avivkeller, line 70: If you're verifying the label, there is no need for this extra step. Just use github.event.pull_request.head.sha

dpgaspar: unsure if `github.event.pull_request.updated_at` could be greater then the time of the last push

avivkeller: No, it’s the SHA at the time of label, which is the most secure

mistercrunch: `GITHUB_RUN_ID` can we useful at times too, not sure if it's better than SHA here, but would allow concurrent runs on the same SHA (do we want that? how do we want to handle potential SHA collisions?)

mistercrunch: I see we have the `concurrency` GHA thing set up properly, meaning a new request will kill the previous ones, which I believe is what we want. Using the `GITHUB_RUN_ID` as a unique id on the AWS-side could help segment/isolate/clarify the runs.

mistercrunch: I guess we need something deterministic that we can use to kill previous instances on PR close or unlabelling on the AWS side, so `GITHUB_RUN_ID` isn't the best for that ...

.github/workflows/ephemeral-env.yml
```diff
@@ -1,145 +1,152 @@
...
-# Example manual trigger:  gh workflow run ephemeral-env.yml --ref fix_ephemerals  --field comment_body="/testenv up" --field issue_number=666
+# Example manual trigger:
+# gh workflow run ephemeral-env.yml --ref fix_ephemerals --field label_name="testenv-up" --field issue_number=666
 
 on:
-  issue_comment:
-    types: [created]
+  pull_request:
+    types:
+      - labeled
```
avivkeller, line 9: Keep in mind, pull_request_target must be used if you want to use secrets.

.github/workflows/ephemeral-env.yml
mistercrunch: Side note, I created `supersetbot` to make it easier to develop/test/deploy things outside of GHA. We're pretty deep already here, but we could move the bulk of the logic to `supersetbot ephemeral deploy {image_id}`, `supersetbot ephemeral stop {image_id}` ...

.github/workflows/ephemeral-env.yml
```diff
@@ -1,145 +1,152 @@
...
+        id: get-sha
+        env:
+          COMMENT_AT: ${{ github.event.pull_request.updated_at }}
+          PUSHED_AT: ${{ fromJSON(steps.get-pr-info.outputs.result).pushed_at }}
+        run: |
+          if [[ $(date -d "$PUSHED_AT" +%s) -gt $(date -d "$COMMENT_AT" +%s) ]]; then
+           echo "Commit was pushed after the PR was updated."
+           exit 1
+          fi
+          echo "sha=${{ fromJSON(steps.get-pr-info.outputs.result).head.sha }}" >> $GITHUB_OUTPUT
```
avivkeller, line 87: as I mentioned in my other review, this entire segment is not needed when not triggering on a dispatch, and could result in a race condition (albeit harmless unless the trigger is pull_request)

I suggest ONLY fetching the SHA via this method when executing on dispatch rather than on PR.

avivkeller (commented):
I'm sorry if my comments are confusing. If you'd like, I can PR into your branch with my suggested changes

mistercrunch (approved):
LGTM! - and they envs are broken so this can only help. 

I did some complementary work here https://github.com/apache/superset/pull/32017

.github/workflows/ephemeral-env.yml
dpgaspar: concurrency seems fine to me (pushed a fix). it's unique at the PR level, so label events for new commits will cancel previous runs.

+1 on using supersetbot ephemeral deploy

avivkeller (approved):

dpgaspar:
@mistercrunch @avivkeller thank you for the reviews!

PR #30956, opened 2024-11-18
fix: added missing pod labels for init job

glothriel:
### SUMMARY
Added missing pod labels for init job. Without them it's impossible to dynamically set labels in environments with Network Policies engaged.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
skip

### TESTING INSTRUCTIONS
Running helm template will suffice.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #31346, opened 2024-12-09
fix(helm): Add common labels for init job pods

LipuFei:
Add two common labels for the init-job pods, similar to other deployments.

In our environment, we need to define NetworkPolicies to control network traffic. Without labels on the init-job pods, I cannot whitelist any network traffic for them.


villebro:
@LipuFei thanks for the PR, but we just merged #30956 that fixes this. So I'm closing this PR, as this is already done.

PR #31349, opened 2024-12-09
feat: improve pdf export quality

tahvane1:
This PR uses img2pdf to export charts to pdf. Current solutions uses compression in such way that pdf is grainy.
Using img2pdf removes this issue mostly.

### SUMMARY
Current pdf quality is poor and this improves it significantly

### TESTING INSTRUCTIONS
img2pdf needs to be added to docker images.

### DIFF
--- a/superset/utils/pdf.py
+++ b/superset/utils/pdf.py
@@ -16,7 +16,7 @@
 # under the License.
 
 import logging
-from io import BytesIO
+import img2pdf
 
 from superset.commands.report.exceptions import ReportSchedulePdfFailedError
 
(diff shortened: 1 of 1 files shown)


villebro (changes requested):
This PR apparently adds a new requirement. I don't oppose this change, but I would propose the following:
- Add `img2pdf` to one of the `.in` files in `requirements/`. Based on this change it would need to go into `base.in`. Also do check that the license is Apache 2.0 compliant.
- Add before/after screenshots to show how quality changes.
- Add info about how the pdf file size changed before/after. Lossless compression is great, but can bloat the attachment size.

tahvane1:
img2pdf is using GNU Lesser General Public License v3 and it seems that it is not accepted in ASF projects. So need to look for alternative approach then.

villebro:
> img2pdf is using GNU Lesser General Public License v3 and it seems that it is not accepted in ASF projects. So need to look for alternative approach then.

Dang, that's a real shame! 🙁

rusackas:
Do we know if the low resolution is caused by the capture method (`Image.open(BytesIO(snap))`) or by the output method (`images[0].save(...)`)? Apparently if it's the latter, you can add a `resolution=300.0` parameter.

tahvane1:
Resolution does not fix issue. I can get similar quality by setting quality=100 but this results to a pdf that has double the size what img2pdf produces. I will continue my testings...

villebro:
FYI I converted this to draft. @tahvane1 please feel free to change it back to non-draft if you find a new ASF-compliant way to fix this issue.

PR #31363, opened 2024-12-10
fix: clean up permissions errors

eschutho:
### SUMMARY

Testing to fix the ephemeral builds
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -163,6 +163,7 @@ RUN if [ "$BUILD_TRANSLATIONS" = "true" ]; then \
 
 # Add server run script
 COPY --chmod=755 ./docker/run-server.sh /usr/bin/
+RUN chown superset:superset /usr/bin/run-server.sh
 
 # Set user and healthcheck
 USER superset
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):

eschutho:
That first commit worked but not the last one. I need to tweak the env var directory lookup a bit.

eschutho:
This version is working now if we want to merge. I can iterate on finding a way to use the environment variables so that we don't need to hard code the app path.

Dockerfile
```diff
@@ -121,6 +121,8 @@ RUN --mount=type=bind,source=./docker,target=/docker \
         libecpg-dev \
         libldap2-dev \
     && touch superset/static/version_info.json \
+    && chown -R superset:superset /app \
+    && chmod -R 775 /app \
```
mistercrunch, line 125: wondering if these 2 commands would be belong right before the `USER superset` command, and if we do so we might be able to simplify the `COPY` commands and other chown/chmod throughout the Dockerfile

mistercrunch: https://github.com/apache/superset/pull/31385

PR #31374, opened 2024-12-10
fix(Dashboard): Sync color configuration via dedicated endpoint

geido:
### SUMMARY
This PR introduces a new endpoint which purpose is to store color configurations for a dashboard. Dashboard color configurations might need syncing and persistence in the backend even when the user takes no action. For instance, new data might appear on the dashboard that needs to be synced with the color maps and persisted. Previously, these changes would be saved using the dashboard PUT endpoint, causing that the dashboard last modified date would show up as updated even though no change was made from the user perspective, causing confusion. Also, issues have surfaced where metadata that wasn't meant to be persisted yet would go along with the updated color metadata, causing unexpected behaviors.

N.A.

### TESTING INSTRUCTIONS
1. Open a Dashboard
2. Make sure new data appeared (new dimensions or metrics)
3. Observe how a network request is made to store the color metadata when leaving the dashboard
4. Make sure the last modified date of the Dashboard hasn't changed

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/actions/dashboardState.js
+++ b/superset-frontend/src/dashboard/actions/dashboardState.js
@@ -73,8 +73,9 @@ import {
   isLabelsColorMapSynced,
   getColorSchemeDomain,
   getColorNamespace,
-  getLabelsColorMapEntries,
+  getFreshLabelsColorMapEntries,
   getFreshSharedLabels,
+  getDynamicLabelsColors,
 } from '../../utils/colorScheme';
 
 export const SET_UNSAVED_CHANGES = 'SET_UNSAVED_CHANGES';
@@ -253,15 +254,18 @@ export function setDashboardSharedLabelsColorsSynced() {
   return { type: SET_DASHBOARD_SHARED_LABELS_COLORS_SYNCED };
 }
 
-export const setDashboardMetadata = updatedMetadata => async dispatch => {
-  dispatch(
-    dashboardInfoChanged({
-      metadata: {
-        ...updatedMetadata,
-      },
-    }),
-  );
-};
+export const setDashboardMetadata =
+  updatedMetadata => async (dispatch, getState) => {
+    const { dashboardInfo } = getState();
+    dispatch(
+      dashboardInfoChanged({
+        metadata: {
+          ...(dashboardInfo?.metadata || {}),
+          ...updatedMetadata,
+        },
+      }),
+    );
+  };
 
 export function saveDashboardRequest(data, id, saveType) {
   return (dispatch, getState) => {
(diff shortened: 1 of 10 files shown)


superset/dashboards/api.py
```diff
@@ -767,6 +773,88 @@ def put_filters(self, pk: int) -> Response:
...
+              $ref: '#/components/responses/403'
+            404:
+              $ref: '#/components/responses/404'
+            422:
+              $ref: '#/components/responses/422'
+            500:
+              $ref: '#/components/responses/500'
+        """
+        try:
+            item = self.update_colors_model_schema.load(request.json, partial=True)
```
Antonio-RiveroMartnez, line 832: Since we can send just partial updates, isn't better to use a PATCH call instead?

superset/commands/dashboard/update.py
```diff
@@ -202,3 +203,29 @@ def run(self) -> Model:
         )
 
         return configuration
+
+
+class UpdateDashboardColorsConfigCommand(UpdateDashboardCommand):
+    @transaction(
```
Antonio-RiveroMartnez, line 209: Isn't this decorator supposed to be used on the `def run`? hence no need to disable the pylint rule about transaction

superset/dashboards/api.py
geido: PATCH isn't widely supported in Superset at the moment. I could not find a single endpoint using PATCH atm

superset/commands/dashboard/update.py
geido: You are right that this should be decorating the `run`. Just moved it. As for disabling the rule, that's to commit the modified date back to its original state when the `mark_updated` parameter is set to `False`. In all other cases, it should commit normally using the `@transaction`.

superset-frontend/src/dashboard/actions/dashboardState.js
```diff
@@ -867,23 +896,24 @@ export const ensureSyncedSharedLabelsColors =
...
       const sharedLabelsColors = enforceSharedLabelsColorsArray(
         metadata.shared_label_colors,
       );
       const freshLabelsColors = getFreshSharedLabels(
         forceFresh ? [] : sharedLabelsColors,
       );
       const isSharedLabelsColorsSynced = isEqual(
-        sharedLabelsColors,
-        freshLabelsColors,
+        sortBy(sharedLabelsColors),
```
kgabryje, line 906: why `sortBy` instead of just `sort`?

superset-frontend/src/utils/colorScheme.ts
```diff
@@ -91,21 +93,40 @@ export const getSharedLabelsColorMapEntries = (
...
+ *
+ * @param labelsColorMap - the labels color map
+ * @param customLabelsColor - the custom label colors in label_colors field
+ * @returns all color entries except custom label colors
+ */
+export const getDynamicLabelsColors = (
+  fullLabelsColors: Record<string, string>,
+  customLabelsColor: Record<string, string> = {},
+): Record<string, string> => {
+  const labelsColors = { ...fullLabelsColors };
```
kgabryje, line 121:
```suggestion
  return omit(fullLabelsColors, Object.keys(customLabelsColor))
```

Antonio-RiveroMartnez (approved):
lgmt after addressing the comments

kgabryje (approved):

PR #31379, opened 2024-12-10
fix: docker-compose up ModuleNotFoundError: {...} superset

mistercrunch:
As reported [here](https://github.com/apache/superset/discussions/31357) our Dockerfile is currently broken. Somehow the copy command was copying the `superset/superset` folder instead of `superset/` in some setups.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -191,7 +191,7 @@ RUN --mount=type=bind,source=./docker,target=/docker \
 
 # Install Playwright and its dependencies
 RUN --mount=type=cache,target=/root/.cache/pip \
-    pip install --system playwright \
+    pip install playwright \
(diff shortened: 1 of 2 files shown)


rusackas:
I have all the clueless questions about how it broke, why it worked before, and why more people aren't running into this, etc.

This looks like it'll copy in a lot more stuff, will that lead to any bloat/slowdown?

Otherwise this looks very stampable to me when it's ready for review.

mistercrunch:
from my troubleshoot 
- uv has some quirks around switching users
- docker cache works in mysterious ways, somehow my setup was working with what I assume was older docker cache that should have been validated (?)

Rolling back using `uv` in docker, tested it works from clean cache

PR #31386, opened 2024-12-10
feat(gha): various docker / docker-compose build improvements

mistercrunch:
Spent time looking at every single step of the `docker` / `docker compose` build process, using `dive` and `docker stats` to inspect every layer/stage/builds. Found a lot of things to improve in and out of docker.

Improvements:
- after building the container, we now fire up `docker-compose up superset-init` to make sure we can actually run the container and perform init-related operations (loading examples, ...), this will catch issues we didn't catch before, whether it's related to the container, docker-compose or the app itself
- during GHA builds, pass ARGs to docker command to NOT include a headless browser, keeping image sizes smaller, builds faster at least in the context of GHA and dev env.
- install the latest docker compose plugin (our docker-compose files require a newer version than what's in the default ubuntu repos)
- print some docker stats after the GHA build, namely `docker images $IMAGEID` `docker history $IMAGEID`, so it's easy to look at stage sizes in the GHA logs, makes it easier to introspect what's happening, even compare different branches/sha more easily just by looking at logs
- docs: documenting new ENV var flags in our docker-compose docs, making things more discoverable/configurable
- calling `uv pip` instead of plain `pip` in more (all?) places, speeding things up
- on PR push, running a matrix of both `lean` and `dev` (instead of just `dev`) and make sure that they both complete. Added something to make sure postgres driver is installed at load (`up`) time
- removing `playwright` from `dev` requirements as it's pretty heavy and don't need it for most development use cases, keeping builds faster/leaner
- found things to add to `.dockerignore` that were bloating the image
- mounting some cache directories when using npm/webpack (preventing 3-4GB of cache to make it inside the frontend build stage), this should greatly improve memory usage while building the frontend
- removed .map files from output when running `npm run build`, slashing the size of the static assets by half


.dockerignore
```diff
@@ -42,6 +42,7 @@ docs/
 install/
 superset-frontend/cypress-base/
 superset-frontend/coverage/
+superset-frontend/.temp_cache/
```
mistercrunch, line 45: this is a big win, that folder (storing the webpack cache) gets real big, like GBs, not bringing it in the image really improve the frontend build stage

.github/workflows/docker.yml
```diff
@@ -15,20 +15,20 @@ concurrency:
...
   setup_matrix:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       matrix_config: ${{ steps.set_matrix.outputs.matrix_config }}
     steps:
       - id: set_matrix
         run: |
-          MATRIX_CONFIG=$(if [ "${{ github.event_name }}" == "pull_request" ]; then echo '["dev"]'; else echo '["dev", "lean", "py310", "websocket", "dockerize", "py311"]'; fi)
+          MATRIX_CONFIG=$(if [ "${{ github.event_name }}" == "pull_request" ]; then echo '["dev", "lean"]'; else echo '["dev", "lean", "py310", "websocket", "dockerize", "py311"]'; fi)
```
mistercrunch, line 24: build matrix now builds `dev` + `lean` on PR push

eschutho (approved):
LGTM

PR #31407, opened 2024-12-11
fix: Big Number side cut fixed

### LINKED ISSUE #31010: Big Number + Big Number with Trendline get cut off on dashboards
### Bug description

Repro: 
1. Create a Big Number chart or a Big Number with Trendline chart 
2. in the Customize tab, set the font size to Normal, large or huge (this bug seems harder to repro when the font size is Small or Tiny, but it might still be possible) 
3. Save the chart and add it to a dashboard 

Currently, the numbers on the right get cut off easily. Typically, giving the chart a lot more horizontal space will fix the issue, but the font should resize correctly to not be cut off if there's not enough space. 

### Screenshots/recordings

image
image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

fardin-developer:
### SUMMARY
Fixes #31010
### BEFORE/AFTER SCREENSHOTS
Before
Screenshot 2024-12-11 at 10 35 47 PM

After
[
Screenshot 2024-12-11 at 10 36 07 PM
](url)

Solution

- Adjusted the chart rendering logic to dynamically resize the font to fit the available space.
- Ensured compatibility with all font size options (Small, Tiny, Normal, Large, Huge).


rusackas:
Updated the PR description to correctly link to and close the issue.

kgabryje:
@fardin-developer thanks for the contribution! Could you provide some repro steps? TBH I haven't seen that bug in quite some time

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -174,9 +174,12 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
       <div
         className="header-line"
         style={{
+          display: 'flex',
+          alignItems: 'center',
           fontSize,
-          height: maxHeight,
+          height: 'auto',
           color: numberColor,
+          marginBottom: '10px',
```
kgabryje, line 182: is this additional margin needed here? If so, can we use `theme.gridUnit * 2` instead?

fardin-developer: Yes, you are right we actually don't need  marginBottom: '10px'

fardin-developer: Yes  @kgabryje we need  margin-bottom here to give some space between header and subheader. You are right we can use `theme.gridUnit * 2`
I've made the change.

fardin-developer:
To repro the same bug:
Please follow the steps

1. Go to Dashboards
2. add `+ Dashboard`
3. create a new chart
4. `search big number`
5. Add big number
6. Add dataset (search sales and add - cleaned_sales_data)
7. from left side Colums drag sales and drop to Metric
8. Simply save it (sum)
9. Go to customize and increase the font to huge
10. Format .3% or ,d like that which gives more number in header
11. save and goto dashboard u can see the bug --- Thank You
I will also share a video

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
fardin-developer: @kgabryje 
Please review my pull request and suggest any additional changes if needed!

geido (commented):
@fardin-developer Thanks for this. Testing locally works. I have just some question about the prop being removed.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -120,7 +120,7 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
         className="kicker"
         style={{
           fontSize,
-          height: maxHeight,
+          height: 'auto',
```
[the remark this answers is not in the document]
fardin-developer: The maxHeight is used here for two purposes:

- To calculate the fontSize dynamically, ensuring the text fits well.
- To set the height of the div elements, like .header-line and .subheader-line.

Since these div elements only contain text, their height could automatically adjust based on the calculated fontSize. However, maxHeight is still important because it helps determine what the fontSize should be.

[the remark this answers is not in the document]
fardin-developer: https://painted-tune-041.notion.site/Results-1664127ab6d4806dbab9f3316356e257
Here I have uploaded all the changes and results related to updating height: maxHeight to height: auto.
Let me know if you’d like more details or have any suggestions.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -204,7 +206,7 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
       document.body.append(container);
       fontSize = computeMaxFontSize({
         text,
-        maxWidth: width,
+        maxWidth: width * 0.9, // max with reduced
```
geido, line 209:
```suggestion
        maxWidth: width * 0.9, // max width reduced
```

geido (approved):

PR #31413, opened 2024-12-12
feat: add date format to the email subject

US579:
### SUMMARY

Problem
If the subject remains the same, all scheduled reports will be delivered in the same email thread, requiring the recipient to scroll back to the original thread to find the latest email.

However, if a date is included in the email subject, the emails will not be grouped in the same thread. To address this, I propose adding a feature that allows users to specify a date format in the email subject, enabling the email client to correctly thread the emails and display them in a more organized manner.

Before
Screenshot 2024-12-12 at 11 57 06 am

After
Screenshot 2024-12-12 at 11 57 30 am

Screenshot 2024-12-12 at 12 01 34 pm

### How to use it

Set the `DATE_FORMAT_IN_EMAIL_SUBJECT=True` to allow users to include a date format in the email subject, which will automatically be replaced with the corresponding actual date. 

available date format  https://strftime.org/

Example

In this example i defined Year-Day-Minutes

Year - %Y
Day - %D
Minutes - %M

Screenshot 2024-12-12 at 11 59 55 am

Output

Screenshot 2024-12-13 at 11 10 39 am

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: DATE_FORMAT_IN_EMAIL_SUBJECT (by default is disabled)
- [ ] Changes UI


oOhyeahh (approved):

superset/config.py
```diff
@@ -563,6 +563,8 @@ class D3TimeFormat(TypedDict, total=False):
     # If on, you'll want to add "https://avatars.slack-edge.com" to the list of allowed
     # domains in your TALISMAN_CONFIG
     "SLACK_ENABLE_AVATARS": False,
+    # Allow users to optionally specify date formats in email subjects, which will be parsed if enabled.
+    "DATE_FORMAT_IN_EMAIL_SUBJECT": False,
```
sadpandajoe, line 567: Can you mention this new feature flag in UPDATING.md

rusackas (dismissed):
The flag should also be documented/categorized in [FEATURE_FLAGS.md](https://github.com/apache/superset/blob/master/RESOURCES/FEATURE_FLAGS.md)

superset/config.py
US579: thanks for reviewing! i have updated

US579:
> @US579 can you also add some documentation on how this is used? It won't be clear for users who don't know about this feature that they can start doing this post merge.

Could you point me where i could add this doc, currently i just added MR link (which has description for how to use it)  next to the FF

sadpandajoe:
> > @US579 can you also add some documentation on how this is used? It won't be clear for users who don't know about this feature that they can start doing this post merge.
> 
> hey @sadpandajoe Could you point me where is the bast place to add this doc, currently i just added [here](https://github.com/apache/superset/pull/31413#diff-a35d90f22c0fe6afe6e6999c4291dc377ec9177a3aae2d1c194fe47f34433fadR32)

@US579 great question. I would say you can either put it in the alerts and reports sections: https://github.com/apache/superset/blob/4ff9aac1fae2d88fd7667317c706256d3cfcb4f4/docs/docs/configuration/alerts-reports.mdx or in the faq https://github.com/apache/superset/blob/4ff9aac1fae2d88fd7667317c706256d3cfcb4f4/docs/docs/faq.mdx

RESOURCES/FEATURE_FLAGS.md
```diff
@@ -46,7 +46,8 @@ These features are **finished** but currently being tested. They are usable, but
 - CONFIRM_DASHBOARD_DIFF
 - DRILL_TO_DETAIL
 - DYNAMIC_PLUGINS
-- ENABLE_SUPERSET_META_DB: [(docs)](https://superset.apache.org/docs/configuration/databases/#querying-across-databases)
+- DATE_FORMAT_IN_EMAIL_SUBJECT: [(docs)](https://superset.apache.org/docs/configuration/databases/#querying-across-databases)
+- ENABLE_SUPERSET_META_DB: [(docs)](https://github.com/apache/superset/pull/31413)
```
rusackas: It looks like this docs link and the one above it are swapped. 

Also, it would nice to add a docs entry rather than linking to the PR. You can run the docs locally and just add some markdown. Docs on how to do that are [here](https://superset.apache.org/docs/contributing/howtos#contributing-to-documentation).

mistercrunch:
Totally get the use case. Hey I'm wondering if it may make sense to support more templating in emails (subject and body) in general to tackle this use case but similar ones as well. I know jinja isn't safe for untrusted (user-provided) content, maybe we could use some pybars or something untrusted safe to allow for bringing arbitrary context in arbitrary places in the subject/body.

RESOURCES/FEATURE_FLAGS.md
US579: thanks for pointing me to the right direction! i have added an entry and updated the link

US579:
hey @rusackas thanks for reviewing! any chance we could merge this?

US579:
> @US579 looks like a rebase is needed to fix the e2e test that is failing.

thanks, pipeline has been fixed, let me know if there is anything bloc

US579:
@sadpandajoe any choice we could merge this? please let me know if there is any pending

sadpandajoe (approved):
LGTM

PR #31414, opened 2024-12-12
fix(Pivot Table): Fix column width to respect currency config

### LINKED ISSUE #31377: Pivot Table doesn't consider metric currency for column width
## Screenshot
imagen
imagen

## Description
When adding metrics with currency to a pivot table chart, the column width is calculated based on the size of the number, but it doesn't account for the additional character of the currency symbol, forcing the symbol to be truncated into an additional line, which increases the rows' height.

## Design input
I suggest to increase columns' width if the metric that it represents has any currency at all.

Vitor-Avila:
### SUMMARY
Fixes https://github.com/apache/superset/issues/31377.

**Before**
image

**After**
image

### TESTING INSTRUCTIONS
1. Create a Pivot table using the `cleaned_sales_data` dataset.
2. Create a metric at the dataset level and apply a currency config.
3. Set `country` as a **column**, `product_line` as a **row** and use the metric created in the previous step.
4. Validate that the column width respects the currency config. 

### ADDITIONAL INFORMATION
- [x] Has associated issue: https://github.com/apache/superset/issues/31377
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/superset-frontend/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -51,6 +51,7 @@ const Styles = styled.div<PivotTableStylesProps>`
       width: ${
         typeof width === 'string' ? parseInt(width, 10) : width - margin * 2
       }px;
+      white-space: nowrap;
  `}
 `;
 


kgabryje (approved):
Looks good!

martin-raymond:
we saw a regression on all table pivot due to this fix.

The texts in cells are no longer cut if too long, and all tables with long texts in cells got horizontal scrolls bars to death

PR #31421, opened 2024-12-12
refactor(Shared_url_query): Fix shared  query URL access for SQL Lab users.

### LINKED ISSUE #31367: Unable to share saved SQL queries - user gets an error wen trying to load the query
## Description
Users are unable to share saved SQL queries to other users. The owner of a saved query can access it via the shared URL, but any other user attempting to access the same URL receives an error message: “The query couldn’t be loaded."

## How to reproduce the bug
1. Log in and save a query in SQL Lab.
2. Copy the query URL using the "Copy Query URL" option.
3. Share the URL with a different user.
4. Have the other user try to load the shared query URL.
5. Observe that the second user receives the error message: “The query couldn’t be loaded.”

## Expected results
Any user with access to SQL Lab should be able to load and view saved SQL queries via a shared URL.

## Actual results
Only the creator of the query can load the query from the shared URL. Other users receive an error message stating: "The query couldn’t be loaded."

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

LevisNgigi:
### SUMMARY
This update fixes the copy sharing to use the newly introduced permalink.

Fixes #31367

fix(sql-lab): fix the copy url to use the permalink api and allow query sharing for users with appropriate permissions.

### TESTING INSTRUCTIONS
1. Log in and save a query in SQL Lab.
2. Copy the query URL using the "Copy Query URL" option.
3. Share the URL with a different user (one with the `sqllab` role and access to the dataset).
4. Have the other user try to load the shared query URL.
5. Observe that the second user with sqllab_role can successfully load the shared query.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #31367
- [ ] Required feature flags:
- [ ] Changes UI


villebro:
Tagging @yousoph @geido as you were active on the original issue. Has there has been discussion about making queries shareable via permalinks? Was this considered as a solution? As @rusackas commented, exposing saved queries to anyone with a link to it can be problematic, as per the current design they are private objects. Implementing this is of course more work, but is aligned with how sharing is done in Explore and the Dashboard views.

LevisNgigi:
> Tagging @yousoph @geido as you were active on the original issue. Has there has been discussion about making queries shareable via permalinks? Was this considered as a solution? As @rusackas commented, exposing saved queries to anyone with a link to it can be problematic, as per the current design they are private objects. Implementing this is of course more work, but is aligned with how sharing is done in Explore and the Dashboard views.

Hey, just a clarification.Not anyone with the link can view the query.Any user with public, Gamma, or Alpha roles cannot view the saved query link. Access is strictly limited to users who have been explicitly granted permission by the admin under the SQL Lab role. This ensures that only those with appropriate SQL Lab permissions and dataset access are able to load and view the query. All other users, including those with general roles like Gamma or Alpha, will receive an "Access Denied" message when attempting to access the link.

villebro:
Thanks for the clarification @LevisNgigi 👍

LevisNgigi:
> Thanks for the clarification @LevisNgigi 👍

@villebro Much Welcome.

justinpark:
@LevisNgigi Has there has been discussion about making queries shareable via permalinks (using "copy link")?

LevisNgigi:
> @LevisNgigi Has there has been discussion about making queries shareable via permalinks (using "copy link")?

Hey @justinpark I do not think so. I'm not aware of any discussion in reference to the above.

geido:
Hey @LevisNgigi I am a bit confused about the latest changes where it is just showing a newly introduced endpoint. Would you clarify the approach here? Thank you!

LevisNgigi:
> Hey @LevisNgigi I am a bit confused about the latest changes where it is just showing a newly introduced endpoint. Would you clarify the approach here? Thank you!

Hey @geido This endpoint already existed in the project "/api/v1/query/{pk}". I added custom implementation on the endpoint to add ownership validation to ensure only the query owner or authorized users can access it.Those with "can_read", "SavedQuery" permission. The custom implementation I added addresses this gap,by ensuring that users with sqllab role can view shared query via the url.

superset/queries/saved_queries/api.py
```diff
@@ -196,6 +199,102 @@ def pre_add(self, item: SavedQuery) -> None:
...
+            500:
+              $ref: '#/components/responses/500'
+        """
+        try:
+            saved_query = db.session.query(SavedQuery).get(pk)
+            if not saved_query:
+                return self.response_404()
+
+            is_owner = saved_query.user_id == g.user.id
+            has_access = security_manager.can_access("can_read", "SavedQuery")
```
dpgaspar, line 257: The `can_read` on `SavedQuery` is already guaranteed by the `@protect` decorator, so it seem redundant to check if the saved query belongs to the user, since it will grant access to all queries to all users.
This does not seem like the correct approach, currently SavedQueries are not shared on Superset, if we want to add this functionality, we can share links using a UUID or wait until we have use group entities and expose queries to groups

LevisNgigi: Thank you for the review and feedback.To clarify, the above implementation does not automatically grant all users access to all saved queries. Instead, it ensures that only users with can_read permissions for SavedQuery can view the query(via a shared url)which is a default role for sqllab. This means that access is already restricted, and not all users will have visibility unless they explicitly have the required permissions.The ownership check was included as an additional layer of security, but I understand your concern about redundancy given the @protect decorator. As you rightly pointed out, one potential vulnerability is the predictability of query IDs in URLs (e.g., http://localhost:9000/sqllab?savedQueryId=3), which makes it possible to guess query IDs. This could expose queries to unauthorized users who have can_read permissions but should not have access to specific queries. Implementing UUIDs for query IDs would mitigate this issue and improve security.Should this be left until query sharing is added in Superset?

LevisNgigi: > @LevisNgigi how hard would it be to implement the UUID approach now?
Implementing the UUID approach would require running a migration script to update existing query IDs to UUIDs in the database while maintaining backward compatibility. Additionally, the logic to verify if the user has the sqllab role would need to be integrated to ensure the correct permissions are applied.I think this adds some complexity especially with the migrations. Unless there is a different approach?

legiondean:
@LevisNgigi do you know if there's any workaround setup to make the sharing work before this fix released?
Or any chance the fix could be released soon? Thanks!
The issue is pretty annoying when come to team collaboration.

LevisNgigi:
> @LevisNgigi do you know if there's any workaround setup to make the sharing work before this fix released? Or any chance the fix could be released soon? Thanks! The issue is pretty annoying when come to team collaboration.

@legiondean Thanks for bringing this up! I am not aware of any workaround as queries were not meant to be shared originally, but I'm actively working on resolving the issue by implementing a UUID-based approach, which will address the predictability of query IDs and make sharing more secure and reliable.Hopefully it will be released and approved soon. I appreciate your patience and understanding while we work on this!

superset-frontend/src/pages/SavedQueryList/index.tsx
```diff
@@ -233,16 +232,31 @@ function SavedQueryList({
...
-          addSuccessToast(t('Link Copied!'));
-        })
-        .catch(() => {
-          addDangerToast(t('Sorry, your browser does not support copying.'));
+    async (savedQuery: SavedQueryObject) => {
+      try {
+        const payload = {
+          dbId: savedQuery.db_id,
+          name: savedQuery.label,
+          schema: savedQuery.schema,
```
betodealmeida, line 240: We need to store the catalog as well, for databases that support it.
```suggestion
          catalog: savedQuery.catalog,
          schema: savedQuery.schema,
```

betodealmeida (changes requested):
This looks very clean! A couple things:

1. Can you test this with a database that supports catalogs? You can use Postgres, but you need to enable "Allow changing catalogs" under Advanced -> Other.
2. Can you add a small test for it?

superset-frontend/src/pages/SavedQueryList/index.tsx
```diff
@@ -233,16 +232,31 @@ function SavedQueryList({
...
+        };
+
+        const response = await SupersetClient.post({
+          endpoint: '/api/v1/sqllab/permalink',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify(payload),
         });
+
+        const { key } = response.json;
+        const permalink = `${window.location.origin}/sqllab/p/${key}`;
```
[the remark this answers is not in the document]
justinpark: +1

justinpark:
LGTM

superset-frontend/src/pages/SavedQueryList/index.tsx
LevisNgigi: Hey, so currently the catalog property does not exist on the SavedQueryObject type [https://github.com/apache/superset/blob/master/superset-frontend/src/views/CRUD/types.ts](Here).

betodealmeida: Ok, let me add it before you merge this, we need it.

LevisNgigi: okay sounds good.

betodealmeida: OK, I tested and it you add `catalog` to `superset-frontend/src/views/CRUD/types.ts` it will be populated automatically by the hook, so you just need to add:

```js
catalog: string | null;
```

to `SavedQueryObject`.

Also, it needs to be added to the Schemas and Types as well, because if a database supports multiple catalogs (like BigQuery or Trino) the query with just the schema is not enough information.

betodealmeida (approved):
Awesome! Thanks for making sure it works with catalogs as well!

drummerwolli:
@sadpandajoe since i see you are [managing](https://lists.apache.org/thread/m659odpm4mmgsnjrnzvc6c5hwy7qz3pf) the release of 4.1.3, any chance to include this fix still there or in 4.1.4? Here in the PR and in the linked issue, this sounds very severe. And we have in our organization with ca. 2000 users now also this Problem after upgrading to 4.1.1 just last week. I dont think this is a feature release (as someone wrote in the issue) but rather a pretty hefty bug. happy to hear your thoughts.

sadpandajoe:
> @sadpandajoe since i see you are [managing](https://lists.apache.org/thread/m659odpm4mmgsnjrnzvc6c5hwy7qz3pf) the release of 4.1.3, any chance to include this fix still there or in 4.1.4? Here in the PR and in the linked issue, this sounds very severe. And we have in our organization with ca. 2000 users now also this Problem after upgrading to 4.1.1 just last week. I dont think this is a feature release (as someone wrote in the issue) but rather a pretty hefty bug. happy to hear your thoughts.

@drummerwolli we don't normally throw refactors into official releases as there could be regression bugs coming out of it. Have you checked out the 5.0rc? This should be part of that official release

PR #31427, opened 2024-12-12
fix(tags): clean up bulk create api and schema

villebro:
### SUMMARY
While working on tags I noticed that the bulk create API spec had multiple issues. This cleans up most of the issues I noticed. The OAS3 schema is also regenerated by running `superset update-api-docs`. However, the whole `tags` resource is pretty funky (including the data model and the related enums), so I would almost suggest doing a big refactor on it at some point.

### BEFORE
The description, request type, and the 200 response are wrong:
image

### AFTER
Now they are in line with what the method actually expects/returns. Note that the tuple types are shown as "string" - this appears to be a limitation of Marshmallow:
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/static/resources/openapi.json
+++ b/docs/static/resources/openapi.json
@@ -100,8 +100,13 @@
                           "QUERY_SECURITY_ACCESS_ERROR",
                           "MISSING_OWNERSHIP_ERROR",
                           "USER_ACTIVITY_SECURITY_ACCESS_ERROR",
+                          "DASHBOARD_SECURITY_ACCESS_ERROR",
+                          "CHART_SECURITY_ACCESS_ERROR",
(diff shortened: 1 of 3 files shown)


mistercrunch (approved):
LGTM

michael-s-molina:
> However, the whole tags resource is pretty funky (including the data model and the related enums), so I would almost suggest doing a big refactor on it at some point.

@villebro I wrote a feedback about the Tags feature [here](https://github.com/apache/superset/discussions/25918) which includes some suggestions for the API.

villebro:
Thanks @michael-s-molina , that was a great writeup! I agree with all points raised 👍

PR #31437, opened 2024-12-13
fix(database import): Gracefully handle error to get catalog schemas

### LINKED ISSUE #32214: `import_database` does not work on AWS RDS databases. Because Superset tries to add permissions to all catalogs in RDS database, rather than just one catalog
### Bug description

This issue is regarding `import_database` in `superset/commands/database/importers/v1/utils.py`. 

When importing dashboard assets using `superset import_directory` cli command the `import_database` function is called. This function tries to connect to and add permissions to **all catalogs** in a database, rather than the specific catalog specified in the `sqlalchemy_uri` in `database.yaml`. 

Why is this an issue? If you use an AWS RDS database there is a catalog called `rdsadmin`. This catalog should only be access by AWS to manage automatic backups. When Superset tries run `security_manager.add_permission_view_menu` on all schemas in `rdsadmin` it fails, because it doesn't have access. This also breaks the least privileges principle. 

We solved this issue using `catalog_names.discard("rdsadmin")`, but there should be a proper solution on master. 

My suggestion is that superset only runs `add_permissions` on the catalog specified in the `database.yaml`.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

I don't know

### Node version

Not applicable

### Browser

Not applicable

### Additional context

_No response_

### Checklist

Vitor-Avila:
### SUMMARY
As part of the DB import logic, Superset creates all permissions for the connection (`database access on ...`, `catalog access on ...`, `schema access on ...`, etc). It's possible that the credentials used for the connection has access to list all catalogs, but can't get the schemas for every catalog (this is reproducible if you create an AWS RDS PSQL - the master account will list the `rdsadmin` DB but can't list its schemas).

There was already logic in place to gracefully handle this scenario in the DB creation method. This PR implements the following changes:
1. Moves the `add_permissions` helper that was being used in `superset/commands/database/importers/v1/utils.py` to `superset/commands/database/utils.py` (since now it's used both by import and create commands).
2. Includes the `try/except` logic that was already present in the create command to the `add_permissions` helper. 
3. Update both commands to use this helper, so that it's DRYer. 

fixes: #32214
No UI changes. Import should work without errors.

### TESTING INSTRUCTIONS

Unit test added. For manual testing:

1. Connect to an AWS RDS PSQL DB.
2. Export the connection.
3. Validate that importing it works.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/create.py
+++ b/superset/commands/database/create.py
@@ -39,9 +39,9 @@ from superset.commands.database.ssh_tunnel.exceptions import (
     SSHTunnelInvalidError,
 )
 from superset.commands.database.test_connection import TestConnectionDatabaseCommand
+from superset.commands.database.utils import add_permissions
 from superset.daos.database import DatabaseDAO
 from superset.databases.ssh_tunnel.models import SSHTunnel
-from superset.db_engine_specs.base import GenericDBException
 from superset.exceptions import OAuth2RedirectError, SupersetErrorsException
 from superset.extensions import event_logger, security_manager
 from superset.models.core import Database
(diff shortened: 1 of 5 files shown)


superset/commands/database/create.py
```diff
@@ -148,26 +127,6 @@ def run(self) -> Model:
 
         return database
 
-    def add_schema_permissions(
```
Vitor-Avila, line 151: Searched for `add_schema_permissions` in the code and couldn't find any other occurrence. Happy to leave it there if we feel like it could be useful in the future tho!

betodealmeida (approved):
Nice! Thanks for fixing this and consolidating the logic!

PR #31450, opened 2024-12-14
feat: add a few docker-compose env vars to control the build + docs

mistercrunch:
Living in docker/compose over the past few days, I figure it'll be good to attach build
args to env vars so it's easy to turn things on and off.

As part of this, I'm turning off installing chromium in the build as in most cases developers
won't be working on something that requires it. If you need though, it's
easy to set an env var to switch it on.

Should speed up the local builds a little and make images significantly
smaller locally.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,6 +40,9 @@ x-common-build: &common-build
     - apache/superset-cache:3.10-slim-bookworm
   args:
     DEV_MODE: "true"
+    INCLUDE_CHROMIUM: ${INCLUDE_CHROMIUM:false}
+    INCLUDE_FIREFOX: ${INCLUDE_FIREFOX:false}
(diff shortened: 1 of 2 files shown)


mistercrunch:
Oh I think the bulk of the changes here got merged in my other PR [here](https://github.com/apache/superset/pull/31386).   I kept the default in Dockerfile and set new aligned defaults in `docker-compose.yml` . 

The main change that may affect people is that Chromium is no longer in the `dev` docker image, but only for `docker compose`. Adding it back if/when needed is now well documented.

The benefit is we all get leaner dev environments that build/load faster.

I could add a note in `UPDATING.md` but that shouldn't affect docker images.

sadpandajoe:
@mistercrunch looks like there are some failures when pushing docker in master looking at the master commits. Not sure why they aren't failing in branches but wonder if we could catch these things before they merge

mistercrunch:
Yeah I saw, will fix tomorrow

PR #31451, opened 2024-12-14
chore: remove numba and llvmlite deps as they are large and we don't use them

mistercrunch:
After realizing llvmlite is about 200mb and numba 25mb, looked to see if we need this.

Apparently they come from pandas[performance], but numba is never user unless we tell pandas to use it explicitely in specific calls
Screenshot 2024-12-13 at 4 46 40 PM


eschutho (approved):
Nice!

PR #31460, opened 2024-12-15
docs: Removed mentioning of .env-non-dev in docker/README.md

nikelborm:
### SUMMARY
Even though `.env-non-dev` (not to be confused with `.env-local`) is mentioned in `docker/README.md`, it's not mentioned or used anywhere else in the repo, except it's being excluded from the `linkinator` job inside [`superset-docs-verify.yml`](https://github.com/apache/superset/blob/master/.github/workflows/superset-docs-verify.yml#L31).

The sentence "`docker-compose-non-dev.yml` uses `.env-non-dev` which sets the variable `SUPERSET_ENV` to `production`" is misleading because `docker-compose-non-dev.yml` **doesn't** use `.env-non-dev` and `.env-non-dev` **doesn't** set env var to anything because `.env-non-dev` **doesn't even exist**.

So I went ahead and removed it.

### TESTING INSTRUCTIONS
1. clone repo
2. run `cd superset; grep -rin env-non-dev`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/superset-docs-verify.yml
+++ b/.github/workflows/superset-docs-verify.yml
@@ -28,7 +28,6 @@ jobs:
           linksToSkip: >-
             ^https://github.com/apache/(superset|incubator-superset)/(pull|issue)/\d+,
             http://localhost:8088/,
-            docker/.env-non-dev,
             http://127.0.0.1:3000/,
             http://localhost:9001/,
             https://charts.bitnami.com/bitnami,
--- a/docker/README.md
+++ b/docker/README.md
@@ -68,7 +68,7 @@ Don't forget to reload the page to take the new frontend into account though.
 
 ## Production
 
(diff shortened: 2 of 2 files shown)


mistercrunch:
Right. The idea is that this file is in `.gitgnore` (to prevent people altering/pushing local changes by mistake) and can be used locally to override defaults. Say for instance I could use it to connect to a remove postgres database in my local environment for some reason. It's documented here https://superset.apache.org/docs/installation/docker-compose/#configuring-further so seems fair to remove here as you did in this PR.

Screenshot 2024-12-16 at 6 34 14 PM

It's great to have an easy way to override configs, but in practice I've seen people (me included) getting in a bad state after messing with these files, switching branches, and forgetting they have a "non-standard" environment.

mistercrunch (approved):

PR #31477, opened 2024-12-16
fix: Filter null child names in treeBuilder utility

gerbermichi:
### SUMMARY
This update modifies the treeBuilder utility to filter out null child names. This ensures that the echart sunburst chart does not display null children as <Null>.

Before:
image

After:
image

### TESTING INSTRUCTIONS

Create a sunburst diagram with multiple hierarchies and some null values.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
@@ -98,7 +98,7 @@ const config: ControlPanelConfig = {
             config: {
               type: 'CheckboxControl',
               label: t('Show Null Values'),
-              default: false,
+              default: true,
               renderTrigger: true,
               description: t('Whether to display null values'),
             },
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
@@ -100,7 +100,7 @@ const config: ControlPanelConfig = {
               label: t('Show Null Values'),
               default: true,
               renderTrigger: true,
-              description: t('Whether to display null values'),
+              description: t('Whether to display entries with null values in the hierarchy'),
             },
           },
         ],
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
@@ -100,7 +100,9 @@ const config: ControlPanelConfig = {
               label: t('Show Null Values'),
               default: true,
               renderTrigger: true,
-              description: t('Whether to display entries with null values in the hierarchy'),
+              description: t(
+                'Whether to display entries with null values in the hierarchy',
+              ),
             },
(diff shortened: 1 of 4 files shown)


rusackas:
Code LGTM, but I wonder if there's a use case where people will want to know/see null entries. We could add a "display null values" checkbox, and default it to off/false.

gerbermichi:
Thank you for your feedback, I've update the PR.

superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
```diff
@@ -91,6 +91,18 @@ const config: ControlPanelConfig = {
             },
           },
         ],
+        [
+          {
+            name: 'show_null_values',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show Null Values'),
+              default: false,
```
rusackas:
```suggestion
              default: true,
```

rusackas: We should set this to use the default/existing behavior. Not sure if this will break your tests, but I'll commit this :)

rusackas:
This PR hasn't been updated in a long time and is blocked by pre-commit CI failures. We'll mark it as draft for now — please fix the pre-commit issues, rebase against `master`, and set it back to 'ready for review' when it's ready.

rusackas:
@gerbermichi this already LGTM'd once you added the toggle. Just needs a rebase on `master` and the pre-commit fixes, then back to ready for review. I'll see if I can help with this to move it forward.

rusackas (approved):
Toggle's in, rebase and pre-commit are clean, tests cover both paths. LGTM, thanks for helping this get through. Will merge soon.

superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
```diff
@@ -92,6 +92,18 @@ const config: ControlPanelConfig = {
...
         ],
+        [
+          {
+            name: 'show_null_values',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show Null Values'),
+              default: true,
+              renderTrigger: true,
+              description: t('Whether to display null values'),
```
rusackas:
```suggestion
              description: t('Whether to display entries with null values in the hierarchy'),
```

rusackas:
Almost there @gerbermichi, but codeant raises two fair points: `showNullValues` reads as `undefined` for legacy charts (transformProps.ts:255), and single-level/root null names still slip through (treeBuilder.ts:81). Got time to look into these last touch-ups?

rusackas:
Thank you for this! Merging :D

rusackas:
Oh jeez... I just merged this, having ignored my prior comment. I'll open up a follow-up PR, and link it here. Would appreciate if you can check it out, as this merged as a _kind of_ breaking change for legacy charts. What I'm about to open should make it non-breaking for pre-existing charts, and still support the intent of this PR.

rusackas:
https://github.com/apache/superset/pull/41442 up for grabs as a fast-follow

PR #31478, opened 2024-12-16
feat(sunburst): Add node click functionality to sunburst chart

gerbermichi:
### SUMMARY
This feature introduces node click functionality to the sunburst chart. Users can now click on nodes to focus on one node, enhancing the interactivity and usability of the chart.

image
image

### TESTING INSTRUCTIONS

Create a sunburst chart with multiple hierarchies.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
@@ -93,6 +93,18 @@ const config: ControlPanelConfig = {
         ],
         [
           {
+            name: 'node_click',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Node Click'),
+              default: false,
+              renderTrigger: true,
+              description: t('Whether to enable drill down by node clicking'),
+            },
+          },
+        ],
+        [
+          {
             name: 'label_type',
             config: {
               type: 'SelectControl',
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Sunburst/controlPanel.tsx
```diff
@@ -91,6 +91,18 @@ const config: ControlPanelConfig = {
             },
           },
         ],
+        [
+          {
+            name: 'node_click',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Node Click'),
```
kgabryje, line 99: Can we use a label that describes the feature it enables rather than action? Like `label: t('Enable drill down'),`?

michael-s-molina: @kgabryje I'm wondering what's the relationship of this feature with cross-filtering and drill by? By default, charts offer the left click for cross-filtering and the right-click for the menu options. They don't offer controls to enable/disable this behavior if I'm not mistaken. I wonder if we should just add support for left/right clicks as the other chart types do.

kgabryje: Oh you're right, this feature would be in conflict with cross filtering. I wonder if it would make sense to trigger drill down with some combination like ctrl + left click. Of course then we'd need some tooltip in chart itself for discoverability

michael-s-molina: Why not implement drill down using drill by as we do for all chart types?

PR #31483, opened 2024-12-16
fix: Card component background color

kgabryje:
### SUMMARY
After upgrading the Card component to antd 5, we forgot about the background color token, resulting in cards having blue-ish background color instead of gray.

Before:

image

image

After:

image

image

### TESTING INSTRUCTIONS
Verify that the card components in SQL Lab (the query in the Results tab) and Dataset edit -> Settings have gray backgrounds

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/theme/index.ts
+++ b/superset-frontend/src/theme/index.ts
@@ -75,6 +75,7 @@ const baseConfig: ThemeConfig = {
     Card: {
       paddingLG: supersetTheme.gridUnit * 6,
       fontWeightStrong: supersetTheme.typography.weights.medium,
+      colorBgContainer: supersetTheme.colors.grayscale.light4,
     },
     Input: {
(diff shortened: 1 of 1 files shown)


michael-s-molina:
@kgabryje @geido I'm following the fixes for the Ant Design migration. Some bugs are hard to detect but this one seems really easy to check during the PR review. Are we missing some manual tests when reviewing these migrations?

rusackas (approved):

kgabryje:
@michael-s-molina on 1 hand I agree that this should have been caught by looking at the storybook during the review, but on the other - can't say for others, but I _probably_ wouldn't notice that the color changed. I wish details like this were caught by visual regression tests

PR #31485, opened 2024-12-16
feat(chart): download CSV chart contents from S3 for Athena charts

### LINKED ISSUE #31482: [SIP-150] Use existing Athena Presto functionality for large downloads from S3
### Motivation

Superset inbuilt download functionality is slow for larger file, attempting to download a 100,000 row CSV from Presto Athena AWS DB took just shy of 9 mins from start to completion.

However, this can be vastly sped up with little extra overhead cost to Superset (for Presto DB users). Presto DB can be configured to automatically persist query results to S3, this takes seconds to query DB and save CSV in S3 bucket.

By starting the query through the Athena API and then checking in to get the output location of the CSV, this URL can be returned to user immediately without the raw results, resulting in the real world use case we had, the 8 min 45 second download time was reduced down to 11 seconds.

### Proposed Change

The proposed change is to add in functionality to return only the existing CSV file output_location to user when the user requests a download of a chart that uses an Athena Presto DB.

This change will be protected with feature flags to only turn on for Presto DB users and they will need to set environment variables for 
AWS region/Athena workgroup and Athena DB name, example below from .env file

SUPERSET_REGION=eu-west-1
SUPERSET_WORKGROUP=superset-etl
SUPERSET_ATHENA_DB=my_superset_db

Feature flags are to enable S3 download functionality and to hide existing CSV/XLSX default options (S3 download is faster than default download), from featureFlags.ts

DownloadCSVFromS3 = 'DOWNLOAD_CSV_FROM_S3',
~~ShowDefaultCSVOptions = 'SHOW_DEFAULT_CSV_OPTIONS',~~

~~Option will appear in right click context menu~~

image

UI is now the same to user, functionality will trigger on normal CSV/Full CSV export if feature flag is set and if database.backend is awsathena.

### New or Changed Public Interfaces

Reusing existing data endpoint used by CSV and XLSX default and full download.

Passing in 'result_location', a new parameter to specify if exported file is to be built within Superset (current export) or S3 (new flow for Presto Athena).

Changes to model for output_location, the returned presigned URL which is used to specify the file inside S3 bucket.

All other changes in PR code changes.

### New dependencies

No changes here

### Migration Plan and Compatibility

No DB changes

### Rejected Alternatives

Describe alternative approaches that were considered and rejected.

Using lambda to get file from S3 and returning through proprietary application, rejected as users are already using Superset and it makes sense to allow them to download large files through Superset UI.

william-fundrecs:
### SUMMARY
Add a right click context menu option if the chart is an Athena table to download full chart contents directly from S3 bucket instead of through Superset

image

### TESTING INSTRUCTIONS
Set required env vars
SUPERSET_REGION=<my_aws_region>
SUPERSET_WORKGROUP=<my_superset-workgroup>
SUPERSET_ATHENA_DB=<my_superset_db>

Enable feature flags
'DOWNLOAD_CSV_FROM_S3'
'SHOW_DEFAULT_CSV_OPTIONS'

Ensure Athena is set to automatically persist query results to S3 bucket in CSV format

Test download of chart through right click option.

### ADDITIONAL INFORMATION
- [x] Has associated issue:
- [x] Required feature flags:
- [x] Changes UI

No DB changes

resolves #31482

### DIFF
--- a/superset/common/query_actions.py
+++ b/superset/common/query_actions.py
@@ -107,7 +107,6 @@ def _get_full(
     if datasource.database.db_engine_spec.supports_remote_download(query_context.result_location):
         # Generate presigned URL for output CSV
         presigned_output_location = generate_presigned_url(payload["output_location"])
-        logger.info("ATHENA OUTPUT LOCATION PRESIGNED %s", presigned_output_location)
         return {
             "output_location": presigned_output_location,
         }
--- a/superset/common/query_actions.py
+++ b/superset/common/query_actions.py
@@ -21,11 +21,12 @@ from typing import Any, Callable, TYPE_CHECKING
 
 from flask_babel import _
 
-from superset import app
-from superset.common.chart_data import ChartDataResultType
+from superset import app, feature_flag_manager
+from superset.common.chart_data import ChartDataResultType, ChartDataResultLocation
 from superset.common.db_query_status import QueryStatus
 from superset.connectors.sqla.models import BaseDatasource
 from superset.exceptions import QueryObjectValidationError
+from superset.utils.aws import generate_presigned_url
 from superset.utils.core import (
     extract_column_dtype,
     extract_dataframe_dtypes,
@@ -101,6 +102,16 @@ def _get_full(
     datasource = _get_datasource(query_context, query_obj)
     result_type = query_obj.result_type or query_context.result_type
     payload = query_context.get_df_payload(query_obj, force_cached=force_cached)
+    presigned_output_location = None
+
+    if datasource.database.db_engine_spec.supports_remote_download(query_context.result_location):
+        # Generate presigned URL for output CSV
+        presigned_output_location = generate_presigned_url(payload["output_location"])
+        logger.info("ATHENA OUTPUT LOCATION PRESIGNED %s", presigned_output_location)
+        return {
+            "output_location": presigned_output_location,
+        }
+    
     df = payload["df"]
     status = payload["status"]
     if status != QueryStatus.FAILED:
@@ -109,6 +120,7 @@ def _get_full(
         payload["coltypes"] = extract_dataframe_dtypes(df, datasource)
         payload["data"] = query_context.get_data(df, payload["coltypes"])
         payload["result_format"] = query_context.result_format
+        payload["output_location"] = presigned_output_location
     del payload["df"]
 
     applied_time_columns, rejected_time_columns = get_time_filter_status(
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -63,6 +63,7 @@ from sqlalchemy.sql import ColumnElement, expression, Select
 
 from superset import app, db, db_engine_specs, is_feature_enabled
 from superset.commands.database.exceptions import DatabaseInvalidError
+from superset.common.chart_data import ChartDataResultLocation
 from superset.constants import LRU_CACHE_MAX_SIZE, PASSWORD_MASK
 from superset.databases.utils import make_url_safe
 from superset.db_engine_specs.base import MetricType, TimeGrain
@@ -83,6 +84,7 @@ from superset.superset_typing import (
     ResultSetColumnType,
 )
 from superset.utils import cache as cache_util, core as utils, json
+from superset.utils.aws import run_query_and_get_s3_url
 from superset.utils.backports import StrEnum
 from superset.utils.core import get_username
 from superset.utils.oauth2 import (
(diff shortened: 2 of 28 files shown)


villebro:
@william-fundrecs I assume this PR was mistakenly opened against upstream Superset, not your fork?

rusackas:
Can we do it without adding two feature flags, perhaps leaning on config? And do we need to single out S3? Could there be numerous other options for where people get/store their CSV?

william-fundrecs:
Potentially yes, it made sense to me to add the feature flag to enable the S3 feature but perhaps not as sensible to use a feature flag to hide the default CSV/XLSX right click options, if there is a better way to hide these I can change to it.

This is coloured by our own needs, we wanted to hide the default options in favour of using the S3 download as it's a lot faster, our users will only be using full export and retrieving the file handle from S3 is multitudes faster than creating the dataframe and processing the file.

S3 was specified as this is existing Athena functionality, it can be configured to automatically persist query results to a CSV inside a configured bucket. Potentially there are other use cases for other DB engines but we only use Athena Presto and PostGreSQL in-house.

If someone wants to use this PR as a foundation for expanding file output locations that is fine with me.

william-fundrecs:
This has been updated to remove the S3 download right click context menu option. 

Instead the db type is passed into component props and if feature flag is enabled and db type is athena, the new export functionality will be used, otherwise the existing/default export functionality will be used.

To the user it will look exactly the same (aside from vastly faster download speeds when using Athena with query results persisted to S3 bucket)

superset/common/query_actions.py
```diff
@@ -101,14 +107,28 @@ def _get_full(
     datasource = _get_datasource(query_context, query_obj)
     result_type = query_obj.result_type or query_context.result_type
     payload = query_context.get_df_payload(query_obj, force_cached=force_cached)
+    presigned_output_location = None
+
+    if datasource.database.backend == "awsathena" and "presto" in datasource.database.name.lower() \
```
betodealmeida, line 112: We should move this logic to the DB engine specs themselves, to abstract the feature.

superset/models/core.py
```diff
@@ -688,6 +691,12 @@ def _log_query(sql: str) -> None:
                     security_manager,
                 )
 
+        if self.db_engine_spec.engine == 'awsathena' and is_feature_enabled("DOWNLOAD_CSV_FROM_S3") and result_location == ChartDataResultLocation.S3:
```
betodealmeida, line 694: This too, should be moved to the DB engine spec so we don't have to explicitly test for Athena.

Something like:

```python
if self.db_engine_spec.supports_remote_download:
    url = self.db_engine_spec.get_remote_download_url(sqls[-1])
    ...
```

william-fundrecs:
Can someone look at this PR, I see Github Action tests are failing due to the boto3 import I added, how do I address? 

I see hive.py has a boto3 reference as well but does not fail on tests.

I'm running on a windows machine, unable to run some quality controls locally.

sadpandajoe:
@william-fundrecs when was the last time you rebased? I'd do that first. I'd then maybe take a look at the docker build errors if they are still there as those should be pretty easy to fix. The pre-commit and and frontend build ci seems to be pretty easy fixes too.

rusackas:
Converting this to draft while it awaits a rebase. Mark it as ready for review when it's good to go!

Vitor-Avila:
Does it make sense to actually expose these configs (`SUPERSET_REGION`, `SUPERSET_WORKGROUP` and `SUPERSET_ATHENA_DB`, etc) in the DB connection form? @betodealmeida I don't think we can currently use any of the existing fields in Other/etc to add a JSON with these settings, right? It might require creating a dynamic form for Athena... Just thinking that if the Superset installation has multiple Athena connections, you might need different values for each based on the DB connection, and having it at the connection dialog would be simpler for users to manage it.

rusackas:
Thanks for working on this — Athena S3 download is a genuinely useful feature. We took a look at the branch with an eye toward rebasing it, but found a few issues that would need to be addressed before it's ready to merge:

**Blocking bugs**
- `Chart.jsx` `exportFromS3` uses `this.props.slice.slice_id` and `this.props.isCached`, but `Chart` is a functional component — `this` is undefined there and this would crash at runtime.
- `query_context.py` references `ChartDataResultLocations` (extra `s`) which won't compile — should be `ChartDataResultLocation`.
- `athena.py` `get_remote_download_url` is annotated `-> str` but returns a `pd.DataFrame` (or implicitly `None`).

**Design concern**
- `superset/utils/aws.py` is imported unconditionally in `models/core.py`, making `boto3` a hard dependency for every Superset install, even those not using Athena. The import (and `boto3` itself) should be guarded so it's only required when the Athena engine is actually in use.

**Minor issues**
- `aws.py` is missing the Apache license header (will fail the License Check CI job).
- `aws.py` `run_query_and_get_s3_url` has an unbounded `while True:` polling loop with no timeout — if Athena never responds, the request hangs indefinitely.
- Missing newline at end of `aws.py`.

We're going to close this for now to keep the queue tidy. Please feel free to reopen (or open a new PR) once the above are addressed — happy to help get it over the line at that point.

PR #31489, opened 2024-12-17
fix(dockerfile): podman compose compatibility with BUILDPLATFORM

### LINKED ISSUE #31488: [Bug] redefining BUILDPLATFORM when using Podman
### Bug description

The following error occurs:
> Error: reading multiple stages: attempted to redefine "BUILDPLATFORM": invalid argument

 when using podman with the command:
 ```
 podman compose -f docker-compose-non-dev.yml up --build
 ```

### Screenshots/recordings

image

### Superset version

master / latest-dev

### Python version

3.11

### Node version

18 or greater

### Browser

Firefox

### Additional context

Works with Docker, issue only occus with Podman.

```
OS: macOS Sequoia 15.0.1 arm64
Host: MacBook Pro (16-inch, 2023)
Kernel: Darwin 24.0.0
```

### Checklist

nktnet1:
### SUMMARY

Fixes #31488.

Using a different variable, `TARGET_BUILDPLATFORM`, inside the `Dockerfile` to avoid name conflict/redefining `BUILDPLATFORM` which is an issue in podman.
```Dockerfile
ARG TARGET_BUILDPLATFORM=${BUILDPLATFORM:-amd64}
```

### TESTING INSTRUCTIONS

Run the command
```
podman compose -f docker-compose-non-dev.yml up --build
```
and observe that the error
> Error: reading multiple stages: attempted to redefine "BUILDPLATFORM": invalid argument

Is no longer there.

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sfirke:
This looks good to me, thanks for contributing it! Is there a risk / what size risk of breaking changes for someone who uses the current variable name in a workflow? If significant, I would say let's merge this during the upcoming breaking window (I think it would be excessive to say this requires a vote). If not, we can merge now IMO.

Dockerfile
```diff
@@ -21,12 +21,12 @@
 ARG PY_VER=3.10-slim-bookworm
 
 # If BUILDPLATFORM is null, set it to 'amd64' (or leave as is otherwise).
-ARG BUILDPLATFORM=${BUILDPLATFORM:-amd64}
+ARG TARGET_BUILDPLATFORM=${BUILDPLATFORM:-amd64}
```
mistercrunch, line 24: would `ARG TARGET_BUILDPLATFORM=${TARGET_BUILDPLATFORM:-${BUILDPLATFORM:-amd64}}` be better? - haven't tested it?

I think it would provide a default if somehow both `TARGET_BUILDPLATFORM` and `BUILDPLATFORM` are not in context.

mistercrunch:
I did a bit of research and this should not be an issue AFAICT.  Especially if `BUILDPLATFORM` is always in context.

Looked a bit into podman, seems interesting but not sure whether it'd be beneficial for us to use as our build engine.

Dockerfile
nktnet1: Thanks @mistercrunch - I can confirm that it does work, although I'm not sure if it's necessary. `TARGET_BUILDPLATFORM` was just a variable I made up with the sole reason being that it's a different name to `BUILDPLATFORM`. Apologies if I've misunderstood though. Happy to make the change regardless

mistercrunch: oh why can't you use `BUILDPLATFORM`?

nktnet1: From #31488 - podman gives this error:
image

mistercrunch:
Still on the fence but would require a line in `UPDATING.md` to notify release managers.

nktnet1:
Thanks ＠mistercrunch (and ＠sfirke) - it doesn't look like podman is that widely used by users of apache/superset anyway (I only found 1-2 posts on it) so I'm happy to let this sit (or just close the PR) and work off a fork.

rusackas:
Looks like this is inactive and in need of a rebase. Do we want to get this through, or have we lost interest? I'll convert this to draft for now, but it might be closed eventually if it remains inactive.

nktnet1:
Thanks for the bump @rusackas.

I've made some other podman-compatible changes, e.g. prefixing redis/postgres image names with `docker.io/` (https://github.com/dotnet/aspire/issues/3423).

---

I think the only remaining incompatibility is with caching in line 33:

https://github.com/apache/superset/blob/ef14b529b82798989b3aa850a5442a5e5b127dd9/docker-compose-non-dev.yml#L32-L33

where podman (specifically buildah) does not currently support cache in tags (https://github.com/containers/buildah/issues/4935).

I believe what is expected at the moment is simply removing the tag:

```yaml
 cache_from: 
   - apache/superset-cache
```

However, I don't think this is a blocking incompatibility.

rusackas:
This draft hasn't been updated in >90 days, so it will be closed. Please feel free to reopen it at any time, or ping us on this thread for further help.

PR #31490, opened 2024-12-17
feat: use docker in frontend GHA to parallelize work

mistercrunch:
Optimizing/parallelizing work around frontend builds.

- docker build is typically much faster than CI builds, able to reuse layers
- build happening only once, before parallelizing the work on top
- sharding jest tests, running 8 in parallel (!)
- maintained the codecov coverage reporting though upload is broken both in `master` and here, requires more attention

Clock time savings from 25 to 5-10 minutes depending on docker layer cache

Machine time savings also significant VS parallelizing the build steps.

Oh also, bycatch **fixing the multiple alembic/db migrations heads currently in `master`**

## workflow topology
Screenshot 2025-01-07 at 3 20 27 PM


mistercrunch:
NOTE: this needs to work in forks/without dockerhub creds to be viable

rusackas (approved):
LGTM!

eschutho (approved):

PR #31498, opened 2024-12-17
feat(pivot table): Add feature flag to allow HTML rendering in pivot tables

ERGO1995:
### SUMMARY
After analyzing the source code, a variable is set to `true` by default in the Table chart (`superset/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx`) but not in the Pivot Table (`superset/superset-frontend/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx`). The variable in question is `allowRenderHtml` in the function `export default function TableChart<D extends DataRecord = DataRecord>` or `export default function PivotTableChart(props: PivotTableProps)`.

The goal is to make this value configurable via a feature flag. Therefore, a new feature flag needs to be declared in the file `superset/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts` as follows:
`AllowPivotTableHtml = 'ALLOW_PIVOT_TABLE_HTML'`.

Then, retrieve this feature flag in the source code of the Pivot Table like this:
`allowRenderHtml = isFeatureEnabled(FeatureFlag.AllowPivotTableHtml)`

Finally, declare the new feature flag in the superset_config.py file and set it to true as follows:
```
FEATURE_FLAGS = {
    "ALLOW_PIVOT_TABLE_HTML": True,
}
```

Capture d'écran 17 12 2024 à 15 03 50 PM
Capture d'écran 17 12 2024 à 15 04 10 PM

### TESTING INSTRUCTIONS
Set the feature flag ALLOW_PIVOT_TABLE_HTML to "True" in your superset_config.py and look if your html is formated or not in the Pivot Table.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [x] Required feature flags: ALLOW_PIVOT_TABLE_HTML
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
@@ -48,6 +48,7 @@ const d3Formatted: [string, string][] = [
   ',.3f',
   '+,',
   '$,.2f',
+  ' ,.2f',
 ].map(fmt => [fmt, `${fmt} (${getNumberFormatter(fmt).preview()})`]);
 
 // input choices & options
--- a/superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
@@ -48,7 +48,6 @@ const d3Formatted: [string, string][] = [
   ',.3f',
   '+,',
   '$,.2f',
-  ' ,.2f',
 ].map(fmt => [fmt, `${fmt} (${getNumberFormatter(fmt).preview()})`]);
 
 // input choices & options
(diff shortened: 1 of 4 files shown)


rusackas:
CC @kasiazjc @mistercrunch since they've been looking at this sort of thing.

ERGO1995:
> I'm curious more about the use cases here...
> 
> Should this perhaps be user-controllable (there's a switch for it in SQL Lab results, for example), rather than a feature flag? I'm not sure if we want to _always_ enable or disable this for everyone.
> 
> If we DO need a feature flag, perhaps we should have a more generic "RENDER_HTML_IN_DATA" flag, that can be used for any tables (pivot or otherwise) or results displays?

Hey, when I checked the code for the "Table" and "Pivot Table" charts, I noticed that the variable is set to true by default only on the "Table" side. If there is a good reason for this, I assume we wouldn't want just any regular user to make a change to allow it. Based on this assumption, I used feature flags so that someone with proper expertise can set this feature flag to True.

sadpandajoe:
@ERGO1995 looks like for pivot tables I'm seeing a checkbox under the `Customize` tab to render html. I think there is probably a bug there where it isn't actually rendering the html so I don't know if this feature flag is actually needed. This will allow the chart creator to choose to render the data as html and I don't think the feature flag is needed if we actually fix the functionality.

PR #31503, opened 2024-12-17
chore: python version to 3.11 (while supporting 3.10)

### LINKED ISSUE #31516: 73 - Drop support for Python 3.9
Check https://github.com/apache/superset/pull/24112 for more context.

mistercrunch:
Bumping python to 3.11 as the default version, supporting 3.10, and deprecating 3.9 support

Currently in this PR I'm setting `next` to `3.11` and `current` to `3.11.11`. This means when 3.11.12 shows up CI will run on two different versions within 3.11. Eventually we can make `next` point to 3.12 as lib support grows.

# about 3.12

Seems 3.12 is not easy at the moment, hit some issues that pointed to upgrading dependencies, but hit some conflicting diamond dependencies as I was doing so. https://github.com/apache/superset/pull/31313 would help sort this out at a faster pace, without the bugs and slowness of `pip-compile-multi`

Closes https://github.com/apache/superset/issues/31516

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -28,7 +28,7 @@ runs:
         if [ "${{ inputs.python-version }}" = "current" ]; then
           echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
-          # TODO: set to 3.12 when libs are ready - seemed hard on 12/17/24
+          # currently disabled in GHA matrixes because of library compatibility issues
           echo "PYTHON_VERSION=3.12" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "previous" ]; then
           echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -26,10 +26,10 @@ runs:
       shell: bash
       run: |
         if [ "${{ inputs.python-version }}" = "current" ]; then
-          echo "PYTHON_VERSION=3.11.11" >> $GITHUB_ENV
+          echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
           # TODO: set to 3.12 when libs are ready - seemed hard on 12/17/24
-          echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
+          echo "PYTHON_VERSION=3.12" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "previous" ]; then
           echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
         else
@@ -44,11 +44,6 @@ runs:
       run: |
         if [ "${{ inputs.install-superset }}" = "true" ]; then
           sudo apt-get update && sudo apt-get -y install libldap2-dev libsasl2-dev
-
-          if [ "${{ env.PYTHON_VERSION }}" = "next" ]; then
-            # No wheel in 3.11 for numpy
-            sudo apt-get -y install build-essential
-          fi
 
           pip install --upgrade pip setuptools wheel uv
 
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -26,11 +26,12 @@ runs:
       shell: bash
       run: |
         if [ "${{ inputs.python-version }}" = "current" ]; then
-          echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
+          echo "PYTHON_VERSION=3.11.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
+          # TODO: set to 3.12 when libs are ready - seemed hard on 12/17/24
           echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "previous" ]; then
-          echo "PYTHON_VERSION=3.9" >> $GITHUB_ENV
+          echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
         else
           echo "PYTHON_VERSION=${{ inputs.python-version }}" >> $GITHUB_ENV
         fi
@@ -43,6 +44,12 @@ runs:
       run: |
         if [ "${{ inputs.install-superset }}" = "true" ]; then
           sudo apt-get update && sudo apt-get -y install libldap2-dev libsasl2-dev
+
+          if [ "${{ env.PYTHON_VERSION }}" = "next" ]; then
+            # No wheel in 3.11 for numpy
+            sudo apt-get -y install build-essential
+          fi
+
           pip install --upgrade pip setuptools wheel uv
 
           if [ "${{ inputs.requirements-type }}" = "dev" ]; then
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,6 @@ authors = [
 classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
-    # "Programming Language :: Python :: 3.12", # not just yet ...
 ]
 dependencies = [
     "backoff>=1.8.0",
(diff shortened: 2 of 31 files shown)


michael-s-molina:
> deprecating 3.9 support

@mistercrunch There's a [5.0 proposal](https://github.com/orgs/apache/projects/345/views/1?filterQuery=python&pane=issue&itemId=59021353) to drop support for 3.9. We should wait for the breaking window unless you remove the 3.9 deprecation of this PR's scope.

mistercrunch:
@michael-s-molina yup I'm taking this on as part of the 5.0 effort.

> Should we first merge https://github.com/apache/superset/pull/31313

@villebro yes, that will help sorting things out. I'll push on that one to go first (pre 5.0)

> should we not wait until 3.12 is supported ?

I don't think that's required, my take is we should support `previous` and `current`, and next is a nice to have as a placeholder to start getting things working. The `next` construct is useful to support the process of getting ready for it, but not needed to deprecate a previous version. Given the state of CI, and the fact that only some checks run on `previous` and `next`, we should strongly recommend using the `current` version, the next/previous are useful if/when we want to support more or while we're in the process of establishing or deprecating support.

rusackas:
Assuming this should effectively resolve #31516, right? Slapping the `hold:next-major` label so we don't merge this until the breaking window.

mistercrunch:
@rusackas @villebro @michael-s-molina we should be able to merge this since the breaking change window is open

.github/actions/setup-backend/action.yml
```diff
@@ -26,11 +26,12 @@ runs:
       shell: bash
       run: |
         if [ "${{ inputs.python-version }}" = "current" ]; then
-          echo "PYTHON_VERSION=3.10" >> $GITHUB_ENV
+          echo "PYTHON_VERSION=3.11.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
+          # TODO: set to 3.12 when libs are ready - seemed hard on 12/17/24
           echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
```
villebro, line 32: Aren't these reversed, i.e. `current` should point to `3.11` and `next` should point to `3.11.11`?

pyproject.toml
```diff
@@ -24,15 +24,15 @@ name = "apache-superset"
...
+requires-python = ">=3.10"
 license = { file="LICENSE.txt" }
 authors = [
     { name = "Apache Software Foundation", email = "dev@superset.apache.org" },
 ]
 classifiers = [
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    # "Programming Language :: Python :: 3.12", # not just yet ...
```
villebro, line 35: If 3.12 isn't supported yet, I'd just remove this for now.
```suggestion
```

villebro (commented):
One main comment about the versions perhaps being inverted on current vs next. My main gripe with this PR is the fact that we're essentially running two 3.11 CI checks, which seems wasteful. As 3.12 really isn't yet supported, I think we should consider just disabling the CI step for next for now until we are actually able to test against it.

mistercrunch:
@villebro I wish it was easier to disable `next`, but given it's marked as a `required_status_checks` in `.asf.yml`, it would require a preliminary PR removing the check, merged in `master`, rebasing this one, and then reenabling it eventually.

I guess we could decide that `next` isn't required now and in the future, more of an indicator on readiness on the next version. I kind of like the idea, except showing any kind of red on PR checks is unsettling/confusing to contributors...

villebro:
> I guess we could decide that `next` isn't required now and in the future, more of an indicator on readiness on the next version. I kind of like the idea, except showing any kind of red on PR checks is unsettling/confusing to contributors...

@mistercrunch I wonder if there's some way to downplay a CI check to yellow or something else to not make it look like a hard failure? I personally always get put off by failing checks, even if they're not required, and I suspect it will be even more confusing to a casual contributor who doesn't have context on this forthcoming work..

mistercrunch:
Screenshot 2025-01-07 at 10 01 44 AM

villebro:
I googled this, and came across this: https://stackoverflow.com/questions/57687369/how-to-write-a-github-action-with-a-neutral-check-run-status. So I don't think it works.

> I discovered through some [old documentation](https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/) that exit code 78 gives a neutral status for the step. But how do I give the entire job a neutral status?
> 
> Update: After testing, I discovered that exit code 78 does not affect the status of the step.

mistercrunch:
Here -> https://github.com/apache/superset/pull/31740

mistercrunch:
Mmmmh, yeah goes back to 2019 so my hopes aren't up on this one, I couldn't find a decent workaround.Maybe it's a good think to make optional as in https://github.com/apache/superset/pull/31740 though?

mistercrunch:
Few options:
- leave this PR as is, and assume **for now** that `next` means the latest 3.11.x
- set 3.12 as next and live with the red, deal with confusion for contributors (I think that's not great)
- set 3.12 as next and force a green status, have to dig in logs to see if it's actually working
- set 3.12 as next and keep pushing on this PR until we get it working

villebro:
@mistercrunch what was missing for 3.12 support? Numpy wheels? Let's give it one more go, and if it just doesn't work out let's settle for one of the compromises listed above..

mistercrunch:
Ok, added a commit bringing in 3.12 in the mix, let's see what fails.

mistercrunch:
Ok I pushed by bumping a bunch of libraries here, and ultimately got to a place where `next` hung on `setup-backend` while installing python deps. Turns out there were no wheels out for pandas/numpy for the version we had pinned. So I bumped it beyond the range that was specified in our `pyproject.toml` (there was no note around why it had an upper bound) and uncovered some other issue we have with this newer version `pandas`.

mistercrunch:
The new pandas `df.to_sql` seem doesn't support `sqlalchemy` datatypes or at least not the same way -> https://github.com/pandas-dev/pandas/issues/57049

villebro:
@mistercrunch after letting this simmer for a while longer, would the following work:
- we run 3.11 on `Dockerfile` and refer to it as the currently supported version
- on the CI matrix, we replace `current` (3.11) and `next` (3.12) with `previous` (3.10) and `current` (3.11). After all, those are the only versions we officially support.

I don't necessarily see any reason to have `next` on CI if it just causes issues like this. So in the future, when we do want to start supporting 3.12, we open up a PR that bumps the versions on `previous` and `next`, and then iterate until both versions actually work.

mistercrunch:
Yes I think that's effectively what it is in current PR, except for next being failing on 3.12 now. Maybe I just remove next from the matrixes for now and we turn it back on when we're ready to start working on supporting 3.12

mistercrunch:
Ok I didn't want to loose the work I did on supporting 3.12 so started this placeholder PR [here](https://github.com/apache/superset/pull/31759). I'll rollback 3.12 support work from this PR as it's breaking things, and remove `next` from our matrixes on this PR RN

michael-s-molina:
@mistercrunch Could you please add the changes made in this PR to `UPDATING.md`?

mistercrunch:
Done! I think this might be ready for merging

michael-s-molina:
> Done! I think this might be ready for merging

I didn't see a change in `UPDATING.md`. Did you push it? I believe it's the last step.

mistercrunch:
@villebro @michael-s-molina anything else needed to get this merged?

michael-s-molina:
> @villebro @michael-s-molina anything else needed to get this merged?

I didn't see a change in UPDATING.md. Did you push it? I believe it's the last required step.

mistercrunch:
Oops, I forgot to `git push` my commit. I just pushed it

michael-s-molina (approved):

michael-s-molina:
@mistercrunch I left a comment about the integration tests in the committers channel.

villebro (approved):
LGTM, but before merging, would you be able to add a note about the `strict=False` bit in the description? I assume we're mostly ok with not being strict about those zips, but In general it's probably usually better to be more strict than not..

mistercrunch:
I think something changed in 3.11 or 3.12 about this particular method, and `ruff` caught it and required this change as lint. It'll enforce this rule moving forward and is upfront about the reason (it's easy to lookup the specific rule and why it's enforced)

PR #31515, opened 2024-12-18
chore: deprecate fossa in favor of liccheck to validate python licenses

mistercrunch:
It appears we are not doing python deps review properly. Fossa is super hard to use/configure given our setup, so moving to a different tool `liccheck` that checks all the boxes


rusackas (approved):
LGTM!

PR #31517, opened 2024-12-18
chore: update browser list

mistercrunch:
Saw a message in my logs about updating the browser list with `npx
update-browserslist-db@latest`, so here we go.


rusackas (approved):

rusackas:
Thanks! I was just about to open the same PR :P

PR #31519, opened 2024-12-19
chore: remove dependency on func_timeout because LGPL

mistercrunch:
func-timeout is a LGPL library, so getting rid of it in this PR.

Luckily we ha another timeout context manager we could just use here in place of this one.


rusackas (approved):
LGTM, thanks for the cleanup!

PR #31580, opened 2024-12-20
feat(doris): add catalog support for Apache Doris

liujiwen-up:
### SUMMARY

add catalog for apache doris
In Apache Doris, in order to be compatible with different BI tools, we have information_schema for each catalog. When no catalog is specified, the default catalog is internal. This feature corresponds to this PR of Doris, https://github.com/apache/doris/pull/28919
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/doris.py
+++ b/superset/db_engine_specs/doris.py
@@ -251,7 +251,7 @@ class DorisEngineSpec(MySQLEngineSpec):
         if uri.database and "." in uri.database:
             current_catalog, _ = uri.database.split(".", 1)
         else:
-            current_catalog = uri.database
+            current_catalog = "internal"
 
         # In Apache Doris, each catalog has an information_schema for BI tool
         # compatibility. See: https://github.com/apache/doris/pull/28919
--- a/superset/db_engine_specs/doris.py
+++ b/superset/db_engine_specs/doris.py
@@ -22,11 +22,13 @@ from urllib import parse
 
 from flask_babel import gettext as __
 from sqlalchemy import Float, Integer, Numeric, String, TEXT, types
+from sqlalchemy.engine.reflection import Inspector
 from sqlalchemy.engine.url import URL
 from sqlalchemy.sql.type_api import TypeEngine
 
 from superset.db_engine_specs.mysql import MySQLEngineSpec
 from superset.errors import SupersetErrorType
+from superset.models.core import Database
 from superset.utils.core import GenericDataType
 
 # Regular expressions to catch custom errors
@@ -111,6 +113,7 @@ class DorisEngineSpec(MySQLEngineSpec):
     )
     encryption_parameters = {"ssl": "0"}
     supports_dynamic_schema = True
+    supports_catalog = supports_dynamic_catalog = True
 
     column_type_mappings = (  # type: ignore
         (
@@ -245,16 +248,44 @@ class DorisEngineSpec(MySQLEngineSpec):
         catalog: Optional[str] = None,
         schema: Optional[str] = None,
     ) -> tuple[URL, dict[str, Any]]:
-        database = uri.database
-        if schema and database:
-            schema = parse.quote(schema, safe="")
-            if "." in database:
-                database = database.split(".")[0] + "." + schema
-            else:
-                database = "internal." + schema
-            uri = uri.set(database=database)
+        if uri.database and "." in uri.database:
+            current_catalog, _ = uri.database.split(".", 1)
+        else:
+            current_catalog = uri.database
 
+        # In Apache Doris, each catalog has an information_schema for BI tool
+        # compatibility. See: https://github.com/apache/doris/pull/28919
+        adjusted_database = ".".join(
+            [catalog or current_catalog or "", "information_schema"]
+        ).rstrip(".")
+
+        uri = uri.set(database=adjusted_database)
         return uri, connect_args
+
+    @classmethod
+    def get_default_catalog(cls, database: Database) -> Optional[str]:
+        """
+        Return the default catalog.
+        """
+        if database.url_object.database is None:
+            return None
+
+        return database.url_object.database.split(".")[0]
+
+    @classmethod
+    def get_catalog_names(
+        cls,
+        database: Database,
+        inspector: Inspector,
+    ) -> set[str]:
+        """
+        Get all catalogs.
+        For Doris, the SHOW CATALOGS command returns multiple columns:
+        CatalogId, CatalogName, Type, IsCurrent, CreateTime, LastUpdateTime, Comment
+        We need to extract just the CatalogName column.
+        """
+        result = inspector.bind.execute("SHOW CATALOGS")
+        return {row.CatalogName for row in result}
 
     @classmethod
     def get_schema_from_engine_params(
--- a/superset/db_engine_specs/doris.py
+++ b/superset/db_engine_specs/doris.py
@@ -258,7 +258,7 @@ class DorisEngineSpec(MySQLEngineSpec):
         # In Apache Doris, each catalog has an information_schema for BI tool
         # compatibility. See: https://github.com/apache/doris/pull/28919
         schema = schema or "information_schema"
-        database = ".".join([catalog or "", schema])
+        database = ".".join([catalog, schema])
         uri = uri.set(database=database)
         return uri, connect_args
 
--- a/superset/db_engine_specs/doris.py
+++ b/superset/db_engine_specs/doris.py
@@ -257,7 +257,10 @@ class DorisEngineSpec(MySQLEngineSpec):
 
         # In Apache Doris, each catalog has an information_schema for BI tool
         # compatibility. See: https://github.com/apache/doris/pull/28919
-        adjusted_database = ".".join([catalog or "", "information_schema"])
+        if schema:
+            adjusted_database = ".".join([catalog or "", schema])
+        else:
+            adjusted_database = ".".join([catalog or "", "information_schema"])
         uri = uri.set(database=adjusted_database)
         return uri, connect_args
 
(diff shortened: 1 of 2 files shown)


michael-s-molina:
@mistercrunch Any idea about the [docker image error](https://github.com/apache/superset/actions/runs/12429138962/job/34702037859?pr=31580)?

rusackas:
@betodealmeida is definitely the subject matter expert on catalog support lately!

tests/unit_tests/db_engine_specs/test_doris.py
```diff
@@ -145,3 +151,52 @@ def test_get_schema_from_engine_params() -> None:
...
+    database.url_object.database = "catalog1.schema1"
+    assert DorisEngineSpec.get_default_catalog(database) == "catalog1"
+
+    # Test with only catalog format
+    database.url_object.database = "catalog1"
+    assert DorisEngineSpec.get_default_catalog(database) == "catalog1"
+
+    # Test with None
+    database.url_object.database = None
+    assert DorisEngineSpec.get_default_catalog(database) is None
```
villebro, line 175: Could we parametrize this test with `pytest.mark.parametrize`?

tests/unit_tests/db_engine_specs/test_doris.py
```diff
@@ -85,25 +86,25 @@ def test_get_column_spec(
         (
             "doris://user:password@host/db1",
             {"param1": "some_value"},
-            "db1",
+            "internal.information_schema",
```
villebro, line 89: Is this a breaking change? To me it seems `db1` is indeed the database, not `internal.information_schema`?

tests/unit_tests/db_engine_specs/test_doris.py
```diff
@@ -120,6 +121,11 @@ def test_adjust_engine_params(
     returned_url, returned_connect_args = DorisEngineSpec.adjust_engine_params(
         url, connect_args
     )
+    # # Update expected schema to include .information_schema
+    # if return_schema not in ".":
+    #     expected_schema = f"{return_schema}.information_schema"
+    # else:
+    #     expected_schema = return_schema
```
villebro, line 128: Why is this commented out block included here?

villebro (changes requested):
This PR is missing a proper description, making it difficult to understand key details in the featured changes. Also, a few general comments that caught my eye while reviewing.

tests/unit_tests/db_engine_specs/test_doris.py
liujiwen-up: @villebro  Yes, in Apache Doris, in order to be compatible with different BI tools, we have information_schema for each catalog. When no catalog is specified, the default catalog is internal. This feature corresponds to this PR of Doris, https://github.com/apache/doris/pull/28919

tests/unit_tests/db_engine_specs/test_doris.py
liujiwen-up: Sorry, this is redundant, I will delete it

tests/unit_tests/db_engine_specs/test_doris.py
liujiwen-up: Thanks, I have changed it as requested.

tests/unit_tests/db_engine_specs/test_doris.py
villebro: Got it, thanks for the clarification 👍

superset/db_engine_specs/doris.py
```diff
@@ -245,17 +248,45 @@ def adjust_engine_params(
...
+        if uri.database and "." in uri.database:
+            current_catalog, _ = uri.database.split(".", 1)
+        else:
+            current_catalog = "internal"
+
+        # In Apache Doris, each catalog has an information_schema for BI tool
+        # compatibility. See: https://github.com/apache/doris/pull/28919
+        adjusted_database = ".".join(
+            [catalog or current_catalog or "", "information_schema"]
+        ).rstrip(".")
```
villebro, line 260: This may sound like a total nit, but I actually had some issues following what's going on here, especially the `catalog or current_catalog or ""` logic. As `current_catalog` is unnecessary if `catalog` is defined, I would have maybe just reused the latter variable for all these uses. Something like:

```python
        if catalog:
            pass
        elif uri.database and "." in uri.database:
            catalog, _ = uri.database.split(".", 1) or ""  # notice how I also moved the `or ""` part here
        else:
            catalog = "internal"
```
Then later just

```python
        adjusted_database = ".".join([catalog, "information_schema"])
```
Also, why is `.rstrip(".")` needed? I don't see how we can ever hit that, as `adusted_database` will always end with `.information_schema`, right?

villebro (commented):
Thanks for the improvements @liujiwen-up! One last comment, after that I feel this is good to go 👍

superset/db_engine_specs/doris.py
liujiwen-up: @villebro Yes, you are right. Thank you very much for your suggestion. I have modified it according to your suggestion.

liujiwen-up: @villebro Thanks for your advice. After in-depth testing with Doris, we found that there is still a problem. The previous test only tested the case of linking data sources. When operating on SQL Lab, it will also go to this function and cannot use the information_schema library fixedly. When there is a schema value, the user-provided schema should be used for querying. This implementation is the correct behavior at present.
1. When linking data sources, the schema is empty and the information_schema library is used uniformly
2. When the schema has a value, the schema value provided by the user is used

liujiwen-up:
@villebro Please help me push it forward. Thank you.
In addition, I would like to ask, after the PR is merged, in which superset release version will it be released? I need to update the official documentation of apache doris based on this information for reference by superset users.

superset/db_engine_specs/doris.py
```diff
@@ -245,17 +248,47 @@ def adjust_engine_params(
...
+        else:
+            catalog = "internal"
+
+        # In Apache Doris, each catalog has an information_schema for BI tool
+        # compatibility. See: https://github.com/apache/doris/pull/28919
+        if schema:
+            adjusted_database = ".".join([catalog or "", schema])
+        else:
+            adjusted_database = ".".join([catalog or "", "information_schema"])
+        uri = uri.set(database=adjusted_database)
```
villebro, line 264: Schema assignment can be simplified, and I'd prefer to just rename `adjusted_database` to `database`:
```suggestion
        # In Apache Doris, each catalog has an information_schema for BI tool
        # compatibility. See: https://github.com/apache/doris/pull/28919
        schema = schema or "information_schema"
        database = ".".join([catalog or "", schema])
        uri = uri.set(database=database)
```

villebro (commented):
One more round to simplify the code (it'll be easier for maintainers to carry this code forward with less duplication and ambiguity). We should be able to get this into 5.0 as long as we can get this merged before the release cut (no rush yet, but let's try to finish this as soon as possible).

One more thing that comes to mind: is it really necessary to assign `information_schema` to the connection string if no schema is selected? Typically we just leave it unspecified (if someone wants to access tables in the `information_schema`, they can just choose that schema explicitly).

superset/db_engine_specs/doris.py
liujiwen-up: @villebro 
One more thing that comes to mind: is it really necessary to assign information_schema to the connection string if no schema is selected?
This is necessary here. In Doris, there must be a schema to link correctly. Catalog is not necessary, but schema is necessary.
When switching catalogs in the sqllab front-end interface, when you select a catalog, you don't know which schemas will be under the catalog. If there is no schema in the link string, an error will be reported. However, one thing is known: there is an information_schema library in each catalog, so using it to link to the search is a very good choice, which can ensure that every time you switch catalogs, it will succeed. After obtaining the schma list under the current catalog, users can select the corresponding schema to obtain all tables.

villebro: Makes sense, thanks for the clarification 👍

villebro (approved):

superset/db_engine_specs/doris.py
```diff
@@ -245,17 +248,45 @@ def adjust_engine_params(
...
+            pass
+        elif uri.database and "." in uri.database:
+            catalog, _ = uri.database.split(".", 1)
+        else:
+            catalog = "internal"
+
+        # In Apache Doris, each catalog has an information_schema for BI tool
+        # compatibility. See: https://github.com/apache/doris/pull/28919
+        schema = schema or "information_schema"
+        database = ".".join([catalog or "", schema])
```
betodealmeida, line 261: Looks like the `catalog or ""` is not needed here, since `catalog` has to be a string (if a `None` is passed then it's a string either from line 254 or 256). So if it is falsy, it has to be an empty string already.

```suggestion
        database = ".".join([catalog, schema])
```

betodealmeida (approved):

superset/db_engine_specs/doris.py
```diff
@@ -245,17 +248,45 @@ def adjust_engine_params(
...
+            pass
+        elif uri.database and "." in uri.database:
+            catalog, _ = uri.database.split(".", 1)
+        else:
+            catalog = "internal"
+
+        # In Apache Doris, each catalog has an information_schema for BI tool
+        # compatibility. See: https://github.com/apache/doris/pull/28919
+        schema = schema or "information_schema"
+        database = ".".join([catalog, schema])
```
villebro, line 261:
```suggestion
        database = ".".join([catalog or "", schema])
```

villebro:
@betodealmeida it appears mypy wasn't smart enough to understand that `catalog` will always be `str`. I committed the change back, and will merge when CI is ready.

liujiwen-up:
@villebro Hello, in which version will this feature be released?

Issue #31510, opened 2024-12-18
docker compose -f docker-compose-image-tag.yml up fails with "uv: command not found"


2af:
### Bug description

I follow the quickstart guide and at the step of 
**docker compose -f docker-compose-image-tag.yml up**
It downloaded all the images and after the DB and REDIS containers initialized all the other containers were constantly spamming the output with messages like:
```

superset_app          | Reinstalling the app in editable mode
superset_app          | /app/docker/docker-bootstrap.sh: line 24: uv: command not found
superset_worker_beat  | Reinstalling the app in editable mode
superset_worker_beat  | /app/docker/docker-bootstrap.sh: line 24: uv: command not found

```

I was able to start Superset with 
**docker compose -f docker-compose-non-dev.yml up**

How can I troubleshoot what happened with the suggested 
**docker-compose-image-tag.yml**
flow?

### Superset version

master / latest-dev

### Python version

Not applicable

### Node version

Not applicable

### Browser

Not applicable

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

kidzen:
same here

small-flowers:
I have also encountered the same problem. How can I solve it?

maeddes:
I have the same issue on different platfors. It seems that the `uv` command got added in some files, e.g. `docker/docker-boostrap.sh` however it is not present in the underlying superset container image.

Temporary workaround is to go back to a version previous to the commits of December, e.g. https://github.com/apache/superset/releases/tag/4.1.1

`git fetch --tags` if you cloned with `depth=1` and `git checkout 4.1.1` made my docker compose work again with the docker-compose-image-tag.yml file.

I still think there will be a patched base image soon.

javalram-ow:
same here

Lucas-87an:
1. Lack of the package that provides the uv command:
The uv command may be related to the uvicorn package, which is a popular ASGI server for Python applications such as Apache Superset.
If uvicorn is not installed correctly in the environment, the system will not be able to find the uv command (which would normally be a shortcut to uvicorn itself).

in the docker-bootstrap.sh file it worked like this for me
set -eo pipefail

# Ensure uvicorn is installed
pip install uvicorn

# Make python interactive
[...]

PR #31583, opened 2024-12-20
fix(docs): get quickstart guide working again

sfirke:
### SUMMARY
A short-term fix for #31510. The install process of `docker-compose-image-tag.yml` is broken because the master branch files like [docker/docker-bootstrap.sh](https://github.com/apache/superset/blob/master/docker/docker-bootstrap.sh) now use `uv` but that package is missing in official versions like `4.1.1`.

Hard-coding a version here is not ideal but this gets new users going again. After the next official release cut from the main/master branch, if `docker compose -f docker-compose-image-tag.yml up` again works without specifying a certain tag, we can revert this PR.

### TESTING INSTRUCTIONS
Test the steps in the quickstart and see that it works now.

### ADDITIONAL INFORMATION
- [x] Has associated issue: #31510

### DIFF
--- a/docs/docs/quickstart.mdx
+++ b/docs/docs/quickstart.mdx
@@ -31,6 +31,9 @@ $ git clone https://github.com/apache/superset
 # Enter the repository you just cloned
 $ cd superset
 
+# Set the repo to the state associated with the latest official version
+$ git checkout tags/4.1.1
+
 # Fire up Superset using Docker Compose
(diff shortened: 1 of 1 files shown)


rusackas (approved):

rusackas:
LGTM, but CC @mistercrunch

mistercrunch (approved):
Related: https://github.com/apache/superset/pull/31588

Issue #26395, opened 2024-01-02
[SIP-111] Proposal for improved database, schema, and table selection UI in SQL Lab sidebar


justinpark:
### Motivation
Currently, when users extend the SQL editor by collapsing the left sidebar, the database and schema selectors become inaccessible.

|Fig. 1. SQL editor with the sidebar|Fig. 2. SQL Editor with folding the sidebar|
|--|--|
|![Image](https://github.com/apache/superset/assets/1392866/fb5f86e2-6418-4a90-bdfe-2d91fbbbab39)|![Image](https://github.com/apache/superset/assets/1392866/f351dee1-918e-46b5-9146-aa3cffde570c)|

This not only makes it difficult for users to utilize the expanded editor layout UI, but it also limits the use of the left sidebar area, which is dedicated to the database and schema selector options. (We aim to customize the left sidebar in a manner similar to the Visual Studio Code sidebar, enabling a more interactive layout that includes features like a ChatGPT prompt and more.) Additionally, the dropdown UI for the table selector lacks usability and does not allow for easy navigation to the corresponding datasource without the need for the database and schema selectors. Moreover, the table metadata, including column schema and record preview, is displayed separately from the table schema dropdown, making it challenging for users to manage and explore their table schema view effectively. Furthermore, there is no search functionality available to find matches in the datasource catalog and table metadata and schema, which hampers users' ability to locate specific information quickly.

### Proposed Change [Updated]

![Screenshot 2024-09-25 at 6 09 33 PM](https://github.com/user-attachments/assets/2b38eb3e-e2cf-40f9-b391-a5b587e0f38b)

The proposed change is to:

> **The tree list starting from the existing database list has been excluded due to performance issues related to search and the complexity of the UI involving catalogs. The list of schemas included in the respective database will appear at the top of the tree view through the database selector. [see the above screenshot]**

1. The database selector will be maintained, while the schema selector will be moved to a tree view.
2. Change the dropdown UI for the table selector to a tree view for improved navigation.
3. For databases that support catalogs, the catalog selector will be displayed under the database selector, and the tree will show the list of schemas for the corresponding category. [see the below screenshot]
![Screenshot 2024-09-25 at 6 12 47 PM](https://github.com/user-attachments/assets/45e09454-3b9b-4a57-a23e-e616ebdce1d4)
[...]

PR #31587, opened 2024-12-20
fix(table_metadata): longType truncate the ROW/MAP structure

justinpark:
### SUMMARY
Currently, coltype stringify does not properly explain the structure of multi-depth in the form of MAP and ROW types.
This code has been modified to help clarify the structure of MAP/ROW types in Trino. (FYI: Superset does *NOT* consume `longType` value yet but it will be used for https://github.com/apache/superset/issues/26395) 

api/v1/database/<dbid>/table_metadata

Before:
```
    {
      "comment": null,
      "keys": [],
      "longType": "BOOLEAN",
      "name": "is_boolean",
      "type": "BOOLEAN"
    },
    {
      "comment": null,
      "keys": [],
      "longType": "ROW",
      "name": "map_value",
      "type": "ROW"
    },
```

After:

```
    {
      "comment": null,
      "keys": [],
      "longType": "BOOLEAN()",
      "name": "is_boolean",
      "type": "BOOLEAN"
    },
    {
      "comment": null,
      "keys": [],
      "longType": "ROW([('reason', VARCHAR()), ('response', VARCHAR()), ('is_bool', BOOLEAN()), ('ids', ARRAY(VARCHAR()))])",
      "name": "map_value",
      "type": "ROW"
    },
```

### TESTING INSTRUCTIONS
Unit tests are added

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/databases/utils.py
```diff
@@ -55,7 +55,7 @@ def get_indexes_metadata(
 
 def get_col_type(col: dict[Any, Any]) -> str:
     try:
-        dtype = f"{col['type']}"
+        dtype = f"{[col['type']]}"[1:-1]
```
mistercrunch, line 58: Hard for me to make sense of this line of code before and after the change. Maybe needs a comment, docstring and/or better typing?

Also if this is solving a trino-specific issue, does it belong somewhere in db_engine_spec?

mistercrunch:
Not directly related, but while trying to add support for python 3.12 on another PR, noticed some things that are related to this.
- newer version of pandas removes support for to_sql's dtype arg to receive sqlalchemy types
- some code specific to presto around the codebase doing just this - `DateTime if "presto" else string` ...

justinpark:
@mistercrunch Then, what do you think about changing it to access via `repr`?
In the case of SQLAlchemy, `repr` contains information that matches the `longType` better than `str`, and I don't think it's risky to convert this part used for `longType` to `repr()`.
What do you think?

rusackas:
@justinpark @mistercrunch see a way forward with this one?

superset/databases/utils.py
```diff
@@ -55,7 +55,7 @@ def get_indexes_metadata(
 
 def get_col_type(col: dict[Any, Any]) -> str:
     try:
-        dtype = f"{col['type']}"
+        dtype = repr(col["type"])
```
mistercrunch, line 58: noting that better/more-specific types in the method signature would help understand what we're dealing with here.

rusackas:
Any hopes of moving this forward @justinpark or should we move it to draft?

rusackas (approved):
This looks right to me, @justinpark — `repr()` preserves the ROW/MAP structure in `longType` and the tests cover it. Happy to approve.

PR #31596, opened 2024-12-22
feat: reduce some filter width in list views

mistercrunch:
Currently in our CRUD list views (dashboards, charts, datasets) we use a fixed width of 200px for all filters, regardless of their content. This leads to some filter rows overflowing depending on use screen width. Here I'm setting some of the filters to be thinner, preventing filters from wrapping into 2 rows.

In the future version of antd, we'll be able to shrink all filters to thinner width with `popupMatchSelectWidth={false}`, allowing for the dropdown width to be wider than the input box.

### before
Screenshot 2024-12-23 at 1 09 14 AM

### after
Screenshot 2024-12-23 at 1 04 27 AM

### DIFF
--- a/superset-frontend/src/pages/ChartList/index.tsx
+++ b/superset-frontend/src/pages/ChartList/index.tsx
@@ -562,6 +562,7 @@ function ChartList(props: ChartListProps) {
         { label: t('Yes'), value: true },
         { label: t('No'), value: false },
       ],
+      width: 100,
     }),
     [],
   );
@@ -600,6 +601,7 @@ function ChartList(props: ChartListProps) {
 
             return 0;
           }),
+        width: 150,
       },
       {
         Header: t('Dataset'),
@@ -669,6 +671,7 @@ function ChartList(props: ChartListProps) {
           { label: t('Yes'), value: true },
           { label: t('No'), value: false },
         ],
+        width: 100,
       },
       {
         Header: t('Modified by'),
--- a/superset-frontend/src/pages/ChartList/index.tsx
+++ b/superset-frontend/src/pages/ChartList/index.tsx
@@ -647,6 +647,7 @@ function ChartList(props: ChartListProps) {
           props.user,
         ),
         paginate: true,
+        width: 150,
       },
       {
         Header: t('Dashboard'),
--- a/superset-frontend/src/components/ListView/Filters/Base.ts
+++ b/superset-frontend/src/components/ListView/Filters/Base.ts
@@ -25,11 +25,17 @@ export interface BaseFilter {
   initialValue: any;
 }
 
-export const FilterContainer = styled.div`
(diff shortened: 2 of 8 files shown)


sadpandajoe:
@kasiazjc take a look and see if there are any other changes you would want to be made here.

superset-frontend/src/pages/ChartList/index.tsx
```diff
@@ -645,6 +647,7 @@ function ChartList(props: ChartListProps) {
           props.user,
         ),
         paginate: true,
+        width: 150,
```
michael-s-molina, line 650: Owner contain people names which require more width.
```suggestion
```

mistercrunch: I can bump it back to 200px but was trying to save some more space to fit things in a single row. Arguably many of the dropdowns need more dropdown width than in the pop-down list (datasets, charts, dashboard names can be way beyond 200px). Luckily they all have a title prop which means hovering for a moment will show the full name. 

I think antd v5 will make it easier to have wider with in the popdown than in the trigger/input

kasiazjc:
> Currently in our CRUD list views (dashboards, charts, datasets) we use a fixed width of 200px for all filters, regardless of their content. This leads to some filter rows overflowing depending on use screen width. Here I'm setting some of the filters to be thinner, preventing filters from wrapping into 2 rows.
> 
> In the future version of antd, we'll be able to shrink all filters to thinner width with `popupMatchSelectWidth={false}`, allowing for the dropdown width to be wider than the input box.
> 
> ### before
> Screenshot 2024-12-23 at 1 09 14 AM
> ### after
> Screenshot 2024-12-23 at 1 04 27 AM

Is this still a fixed width or it fills in the space it has? I like the ideas of keeping the filters in one line, but different widths in here make it a little bit messy. I would consider (probably not eng language, so let me know if it needs clarification): 

- making with of the fields flexible - it fills the space it has based on the width of the space/screen (it will be different for 1440px and 1720px) so it is all in one line
- adding min width for fields - for example 100px, if it doesn't fit in one line, then we wrap

as @mistercrunch said, we have values visible on hover so it for sure helps 

second solution which is notable more time consuming is implementing "compact" horizontal filter bar. We've heard from customers that depends on the dashboard they could use compact or current horizontal filter bar (it would be an option in dashboard). For CRUDs etc we could use compact - preview below. 

image
image


Thoughts on both? @mistercrunch @michael-s-molina

mistercrunch:
Feels like the best solution here is to have wider with for the popdown content than the input box, not sure why it's so hard. Will give it another shot. 

Could also try and use flex with min-width / max-width

mistercrunch:
Closing and hoping to use more compact widths once we upgrade <Select/> to antd v5

PR #31636, opened 2024-12-28
chore(deps): Updated jinja2 from 3.1.4 to 3.1.5

chickahoona:
### SUMMARY
jinja2 @ 3.1.4  is vulnerable to [GHSA-q2x7-8rv6-6q7h](https://github.com/pallets/jinja/security/advisories/GHSA-q2x7-8rv6-6q7h) and [GHSA-gmj6-6f8f-6699](https://github.com/pallets/jinja/security/advisories/GHSA-gmj6-6f8f-6699) which were addressed with 3.1.5


mistercrunch:
you'll want to run `pip-compile-multi -P jinja2` or CI is going to fail, could also be good to bump the lower bound here -> https://github.com/apache/superset/blob/master/pyproject.toml#L64 .

I just realized that `jinja2` isn't listed as a direct dependency in our `pyproject.toml` though it should be since we use it directly in the package. Guessing we never added it since it's a sub-dependency of Flask and other packages we import.

Since we use it directly, let's add an entry:
```
# 3.1.4 has a known security vulnerability
"jinja2>=3.1.5",
```

You'll want to:
1. add this line to pyproject.toml
2. run `pip-compile-multi -P jinja2`
3. commit and push

chickahoona:
Followed your instructions, but the change seems oddly huge ... if you want to take over this endeavor of upgrading jinja and dismissing my PR, I'd have no hard feelings ;)

mistercrunch:
Ooops, we changed our package manager since I commented (from `pip-compile-multi` to `uv`). Links to docs -> https://superset.apache.org/docs/contributing/development/#dependencies

In short, you'll have to `git reset`  to the commit prior to applying `pip-compile-multi`, and run `./scripts/uv-pip-compile.sh -P jinja2`. Sorry about the confusion!

mistercrunch (changes requested):
Please use `./scripts/uv-pip-compile.sh -P jinja2` to regenerate the reqs files as we've changed package manager. This command requires those reqs files to be in the right state so you'll have to reset to a previous commit or checkout those files from the latest `master`

chickahoona:
I am sorry but no. I just wanted to bump a version for a quick security fix, not fixing pipelines, running scripts, adding dependencies somewhere to please whoever.

PR #31688, opened 2025-01-02
fix(helm): change values.yaml comments

sule26:
### SUMMARY

Change the comment which says extraConfigs is mounted in `/app/pythonpath`, when it is actually mounted in `/app/configs`
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


villebro:
@sule26 you need to regenerate the `helm-docs` so they reflect the updated path.

sule26:
@villebro Done!

helm/superset/README.md
```diff
@@ -66,7 +66,7 @@ On helm this can be set on `extraSecretEnv.SUPERSET_SECRET_KEY` or `configOverri
 | envFromSecret | string | `"{{ template \"superset.fullname\" . }}-env"` | The name of the secret which we will use to populate env vars in deployed pods This can be useful for secret keys, etc. |
 | envFromSecrets | list | `[]` | This can be a list of templated strings |
 | extraConfigMountPath | string | `"/app/configs"` |  |
-| extraConfigs | object | `{}` | Extra files to mount on `/app/pythonpath` |
+| extraConfigs | object | `{}` | Extra files to mount on `/app/configs` |
```
villebro, line 69: While we're at it, maybe we can update the description for both `extraConfigs` and `extraSecrets` to explain that the mount points aren't static, but will in fact be mounted on the path defined in `extraConfigMountPath` and `configMountPath`?

villebro (commented):
LGTM, but I'd prefer to touch up the comments some more while we're at it

helm/superset/README.md
sule26: I think the part about mounting paths could use some clarification.

`extraConfigMountPath` is exclusively for mounting extraConfigs.
`configMountPath` , on the other hand, is where both `extraSecrets` and `configOverrides` will be mounted.

This distinction might not be immediately obvious and could lead to confusion. It might help to make this more explicit in the documentation or code comments.

helm/superset/values.yaml
```diff
@@ -106,7 +106,7 @@ extraSecretEnv: {}
   #   # Generate your own secret key for encryption. Use openssl rand -base64 42 to generate a good key
   #  SUPERSET_SECRET_KEY: 'CHANGE_ME_TO_A_COMPLEX_RANDOM_SECRET'
 
-# -- Extra files to mount on `/app/pythonpath`
+# -- Extra files to be mount as ConfigMap on the path specified in `extraConfigMountPath`
```
villebro, line 112:
```suggestion
# -- Extra files to be mounted as ConfigMap on the path specified in `extraConfigMountPath`
```

helm/superset/values.yaml
```diff
@@ -120,10 +124,11 @@ extraConfigs: {}
   #       sqlalchemy_uri: example://example-db.local
   #       tables: []
 
-# -- Extra files to mount on `/app/pythonpath` as secrets
+# -- Extra files to be mount as Secrets on the path specified in `configMountPath`
```
villebro, line 127:
```suggestion
# -- Extra files to be mounted as Secrets on the path specified in `configMountPath`
```

villebro (commented):
Two suggestions (typos). Also, maybe clarify in the description that there's some general cleanup to harmonize quotes etc (how were these applied btw: automatically or manually?)

helm/superset/values.yaml
sule26: Honestly, some changes were made automatically when I saved the file in VS Code. I'm planning to revert some of them, but I think a few might be beneficial since many projects use this format.

Before:
```yaml
test:
  {}
```

Now:
```yaml
test: {}
```

I believe the updated format is more concise and aligns better with common YAML conventions. Let me know your thoughts.

villebro: I think this makes sense. In the long term it'd be a good idea to have format linting for this file, but for now I feel this is a step in the right direction.

villebro (approved):
LGTM, thanks for all the iterations!

Issue #30334, opened 2024-09-19
Issues with Language Settings in Superset: Configuration Not Saved, Partial Translations, and Redirects


aprendizlaura:
I am experiencing multiple issues related to the language settings in Superset:

1. **Partial Translations and Incorrect Translations**: While testing different languages, I noticed that not all parts of Superset are translated when switching from English to another language. This issue is compounded by incorrect translations, such as those shown in the attached image, where both ascending and descending order options are labeled as "orden descendente" in Spanish. This forces me to revert to English to differentiate between options. I would prefer that parts of Superset remain untranslated rather than being mistranslated.
![Captura de pantalla 2024-09-19 103056](https://github.com/user-attachments/assets/05816252-ca38-4b3c-81ab-20dae4995877)
2. **Language Configuration Not Saved**: When changing the language settings in Superset, the changes are not saved upon closing and reopening the application. For example, if the default language is English and I change it to Spanish, Superset reverts back to English after restarting. Additionally, if I open Superset in a different browser after setting it to Spanish, it also defaults back to English. This becomes problematic when multiple users are involved. If User 1 prefers Spanish and User 2 prefers English, only one default language can be set, requiring one user to change the language every time they log in. The expected behavior is that once a user changes the language setting, it should persist across sessions.
3. **Redirect to Homepage**: When changing the language, Superset redirects the user to the homepage, regardless of the current page. Although I understand that a restart might be necessary for language changes to take effect, it would be more user-friendly if the user remained on the page they were working on after the restart.

Suvendu-UI:
is this issue resolved?

aprendizlaura:
@Suvendu-UI No, it hasn't been resolved. The issue that concerns me the most is the first one, and even with the configuration that Dosu suggested (which I already had because I read other threads), it still works partially.

rusackas:
Partial translations are not considered bugs, as we'll be chasing that down approximately forever. If you'd like to contribute or fix translations, we encourage you to edit the `.po` files and open a PR with your changes.

rusackas:
I haven't looked much at the other two issues, but the second one sounds legit, and maybe you can set a cookie or use localstorage to resolve this. The third, I believe, is a duplicated of another open issue. If you or I find that, we _might_ want to close this as a duplicate... I'd encourage you to file issues separately for reasons like this :)

aprendizlaura:
[...]

PR #31692, opened 2025-01-03
fix(lang): patch FAB's LocaleView to redirect to previous page

### LINKED ISSUE #299: Superset i18n: redirect back to previous page after changing language setting
### Description

The original issue states:

> When changing the language, Superset redirects the user to the homepage, regardless of the current page. Although I understand that a restart might be necessary for language changes to take effect, it would be more user-friendly if the user remained on the page they were working on after the restart.

However there seems to be a "previous page" stored in the user's session somewhere, because after visiting the User Info page (`/users/userinfo/`), this became my go-to page instead of the home page for redirects after changing my language.

### Supporting Information

* Part 3 of https://github.com/apache/superset/issues/30334
* Also reported here: https://github.com/apache/superset/issues/12768

### Proposed solution

Investigate further, and choose an appropriate option.

**Option 1: Fix "previous page" setting**

* Determine where the "previous page" setting is stored and updated in the user's session
* Ensure this setting gets updated whenever the "current page" changes.

**Option 2: Store "next page" as URL param on language change**

* As with the "login" screen, store the "current page" path as a `next` URL parameter when changing the user's preferred language
* Ensure this `next` parameter is used when deciding where to redirect after changing the user's preferred language.

### Aspects specifics

- [ ] Ensure the solution is backported to the Superset version used by Aspects. ❌ see [comment](https://github.com/openedx/openedx-aspects/issues/300#issuecomment-2767884629)

### LINKED ISSUE #12768: Redirect happens whenever user changes language
When the user presses on change language a redirect to `/users/list` happens. I think, in terms of UX, it should remain on the same page as before changing the language.

### Expected results

Remain on the same page, but translated into a different language.

### Actual results

Redirects to `users/list`.

#### Screenshots

gif-gif

#### How to reproduce the bug

1. Enable languages
2. Open any page
3. Change the language

### Environment

- superset version: `0.999.0dev`
- python version: `3.7.4`
- node.js version: `v14.11.0`

### Checklist

pomegranited:
### SUMMARY
Overrides FAB's LocaleView to redirect to the request's referrer instead of relying on the session page history as [suggested here](https://github.com/apache/superset/issues/12768#issuecomment-2376141202).

This isn't a perfect solution, but it keeps the change very close to the actual issue.

Note that this view is marked out in the tests as an "unsecured" view, because we need to be able to change the current locale even when logged out (so that the login page can be presented in the user's preferred language).

**Rejected alternatives**

1. Call `update_redirect()` whenever a GET view is rendered ([like this](https://github.com/apache/superset/compare/master...open-craft:superset:jill/update_redirect)).
   This is more in line with how Flask AppBuilder works (and was [recommended here](https://github.com/dpgaspar/Flask-AppBuilder/pull/1679#discussion_r702709749)).
   However, this only works for pages rendered by Superset's backend, and most Superset pages are rendered by the frontend.
2. Hook into the [frontend's action logging](https://github.com/apache/superset/blob/7c903236498281f34970012e701250edef1b1493/superset-frontend/src/views/App.tsx#L54-L69) to update the page history, e.g when a `LOG_ACTIONS_SPA_NAVIGATION` event is logged to the backend.
  This approach felt even hackier than the one posted here, and farther from the issue it addresses.
  Also, Flask AppBuilder only provides the `update_redirect()` method which uses the current `request.url`, so we'd need to add another method that could use a different `url`.

Fixes: https://github.com/apache/superset/issues/12768
Part of: https://github.com/apache/superset/issues/30334
Fixes: https://github.com/openedx/openedx-aspects/issues/299

Before:

https://github.com/user-attachments/assets/7bcdefa5-46d5-4173-98cc-884384af233d

After:

https://github.com/user-attachments/assets/70b32340-7f23-482d-a33b-8347aa5000b0

### TESTING INSTRUCTIONS

1. Load Superset with multiple languages enabled.
2. Visit any page.
3. Change your current language using the language switcher menu on the top right.
4. Ensure that your language is changed and you're redirect back to the page in step 2.

### ADDITIONAL INFORMATION

- [x] Fixes https://github.com/apache/superset/issues/12768
- [ ] Required feature flags:
- [ ] Changes UI


superset/initialization/__init__.py
```diff
@@ -701,3 +701,21 @@ class SupersetIndexView(IndexView):
...
+        """
+        Change user's locale and redirect back to the previous page.
+
+        Overrides FAB's babel.views.LocaleView so we can use the request
+        Referrer as the redirect target, in case our previous page was actually
+        served by the frontend (and thus not added to the session's page_history
+        stack).
+        """
+        if locale not in self.appbuilder.bm.languages:
+            abort(404, description="Locale not supported.")
```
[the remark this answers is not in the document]
pomegranited: Leaving this one up to the reviewers - see original [LocaleView](https://github.com/dpgaspar/Flask-AppBuilder/blob/418ab8a93907669be4ccbb99d7aefa5283f3e013/flask_appbuilder/babel/views.py#L7-L19)

rusackas:
Installing the pre-commit hook should resolve at least some of the CI issues. 

/
```
pre-commit install
```
```
pre-commit run --all-files
```

superset/initialization/__init__.py
```diff
@@ -701,3 +702,34 @@
...
+        session["locale"] = locale
+        refresh()
+        self.update_redirect()
+
+        redirect_to = request.headers.get("Referer")
+        if not redirect_to or not self.is_safe_url(redirect_to):
+            redirect_to = self.get_redirect()
+        return redirect(redirect_to)
```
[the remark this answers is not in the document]
pomegranited: Leaving this one up to the reviewers - see original [LocaleView](https://github.com/dpgaspar/Flask-AppBuilder/blob/418ab8a93907669be4ccbb99d7aefa5283f3e013/flask_appbuilder/babel/views.py#L7-L19)

pomegranited:
Hi @rusackas :) Is there anything more I can do here to get this PR approved?

rusackas:
I was hoping someone more FABtastic than me would come along to review this. If not @dpgaspar, perhaps @mistercrunch can take a look.

pomegranited:
Hi @jpchev , would you like to review this PR for me? Might help get it merged if there's some community testing/support :)

jpchev:
> Hi @jpchev , would you like to review this PR for me? Might help get it merged if there's some community testing/support :)

yes, I've tested the changes and can confirm I am not redirected to another page after switching language

superset/initialization/__init__.py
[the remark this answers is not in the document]
dpgaspar: we only redirect to the referer after checking with `self.is_safe_url` so I think this one is not correct by CodeQL.

dpgaspar (commented):
Looks good, would be great if you could add a couple of tests

superset/initialization/__init__.py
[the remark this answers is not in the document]
pomegranited: Ah.. that makes sense, thanks! See [3dc36a2](https://github.com/apache/superset/pull/31692/commits/3dc36a2e309e3af8812cacdc63e8e2698dec4c01).

pomegranited:
hi @dpgaspar , is there anything else you need from me before this can be merged?

pomegranited:
Hi @rusackas , I've applied @dpgaspar 's suggestions, but this has been needing another review/approval for a while now.. What can I do to get this merged? Would love to make the next release candidate.

pomegranited:
@mistercrunch Can you help me get this PR merged?

superset/initialization/__init__.py
```diff
@@ -701,3 +703,34 @@ class SupersetIndexView(IndexView):
     @expose("/")
     def index(self) -> FlaskResponse:
         return redirect("/superset/welcome/")
+
+    @staticmethod
+    def is_safe_url(target: str) -> bool:
```
mistercrunch, line 708: might belong in `superset/utils/urls.py` would be surprised if we don't already have a similar method there, grepping for `urlparse` or `next=` could point to a similar method.

mistercrunch: otherwise PR LGTM

pomegranited: Thank you @mistercrunch ! I didn't find a similar method anywhere, so [4bb4916](https://github.com/apache/superset/pull/31692/commits/4bb49161c04f4cedeaeba30d9fabf9889fd9497c) creates `superset.utils.urls.is_safe_redirect_url(source_url, target_url)`.

Does this look ok now?

superset/initialization/__init__.py
```diff
@@ -701,3 +703,34 @@
...
+    def patch_flask_locale(self, locale: str) -> FlaskResponse:
+        """
+        Change user's locale and redirect back to the previous page.
+
+        Overrides FAB's babel.views.LocaleView so we can use the request
+        Referrer as the redirect target, in case our previous page was actually
+        served by the frontend (and thus not added to the session's page_history
+        stack).
+        """
+        if locale not in self.appbuilder.bm.languages:
```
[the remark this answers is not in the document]
pomegranited: Mitigated by https://github.com/apache/superset/pull/31692/files#diff-ab057099067874ff6f10f77be2e3e6ed55ba08b7f743a40d5f3905fb836a7ee0R67

mistercrunch (approved):
LGTM - thank you for the contribution!

pomegranited:
Hi @rusackas , could this be merged soon?

superset/utils/urls.py
```diff
@@ -62,3 +62,20 @@ def is_secure_url(url: str) -> bool:
     """
     parsed_url = urlparse(url)
     return parsed_url.scheme == "https"
+
+
+def is_safe_redirect_url(source_url: str, target_url: str) -> bool:
```
mistercrunch, line 67: sorry for the multi-phase review, but GPT recommends further checking here, and since this is security for XSS I'm thinking let's do it...

```
def is_safe_redirect_url(source_url: str, target_url: str) -> bool:
    if not target_url:
        return False
    joined = urljoin(source_url, target_url)  # resolves relative URLs
    parsed_source = urlparse(source_url)
    parsed_target = urlparse(joined)
    return (
        parsed_source.scheme == parsed_target.scheme and
        parsed_source.hostname == parsed_target.hostname
    )
```
GPT says: This handles edge cases like user-supplied target_url values starting with // (which browsers interpret as external redirects) or other relative path tricks. Using urljoin() ensures we're validating the fully resolved URL against the expected scheme and host. Safer for open internet exposure.

rusackas:
@pomegranited just approved CI and it looks like this is mergeable pending feedback/action on the last comment from @mistercrunch.

rusackas (dismissed):
I was getting itchy over the merge button, so I'll block this awaiting feedback/implementation of the security-based code suggestion provided by @mistercrunch. CC @dpgaspar / @Antonio-RiveroMartnez for any other additional feedback or suggestions on that approach.

mistercrunch:
We just need to be super careful around the `?next={some_url}` pattern. Might be a good time to review what we already have here, and merge all codepaths that use this pattern into a state-of-the-art function that handles it all.

Couldn't find the pattern in the codebase after a quick search, maybe FAB exposes something we piggy-back on? @dpgaspar

superset/utils/urls.py
pomegranited: Easy done, thank you for this security fix @mistercrunch ! Applied with [08010c9](https://github.com/apache/superset/pull/31692/commits/08010c9755ac5485a3480d615437d28cafacdcc7)

However, should we to use the joined `target_url` when doing the actual redirect?

mistercrunch (approved):
LGTM

mistercrunch:
Still would love for @dpgaspar to chime in, maybe FAB has builtin functions for that already

superset/utils/urls.py
pomegranited: Ok, I've found what we needed in FAB: [get_safe_redirect](https://github.com/dpgaspar/Flask-AppBuilder/blob/6d44e6d581433dcea475764c4bb1270c24bbd6de/flask_appbuilder/utils/base.py#L36). My only worry is it's not exposed as an explicit part of the API, and so might change out from under us. But it's used in several places in the FAB code base, so it's a calculated risk.

pomegranited:
Hi @rusackas I found a FAB method to do what we need, so I think this resolves @mistercrunch 's concern?

dpgaspar:
> Hi @rusackas I found a FAB method to do what we need, so I think this resolves @mistercrunch 's concern?

Nice! yes `is_safe_redirect_url` or `get_safe_redirect`

dpgaspar (approved):
Nice, not a perfect solution but an improvement

pomegranited:
Hi @rusackas , I think we're good to merge now?

pomegranited:
Hi @rusackas , I think I've addressed everything here, and we got an approval from our FAB expert.. can this be merged?

mistercrunch:
hitting a weird GH bug I've never seen before ->
Screenshot 2025-04-15 at 9 34 10 AM

Will try again a little later...

pomegranited:
Yay!! Thank you @mistercrunch @rusackas  @dpgaspar for your help here!

PR #31726, opened 2025-01-06
feat(filter): adding inputs to Numerical Range Filter

### LINKED ISSUE #31345: Update filter numerical range slider to user input fields
## Screenshot

image

## Description

The slider as a filter isn't an intuitive pattern and it's hard to set the filter values at a large value. Move to two input boxes instead (see below) 

## Design input
image

alexandrusoare:
### SUMMARY
Adding inputs to the Numerical Range Filter

Fixes #31345

BEFORE
image

AFTER
image

BEFORE
image

AFTER
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

Add numerical input fields to the Numerical Range Filter and update tests to reflect and verify this new functionality.

The current slider interface is being replaced with numerical input fields to improve precision and user control in defining numerical ranges. This enhancement also ensures that the inputs sync to maintain valid ranges dynamically, improving usability. The accompanying tests are updated to ensure the new input behavior works as expected and edge cases are properly handled.


villebro:
@alexandrusoare would it be a good idea to make the inputs optional? Especially in horizontal filter bar orientation the added inputs may be problematic.

alexandrusoare:
> @alexandrusoare would it be a good idea to make the inputs optional? Especially in horizontal filter bar orientation the added inputs may be problematic.

@villebro I think having both the slider and the inputs offer greater control to the user, it's hard setting values just from the slider alone. For the horizontal bar I do agree that it doesn't look good, so what I propose is to have the slider next to the inputs instead of having them one on top of the other. Let me know what you think about this

michael-s-molina:
Reviewers: this PR just left draft state.

sadpandajoe (commented):
👍 on the cypress test. I'll leave the other two files for people that have more experience with those components.

superset-frontend/src/filters/components/Range/RangeFilterPlugin.tsx
```diff
@@ -25,113 +25,32 @@ import {
...
-      };
-    }
-
-    .ant-slider-track {
-      background-color: ${DARK_GRAY};
-    }
-  }
-  `}
+const StyledDivider = styled.span`
+  margin: 0 10px;
```
kgabryje, line 36: nit: could we use `{theme.gridUnit * 2}px` instead of 10 here?

geido:
I am still seeing a bunch of different issues:

- When selecting the Single value option the filter goes out of scope for some reasons
- When working with the default value I see the value being reset to zero and sometimes I have also seen it blink from the lowest value to the highest, not sure

https://github.com/user-attachments/assets/705c8a98-02f1-45c5-8192-ae4e570cde9b

alexandrusoare:
> I am still seeing a bunch of different issues:
> 
> * When selecting the Single value option the filter goes out of scope for some reasons
> * When working with the default value I see the value being reset to zero and sometimes I have also seen it blink from the lowest value to the highest, not sure
> 
>  Tabbed.Dashboard.mp4

Hey @geido , thank you for your feedback. Regarding the first issue, this is something that I could reproduce on master as well, and with all the types of filters, not just the Numerical Range one

rusackas:
Seems like on the filter's configuration modal, there could be checkbox or radio buttons to pick 
```(•) slider ( ) inputs ( ) both```

geido:
> Seems like on the filter's configuration modal, there could be checkbox or radio buttons to pick `(•) slider ( ) inputs ( ) both`

We have got feedback from @kasiazjc that inputs are the preferred approach as the slider tends to give a worst UX experience. I guess we'll try without this option and if people feel strongly about the slider we can bring it back.

kgabryje (approved):
Thanks for the contribution!
I tested the PR, and while the filter goes to "out of scope" section after changing between single and multi value, the same bug is already present on master - let's fix it in a separate PR

villebro:
I added the "hold!" label - there's a few things I want to check here as we're making a major change here. I'll drop my review shortly and remove the hold label after that to unblock.

villebro:
@alexandrusoare I don't fully agree with the new UX. Currently when the filter renders, it shows the min and max values as set, despite them not applying to any charts (see how the filter indicator is unset):

image

In this example, the minimum and maximum values of the reference column are 5 and 37296 respectively. When I bump the lower bound from 5 to 6, it starts applying the lower bound (setting it explicitly to a value that's lower than the min value removes the bound):

image

The same happens for the upper bound.

It feels like it would be more intuitive if the min/max values were shown as placeholders when the lower/upper bounds are unset, and only when a value is entered, the value would apply. With this change it would also be possible to explicitly set a lower/upper bound when it coincides with the min/max.

Thoughts? Ping @kgabryje @kasiazjc

kgabryje:
I agree with @villebro that using a placeholder might be a better solution. Also, I'm not sure about the value resetting to minimal/maximal if I type a lower/higher value. Perhaps some kind of info tooltip like "This value is too low/high and won't be applied"? (of course that's sounds horrible, but hopefully ChatGPT would help refine it 😄)

kasiazjc:
> I agree with @villebro that using a placeholder might be a better solution. Also, I'm not sure about the value resetting to minimal/maximal if I type a lower/higher value. Perhaps some kind of info tooltip like "This value is too low/high, the filter won't have an effect"? (of course that sounds horrible, but hopefully ChatGPT would help refine it 😄)

Agree with @villebro that there should be placeholders and honestly I was sure we had this flow (with value is too/low high, change it so that it is in the range) somewhere, but it might have stayed in one of the docs 😅 So I agree with you both and I think that that should be the path forward

geido:
Hello. I have noticed this bug from manual testing:

When adding default values, leaving the Dashboard and then coming back shows the inputs empty instead of the default values.

geido (approved):

PR #31732, opened 2025-01-06
fix: docker builds in forks

### LINKED ISSUE #31731: Build and push docker steps may not always work for branches
### Bug description

When working with a branch in Superset sometimes the build and push to docker steps may not always be successful. We can see these in both dependabot branches and some committer branches, however, post-merge the job seems to run fine.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

mistercrunch:
Recently observed an issue where the GHA for docker builds fails when the PR is opened from a fork (as opposed to the main repo itself). This is related to the fact that the job tried to push the image to dockerhub, and doesn't have the creds `${{ secrets.DOCKERHUB_TOKEN }}` in that context.

Now there are some intricacies here:
- we should only push images on `push` events on the main branch (we don't need to publish images for every PR)
- we can only `--load` when building single-platform manifests, docker will just fail if you try and load the image from the wrong platform, --load is just not supported for multi-platform builds
- so, in push-to-master context, we simply multi-platform builds to docker hub, and then `docker pull` right after so that we can use the image in subsequent steps

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -77,16 +77,12 @@ jobs:
           fi
 
           supersetbot docker \
-            --push \
+            --load \
             --preset ${{ matrix.build_preset }} \
             --context "$EVENT" \
             --context-ref "$RELEASE" $FORCE_LATEST \
             --extra-flags "--build-arg INCLUDE_CHROMIUM=false" \
             $PLATFORM_ARG
-
-      - name: Docker pull
-        if: steps.check.outputs.python || steps.check.outputs.frontend || steps.check.outputs.docker
-        run:  docker pull apache/superset:GHA-${GITHUB_RUN_ID}
 
       - name: Print docker stats
         if: steps.check.outputs.python || steps.check.outputs.frontend || steps.check.outputs.docker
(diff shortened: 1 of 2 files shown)


villebro (approved):
LGTM!

superset/config.py
```diff
@@ -12,8 +12,8 @@
 # software distributed under the License is distributed on an
 # "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 # KIND, either express or implied.  See the License for the
-# specific language governing permissions and limitations
-# under the License.
+# specific language governing permissions and limitations under
+# the License.
```
villebro, line 16: Maybe this change was done unintentionally?

mistercrunch: oh I needed to trigger the build... let me change it back

.github/workflows/docker.yml
```diff
@@ -37,6 +37,7 @@ jobs:
     env:
       DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
       DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      IMAGE_TAG: apache/superset:GHA-${{ matrix.build_preset }}-${{ github.run_id }}
```
mistercrunch, line 40: turns out `run_id` isn't unique in the context of a matrix...

sadpandajoe:
@mistercrunch couple of questions from the description:
> we should only push images on push events on the main branch (we don't need to publish images for every PR)
- Would this affect the release branches we have? My guess is that those tags should still be pushed to docker.
- For ephemerals, we don't currently download the docker image to run, right?

mistercrunch:
> Would this affect the release branches we have? My guess is that those tags should still be pushed to docker.

Nope, there's a different workflow for releases that's similar and trigger on-release events and workflow-dispatch (manually)

> For ephemerals, we don't currently download the docker image to run, right?

There's also another isolated workflow for that that's triggered on-comment


The good thing with the recent changes that broke what I'm fixing here is that we have a much more comprehensive/DRY `setup-docker` reusable action now that all these 3 workflows use

PR #31747, opened 2025-01-08
feat: improve docker-compose services boot sequence

mistercrunch:
Recently I noticed that despite the service dependencies state in our docker-compose files, the all start on top of each other, which isn't good or as expected. We want for the `superset-init` service/scripts to finish prior to starting other services, and ideally services being in a healthy state before booting up the others that depend on them. There are newish semantics in docker-compose to enable this, so will implement here in this PR.

Improvements on this PR:
- `superset-init` waits for redis and db to be started (not really an issue, but cleaner)
- `superset-app` and worker-related stuff waits on `superset-init` to be fully completed and successful, so the app shouldn't be there and confused until the superset's database is ready (setup, migrations, load_examples, ...)
- removing `--force` from loading examples (this mostly forces a data reload from my understanding), and we really don't need to reload at the examples every time we fire up compose! this saves a bunch of work and log output
- somehow docker-compose doesn't see `print()` outputs, so moving some example-related stuff to `logger.info` and `logger.debug` to clean the startup output and see what it's actually doing
- moving default log level in dev to `info` (from `debug`)

Reorder Docker Compose service startup sequence and enhance logging by implementing custom conditions for service dependencies, introduce environment variable configuration for logging levels, and refactor logging statements from print to use the logger across various example scripts.

The changes ensure that services like `superset-init` start only after dependent services are ready, reducing startup issues. Additionally, the enhanced logging options improve transparency and troubleshooting during the service lifecycle. These adjustments increase reliability and provide better insights into the service operation without print statements, utilizing configurable logging levels instead.


mistercrunch:
Noticing here too that example load/reload is slow. Looking to do some work on making example reload super fast, meaning if you already loaded up the examples, upserting them should be extremely quick.

docker-compose-image-tag.yml
```diff
@@ -22,9 +22,6 @@
 # unique random secure passwords and SECRET_KEY.
 # -----------------------------------------------------------------------
 x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest-dev}
-x-superset-depends-on: &superset-depends-on
```
Antonio-RiveroMartnez, line 25: nit: We could keep the anchor and just update the values of it with the new conditions you have and also create a new one for the `superset-init: condition: service_completed_successfully` so we can then reuse and refer to these

superset/utils/encrypt.py
```diff
@@ -22,7 +22,12 @@
 from flask_babel import lazy_gettext as _
 from sqlalchemy import text, TypeDecorator
 from sqlalchemy.engine import Connection, Dialect, Row
-from sqlalchemy_utils import EncryptedType
+from sqlalchemy_utils import EncryptedType as SqlaEncryptedType
+
+
+class EncryptedType(SqlaEncryptedType):
```
Antonio-RiveroMartnez, line 28: Why is this needed?

mistercrunch: oh! finally took on removing WARNING log message that became apparent while I paid attention to the boot sequence's STDOUT

docker-compose-image-tag.yml
mistercrunch: As much as I like DRY code, I felt like in yaml it was mostly just an indirection as there's not a lot that's repeated here. Personally prefer have it spelled out inline as opposed to having to look up a thing that has a few line and is repeated a few times.

mistercrunch: Could see it both ways though, let me know what you think

Antonio-RiveroMartnez: Yeah, I don't have a strong preference, so we can kept it spelled out inline

Antonio-RiveroMartnez (approved):
lgtm

eschutho (approved):

PR #31754, opened 2025-01-08
refactor: Removes Apply to all panels filters scope configuration

michael-s-molina:
### SUMMARY
This PR implements the 5.0 proposal called [87 - Remove Apply to all panels filters scope configuration](https://github.com/orgs/apache/projects/345?pane=issue&itemId=84839864) to make it explicit which charts will be affected when setting filter scopes.

https://github.com/user-attachments/assets/f6c07d75-40f7-4ca1-8fdc-384205c3b26c

### TESTING INSTRUCTIONS
Check that filter scopes continues to work as expected.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/FilterScope.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/FilterScope.tsx
@@ -17,13 +17,11 @@
  * under the License.
  */
 
-import { FC, useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { NativeFilterScope, styled, t } from '@superset-ui/core';
-import { Radio } from 'src/components/Radio';
-import { AntdForm, Typography } from 'src/components';
-import { ScopingType } from './types';
+import { FC, useCallback, useEffect, useMemo, useState } from 'react';
+import { NativeFilterScope, styled } from '@superset-ui/core';
+import { AntdForm } from 'src/components';
 import ScopingTree from './ScopingTree';
-import { getDefaultScopeValue, isScopingAll } from './utils';
+import { getDefaultScopeValue } from './utils';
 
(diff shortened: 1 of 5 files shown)


superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/FilterScope.tsx
```diff
@@ -98,49 +83,20 @@ const FilterScope: FC<FilterScopeProps> = ({
...
-        >
-          <Radio value={ScopingType.All}>{t('Apply to all panels')}</Radio>
-          <Radio value={ScopingType.Specific}>
-            {t('Apply to specific panels')}
-          </Radio>
-        </Radio.Group>
-      </CleanFormItem>
-      <Typography.Text type="secondary">
-        {(formScopingType ?? initialScopingType) === ScopingType.Specific
-          ? t('Only selected panels will be affected by this filter')
```
kgabryje, line 131: Should we keep this text somewhere? In some sort of tooltip maybe?

kgabryje (approved):
LGTM, 1 non-blocking discussion item

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/FilterScope.tsx
michael-s-molina: I think this made more sense when we had the radio group to explain each option. Now that the tree is always visible it seems pretty clear that only selected panels will be affected.

michael-s-molina: Do you agree? If not, I can restore the message and put it at the top.

PR #31757, opened 2025-01-08
fix(thumbnail cache): Enabling force parameter on screenshot/thumbnail cache

fisjac:
### SUMMARY

Both the `/api/v1/chart/{{id}}/cache_screenshot/` and `/api/v1/dashboard/{{id}}/cache_dashboard_screenshot/` endpoints enable generating and retrieve a screenshot from the respective resource.

The endpoints accept sending force=true in the query parameters, but currently this parameter does nothing. It should instead invalidate the previous image and generate a new one.

This PR introduces a fix as well as some improvements to the functioning of the thumbnail cache, such that the status of the screenshot can be tracked by the relevant api endpoints as well.

Rather than storing plain bytes objects in the cache, this PR introduces a new class `ScreenshotCachePayload` which can store the image bytes as well as `timestamp` and `status` values. Status values include 
`Pending`: screenshot is awaiting being processed
`Computing`: screenshot job has been picked up by the worker and is processing
`Updated`: screenshot has been updated
`Error`: an error occured while the webdriver was processing the screenshot task.

While in a `Computing` or `Updated` state, requests to regenerate the screenshot will be blocked unless the `force` parameter is set to `True`

Additional `task_status` and `updated_at` fields have been added to 202 response payload for endpoints as follows:
```
{
  "cache_key": "ef13dd170eda0581d74a5d1c84983778",
  "dashboard_url": "http://superset_app:8088/superset/dashboard/15/",
  "image_url": "http://superset_app:8088/api/v1/dashboard/15/thumbnail/ef13dd170eda0581d74a5d1c84983778/",
  "task_status": "Pending",
  "updated_at": "2025/01/16-14:58:34"
}
```

How to reproduce the bug
Create a chart.
Perform a modification in the data.
Send a GET request to /api/v1/chart/{{id}}/cache_screenshot/?q=(force:!t).
Access the URL returned in the image_url value.

### TESTING INSTRUCTIONS

tests have been added in the `tests/unit_tests/utils/screenshot_tests.py`

For manual testing
- [ ] ping the `/api/v1/chart/<pk>/cache_screenshot/` endpoint. Ensure that a 202 response is received.
- [ ] refresh periodically until status changes to "Computing." Then again to "Updated"
- [ ] once updated, ensure the screenshot url returns the proper screenshot.
- [ ] refresh again, and ensure the status has not changed from "Updated"
- [ ] ping the endpoint again with force parameter `/api/v1/chart/cache_screenshot/<pk>/?q=(force:!t)` to ensure that the status changes to "Pending"

Repeat the steps for the `POST /api/v1/dashboard/<pk>/cache_dashboard_screenshot/`

- [ ] access the `GET /api/v1/chart/<pk>/thumbnail/<digest>` endpoint to ensure that a 202 response is received.
- [ ] access again once the status changes to "Updated" ensure that the endpoint results in a 200 response with the screenshot.
- [ ] repeat with `GET /api/v1/dashboard/<pk>/thumbnail/<digest>`
 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: THUMBNAILS, ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS
- [ ] Changes UI

### DIFF
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -621,14 +621,9 @@ class ChartRestApi(BaseSupersetModelRestApi):
                 task_status=cache_payload.get_status(),
             )
 
-        error_cache_ttl = config["THUMBNAIL_ERROR_CACHE_TTL"]
-        error_cache_expired = (
-            datetime.now()
-            - datetime.strptime(cache_payload.get_timestamp(), "%Y/%m/%d-%H:%M:%S")
-        ).total_seconds() > error_cache_ttl
         if (
             cache_payload.status == StatusValues.PENDING
-            or (cache_payload.status == StatusValues.ERROR and error_cache_expired)
+            or cache_payload.is_error_cache_ttl_expired()
             or force
         ):
             logger.info("Triggering screenshot ASYNC")
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -576,7 +576,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
                     schema:
                       $ref: "#/components/schemas/ChartCacheScreenshotResponseSchema"
             202:
-              description: Chart async task created
+              description: Chart screenshot task created
               content:
                 application/json:
                   schema:
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -621,7 +621,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
                 task_status=cache_payload.get_status(),
             )
 
-        if cache_payload.should_trigger_task() or force:
+        if cache_payload.should_trigger_task(force):
             logger.info("Triggering screenshot ASYNC")
             screenshot_obj.cache.set(cache_key, ScreenshotCachePayload())
             cache_chart_thumbnail.delay(
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -621,7 +621,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
                 update_status=cache_payload.get_status(),
             )
 
-        if cache_payload.status == StatusValues.ERROR or force:
+        if cache_payload.status in [StatusValues.PENDING, StatusValues.ERROR] or force:
             logger.info("Triggering screenshot ASYNC")
             screenshot_obj.cache.set(cache_key, ScreenshotCachePayload())
             cache_chart_thumbnail.delay(
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -30,7 +30,7 @@ from marshmallow import ValidationError
 from werkzeug.wrappers import Response as WerkzeugResponse
 from werkzeug.wsgi import FileWrapper
 
-from superset import app, is_feature_enabled, thumbnail_cache
+from superset import app, is_feature_enabled
 from superset.charts.filters import (
     ChartAllTextFilter,
     ChartCertifiedFilter,
@@ -84,7 +84,12 @@ from superset.models.slice import Slice
 from superset.tasks.thumbnails import cache_chart_thumbnail
 from superset.tasks.utils import get_current_user
 from superset.utils import json
-from superset.utils.screenshots import ChartScreenshot, DEFAULT_CHART_WINDOW_SIZE
+from superset.utils.screenshots import (
+    ChartScreenshot,
+    DEFAULT_CHART_WINDOW_SIZE,
+    ScreenshotCachePayload,
+    StatusValues,
+)
 from superset.utils.urls import get_url_path
 from superset.views.base_api import (
     BaseSupersetModelRestApi,
@@ -564,8 +569,14 @@ class ChartRestApi(BaseSupersetModelRestApi):
                 schema:
                   $ref: '#/components/schemas/screenshot_query_schema'
           responses:
+            200:
+                description: Chart async result
+                content:
+                  application/json:
+                    schema:
+                      $ref: "#/components/schemas/ChartCacheScreenshotResponseSchema"
             202:
-              description: Chart async result
+              description: Chart async created
               content:
                 application/json:
                   schema:
@@ -580,6 +591,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
               $ref: '#/components/responses/500'
         """
         rison_dict = kwargs["rison"]
+        force = rison_dict.get("force")
         window_size = rison_dict.get("window_size") or DEFAULT_CHART_WINDOW_SIZE
 
         # Don't shrink the image if thumb_size is not specified
@@ -591,25 +603,36 @@ class ChartRestApi(BaseSupersetModelRestApi):
 
         chart_url = get_url_path("Superset.slice", slice_id=chart.id)
         screenshot_obj = ChartScreenshot(chart_url, chart.digest)
-        cache_key = screenshot_obj.cache_key(window_size, thumb_size)
+        cache_key = screenshot_obj.get_cache_key(window_size, thumb_size)
+        cache_payload = (
+            screenshot_obj.get_from_cache_key(cache_key) or ScreenshotCachePayload()
+        )
         image_url = get_url_path(
             "ChartRestApi.screenshot", pk=chart.id, digest=cache_key
         )
 
-        def trigger_celery() -> WerkzeugResponse:
+        def build_response(status_code: int) -> WerkzeugResponse:
+            return self.response(
+                status_code,
+                cache_key=cache_key,
+                chart_url=chart_url,
+                image_url=image_url,
+                updated_at=cache_payload.get_timestamp(),
+                update_status=cache_payload.get_status(),
+            )
+
+        if cache_payload.status == StatusValues.ERROR or force:
             logger.info("Triggering screenshot ASYNC")
+            screenshot_obj.cache.set(cache_key, ScreenshotCachePayload())
             cache_chart_thumbnail.delay(
                 current_user=get_current_user(),
                 chart_id=chart.id,
-                force=True,
                 window_size=window_size,
                 thumb_size=thumb_size,
+                force=force,
             )
-            return self.response(
-                202, cache_key=cache_key, chart_url=chart_url, image_url=image_url
-            )
-
-        return trigger_celery()
+            return build_response(202)
+        return build_response(200)
 
     @expose("/<pk>/screenshot/<digest>/", methods=("GET",))
     @protect()
@@ -635,7 +658,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
             name: digest
           responses:
             200:
-              description: Chart thumbnail image
+              description: Chart screenshot image
               content:
                image/*:
                  schema:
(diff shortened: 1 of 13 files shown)


superset/charts/api.py
```diff
@@ -564,8 +569,14 @@ def cache_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse:
...
           responses:
+            200:
+                description: Chart async result
+                content:
+                  application/json:
+                    schema:
+                      $ref: "#/components/schemas/ChartCacheScreenshotResponseSchema"
             202:
-              description: Chart async result
+              description: Chart async created
```
villebro, line 579: nit, while we're at it:
```suggestion
              description: Chart async task created
```

superset/charts/api.py
```diff
@@ -591,25 +603,36 @@ def cache_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse:
...
+            return self.response(
+                status_code,
+                cache_key=cache_key,
+                chart_url=chart_url,
+                image_url=image_url,
+                updated_at=cache_payload.get_timestamp(),
+                update_status=cache_payload.get_status(),
+            )
+
+        if cache_payload.status in [StatusValues.PENDING, StatusValues.ERROR] or force:
```
villebro, line 624: Typically when a caching operation has errored out, we'd actually like to leave it in the cache with a specific TTL, as there's usually some access issue, or the chart is broken or something similar. Rerunning that screenshot every time is very expensive. Could we maybe introduce a config flag with an error TTL? This will likely save considerable compute on large deployments.

superset/charts/schemas.py
```diff
@@ -304,6 +304,21 @@ class ChartCacheScreenshotResponseSchema(Schema):
     image_url = fields.String(
         metadata={"description": "The url to fetch the screenshot"}
     )
+    update_status = fields.String(
+        metadata={"description": "The status of the async screenshot"}
+    )
+    updated_at = fields.String(
+        metadata={"description": "The timestamp of the last change in status"}
+    )
```
villebro, line 312: "update" seems ambiguous here:
```suggestion
    task_status = fields.String(
        metadata={"description": "The status of the async screenshot task"}
    )
    task_updated_at = fields.String(
        metadata={"description": "The timestamp of the last change in task status"}
    )
```

superset/config.py
```diff
@@ -729,6 +729,7 @@ class D3TimeFormat(TypedDict, total=False):
 
 THUMBNAIL_CACHE_CONFIG: CacheConfig = {
     "CACHE_TYPE": "NullCache",
+    "CACHE_DEFAULT_TIMEOUT": 3600,
```
villebro, line 732: we should be using `timedelta` here. Also, 3600 seems rather short for a thumbnail. Maybe a week? Or even a month?
```suggestion
    "CACHE_DEFAULT_TIMEOUT": int(timedelta(days=7).total_seconds()),
```

superset/models/slice.py
```diff
@@ -380,9 +380,7 @@ def event_after_chart_changed(
     _mapper: Mapper, _connection: Connection, target: Slice
 ) -> None:
     cache_chart_thumbnail.delay(
-        current_user=get_current_user(),
-        chart_id=target.id,
-        force=True,
+        current_user=get_current_user(), chart_id=target.id, force=True
```
villebro, line 383: Hmm, I'd maybe be ok with forcefully invalidating thumbs on chart/dashboard change, but recaching it? I'd propose adding a config flag `THUMBNAIL_ON_CHANGE: "invalidate" | "recalculate" | None` for admins to be able to specify if they want to just ignore changes, invalidate or recache every time the object changes.

superset/utils/screenshots.py
```diff
@@ -17,12 +17,14 @@
 from __future__ import annotations
 
 import logging
+from datetime import datetime
+from enum import Enum
 from io import BytesIO
-from typing import TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING
```
villebro, line 23: Nit: we should not use `Optional` anymore, always just `from __future__ import annotations` and then `| None`.

villebro (commented):
First pass comments

superset/config.py
kgabryje: lets go with a week and we can tweak it later if it's too short

superset/models/slice.py
kgabryje: discussed on slack - out of scope

superset/utils/screenshots.py
```diff
@@ -54,6 +56,57 @@
...
+    PENDING = "Pending"
+    COMPUTING = "Computing"
+    UPDATED = "Updated"
+    ERROR = "Error"
+
+
+class ScreenshotCachePayload:
+    def __init__(self, image: bytes | None = None):
+        self._image = image
+        self._timestamp = datetime.now().strftime("%Y/%m/%d-%H:%M:%S")
```
villebro, line 69: Hmm, I'd prefer not to use a custom format here. Let's just use `isoformat()`
```suggestion
        self._timestamp = datetime.now().isoformat()
```

superset/utils/screenshots.py
```diff
@@ -54,6 +56,57 @@
...
+        self.status = StatusValues.PENDING
+        if image:
+            self.status = StatusValues.UPDATED
+
+    def update_timestamp(self) -> None:
+        self._timestamp = datetime.now().strftime("%Y/%m/%d-%H:%M:%S")
```
villebro, line 75: same
```suggestion
        self._timestamp = datetime.now().isoformat()
```

superset/charts/api.py
```diff
@@ -591,25 +603,45 @@ def cache_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse:
...
+                chart_url=chart_url,
+                image_url=image_url,
+                task_updated_at=cache_payload.get_timestamp(),
+                task_status=cache_payload.get_status(),
+            )
+
+        error_cache_ttl = config["THUMBNAIL_ERROR_CACHE_TTL"]
+        error_cache_expired = (
+            datetime.now()
+            - datetime.strptime(cache_payload.get_timestamp(), "%Y/%m/%d-%H:%M:%S")
```
villebro, line 627: same
```suggestion
            - datetime.fromisoformat(cache_payload.get_timestamp())
```

superset/charts/api.py
```diff
@@ -564,8 +569,14 @@ def cache_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse:
...
           responses:
+            200:
+                description: Chart async result
+                content:
+                  application/json:
+                    schema:
+                      $ref: "#/components/schemas/ChartCacheScreenshotResponseSchema"
             202:
-              description: Chart async result
+              description: Chart async task created
```
villebro, line 579: Sorry about the back and forth here, but this seems more appropriate
```suggestion
              description: Chart screenshot task created
```

superset/charts/api.py
```diff
@@ -712,44 +745,41 @@ def thumbnail(self, pk: int, digest: str, **kwargs: Any) -> WerkzeugResponse:
...
+        url = get_url_path("Superset.slice", slice_id=chart.id)
+        screenshot_obj = ChartScreenshot(url, chart.digest)
+        cache_key = screenshot_obj.get_cache_key()
+        cache_payload = (
+            screenshot_obj.get_from_cache_key(cache_key) or ScreenshotCachePayload()
         )
-        # If not screenshot then send request to compute thumb to celery
-        if not screenshot:
+
+        if cache_payload.status in [StatusValues.PENDING, StatusValues.ERROR]:
```
villebro, line 762: Don't we want to check the error TTL here?

superset/dashboards/api.py
```diff
@@ -1197,13 +1094,29 @@ def cache_dashboard_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse
...
+                status_code,
+                cache_key=cache_key,
+                dashboard_url=dashboard_url,
+                image_url=image_url,
+                task_updated_at=cache_payload.get_timestamp(),
+                task_status=cache_payload.get_status(),
+            )
 
-        def trigger_celery() -> WerkzeugResponse:
+        if cache_payload.status in [StatusValues.PENDING, StatusValues.ERROR] or force:
```
villebro, line 1117: same here about checking error TTL

superset/dashboards/api.py
```diff
@@ -1296,13 +1205,120 @@ def screenshot(self, pk: int, digest: str) -> WerkzeugResponse:
...
+        screenshot_obj = DashboardScreenshot(dashboard_url, dashboard.digest)
+        cache_key = screenshot_obj.get_cache_key()
+        cache_payload = (
+            screenshot_obj.get_from_cache_key(cache_key) or ScreenshotCachePayload()
+        )
+        image_url = get_url_path(
+            "DashboardRestApi.thumbnail", pk=dashboard.id, digest=cache_key
+        )
+
+        if cache_payload.status in [StatusValues.PENDING, StatusValues.ERROR]:
```
villebro, line 1294: here, too

villebro (commented):
Second pass comments. I'd propose breaking the status check into a method of `CachePayload` to avoid duplication of the logic. Something like `.should_trigger_task()` that returns `bool` based on status, error TTL etc.

superset/utils/screenshots.py
```diff
@@ -54,6 +56,68 @@
...
+
+    def get_status(self) -> str:
+        return self.status.value
+
+    def should_trigger_task(self) -> bool:
+        def is_error_cache_ttl_expired() -> bool:
+            error_cache_ttl = app.config["THUMBNAIL_ERROR_CACHE_TTL"]
+            return (
+                datetime.now() - datetime.fromisoformat(self.get_timestamp())
+            ).total_seconds() > error_cache_ttl
```
villebro, line 114: Is there some reason we're nesting this? Couldn't it just be a regular method?

superset/charts/api.py
```diff
@@ -591,25 +603,36 @@ def cache_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse:
...
+            return self.response(
+                status_code,
+                cache_key=cache_key,
+                chart_url=chart_url,
+                image_url=image_url,
+                task_updated_at=cache_payload.get_timestamp(),
+                task_status=cache_payload.get_status(),
+            )
+
+        if cache_payload.should_trigger_task() or force:
```
villebro, line 624: very very nit, but flipping this around will likely save some CPU. To DRY it even more, we could maybe even consider making `force` an optional parameter in `should_trigger_task` that defaults to `False` (not highly opinionated here)
```suggestion
        if force or cache_payload.should_trigger_task():
```

superset/dashboards/api.py
```diff
@@ -1197,13 +1093,29 @@ def cache_dashboard_screenshot(self, pk: int, **kwargs: Any) -> WerkzeugResponse
...
+                status_code,
+                cache_key=cache_key,
+                dashboard_url=dashboard_url,
+                image_url=image_url,
+                task_updated_at=cache_payload.get_timestamp(),
+                task_status=cache_payload.get_status(),
+            )
 
-        def trigger_celery() -> WerkzeugResponse:
+        if cache_payload.should_trigger_task() or force:
```
villebro, line 1116: same
```suggestion
        if force or cache_payload.should_trigger_task():
```

villebro (approved):
A few last non-blocking comments. Great to see this land before 5.0 cut, this will save considerable compute where enabled!

superset/utils/screenshots.py
kgabryje: I figured that we don't use it anywhere else and we only need to check it inside of `should_trigger_task` method. I can make it a regular method or a private one instead of nested if you think it's better

villebro: I'd prefer a separate method, as redefining it each time will incur some extra overhead.

ganeshvaidee:
Is there a workaround for this in 4.1.10? We recently upgraded to 4.0 and then 4.1.0 and are seeing our DB queries increase because it looks like refresh from cache is not working as expected. We have no plans to upgrade to 5.0 yet.

PR #31758, opened 2025-01-08
chore: move label-draft-pr from action to workflow

sadpandajoe:
### SUMMARY

Looks like my initial change wasn't triggering the label. Trying to move this to github workflow vs action to see if it'll trigger it.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -13,141 +13,147 @@
 ############################################
 
 "risk:db-migration":
-- changed-files:
-  - any-glob-to-any-file:
-    - 'superset/migrations/**'
+  - changed-files:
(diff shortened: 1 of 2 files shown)


sadpandajoe:
@mistercrunch as the resident github actions expert, let me know if this isn't the right place to move the file as it wasn't being triggered earlier

mistercrunch (approved):
LGTM - You would have gotten extra points by using `git mv` insetad of `mv` and keeping the history attached, but I don't think that's important here.

PR #31759, opened 2025-01-08
feat: WiP support for python 3.12

mistercrunch:
Gave supporting `3.12` a shot while deprecating `3.9`, made some progress but hit some blockers. Keeping this PR open to keep tabs on what's blocking at this time.

`pandas==2.0.3` does not support py3.12, so bumped to latest, hit some issues, found some worrisome changes apparently in `pandas>=2.2.0` related to them no supporting `dtype` arg of type `sqlalchemy` in `df.to_sql()`, so reverter to the highest under 2.2, and seems there are issues there. Pandas folks not respecting semver it seems but given the API surface pandas have, gonna be have to evolve that lib without breaking a million things.

Python wizards, help is very welcome to help get this through

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -28,8 +28,7 @@ runs:
         if [ "${{ inputs.python-version }}" = "current" ]; then
           echo "PYTHON_VERSION=3.11.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
-          # TODO: set to 3.12 when libs are ready - seemed hard on 12/17/24
-          echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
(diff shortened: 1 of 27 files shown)


rad-pat:
In our app, to upgrade Pandas to 2.2.3 and still work with SQLAlchemy 1.4.x, we had to change the way we called e.g. `pd.read_sql_query` 

So, the likes of this call: 
https://github.com/apache/superset/blob/aae8fda11dafdd96298884962e790d10bd0104d0/superset/models/helpers.py#L1381

Would change to read:
```
df = pd.read_sql_query(sql=sql, con=engine.raw_connection())
```

Worth a try in the places it fails 👍

rusackas:
This one hasn't been updated since January, should we close it out?

mistercrunch:
i think this isn't relevant anymore, looks like we're running against `next` as 3.12 already

PR #31582, opened 2024-12-20
refactor: Removes 5.0 approved legacy charts

michael-s-molina:
### SUMMARY
As part of the 5.0 breaking window, this PR executes the following proposals:
- [23 - Force legacy Area chart migration and delete NVD3 plugin](https://github.com/orgs/apache/projects/345/views/1#)
- [42 - Migrate Line Chart (legacy) to ECharts Line Chart and remove code (was previously deprecated)](https://github.com/orgs/apache/projects/345/views/1#)
- [54 - Migrate and remove the legacy Heatmap chart](https://github.com/orgs/apache/projects/345/views/1#)
- [80 - Migrate and remove the NVD3 Bar Chart in favor of ECharts](https://github.com/orgs/apache/projects/345/views/1#)
- [84 - Migrate and remove the legacy Sankey chart](https://github.com/orgs/apache/projects/345/views/1#)
- [85 - Remove the unused Sankey Loop chart](https://github.com/orgs/apache/projects/345/views/1#)
- [86 - Migrate and remove the legacy Histogram chart](https://github.com/orgs/apache/projects/345/views/1#)
- [88 - Remove the Event Flow chart](https://github.com/orgs/apache/projects/345/views/1#)

I decided to do it all in a single PR to have a single migration for the charts. There's also many code blocks that were removed that touches multiple visualization types at the same time and it would be more complex to remove it in parts. Another important factor was the ability to execute the tests once, given that we need to check if the example charts are loaded correctly. In summary, the benefits outweigh the review complexity.

### TESTING INSTRUCTIONS
- Check the examples are loaded correctly
- Check that the charts in your production environment are all migrated and loaded correctly

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

### What change is being made?

Remove legacy charts (Area, Bar, Event Flow, Heatmap, Histogram, Line, Sankey, and Sankey Loop) and migrate them to their ECharts counterparts, while updating relevant tests and dependencies.

### Why are these changes being made?

[...]

PR #31791, opened 2025-01-10
chore: Skips integration tests affected by legacy charts removal

michael-s-molina:
### SUMMARY
This PR skips all tests affected by https://github.com/apache/superset/pull/31582 to unblock master and avoid reverting an important PR. I'll open a follow-up to fix these tests to use non-legacy assets.

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/async_events/api_tests.py
+++ b/tests/integration_tests/async_events/api_tests.py
@@ -17,6 +17,7 @@
 from typing import Any, Optional, Type
 from unittest import mock
 
+import pytest
 import redis
 
 from superset.async_events.cache_backend import (
@@ -30,6 +31,7 @@ from tests.integration_tests.constants import ADMIN_USERNAME
 from tests.integration_tests.test_app import app
 
 
+@pytest.skip(reason="Needs to investigate this test", allow_module_level=True)
 class TestAsyncEventApi(SupersetTestCase):
     UUID = "943c920-32a5-412a-977d-b8e47d36f5a4"
 
--- a/tests/integration_tests/cache_tests.py
+++ b/tests/integration_tests/cache_tests.py
@@ -29,6 +29,10 @@ from tests.integration_tests.fixtures.birth_names_dashboard import (
     load_birth_names_data,  # noqa: F401
 )
 
+pytest.skip(
+    reason="These tests will be changed to use the api/v1/data", allow_module_level=True
+)
+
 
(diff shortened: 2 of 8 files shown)


tests/integration_tests/async_events/api_tests.py
```diff
@@ -30,6 +31,7 @@
 from tests.integration_tests.test_app import app
 
 
+@pytest.skip(reason="Needs to investigate this test", allow_module_level=True)
```
michael-s-molina, line 34: This one does not seem related to the legacy charts removal. Need to investigate in more detail.

Vitor-Avila (approved):
LGTM!

tests/integration_tests/async_events/api_tests.py
michael-s-molina: No. I will investigate when re-enabling the test in a follow-up.

Vitor-Avila: gotcha. thank you!

michael-s-molina: @Vitor-Avila I tried to restore this test but it's throwing an un-related error. If you want to take a stab at it.. 🤷🏼‍♂️

Vitor-Avila: @michael-s-molina I tested locally and it seems the tests work if `GLOBAL_ASYNC_QUERIES` FF is set to `True`, which I think makes sense? I'll double check that and if so push a PR.

Vitor-Avila: @michael-s-molina pushed the change here: https://github.com/apache/superset/pull/32062 it worked locally -- hopefully it works through CI as well 🙏

Vitor-Avila: ooof, failed in the CI 😢

michael-s-molina: Yeah, I faced the same error trying to make it work 😢

Vitor-Avila: ok @michael-s-molina got it to work 🙌 could you please review? https://github.com/apache/superset/pull/32062

PR #31794, opened 2025-01-10
chore: Removing DASHBOARD_CROSS_FILTERS flag and all that comes with it.

rusackas:
### SUMMARY

Approved 5.0 change. The feature is now ready for prime time!

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/components/DashboardBuilder/DashboardBuilder.tsx
```diff
@@ -475,8 +472,7 @@ const DashboardBuilder = () => {
     ELEMENT_ON_SCREEN_OPTIONS,
   );
 
-  const showFilterBar =
-    (crossFiltersEnabled || nativeFiltersEnabled) && !editMode;
```
sadpandajoe, line 479: was `nativeFiltersEnabled` already cleaned up?

rusackas: Not sure, actually, but it's part of an OR statement, so with crossFiltersEnabled always being true, it doesn't actually matter, and will just be the same as `!editMode`

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterBarSettings/FilterBarSettings.test.tsx
```diff
@@ -104,44 +104,14 @@ test('Dropdown trigger does not render without dashboard edit permissions', asyn
   expect(screen.queryByRole('img', { name: 'gear' })).not.toBeInTheDocument();
 });
 
-test('Dropdown trigger renders with FF DASHBOARD_CROSS_FILTERS on', async () => {
```
sadpandajoe, line 107: Did you want to update this test as just "Dropdown trigger renders"?

rusackas: The dropdown trigger is clicked in other tests so I wasn't worried about keeping it as a separate one. Those other tests would fail if it didn't render.

sadpandajoe (approved):
LGTM

UPDATING.md
```diff
@@ -23,6 +23,7 @@ This file documents any backwards-incompatible changes in Superset and
 assists people when migrating to a new version.
 
 ## Next
+- [31794](https://github.com/apache/superset/pull/31794) Removed the previously deprecated `DASHBOARD_CROSS_FILTERS` feature flag
```
michael-s-molina, line 26:
```suggestion
- [31794](https://github.com/apache/superset/pull/31794) Removed the previously deprecated `DASHBOARD_CROSS_FILTERS` feature flag.
```

superset-frontend/src/dashboard/components/nativeFilters/utils.test.ts
```diff
@@ -45,6 +45,12 @@ describe('nativeFilterGate', () => {
     });
 
     it('should return false for native filter chart with cross filter support', () => {
+      isFeatureEnabledMock.mockImplementation(feature => {
+        if (feature === FeatureFlag.DASHBOARD_CROSS_FILTERS) return false;
```
michael-s-molina, line 49: I believe `FeatureFlag.DASHBOARD_CROSS_FILTERS` will not exist anymore given that we are removing the feature flag. We need to remove all references of it.

superset-frontend/src/components/Chart/ChartRenderer.jsx
```diff
@@ -90,10 +88,7 @@ class ChartRenderer extends Component {
     )?.suppressContextMenu;
     this.state = {
       showContextMenu:
-        props.source === ChartSource.Dashboard &&
-        !suppressContextMenu &&
-        (isFeatureEnabled(FeatureFlag.DrillToDetail) ||
```
kgabryje, line 95: I think we should keep the check for drill to detail here

superset-frontend/src/dashboard/actions/dashboardState.js
```diff
@@ -437,10 +435,8 @@ export function saveDashboardRequest(data, id, saveType) {
     ) {
       let chartConfiguration = {};
       let globalChartConfiguration = {};
-      if (isFeatureEnabled(FeatureFlag.DashboardCrossFilters)) {
-        ({ chartConfiguration, globalChartConfiguration } =
-          handleChartConfiguration());
-      }
+      ({ chartConfiguration, globalChartConfiguration } =
```
kgabryje, line 438: We can remove the `let` declarations above and use const here

superset-frontend/src/dashboard/components/nativeFilters/selectors.ts
```diff
@@ -379,16 +375,14 @@ export const selectNativeIndicatorsForChart = (
...
-  if (isFeatureEnabled(FeatureFlag.DashboardCrossFilters)) {
-    crossFilterIndicators = selectChartCrossFilters(
-      dataMask,
-      chartId,
-      chartLayoutItems,
-      chartConfiguration,
-      appliedColumns,
-      rejectedColumns,
-    );
-  }
```
kgabryje, line 391: Same as above, let’s just use const here 🙂

superset-frontend/src/components/Chart/ChartRenderer.jsx
michael-s-molina: +1

kgabryje:
@rusackas I think the tests should pass now

superset-frontend/src/dashboard/components/nativeFilters/utils.test.ts
```diff
@@ -16,75 +16,28 @@
...
-import * as uiCore from '@superset-ui/core';
+import { Behavior } from '@superset-ui/core';
 import { DashboardLayout } from 'src/dashboard/types';
 import { CHART_TYPE } from 'src/dashboard/util/componentTypes';
 import { nativeFilterGate, findTabsWithChartsInScope } from './utils';
 
-let isFeatureEnabledMock: jest.MockInstance<boolean, [feature: FeatureFlag]>;
-
 describe('nativeFilterGate', () => {
-  describe('with all feature flags disabled', () => {
```
kgabryje, line 28: I removed those tests as they didn't make sense now since we got rid of the featuer flags

PR #31796, opened 2025-01-10
feat: get docker-compose to work as the backend for Cypress tests

mistercrunch:
Screenshot 2025-01-10 at 7 15 20 PM

Running Cypress tests as depicted in docs is difficult and error-prone. Some of the challenges include:
- having a reproducible deterministic setup (version of python that matches CI)
- switching to a different `superset_config`
- requires a local postgres
- database state from previous branch can interfere in and out of a cypress setup (different set of examples, database migrations might be in a different checkpoint, ...)
- switching to a different port
- having celery workers working
- caching is nice to have to accelerate tests, which may or may not be set up
- ...

Given this, I thought I'd just make it such that you can just `CYPRESS_CONFIG=true docker compose up` to fire up a backend against which you can run cypress tests locally. Given the reference to `CYPRESS_CONFIG` in our docker-compose.yml and in various `docker/`-related scripts, it looks like either it was supported in the past, or someone tried to do work in this direction, but clearly isn't working today as is.

To get this working here I'm:
- setting up `CYPRESS_CONFIG=true` to point to a dedicated database schema `superset_cypress`, this gets created on Postgres first-boot, and currently will require deleting the docker volume for Postgres to take effect
- starting up superset app to run on `8081`, which seems to be the default port assumption for Cypress runs
- pointing to the config file at `tests/integration_tests/superset_test_config.py`, which is need for tests, eventually might want to merge this into `docker/pythonpath/superset_config.py` for simplicity. For now I conditionally import it
- removing duplicate docs entry existing in both `docs/docs/contributing/development.mdx` and `docs/docs/contributing/howtos.mdx`
- adjusting docs

## addressing current issues 

I won't lie getting this back on track feels like wacking a machete through the jungle. Borderline disrepair territory. Turns out probably no one has been running tests locally for a while cause the setup was broken af.


docker-compose.yml
```diff
@@ -89,6 +89,8 @@ services:
     restart: unless-stopped
     ports:
       - 8088:8088
+      # When in cypress-mode ->
+      - 8081:8081
```
mistercrunch, line 93: NOTE: debated whether to try and open this port conditionally, but seemed impossible and researched whether there's downside in opening a port-to-nowhere in other scenarios and I seemed there's no downside. Could also just run Cypress against 8088 too if we prefer

docker/pythonpath_dev/superset_config.py
```diff
@@ -107,6 +108,18 @@ class CeleryConfig:
 log_level_text = os.getenv("SUPERSET_LOG_LEVEL", "INFO")
 LOG_LEVEL = getattr(logging, log_level_text.upper(), logging.INFO)
 
+if os.getenv("CYPRESS_CONFIG") == "true":
+    # When running the service as a cypress backend, we need to import the config
+    # located @ tests/integration_tests/superset_test_config.py
+    base_dir = os.path.dirname(__file__)
+    module_folder = os.path.abspath(
+        os.path.join(base_dir, "../../tests/integration_tests/")
```
mistercrunch, line 116: NOTE: conditionally loading config located at `tests/integration_tests/`, but would much rather deprecate that file and merge non-conflicting/duplicated configs directly in here. Doing this would need adjustments on CI scripts and could break local envs (probably ok as we would instruct devs to work in `docker compose`)

docs/docs/contributing/development.mdx
```diff
@@ -631,71 +631,6 @@ To run a single test file:
 npm run test -- path/to/file.js
 ```
 
-### Integration Testing
```
mistercrunch, line 634: NOTE: these docs, and many other entries in these files are duplicated in `./howtos.mdx` . It's a bit tricky as they are not pure dups as there has been edits on both sides

superset/config.py
```diff
@@ -1923,7 +1923,7 @@ class ExtraDynamicQueryFilters(TypedDict, total=False):
             "Failed to import config for %s=%s", CONFIG_PATH_ENV_VAR, cfg_path
         )
         raise
-elif importlib.util.find_spec("superset_config") and not is_test():
```
mistercrunch, line 1926: NOTE: this seemed to say "if you're in test mode, DONT load ANY custom config", which is a confusing hack and probably doesn't belong here or anywhere

superset-frontend/cypress-base/cypress/support/e2e.ts
```diff
@@ -26,6 +26,10 @@ require('cy-verify-downloads').addCustomCommand();
 
 // fail on console error, allow config to override individual tests
 // these exceptions are a little pile of tech debt
+//
+
+// DISABLING FOR NOW
```
mistercrunch, line 31: Ok so now I know no one has used cypress locally in a while since any console error/warning including React ones fails the tests. From my understanding the reason why it works in CI is because the `npm run build` suppresses these console errors.

I'm thinking we disable for now and re-enable later after cleaning up all React warnings, which is beyond the scope of this PR

superset-frontend/cypress-base/cypress.config.ts
```diff
@@ -27,7 +27,8 @@ export default eyesPlugin(
     chromeWebSecurity: false,
     defaultCommandTimeout: 8000,
     numTestsKeptInMemory: 0,
-    experimentalFetchPolyfill: true,
+    // Disabled after realizing this MESSES UP rison encoding in intricate ways
+    experimentalFetchPolyfill: false,
```
mistercrunch, line 31: This bug was gnarly to debug. Screw this experimental feature.

mistercrunch:
@sadpandajoe , could use help here around:
- understanding how I broke the few tests that are failing here given the content of this PR. I'm not really changing the code/logic in any way, and most tests pass now. What did i change in this PR that could have broken those few tests? or is it just semi-random/flakiness (?)
- help fixing the tests, I'm a bit at a loss with Cypress (though I did manage to fix some intricate issues related to timing/flaky tests. It was around an async call that would trigger a react re-render and some race-condition locally, maybe cause my laptop is faster than CI?)

superset-frontend/cypress-base/cypress.config.ts
mistercrunch: Now wondering if switching this off could be what's failing the few tests (?)

superset-frontend/cypress-base/cypress.config.ts
```diff
@@ -70,6 +72,7 @@ export default eyesPlugin(
       },
       baseUrl: 'http://localhost:8088',
       excludeSpecPattern: [],
+      experimentalRunAllSpecs: true,
```
mistercrunch, line 75: NOTE: this adds a button to "run all tests" in-context in the Cypress UI

superset-frontend/cypress-base/cypress/support/e2e.ts
```diff
@@ -160,7 +166,18 @@ Cypress.Commands.add('login', () => {
     url: '/login/',
     body: { username: 'admin', password: 'general' },
   }).then(response => {
-    expect(response.status).to.eq(200);
+    if (response.status === 302) {
```
mistercrunch, line 169: NOTE: not sure if this is required, but at some point was getting some redirect around login, this handles redirect as not a failure

sadpandajoe:
> @sadpandajoe , could use help here around:
> 
> * understanding how I broke the few tests that are failing here given the content of this PR. I'm not really changing the code/logic in any way, and most tests pass now. What did i change in this PR that could have broken those few tests? or is it just semi-random/flakiness (?)
> * help fixing the tests, I'm a bit at a loss with Cypress (though I did manage to fix some intricate issues related to timing/flaky tests. It was around an async call that would trigger a react re-render and some race-condition locally, maybe cause my laptop is faster than CI?)

@mistercrunch do you know if we stored the failed screenshots anywhere? If so it should have a snapshot of the failure state. It looks like the two errors are grouped as:
- antd5 tooltip not being found which could either be the page didn't load correctly or that the element isn't antd5 yet. So having the screenshot could eliminate one of them
- delete modal button not found. same thing here as it could be that the selector has changed or the modal never appears.

mistercrunch:
@sadpandajoe maybe we can pair up and dig in, I have the whole interactive Cypress thing running and interactively debugging but was stuck. Also confused as to how my backend PR creates tooltip/modal issues. Seem like timing / race condition type stuff. Tried disabling animation, longer timeouts, ...

superset-frontend/cypress-base/cypress.config.ts
```diff
@@ -26,8 +26,9 @@ export default eyesPlugin(
   defineConfig({
     chromeWebSecurity: false,
     defaultCommandTimeout: 8000,
-    numTestsKeptInMemory: 0,
-    experimentalFetchPolyfill: true,
+    numTestsKeptInMemory: 1,
```
sadpandajoe, line 29: you want this to be 0 for ci, so i'd see if this could be overridden as part of the command to start up cypress

docker-compose.yml
eschutho: Is it fairly simple to run Cypress against 8088? I know that we can change the port when running locally.

docs/docs/contributing/howtos.mdx
```diff
@@ -225,22 +225,10 @@ npm run test -- path/to/file.js
...
 ```bash
-export SUPERSET_CONFIG=tests.integration_tests.superset_test_config
-export SUPERSET_TESTENV=true
-export CYPRESS_BASE_URL="http://localhost:8081"
-superset db upgrade
-superset load_test_users
-superset init
-superset load-examples --load-test-data
-superset run --port 8081
+CYPRESS_CONFIG=true docker compose up
```
sadpandajoe, line 231: Will this also load all the assets? I had an issue so working from a clean state and it looks like charts/dashboards, didn't load for me at all when spinning this up with this command.

superset-frontend/cypress-base/cypress.config.ts
mistercrunch: ooops that leaked through... how about we force to `--config numTestsKeptInMemory=0` in CI, but remove the setting to use whatever the default is locally?

sadpandajoe: that's fine as long as it's not on in CI.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -78,20 +81,15 @@ describe('Charts list', () => {
       cy.wait('@get');
     });
 
-    it('should show the newly added dashboards in a tooltip', () => {
+    it.only('should show the newly added dashboards in a tooltip', () => {
```
sadpandajoe, line 84: need to remove this only or else the other tests won't run

superset-frontend/cypress-base/cypress.config.ts
mistercrunch: I'm moving now to remove local settings from being hard coded (commenting out the setting), but by setting it to `0` in CI using command line

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
sadpandajoe: @mistercrunch actually on second thought, we should probably add a it.skip to skip the test entirely with the old code. We should leave a comment above the test to say we skipped it because mouseover does not trigger the tooltip in cypress but does do it manually. Currently the code that is removed is the meat of showing the tooltip, so if we aren't going to validate this, we probably don't need the tests to run all the steps before the validation.

docker-compose.yml
mistercrunch: yup it's super easy, I'm just unclear if we want segmentation on PORT similar to the segmentation we have at the database level. In theory would could have both a dev and cypress environment running in parallel and not conflicting on ports ... Though I don't think you can fire two docker-compose at the same time (?)

docker/pythonpath_dev/superset_config.py
```diff
@@ -107,6 +108,18 @@ class CeleryConfig:
...
+    )
+    sys.path.insert(0, module_folder)
+    from superset_test_config import *  # noqa
```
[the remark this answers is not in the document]
mistercrunch: which part of `# noqa` don't you understand @korbit-ai !?

docs/docs/contributing/howtos.mdx
mistercrunch: yes it will, that's somewhere in the docker init file

docker-compose.yml
mistercrunch: One thing that's kind of good with using a different port, is that firing up Cypress against `docker compose up` will fail as it's looking for Superset on 8081, forcing the user to stop docker-compose and to fire it up properly `CYPRES_CONFIG=true docker-compose up`, meaning it enforces having the right backend, with the right test sets and examples loaded up, preventing some confusion around using the wrong backend.

eschutho:
@mistercrunch I tested the branch with running:
`docker compose up db` (opened a shell and logged into postgres and created the superset_cypress db)
` CYPRESS_CONFIG=true docker-compose up`
tests on localhost:8081-- test the login with `admin/general`
`cd superset-frontend/cypress-base`
`npm ci`
`npx cypress open --config numTestsKeptInMemory=3`
And I'm seeing this error (it's looking for port 8081)
```cy.request() failed trying to load:

http://localhost:8088/coverage

We attempted to make an http request to this URL but the request failed without a response.

We received this error at the network level:

> Error: socket hang up

-----------------------------------------------------------

The request we sent was:

Method: GET
URL: http://localhost:8088/coverage```

**Comment by mistercrunch** [comment]
@eschutho you probably have an env var in our current shell pointing to 8088. 
```
$ env | grep -i cypress
$ unset CYPRESS_BASE_URL
```

eschutho:
Ok, I'm running it in a different terminal window and no error, so that was probably it. LGTM!

eschutho (approved):

PR #31798, opened 2025-01-11
fix(Embedded): Skip CSRF validation for dashboard download endpoints

Vitor-Avila:
### SUMMARY
https://github.com/apache/superset/pull/29953 removed the CSRF requirement for endpoints used to download a dashboard as PNG/PDF, so that these work in embedded mode.

This PR changes the FF validation logic for these endpoints from using `before_request` to a new decorator, as `before_request` was causing issues to the `WTF_CSRF_EXEMPT_LIST` config.

No UI changes.

### TESTING INSTRUCTIONS
1. Have async dashboard screenshot download configured.
2. Access an embedded dashboard and validate that the PDF download works.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -19,20 +19,19 @@ import functools
 import logging
 from datetime import datetime
 from io import BytesIO
-from typing import Any, Callable, cast, Optional
+from typing import Any, Callable, cast
 from zipfile import is_zipfile, ZipFile
 
 from flask import g, redirect, request, Response, send_file, url_for
 from flask_appbuilder import permission_name
 from flask_appbuilder.api import expose, protect, rison, safe
-from flask_appbuilder.hooks import before_request
 from flask_appbuilder.models.sqla.interface import SQLAInterface
 from flask_babel import gettext, ngettext
 from marshmallow import ValidationError
 from werkzeug.wrappers import Response as WerkzeugResponse
 from werkzeug.wsgi import FileWrapper
 
-from superset import db, is_feature_enabled, thumbnail_cache
+from superset import db, thumbnail_cache
 from superset.charts.schemas import ChartEntityResponseSchema
 from superset.commands.dashboard.copy import CopyDashboardCommand
 from superset.commands.dashboard.create import CreateDashboardCommand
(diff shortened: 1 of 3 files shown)


superset/dashboards/api.py
```diff
@@ -1141,6 +1130,7 @@ def thumbnail(self, pk: int, digest: str, **kwargs: Any) -> WerkzeugResponse:
         )
 
     @expose("/<pk>/cache_dashboard_screenshot/", methods=("POST",))
+    @validate_feature_flags(["THUMBNAILS", "ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS"])
```
geido, line 1133: I am thinking if we really need the couple here at this point or `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS` will suffice. What's your thought? I have been meaning to remove this dependency before and this could be a good chance.

geido (approved):

superset/dashboards/api.py
Vitor-Avila: @geido if we remove the requirement to have `THUMBNAILS` enabled for this endpoint, but then we keep it at the `digest` endpoint, then at the end the process would still not work unless you have both enabled. 

Also, currently when you trigger a "screenshot generation" it returns the digest link, which returns a `404` status code until it's being processed. If `THUMBNAILS` is disabled, it would return `404` forever and users wouldn't have an easy way to spot the issue.

I think we could either:
* Remove the `THUMBNAILS` check for this endpoint, and then update the check on the `digest` endpoint to validate for either `THUMBNAILS` OR `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS`. This way you can still download a dashboard screenshot without having `THUMBNAILS` FF enabled; or
* Update the other endpoint to return a different status code if the `THUMBNAILS` FF is disabled. Also update the frontend (if that's not the case yet) to validate if both FFs are enabled to decide if frontend or backend processing should be used.

One potential issue with the first approach is that environments with `ENABLE_DASHBOARD_SCREENSHOT_ENDPOINTS` enabled but `THUMBNAILS` disabled would still expose the `digest` endpoint, so it's possible that disabling `THUMBNAILS` FF would no longer be sufficient to remove thumbnails from displaying in the card views.

Let me know your thoughts!

geido: Alright. I think there is a case here to keep both. Maybe @fisjac can consider the above for the work that he is doing around screenshot generation for future improvements. Thanks for this!

PR #31844, opened 2025-01-14
chore: replace selenium user with fixed user

### LINKED ISSUE #31843: 76 - Remove falling back to fixed selenium executor for thumbnails by default
Currently we have
```python
THUMBNAIL_SELENIUM_USER: str | None = "admin"
THUMBNAIL_EXECUTE_AS = [ExecutorType.CURRENT_USER, ExecutorType.SELENIUM]
```
I propose we change the name of the config + remove defaulting to it
```python
THUMBNAIL_FIXED_USER: str | None = None
THUMBNAIL_EXECUTE_AS = [ExecutorType.CURRENT_USER]
```

Same for Alerts & Reports

villebro:
### SUMMARY
This PR does the following:
- Remove the `THUMBNAILS_SELENIUM_USER` config parameter. To execute as a fixed user you can now use the new `FixedUser` executor that can assume the role of any user. This means you can use a different fixed user for thumbnails and Alerts & Reports if necessary.
- Rename `ALERT_REPORTS_EXECUTE_AS` and `THUMBNAILS_EXECUTE_AS` to `ALERT_REPORTS_EXECUTORS` and `THUMBNAILS_EXECUTORS` for improved clarity.
- Add `CACHE_WARMUP_EXECUTORS` to make it possible to specify executors for cache warmup tasks. With this change all async tasks can be configured to use arbitrary executors, similar to what was already possible for Thumbnails and Alerts & Reports.
- Change the cache warmup strategy interface to have a `get_tasks` method instead of `get_payloads`. This change is needed as the cache warmup tasks now need to provide a username to execute the cache warmup tasks.
- Update docs to reflect the changes.

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes #31843
- [ ] Required feature flags:
- [ ] Changes UI


UPDATING.md
```diff
@@ -24,6 +24,7 @@ assists people when migrating to a new version.
 
 ## Next
 
+- [31844](https://github.com/apache/superset/pull/31844) The `ALERT_REPORTS_EXECUTE_AS` and `THUMBNAILS_EXECUTE_AS` config parameters have been renamed to `ALERT_REPORTS_EXECUTORS` and `THUMBNAILS_EXECUTORS` respectively. A new config flag `CACHE_WARMUP_EXECUTORS` has also been introduced to be able to control which user is used to execute cache warmup tasks. Finally addition, the config flag `THUMBNAILS_SELENIUM_USER` has been removed. To use a fixed executor for async tasks, use the new `FixedExecutor` class. See the config
```
michael-s-molina, line 27: > See the config

Did you forget to complete the sentence?

superset/config.py
```diff
@@ -689,17 +689,29 @@ class D3TimeFormat(TypedDict, total=False):
...
 EXTRA_SEQUENTIAL_COLOR_SCHEMES: list[dict[str, Any]] = []
 
+# User used to execute cache warmup tasks
+# By default, the cache is warmed up using the primary owner. To fall back to using
+# a fixed user (admin in this example), use the following configuration:
+#
+# from superset.tasks.types import ExecutorType, FixedExecutor
+#
+# CACHE_WARMUP_EXECUTORS = [ExecutorType.OWNER, FixedExecutor("admin")]
+CACHE_WARMUP_EXECUTORS = [ExecutorType.OWNER]
```
michael-s-molina, line 699: Would be possible to have a more generic configuration that allows future async tasks and reduces the number of configurations? Something like:

```
ASYNC_TASK_EXECUTORS: dict[AsyncTaskType, ExecutorType] = {}
```

villebro: @michael-s-molina I kinda like this idea. However, I'm not fully clear on what the exact pattern will look like for the forthcoming `AsyncTaskType`. Similarly, a while back when I was implementing a custom `KeyValue` resource for an internal use case, I ran into issues where it was difficult to extend the existing `KeyValueResource` type. In the end I ended up doing a bunch of funky casting to avoid linting issues. So to get around this, we would need to make sure the type (or its equivalent) supports introducing new types beyond the built-in ones.

Thoughts? I'm ok implementing an `AsyncTaskType` enum (or similar) now, but we may still need to refactor it later, causing yet another breaking change.

michael-s-molina: How about:

```
class AsyncTaskType:
    """Base class for async task types."""
    pass

class CacheWarmUpTask(AsyncTaskType):
    pass

class AlertsReportsTask(AsyncTaskType):
    pass

# Usage
ASYNC_TASKS_EXECUTORS: dict[AsyncTaskType, ExecutorType] = {
    CacheWarmUpTask(): ExecutorType.OWNER,
    AlertsReportsTask(): FixedExecutor("admin"),
}
```

villebro: I'd prefer something that serializes more easily. String based enums are good in this regard, but introduce the pluggability issue outlined above.

michael-s-molina: Would a class with predefined values that can be extended be serializable enough? Something like:

```
class AsyncTaskType:
   CACHE_WARMUP = "CACHE_WARMUP",
   ALERTS_REPORTS = "ALERTS_REPORTS",

   _additional_types = set()

   @classmethod
   def add_type(cls, type):
       if type not in cls._additional_types:
           cls._additional_types.add(type)

   @classmethod
   def get_types(cls):
       return {cls.CACHE_WARMUP, cls.ALERTS_REPORTS}.union(cls._additional_types)

```

villebro: This would essentially mean `get_types` just returns `set[str]`, which means `ASYNC_TASKS_EXECUTORS` would be `dict[str, list[ExecutorType]]`. I'm sure there's some good pattern for solving this, but I still feel that's best taken care of in a follow-up PR that implements the new async task framework.

villebro:
@michael-s-molina is there anything else that needs to be addressed, or is this good to go, assuming the async task type change can wait until the GATF feature?

superset/thumbnails/digest.py
```diff
@@ -92,11 +93,15 @@ def _adjust_string_with_rls(
...
-        current_user=get_current_user(),
-    )
+    try:
+        executor_type, executor = get_executor(
+            executors=config["THUMBNAIL_EXECUTORS"],
+            model=dashboard,
+            current_user=get_current_user(),
+        )
+    except ExecutorNotFoundError:
+        return ""
```
michael-s-molina, line 103: Maybe `None` will be a better representation? If you agree, please change all digest generation methods.

michael-s-molina (approved):

superset/charts/schemas.py
```diff
@@ -162,6 +162,7 @@ class ChartEntityResponseSchema(Schema):
     certification_details = fields.String(
         metadata={"description": certification_details_description}
     )
+    thumbnail_url = fields.String(allow_none=True)
```
villebro, line 165: I noticed that `thumbnail_url` was missing on the Marshmallow schema despite being returned by the API

PR #31858, opened 2025-01-15
chore: refactor Alert-related components

mistercrunch:
Chiseling at https://github.com/apache/superset/pull/31590 that has gotten big / unruly, in this PR is a refactor of Alert-related components, going vanilla AntD.

Also here. Deprecating colors.alerts since it's ambiguous/redundant with warning/error and does not exist in antd-v5

Screenshot 2025-01-14 at 11 42 05 PM
Screenshot 2025-01-14 at 11 41 47 PM

### DIFF
--- a/.github/workflows/superset-e2e.yml
+++ b/.github/workflows/superset-e2e.yml
@@ -141,4 +141,4 @@ jobs:
         if: failure()
         with:
           path: ${{ github.workspace }}/superset-frontend/cypress-base/cypress/screenshots
-          name: cypress-artifact-${{ github.run_id }}-${{ github.job }}
+          name: cypress-artifact-${{ github.run_id }}-${{ github.job }}-${{ matrix.browser }}-${{ matrix.parallel_id }}
--- a/superset-frontend/cypress-base/cypress.config.ts
+++ b/superset-frontend/cypress-base/cypress.config.ts
@@ -36,7 +36,7 @@ export default eyesPlugin(
     viewportHeight: 1024,
     projectId: 'ud5x2f',
     retries: {
-      runMode: 3,
+      runMode: 2,
       openMode: 0,
     },
     e2e: {
--- a/superset-frontend/cypress-base/cypress.config.ts
+++ b/superset-frontend/cypress-base/cypress.config.ts
@@ -36,7 +36,7 @@ export default eyesPlugin(
     viewportHeight: 1024,
     projectId: 'ud5x2f',
     retries: {
-      runMode: 2,
+      runMode: 3,
       openMode: 0,
     },
     e2e: {
--- a/superset-frontend/src/components/ErrorMessage/ErrorAlert.stories.tsx
+++ b/superset-frontend/src/components/ErrorMessage/ErrorAlert.stories.tsx
@@ -1,0 +1,151 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { Meta, StoryFn } from '@storybook/react';
+import { Layout, Row, Col, Card } from 'antd-v5';
+import ErrorAlert from './ErrorAlert';
+
+const { Content } = Layout;
+
+const longDescription = `This is a detailed description to test long content display.
+Line breaks are included here to demonstrate pre-wrap styling.
+This is useful for verbose error messages.`;
+
+const sqlErrorDescription = `SQL Error: Syntax error near unexpected token.
+Please check your query and ensure it follows the correct syntax.`;
+
+const detailsExample = `Additional details about the issue are provided here.
+This content is shown when the user clicks "Show more".`;
+
+const ErrorCard: React.FC<{ children: React.ReactNode }> = ({ children }) => (
+  <Card>{children}</Card>
+);
+
+export default {
+  title: 'Components/ErrorAlert',
+  component: ErrorAlert,
+} as Meta;
+
+export const Gallery: StoryFn = () => (
+  <Layout>
+    <Content style={{ padding: '24px' }}>
+      <h2>Non-Compact Errors</h2>
+      <Row gutter={[16, 16]}>
+        <Col xs={48} sm={24} md={16} lg={16} xl={12}>
+          <ErrorCard>
+            <ErrorAlert message="Only message props was passed here" />
+          </ErrorCard>
+        </Col>
+        <Col xs={48} sm={24} md={16} lg={16} xl={12}>
+          <ErrorCard>
+            <ErrorAlert
+              errorType="Database Connection Error"
+              type="warning"
+              message="Failed to connect to database"
+              descriptionDetails={detailsExample}
+              descriptionDetailsCollapsed
+            />
+          </ErrorCard>
+        </Col>
+        <Col xs={48} sm={24} md={16} lg={16} xl={12}>
+          <ErrorCard>
+            <ErrorAlert
+              errorType="Error"
+              message="SQL Syntax Error - No defaults set here"
+              description={sqlErrorDescription}
+              descriptionDetails={detailsExample}
+              descriptionDetailsCollapsed
+              descriptionPre
+            />
+          </ErrorCard>
+        </Col>
+        <Col xs={48} sm={24} md={16} lg={16} xl={12}>
+          <ErrorCard>
+            <ErrorAlert
+              errorType="Error"
+              message="See the details below"
+              type="error"
+              description={longDescription}
+              descriptionDetails={detailsExample}
+              descriptionDetailsCollapsed={false}
+            />
+          </ErrorCard>
+        </Col>
+        <Col xs={48} sm={24} md={16} lg={16} xl={12}>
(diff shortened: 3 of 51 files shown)


superset-frontend/src/components/ErrorMessage/ErrorMessageWithStackTrace.tsx
```diff
@@ -66,28 +68,27 @@ export default function ErrorMessageWithStackTrace({
   if (fallback) {
     return <>{fallback}</>;
   }
+  const computedDescriptionDetails =
+    descriptionDetails ||
+    (link || stackTrace ? (
+      <>
+        {link && (
+          <a href={link} target="_blank" rel="noopener noreferrer">
+            (Request Access)
```
geido, line 77: Should we localize this?

superset-frontend/src/components/ErrorMessage/InvalidSQLErrorMessage.test.tsx
```diff
@@ -44,83 +43,66 @@ const defaultProps = {
...
+    // Validate SQL and caret indicator
+    expect(getByText('SELECT * FFROM table')).toBeInTheDocument();
 
-    const body = getByTestId('body');
-    expect(body).toContainHTML('<pre>SELECT * FFROM table</pre>');
-    expect(body).toContainHTML('<pre>         ^</pre>');
+    // Check for caret (`^`) under the error column
+    const preTags = container.querySelectorAll('pre');
+    const secondPre = preTags[1];
+    console.log({ secondPre });
```
geido, line 72:
```suggestion
```

geido:
From some functional testing:

- A bit of an edge case but the error looks like it is not correctly padded in full page width
Screenshot 2025-01-15 at 17 26 57

- Is a chart failure considered a warning?

Screenshot 2025-01-15 at 17 28 55

- Same, for a chart in Explore. I am thinking these should probably be errors. 

Screenshot 2025-01-15 at 17 30 49

- The positioning of the icon on the left of the text looks off.

- In general, the text color of these alerts seem to stick out a bit too much. I know we want be as vanilla as possible but I'd like to hear @kasiazjc opinion too.

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
```diff
@@ -16,250 +16,128 @@
...
+
+  if (compact) {
+    return (
+      <>
+        <Tooltip title={`${errorType}: ${message}`}>
+          <span role="button" onClick={() => setShowModal(true)} tabIndex={0}>
+            {renderTrigger()}
+          </span>
+        </Tooltip>
+        <Modal
```
geido, line 127: Should we remove the ability to close the alert within a Modal? It seems odd to be able to close an alert which is within a Modal container.

Screenshot 2025-01-15 at 17 39 58

superset-frontend/src/components/ErrorMessage/ErrorAlert.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+const ErrorCard: React.FC<{ children: React.ReactNode }> = ({ children }) => (
+  <Card>{children}</Card>
+);
+
+export default {
+  title: 'Components/ErrorAlert',
+  component: ErrorAlert,
+} as Meta;
+
+export const Gallery: StoryFn = () => (
```
geido, line 44: This is not coming with this PR but I find it odd that an ErrorAlert can be of type info or warning. Naming here could probably be improved.

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
```diff
@@ -16,250 +16,128 @@
...
-interface ErrorAlertProps {
-  body: ReactNode;
-  copyText?: string;
-  level: ErrorLevel;
-  source?: ErrorSource;
-  subtitle: ReactNode;
-  title: ReactNode;
-  description?: string;
+import { useState } from 'react';
+import { Modal, Tooltip } from 'antd';
```
geido, line 20: Why importing from `antd` directly? From one side this is the previous version of `antd` and on the other hand, I think we said to import components from our wrappers not from `antd` directly.

superset-frontend/spec/helpers/theming.tsx
```diff
@@ -55,3 +57,6 @@ export function styledShallow(
     },
   });
 }
+
+export const renderWithTheme = (component: JSX.Element) =>
```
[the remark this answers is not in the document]
mistercrunch: I got confused by this, and remember getting some error about the theme provider not being in-context, maybe I imported `render` directly from the wrong place originally.... Getting rid of it since it's handled in helpers

superset-frontend/src/components/ErrorMessage/ErrorAlert.stories.tsx
mistercrunch: Right, maybe limiting to warning/error here makes more sense here. I think there may be one or two edge cases where `ErrorAlert` is used to `info`, maybe cause someone wanted some of the features in here. We should probably change those to use a normal `Alert`

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
mistercrunch: Probably, would be nice to force this in this context, though the parent component can just pass the prop too. That may be a thing I didn't fully sort out here -> making sure close is only available where it make sense. Might be worth a sweep reviewing everywhere `ErrorAlert` is use and double checking that we set this right in that context.

superset-frontend/src/components/ErrorMessage/ErrorMessageWithStackTrace.tsx
mistercrunch: yes!

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
mistercrunch: ooops. We really need a lint rule to catch this...

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
mistercrunch: Ok i added a line of logic to handle this

superset-frontend/src/theme/index.ts
```diff
@@ -56,15 +56,6 @@ const baseConfig: ThemeConfig = {
     zIndexPopupBase: supersetTheme.zIndex.max,
   },
   components: {
-    Alert: {
```
mistercrunch, line 59: Let's go vanilla!

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
```diff
@@ -16,250 +16,129 @@
...
+    return (
+      <div style={{ cursor: 'pointer' }}>
+        <span style={{ color }}>{icon} </span>
+        {errorType}
+      </div>
+    );
+  };
+  const preStyle = {
+    whiteSpace: 'pre-wrap',
+    // fontFamily: theme.antd.fontFamilyCode,
```
geido, line 77:
```suggestion
```

geido (changes requested):
There was a E2E failing and this brought me to check SQL Lab which seems broken on this branch. Checked on master and could not repro there.

Screenshot 2025-01-22 at 15 36 23

Also not sure what the other Cypress failures are about, first time seeing this error on our CI:

> Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

I have been looking at other E2E tests and I did not see any failures strictly related to this work so I think we'll be good here once that issue on SQL Lab is fixed.

mistercrunch:
Ok, this PR needs a bit more TLC, will ping after addressing the comments and CI issues

superset-frontend/src/components/ErrorMessage/ErrorAlert.tsx
rusackas: There is one, but it's set to warn.... I'll see about flipping it to Error

rusackas: Done! https://github.com/apache/superset/pull/31963

mistercrunch:
Looks like a lot of the flakiness is there on `master` too, also couldn't repro locally, starting to think we should disable the flaky tests.

Lots of red in the last set of commits, no clear trigger as to what triggered it. 
Screenshot 2025-01-22 at 8 54 32 PM

superset-frontend/cypress-base/cypress.config.ts
```diff
@@ -36,7 +36,7 @@ export default eyesPlugin(
     viewportHeight: 1024,
     projectId: 'ud5x2f',
     retries: {
-      runMode: 2,
+      runMode: 3,
```
sadpandajoe, line 39: Heads up that with this and the automatic retry, there is a potential of running this test 9 times which could extend run time.

mistercrunch: Right, I know, getting a bit desperate here, seems the more I disable the more seem to fail...

mistercrunch: Wondering if we're best having a strick no-flakiness policy and less retries (?) Suck to wait 6 or 9 times the time to get a true negative....

.github/workflows/superset-e2e.yml
```diff
@@ -141,4 +141,4 @@ jobs:
         if: failure()
         with:
           path: ${{ github.workspace }}/superset-frontend/cypress-base/cypress/screenshots
-          name: cypress-artifact-${{ github.run_id }}-${{ github.job }}
+          name: cypress-artifact-${{ github.run_id }}-${{ github.job }}-${{ matrix.browser }}-${{ matrix.parallel_id }}
```
mistercrunch, line 144: This prevents collisions across the matrix

eschutho (approved):
LGTM

superset-frontend/cypress-base/cypress.config.ts
sadpandajoe: i wonder if any of them are dependent on other tests we have disabled, but if it's taking more time to pass, i think just go with it for now.

sadpandajoe (approved):
All the cypress changes LGTM

geido (approved):
Thanks!

PR #19598, opened 2022-04-07
feat: add empty states to sqlab editor and select

pkdotson:
### SUMMARY
This pr creates an empty state for the dropdown for the table selector and query editor when there is no database selected.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
after

https://user-images.githubusercontent.com/17326228/162277445-cefc2bdf-6ee5-4ce4-912d-dafb85dda4fc.mov

### TESTING INSTRUCTIONS

Go to sql editor. If there is no db selected to query there should be an empty state for the query editor. If there is no db connected to db select dropdown should show empty state with link to foward user to db listview.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #31860, opened 2025-01-15
chore: Empty state refactor

mistercrunch:
- making `size` a prop instead of one-component-per-size
- fixing up the svg so its themable using `currentColor`
- centralizing/fixing up references
- simplify empty message in home page to simply "Nothing here yet" as the context in which the empty component shows up is clear enough, and the previous approach wast over-engineered


superset-frontend/src/SqlLab/components/SqlEditorLeftBar/index.tsx
```diff
@@ -113,7 +113,7 @@ const SqlEditorLeftBar = ({
     'schema',
   ]);
 
-  const [emptyResultsWithSearch, setEmptyResultsWithSearch] = useState(false);
+  const [_emptyResultsWithSearch, setEmptyResultsWithSearch] = useState(false);
```
geido, line 116: Do we actually need this?

mistercrunch: I think we need `setEmptyResultsWithSearch`, and are flushing position 1 of the array. Code is a bit cryptic to me here. Not sure why we need `setEmptyResultsWithSearch`

mistercrunch: Since it's meant to be flushed, I could put it in a `_` to make it more clear (and keep the spread operator) or hard reference `useState(false)[1]`, WDYT?

superset-frontend/src/features/datasets/AddDataset/LeftPanel/index.tsx
```diff
@@ -178,13 +178,30 @@ export default function LeftPanel({
...
+    />
+  );
 
   return (
     <LeftPanelStyle>
       <TableSelector
         database={dataset?.db}
         handleError={addDangerToast}
-        emptyState={emptyStateComponent(false)}
+        emptyState={getDatabaseEmptyState(false)}
```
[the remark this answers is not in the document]
mistercrunch: Here I simply moved and renamed the function to a better name/location, didn't want to alter the logic.

superset-frontend/src/SqlLab/components/SqlEditorLeftBar/index.tsx
geido: I am not sure that we need this. It looks that originally the [PR that introduced this ](https://github.com/apache/superset/pull/19598/files#diff-bbf91464669d98b9e120f57ce26155804f7238a277fcc30c0d805a943f6c73edR190) was just meant to show some alternate text. I did some testing and I could not see anything breaking with just removing this entirely.

geido (approved):
Looks great. Thanks for doing this. Just a few nits above but approving otherwise.

superset-frontend/package-lock.json
```diff
@@ -1702,15 +1702,18 @@
       }
     },
     "node_modules/@babel/helper-string-parser": {
-      "version": "7.24.8",
-      "resolved": "https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.24.8.tgz",
-      "integrity": "sha512-pO9KhhRcuUyGnJWwyEgnRJTSIZHiT+vMD0kPeD+so0l7mxkMT19g3pjY9GTnHySck/hDzq+dtW/4VgnMkippsQ==",
+      "version": "7.25.9",
```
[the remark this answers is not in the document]
mistercrunch: I think I bumped babel-types (and maybe other libs (?)) to try and prevent some warning/console messages. I think they're not directly related to the PR but net-positive, will leave them in as they might be related to things I changed here.

PR #31888, opened 2025-01-16
fix(import): ensure charts and datasets overwrite on report import

mcdogg17:
### SUMMARY

Previously, when importing a report in Superset, charts and datasets were not being overwritten, unlike when they were imported separately. 

This update fixes the issue by ensuring the overwrite parameter is correctly passed to the import_database, import_dataset, and import_chart functions. Now, when importing a report, charts and datasets will be properly updated, eliminating the need for manual re-importing.

Additionally, query_context handling in slice.py has been improved to correctly pass the datasource to QueryContextFactory.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/dashboard/importers/v1/__init__.py
+++ b/superset/commands/dashboard/importers/v1/__init__.py
@@ -86,7 +86,7 @@ class ImportDashboardsCommand(ImportModelsCommand):
         database_ids: dict[str, int] = {}
         for file_name, config in configs.items():
             if file_name.startswith("databases/") and config["uuid"] in database_uuids:
-                database = import_database(config, overwrite=False)
+                database = import_database(config, overwrite=overwrite)
                 database_ids[str(database.uuid)] = database.id
 
         # import datasets with the correct parent ref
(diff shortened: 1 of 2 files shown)


mcdogg17:
@supersetbot orglabel

rusackas:
I defer to @betodealmeida and @dpgaspar here, but I can imagine somoene importing a report, and _accidentally_ overwriting a bunch of charts and datasets and impacting the dashboards of untold numbers of users. Seems dangerous to me. It seems like it'd be ideal if instead of _overwriting_ an asset, we could create a *version* of the asset. That's been discussed a bit, but is dependent on a yet-to-come SIP.

sadpandajoe:
Might be a feature, but would be good to have a checkbox on imports that ask if the user wants to overwrite. Would also fix workflows where we want to know if things should be overwritten or not.

Vitor-Avila:
I agree this should be optional, like a checkbox in the import dialog to "Also overwrite dependencies" or etc, defaulted to un-checked. 

Overwriting all dependencies is currently already possible via the `/api/v1/assets/import/` endpoint, but having this as a new feature via the UI is definitely helpful! 🙌

mcdogg17:
@supersetbot orglabel

mcdogg17:
Hi! I'll take into account all the comments above and get back to you in a while.

rusackas:
Looks like we don't have that confirmation yet that @sadpandajoe suggested. I'll convert this to draft for now while we await that.

rusackas:
It's been a few months since we heard back here, so I'll close this out. Feel free to re-open if/when the time is right. Many thanks for this and any future contribution!

PR #31905, opened 2025-01-17
chore: Reduces the form_data_key length

michael-s-molina:
### SUMMARY
When the `form_data_key` was created, it was used to share state between Superset users and required a secure key length. When permalinks were introduced, the `form_data_key` role changed to only be used for holding and transferring temporary state for a user session and not more for sharing state between Superset users. This PR reduces the `form_data_key` length given that an 8 bytes key is sufficient for its new role.

As an example, this will be the diff:

```
form_data_key=yjxRBK8zEhgHTs2PhFKLpA3gja9yxkBRdVPP4xRtLFO2fipTttSpFjZEbizBbyhm

form_data_key=z6OiPvci1q0
```

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/key_value/shared_entries.py
+++ b/superset/key_value/shared_entries.py
@@ -42,6 +42,7 @@ def set_shared_value(key: SharedKey, value: Any) -> None:
 def get_permalink_salt(key: SharedKey) -> str:
     salt = get_shared_value(key)
     if salt is None:
-        salt = random_key()
+        # Use a 48 bytes salt
+        salt = random_key(48)
         set_shared_value(key, value=salt)
     return salt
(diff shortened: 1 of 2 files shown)


rusackas:
Justin and I were wondering if this is indeed safe, or if it might be storing longer strings for some and cause side effects.

michael-s-molina:
> Justin and I were wondering if this is indeed safe, or if it might be storing longer strings for some and cause side effects.

This change will only affect the generation of new permalinks. Old ones are not impacted.

rusackas (approved):

PR #31908, opened 2025-01-17
chore(fe): dev deps cleanup 

hainenber:
chore(fe): dev deps cleanup 

### SUMMARY

- I find these babel-plugin-typescript-to-proptypes` and `html-webpack-plugin` are of development nature. They are Babel/Webpack plugin respectively.
image

- `node-fetch` as dev deps is obsolete as `npm run test` can be done without it.

- `xdm` is no longer needed as `npm run storybook` is successful without it. The dep's omission removed 62 packages.

image
image

N/A

### TESTING INSTRUCTIONS

All CI checks pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Tell me what you think, but I think effectively it doesn't matter what's dev or not dev dependencies as we simply build assets here and bundle a webpack package with everything that's needed as part of it. We could decide to put everything here as dev or non-dev and it wouldn't matter. It does matter in `packages/` though, just not in this main `package.json`. I might be missing something though

hainenber:
Yes, there isn't a clear distinction so this is more of a definition correction for them Babel and Webpack plugin. There might be some vulnerability scanners employing heuristics of counting affected `dependencies` towards vulnerability count instead of `devDeps` so it might help there as well.

mistercrunch (approved):
Agreed, didn't think of the security dep scanners and implications there.

PR #31910, opened 2025-01-17
fix: add catalog to cache key when getting tables/views

betodealmeida:
### SUMMARY

The cache key for `get_all_table_names_in_schema` and `get_all_view_names_in_schema` doesn't include the catalog.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -791,7 +791,7 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
         return self.sqlalchemy_uri
 
     @cache_util.memoized_func(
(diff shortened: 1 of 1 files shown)


villebro (approved):
Oof! LGTM. I wish we had unit tests for this logic.

betodealmeida:
> Oof! LGTM. I wish we had unit tests for this logic.

One thing I want to do is write a CI action that checks if the return value of cached functions has changed, and warns in the PR that the cache needs to be purged. We could also make it check if all arguments are used in the cache key.

PR #31929, opened 2025-01-20
fix(api): loading viz_type from request params

hdahme:
### SUMMARY
Previously, when passing `viz_type` via the API, it would default to table. This change allows us to send viz_type and have the charts respect it in the UI

### TESTING INSTRUCTIONS
Make an API request via python and send the following params

```bash
params = {
        "viz_type": "big_number",
        "groupby": columns,
        "metrics": metrics,
        "adhoc_filters": [
            {
                "expressionType": "SQL",
                "sqlExpression": where_clause,
                "clause": "WHERE",
            }
        ]
    }
```

The resulting chart would always be a table. No longer

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/chart/create.py
+++ b/superset/commands/chart/create.py
@@ -43,6 +43,16 @@ class CreateChartCommand(CreateMixin, BaseCommand):
     def __init__(self, data: dict[str, Any]):
         self._properties = data.copy()
 
+        # Extract viz_type from params if it exists
+        if 'params' in self._properties:
+            try:
+                import json
+                params = json.loads(self._properties['params'])
+                if 'viz_type' in params:
+                    self._properties['viz_type'] = params['viz_type']
+            except json.JSONDecodeError as ex:
+                exceptions.append(ex)
+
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM, but adding additional reviewers who might know better :)

rusackas:
This is still failing CI, but we'd love to see it through. Can you address whatever's busted, and (at least preferably) add some test coverage?

rusackas:
Hi @hdahme, thanks for this fix — the underlying bug is real and worth addressing. The PR currently has failing CI and no test coverage for the changed lines. We're converting it to draft for now. When you're able to fix the CI failures and add a test or two covering the `viz_type` loading behavior, please mark it as "Ready for review" and we'll take a look.

rusackas:
Thanks so much for this contribution @hdahme! The fix is solid — we've adopted it and opened #40536 to get it over the finish line. The reason we needed a new PR rather than pushing to your branch directly is that your fork (interval-network/superset) appears to be archived and is read-only. Hope to see you contributing again!

PR #31948, opened 2025-01-21
fix: Load cached DB metadata as DatasourceName and add catalog to schema_list cache key

Vitor-Avila:
### SUMMARY
The `get_all_table_names_in_schema` and `get_all_view_names_in_schema` methods might return data from cache (if `force` isn't set to `True`). Metadata is serialized before getting cached, so currently if cached results are returned, the logic would throw an error as it would try to access these as `DatasourceName` named tuples directly.

This PR changes the logic so that the result of these methods are then loaded as `DatasourceName`s, so that cached results can be used properly.

I also added the catalog name to the `schema_list` cache key.

**Before**
An error is displayed when trying to load cached table list:
image

**After**
List loads properly.

### TESTING INSTRUCTIONS
1. Create a DB connection.
2. Set a cache timeout for table metadata.
3. Load tables from a schema.
4. Refresh the browser.
5. The list should load properly.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/tables.py
+++ b/superset/commands/database/tables.py
@@ -32,6 +32,7 @@ from superset.daos.database import DatabaseDAO
 from superset.exceptions import SupersetException
 from superset.extensions import db, security_manager
 from superset.models.core import Database
+from superset.utils.core import DatasourceName
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Nice catch!

michael-s-molina:
@Vitor-Avila It would be interesting to open a follow-up to add tests for this change. @sadpandajoe mentioned that any contributor could remove the datasource reference without knowing the impact.

Vitor-Avila:
@michael-s-molina @sadpandajoe that's a good point. I've added some additional tests/updated some existing tests and also added a comment to the code on this follow up PR: https://github.com/apache/superset/pull/32054

PR #31959, opened 2025-01-22
refactor: upload data unification, less permissions and less endpoints

### LINKED ISSUE #28352: 94 - Remove `can csv upload`, `can excel upload`, `can columnar upload` in favour of just `can upload on Database`
After the upload data to database refactor, we are ready to simplify permissions and code.
We currently have 6 REST API endpoints to support this feature and their multiple file types and 3 distinct permissions 
(one for each file type). We can easily reduce this to 2 REST API endpoints and 1 permission. 

ref: https://github.com/apache/superset/pull/28192

dpgaspar:
### SUMMARY

This pull request unifies csv, excel and columnar upload endpoints and permissions into a single endpoint and permission type. 

This is a breaking change related with Apache Superset 5.0
Removes the following endpoints:
- POST `/api/v1/database/csv_metadata`
- POST `/api/v1/database/excel_metadata`
- POST `/api/v1/database/columnar_metadata`

- POST `/api/v1/database/<id>/csv_upload`
- POST `/api/v1/database/<id>/excel_upload`
- POST `/api/v1/database/<id>/columnar_upload`

And adds the following generic endpoints:
- POST `/api/v1/database/upload_metadata`
- POST `/api/v1/database/<id>/upload`

Removes the following permissions:
`can csv upload on Database`
`can excel upload on Database`
`can columnar upload on Database`

And adds the following permission: `can upload on Database` 

Includes a db migration that should run fast, depending on the number of roles that include the `can <file-type> upload on Database` permission.
Upgrade and downgrade were tested, but be aware that we loose granularity when downgrading, since we are converging 3 permissions into 1, the downgrade will always expand the final permission into 3 permissions.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
  - [X] Runtime estimates and downtime expectations provided


mistercrunch (approved):
Did a fairly thorough review, paying special attention to the db migration and LGTM!

eschutho:
@dpgaspar I know that we haven't put up the SIP yet (still on my todo list) but I think we should consider api semver as a separate thing from application semver. In other words, we should only remove api endpoints when the entire v1 api is deprecated and removed.

dpgaspar:
> @dpgaspar I know that we haven't put up the SIP yet (still on my todo list) but I think we should consider api semver as a separate thing from application semver. In other words, we should only remove api endpoints when the entire v1 api is deprecated and removed.

Sounds very good for the future. We have been applying semver at the application level only, and taking the opportunity to replace APIs and change specs. This change was approved here: https://github.com/orgs/apache/projects/345?pane=issue&itemId=61955191&issue=apache%7Csuperset%7C28352.
Most probably when we start an API v2 we don't need to remove APIs, both can live side by side, the downside is that we won't be able to remove API permissions immediately, only when we remove v1 for example.
We should also use alpha and beta first for new versions to get the specs stabilised.

eschutho:
@dpgaspar and I talked about the impact of this change, and it's not a commonly-used endpoint, so we'll go with the current flow which is that breaking api changes are permissible during an application breaking change release, and I'll plan on putting up a SIP soon so that we can discuss a new policy for having a separate semver for the api.

eschutho (approved):

ahmedrazakhan1224-alt (approved):

PR #31981, opened 2025-01-24
chore(i8n:spanish): complete-spanish-translations

EmmanuelCbd:
### SUMMARY
Added missing translations to spanish language

### TESTING INSTRUCTIONS
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset/datasource/api.py
+++ b/superset/datasource/api.py
@@ -31,7 +31,12 @@ from superset.exceptions import SupersetSecurityException
 from superset.extensions import cache_manager
 from superset.superset_typing import FlaskResponse
 from superset.utils import json
-from superset.utils.core import apply_max_row_limit, DatasourceType, SqlExpressionType
+from superset.utils.core import (
+    apply_max_row_limit,
+    DatasourceType,
+    parse_boolean_string,
+    SqlExpressionType,
+)
 from superset.views.base_api import BaseSupersetApi, statsd_metrics
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


EmmanuelCbd:
Hi @rusackas ,
Seems like this PR still needs your approval.
Can it be merged, because I'm affraid we'll have many missin translations, while versions are going forward :)
Regards

rusackas:
I'm not sure how this slipped under the collective radar (sorry!) but we're catching up with older PRs and would still love to see this merged. But now it's in need of rebase. Maybe Claude Code can help rebase it easy-peasy nowdays?

tetebueno:
Hey there, I'd be needing this translation complete, is there anything I can do te help?

Also, once this is merged into master, how can I apply this to a Helm installation? I'm currently using 0.15.0 version of the Helm chart. Is there a planned release with this change?

rusackas:
Hi @EmmanuelCbd, thanks again for this contribution — the Spanish translations are much appreciated\! When you get a chance to rebase this on `master`, there are a couple of things worth addressing at the same time. First, the file header currently shows `Language: fr` which should be `Language: es`. Second, there are some word choice differences with the existing terminology in master — for example, master uses `panel` where this PR uses `tablero` for "dashboard", and `añadir` where this PR uses `agregar` for "add". It would be great to align with the existing conventions while you're resolving the conflicts. Let us know if you need any help\!

rusackas (changes requested):
Thanks for sticking with this @EmmanuelCbd — still really want it in. Beyond the rebase and the Language: fr header I flagged earlier (the Plural-Forms also got switched to the French n > 1 rule), I'm seeing a few `msgstr` values that don't match their `msgid`: `%(name)s.pdf` is translated as `%(name)s.csv`, and `% calculation`/`% of parent` come back as "Elegir un tipo de cálculo"/"Padre". Could you sweep those while you're in here? Happy to help if I can. I've also since updated a bunch of spanish translations, maybe the ones on master are better now?

rusackas:
@EmmanuelCbd I'll rebase this and flip the `Language: fr` header to `es` while I'm in there — would love to finally land it. Holler if you'd rather drive it yourself.

rusackas:
Thanks @EmmanuelCbd, and sorry for the glacial turnaround here!

Quick heads up on something I caught while reviewing: the branch had regenerated es/messages.po against a stale catalog, which quietly dropped ~1,350 source strings (most already translated) and pulled coverage down from ~97% to ~72%. So as-is it would have been be a net regression rather than a completion, which I don't think is what either of us was going for :)

I went ahead and re-based the catalog on current master so the dropped translations come back, kept your reworded strings layered on top, and held back a couple dozen that either reverted Spanish to the English source or broke %(...)s placeholders (those would throw at runtime). Pushed it just now... we'll see what CI thinks.

The thing actually worth a conversation: a lot of the rewordings shift register toward formal/LatAm "usted" (Haz clic → Haga clic, panel de control → tablero, etc.), where the existing catalog mostly sits at tú. Both are fine in isolation, but half-and-half is the one result that's worse than either. My inclination would be to go with tú, leaning on impersonal/infinitive phrasing where it reads naturally so we sidestep the choice when possible... but I'd really like input from native Spanish speakers before we lock anything in. Might even be worth writing it down as an es style convention so we don't hit this kind of confusion again on every translation PR. Thoughts?

rusackas:
Went back through my own rebase commit and the `%(name)s.pdf` mismatch I flagged earlier (still translated as `.csv`) is still in there. `msgfmt --check-format` also finds 6 fatal errors on this branch that `master` doesn't have, including `Applied filters (%s)` coming back as `Filtros aplicados (%d)`, which throws at runtime for non-integer values.

Still curious what folks think on the tú vs usted question from before, too. Let me know if you have feelings there, want me to get some opinions (or want to raise it on the dev@ mailing list), or if I should... just make an executive decision here to merge and move forward.

rusackas:
Thoughts about just going with the `tú` approach here, @EmmanuelCbd? Happy to push commits and get this fixed up if you'd like a hand.

EmmanuelCbd:
I'm sorry @rusackas , I missed all the answers.

I currently have a lot to do at work and so, as you might have already seen, I'm less available (at least until september).
I'll take a moment to see what I can do to move forward.

Regards

EmmanuelCbd:
Hi @rusackas, thanks for all the detailed feedback and for keeping this alive.

Quick update: the colleague who did the original Spanish translation work has left the company, so I went ahead and addressed the remaining issues myself (with some help from Claude AI), and I'd like the community's eyes on the wording before we merge.

What I fixed, on top of your rebase:

- **The two runtime bugs you flagged**: `%(name)s.pdf` was mistranslated as `%(name)s.csv`, and `Applied filters (%s)` was using `%d` (would crash on non-integer values). Both fixed.
- **tú vs usted**: went with `tú` throughout, matching master's dominant convention (~30 `Selecciona`-style forms vs a couple of `usted` outliers). Also aligned `tablero`→`panel de control` and `agregar`→`añadir` to match existing terminology.
- **Dropped placeholders**: found ~60 more entries where `%s`/`%(name)s` had been dropped and replaced with a fixed word (e.g. `Edit %s` → `Editar usuario`). Turns out these are shared templates called from multiple places with a runtime value — e.g. `t('Edit %s', datasetLabelLower())` is called for datasets, charts, semantic views, etc. So the hardcoded translation was only ever correct for one of those cases. I checked each one against its actual call site in `superset-frontend/src`/`superset/` and restored the placeholder with a translation built around it.

`pybabel compile`'s format-string check went from 65 errors down to 3, and those 3 remaining (`% calculation`, `% of parent`, `% of total`) are checker false positives — the literal `% c`/`% o` substring gets misread as a format directive even though no argument is ever substituted there.

None of this is native-speaker-reviewed — I don't speak Spanish myself, so please treat the wording as a proposal rather than a final answer. Happy to adjust based on feedback, or if anyone wants to just push corrections directly, that works too.

rusackas (changes requested):
Thanks for digging back in on this, @EmmanuelCbd, looks like great progress! 

Looks like something went sideways with a rebase, though, and some files got pulled in that shouldn't have, which are what's triggering the bot reviews. I'll block the PR from merging for that, just for the moment so it doesn't get merged by accident. Happy to rebase and clean up if needed, let me know.

EmmanuelCbd:
Yes, @rusackas , if you can take it from there I'll be glad.
I'm not very at ease with rebase and stuff (and don't want to let Claude to mess anything), I prefer that a more expertised guy take the next steps.

Thanks for the support :)

rusackas:
Went back through this against current `master` to see what's actually left to merge. Turns out the Spanish catalog has mostly caught up on its own since this branch started. Only a handful of the untranslated/fuzzy strings match anything unique in this diff, and none of it moves the needle.

Closing, but I really appreciate you sticking with this through the rebase mess and the bug hunting, @EmmanuelCbd. Open a fresh PR if you spot untranslated strings out there, still very welcome.

PR #31984, opened 2025-01-24
feat: run prettier before eslint in pre-commit hooks

mistercrunch:
Simply changing the ordering here as prettier is set to mutate/fix things, and eslint simply logs and errors out, often duplicating messaging on a single issues. Running prettier first makes more sense here.


michael-s-molina (approved):

mistercrunch:
Related if we want to add tsx pre-commit checks -> https://github.com/microsoft/TypeScript/issues/27379#issuecomment-2611978121

PR #31996, opened 2025-01-27
feat: cache the frontend's bootstrap data

mistercrunch:
While reviewing https://github.com/apache/superset/pull/30134 which seemed to make heavy use of `getBootstrapData`, I realized that it's used 21 times across the codebase, while it may be a fairly expensive operation (JSON.parse on an object that may be significant in size at times (?)).

Now it's easy to do module-level caching here and ensures that `JSON.parse` happens only once per SPA-load, but opening this as a DRAFT as we probably already store it in other places (didn't dig in, but it's probably stored in a redux store or component state somewhere...).

Tradeoff here is compute VS memory, and clearly we need this object in memory somewhere, but probably in most cases, already have it in the app's state somewhere. Ideally the app would read it once and make it available in its state, as opposed to different modules calling this function on-demand...

PR is more of a request for comments.

Grepping for calls ->
```
22:30 $ git grep "getBootstrapData("
spec/helpers/reducerIndex.ts:const bootstrapData = getBootstrapData();
src/SqlLab/actions/sqlLab.js:      ...getBootstrapData(),
src/SqlLab/components/ScheduleQueryButton/index.tsx:const bootstrapData = getBootstrapData();
src/SqlLab/components/SqlEditor/index.tsx:const bootstrapData = getBootstrapData();
src/SqlLab/components/TabbedSqlEditors/index.tsx:    const bootstrapData = getBootstrapData();
src/dashboard/components/Dashboard.jsx:    const bootstrapData = getBootstrapData();
src/dashboard/components/nativeFilters/utils.ts:  const bootstrapData = getBootstrapData();
src/embedded/api.tsx:const bootstrapData = getBootstrapData();
src/embedded/index.tsx:const bootstrapData = getBootstrapData();
src/explore/components/controls/VizTypeControl/index.tsx:const bootstrapData = getBootstrapData();
src/middleware/asyncEvent.ts:  config = appConfig || getBootstrapData().common.conf;
src/pages/ChartCreation/index.tsx:const bootstrapData = getBootstrapData();
src/pages/Home/index.tsx:const bootstrapData = getBootstrapData();
src/preamble.ts:const bootstrapData = getBootstrapData();
src/setup/setupClient.ts:const bootstrapData = getBootstrapData();
src/utils/getBootstrapData.ts:export default function getBootstrapData(): BootstrapData {
src/utils/hostNamesConfig.js:  const bootstrapData = getBootstrapData();
src/views/App.tsx:const bootstrapData = getBootstrapData();
src/views/RootContextProviders.tsx:const { common } = getBootstrapData();
src/views/menu.tsx:const bootstrapData = getBootstrapData();
src/views/store.ts:const bootstrapData = getBootstrapData();
```

Thinking about it, this PR is probably net-positive, but would say that that sprawling calls to `getBootstrapData()` across the codebase is an anti-pattern.

### DIFF
--- a/superset-frontend/src/utils/getBootstrapData.ts
+++ b/superset-frontend/src/utils/getBootstrapData.ts
@@ -29,5 +29,6 @@ export default function getBootstrapData(): BootstrapData {
       ? JSON.parse(dataBootstrap)
       : DEFAULT_BOOTSTRAP_DATA;
   }
-  return cachedBootstrapData;
+  // Add a fallback to ensure the returned value is always of type BootstrapData
+  return cachedBootstrapData ?? DEFAULT_BOOTSTRAP_DATA;
 }
--- a/superset-frontend/src/utils/getBootstrapData.ts
+++ b/superset-frontend/src/utils/getBootstrapData.ts
@@ -16,12 +16,18 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
 import { BootstrapData } from 'src/types/bootstrapTypes';
 import { DEFAULT_BOOTSTRAP_DATA } from 'src/constants';
 
+let cachedBootstrapData: BootstrapData | null = null;
+
 export default function getBootstrapData(): BootstrapData {
-  const appContainer = document.getElementById('app');
-  const dataBootstrap = appContainer?.getAttribute('data-bootstrap');
-  return dataBootstrap ? JSON.parse(dataBootstrap) : DEFAULT_BOOTSTRAP_DATA;
+  if (cachedBootstrapData === null) {
+    const appContainer = document.getElementById('app');
+    const dataBootstrap = appContainer?.getAttribute('data-bootstrap');
+    cachedBootstrapData = dataBootstrap
+      ? JSON.parse(dataBootstrap)
+      : DEFAULT_BOOTSTRAP_DATA;
+  }
+  return cachedBootstrapData;
 }


mistercrunch:
Actually decided to put up for review (from DRAFT) given there should be some benefit given the current 21 calls across the codebase

rusackas:
I guess my question here is whether or not we expect any of the bootstrap data to change while the user has the site loaded. I hardly remember the details of what's in there. This doesn't just cache the data, but seems like it effectively locks it in place with no cache invalidation. We should also confirm it's dumped on logout.

villebro (commented):
Bootstrap data does indeed need to get refreshed from time to time, but we currently have a TTL of 1 min in the backend, which is IMO totally acceptable. However, I don't believe it currently ever changes after initial browser load, so for now I agree it's fine to cache this in the frontend like this. Ofc this will all be totally rewritten once User Preferences lands, but by then we'll have other mechanisms in place for this.

mistercrunch:
I kind of needed a refresher on what's in there or not:
- [root call](https://github.com/apache/superset/blob/827042f12fb4710f75c26abbc3ea67c10a942ede/superset/views/base.py#L208-L212)
- [user-related](https://github.com/apache/superset/blob/827042f12fb4710f75c26abbc3ea67c10a942ede/superset/views/utils.py#L71-L99)
- [common](https://github.com/apache/superset/blob/827042f12fb4710f75c26abbc3ea67c10a942ede/superset/views/base.py#L325-L337)

Quite a bit of junk in there. Trying to summarize:
- basic user info (unclear why it's needed)
- perms (optionally?), guessing perms can be quite large at times (!), probably used to check whether we should enable/disable action buttons and/or menu items (?)
- some users preferences + locale-related stuff
- dynamic menu configs, probably based on user-permissions, like only show subset of the Settings menu with what the user has access to - probably should be merged with the perm-related portion and served by a different mechanism
- flash user messages (this is used for toasts, build on top of flask.flash), this is legacy and we should probably kill this eventually and have the frontend fully handle API calls and user messaging. Low volume, but the real issue here is that really can't count on the next round trip to flash a message to a user in the context of a SPA
- language_pack, probably the most concerning thing here, meaning if you pick a language other than `en` you get a massive payload with all translated strings through this mechanism. That belongs in its own blocking, pre-render fetch, with proper browser caching set up. Here no caching would apply/work I think or lead to large memory usage. Ideally would be served as a static asset from a CDN.

Unclear how dated/bad of a pattern this is in more modern app design... My take would be that it's somewhat reasonable for minimal must-have-data-to-render-anything stuff. Clearly has been abused here... Screams for modern server-side rendering, maybe or just a set of sensible refactor(s).

eschutho (approved):

PR #32004, opened 2025-01-27
refactor(Radio): Upgrade Radio Component to Ant Design 5

EnxDev:
### SUMMARY

Migration of Radio component from Ant Design v4 to Ant Design v5.

- Calendar frame before
Screenshot 2025-01-25 181611

- CalendarFrame after
Screenshot 2025-01-27 213626

- CommonFrame before
Screenshot 2025-01-25 183010

- CommonFrame after
Screenshot 2025-01-25 183220

- CurrentFrame before
Screenshot 2025-01-25 183925

- CurrentFrame after
Screenshot 2025-01-25 183737

- DisplayModeToggle before
Screenshot 2025-01-25 185257

- DisplayModeToggle after
Screenshot 2025-01-25 193315

https://github.com/user-attachments/assets/ca303698-23f1-4529-bca2-5655ad869366

- HeaderWithRadioGroup before
Screenshot 2025-01-25 211904

- HeaderWithRadioGroup after
Screenshot 2025-01-25 212956

- ReportModal before
Screenshot 2025-01-27 173413

- ReportModal  after
Screenshot 2025-01-27 214013

- SaveModal before
Screenshot 2025-01-25 213500

- SaveModal After
Screenshot 2025-01-25 213500

- Before
Screenshot 2025-01-25 211904

https://github.com/user-attachments/assets/83582070-cce2-43c7-850e-98c21dbd2bd9

- After
Screenshot 2025-01-25 212956

https://github.com/user-attachments/assets/de5cc25a-2b3e-46d1-a93e-4c897ea7d6da

### TESTING INSTRUCTIONS

1) CI should pass
2) All existing Radios should work as expected with minimal visual differences

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/Radio/index.tsx
```diff
@@ -16,46 +16,60 @@
...
-    left: 2px;
-    width: ${({ theme }) => theme.gridUnit * 4}px;
-    height: ${({ theme }) => theme.gridUnit * 4}px;
-    border-width: 2px;
-    border-color: ${({ theme }) => theme.colors.grayscale.light2};
-  }
+export type RadioGroupWrapperProps = RadioGroupProps & {
+  useSpace?: boolean;
+  direction?: SpaceProps['direction'];
+  spaceSize?: SpaceProps['size'];
```
geido, line 31: I think a better approach here could be something like this

`spaceConfig={...}`

If this is used then include the Space component. This is to avoid dependencies across different props.

superset-frontend/src/components/Radio/index.tsx
```diff
@@ -16,46 +16,60 @@
...
+  children,
+  ...props
+}: RadioGroupWrapperProps) => {
+  const content = options
+    ? options.map((option: CheckboxOptionType) => (
+        <Radio key={option.value} value={option.value}>
+          {option.label}
+        </Radio>
+      ))
+    : children;
```
geido, line 51: I think we should not be supporting both `children` and `options`. Let's be opinionated here. I think the question is:

- Do we believe people will need to make customizations to Radios within the RadioGroup, if yes, go with `children`. If not, then `options` is a better candidate. Right now it seems the codebase does not require customizations so `options` might be a better solution but let's consider all possible use cases and also whether Ant Design even allows to put components in between Radios within a RadioGroup (like dividers or other elements) or if it only accept a plain sequence of Radios. That might lead us to the right direction.

EnxDev: Hi, Radio.Group doesn't directly support inserting non-radio elements like 'hr' element. 

For this reason, I would prefer to keep the children prop as well, so that in the future, if we need more customized elements, we won't have to modify the component and the entire codebase accordingly.

It's also true that if we need a more complex layout, we can use Radio.Group with Radio components inside.
However, in this way, we wouldn't benefit from standardizing the use of Radio.Group with the relative best practices.

For example, I've often noticed that Radio components are used without being wrapped in a Radio.Group, sometimes styled directly for alignment or combined with a Space component for alignment.

geido: Ok so if it is not supporting custom components, then we can just keep options as you originally suggested and remove children? I am not sure I understand what children is for if we cannot make customizations anyways. Also, the code base does not require that customization right now so I am thinking let's worry about that later if it will ever be needed and for now options would just do the job. Your call.

EnxDev: Inside Radio.Group, it is possible to insert components. 
Within the options prop, you can provide an array structured like this: [{ value: 1, label: (<Space></Space>)}, ]. 
However, if we want to insert an 'hr' element, for example, between one option and another, it is not possible.

this is options that used in Radio.Group: 
options?: (CheckboxOptionType<T> | string | number)[]; 

this is the iterface: 
export interface CheckboxOptionType<T = any> {
    label: React.ReactNode;
    value: T;
    style?: React.CSSProperties;
    disabled?: boolean;
    title?: string;
    id?: string;
    onChange?: (e: CheckboxChangeEvent) => void;
    required?: boolean;
}

superset-frontend/src/components/Chart/DrillBy/useDisplayModeToggle.tsx
```diff
@@ -19,49 +19,38 @@
...
-import { Radio } from 'src/components/Radio';
+import { Radio, CheckboxOptionType } from 'src/components/Radio';
 import { DrillByType } from '../types';
 
 export const useDisplayModeToggle = () => {
   const [drillByDisplayMode, setDrillByDisplayMode] = useState<DrillByType>(
     DrillByType.Chart,
   );
-
+  const customButtons: CheckboxOptionType[] = [
```
kgabryje, line 29: please either wrap it in useMemo or move it out of the hook (it has only static values so should work fine) - otherwise it will trigger rerenders on every render cycle

kgabryje:
Looks like the radio buttons in drill by are separated

image

EnxDev:
> Looks like the radio buttons in drill by are separated
> 
> image

Yes, when I modified the Radio.GroupWrapper, it broke the spacing of the drill-by. I reattached it again, thanks

kgabryje (approved):
lgtm

michael-s-molina:
I believe this broke master CI.

Screenshot 2025-01-31 at 15 54 45

We need to try to fix it before Monday when we'll cut 5.0. Otherwise, we'll need to revert.

michael-s-molina:
It looks like this was a false alarm and we have another flaky test. It succeeded on the third run 😞

PR #32005, opened 2025-01-28
fix(sqllab): tab layout truncated

justinpark:
### SUMMARY
Fixes https://github.com/apache/superset/pull/31858#pullrequestreview-2567399027

In #31858, unnecessary space was created on all pages due to the div with added padding.
|Before|After|
|--|--|
|Screenshot_2025-01-27_at_6_20_03 PM|Screenshot_2025-01-27_at_6_19_44 PM

The intention of this padding was to provide [space for the alert container](https://github.com/apache/superset/pull/31858#issuecomment-2593398152). To resolve this issue, I removed the div and added margin to the ErrorBoundary in the App container instead. 

Screenshot 2025-01-27 at 6 03 47 PM
Additionally, the Layout.Content component caused issues with the full page height, resulting in improper rendering of the SQL Lab tab. This was fixed by adding `display:flex`.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

|Before|After|
|--|--|
|Screenshot 2025-01-27 at 6 20 26 PM|Screenshot 2025-01-27 at 6 20 10 PM|
|Screenshot 2025-01-27 at 6 19 44 PM|Screenshot 2025-01-27 at 6 20 03 PM|

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #32006, opened 2025-01-28
feat: Fix explore dynamic height issues by adding style to ErrorAlertBoundary

mistercrunch:
Recently brought in an intricate issue while trying to style the ErrorBoundary. Somehow was triggering a loop of auto-height-expanding in the Explore app.

Another related issue was around introducing `Layout.Content` in the App.tsx file, was creating issues with SQL Lab rendering the tab-content.

Here my solution is to pass a css prop to the boundary, which carries through to the `Alert` component. Here styling the boundary object ultimately styles the Alert, while passing the style props through ErrorAlert.

Could implement full emotion interface with `className` and `css` but decided to go minimalist and only pass what we need (`style`)
Screenshot 2025-01-27 at 7 28 28 PM

### DIFF
--- a/superset-frontend/src/views/App.tsx
+++ b/superset-frontend/src/views/App.tsx
@@ -29,6 +29,7 @@ import { GlobalStyles } from 'src/GlobalStyles';
 import ErrorBoundary from 'src/components/ErrorBoundary';
 import Loading from 'src/components/Loading';
 import { Layout } from 'src/components';
+import { useTheme } from '@superset-ui/core';
 import Menu from 'src/features/home/Menu';
 import getBootstrapData from 'src/utils/getBootstrapData';
 import ToastContainer from 'src/components/MessageToasts/ToastContainer';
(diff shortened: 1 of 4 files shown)


superset-frontend/src/views/App.tsx
```diff
@@ -69,34 +70,36 @@ const LocationPathnameLogger = () => {
...
-      <Menu
-        data={bootstrapData.common.menu_data}
-        isFrontendRoute={isFrontendRoute}
-      />
-      <Switch>
-        {routes.map(({ path, Component, props = {}, Fallback = Loading }) => (
-          <Route path={path} key={path}>
-            <Suspense fallback={<Fallback />}>
-              <Layout.Content>
-                <div style={{ padding: '16px' }}>
```
mistercrunch, line 87: that was the issue here ^^^

superset-frontend/src/views/App.tsx
```diff
@@ -83,13 +83,9 @@ const App = () => (
...
           <Route path={path} key={path}>
             <Suspense fallback={<Fallback />}>
-              <Layout.Content>
-                <div style={{ padding: '16px' }}>
-                  <ErrorBoundary>
-                    <Component user={bootstrapData.user} {...props} />
-                  </ErrorBoundary>
-                </div>
-              </Layout.Content>
+              <ErrorBoundary style={{ margin: '16px' }}>
```
mistercrunch, line 86: wanted to do `theme.sizeUnit * 4` but theme isn't in context here so bailed and went for hard-coded

kgabryje (approved):

kgabryje:
I think we can close this PR now as https://github.com/apache/superset/pull/32005 is merged

PR #32008, opened 2025-01-28
fix: re-enable cypress checks

mistercrunch:
I got a bit confused and reckless on https://github.com/apache/superset/pull/31858 while dealing with a sizable merge conflict, proceeded to disable many cypress tests, many of which are or may be flaky at times, but in this instance were probably triggered by some real issues in that PR. 

Renabling them all and will monitor PRs and master merges to re-disable the ones flaking as needed.

Note that prior to the PR issues I created on https://github.com/apache/superset/pull/31858, there were all sorts of cypress issues that we couldn't pinpoint to any particular merges, and that led a few of us to make to call to "disable all flaking tests" with the intent to fix or replace with unit tests:
Screenshot 2025-01-27 at 9 41 17 PM

Seems there are other new issues with the way we goup/report tests as well. Still needs weeding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 *.swp
 __pycache__
 
+.aider.*
 .local
 .cache
 .bento*
--- a/scripts/cypress_run.py
+++ b/scripts/cypress_run.py
@@ -60,6 +60,7 @@ def run_cypress_for_test_file(
                 f"--browser {browser} "
                 f"--record --group {group_id} --tag {REPO},{GITHUB_EVENT_NAME} "
                 f"--ci-build-id {build_id} "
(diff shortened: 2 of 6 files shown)


sadpandajoe (approved):
LGTM

michael-s-molina:
@mistercrunch This seems to have broken CI. Should we revert?

Screenshot 2025-01-30 at 12 07 14

mistercrunch:
Oh yes, this is due to checks being different on `master` and PRs, I was trying to address (amongst other things) a master-specific bug on retries, but flying blind. Let's revert for now.

mistercrunch:
Revert is here -> https://github.com/apache/superset/pull/32045

Issue #31723, opened 2025-01-06
No input validations there on UI


chakilamsuryat:
## Screenshot

<img width="1790" alt="Screenshot 2025-01-06 at 4 40 40 PM" src="https://github.com/user-attachments/assets/38e53069-3745-4006-8b8d-2742a7e7a5f4" />

## Description

There is no input validation present for Dashboard, Chart, Annotation Layers, row-level-security, and for user info details

## Design input
In the module specific {charts, dashboards, annotation_layers, row_level_security, etc} schemas.py files, can add input validation like:- validate=[validate.And(Length(1, 250),validate.Regexp(regex='^[a-zA-Z0-9_]+$',error='Special Characters are not Allowed!'))] for the fields for which it is needed like name, description because if no validation is there then any attack is possible like they can give some scripts to be in name/description which might cause a vulnerability. For Userinfo input validations for first_name and last_name can create new view in our superset/security/manager.py and can add like:- [userinfoeditview = NewUserInfoEditView] in SuperSetSecurityManager.

chakilamsuryat:
I have made these all changes as we have some security vulnerability issues raised and we are currently running updated code from our end.

rusackas:
A limit on length sounds reasonable, but I'm not sure what it really ought to be... perhaps that should be configurable.

Similarly, for disallowing characters in names/titles, if there is an XSS vulnerability or something, I'm onboard, but otherwise I think special characters should be fine. I might want a chart of "Top 1% of songs written in D# minor, by $ earned" 

Of course, again I'm fine with imposing a list of "banned" characters, so long as that's part of the config. 

A PR would be welcome. Meanwhile, I'll convert this to an Ideas discussion, since I don't think it's a bug report. CC @dpgaspar as well since this might border on security.

PR #32012, opened 2025-01-28
input validation added for various modules

chakilamsuryat:
### SUMMARY

No input validations there on UI

Before: 
before-dashboard
After:
Screenshot 2025-01-28 at 3 56 48 PM
We will restricting the user for input validation and will thrown error as shown above.

### TESTING INSTRUCTIONS

Can test charts, dashboards, annotations, annotation_layers, row_level_security, user info editing views on the UI for input validation given for names and descriptions.

### ADDITIONAL INFORMATION

There is no input validation present for Dashboard, Chart, Annotation Layers, row-level-security, and for user info details.
This is as part of security fixes for vulnerabilities raised as able to input scripts also currently in the fields of name and descriptions of various views.

### DIFF
--- a/superset/annotation_layers/annotations/schemas.py
+++ b/superset/annotation_layers/annotations/schemas.py
@@ -16,7 +16,7 @@
 # under the License.
 from typing import Union
 
-from marshmallow import fields, Schema, ValidationError
+from marshmallow import fields, Schema, ValidationError, validate
 from marshmallow.validate import Length
 
 from superset.utils import json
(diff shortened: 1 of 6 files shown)


chakilamsuryat:
Can suggest if any changes required from my end as well.

sadpandajoe:
@chakilamsuryat so what happens to any dashboard/chart now that has characters that are no longer allowed? Will they break when a user tries to save?

rusackas:
I added a comment on an [issue](https://github.com/apache/superset/issues/31723#issuecomment-2581153471) that there are plenty of valid use cases for these characters, and we're unaware of any risk (XSS or otherwise). If you spot a security issue, please report it privately to security@superset.apache.org, but otherwise, I'd like to know more about the rationale for the change.

chakilamsuryat:
@sadpandajoe , if given input invalid characters as shared in screenshot, it will give popup of error messaging saying that its not allowed and that name will not save.

chakilamsuryat:
@rusackas ,  Improper Input Validation- It is observed that application fails to validate and sanitize user supplied input containing malicious JavaScript which may be executed on browser leading to Stored Cross-Site Scripting.

sadpandajoe:
> @sadpandajoe , if given input invalid characters as shared in screenshot, it will give popup of error messaging saying that its not allowed and that name will not save.

@chakilamsuryat so we only show that this isn't possible after a user may have already made edits on an existing dashboard and trying to save then we let them know that they can't save? That seems pretty bad for the user. Also if they save as a new dashboard, then anything shared will have to be reshared with users.

sadpandajoe:
I believe we have already fixed this in the past and that's why we do allow these characters. I'm going to mention @dpgaspar as he might remember.

rusackas:
Closing this one out... I don't think it's needed, but happy to continue the discussion if you feel otherwise! We appreciate the effort (especially when security is concerned) so thank you for your contribution!

Issue #31890, opened 2025-01-16
docker compose up --build is broken, fails with werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. 


ludaxiansheng:
### Bug description

Bug description
```
docker 构建正常,启动并登录后跳转页面错误,并且superset_app出现错误,错误如下
2025-01-16 11:35:39,016:WARNING:superset.views.error_handling:HTTPException
superset_app          | Traceback (most recent call last):
superset_app          |   File "/app/.venv/lib/python3.11/site-packages/flask/app.py", line 1484, in full_dispatch_request
superset_app          |     rv = self.dispatch_request()
superset_app          |          ^^^^^^^^^^^^^^^^^^^^^^^
superset_app          |   File "/app/.venv/lib/python3.11/site-packages/flask/app.py", line 1469, in dispatch_request
superset_app          |     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
superset_app          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
superset_app          |   File "/app/.venv/lib/python3.11/site-packages/flask/app.py", line 502, in <lambda>
superset_app          |     view_func=lambda **kw: self_ref().send_static_file(**kw),  # type: ignore # noqa: B950
superset_app          |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
superset_app          |   File "/app/.venv/lib/python3.11/site-packages/flask/scaffold.py", line 317, in send_static_file
superset_app          |     return send_from_directory(
superset_app          |            ^^^^^^^^^^^^^^^^^^^^
superset_app          |   File "/app/.venv/lib/python3.11/site-packages/flask/helpers.py", line 557, in send_from_directory
superset_app          |     return werkzeug.utils.send_from_directory(  # type: ignore[return-value]
superset_app          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
superset_app          |   File "/app/.venv/lib/python3.11/site-packages/werkzeug/utils.py", line 575, in send_from_directory
superset_app          |     raise NotFound()
superset_app          | werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
[...]

PR #32017, opened 2025-01-29
fix: eph env + improve docker images to run in userspace

mistercrunch:
While digging into fixing ephemeral envs, I realized that running docker-compose as `superset` didn't work out of the box. `docker-compose.yml` sets `root` as the user which goes around the issue while running `compose`. When switching to user `superset`, there were some permission issues around excecuting certain scripts and such.

Clearly it's better/more-secure to run the app as a non-root user. So here I'm making some changes and simplifying things, paying attention to a few layers, and avoiding doing any sorts of installation of anything while starting the app as the `superset` user in the tree of entrypoint scripts.

Adding notes in PR/code to explain the changes step-by-step

Tested locally:
-  `docker-compose up` while using the default `root` user
-  `docker-compose up` while using `superset` user
-  `docker-compose -f docker-compose-image-tag.yml up`
-  `docker-compose build --build-arg BUILD_TRANSLATIONS=true`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ COPY superset-frontend /app/superset-frontend
 FROM superset-node-ci AS superset-node
 
 # Build the frontend if not in dev mode
-RUN --mount=type=cache,target=$/root/.npm \
+RUN --mount=type=cache,target=/root/.npm \
     if [ "$DEV_MODE" = "false" ]; then \
         echo "Running 'npm run ${BUILD_CMD}'"; \
         npm run ${BUILD_CMD}; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,8 +76,7 @@ COPY superset-frontend /app/superset-frontend
 FROM superset-node-ci AS superset-node
 
 # Build the frontend if not in dev mode
-RUN --mount=type=cache,target=/app/superset-frontend/.temp_cache \
-    --mount=type=cache,target=/root/.npm \
+RUN --mount=type=cache,target=$/root/.npm \
     if [ "$DEV_MODE" = "false" ]; then \
         echo "Running 'npm run ${BUILD_CMD}'"; \
         npm run ${BUILD_CMD}; \
@@ -100,21 +99,14 @@ RUN if [ "$BUILD_TRANSLATIONS" = "true" ]; then \
 # Base python layer
 ######################################################################
 FROM python:${PY_VER} AS python-base
-ARG BUILD_TRANSLATIONS="false" # Include translations in the final build
-ENV BUILD_TRANSLATIONS=${BUILD_TRANSLATIONS}
-ARG DEV_MODE="false"           # Skip frontend build in dev mode
-ENV DEV_MODE=${DEV_MODE}
 
-ENV LANG=C.UTF-8 \
-    LC_ALL=C.UTF-8 \
-    SUPERSET_ENV=production \
-    FLASK_APP="superset.app:create_app()" \
-    PYTHONPATH="/app/pythonpath" \
-    SUPERSET_HOME="/app/superset_home" \
-    SUPERSET_PORT=8088
+ARG SUPERSET_HOME="/app/superset_home"
+ENV SUPERSET_HOME=${SUPERSET_HOME}
 
-
-RUN useradd --user-group -d ${SUPERSET_HOME} -m --no-log-init --shell /bin/bash superset
+RUN mkdir -p $SUPERSET_HOME
+RUN useradd --user-group -d ${SUPERSET_HOME} -m --no-log-init --shell /bin/bash superset \
+    && chmod -R 1777 $SUPERSET_HOME \
+    && chown -R superset:superset $SUPERSET_HOME
 
 # Some bash scripts needed throughout the layers
 COPY --chmod=755 docker/*.sh /app/docker/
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -76,8 +76,7 @@ COPY superset-frontend /app/superset-frontend
 FROM superset-node-ci AS superset-node
 
 # Build the frontend if not in dev mode
-RUN --mount=type=cache,target=/app/superset-frontend/.temp_cache \
-    --mount=type=cache,target=/root/.npm \
+RUN --mount=type=cache,target=$/root/.npm \
```
mistercrunch, line 79: Here caching the cache mount was preventing the docker-compose-level mount from hooking up the cache to the host. Getting a nice perf boost by letting the host mount do what it's supposed to.

Dockerfile
```diff
@@ -100,21 +99,14 @@ RUN if [ "$BUILD_TRANSLATIONS" = "true" ]; then \
 # Base python layer
 ######################################################################
 FROM python:${PY_VER} AS python-base
-ARG BUILD_TRANSLATIONS="false" # Include translations in the final build
```
mistercrunch, line 103: Moving this section to the python-common layer that's used for the app itself.

Dockerfile
```diff
@@ -100,21 +99,14 @@ RUN if [ "$BUILD_TRANSLATIONS" = "true" ]; then \
...
-ENV BUILD_TRANSLATIONS=${BUILD_TRANSLATIONS}
-ARG DEV_MODE="false"           # Skip frontend build in dev mode
-ENV DEV_MODE=${DEV_MODE}
-
-ENV LANG=C.UTF-8 \
```
mistercrunch, line 108: LANG/ LC_ALL are not needed anymore as of 3.9 I believe. Other ENVs we don't need in this layer. Turns out ENV don't carry through a layer with a different base and led to some confusion on my part. As it was ENV would not carry through to the final layers, but they do when located in python-common

Dockerfile
```diff
@@ -125,27 +117,14 @@ RUN pip install --no-cache-dir --upgrade uv
 RUN uv venv /app/.venv
 ENV PATH="/app/.venv/bin:${PATH}"
 
-# Install Playwright and optionally setup headless browsers
-ARG INCLUDE_CHROMIUM="true"
-ARG INCLUDE_FIREFOX="false"
-RUN --mount=type=cache,target=/root/.cache/uv\
```
mistercrunch, line 131: Moving this section to the python-common layer that's used for the app itself as the python/backend translation builder layer does not need this

Dockerfile
```diff
@@ -159,13 +138,20 @@ RUN if [ "$BUILD_TRANSLATIONS" = "true" ]; then \
 # Python APP common layer
 ######################################################################
 FROM python-base AS python-common
+
+ENV SUPERSET_HOME="/app/superset_home" \
```
mistercrunch, line 142: these now carry through to the 3 final images

Dockerfile
```diff
@@ -238,12 +236,13 @@ RUN /app/docker/apt-install.sh \
...
 # Install Python dependencies using docker/pip-install.sh
-RUN --mount=type=cache,target=/root/.cache/uv \
+RUN --mount=type=cache,target=${SUPERSET_HOME}/.cache/uv \
     /app/docker/pip-install.sh --requires-build-essential -r requirements/development.txt
 # Install the superset package
-RUN --mount=type=cache,target=/root/.cache/uv \
+RUN --mount=type=cache,target=${SUPERSET_HOME}/.cache/uv \
     uv pip install .
 
+RUN uv pip install .[postgres]
```
mistercrunch, line 245: going back to installing postgres client as part of the build process for the `dev` and `ci` layer (still not packaging the client on `lean`)

docker/docker-bootstrap.sh
```diff
@@ -18,9 +18,12 @@
 
 set -eo pipefail
 
+# Ensure the temporary directory exists
+mkdir -p /tmp/superset_temp
+
 # Make python interactive
 if [ "$DEV_MODE" == "true" ]; then
-    if command -v uv > /dev/null 2>&1; then
+    if [ "$(whoami)" = "root" ] && command -v uv > /dev/null 2>&1; then
```
mistercrunch, line 26: only doing this when `root` as `superset` doesn't have the rights to touch the python packaging.

docker/docker-bootstrap.sh
```diff
@@ -34,7 +37,8 @@ if [ "$CYPRESS_CONFIG" == "true" ]; then
     export SUPERSET__SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://superset:superset@db:5432/superset_cypress
     PORT=8081
 fi
-if [[ "$DATABASE_DIALECT" == postgres* ]] ; then
+if [[ "$DATABASE_DIALECT" == postgres* ]] && [ "$(whoami)" = "root" ]; then
```
mistercrunch, line 40: We need to keep this for the use case of other docker-compose files that mount the docker/ folder

mistercrunch:
OMG a working ephemeral env! ^^^^

mistercrunch:
back

dpgaspar (approved):
Nice, fingers crossed, we need to create a new label named testenv-up

drummerwolli:
@mistercrunch can you check [this issue](https://github.com/apache/superset/issues/31890)? I have problems running the `docker compose up --build` command from the [development instructions](https://superset.apache.org/docs/contributing/development#docker-compose-recommended) on a fresh clean machine, and if i try without the changes from this PR (the commit just before), it works. I'm not too deep into Docker to understand exactly what's going on, but assuming you changed some caching here, it sounds very suspicious that this could be related. thanks for checking!

cc @sfirke who helped on the investigation as well.

mistercrunch:
Can you confirm you have issues when using `localhost:9000` if so I can't recreate here. Also noting that there may be some false-negative log messages around the `/ws` endpoint. I haven't had the time to look into those, but should get to the bottom of it/tackled them.

PR #32030, opened 2025-01-30
chore(timeseries charts): adjust legend width by padding

eschutho:
### SUMMARY

Small adjustment to add an ellipsis and width to the legend based on the margin so that the legend does not overlap the chart when on the sides. Currently you have to manually adjust the margin to make sure that it all fits. There are still some cases in very small margins where there is still some overlap, and manual adjustment to the margins is necessary.

Before:
Screenshot 2025-01-31 at 3 04 01 PM
After:

https://github.com/user-attachments/assets/02e592c4-1fcd-4e33-a463-8420a32351e4

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/utils/series.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/utils/series.ts
@@ -33,7 +33,7 @@ import {
   TimeFormatter,
   ValueFormatter,
 } from '@superset-ui/core';
-import { SortSeriesType } from '@superset-ui/chart-controls';
+import { SortSeriesType, LegendPaddingType } from '@superset-ui/chart-controls';
 import { format } from 'echarts/core';
 import type { LegendComponentOption } from 'echarts/components';
 import type { SeriesOption } from 'echarts';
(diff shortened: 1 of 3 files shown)


geido (approved):
Looks good. Curious about that big margin when the legend is on top or the bottom? Is it intended?

superset-frontend/plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -443,13 +444,30 @@ export function getLegendProps(
       borderColor: theme.colors.grayscale.base,
     },
   };
+  const MIN_LEGEND_WIDTH = 0;
+  const MARGIN_GUTTER = 45;
+  const getLegendWidth = (paddingWidth: number) =>
+    Math.max(paddingWidth - MARGIN_GUTTER, MIN_LEGEND_WIDTH);
```
kgabryje, line 450: I kinda agree with bot's comment above - maybe we should set a minimal width value where the legend is still readable?

eschutho: Yeah, I tried out a few different min widths, but the problem is that when applying a min width, then we have the same overlap issue with the chart. So either way, there will be a UI issue, and a larger padding will need to be set. Someone needs to explicitly set a 0 padding in order for this to apply. If you think having the overlap is better than a readable legend, we can def do that.

eschutho:
> Looks good. Curious about that big margin when the legend is on top or the bottom? Is it intended?

Oh, good eye. That was just because I had set the margin to 200 in the video. This is the margin top when not entered:
Screenshot 2025-02-03 at 2 06 44 PM

PR #32035, opened 2025-01-30
fix(fe/dashboard-list): display modifier info for `Last modified` data

### LINKED ISSUE #31258: Hovering over a "last modified" date on the dashboard list does not show who modified it
### Bug description

This is a regression from 4.0.2 to 4.1.1. It only affects the Dashboard list, not Datasets or Charts lists -- those retain this functionality and did not regress.

[This test ](https://github.com/apache/superset/blob/339d491dfc81d28d0caf175f215e58c1383066f0/superset-frontend/src/components/AuditInfo/ModifiedInfo.test.tsx#L4) failed to catch the regression - I suspect the username was not available, but instead of that being a failure, it was a pass because the tooltip was not present.

I tried to hunt it down to fix it but couldn't.  Maybe caused by https://github.com/apache/superset/pull/28609 and the fact that the `changed by` comes from superset-frontend/src/components/AuditInfo/index.tsx ? Just one guess.

### Screenshots/recordings

**Before - 4.0.2**

image

**Current - 4.1.1**
image

### Superset version

4.1.0

### Python version

I don't know

### Node version

I don't know

### Browser

Firefox

### Additional context

_No response_

### Checklist

hainenber:
### SUMMARY

fix(fe/dashboard-list): display modifier info for `Last modified` data

Fix #31258

The tooltip showing modifier data on `Last modified` cell is now restored.

Before

https://github.com/user-attachments/assets/0dbb18fc-e6f9-4426-93ef-6bcd672de797

After

https://github.com/user-attachments/assets/ad6b77ce-9506-4c82-b0a9-941aefe588be

### ADDITIONAL INFORMATION

- [x] Has associated issue: 31258
- [ ] Required feature flags:
- [ ] Changes UI


sadpandajoe (approved):

sfirke:
Thanks for this fix! Did you see that in the issue, I identified what I think is an end-to-end test that failed to catch this regression? If my assessment is correct - I'm new to that part of the project - it would be ideal if this test got updated to avoid such a regression in the future.

hainenber:
@sfirke the linked test is not an E2E one though, it's a component test. I'd like to add an E2E test to cover this regression but Cypress doesn't _really_ support hovering so it's a real hassle atm. Once we migrated to Playwright, it'll be a breeze instead. Hope this helps!

sfirke:
Got it @hainenber - thanks for explaining that and thanks for the fix!

superset-frontend/src/pages/DashboardList/index.tsx
```diff
@@ -118,12 +118,13 @@ const DASHBOARD_COLUMNS_TO_FETCH = [
...
   'slug',
-  'changed_by',
+  'changed_by.id',
+  'changed_by.first_name',
+  'changed_by.last_name',
   'changed_on_delta_humanized',
   'owners.id',
   'owners.first_name',
   'owners.last_name',
-  'owners',
```
michael-s-molina, line 126: @hainenber Can you confirm that the owners' fields, apart from `id`, `first_name` and `last_name`, are not being used?

hainenber: The attributes like `owners.last_name` are the one populating the owners field. For some reasons, you have to specify atributes for nested fields like `owners` and `changed_by`.

michael-s-molina: Maybe you misunderstood what I mean. Previously, we had `owners` being declared which means that ALL attributes of `owners` were accessible like `full_name`. Now that you excluded `owners` from the fetched columns, if there's some component accessing `full_name`, it will break. I'm ok with removing the whole owners object but we need to check which attributes are in fact being used so that the projection is correct.

hainenber: I see. Then in that case, I'll only add the missing attributes and not to remove the nested field names.

hainenber: Re-added old nested fields. Issue is still fixed with less risks :D

image

rusackas (approved):
LGTM! Thank you for addressing the comments/concerns!

PR #32043, opened 2025-01-30
chore: Skip the creation of secondary perms during catalog migrations

Vitor-Avila:
### SUMMARY
The migrations to add catalog permissions can take considerably long, as it would get all catalogs exposed to the credentials, then get all schemas for each of these catalogs and create these permissions accordingly. 

This requires longer migrations that typically run with downtime. 

This PR adds a flag to the `config` that allows to skip these permissions, as these can be created later once the app is up by just editing the DB connection.

No UI changes.

### TESTING INSTRUCTIONS
Unit tests added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -1905,6 +1905,15 @@ class ExtraDynamicQueryFilters(TypedDict, total=False):
 EXTRA_DYNAMIC_QUERY_FILTERS: ExtraDynamicQueryFilters = {}
 
 
+# The migrations that add catalog permissions might take a considerably long time
+# to execute as it has to create permissions to all schemas and catalogs from all
+# other catalogs accessible by the credentials. This flag allows to skip the
+# creation of these secondary perms, and focus only on permissions for the default
+# catalog. These secondary permissions can be created later by editing the DB
+# connection via the UI (without downtime).
+CATALOGS_SIMPLIFIED_MIGRATION: bool = False
+
+
 # -------------------------------------------------------------------
 # *                WARNING:  STOP EDITING  HERE                    *
 # -------------------------------------------------------------------
--- a/superset/migrations/shared/catalogs.py
+++ b/superset/migrations/shared/catalogs.py
@@ -23,6 +23,7 @@ from typing import Any, Type, Union
 
 import sqlalchemy as sa
 from alembic import op
+from flask import current_app
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import Session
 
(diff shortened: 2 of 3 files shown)


dpgaspar (approved):

michael-s-molina:
> This flag allows to skip the
> creation of these secondary perms, and focus only on permissions for the default
> catalog. These secondary permissions can be created later by editing the DB
> connection via the UI (without downtime).

@Vitor-Avila if these migrations are optional, why do they are executed by default in the first place? Isn't better to just remove them of upgrading script than creating a feature flag? If they are not optional, what happens when they are skipped because of the feature flag?

betodealmeida (approved):
Nice!!

Vitor-Avila:
> @Vitor-Avila if these migrations are optional, why are they executed by default in the first place? Isn't better to just remove them from the upgrade script instead of create a new feature flag? If they are not optional, what happens when they are skipped because of the feature flag?

Hey @michael-s-molina,

Before catalogs, to query something from another database in the DB instance/host users would have to do it via a virtual dataset using `select * from <other_catalog>.<schema>.<table>`. These "secondary permissions" are created so that after the migration, limited users need access on these other catalogs in order to query them in SQL Lab. 

For existing datasets and charts, they'll all be associated with the `default_catalog` in their permission setting (even if they're querying other catalog) as we won't parse the query to determine (the same happens today if you create a dataset with `schema_a` selected in the dropdown and use `select * from schema_b.table`).

That said, ideally creating these permissions during the migration is a good thing, but we've seen cases where this process takes too long, and that causes a very long migration/downtime. For example, there used to be a Postgres Cloud service named `ElephantSQL`, and your credentials would be able to list all DBs in the instance (even though you actually don't have access to these other DBs). 

These permissions are going to be added by the time you edit the DB configuration (either to enable the ability to change catalogs, or any other change actually).

I thought about adding it as a config because if you're able to run the migration with the config disabled, then it's great! All perms created. But if you start facing issues, a long downtime, then enabling the setting might be ideal. 

Let me know your thoughts.

betodealmeida:
> @Vitor-Avila if these migrations are optional, why are they executed by default in the first place? Isn't better to just remove them from the upgrade script instead of create a new feature flag? If they are not optional, what happens when they are skipped because of the feature flag?

@michael-s-molina we want to apply these migrations as soon as possible, because without them the data in non-default catalogs is vulnerable — users with access to schema `foo` in the default catalog will be able to access schemas with the same name in the non-default catalogs, which is probably not what the administrator wants.

In most cases the migration is quick and painless, but of course there are people with huge databases where this can take considerable time. So this is a compromise between security and performance.

michael-s-molina:
Got it. Thanks for the additional context @Vitor-Avila @betodealmeida!

michael-s-molina (approved):

PR #32053, opened 2025-01-30
fix: enforce `ALERT_REPORTS_MAX_CUSTOM_SCREENSHOT_WIDTH`

betodealmeida:
### SUMMARY

We only enforce `ALERT_REPORTS_MAX_CUSTOM_SCREENSHOT_WIDTH` when an alert/report is created. This means that if the feature flag is later changed to a smaller value it might not have any effect on pre-existing alerts/reports.

This PR adds a check during screenshot generation.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/report/execute.py
+++ b/superset/commands/report/execute.py
@@ -300,13 +300,16 @@ class BaseReportState:
         )
         user = security_manager.find_user(username)
 
+        max_width = app.config["ALERT_REPORTS_MAX_CUSTOM_SCREENSHOT_WIDTH"]
+
         if self._report_schedule.chart:
             url = self._get_url()
+
             window_width, window_height = app.config["WEBDRIVER_WINDOW"]["slice"]
-            window_size = (
-                self._report_schedule.custom_width or window_width,
-                self._report_schedule.custom_height or window_height,
-            )
+            width = min(max_width, self._report_schedule.custom_width or window_width)
+            height = self._report_schedule.custom_height or window_height
+            window_size = (width, height)
+
             screenshots: list[Union[ChartScreenshot, DashboardScreenshot]] = [
                 ChartScreenshot(
                     url,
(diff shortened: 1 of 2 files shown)


superset/commands/report/execute.py
```diff
@@ -300,13 +300,16 @@ def _get_screenshots(self) -> list[bytes]:
         )
         user = security_manager.find_user(username)
 
+        max_width = app.config["ALERT_REPORTS_MAX_CUSTOM_SCREENSHOT_WIDTH"]
```
michael-s-molina, line 303: @betodealmeida Maybe we should deprecate the `ALERT_REPORTS_MAX_CUSTOM_SCREENSHOT_WIDTH` feature flag to remove it in a major version. It seems that this feature should always be enabled in case alerts and reports are too.

tests/unit_tests/commands/report/execute_test.py
```diff
@@ -365,3 +370,113 @@ def test_get_tab_url(
...
+        ("exceeds_max", 2000, 1600, 800, 1600),
+        # Test when custom width is less than max width
+        ("under_max", 1200, 1600, 800, 1200),
+        # Test when custom width is None (should use window width)
+        ("no_custom", None, 1600, 800, 800),
+        # Test when custom width equals max width
+        ("equals_max", 1600, 1600, 800, 1600),
+    ],
+)
+def test_screenshot_width_calculation(
```
michael-s-molina, line 407: Nice!

michael-s-molina (approved):
Thank you for the improvement!

superset/commands/report/execute.py
betodealmeida: Oh, it's not a feature, just a configuration value.

PR #32137, opened 2025-02-04
fix: copy oauth2 capture to `get_sqla_engine`

betodealmeida:
### SUMMARY

Copied the "OAuth2 needed" logic to `get_sqla_engine`, since for some databases the error happens when calling `create_engine()` (eg, BigQuery) instead of when calling `get_raw_connection`. This way we always catch the error and raise the proper exception.

N/A

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -451,13 +451,20 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
 
             engine_context_manager = config["ENGINE_CONTEXT_MANAGER"]
             with engine_context_manager(self, catalog, schema):
-                yield self._get_sqla_engine(
-                    catalog=catalog,
-                    schema=schema,
-                    nullpool=nullpool,
-                    source=source,
-                    sqlalchemy_uri=sqlalchemy_uri,
-                )
+                try:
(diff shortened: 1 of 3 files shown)


superset/models/core.py
```diff
@@ -583,24 +590,18 @@ def get_raw_connection(
...
-                        catalog=catalog,
-                        schema=schema,
-                    ):
-                        cursor = conn.cursor()
-                        cursor.execute(prequery)
-
-                    yield conn
+            with closing(engine.raw_connection()) as conn:
+                # pre-session queries are used to set the selected schema and, in the
+                # future, the selected catalog
```
Vitor-Avila, line 595: I think it's finally the future and we can remove the `, in the future, the selected catalog` part?

betodealmeida: Ah, good point!

sadpandajoe:
@michael-s-molina it'd be good to get this fix in as part of 5.0

Vitor-Avila (approved):
LGTM! Thanks for adding the test

PR #32154, opened 2025-02-05
fix(sqllab): correct URL format for SQL Lab permalinks

LevisNgigi:
fix(sqllab): correct URL format for SQL Lab permalinks

### SUMMARY
This PR fixes an issue from this [PR](https://github.com/apache/superset/pull/31962) by ensuring that permalinks use the correct /sqllab/p/{key} format instead of ?key={key} while still using url_for. Previously , url_for("SqllabView.root", key=key, _external=True) generated URLs in the query parameter format (?key={key}), which caused issues when sharing permalinks across users.

### TESTING INSTRUCTIONS
1. Log in and save a query in SQL Lab.
2. Copy the query URL using the "Copy Query URL" option.
3. Share the URL with a different user (one with the `sqllab` role and access to the dataset).
4. Have the other user try to load the shared query URL.
5. Observe that the second user with sqllab_role can successfully load the shared query.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/sqllab/permalink/api.py
```diff
@@ -88,6 +88,8 @@
             state = self.add_model_schema.load(request.json)
             key = CreateSqlLabPermalinkCommand(state=state).run()
             url = url_for("SqllabView.root", key=key, _external=True)
+            if "/sqllab/?key=" in url:
+                url = url.replace("/sqllab/?key=", "/sqllab/p/")
```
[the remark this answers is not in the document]
LevisNgigi: The pattern occurs only once in the url.

superset/sqllab/permalink/api.py
```diff
@@ -88,6 +88,8 @@ def post(self) -> Response:
...
```
betodealmeida, line 92: Oops, my bad. Looking at the view:

```python
    @expose("/", methods=["GET", "POST"])
    @expose("/p/<string:permalink>/", methods=["GET"])
    @has_access
    @permission_name("read")
    @event_logger.log_this
    def root(self, **kwargs: Any) -> FlaskResponse:
        payload = {}
        if form_data := request.form.get("form_data"):
            with contextlib.suppress(json.JSONDecodeError):
                payload["requested_query"] = json.loads(form_data)
        return self.render_app_template(payload)
```

Looks like this is what we want:

```suggestion
            url = url_for("SqllabView.root", permalink=key, _external=True)
```

In general, we should always build internal URLs using `url_for`.

betodealmeida (dismissed):
Thanks for fixing this! I suggested a more robust solution.

superset/sqllab/permalink/api.py
LevisNgigi: So after testing this locally it still will not work.It displays the same query despite copying different queries.

betodealmeida: Can you try this:

```python
    @expose("/", methods=["GET", "POST"], endpoint="root")
    @expose("/p/<string:permalink>/", methods=["GET"], endpoint="root_p")
    @has_access
    @permission_name("read")
    @event_logger.log_this
    def root(self, **kwargs: Any) -> FlaskResponse:
        payload = {}
        if form_data := request.form.get("form_data"):
            with contextlib.suppress(json.JSONDecodeError):
                payload["requested_query"] = json.loads(form_data)
        return self.render_app_template(payload)
```

And then do:

```python
url = url_for("SqllabView.root_p", permalink=key, _external=True)
```

LevisNgigi: Updated the pr and separated permalink_view from root and works well.

justinpark (approved):
LGTM! Thanks for hotfix

PR #32170, opened 2025-02-06
feat(charts): add two new boxplot parameter sets

sfirke:
### SUMMARY
I have a user who needs 90th percentile values for a reporting requirement. After discussion with committers I added two common options, `10/90` and `5/95`.

### AFTER
image

### TESTING INSTRUCTIONS
I spun up Superset with this addition and confirmed that the min/max box values change appropriately.


sfirke:
Sorry I missed the window for 5.0.0 - if there's an RC2 I would love for this to get cherried in.

villebro (approved):
LGTM

superset-frontend/plugins/plugin-chart-echarts/src/BoxPlot/types.ts
```diff
@@ -36,7 +36,9 @@ export type BoxPlotFormDataWhiskerOptions =
   | 'Tukey'
   | 'Min/max (no outliers)'
   | '2/98 percentiles'
-  | '9/91 percentiles';
+  | '5/95 percentiles'
+  | '9/91 percentiles'
+  | '10/90 percentiles';
```
giftig, line 41: Silly question but these are just descriptive strings; what's actually applying the percentiles for these new options?

sfirke: I think those strings get pulled out here: https://github.com/apache/superset/blob/master/superset/utils/pandas_postprocessing/boxplot.py#L93

sfirke:
I think that CI failure is just a flaky test? If someone confirms, we can merge.

superset-frontend/plugins/plugin-chart-echarts/src/BoxPlot/types.ts
villebro: The percentiles are regexed out of the string, so as long as long as the string is compliant it'll work: https://github.com/apache/superset/blob/acf91e1f609f2106d019aa40fcf66a7c147e36d0/superset-frontend/packages/superset-ui-chart-controls/src/operators/boxplotOperator.ts#L28

giftig (approved):

superset-frontend/plugins/plugin-chart-echarts/src/BoxPlot/types.ts
giftig: Seems like ab it of a flaky way to handle it but it's not new code so fair enough :+1:

rusackas:
The test failure sure does look unrelated, but I re-ran it and got the same result. I'm not (yet, anyway) sure what to make of it.

sfirke:
Tried to rebase to address a flaky CI failure and spammed everyone, sorry

villebro (dismissed):
Oops, please fix the bad rebase

sfirke:
@villebro I fixed the rebase, ugh.

The same test is still failing, like @rusackas I don't see how it's related and yet it keeps failing on this PR.

sfirke:
@villebro I fixed the rebase and the flaky test finally passed, I believe this is ready for merge but needs your approval

Antonio-RiveroMartnez (approved):
lgtm

PR #31634, opened 2024-12-27
chore(fe): migrate 4 Enzyme-based tests to RTL

hainenber:
chore(fe): migrate 4 Enzyme-based tests to RTL

### SUMMARY

Part of ongoing effort to refactor away Enzyme-based unit test to RTL

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

N/A

### TESTING INSTRUCTIONS

The CI checks should all pass for this particular PR

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #32191, opened 2025-02-07
chore: remove all enzyme tests and references

mistercrunch:
[Enzyme is dead, now what?](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl)

We're deep in RTL already, and the enzyme tests are doing more harm than good to us as maintainers.

This has implication on lowering test coverage, but at the same time many of the enzyme tests are against fairly settled, untouched components. I suggest that we delete all the Enzyme tests and agree that touching now-untested component requires writing RTL test to cover the components you touch.

Overall a good step forward that allows us to drop a large chunk of complexity and branches in our dep tree.

-----
NOTE: overall there are more enzyme tests than I thought orignally as grepping for `enzyme` missed out on some reusable shims/constructs that indirectly used enzyme. Putting out as DRAFT so we can see what we want to do with this.


villebro:
Nice! 🔥  Next let's go after Cypress!

image

mistercrunch:
afraid to look at coverage before/after... whatever happened to covecov?!

villebro:
> afraid to look at coverage before/after...

I wouldn't worry about it, after this we have a much better understanding of high quality code coverage.

hainenber:
There are 4 Enzyme tests I have converted to RTL in this [PR](https://github.com/apache/superset/pull/31634). Might be a tad better than to delete them away? :D

mistercrunch:
I think @rusackas has this one too https://github.com/apache/superset/pull/32207, so I'm going to close this PR and slowly back off.

If anything we learned a few things in this PR
1. there are about 50 enzyme-backed test files around (not only a dozen like I thought originally)
1. other people seem interested to convert the tests to RTL

I think overall we think that more test coverage is better, even if the tests aren't fully redesigned for RTL. Anyhow, closing this PR and passing the baton here...

I anything I'd advocate to be radical around approaching the migration. If some tests are hard to convert, it's acceptable to delete chunks or the file altogether.

michael-s-molina:
@mistercrunch I think the difference between Enzyme tests and Cypress tests are:
- There are active efforts regarding converting Enzyme to RTL.
- The level of flakiness during CI runs is very different.
- We capture more regressions with Enzyme tests than Cypress tests.

The radical approach might be more appropriate for Cypress tests.

PR #32195, opened 2025-02-09
fix: handlebars html and css templates reset on dataset update

### LINKED ISSUE #32049: When updating the dataset associated with an existing Handlebar chart, all the previous HTML and CSS gets lost
## Description
When you have an existing Handlebar chart with personalized HTML and CSS code, if you update the dataset that powers the chart, all the previous HTML and CSS get removed.

## How to reproduce the bug
1. Create a virtual dataset with something simple like `select * from YourTable`
2. Create a handlebar chart with that dataset and add some HTML and CSS code in the customize section
3. From the Chart section, click on the 3 dots on the left and “edit dataset” 
4. Change the query by adding a where clause or any other change you want.
5. Save the dataset and click on “Update Chart”

## Expected results
Your dataset gets updated and you have the same previous HTML and CSS applied to your handlebars chart

## Actual results
Your dataset gets updated but all the previous HTML and CSS code you had gets lost. It also seems that the textarea "blinks" going back and forth between states.

### Screenshots/recordings

Image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

DamianPendrak:
Fixes #32049 

### SUMMARY

There is sharedControls.entity included in the config of the handlebarsTemplateControlSetItem and styleControlSetItem. The "valueKey" property from this sharedConfig is used in the exploreReducer in UPDATE_FORM_DATA_BY_DATASOURCE action to reset the control values that are related to a column. If I understand it correctly, the handlebars templates are not column related drag and drop controls. There might be other reasons (based on the comment in dndControls.tsx:45) to have the sharedControls.entity config included, so I just unset the "valueKey" property to exclude the handlebars templates control values from the reset on datasource update.

"It also seems that the textarea "blinks" going back and forth between states."
After the fix I do not experience the "blinks" anymore. Let me know if it is still an issue.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #32049 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-handlebars/src/plugin/controls/handlebarTemplate.tsx
+++ b/superset-frontend/plugins/plugin-chart-handlebars/src/plugin/controls/handlebarTemplate.tsx
@@ -65,6 +65,7 @@ export const handlebarsTemplateControlSetItem: ControlSetItem = {
 </ul>`,
     isInt: false,
     renderTrigger: true,
+    valueKey: null,
 
     validators: [validateNonEmpty],
     mapStateToProps: ({ controls }) => ({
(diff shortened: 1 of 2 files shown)


kgabryje (approved):
LGTM, thanks for the contribution!

PR #32211, opened 2025-02-10
fix: hydrate datasetsStatus

betodealmeida:
### SUMMARY

When the dashboard is hydrated we're not using the `datasetsStatus` from the redux state; instead we're setting it to `loading`, which prevents the OAuth2 link to show up.

Before:

Screenshot 2025-02-10 at 16-29-51 A dash

It's always spinning because the state is `loading`.

After:

Screenshot 2025-02-10 at 16-28-21 A dash

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila (approved):
LGTM! Unsure if we could cover this in a test scenario, but also think we could cover that in a follow up if needed (and possible).

geido (approved):

PR #32212, opened 2025-02-10
fix: set `Rich tooltip` -> 'Show percentage' to false by default

mistercrunch:
Many users have reported that they prefer having the show percentage option in timeseries as false by default. While it's generally a good feature and many might want it, default should be lightweight and optimize for less clutter.

Open to input or debatting which default option is best, but given that many users reported it, and that it's a bit of a visual regression / change in some dashboard, I thought we'd switch it off.

### after 
Screenshot 2025-02-10 at 1 46 42 PM

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/controls.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/controls.tsx
@@ -230,7 +230,7 @@ const tooltipPercentageControl: ControlSetItem = {
     type: 'CheckboxControl',
     label: t('Show percentage'),
     renderTrigger: true,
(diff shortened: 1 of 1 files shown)


eschutho (approved):

mistercrunch:
@michael-s-molina checking in to make sure you agree on switching this default. Could add in a feature flag `TIMESERIES_SHOW_PERCENTAGE_DEFAULT`, but feeling like we have enough feature flags without adding this new one.

michael-s-molina:
Makes sense @mistercrunch!

michael-s-molina (approved):

PR #32218, opened 2025-02-11
fix(Datasource): handle undefined datasource_type in fetchSyncedColumns

tahvane1:
### SUMMARY
Sync columns from source is not working on dataset edit columns page. This fixes that.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Datasource/utils.js
+++ b/superset-frontend/src/components/Datasource/utils.js
@@ -121,7 +121,7 @@ export function updateColumns(prevCols, newCols, addSuccessToast) {
 
 export async function fetchSyncedColumns(datasource) {
   const params = {
-    datasource_type: datasource.type,
+    datasource_type: datasource.type || datasource.datasource_type,
     database_name:
       datasource.database?.database_name || datasource.database?.name,
     catalog_name: datasource.catalog,


superset-frontend/src/components/Datasource/utils.js
```diff
@@ -121,7 +121,7 @@ export function updateColumns(prevCols, newCols, addSuccessToast) {
 
 export async function fetchSyncedColumns(datasource) {
   const params = {
-    datasource_type: datasource.type,
+    datasource_type: datasource.type || datasource.datasource_type,
```
michael-s-molina, line 124: @fisjac Are these fields the same thing and we should consolidate them? I'm asking because of https://github.com/apache/superset/pull/30903.

sadpandajoe: adding @kgabryje as he might have knowledge on this too.

sadpandajoe: @michael-s-molina I wonder if this is similar to the line below where we have `database.database_name` or `database.name`. Maybe the API is returning `datasource_type` vs `type` which could be whats causing an issue. Also I do agree with Korbit that we should probably add null check and we should probably add tests for `fetchSyncedColumns`

eschutho: Hi, yes the dataset object from the chart api returns `datasource.type` and the dataset api returns `datasource_type`. There is another example of it [here](https://github.com/apache/superset/blob/master/superset-frontend/src/components/Datasource/DatasourceEditor.jsx#L615).

eschutho: It's probably outside of the scope of this ticket to fix it, but might be good to look at for a future major release?

sadpandajoe (approved):
Tested this out and it's working now.

PR #32239, opened 2025-02-12
docs: adding notes about using uv instead of raw pip

mistercrunch:
Since the last release, we moved to using `uv` instead of `pip` and related packages. This can break deployments using our docker files that rely on `pip` for say installing drivers or custom libraries. Adding a note in UPDATING.md. This PR should be cherried in 5.0 branch


sadpandajoe:
@michael-s-molina we should probably cherry this since there are a few changes which might come back as the build being broken.

sadpandajoe (approved):

Issue #27379, opened 2024-03-04
Add Romania to countries available in choropleth


omniconvert:
## Description

Romania as a country is missing from the choropleth maps.

omniconvert:
Not sure if bug or feature request at this point :)

PR #32261, opened 2025-02-14
feat(type-checking): Add type-checking pre-commit hooks

alveifbklsiu259:
feat(type-checking): Add type-checking pre-commit hooks

### SUMMARY

This PR introduces some pre-commit hooks to check TypeScript files using `tscw-config`. This enhancement aims to streamline the development process by running type checks **only on changed files**.

#### NOTES

- Since different packages (e.g., `superset-frontend`, `superset-websocket`...) use their own `tsconfig.json` and may require different declaration files, ~~multiple type-checking hooks are used~~. (edit: after discussion, only one hook (`type-check-frontend`) is used, if future requirement changes, one can always reference [commits](https://github.com/apache/superset/pull/32261/commits)).
- JS/JSX files will be processed by the type-checking hooks, but `checkJs` is not enabled by default. To enable type-checking for JS/JSX files, either include `//@ts-check` at the top of each JS/JSX file or enable it in the package's `tsconfig.json`.
- Each package, such as `superset-frontend`, `superset-websocket`... will include all files ending with `.d.ts` in their respective directories (recursively). However, if one needs to include third-party declaration files, one can either import them directly in the files that need them or declare them in the `type` field of the package's `tsconfig.json`. For example `import "@testing-library/jest-dom"`
- Under the hood, what this hook does it create a `tsconfig.json` from the current package (`superset-frontend/tsconfig.json` for `superset-frontend`), remove the `include` filed, and include all the files passed to this hook (and some declaration files) in the `files` field. This means that ANY files passed to the hook will be type-cheked even if they are not included in the original `tsconfig.json`'s `include` field. For example, at the time of writing, `superset-frontend/tsconfig.json` does not include files from `./*/storybook/**/*`, but if one modifies `superset-frontend/packages/superset-ui-demo/storybook/stories/superset-ui-color/RenderPalettes.tsx`, this file will still be type-checked. IMHO type-checking the files even though they will not be compiled to javascript is not a bad idea, but if one really wants to disable this behavior, one can:
   - Use `// @ts-nocheck`.
   - Prevent those files from being passed to the hook, for example, use `exclude: ^(superset-frontend/cypress-base/|superset-frontend/.*/storybook/)` in `pre-commit-config.yaml`.

### DISCLOSURE

I am the author of `tscw-config`. I genuinely believe that this library offers an effective solution to the issue discussed in [Issue #27379](https://github.com/microsoft/TypeScript/issues/27379#issuecomment-2565940933).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sadpandajoe:
@mistercrunch don't we already have some sort of type checking? Adding you since you did a lot of work to clean up our pre-commits

.pre-commit-config.yaml
```diff
@@ -68,6 +68,39 @@ repos:
         language: system
         pass_filenames: true
         files: \.(js|jsx|ts|tsx)$
+  - repo: local
```
mistercrunch, line 71: wondering if you could get away with a single entry here (?)

mistercrunch:
@sadpandajoe yes we do in CI, but this add the checks to `pre-commit` so issues can be caught at commit time, which is better. This adds in a fair amount of complexity to deliver on it. 

Wondering if it would make sense to only do it on superset-frontend/**/js-and-ts-files to keep pre-commit.yml a bit simpler.

@alveifbklsiu259  is it all super fast?

.pre-commit-config.yaml
```diff
@@ -68,6 +68,39 @@ repos:
...
+    hooks:
+      - id: type-checking-frontend
+        name: Type-Checking (Frontend)
+        entry: ./scripts/check-type.js
```
mistercrunch, line 75: From my understanding this will check all the files, not just the ones touched in the commit is that right?  Same for all other entries. 

If so it seems this will greatly slow down commit speed.

You can see how I did it for eslint, but by using `$@`, I'm able to pass the list of files to check to the command, so it will only check the files touched in that specific commit.

mistercrunch: know that `npm run type` take 10-30 seconds, we don't want to add this kind of lag for each commit, especially when you're not even touching these files. 

Key here is to somehow pass the list of files touched in the commit (provided by pre-commit), and only run the checks on those files. Now guessing `tsc` may have to go and import all the related files referenced in that seed file, and could end up taking quite some time if you end up covering files-that-import-files-that-import-other-files, ...

alveifbklsiu259: > From my understanding this will check all the files, not just the ones touched in the commit is that right?

It seems that we don't have to explicitly pass `$@`, pre-commit will only passed the touched files:

```js
// check-type.js
void (async () => {
  const args = process.argv.slice(2);
  console.log(args)
  exit(1)
```


> is it all super fast?

I wouldn't say it is **super fast**, but it is faster than checking all files.




> Key here is to somehow pass the list of files touched in the commit (provided by pre-commit), and only run the checks on those files.

This is exactly what this pre-commit hook does.

> ... may have to go and import all the related files referenced in that seed file, and could end up taking quite some time if you end up covering files-that-import-files-that-import-other-files

This is still true for this hook. For example:

```ts
// superset-frontend/src/components/AsyncAceEditor/index.tsx
import AsyncEsmComponent, {
  PlaceholderProps,
} from 'src/components/AsyncEsmComponent';
```

If `superset-frontend/src/components/AsyncAceEditor/index.tsx` is the only file in the staging area, running this hook will also check `src/components/AsyncEsmComponent/index.tsx` and the files it imports and so on.

> Wondering if it would make sense to only do it on superset-frontend/**/js-and-ts-files to keep pre-commit.yml a bit simpler.

Sure, as long as we only want to type-check the files under `superset-frontend`, but we still need to exclude `superset-frontend/cypress-base/`, since `superset-frontend/cypress-base/` uses a different `tsconfig.json` and it contains a conflicting declaration file. i.e. `superset-frontend/cypress-base/cypress/support/index.d.ts`

scripts/check-type.js
```diff
@@ -0,0 +1,180 @@
...
+    remainingArgs,
+  } = extractArgs(args, [PACKAGE_ARG_REGEX, EXCLUDE_DECLARATION_DIR_REGEX]);
+
+  if (!packageArg) {
+    console.error("package is not specified");
+    exit(1);
+  }
+
+  const packageRootDir = getPackage(packageArg);
+  const packagePathRegex = new RegExp(`^${packageRootDir}\/`);
```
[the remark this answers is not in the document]
alveifbklsiu259: [fixed](https://github.com/apache/superset/pull/32261/commits/503378e5bf4e8b0ae54044e5699588756041a7d1).

.pre-commit-config.yaml
mistercrunch: Oh nice. Great to hear it's all working. And I understand now that we need one entry per tsconfig.

scripts/check-type.js
```diff
@@ -0,0 +1,244 @@
...
+      if (file.isDirectory() && !shouldExclude) {
+        recursivePromises.push(
+          getFilesRecursively(fullPath, regex, excludedDirs)
+        );
+      } else if (regex.test(file.name)) {
+        result.push(fullPath);
+      }
+    }
+
+    const recursiveResults = await Promise.all(recursivePromises);
```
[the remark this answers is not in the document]
alveifbklsiu259: @korbit-ai

After analyzing the memory impact, I found that the increase is not significant enough to warrant the added complexity of introducing a semaphore or concurrency limit. Implementing such a solution would complicate the codebase.

I believe the current implementation is sufficient for our use case and does not pose any immediate performance concerns. Maybe we will consider revisiting this approach if we encounter any memory-related issues in the future.

```ts
  const initial = process.memoryUsage();

  let declarationFiles = await getFilesRecursively(
    packageRootDir,
    DECLARATION_FILE_REGEX,
    excludedDeclarationDirs
  );
  const final = process.memoryUsage();

  const memoryDiff = {
    rss: `${Math.round((final.rss - initial.rss) / (1024 * 1024))} MB`,
    heapTotal: `${Math.round(
      (final.heapTotal - initial.heapTotal) / (1024 * 1024)
    )} MB`,
    heapUsed: `${Math.round(
      (final.heapUsed - initial.heapUsed) / (1024 * 1024)
    )} MB`,
    external: `${Math.round(
      (final.external - initial.external) / (1024 * 1024)
    )} MB`,
  };

  console.log("Memory Difference:", memoryDiff); 
  // Memory Difference: { rss: '16 MB', heapTotal: '9 MB', heapUsed: '3 MB', external: '0 MB' }
```

scripts/check-type.js
```diff
@@ -0,0 +1,220 @@
...
+    packageRootDir,
+    DECLARATION_FILE_REGEX,
+    excludedDeclarationDirs
+  );
+  declarationFiles = removePackageSegment(declarationFiles, packageRootDir);
+  const declarationFilesStr = declarationFiles.join(" ");
+
+  const packageRootDirAbsolute = join(SUPERSET_ROOT, packageRootDir);
+  const tsConfig = getTsConfig(packageRootDirAbsolute);
+  const command = `--noEmit --allowJs --composite false --project ${tsConfig} ${argsStr} ${declarationFilesStr}`;
```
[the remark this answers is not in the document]
mistercrunch: Agreed, not an issue here

mistercrunch (approved):
Thanks for doing this! This will certainly save extra CI cycles by catching type errors at commit time.

rusackas (approved):
Looks safe from a repo secrets perspective. Thanks for the hard work here, and looking forward to the performance bump!

Issue #17852, opened 2021-12-22
[SIP-78] Proposal for Advanced Types


cccs-RyanS:
# [SIP-78] Proposal for Advanced Types

## Motivation

Our users are leveraging Superset to explore and visualize network security logs. These logs often contain IP, CIDR range, port, hostname, and/or protocol fields, among others. While such fields might be stored as simple INTEGERS or VARCHARS in the backend database, these fields also have a “advanced” type which, if leveraged, can provide additional context and improve the user experience.

**User-Friendly Display**

A field’s advanced type can influence how that field is interacted with by the users. Consider, for example, an IP field. For performance reasons, we store IP values as INTEGERs in our DW e.g. the IP “127.0.0.1” would be stored as its integer value 2130706433. However, the string representation is substantially more human-readable and, as such, we wish to display IP fields in dot-decimal notation in visualizations and filters. This numeric format is fairly standard (link)and having this display functionality would be valuable in superset in case the underlying DB does not support this conversion functionality out of the box. As another example, a Port field stored as an integer (e.g., 443) could be displayed as a protocol name where such mapping is known (e.g., HTTPS).

**Applicable Operators for filters**

Certain operators (e.g., `LIKE`, `=`, `<`, `IS NULL`, `IN`, etc.) may be valid for a database type, but may not be applicable to a advanced type. For example, an organization might decide that the `LIKE` operator is not relevant to Port fields. Ideally, we want to be able to configure a list of applicable operators that will appear for a field of a certain advanced type.

**Input Validation**

Advanced type can be associated with (configurable) input validation and type conversion rules. For example, an input validator for a Port field could validate that the filter string is a parsable number between 0 and 65536, or that it contains known port names such as HTTPS, SSH, FTP.

**Human Readable input values**

As with displaying the human readable format of a value to the user (i.e the dot notation of an IP) users should be able to enter the more readable form of a value and have it be applied properly to a query. This would mean a user could create a filter such as `{col: "ip_col", op: "IN", val: [“1.1.1.1”, “1.1.1.2”]}` and have the query be executed as `ip_col IN [16842752, 16842753]`. The functionality for a user to simply enter `{col: "ip_col", op: "IN", val: [16842752, 16842753]}` should also be preserved.

**Behavior specified in the config**

The code to specify the behavior of each advanced type should be defined outside of the superset codebase (i.e the config). This way those who wish to implement a advanced type may do so freely without having to make any code changes to superset.

**Proof of concept**

[...]

PR #32264, opened 2025-02-14
feat(number-format): adds memory data transfer rates in binary and decimal format

tshallenberger:
### SUMMARY

At Yahoo we needed custom number formatters for data transfer rates, in either IEC or Standard format (binary or decimal). This change adds the following NumberFormats:

```
MEMORY_TRANSFER_RATE_DECIMAL
MEMORY_TRANSFER_RATE_BINARY
```

Where decimal records rates in terms of gigabyte, terabyte, and petabytes, and binary records rates in terms of gibibytes, tebibytes, and pebibytes. The number formatter expects absolute values in bytes. 

image

### TESTING INSTRUCTIONS

1. Create a Big Number chart for any dataset
2. Add a metric with custom sql for the value: `123456789`
3. Nav to Customize > Chart Options > Number Format and select either Bytes (SI), Bytes (IEC), Byterate (SI), Byterate (IEC)
4. You should see either of the following for your selection:

MEMORY_TRANSFER_RATE_DECIMAL: 123.46 MB/s
MEMORY_TRANSFER_RATE_BINARY (IEC): 117.74 MiB/s

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


superset-frontend/packages/superset-ui-core/src/number-format/NumberFormats.ts
```diff
@@ -54,6 +54,13 @@ const SMART_NUMBER = 'SMART_NUMBER';
 const SMART_NUMBER_SIGNED = 'SMART_NUMBER_SIGNED';
 const OVER_MAX_HIDDEN = 'OVER_MAX_HIDDEN';
 
+const BYTES_SI = 'BYTES_SI';
+const BYTES_IEC = 'BYTES_IEC';
+const BITRATE_SI = 'BITRATE_SI';
+const BYTERATE_SI = 'BYTERATE_SI';
+const BITRATE_IEC = 'BITRATE_IEC';
+const BYTERATE_IEC = 'BYTERATE_IEC';
```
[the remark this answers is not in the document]
tshallenberger: False flag I think?

superset-frontend/packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
```diff
@@ -60,6 +60,10 @@ export const D3_FORMAT_OPTIONS: [string, string][] = [
   ['DURATION_COL', t('Duration in ms (10500 => 0:10.5)')],
   ['MEMORY_DECIMAL', t('Memory in bytes - decimal (1024B => 1.024kB)')],
   ['MEMORY_BINARY', t('Memory in bytes - binary (1024B => 1KiB)')],
+  [NumberFormats.BYTES_SI, 'Bytes in SI (kB, MB)'],
+  [NumberFormats.BYTES_IEC, 'Bytes in IEC (kiB, MiB)'],
+  [NumberFormats.BYTERATE_SI, 'Byterate in SI (kB/s, MB/s)'],
+  [NumberFormats.BYTERATE_IEC, 'Byterate in IEC (kiB/s, MiB/s)'],
```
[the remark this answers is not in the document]
tshallenberger: The lowercase k can be synonymous with uppercase K when dealing with kilo / kibi

superset-frontend/packages/superset-ui-core/src/number-format/factories/createNetworkNumberFormatter.ts
```diff
@@ -0,0 +1,156 @@
...
+const bytesIECLabels = [
+  'Bytes',
+  'KiB',
+  'MiB',
+  'GiB',
+  'TiB',
+  'PiB',
+  'EiB',
+  'ZiB',
+];
```
villebro, line 37: Is there overlap here with the code in `createMemoryFormatter.ts`? I think ideally we want to avoid duplication and double down on the same terms. Some quick observations:
- Here we're talking about `Bytes`, while in memory formatter we have `B`
- Here we're only going to `ZiB`/`ZB`, while in memory formatter we're going all the way to `YiB` and for SI to `QB`.

superset-frontend/packages/superset-ui-core/src/number-format/factories/createNetworkNumberFormatter.ts
```diff
@@ -0,0 +1,173 @@
...
+const byterateIECLabels = [
+  'Bytes/s',
+  'KiB/s',
+  'MiB/s',
+  'GiB/s',
+  'TiB/s',
+  'PiB/s',
+  'EiB/s',
+  'ZiB/s',
+];
```
villebro, line 57: To avoid duplication, can we just use the bytes above and then slap on a `/s` after them?

superset-frontend/packages/superset-ui-core/src/number-format/factories/createNetworkNumberFormatter.ts
```diff
@@ -0,0 +1,173 @@
...
+      formatterCache.set(decimals, d3Format(`.${decimals}s`));
+    }
+    return formatterCache.get(decimals);
+  } catch (error) {
+    throw new Error(
+      `Failed to create SI formatter for ${decimals} decimals: ${error.message}`,
+      { cause: error },
+    );
+  }
+}
```
villebro, line 82: Is this performance optimization needed? Was there performance issues without this?

superset-frontend/packages/superset-ui-core/src/number-format/factories/createNetworkNumberFormatter.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+  if (absoluteValue > 0.000001) {
+    return `${siFormatter(decimals)(value * 1000000)}µ ${labels[0]}`;
+  }
+  return `${float4PointFormatter(value)} ${labels[0]}`;
+}
+
+function formatBytesSI(value: number, decimals: number) {
+  return formatValue(value, bytesSILabels, 1000, decimals);
+}
```
villebro, line 118: There's a fair amount of logic here that we'd want to have unit tests for. Can we add those to get full test coverage?

villebro (commented):
Awesome stuff, I'm sure this will be valuable to many orgs! First pass comments.

rusackas:
I think this approach is great, but I also wonder if you're aware of the Advanced Types feature. It's a bit lesser-known, but this might be an interesting application for it. The original implementation of it was to match different formats of IP addresses to one another (acting as a bit of a translation layer, if you will). Having a "fille size" type might allow you to match data where columns contain different units, and do the appropriate conversions, e.g. 1000KB = 1MiB, or 1024K = 1MB, etc.

Here's the original SIP that explains it better than I'm doing :)
https://github.com/apache/superset/issues/17852

superset-frontend/packages/superset-ui-core/src/number-format/factories/createNetworkNumberFormatter.ts
tshallenberger: Yes! I'm still in the process of figuring out how to run tests from my dev container, once I get that running I'll see about getting coverage on all new branches

superset-frontend/packages/superset-ui-core/src/number-format/factories/createNetworkNumberFormatter.ts
tshallenberger: No, this was to satisfy the korbit-ai review bot 😅 if you don't think it's really needed I can simplify it back to its original form

villebro: Ugh, sorry about our house bot giving you bad advice 🤦 Yes, please roll this back, I didn't quite understand how it would help in this context. So let's keep it simple and add memoization later if it's really needed.

rusackas:
re-running CI 🤞

tshallenberger:
Sorry for the delayed update on the PR: @rusackas we weren't able to get advanced types to work for our particular use case, so I took the comments and revisited the code change. It's still a work in progress but I think the best approach is to simply add an extra config to `createMemoryFormatter` that specifies if it's a data transfer rate or not, and therefore appends the '/s' (since data transfer rates are usually specified in bits/bytes per second, I don't think we need to worry about supporting per minute or per hour). 

so the final `setupFormatters.ts` would look kinda like:

```
.registerValue('MEMORY_DECIMAL', createMemoryFormatter({ binary: false }))
.registerValue('MEMORY_BINARY', createMemoryFormatter({ binary: true }))
// new
.registerValue('MEMORY_TRANSFER_RATE_DECIMAL', createMemoryFormatter({ binary: false, transfer: true }))
.registerValue('MEMORY_TRANSFER_RATE_BINARY', createMemoryFormatter({ binary: true, transfer: true }))
```

This way we heavily re-use the createMemoryFormatter and simplify things a bit. 

@villebro Thoughts?

rusackas:
Thanks for staying on top of this! Re-running CI 🤞

villebro (approved):
Great work, LGTM!

renaudk:
Hey guys, that's awesome feature! But ... we are working with bitrates in bits/sec. So the unit B in uppercase is not appropriated. Is it possible to add b/s equivalents?

PR #32269, opened 2025-02-14
fix: SUPERSET_LOAD_EXAMPLES=no are not respect when using environment variables

thiagodamas:
### SUMMARY

SUPERSET_LOAD_EXAMPLES=no are not respect when using environment variables

You can check by using "docker compose config"

It maybe related to https://github.com/docker/compose/issues/12449

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docker-compose-image-tag.yml
+++ b/docker-compose-image-tag.yml
@@ -87,7 +87,6 @@ services:
     healthcheck:
       disable: true
     environment:
-      SUPERSET_LOAD_EXAMPLES: "${SUPERSET_LOAD_EXAMPLES:-yes}"
       SUPERSET_LOG_LEVEL: "${SUPERSET_LOG_LEVEL:-info}"
 
   superset-worker:
--- a/docker-compose-non-dev.yml
+++ b/docker-compose-non-dev.yml
@@ -94,7 +94,6 @@ services:
     healthcheck:
       disable: true
     environment:
-      SUPERSET_LOAD_EXAMPLES: "${SUPERSET_LOAD_EXAMPLES:-yes}"
       SUPERSET_LOG_LEVEL: "${SUPERSET_LOG_LEVEL:-info}"
 
   superset-worker:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -154,7 +154,6 @@ services:
     volumes: *superset-volumes
     environment:
       CYPRESS_CONFIG: "${CYPRESS_CONFIG:-}"
-      SUPERSET_LOAD_EXAMPLES: "${SUPERSET_LOAD_EXAMPLES:-yes}"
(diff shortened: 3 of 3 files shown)


mistercrunch:
Wait, it works for me, and would work after this change...

Trace:
Screenshot 2025-02-14 at 1 54 13 PM
Screenshot 2025-02-14 at 1 54 04 PM

My env:
Screenshot 2025-02-14 at 1 55 16 PM

mistercrunch:
if it's a docker compose bug, maybe you can force install the same version as the one I use (?)

mistercrunch:
If there's a known issue on some docker-compose versions, we could alter the check-env.py script to disallow certain docker-compose versions.

thiagodamas:
> Wait, it works for me, and would work after this change...
> 
> Trace: how I ran it: Screenshot 2025-02-14 at 1 54 04 PM shows only 3 steps in init instead of 4 (excludes loading the examples) Screenshot 2025-02-14 at 1 54 13 PM
> 
> My env: Screenshot 2025-02-14 at 1 55 16 PM

Its works this way, but it doesn't work when SUPERSET_LOAD_EXAMPLES=no is inside docker/.env or docker/.env-local, I already tried both ways.

mistercrunch:
Oh interesting. I'm not sure what that informs though.

So then the issue is that `env_file:` directives don't play nice with `envrionment:` directives? Is this a bug or a misunderstanding?

thiagodamas:
> Oh interesting. I'm not sure what that informs though.
> 
> So then the issue is that `env_file:` directives don't play nice with `envrionment:` directives? Is this a bug or a misunderstanding?
> 
> Should we move away from `environment` and set all defaults in `.env`, tell people to use `.env-local`? Gets a little messy as these .env files are used across services and things, and have a mix of things people should and shouldn't mess with...

For me it appears to be a bug, conflict order when using, inside docker compose, environment and env_file directives; when using like above (using with export or prefixed the docker command) (https://github.com/apache/superset/pull/32269#issuecomment-2660344469) it works

mistercrunch:
While the .env/.env-local are used globally across services, it feels the `envrionment` directive would be great as per-service overrides. Now i don't think we ever need to diverge env vars from a service to another (?) Maybe that informs we could nuke all `envrionment` directives and instruct people to either use `.env-local` or run their own `export K=V`, assuming this all works as expected (?)

rusackas:
This PR has been on hold for a while and now has conflicts. @mistercrunch - should we close this in favor of a different approach, or would you like the author to update it?

rusackas:
Closing due to inability to reproduce, and lack of follow-ups/rebase to keep it moving forward. Feel free to reopen if/when it's brushed up and we have reliable repro steps!

PR #32284, opened 2025-02-17
fix(roles): Add SqlLabPermalinkRestApi as default sqlab roles.

LevisNgigi:
feat(roles): Add SqlLabPermalinkRestApi as default sqlab roles.

### SUMMARY
This PR ensures that SQL Lab users automatically receive can_read and can_write permissions on SqlLabPermalinkRestApi by default when superset init runs.

Previously, these permissions were missing, requiring manual assignment. Now, they are included in the SupersetSecurityManager, making setup more seamless.

N/A
### TESTING INSTRUCTIONS
1)Run superset init after applying this patch.
2)Go to Security → List Roles → SQL Lab.
3)Confirm that:
a.can_read on SqlLabPermalinkRestApi is assigned.
b.can_write on SqlLabPermalinkRestApi is assigned.
4)Log in as a SQL Lab user and verify that permissions are applied.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila (approved):
LGTM!

michael-s-molina:
If these permissions are required, it would be a good idea to use the `fix` prefix for this PR so it can be cherry-picked into 5.0.

superset/security/manager.py
```diff
@@ -330,6 +330,8 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
         ("menu_access", "SQL Editor"),
         ("menu_access", "Saved Queries"),
         ("menu_access", "Query Search"),
+        ("can_read", "SqlLabPermalinkRestApi"),
+        ("can_write", "SqlLabPermalinkRestApi"),
```
justinpark, line 334: The permalink rest api follows `sqllab` resource ([here](https://github.com/apache/superset/pull/29163/files#diff-c4e8d9dc9eec6a02dbb7e8689350aaa761339f1ef0814ec9cfa0758757f34ebdR40)) so it's unnecessary to add this new role.
Additionally, this is where additional permissions related to view access are defined, and the permalink does not have a menu either.

justinpark (changes requested):
Could you attach the access error without the new role? Otherwise the permalink should be working with the existing (sqllab, can_read/write) roles

LevisNgigi:
> Could you attach the access error without the new role? Otherwise the permalink should be working with the existing (sqllab, can_read/write) roles



> Could you attach the access error without the new role? Otherwise the permalink should be working with the existing (sqllab, can_read/write) roles

@justinpark so without manually adding the permissions via the ui,("can_read", "SqlLabPermalinkRestApi"),,("can_write", "SqlLabPermalinkRestApi"),one gets the error that "there was an error generating the permalink" when copying the query link despite having the sqllab role.

justinpark (approved):
It's interesting that other permalink REST APIs (i.e., ExplorePermalinkRestApi) have been generated without this additional permission declaration. I'll stamp it now to unblock the change.

sadpandajoe:
@michael-s-molina it's a fix to help with https://github.com/apache/superset/pull/31421 which is already in 5.0.

pang-wu:
we tested this in 5.0 but it doesn't work -- we have all permission added to the role. Here's the process to reproduce the problem:
- Click "Copy Query URL" in saved Queries.
- Close all tabs in SQLLab to make sure there is no confusion.
- Past the copied link in the same browser: There is nothing show up.

Any thoughts?

LevisNgigi:
> we tested this in 5.0 but it doesn't work -- we have all permission added to the role. Here's the process to reproduce the problem:
> 
> * Click "Copy Query URL" in saved Queries, we got something like `https://[superset hosting address]/sqllab/?key=xx`
> * Close all tabs in SQLLab to make sure there is no confusion.
> * Past the copied link in the same browser: There is nothing show up.
> 
> Any thoughts?

Hey @pang-wu  are you on the latest version of 5.0?I can see the issue is how your url is being built which was fixed by this [pr](https://github.com/apache/superset/pull/32154). The generated url should look something like this https://[superset hosting address]/sqllab/p/k34OzpRjqWB

pang-wu:
@LevisNgigi I see... I am on the 5.0.0rc1 docker image [here](https://hub.docker.com/layers/apache/superset/5.0.0rc1/images/sha256-1191d03e7fd206794272ebe9d730d2a6d7a515ed5633951bed97d48d04d3c4a7), any plan to release a new image or backport this to 4.1.2? The issue is very annoying when team collaboration is needed.

LevisNgigi:
> @LevisNgigi I see... I am on the 5.0.0rc1 docker image [here](https://hub.docker.com/layers/apache/superset/5.0.0rc1/images/sha256-1191d03e7fd206794272ebe9d730d2a6d7a515ed5633951bed97d48d04d3c4a7), any plan to release a new image or backport this to 4.1.2? The issue is very annoying when team collaboration is needed.

@pang-wu This is a new feature that comes with v5.0. Not sure whether it will be backported. Maybe @sadpandajoe can clarify on that.

sadpandajoe:
> > @LevisNgigi I see... I am on the 5.0.0rc1 docker image [here](https://hub.docker.com/layers/apache/superset/5.0.0rc1/images/sha256-1191d03e7fd206794272ebe9d730d2a6d7a515ed5633951bed97d48d04d3c4a7), any plan to release a new image or backport this to 4.1.2? The issue is very annoying when team collaboration is needed.
> 
> @pang-wu This is a new feature that comes with v5.0. Not sure whether it will be backported. Maybe @sadpandajoe can clarify on that.

This won't be backported as part of 4.1 since don't pull in features to minor releases. As for a new docker image, we'll probably have a 5.0.0rc2 which will have this fix

marianysilva:
Should the [STANDARD_ROLES](https://github.com/apache/superset/blob/5.0/RESOURCES/STANDARD_ROLES.md) be updated?

LevisNgigi:
> Should the [STANDARD_ROLES](https://github.com/apache/superset/blob/5.0/RESOURCES/STANDARD_ROLES.md) be updated?

Yes I think they should,will try raise a pr for the same. @marianysilva

PR #32317, opened 2025-02-19
feat(flag flip): Setting Horizontal Filters to True by default.

rusackas:
### SUMMARY
I was a bit puzzled as to why this wasn't enable on an ephemeral environment. I think enough people are using this feature now that it ought to be on by default as part of 5.0 or a subsequent minor. I don't think this is a breaking change, but I added a note to UPDATING.md and the Feature Flags doc file accordingly. 

Tagging @kgabryje @michael-s-molina @mistercrunch in case they have any feelings on controversy here :) 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -24,7 +24,7 @@ assists people when migrating to a new version.
 
 ## Next
 
-- [32317](https://github.com/apache/superset/pull/32317) The `HORIZONTAL_FILTER_BAR` feature flag is now set to True by default.
+- [32317](https://github.com/apache/superset/pull/32317) The `HORIZONTAL_FILTER_BAR` feature is now out of testing/beta development and has been removed.
 - [31976](https://github.com/apache/superset/pull/31976) Removed the `DISABLE_LEGACY_DATASOURCE_EDITOR` feature flag. The previous value of the feature flag was `True` and now the feature is permanently removed.
 - [31959](https://github.com/apache/superset/pull/32000) Removes CSV_UPLOAD_MAX_SIZE config, use your web server to control file upload size.
 - [31959](https://github.com/apache/superset/pull/31959) Removes the following endpoints from data uploads: `/api/v1/database/<id>/<file type>_upload` and `/api/v1/database/<file type>_metadata`, in favour of new one (Details on the PR). And simplifies permissions.
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -24,6 +24,7 @@ assists people when migrating to a new version.
 
 ## Next
 
+- [32317](https://github.com/apache/superset/pull/32317) The `HORIZONTAL_FILTER_BAR` feature flag is now set to True by default.
 - [31976](https://github.com/apache/superset/pull/31976) Removed the `DISABLE_LEGACY_DATASOURCE_EDITOR` feature flag. The previous value of the feature flag was `True` and now the feature is permanently removed.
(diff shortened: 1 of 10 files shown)


mistercrunch (approved):

mistercrunch:
Agreed, should be the default!

michael-s-molina:
@rusackas We had a proposal about this in 4.0 that was blocked because cross-filters worked pretty poorly with horizontal filters when compared to the vertical approach. I don't know if we had a design review since then but it would be good to check before making this the default. @kasiazjc

sadpandajoe:
@kasiazjc For reference, this is what it'll look like with a horizontal filter bar with cross filtering:
Screenshot 2025-02-20 at 10 22 09 AM

This is for vertical:
Screenshot 2025-02-20 at 10 21 54 AM

mistercrunch:
Oh trying to distinguish between:
- horizontal filters are ENABLED to be used
- horizontal filters are USED by default

Does the feature flag here does only 1 or 1 and 2?

michael-s-molina (changes requested):
@mistercrunch @rusackas I checked the code and this feature flag does NOT change the default which is vertical. This is a development feature flag and not a configuration feature flag. For this reason, we can delete the feature flag instead of setting it to True given that we don't need the whole process of a breaking change. For users, it would just look like we added a new feature that allows to change the filter bar orientation.

rusackas:
@michael-s-molina works for me, if it works for you. Indeed, enabling OR deleting the flag adds the feature. If we don't perceive ords wanting to _disable_ the feature, I'm 100% happy to delete it :) Revision coming soon...

rusackas:
@michael-s-molina should be all good now! :)

michael-s-molina:
> @michael-s-molina should be all good now! :)

@rusackas Did you forget to commit something? The feature flag is still there.

rusackas:
Yeesh, looks like my intended push didn't make it through earlier. Should be there now.

UPDATING.md
```diff
@@ -24,6 +24,7 @@ assists people when migrating to a new version.
 
 ## Next
 
+- [32317](https://github.com/apache/superset/pull/32317) The `HORIZONTAL_FILTER_BAR` feature is now out of testing/beta development and has been removed.
```
michael-s-molina, line 27:
```suggestion
- [32317](https://github.com/apache/superset/pull/32317) The horizontal filter bar feature is now out of testing/beta development and its feature flag `HORIZONTAL_FILTER_BAR` has been removed.
```

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterBarSettings/index.tsx
```diff
@@ -193,9 +185,7 @@ const FilterBarSettings = () => {
           </FilterConfigurationLink>
         ),
       });
-      if (canSetHorizontalFilterBar) {
-        items.push({ type: 'divider' });
-      }
+      items.push({ type: 'divider' });
```
michael-s-molina, line 188:
```suggestion
      if (canEdit) {
         items.push({ type: 'divider' });
      }
```

michael-s-molina (approved):

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterBarSettings/index.tsx
rusackas: Ooooh, nice catch!

PR #32318, opened 2025-02-19
feat(feature flag): Setting `TAGGING_SYSTEM` to True by default

rusackas:
### SUMMARY
Following on the heels of https://github.com/apache/superset/pull/32317, and a discussion with @eschutho, we feel that this feature is ready for promotion to being on by default, whether this makes it into 5.0 or a subsequent minor. Again tagging @kgabryje @michael-s-molina @mistercrunch in case there are any concerns here preventing this change.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/RESOURCES/FEATURE_FLAGS.md
+++ b/RESOURCES/FEATURE_FLAGS.md
@@ -31,7 +31,6 @@ These features are considered **unfinished** and should only be used on developm
 - ENABLE_ADVANCED_DATA_TYPES
 - PRESTO_EXPAND_DATA
 - SHARE_QUERIES_VIA_KV_STORE
-- TAGGING_SYSTEM
 - CHART_PLUGINS_EXPERIMENTAL
 
 ## In Testing
@@ -65,6 +64,7 @@ These features flags are **safe for production**. They have been tested and will
 ### Flags on the path to feature launch and flag deprecation/removal
 
 - DASHBOARD_VIRTUALIZATION
+- TAGGING_SYSTEM
 
 ### Flags retained for runtime configuration
 
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -24,6 +24,7 @@ assists people when migrating to a new version.
 
 ## Next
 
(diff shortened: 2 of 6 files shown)


mistercrunch (approved):
Yup!

rusackas:
Whoa... CI is not happy. Not sure yet what can of worms I've opened here :/

michael-s-molina:
@rusackas We had a proposal about this in 4.0 that was blocked because cross-filters worked pretty poorly with horizontal filters when compared to the vertical approach. I don't know if we had a design review since then but it would be good to check before making this the default. @kasiazjc

mistercrunch:
Wondering who's the best person to help, though should be a simple glitch around feature flags handling in tests/builds(?)

sadpandajoe:
Shouldn't we mention this in UPDATING.md to let people know it's on by default now?

mistercrunch:
> Shouldn't we mention this in UPDATING.md to let people know it's on by default now?

I agree that feature flag default changes aren't breaking changes, but should be noted in `UPDATING.md`, where more heads up to the person performing the release is better. It allows them to know about the change in behavior, and prevent it if they decide so.

sadpandajoe:
@rusackas I pushed up a possible fix for the e2e cypress test that's failing. When tags are turned on, we add a `Tags` column between `Status` and `Owners`, so the e2e was failing since it was expecting `Owners` but it was returning `Tags`

sfirke:
@rusackas I'm wondering what the state of this is - it looks above like there was consensus to make this happen but it got aborted?

rusackas:
@sfirke I.... don't recall! I think perhaps there were some frustrating test failures, or the feature was deemed not-quite-ready for prime time in another conversation somewhere. 

If we wanna do this, I can try to reopen/recreate it, but I'm not sure if @sadpandajoe @msyavuz @yousoph have any feelings that we shouldn't (yet) flip it on for everyone. 

I suppose a "lazy consensus" thread couldn't hurt if there is any contentiousness about flipping it on when we're not in a major release cycle.

sfirke:
Hm I'll add it to the next town hall agenda!

PR #31972, opened 2025-01-23
refactor(Dropdown): Migrate Dropdown to Ant Design 5

msyavuz:
### SUMMARY

Migrate Dropdown to Ant Design 5

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:
image
image

After:
image
image

### TESTING INSTRUCTIONS

Check for visual changes manually

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #31952
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #32334, opened 2025-02-20
chore: Upgrade AG Grid to use tree shaking

kgabryje:
### SUMMARY
Upgrade AG Grid to version 33. This version improved [modules](https://www.ag-grid.com/javascript-data-grid/modules/) architecture, which lets us specify exactly which parts of the library we use and tree-shake the rest, thus improving the bundle size.
Additionally, I wrapped some props passed to AG Grid in `useMemo` to avoid rerenders

No visual changes

### TESTING INSTRUCTIONS
The AG Grid table in SQL Lab should work like before

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


kgabryje:
@justinpark could you please verify if I included all modules that we use in the table in SQL Lab? There's a tool that helps selecting the correct modules [here](https://www.ag-grid.com/javascript-data-grid/modules/#selecting-modules)

villebro (approved):
LGTM

michael-s-molina:
@kgabryje I pinged @justinpark and he'll take a look. Please don't merge it just yet 😉

superset-frontend/src/SqlLab/components/ResultSet/ResultSet.test.tsx
```diff
@@ -363,7 +368,7 @@ describe('ResultSet', () => {
       );
     });
     const { getByRole } = setup(mockedProps, mockStore(initialState));
-    expect(getByRole('treegrid')).toBeInTheDocument();
+    expect(getByRole('grid')).toBeInTheDocument();
```
kgabryje, line 371: For some reason, in jest runtime, the ag grid root had a role "grid", not "treegrid"

justinpark (approved):
LGTM! Thanks for upgrade

justinpark (changes requested):
wait. it looks like the header menu has gone in this version. Could you check what is missing?

superset-frontend/src/setup/setupAGGridModules.ts
```diff
@@ -0,0 +1,59 @@
...
+  TextFilterModule,
+  NumberFilterModule,
+  DateFilterModule,
+  ExternalFilterModule,
+  CsvExportModule,
+  ColumnApiModule,
+  RowApiModule,
+  CellApiModule,
+  RenderApiModule,
+  ClientSideRowModelModule,
```
justinpark, line 37: Could you check if there's an api for column header?

kgabryje:
> wait. it looks like the header menu has gone in this version. Could you check what is missing?

Which header menu? I'm comparing before/after side by side and I can't see a difference. Also nothing happens on right-click in both versions.

In the recording, I'm running master on localhost, and the other tab is the ephemeral env for this PR.

https://github.com/user-attachments/assets/4ad81d63-58f5-459a-93c0-f2c1e2f76daf

superset-frontend/src/setup/setupAGGridModules.ts
kgabryje: There isn't, the closest thing is ColumnApiModule which we include here

justinpark: CustomFilterModule might be the one.

superset-frontend/src/components/GridTable/index.tsx
```diff
@@ -216,21 +231,16 @@ function GridTable<RecordType extends object>({
         `}
       >
         <AgGridReact
+          // TODO: migrate to Theme API - https://www.ag-grid.com/react-data-grid/theming-migration/
+          theme="legacy"
```
justinpark, line 235: with this theme value, the existing ag-theme-quartz styles are broken.
_DEV__Superset

Please migrate to theme api first to apply the theme name here.

justinpark:
> > wait. it looks like the header menu has gone in this version. Could you check what is missing?
> 
> Which header menu? I'm comparing before/after side by side and I can't see a difference. Also nothing happens on right-click in both versions.
> 
> In the recording, I'm running master on localhost, and the other tab is the ephemeral env for this PR.
> 
>  Screen.Recording.2025-02-21.at.19.18.25.mov

I found that it's not related to this change but this [ant dropdown migration](https://github.com/apache/superset/pull/31972/files#diff-b35ab989924f8adea06aafc3caa5c8814d9aaf24146d7e0505cf25a675b8f785R101).

I will fix the dropdown menu issue separately.

superset-frontend/src/setup/setupAGGridModules.ts
kgabryje: Done, filtering works now 👍 

image

kgabryje:
@justinpark could you take another look?

justinpark (approved):
LGTM. Thanks for all hard work

PR #32336, opened 2025-02-20
fix(Slack): Fix Slack recipients migration to V2

### LINKED ISSUE #31936: Failed to update slack recipients to v2
### Bug description

I upgraded Superset to 4.1.1 and attempted to migrate the Slack API due to the old version being deprecated. However, the new version isn't functioning properly. The logs show an error, and while old reports and alerts work fine, the new version seems unable to upgrade or send reports.

I've already added `channels:read` scope. There is no problem to adding new alert. 

```
superset-worker-1  | [2025-01-21 05:55:13,723: INFO/MainProcess] Slack API v2 is available
superset-worker-1  | [2025-01-21 05:55:13,723: INFO/MainProcess] Attempting to upgrade the report to Slackv2: Report should not be run with the Slack V1 api.
superset-worker-1  |     Attempting to run with V2 if required Slack scopes are available
superset-worker-1  | [2025-01-21 05:55:15,462: WARNING/MainProcess] Failed to update slack recipients to v2: No recipients saved in the report
```

### Screenshots/recordings

_No response_

### Superset version

4.1.1

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

Vitor-Avila:
### SUMMARY
This PR fixes a few things with the Slack method migration from V1 to V2 (required as Slack is fully deprecating `files.upload()` in March):
* `get_channels_with_search` was returning a list of dictionaries (one per channel found) and we were dumping this directly in the `target` configuration for a report. The `target` field expects a comma-separated list of channels IDs for the V2 approach.
* I've added a schema that drops any additional fields from each channels to reduce information (some organizations could have thousands of channels).
* Enforcing `exact_match` search during the migration to avoid for example migrating a report from `other-channel` to `another-channel`.
* Stripping channel names before the search to account for `one-channel, second-channel`. Also removing leading `#` to channel names that were supported with the V1 method.
* Considering that V1 is reaching EOL and won't work anymore, if the bot has the required scopes for the migration to V2 but the migration fails, the execution now fails with an error that should notify owners for manual fix. The channel names are preserved in the form to make the process easier. Before this PR, the execution was logged as `SUCCESS` even though the notification was not sent, so this seems more appropriate.

One known issue is that the V1 integration allowed sending reports as a DM (by using the **Member ID**), and this is no longer possible with V2. The V2 APIs consider a DM as a conversation as well, so the bot would need to parse all Workspace members + public channels + private channels to support this, which would definitely hit the Slack rate limit (20+ requests per minute). We can revisit this once we have a better strategy to avoid rate limits.

Fixes https://github.com/apache/superset/issues/31936

**Report form when the migration failed**
image

**Execution logs when the migration failed**
image

### TESTING INSTRUCTIONS
Added testing.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/31936
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/report/execute.py
+++ b/superset/commands/report/execute.py
@@ -140,6 +140,7 @@ class BaseReportState:
                     # we need to ensure that existing reports can also fetch
                     # ids from private channels
                     channels_list = get_recipients_list(slack_recipients["target"])
+                    channels_list = [channel.lstrip("#") for channel in channels_list]
                     channels = get_channels_with_search(
                         search_string=channels_list,
                         types=[
@@ -164,10 +165,11 @@ class BaseReportState:
                         }
                     )
         except Exception as ex:
-            logger.warning(
-                "Failed to update slack recipients to v2: %s", str(ex), exc_info=True
-            )
-            raise UpdateFailedError from ex
+            # Revert to v1 to preserve configuration (requires manual fix)
+            recipient.type = ReportRecipientType.SLACK
+            msg = f"Failed to update slack recipients to v2: {str(ex)}"
+            logger.error(msg, exc_info=True)
+            raise UpdateFailedError(msg) from ex
 
     def create_log(self, error_message: Optional[str] = None) -> None:
         """
@@ -584,10 +586,13 @@ class BaseReportState:
                     recipient.type = ReportRecipientType.SLACKV2
                     notification = create_notification(recipient, notification_content)
                     notification.send()
-                except (UpdateFailedError, NotificationParamException) as err:
-                    # log the error but keep processing the report with SlackV1
-                    logger.warning(
-                        "Failed to update slack recipients to v2: %s", str(err)
+                except (UpdateFailedError, NotificationParamException) as mig_err:
+                    notification_errors.append(
+                        SupersetError(
+                            message=mig_err.message,
+                            error_type=SupersetErrorType.REPORT_NOTIFICATION_ERROR,
+                            level=ErrorLevel.ERROR,
+                        )
                     )
             except (NotificationError, SupersetException) as ex:
                 # collect errors but keep processing them
--- a/superset/commands/report/execute.py
+++ b/superset/commands/report/execute.py
@@ -149,7 +149,14 @@ class BaseReportState:
                         exact_match=True,
                     )
                     if len(channels_list) != len(channels):
-                        raise UpdateFailedError("Not all channels could be found")
+                        missing_channels = set(channels_list) - {
+                            channel["name"] for channel in channels
+                        }
+                        msg = (
+                            "Could not find the following channels: "
+                            f"{', '.join(missing_channels)}"
+                        )
+                        raise UpdateFailedError(msg)
                     channel_ids = ", ".join(channel["id"] for channel in channels)
                     recipient.recipient_config_json = json.dumps(
                         {
--- a/superset/commands/report/execute.py
+++ b/superset/commands/report/execute.py
@@ -72,7 +72,7 @@ from superset.reports.notifications.exceptions import (
 )
 from superset.tasks.utils import get_executor
 from superset.utils import json
-from superset.utils.core import get_recipients_list, HeaderDataType, override_user
+from superset.utils.core import HeaderDataType, override_user, recipients_string_to_list
 from superset.utils.csv import get_chart_csv_data, get_chart_dataframe
 from superset.utils.decorators import logs_context, transaction
 from superset.utils.pdf import build_pdf_from_screenshots
@@ -137,18 +137,19 @@ class BaseReportState:
                 if recipient.type == ReportRecipientType.SLACK:
                     recipient.type = ReportRecipientType.SLACKV2
                     slack_recipients = json.loads(recipient.recipient_config_json)
+                    # V1 method allowed to use leading `#` in the channel name
+                    channel_names = (slack_recipients["target"] or "").replace("#", "")
                     # we need to ensure that existing reports can also fetch
                     # ids from private channels
-                    channels_list = get_recipients_list(slack_recipients["target"])
-                    channels_list = [channel.lstrip("#") for channel in channels_list]
                     channels = get_channels_with_search(
-                        search_string=channels_list,
+                        search_string=channel_names,
                         types=[
                             SlackChannelTypes.PRIVATE,
                             SlackChannelTypes.PUBLIC,
                         ],
                         exact_match=True,
                     )
+                    channels_list = recipients_string_to_list(channel_names)
                     if len(channels_list) != len(channels):
                         missing_channels = set(channels_list) - {
                             channel["name"] for channel in channels
@@ -158,7 +159,7 @@ class BaseReportState:
                             f"{', '.join(missing_channels)}"
                         )
                         raise UpdateFailedError(msg)
-                    channel_ids = ", ".join(channel["id"] for channel in channels)
+                    channel_ids = ",".join(channel["id"] for channel in channels)
                     recipient.recipient_config_json = json.dumps(
                         {
                             "target": channel_ids,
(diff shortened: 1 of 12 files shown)


superset/commands/report/execute.py
```diff
@@ -139,15 +139,21 @@
                     slack_recipients = json.loads(recipient.recipient_config_json)
                     # we need to ensure that existing reports can also fetch
                     # ids from private channels
+                    current_target_count = slack_recipients["target"].split(",")
```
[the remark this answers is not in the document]
Vitor-Avila: In general a valid thought, but this is post `validate()` when the report/alert is created so the `target` configuration should have been properly handled.

superset/commands/report/execute.py
```diff
@@ -139,15 +139,21 @@ def update_report_schedule_slack_v2(self) -> None:
...
+                    channels = get_channels_with_search(
+                        slack_recipients["target"],
+                        types=[
+                            SlackChannelTypes.PRIVATE,
+                            SlackChannelTypes.PUBLIC,
+                        ],
+                        exact_match=True,
+                    )
+                    if len(current_target_count) != len(channels):
+                        raise UpdateFailedError("Not all channels could be found")
```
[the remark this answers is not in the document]
Vitor-Avila: This is good feedback! I'll add this in a following commit.

superset/reports/api.py
```diff
@@ -574,7 +575,7 @@ def slack_channels(self, **kwargs: Any) -> Response:
         """
         try:
             params = kwargs.get("rison", {})
-            search_string = params.get("search_string")
+            search_string = get_recipients_list(params.get("search_string"))
```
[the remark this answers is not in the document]
Vitor-Avila: I was using `get_recipients_list` here just to `strip()` and return a `list`. I'll remove it

superset/reports/api.py
```diff
@@ -574,7 +574,7 @@ def slack_channels(self, **kwargs: Any) -> Response:
         """
         try:
             params = kwargs.get("rison", {})
-            search_string = params.get("search_string")
+            search_string = [params.get("search_string", "").strip()]
```
eschutho, line 577: Any benefit for doing the search string cleanup here (and other place where you pass it to `get_channels_with_search`) as opposed to letting the `get_channels_with_search` be responsible for it in one place?

superset/commands/report/execute.py
```diff
@@ -139,22 +139,37 @@ def update_report_schedule_slack_v2(self) -> None:
...
                     )
         except Exception as ex:
-            logger.warning(
-                "Failed to update slack recipients to v2: %s", str(ex), exc_info=True
-            )
-            raise UpdateFailedError from ex
+            # Revert to v1 to preserve configuration (requires manual fix)
+            recipient.type = ReportRecipientType.SLACK
+            msg = f"Failed to update slack recipients to v2: {str(ex)}"
+            logger.error(msg, exc_info=True)
```
eschutho, line 171: You can change this to logger.exception and remove the exc_info

superset/commands/report/execute.py
```diff
@@ -571,10 +586,13 @@ def _send(
                     recipient.type = ReportRecipientType.SLACKV2
                     notification = create_notification(recipient, notification_content)
                     notification.send()
-                except (UpdateFailedError, NotificationParamException) as err:
-                    # log the error but keep processing the report with SlackV1
-                    logger.warning(
-                        "Failed to update slack recipients to v2: %s", str(err)
+                except (UpdateFailedError, NotificationParamException) as mig_err:
```
eschutho, line 589: can you catch these errors below on line 597?

superset/utils/core.py
```diff
@@ -811,7 +811,13 @@ def send_mime_email(
     smtp.quit()
 
 
-def get_email_address_list(address_string: str) -> list[str]:
+def get_recipients_list(address_string: str | None) -> list[str]:
```
eschutho, line 814: Since this isn't really fetching anything, can we change it to sound more like a utility, like "recipient_string_to_list"?

superset/reports/api.py
eschutho: My main concern here is really because we have this utility that converts a recipient string to a list, and so if we are going to pass around this search_string as a list, we should probably rename it as well.

superset/commands/report/execute.py
```diff
@@ -139,22 +139,37 @@ def update_report_schedule_slack_v2(self) -> None:
...
+                    )
+                    if len(channels_list) != len(channels):
+                        missing_channels = set(channels_list) - {
+                            channel["name"] for channel in channels
+                        }
+                        msg = (
+                            "Could not find the following channels: "
+                            f"{', '.join(missing_channels)}"
+                        )
+                        raise UpdateFailedError(msg)
```
eschutho, line 160: Do we want to block the conversion to v2 if channel ids can't be found or just log them and continue?

superset/reports/api.py
Vitor-Avila: I think that's a valid concern. I can probably move this back there but still rely on the `util` we already have.

superset/commands/report/execute.py
Vitor-Avila: great catch! I'll do that

superset/utils/core.py
Vitor-Avila: Sure! I'll do that in another commit

superset/commands/report/execute.py
Vitor-Avila: @eschutho from my tests currently (before this PR) what's happening is that the execution itself is listed as successful in the execution logs, but nothing is sent to Slack. I thought about calling `send()` with the type set to `V1` again, but that would cause an infinite loop as we would validate again if there are scopes to migrate, attempt the migration, fail again, etc. Calling it with `V2` would also fail as we don't have all IDs. 

Considering that the Slack deprecation is next Tuesday,  I think it makes sense to consider any issues to migrate to V2 an error that should be fixed asap. Let me know your thoughts -- happy to think of other alternatives.

eschutho: Ok, this makes sense.

superset/commands/report/execute.py
```diff
@@ -139,22 +139,37 @@ def update_report_schedule_slack_v2(self) -> None:
...
+                    channel_ids = ", ".join(channel["id"] for channel in channels)
```
eschutho, line 161: nit, do we need the space after the comma here?
```suggestion
                    channel_ids = ",".join(channel["id"] for channel in channels)
```

superset/commands/report/execute.py
```diff
@@ -139,22 +139,37 @@ def update_report_schedule_slack_v2(self) -> None:
...
                     recipient.recipient_config_json = json.dumps(
                         {
-                            "target": get_channels_with_search(
-                                slack_recipients["target"],
-                                types=[
-                                    SlackChannelTypes.PRIVATE,
-                                    SlackChannelTypes.PUBLIC,
-                                ],
-                            )
+                            "target": channel_ids,
```
eschutho, line 164: Thanks for catching this error/missing step, btw!

eschutho (approved):
This looks great. Small formatting suggestions, but good to merge when you're ready!

superset/commands/report/execute.py
Vitor-Avila: Thanks! Won't resolve this as I think it's good context for someone looking at the changes in the future

superset/commands/report/execute.py
Vitor-Avila: Oh, we don't. Removing this now.

superset/commands/report/execute.py
Vitor-Avila: This is actually inside an except block that's parallel to 597. But I can move that line to an "outer" try/except and have this specific except nested and I think that would remove this duplicated syntax. I'll try that now.

superset/reports/api.py
Vitor-Avila: Ohh, I know why I moved this outside. It's because during the v1 -> v2 migration, I need to know the list length before calling `get_channels_with_search` to confirm if all channels were found. Doing it outside and then again in the method felt repetitive, but considering this is specific to the migration, and this code will be removed soon as only V2 is going to be supported going forward I think that's ok.

PR #32339, opened 2025-02-21
feat: allow importing encrypted_extra

### LINKED ISSUE #31438: Unable to import datasources with `encrypted_extra` params via REST
### Bug description

At present it is possible to create a bigquery datasource which uses an explicit key by creating an `import_datasources.yaml` file that contains something similar to the block below:

```
databases:
  -
    database_name: mydatabase
    sqlalchemy_uri: bigquery://gcp-project-id/
    cache_timeout: 0
    expose_in_sqllab: true
    allow_run_async: false
    allow_ctas: false
    allow_cvas: false
    allow_dml: false
    allow_file_upload: false
    encrypted_extra: |
      {"credentials_info": {  "type": "service_account",  "project_id": "gcp-project-id",  "private_key_id": "aca359ef9828567aa1f031f367ba547fe26bf54c",  "private_key": "-----BEGIN PRIVATE KEY-----VERY_BIG_PRIVATE_KEY_STRING\n-----END PRIVATE KEY-----\n",  "client_email": "service-account@gcp-project-id.iam.gserviceaccount.com",  "client_id": "1149993100000000000",  "auth_uri": "https://accounts.google.com/o/oauth2/auth",  "token_uri": "https://oauth2.googleapis.com/token",  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/service-account%40gcp-project-id.iam.gserviceaccount.com",  "universe_domain": "googleapis.com"} }
    extra: |
      { "allows_virtual_table_explore": true }
    uuid: 00000000-0000-0000-0000-123456789001  
```

This can be imported using `superset import_datasources -p filename.yaml` from a running instance.

However, if I wish to import the same data structure using the REST API `/api/v1/database/import` it will fail.
It seems that the REST version applies additional schema checks on the bundled zipfile and fails because `encrypted_extra` is not part of `ImportV1DatabaseSchema`

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.11

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

betodealmeida:
### SUMMARY

Even though we don't include `encrypted_extra` in database exports, we allow it to be imported using the CLI. This PR brings the same behavior to the API, so that the `encrypted_extra` can be manually added to the bundle and the database imported with it.

Fixes https://github.com/apache/superset/issues/31438

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/databases/schemas.py
+++ b/superset/databases/schemas.py
@@ -229,7 +229,7 @@ def server_cert_validator(value: str) -> str:
     return value
 
 
-def encrypted_extra_validator(value: str) -> str:
+def encrypted_extra_validator(value: str | None) -> None:
     """
     Validate that encrypted extra is a valid JSON string
     """
@@ -240,7 +240,6 @@ def encrypted_extra_validator(value: str) -> str:
             raise ValidationError(
                 [_("Field cannot be decoded by JSON. %(msg)s", msg=str(ex))]
             ) from ex
-    return value
 
 
 def extra_validator(value: str) -> str:
@@ -855,6 +854,7 @@ class ImportV1DatabaseSchema(Schema):
     database_name = fields.String(required=True)
     sqlalchemy_uri = fields.String(required=True)
     password = fields.String(allow_none=True)
+    encrypted_extra = fields.String(allow_none=True, validate=encrypted_extra_validator)
     cache_timeout = fields.Integer(allow_none=True)
     expose_in_sqllab = fields.Boolean()
     allow_run_async = fields.Boolean()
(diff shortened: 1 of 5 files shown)


rusackas:
Is this a fix rather than a feature?

mistercrunch (approved):
LGTM

PR #32341, opened 2025-02-21
chore(build): reduce Lodash usage in `superset-frontend`

hainenber:
chore(build): reduce Lodash usage in `superset-frontend`

### SUMMARY

Reduce some Lodash usage for cases modern JS can handle by its own.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/home/ChartTable.tsx
+++ b/superset-frontend/src/features/home/ChartTable.tsx
@@ -18,7 +18,6 @@
  */
 import { useEffect, useMemo, useState } from 'react';
 import { t } from '@superset-ui/core';
-import { filter } from 'lodash';
 import {
   useChartEditModal,
   useFavoriteStatus,
@@ -76,7 +75,7 @@ function ChartTable({
     TableTab.Other,
   );
 
-  const filteredOtherTabData = filter(otherTabData, obj => 'viz_type' in obj);
+  const filteredOtherTabData = otherTabData?.filter(obj => 'viz_type' in obj);
 
   const {
     state: { loading, resourceCollection: charts, bulkSelectEnabled },
--- a/superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Screengrid/Screengrid.tsx
+++ b/superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Screengrid/Screengrid.tsx
@@ -23,7 +23,6 @@
 import { memo, useCallback, useEffect, useRef, useState } from 'react';
 import { ScreenGridLayer } from '@deck.gl/aggregation-layers';
 import { JsonObject, JsonValue, QueryFormData, t } from '@superset-ui/core';
-import { noop } from 'lodash';
 import sandboxedEval from '../../utils/sandbox';
 import { commonLayerProps } from '../common';
 import TooltipRow from '../../TooltipRow';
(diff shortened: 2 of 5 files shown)


superset-frontend/src/dashboard/hooks/useDownloadScreenshot.ts
```diff
@@ -44,9 +43,12 @@ export const useDownloadScreenshot = (
   const activeTabs = useSelector(
     (state: RootState) => state.dashboardState.activeTabs || undefined,
   );
-  const anchor = useSelector(
-    (state: RootState) =>
-      last(state.dashboardState.directPathToChild) || undefined,
+  const anchor = useSelector((state: RootState) =>
```
kgabryje, line 46: I think that last looks much more readable, especially when we have such long array names. Maybe we should either keep lodash or implement our own `last` util function?

hainenber: Yes can, I'll put in our own `last` impl.

superset-frontend/src/components/Select/AsyncSelect.tsx
```diff
@@ -574,7 +574,9 @@ const AsyncSelect = forwardRef(
         }
       } else {
         const token = tokenSeparators.find(token => pastedText.includes(token));
-        const array = token ? uniq(pastedText.split(token)) : [pastedText];
+        const array = token
```
michael-s-molina, line 577: Even though we can implement this using simple Javascript, `uniq` still offers semantic meaning that helps with readability.

superset-frontend/src/dashboard/hooks/useDownloadScreenshot.ts
michael-s-molina: I agree with @kgabryje but don't think we should implement our own function. We need to keep in mind that any function that we implement needs to be maintained and in general it's best to depend on libraries that are greatly supported like Lodash.

superset-websocket/src/config.ts
```diff
@@ -142,6 +140,50 @@ function applyEnvOverrides(config: ConfigType): ConfigType {
 }
 
 export function buildConfig(): ConfigType {
-  const config = _merge(defaultConfig(), configFromFile());
```
michael-s-molina, line 145: Previous version is way more readable and less error prone.

superset-frontend/src/features/home/ChartTable.tsx
```diff
@@ -76,7 +75,7 @@ function ChartTable({
     TableTab.Other,
   );
 
-  const filteredOtherTabData = filter(otherTabData, obj => 'viz_type' in obj);
+  const filteredOtherTabData = otherTabData?.filter(obj => 'viz_type' in obj);
```
michael-s-molina, line 78: 👍🏼

michael-s-molina (changes requested):
Thank you for the PR @hainenber. I think some of the changes are beneficial like `filter` changes but others decrease readability and increase our maintenance cost. I left comments on the ones that I think we should preserve. As a general guidance, I think it's always better to rely on greatly supported libraries like Lodash instead of rewriting code which requires testing and support.

hainenber:
Thanks all for the feedbacks. Looking at this PR retrospectively, I think I optimized too much at the expense of other factors like readability and module maturity. I'll retain the .filter` and `.noop` replacements and discard others.

michael-s-molina (approved):
Thank you for addressing the comments @hainenber. You have been one of the best additions to our committers roster. Keep up the excellent work 👏🏼

PR #32344, opened 2025-02-21
fix: ensure metric_macro expands templates

betodealmeida:
### SUMMARY

Make sure `{{ metric(...) }}` is able to expand templates, so that this works (pseudocode):

```
a = "{{ current_user_id() }}"  # 42
b = "{{ metric('a') }}"

{{ metric('b') }} => renders "42"
```

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -604,7 +604,9 @@ class BaseTemplateProcessor:
         try:
             return template.render(context)
         except RecursionError as ex:
-            raise SupersetTemplateException("Infinite recursion detected in template") from ex
+            raise SupersetTemplateException(
+                "Infinite recursion detected in template"
+            ) from ex
 
 
 class JinjaTemplateProcessor(BaseTemplateProcessor):
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -604,7 +604,7 @@ class BaseTemplateProcessor:
         try:
             return template.render(context)
         except RecursionError as ex:
-            raise SupersetTemplateException("Cyclic filters detected") from ex
+            raise SupersetTemplateException("Infinite recursion detected in template") from ex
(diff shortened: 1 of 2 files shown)


superset/jinja_context.py
```diff
@@ -602,7 +601,10 @@ def process_template(self, sql: str, **kwargs: Any) -> str:
         kwargs.update(self._context)
 
         context = validate_template_context(self.engine, kwargs)
-        return template.render(context)
+        try:
+            return template.render(context)
+        except RecursionError as ex:
+            raise SupersetTemplateException("Cyclic filters detected") from ex
```
Vitor-Avila, line 607: nit: Should we say "Cyclic Jinja macros detected"? Thinking that if the cyclic flow is caused by a  `{{ metric() }}` macro, `filters` might be confusing.

Vitor-Avila (approved):
LGTM! Also tested manually to validate both 1) the error message on a cyclic occurrence and 2) it works properly.

Left a non-blocking comment about the error message.

superset/jinja_context.py
betodealmeida: Yeah, the thing is, `metric` is neither a macro nor a filter. I need a better wording here.

betodealmeida, line 607:
```suggestion
            raise SupersetTemplateException("Infinite recursion detected in template") from ex
```

Vitor-Avila: good point

adamcunnington-mlg: Perhaps just Cyclic jinja detected or Cyclic references detected

PR #32358, opened 2025-02-24
feat: Adding the option and feature to enable borders with color, opacity and width control on heatmaps along with white borders on emphasis 

Dev10-34:
### SUMMARY
 In heatmaps, when there can be several continuous same colored regions, making it hard to distinguish between each heatmap element itself. Depending on need of the user, this differentiation capability might be required or crucial to their work. 
Added a color picker color control which keeps the borders transparent by default. Users can set the opacity as needed.
Added a slider control to set the border width as needed.  
Added a feature which pops out the element upon emphasis with white borders and shadows.

BEFORE
image

AFTER
image
image
image

### TESTING INSTRUCTIONS
No test files found for the same.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI: Adds an 'Enable Borders' Checkbox to Heatmap Customization, adds border and emphasis feature to heatmaps.


superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -278,6 +278,20 @@ const config: ControlPanelConfig = {
             },
           },
         ],
+        [
+          {
+            name: 'enable_borders',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Enable Borders'),
```
sadpandajoe, line 286: This should be sentence case

Dev10-34: Hello, 
This is my first time creating a PR. What should I be doing next?
Thanking you,
Regards
Dev10-34

michael-s-molina:
Thanks for the PR @Dev10-34. Could you instead of adding a checkbox to enabling black borders, add a control to set the border color which by default is transparent (no border)? This is important as the color border may conflict with the color scheme used by the chart.

Dev10-34:
> Thanks for the PR @Dev10-34. Could you instead of adding a checkbox to enabling black borders, add a control to set the border color which by default is transparent (no border)? This is important as the color border may conflict with the color scheme used by the chart.

Thank you for the response! I don't quite get what you mean by a control. Do you mean something like a slider?

michael-s-molina:
> Do you mean something like a slider?

Something like the `Series colors `control of the Waterfall chart.
Screenshot 2025-02-24 at 16 30 38

Dev10-34:
> > Do you mean something like a slider?
> 
> Something like the `Series colors `control of the Waterfall chart.
> 
> Screenshot 2025-02-24 at 16 30 38

I couldn't quite get the opacity part down. So for now, I have added 2 things. One checkbox which will let you control whether you want the borders or not and secondly a color picker so that you are able to change the color of your border as needed. The opacity function doesn't seem to work for waterfall chart either.

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +224,20 @@ const config: ControlPanelConfig = {
         ['y_axis_format'],
         ['x_axis_time_format'],
         ['currency_format'],
+        [
+          {
+            name: 'element_border',
+            config: {
+              type: 'ColorPickerControl',
+              label: t('Border Color'),
```
michael-s-molina, line 232:
```suggestion
              label: t('Border color'),
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +224,20 @@ const config: ControlPanelConfig = {
...
+              renderTrigger: true,
+              description: t(
+                'Select border color for Heatmap elements. Initial opacity is 0. Remember to change as per need',
```
michael-s-molina, line 235:
```suggestion
                'Select border color for Heatmap elements. Initial opacity is 0.',
```

michael-s-molina (commented):
Could you also remove the `package-lock.json` changes as there was no package change in this PR?

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
Dev10-34: I am sorry. I  made another mistake. Need to remove "Initial opacity set to 0" as changing opacity bar (a value) doesn't seem to have any effect

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
Dev10-34: Ok. On it

Dev10-34:
I added a function in utils.ts which lets you change the opacity as well now

superset-frontend/packages/superset-ui-core/src/color/utils.ts
```diff
@@ -120,3 +120,29 @@ export function rgbToHex(red: number, green: number, blue: number) {
 
   return `#${r}${g}${b}`;
 }
+
+export function rgbToHexWithAlpha(
```
michael-s-molina, line 124: We already have the `addAlpha` function for this purpose.

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +225,31 @@ const config: ControlPanelConfig = {
...
+              label: t('Border Color'),
+              renderTrigger: true,
+              description: t('Select border color for Heatmap elements'),
+              default: { r: 255, g: 255, b: 255, a: 1 },
+            },
+          },
+        ],
+        [
+          {
+            name: 'enable_border',
```
michael-s-molina, line 243: If the border color can be transparent by default, you don't need a checkbox to enable the border, you just need to set a color. You could also have a control for border width with a default value of 0 indicating no border.

michael-s-molina, line 243: If the border color can be transparent by default, you don't need a checkbox to enable the border, you just need to set a color. You could also have a control for border width with a default value of 0 indicating no border.

Dev10-34: Thank you for being so patient and helpful! I have added the changes as you mentioned. Earlier I didn't realize how to exactly use the addAlpha() function that was defined. I was creating a separate variable for opacity and passing that through the function which ended up not letting it function it properly. Passing elementBorder.a as the second argument was the way to go. 
I removed the function from ui-core packages along with the enable border checkbox. 
I added a slider control for border width as that looked nice. However, I set its default width to 1 as the effects should be visible upon increasing the opacity immediately and the user wouldn't need to work with 2 controls for their effect to show. 
I have a suggestion as well. Instead of having the opacity set to 0 (transparent), I keep the opacity to max, but keep the border width to 0. My reasoning is that for someone who doesn't really care for how the borders look, such a user will need to tweak just 1 visible control for them to be able to visualize their needs properly. Someone who is looking to customize and style their charts can use all the controls as needed.

Dev10-34: I have implemented my suggestion for the review. Please let me know what default to keep

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -19,6 +19,7 @@
 import { t, validateNonEmpty } from '@superset-ui/core';
 import {
   ControlPanelConfig,
+  ControlSubSectionHeader,
```
michael-s-molina, line 22:
```suggestion
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +225,31 @@ const config: ControlPanelConfig = {
         ['y_axis_format'],
         ['x_axis_time_format'],
         ['currency_format'],
+        [<ControlSubSectionHeader>Border</ControlSubSectionHeader>],
```
michael-s-molina, line 228:
```suggestion
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +225,31 @@ const config: ControlPanelConfig = {
...
+        [
+          {
+            name: 'element_border',
+            config: {
+              type: 'ColorPickerControl',
+              label: t('Border color'),
+              renderTrigger: true,
+              description: t('Select border color for Heatmap elements'),
+              default: { r: 255, g: 255, b: 255, a: 1 },
+            },
```
michael-s-molina, line 238: Could you also move the configuration after `linear_color_scheme` as it makes sense to be close to the color scheme?

```suggestion
          {
            name: 'border_color',
            config: {
              type: 'ColorPickerControl',
              label: t('Border color'),
              renderTrigger: true,
              description: t('The color of the elements border'),
              default: { r: 0, g: 0, b: 0, a: 1 },
            },
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +225,31 @@ const config: ControlPanelConfig = {
...
+            name: 'borders_width',
+            config: {
+              type: 'SliderControl',
+              label: t('Border width'),
+              renderTrigger: true,
+              min: 0,
+              max: 1,
+              step: 0.1,
+              description: t('Select border width'),
+            },
```
michael-s-molina, line 250:
```suggestion
          {
            name: 'border_width',
            config: {
              type: 'SliderControl',
              label: t('Border width'),
              renderTrigger: true,
              min: 0,
              max: 2,
              default: 0,
              step: 0.1,
              description: t('The width of the elements border'),
            },
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/transformProps.ts
```diff
@@ -86,6 +89,8 @@ export default function transformProps(
     metric = '',
     normalizeAcross,
     normalized,
+    elementBorder = { r: 255, g: 255, b: 255, a: 1 },
+    bordersWidth = 0,
```
michael-s-molina, line 93:
```suggestion
    borderColor,
    borderWidth,
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/transformProps.ts
```diff
@@ -161,6 +166,20 @@ export default function transformProps(
           return valueFormatter(paramsValue?.[2] as number | null | undefined);
         },
       },
+      itemStyle: {
+        borderColor: addAlpha(
+          rgbToHex(elementBorder.r, elementBorder.g, elementBorder.b),
+          elementBorder.a,
+        ),
+        borderWidth: bordersWidth,
+      },
```
michael-s-molina, line 175:
```suggestion
      itemStyle: {
        borderColor: addAlpha(
          rgbToHex(borderColor.r, borderColor.g, borderColor.b),
          borderColor.a,
        ),
        borderWidth,
      },
```

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/types.ts
```diff
@@ -33,6 +34,8 @@ export interface HeatmapFormData extends QueryFormData {
   metric: QueryFormMetric;
   normalizeAcross: 'heatmap' | 'x' | 'y';
   normalized?: boolean;
+  elementBorder?: RgbaColor;
+  bordersWidth: number;
```
michael-s-molina, line 38:
```suggestion
  borderColor: RgbaColor;
  borderWidth: number;
```

michael-s-molina (commented):
Thanks for the changes @Dev10-34. I left some suggestion to improve variable names.

Dev10-34:
I applied all the suggestions which you gave, however, upon re running there seems to be a bug sorta where even though the border width is set to 0 as default, on initial creation of the chart. If you move the border width slider, it fixes itself. Everything else works as intended functionally, just the visuals of the initial chart creation shows the borders as well, even if all the defaults are set correctly.
Giving borderWidth variable in transformProps.ts an initial value as 0 seems to fix this issue

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/transformProps.ts
```diff
@@ -86,6 +89,8 @@ export default function transformProps(
     metric = '',
     normalizeAcross,
     normalized,
+    borderColor,
+    borderWidth = 0,
```
Dev10-34, line 93: This fixes the issue

superset-frontend/plugins/plugin-chart-echarts/src/Heatmap/controlPanel.tsx
```diff
@@ -224,6 +224,31 @@ const config: ControlPanelConfig = {
         ['y_axis_format'],
         ['x_axis_time_format'],
         ['currency_format'],
+        [
```
michael-s-molina, line 227: Can you move these configs after `linear_color_scheme`? It makes sense to keep close the color schemes.

Dev10-34: I have made the suggested changes

michael-s-molina (approved):
Thank you so much for this great improvement @Dev10-34 and congrats on your first Superset PR!

Dev10-34:
> Thank you so much for this great improvement @Dev10-34 and congrats on your first Superset PR!

Thank you so much for having me!! Is there any possibility that these changes will be reflected in future tags/releases?

michael-s-molina:
> Is there any possibility that these changes will be reflected in future tags/releases?

It will be present in the latest Docker tag now and it will be officially released in 5.1.

Dev10-34:
> > Is there any possibility that these changes will be reflected in future tags/releases?
> 
> The `latest` Docker tag will contain this feature. It will be officially released in 5.1.

Glad to have been of help! Looking forward to more commits!

PR #29335, opened 2024-06-23
feat(reports): adding flag for removing the index on the report CSV

SkinnyPigeon:
### SUMMARY
Some of our account managers send their reports in CSV format. Currently, the index is shown in the CSV due to the `POST_PROCESSED` value being set in the call to the API. This is causing a few complaints from our clients due to the need for them to remove the column when working with the report data.

There is a current PR that has been ongoing for over a year and appears to have stalled. I have linked this below. Since this is unlikely to be completed, I am opening this in the hopes that we can resolve it sooner. As opposed to their approach, I reformat the `url` used to call the API based on the `CSV_INDEX`, leaving it as-is for its default behaviour.

Issue #[22981](https://github.com/apache/superset/issues/22981) 
[Current PR](https://github.com/apache/superset/pull/23155)
Fixes https://github.com/apache/superset/issues/32113

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
Superset Index

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [x] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [x] Introduces new feature or API
- [ ] Removes existing feature or API

Issue #22981, opened 2023-02-03
Unwanted row number column appears in csv / excel while scheduling in email.


rohitpawar2811:
Hello

I am scheduling a chart(in table form) in email as csv form , I am getting an extra column which has a count from 0 to n-row . 
Which I should not want.(And I think no one wants in CSV because we get numbering in csv opening platforms)

**Like :**

![Screenshot from 2023-02-03 19-10-58](https://user-images.githubusercontent.com/72196393/216618803-c82b78a7-995e-4934-a514-6bb0933431f2.png)

**Findings :** 

- When I directly download csv from chart I didn't getting that extra one column
- And one scenario :  In the case of Email it avoids the extra column add in csv only if the chart is not in table form.
   **Exactly  I would say when the chart is in the form of bar/any other visualization except table then it will not add an extra column in csv while sending in email.**
   
**Using :**  superset-2.0.0-dev

**My-Question :** 

- It is intentionally did ? 
- Or any thing I am forgetting to set by which i can avoid this extra-column(with numbering)
- Or this thing is resolved in the latest releases? 

 @michael-s-molina  @codyml

Rohit-pawar902:
```
def get_data(self, df: pd.DataFrame) -> Union[str, List[Dict[str, Any]]]:
        if self._query_context.result_format == ChartDataResultFormat.CSV:
            include_index = not isinstance(df.index, pd.RangeIndex)
            columns = list(df.columns)
            verbose_map = self._qc_datasource.data.get("verbose_map", {})
            if verbose_map:
                df.columns = [verbose_map.get(column, column) for column in columns]
            result = csv.df_to_escaped_csv(
                df, index=include_index, **config["CSV_EXPORT"]
            )
            return result or ""

        return df.to_dict(orient="records")
```

I have tracked the issue but I am here stuck that both in the case of
csv -direct-download
csv -email
both execute this function and genrate csv which has no index-colmn because **Include_index** is always false here
but some-how in case of csv-email from any else method index column is created again

I think while attching in email or while placing file on disk it goes with index-col

But got nothing , **any suggestion realted to this bug**

Rohit-pawar902:
@eschutho 
Hii I am working on this bug of superset , I need an help
Can you would tell location of this api endpoint
http://superset:8088/api/v1/chart/2470/data/?format=csv&type=post_processed&force=true
where could I get in code
Can anyone ping me the location or folder where this api exists

Rohit-pawar902:
@eschutho 
[...]

PR #23155, opened 2023-02-22
fix: Extra column in csv(or non-raw csv) get while sending on email

rohitpawar2811:
### SUMMARY

While sending csv on Email getting an extra column(non-raw csv get) which has a count from 0 to n-row .
Which I should not want.( And I think no one wants in CSV because we get numbering while opening it on platforms )

I have fix this thing and added CSV_INDEX=False //In superset_config.py
Which make to not create an index 
If You want index in csv make it True.

For More detail on this fix Refer this issue
https://github.com/apache/superset/issues/22981

### TESTING INSTRUCTIONS

- Simply Checked By sending chart as csv on email extra index-column is not coming

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes 22981
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #32366, opened 2025-02-24
feat(reports): removing index column

SkinnyPigeon:
### SUMMARY
This supersedes this PR: [29335](https://github.com/apache/superset/pull/29335)
By default, the URL generated for CSVs includes the argument `type=post_processed`. This can lead to an index column being included. This is often an undesirable inclusion and our analytics team has repeatedly asked for it to be removed. Following feedback from a member of the review team, it has been decided that this column should simply be removed.

Before
Screenshot 2025-03-12 at 21 35 25

After
Screenshot 2025-03-12 at 21 35 33

### TESTING INSTRUCTIONS
1. Build Docker image using current codebase
2. Create a dataset and table chart:
```sql
WITH dates AS (
    SELECT generate_series(
        (CURRENT_DATE - INTERVAL '90 days')::date,
        CURRENT_DATE::date,
        '1 day'::interval
    ) AS order_date
),
products AS (
    SELECT 'Product A' AS product_name, 100 AS base_price
    UNION ALL SELECT 'Product B', 200
    UNION ALL SELECT 'Product C', 150
    UNION ALL SELECT 'Product D', 75
),
regions AS (
    SELECT 'North' AS region
    UNION ALL SELECT 'South'
    UNION ALL SELECT 'East'
    UNION ALL SELECT 'West'
),
sales AS (
    SELECT 
        d.order_date,
        p.product_name,
        r.region,
        p.base_price,
        ROUND((p.base_price * (0.8 + RANDOM() * 0.4))::numeric, 2) AS sale_price,
        FLOOR(10 + RANDOM() * 90)::int AS quantity
    FROM 
        dates d
        CROSS JOIN products p
        CROSS JOIN regions r
    WHERE 
        RANDOM() < 0.3  -- Only keep some random combinations
)
SELECT 
    ROW_NUMBER() OVER (ORDER BY order_date, product_name, region) AS id,
    order_date,
    product_name,
    region,
    base_price,
    sale_price,
    quantity,
    (sale_price * quantity) AS total_revenue
FROM 
    sales
ORDER BY 
    order_date, product_name, region
LIMIT 100;
```
3. Schedule report
4. Build Docker image with these changes
5. Resend existing report

### ADDITIONAL INFORMATION

- [x] Has associated issue: FIxes [22981](https://github.com/apache/superset/issues/22981)
- [ ] Required feature flags:
- [ ] Changes UI


superset/commands/report/execute.py
```diff
@@ -363,6 +364,8 @@ def _get_pdf(self) -> bytes:
 
     def _get_csv_data(self) -> bytes:
         url = self._get_url(result_format=ChartDataResultFormat.CSV)
+        if not self._report_schedule.include_index:
+            url = remove_post_processed(url)
```
SkinnyPigeon, line 368: Adding logging above and below this shows the URL being mutated when the include index option is left unchecked

sadpandajoe:
@SkinnyPigeon I'd change this to be the default behavior by default as that is how the current app behaves.

geido:
I am restarting failed jobs, hopefully that was a flaky one

SkinnyPigeon:
> I am restarting failed jobs, hopefully that was a flaky one

Thanks @geido. I spotted that the failed test was for something I had not touched. Looks like we're all green now

rusackas:
At a glance, this looks good to me. Just needs a code owner review from @betodealmeida, @eschutho, @michael-s-molina, and/or @mistercrunch.

mistercrunch:
I'm not very familiar with the code here so I will pass on the review. @betodealmeida, @eschutho, @michael-s-molina, @dpgaspar

eschutho:
I'll check with @betodealmeida but it seems like this was a valid approach [here](https://github.com/apache/superset/pull/23155/files) except that the recommendation was to remove the index earlier.

SkinnyPigeon:
Thanks @eschutho for taking a look. I had a similar approach to use a feature flag [here](https://github.com/apache/superset/pull/29335) but it was suggested using the modal would be a better approach so it was not a global setting

eschutho:
> Thanks @eschutho for taking a look. I had a similar approach to use a feature flag [here](https://github.com/apache/superset/pull/29335) but it was suggested using the modal would be a better approach so it was not a global setting

Hi @SkinnyPigeon I agree, usually it is better to put options like that in a modal instead of a config, but in this case, I believe that we can simplify this and simply call this extra column a bug. We have heard a bit of feedback that most people do not want this column. To @betodealmeida's point in this [PR](https://github.com/apache/superset/pull/23155/files) it's ok to remove the column, but suggested just to remove it earlier in the process. 

I think a viable approach would be to remove the column in the charts/client_processing.py file.. something like this when a default index is used:
```
if isinstance(df.index, pd.RangeIndex):
            show_default_index = False
    
        processed_df.columns = [
            (
                " ".join(str(name) for name in column).strip()
                if isinstance(column, tuple)
                else column
            )
            for column in processed_df.columns
        ]

        processed_df.index = [
            (
                " ".join(str(name) for name in index).strip()
                if isinstance(index, tuple)
                else index
            )
            for index in processed_df.index
        ]

        if query["result_format"] == ChartDataResultFormat.JSON:
            if show_default_index:
                query["data"] = processed_df.to_dict()
            else:
                query["data"] = processed_df.to_dict(orient="index")
        elif query["result_format"] == ChartDataResultFormat.CSV:
            buf = StringIO()
            processed_df.to_csv(buf, index=show_default_index)
            buf.seek(0)
            query["data"] = buf.getvalue()

    return result
```
Let me know if this helps!

SkinnyPigeon:
Hey @eschutho 👋., that's interesting. So, to confirm, do you want the index column removed entirely and not just a toggleable option? Should we also remove the parameter from the URL or does that serve another purpose?

eschutho:
> Hey @eschutho 👋., that's interesting. So, to confirm, do you want the index column removed entirely and not just a toggleable option? Should we also remove the parameter from the URL or does that serve another purpose?

Hi @SkinnyPigeon yeah we can remove the index column when it is the default RangeIndex. I would suggest keeping the param in the URL because the client (js) needs to create the post processing and store it later for the report, and it's generated when that post_processed param is passed in.

superset/migrations/versions/2025-02-24_17-52_3b86f563edbc_add_remove_index_to_report_schedule.py
```diff
@@ -0,0 +1,41 @@
...
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "3b86f563edbc"
+down_revision = "74ad1125881c"
+
+
+def upgrade():
+    op.add_column(
```
sadpandajoe, line 34: We should be using the util function to add columns: https://github.com/apache/superset/blob/master/superset/migrations/shared/utils.py#L294

superset/migrations/versions/2025-02-24_17-52_3b86f563edbc_add_remove_index_to_report_schedule.py
```diff
@@ -0,0 +1,41 @@
...
+        "report_schedule",
+        sa.Column("remove_index", sa.Boolean(), default=False),
+    )
+
+
+def downgrade():
+    op.drop_column("report_schedule", "remove_index")
```
sadpandajoe, line 41: same comment as above to drop a column.

SkinnyPigeon:
Hey @eschutho, thanks for all of that. I will try to get that done soon

SkinnyPigeon:
Hey @eschutho, you pretty much handed me that on a plate. I've updated the PR with almost exactly what you gave me. However, I have managed to test it. First, I created a Dataset (PostgreSQL compatible):
```sql
WITH dates AS (
    SELECT generate_series(
        (CURRENT_DATE - INTERVAL '90 days')::date,
        CURRENT_DATE::date,
        '1 day'::interval
    ) AS order_date
),
products AS (
    SELECT 'Product A' AS product_name, 100 AS base_price
    UNION ALL SELECT 'Product B', 200
    UNION ALL SELECT 'Product C', 150
    UNION ALL SELECT 'Product D', 75
),
regions AS (
    SELECT 'North' AS region
    UNION ALL SELECT 'South'
    UNION ALL SELECT 'East'
    UNION ALL SELECT 'West'
),
sales AS (
    SELECT 
        d.order_date,
        p.product_name,
        r.region,
        p.base_price,
        ROUND((p.base_price * (0.8 + RANDOM() * 0.4))::numeric, 2) AS sale_price,
        FLOOR(10 + RANDOM() * 90)::int AS quantity
    FROM 
        dates d
        CROSS JOIN products p
        CROSS JOIN regions r
    WHERE 
        RANDOM() < 0.3  -- Only keep some random combinations
)
SELECT 
    ROW_NUMBER() OVER (ORDER BY order_date, product_name, region) AS id,
    order_date,
    product_name,
    region,
    base_price,
    sale_price,
    quantity,
    (sale_price * quantity) AS total_revenue
FROM 
    sales
ORDER BY 
    order_date, product_name, region
LIMIT 100;
```
Then this is the before:
Screenshot 2025-03-12 at 21 35 25

And this is the after:
Screenshot 2025-03-12 at 21 35 33

No more index column. I get the feeling you didn't need me for this one, but I'm always happy to help. This is such a great project, and it's always fun to work on the codebase

eschutho:
Hey, great @SkinnyPigeon, glad it worked! And we're happy to have your help on the project as well! I have some tests on my local that I can push up after your PR lands, too.

eschutho (approved):
Thanks @SkinnyPigeon!

PR #32372, opened 2025-02-25
fix(backend/async_events): allow user to configure username for Redis authentication in GLOBAL_ASYNC_QUERIES_CACHE_BACKEND

### LINKED ISSUE #32353: GLOBAL_ASYNC_QUERIES_CACHE_BACKEND ignores username
### Bug description

#30284 removed `GLOBAL_ASYNC_QUERIES_REDIS_CONFIG` for `GLOBAL_ASYNC_QUERIES_CACHE_BACKEND`,
however `GLOBAL_ASYNC_QUERIES_CACHE_BACKEND` even though it declares `CACHE_REDIS_USER` as a parameter  https://github.com/apache/superset/blob/2c37ddb2f63d216665c8d81232500986281ccfc0/superset/config.py#L1755 never passes the username down to the redis client 
 https://github.com/apache/superset/blob/master/superset/async_events/cache_backend.py#L27 which means the user/password pair is never correct.

With `GLOBAL_ASYNC_QUERIES_REDIS_CONFIG` this was possible to do as it had the ability to pass kwargs directly to the redis client. Which means before this change it was possible to use secure username/password redis, but after this change it's not.

It would also be good to restore the `CACHE_OPTIONS` value that `GLOBAL_ASYNC_QUERIES_REDIS_CONFIG` and others have.  This options expands the kwargs passed to the constructor of the underlying redis library.

Since superset it not the one handling the connection, this allowed us to use connections that redis supported but superset does not. e.g. the underlying redis and kombus library supports clustered redis, but superset does not have a cache type for it.

Using the `CACHE_OPTIONS` this worked fine before.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

hainenber:
Fixes #32353

### SUMMARY

`redis-py` can accept [username](https://redis-py.readthedocs.io/en/stable/connections.html) used for authentication. This PR is to restore that functionality.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/async_events/cache_backend.py
+++ b/superset/async_events/cache_backend.py
@@ -97,7 +97,7 @@ class RedisCacheBackend(RedisCache):
         }
 
         # Handle username separately as it's optional for Redis authentication.
-        if configured_username := config.get("CACHE_REDIS_USER", ""):
+        if configured_username := config.get("CACHE_REDIS_USER"):
             kwargs["username"] = configured_username
 
         return cls(**kwargs)
--- a/superset/async_events/cache_backend.py
+++ b/superset/async_events/cache_backend.py
@@ -95,6 +95,10 @@ class RedisCacheBackend(RedisCache):
             "ssl_cert_reqs": config.get("CACHE_REDIS_SSL_CERT_REQS", "required"),
             "ssl_ca_certs": config.get("CACHE_REDIS_SSL_CA_CERTS", None),
         }
+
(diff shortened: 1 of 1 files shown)


villebro (approved):
LGTM with a minor nit.

PR #32401, opened 2025-02-26
fix: prevent nested transactions

betodealmeida:
### SUMMARY

The `@transaction` decorator is used at the command level to ensure a single point of `commit`/`rollback` in the application stack. Using the decorator instead of individual calls to `db.session.commit()` in the stack ensure that requests are atomic.

Unfortunately we have a pattern where commands may call other commands. For example, the `UpdateDatabaseCommand.run()` method (which is decorated) calls the `UpdateSSHTunnelCommand.run()` method, which is also decorated with `@transaction`. This can lead to a situation where the SSH tunnel is updated and committed, but the database update fails and tries to rollback, resulting in an inconsistent state.

To ensure atomicity, this PR modifies the decorator to check if it's inside another decorated method. If so, the decorator returns the function unmodified, behaving as a no-op. With this change, for the example above the SSH tunnel update would only be committed together with the database changes.

N/A

### TESTING INSTRUCTIONS

Added basic unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/decorators.py
+++ b/superset/utils/decorators.py
@@ -255,6 +255,11 @@ def transaction(  # pylint: disable=redefined-outer-name
         def wrapped(*args: Any, **kwargs: Any) -> Any:
             from superset import db  # pylint: disable=import-outside-toplevel
 
+            if getattr(g, "in_transaction", False):
+                # If already in a transaction, call the function directly
+                return func(*args, **kwargs)
+
+            g.in_transaction = True
             try:
                 result = func(*args, **kwargs)
                 db.session.commit()  # pylint: disable=consider-using-transaction
@@ -266,6 +271,8 @@ def transaction(  # pylint: disable=redefined-outer-name
                     return on_error(ex)
 
                 raise
+            finally:
+                g.in_transaction = False
 
         return wrapped
 
--- a/tests/integration_tests/databases/api_tests.py
+++ b/tests/integration_tests/databases/api_tests.py
@@ -391,6 +391,7 @@ class TestDatabaseApi(SupersetTestCase):
         db.session.delete(model)
         db.session.commit()
 
+    @pytest.mark.skip("buggy")
     @mock.patch(
         "superset.commands.database.test_connection.TestConnectionDatabaseCommand.run",
     )
--- a/tests/unit_tests/utils/test_decorators.py
+++ b/tests/unit_tests/utils/test_decorators.py
@@ -24,6 +24,7 @@ from typing import Any, Optional
 from unittest.mock import call, Mock, patch
 
 import pytest
+from pytest_mock import MockerFixture
 
 from superset import app
 from superset.utils import decorators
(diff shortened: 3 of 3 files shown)


Vitor-Avila (approved):
This is super cool, and I think it's a more elegant solution than calling a `run()` (or similar method) with or without the `@transaction` decorator. I've tested this locally (as it will simplify https://github.com/apache/superset/pull/32231) and confirmed it works 🙌

michael-s-molina:
@betodealmeida @Vitor-Avila I tried to cherry-pick this PR into 5.0 but it's consistently throwing the follow error:

```
_____________ TestDatabaseApi.test_create_database_with_ssh_tunnel _____________

self = <tests.integration_tests.databases.api_tests.TestDatabaseApi testMethod=test_create_database_with_ssh_tunnel>
mock_get_all_schema_names = <MagicMock name='get_all_schema_names' id='140131078462096'>
mock_get_all_catalog_names = <MagicMock name='get_all_catalog_names' id='140131074401616'>
mock_create_is_feature_enabled = <MagicMock name='is_feature_enabled' id='140131069006096'>
mock_test_connection_database_command_run = <MagicMock name='run' id='140131060802576'>

    @mock.patch(
        "superset.commands.database.test_connection.TestConnectionDatabaseCommand.run",
    )
    @mock.patch("superset.commands.database.create.is_feature_enabled")
    @mock.patch("superset.models.core.Database.get_all_catalog_names")
    @mock.patch("superset.models.core.Database.get_all_schema_names")
    def test_create_database_with_ssh_tunnel(
        self,
        mock_get_all_schema_names,
        mock_get_all_catalog_names,
        mock_create_is_feature_enabled,
        mock_test_connection_database_command_run,
    ):
        """
        Database API: Test create with SSH Tunnel
        """
        mock_create_is_feature_enabled.return_value = True
        self.login(ADMIN_USERNAME)
        example_db = get_example_database()
        if example_db.backend == "sqlite":
            return
        ssh_tunnel_properties = {
            "server_address": "123.132.123.1",
            "server_port": 8080,
            "username": "foo",
            "password": "bar",
        }
        database_data = {
            "database_name": "test-db-with-ssh-tunnel",
            "sqlalchemy_uri": example_db.sqlalchemy_uri_decrypted,
            "ssh_tunnel": ssh_tunnel_properties,
        }
        uri = "api/v1/database/"
        rv = self.client.post(uri, json=database_data)
        response = json.loads(rv.data.decode("utf-8"))
>       assert rv.status_code == 201
E       assert 422 == 201
E        +  where 422 = <WrapperTestResponse 78 bytes [422 UNPROCESSABLE ENTITY]>.status_code
```

I noticed that you added `@pytest.mark.skip("buggy")` to `test_create_database_with_ssh_tunnel_no_port_no_default` which is a different test. I'll not cherry-pick this fix for 5.0 given that I'm concerned with side-effects throughout the codebase and the fact that https://github.com/apache/superset/pull/32231 is not included in 5.0. It might impact 5.1 though @sadpandajoe.

PR #32405, opened 2025-02-27
fix(daos/tag): prevent non-unique tags getting created along with unique ones

### LINKED ISSUE #30895: Bug with the tag uniqueness check
### Bug description

NB: I can’t currently run Superset 4 locally, but we backported the `TagDAO.create_custom_tagged_objects` to our Superset 3, and I _assume_ that the bug affects Superset 4 too (considering the code logic).

Basically, the reproduction is to call `POST /api/v1/tags/2/<chart_id> {"properties": {"tags": ["<some-tag>"]}}` twice (with the same chart and the same tag). Superset will fail with the tag uniqueness check.

The problem seems to be [here](https://github.com/apache/superset/blob/master/superset/daos/tag.py#L60-L74):
* in line 60, a new TaggedObject is created unconditionally (I think this is wrong)
* ...then, there is an existence check, and then in line 72, the same object is created conditionally.

At least when ported to the Superset 3 codebase, this fails predictably. Maybe Superset 4 somehow prevents that on another layer, but the method looks wrong in any case. (Or I am missing something.)

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.10

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

hainenber:
fix(daos/tag): prevent non-unique tags getting created along with unique ones

### SUMMARY

Fixes #30895

Unconditionally add a tag object before checking its uniqueness can fail the API call with payload combining non-unique and unique tags. This PR is to fix it.

N/A

### TESTING INSTRUCTIONS

Unit test extended.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/tags/dao_tests.py
+++ b/tests/integration_tests/tags/dao_tests.py
@@ -146,6 +146,15 @@ class TestTagsDAO(SupersetTestCase):
         # check if tag exists
         assert db.session.query(Tag).filter(Tag.name == "example tag 1").first()
 
+        # test that a combination of non-unique and unique tags
+        # can be added.
+        TagDAO.create_custom_tagged_objects(
+            object_type=ObjectType.dashboard.name,
+            object_id=1,
+            tag_names=["example tag 1", "example tag 2"],
+        )
+        assert db.session.query(Tag).filter(Tag.name == "example tag 2").first()
+
     @pytest.mark.usefixtures("load_world_bank_dashboard_with_slices")
     @pytest.mark.usefixtures("with_tagging_system_feature")
     @pytest.mark.usefixtures("create_tags")
--- a/superset/daos/tag.py
+++ b/superset/daos/tag.py
@@ -57,9 +57,6 @@ class TagDAO(BaseDAO[Tag]):
         for name in clean_tag_names:
             type_ = TagType.custom
             tag = TagDAO.get_by_name(name, type_)
-            tagged_objects.append(
-                TaggedObject(object_id=object_id, object_type=object_type, tag=tag)
-            )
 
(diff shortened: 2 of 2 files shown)


hainenber:
Re-bump for posterity :D

rusackas:
Pinging @mistercrunch @Vitor-Avila as additional possible reviewers.

mistercrunch:
Curious, didn't dig deep, but should we have a db migration that cleans ones created before this change?

Vitor-Avila:
I think @mistercrunch has a valid point here -- it would be good to:
1. find duplicated tags.
2. re-map association to the oldest one.
3. delete dupes.

tests/integration_tests/tags/dao_tests.py
```diff
@@ -146,6 +146,15 @@ def test_create_tagged_objects(self):
...
         assert db.session.query(Tag).filter(Tag.name == "example tag 1").first()
 
+        # test that a combination of non-unique and unique tags
+        # can be added.
+        TagDAO.create_custom_tagged_objects(
+            object_type=ObjectType.dashboard.name,
+            object_id=1,
+            tag_names=["example tag 1", "example tag 2"],
+        )
+        assert db.session.query(Tag).filter(Tag.name == "example tag 2").first()
```
Vitor-Avila, line 156: are we able to also validate that there's no duplicated tags? or tags association?

mistercrunch (approved):
LGTM actually, there's logic later that checks for uniqueness and appends only if it's not already there

PR #32414, opened 2025-02-27
fix(api): Added uuid to list api calls

### LINKED ISSUE #32413: Unable to correlate exports and live API resource definitions without uuid
### Bug description

At present there are two different types of APIs that can be used to view the resources. The list apis (such as `/api/v1/dashboard/` gives the configuration of the dashboard whilst `/api/v1/dashboard/export/` provides a mechanism for downloading dashboards (and related resources) for storing offline or later importing elsewhere.

Certain runtime fields, such as `last modified` and `owners` are rightly not present in the export APIs, but there are legitimate use cases where you would wish to be able to easily use both APIs together.

However, the list APIs are referenced by their internal PK id which is auto generated on creation and the exported resources are referenced via UUID and so neither have a way to easily map one definition to the other.

Ideally, we should have the UUID available in the list APIs to simplify correlation of the two views.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.11

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

withnale:
### SUMMARY

This is to address issue https://github.com/apache/superset/issues/32413

Ensure that the uuid is passed back with any resource list API call.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes https://github.com/apache/superset/issues/32413
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/charts/api.py
+++ b/superset/charts/api.py
@@ -206,6 +206,7 @@ class ChartRestApi(BaseSupersetModelRestApi):
         "tags.id",
         "tags.name",
         "tags.type",
+        "uuid",
     ]
     list_select_columns = list_columns + ["changed_by_fk", "changed_on"]
     order_columns = [
--- a/superset/dashboards/api.py
+++ b/superset/dashboards/api.py
@@ -217,6 +217,7 @@ class DashboardRestApi(BaseSupersetModelRestApi):
         "tags.id",
         "tags.name",
         "tags.type",
+        "uuid",
     ]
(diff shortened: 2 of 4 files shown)


sadpandajoe (approved):
LGTM, doesn't seem like we're removing or updating the response so not going to be a breaking change

rusackas:
running CI 🤞

PR #32422, opened 2025-02-28
fix(security): bind  SavedQueryView list permission to read permission

codenamelxl:
### SUMMARY

The  `SavedQuery` View has mostly migrated to `read` and `write` permission
This migration [superset/migrations/versions/2020-11-20_14-24_e38177dbf641_security_converge_saved_queries.py](https://github.com/apache/superset/blob/master/superset/migrations/versions/2020-11-20_14-24_e38177dbf641_security_converge_saved_queries.py)  has moved the `list` permission to `read`. However, the views still bind to `list` permission, causing permission lacking if user access that view directly through the link instead of navigating through SPA.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/RESOURCES/STANDARD_ROLES.md
+++ b/RESOURCES/STANDARD_ROLES.md
@@ -20,8 +20,8 @@ under the License.
 |                                                  |Admin|Alpha|Gamma|SQL_LAB|
 |--------------------------------------------------|---|---|---|---|
 | Permission/role description                      |Admins have all possible rights, including granting or revoking rights from other users and altering other people’s slices and dashboards.|Alpha users have access to all data sources, but they cannot grant or revoke access from other users. They are also limited to altering the objects that they own. Alpha users can add and alter data sources.|Gamma users have limited access. They can only consume data coming from data sources they have been given access to through another complementary role. They only have access to view the slices and dashboards made from data sources that they have access to. Currently Gamma users are not able to alter or add data sources. We assume that they are mostly content consumers, though they can create slices and dashboards.|The sql_lab role grants access to SQL Lab. Note that while Admin users have access to all databases by default, both Alpha and Gamma users need to be given access on a per database basis.||
-| can read on SavedQuery                           |:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
-| can write on SavedQuery                          |:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
+| can read on SavedQuery                           |:heavy_check_mark:|O|O|:heavy_check_mark:|
+| can write on SavedQuery                          |:heavy_check_mark:|O|O|:heavy_check_mark:|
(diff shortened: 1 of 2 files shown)


codenamelxl:
The failing test seems un-related.

rusackas:
> The failing test seems un-related.

re-running it 🤞

codenamelxl:
Is there anything I can do to help this get merged?

rusackas:
Some concerns from our friend Claude. Thoughts?

**Breaking change**: This appears to be removing saved query access from Alpha and Gamma users, which could break existing workflows for these users who currently rely on this functionality.

**Inconsistency**: The code change adds a @permission_name("read") decorator, but the documentation changes remove read permissions from Alpha/Gamma roles. This seems contradictory.

**Missing context**: The PR doesn't include:

- A description explaining why these permissions are being removed
- Any migration plan for existing users
- Tests to verify the permission changes work correctly
- Discussion of the security implications

**Documentation accuracy**: If Alpha and Gamma users can no longer read saved queries, the general role descriptions at the top of the document may need updating, as they currently state Alpha users "have access to all data sources."

codenamelxl:
> Some concerns from our friend Claude. Thoughts?
> 
> **Breaking change**: This appears to be removing saved query access from Alpha and Gamma users, which could break existing workflows for these users who currently rely on this functionality.
> 
> **Inconsistency**: The code change adds a @permission_name("read") decorator, but the documentation changes remove read permissions from Alpha/Gamma roles. This seems contradictory.
> 
> **Missing context**: The PR doesn't include:
> 
> * A description explaining why these permissions are being removed
> * Any migration plan for existing users
> * Tests to verify the permission changes work correctly
> * Discussion of the security implications
> 
> **Documentation accuracy**: If Alpha and Gamma users can no longer read saved queries, the general role descriptions at the top of the document may need updating, as they currently state Alpha users "have access to all data sources."

Thanks for the comment

- Alpha and Gamma roles do not have read/write permission on SavedQuery currently. So this is to correct the doc.
- As mentioned in the migration [superset/migrations/versions/2020-11-20_14-24_e38177dbf641_security_converge_saved_queries.py](https://github.com/apache/superset/blob/master/superset/migrations/versions/2020-11-20_14-24_e38177dbf641_security_converge_saved_queries.py) , legacy SavedQuery permissions were simplified to only `read/write` permissions. But the `list` endpoint was not update to `read` but still `list` permission. This MR correct that.
- There should be no migration plan needed as the permission has been granted to users already, this change only correct the permission check on the endpoint.

rusackas:
I'll let @dpgaspar make the call here... just sent him a Slack message about this :D

rusackas:
This PR hasn't been updated in a long time. Testing instructions and before/after screenshots are missing from the PR description. We'll mark it as draft for now — please rebase against `master`, add testing instructions, and set it back to 'ready for review' when it's ready.

rusackas:
The open question about whether this removes saved-query access for Alpha/Gamma was never settled, and it needs testing instructions. Closing as inactive - happy to reopen if @dpgaspar wants to weigh in.

PR #32473, opened 2025-03-01
fix(plugin-chart-echarts): remove erroneous upper bound value

### LINKED ISSUE #32428: Confidence interval shading in timeseries displays the wrong upper bound value
### Bug description

1. Go to timeseries chart
2. Add __yhat, __yhat_upper, __yhat_lower to the metrics names, which activates the confidence interval shading described through here https://github.com/apache/superset/discussions/19145
3. I have set the `__yhat_upper` to be max value of my column, __yhat_lower to be the min value of my column (using SQL such as MAX(column) to set the metric definition of `__yhat_upper` )
4. Save and then interact with chart, you'll see the max `__yhat_upper`  is not actually max (for example, the legend for `__yhat_upper` on hover will show 7.75, but the dot at top of confidence interval shading of that day will show 9.61. This is probably because the top is actually showing the diff between 7.75 and the  `_yhat_lower` -1.86 - instead of max, which is misleading.

We believe this would constitute a bug because we'd want to show the actual upper values upon hover on top of confidence interval, not the diff between lower and upper. Thank you!

### Screenshots/recordings

Image

### Superset version

master

### Python version

3.10

### Node version

20

### Browser

Safari

### Additional context

_No response_

### Checklist

villebro:
### SUMMARY
Currently when selecting "show value" on the chart, the upper bound shows the confidence bound width (400), not the actual value (which is 1.76k). In the tooltip this is fixed by adding the lower bound to the upper bound. However, this is more difficult to implement in the label formatter, as it only receives the single value of the series, while the tooltip formatter receives all series, making it easier to do the addition.

### BEFORE
Notice the incorrect upper bound label (400), as opposed to the actual value seen on the tooltip (1.76k):
image

### AFTER
Now the lower and upper bound labels are no longer shown when hovering:
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes #32428
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
@@ -226,7 +226,7 @@ export function transformSeries(
     stackId = forecastSeries.name;
   } else if (stack && isObservation) {
     // the suffix of the observation series is '' (falsy), which disables
-    // stacking. Therefore we need to set something that is truthy.
+    // stacking. Therefore, we need to set something that is truthy.
     stackId = getTimeCompareStackId('obs', timeCompare, name);
   } else if (stack && isTrend) {
     stackId = getTimeCompareStackId(forecastSeries.type, timeCompare, name);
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM. I haven't checked if the feature is covered by RTL, but maybe there's an enhancement if so. If it's enzyme, then I wouldn't bother at this time.

villebro:
> LGTM. I haven't checked if the feature is covered by RTL, but maybe there's an enhancement if so. If it's enzyme, then I wouldn't bother at this time.

I don't think this is testable with RTL, as we're just wrapping ECharts in a thin React component (ECharts implements all the drawing stuff at a very low level with ZRender). But working on this gave me a good proper headache, so this plugin is definitely in need of some cleanup at some point. Once we have the new extensions architecture in place, I'd love for us to give this plugin some TLC and rethink how we can do some real RTL-style testing.

Issue #32426, opened 2025-02-28
docker compose setup fails with superset_node | npm error Cannot read properties of undefined (reading 'extraneous')


jpchev:
### Bug description

I cloned the git code and followed the instructions on the doc page
https://superset.apache.org/docs/contributing/development/#docker-compose-recommended
to setup a dev environment with docker,
but the superset_node container fails and exists with an error,
can you please help me in understanding what I'm doing wrong?

Thank you

```

**superset_node         | npm error Cannot read properties of undefined (reading 'extraneous')**
superset_node         | npm notice
superset_node         | npm notice New major version of npm available! 10.8.2 -> 11.1.0
superset_node         | npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.1.0
superset_node         | npm notice To update run: npm install -g npm@11.1.0
superset_node         | npm notice
```

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

jpchev:
@dosu `BUILD_SUPERSET_FRONTEND_IN_DOCKER` is undocumented and it's true by default

jpchev:
after pulling again from the master, the error disappeared, but the static resources are not found when loading the web page http://localhost:8088/superset/welcome/

rusackas:
We need to figure out how to evangelize people using :9000 rather than :8088. We need to document that and the `BUILD_SUPERSET_FRONTEND_IN_DOCKER` config, so I'll tag @sfirke, but we're welcome to a docs contribution there. Also tagging @mistercrunch as he was weighing options on if/how/where to do forwarding so people _can't_ land on 8088 anymore, since this is a common issue.

jpchev:
thanks for the info @rusackas actually I was looking at the docker compose and trying to figure out how the service call each other why there are 2 tcp ports exposed.

Now I call http://localhost:9000/ and get the following error in container logs

```
superset_node         | [HPM] Error occurred while proxying request localhost:9000/ to http://host.docker.internal:8088/ [ENOTFOUND] (https://nodejs.org/api/errors.html#errors_common_system_errors)
[...]

PR #32481, opened 2025-03-03
fix(docker compose): replace port 8088 with 9000

jpchev:
### SUMMARY
This PR is to fix the `docker-compose.yml`, see
https://github.com/apache/superset/issues/32426#issuecomment-2692936100

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -176,7 +176,7 @@ services:
       NPM_RUN_PRUNE: false
       SCARF_ANALYTICS: "${SCARF_ANALYTICS:-}"
       # configuring the dev-server to use the host.docker.internal to connect to the backend
-      superset: "http://host.docker.internal:8088"
+      superset: "http://superset:8088"
     ports:
       - "127.0.0.1:9000:9000"  # exposing the dynamic webpack dev server
     container_name: superset_node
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -176,7 +176,7 @@ services:
       NPM_RUN_PRUNE: false
       SCARF_ANALYTICS: "${SCARF_ANALYTICS:-}"
       # configuring the dev-server to use the host.docker.internal to connect to the backend
-      superset: "http://host.docker.internal:8088"
+      superset: "http://superset:8088"
```
rusackas, line 179: This seems to work, but I'm not sure if/why the change is needed.

jpchev: on my linux machine, host.docker.internal doesn't seem to work, I don't know if I need some extraconfiguration, but I'd rather ask why is host.docker.internal, instead of just simply calling the `superset` service?

host.docker.internal calls a service on the underlying host network, why is this needed?

mistercrunch: asked GPT about it and seems this is a better approach -> https://chatgpt.com/share/67c6330b-9500-8010-b26a-0e1666025dea

mistercrunch (approved):

PR #32486, opened 2025-03-03
fix: skip DB filter when doing OAuth2

betodealmeida:
### SUMMARY

When users OAuth2 the backend needs to load the database instance in order to get the OAuth2 configuration. For users without the read database permission this currently doesn't work because we're going through the DAO. This PR changes the call to not apply a filter, so that OAuth2 can be used by everyone.

N/A

### TESTING INSTRUCTIONS

Added a test.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila (approved):
LGTM! Thanks for fixing this

PR #32499, opened 2025-03-04
fix(explore): Glitch in a tooltip with metric's name

kgabryje:
### SUMMARY
Fixes a small visual bug in a tooltip that displays metric's name and label - before, "Metric name", ":", and the name would be displayed in separate lines rather than inline.

Before:

image

After:

image

### TESTING INSTRUCTIONS
1. Go to Explore
2. Add a metric with a long name, so that it's truncated and a tooltip is displayed on hover
3. Verify that the name is displayed inline with the title "Metric name"
4. Check that it works for columns too and in the metrics/columns list on the left

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
@@ -23,20 +23,17 @@ import { ColumnMeta, Metric } from '@superset-ui/chart-controls';
 
 const TooltipSectionWrapper = styled.div`
   ${({ theme }) => css`
-    display: flex;
-    flex-direction: column;
+    display: -webkit-box;
+    -webkit-line-clamp: 40;
+    -webkit-box-orient: vertical;
+    overflow: hidden;
+    text-overflow: ellipsis;
+
     font-size: ${theme.typography.sizes.s}px;
     line-height: 1.2;
 
     &:not(:last-of-type) {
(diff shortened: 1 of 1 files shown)


superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
```diff
@@ -23,21 +23,18 @@ import { ColumnMeta, Metric } from '@superset-ui/chart-controls';
 
 const TooltipSectionWrapper = styled.div`
   ${({ theme }) => css`
-    display: flex;
-    flex-direction: column;
+    display: -webkit-box;
+    -webkit-line-clamp: 40;
```
rusackas, line 27: I haven't looked at all of 'em, but you might be able to live without the webkit prefix these days:
https://caniuse.com/mdn-css_properties_line-clamp

rusackas (approved):
We might not need to be so `webkit` specific, but it looks good to me otherwise.

superset-frontend/packages/superset-ui-chart-controls/src/components/labelUtils.tsx
kgabryje: By the way, some sources claim that `line-clamp` works and then... provide an example with `-webkit-line-clamp`

Issue #32480, opened 2025-03-03
Creating an alert with Slack notifications errors out due to rate-limiting in large Slack workspaces


Usiel:
### Bug description

Preconditions: Be connected to a Slack workspace with thousands of channels and even more archived channels.

1. Make sure Slack is connected (v1 or v2)
2. Edit an alert and open the "Notification method" section
3. Select Slack

Actual: The request to `/api/v1/report/slack_channels/?...` errors out with a rate-limit error from the Slack client.
Expected: User can select a channel/group without much delay.

### Screenshots/recordings

![Image](https://github.com/user-attachments/assets/7dd01bf5-e827-45d3-ac44-4743d968cb7e)

### Superset version

4.1.1

### Python version

3.11

### Node version

Not applicable

### Browser

Chrome

### Why?

From what I could see this is due to Slack's implementation of the `conversation.list` endpoint. When the connected workspace has lots of archived channels, then the cursor will return far less conversations than specified in the request (using `limit=999`). This means we end up doing a lot of requests to fully iterate all conversations, eventually running into the rate-limit.

I applied a diff to properly handle rate-limit errors. This allows the request to eventually finish (6 minutes for our workspace) but it's not a great user experience 😆 (It may still be useful to add the rate-limit handler to preempt other Slack API issues).

```diff
diff --git a/superset/utils/slack.py b/superset/utils/slack.py
index 468429fb60..5d0cbfe507 100644
--- a/superset/utils/slack.py
+++ b/superset/utils/slack.py
@@ -22,6 +22,7 @@ from typing import Optional
 from flask import current_app
 from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
+from slack_sdk.http_retry.builtin_handlers import RateLimitErrorRetryHandler

 from superset import feature_flag_manager
 from superset.exceptions import SupersetException
@@ -43,7 +44,12 @@ def get_slack_client() -> WebClient:
     token: str = current_app.config["SLACK_API_TOKEN"]
     if callable(token):
         token = token()
-    return WebClient(token=token, proxy=current_app.config["SLACK_PROXY"])
+    client = WebClient(token=token, proxy=current_app.config["SLACK_PROXY"])
+
+    rate_limit_handler = RateLimitErrorRetryHandler(max_retry_count=2)
+    client.retry_handlers.append(rate_limit_handler)
+
+    return client


 def get_channels_with_search(
```

### Options

[...]

PR #32510, opened 2025-03-05
feat(slack): adds rate limit error handler for Slack client

Usiel:
### SUMMARY

While investigating https://github.com/apache/superset/issues/32480 I noticed that Superset's Slack client does not handle rate-limiting. Although the original issue may have been intermittent, I still believe it will be useful to handle rate-limiting to make Slack operations with Superset more robust.

The rate-limit is easily triggered by 2 users trying to change a Slack alert with a large Slack workspace (multiple `conversations.list` requests with rate-limit 20 per minute), or lots of reports firing at the same time. 2 retries should be sufficient for most use cases. I didn't want to set it too high to avoid keeping users waiting when there's something else broken (e.g. same Slack token is used by a rogue process constantly using the full quota).

The `RateLimitErrorRetryHandler` feature is available since `slack_sdk>=3.9.0` (Superset requires at least `3.19.0`).

### TESTING INSTRUCTIONS

With a sufficiently large workspace one can trigger the rate-limiting and test the retry handler simply by opening the alert edit screen in 2 or more tabs at the same time.

Here's another way using `superset shell`:

```
from superset.utils.slack import get_channels_with_search

while True:
    get_channels_with_search()
```

At some point we expect to see
```
INFO:slack_sdk.web.base_client:A retry handler found: RateLimitErrorRetryHandler for POST https://slack.com/api/conversations.list - HTTP Error 429: Too Many Requests
```

After which we continue successfully (after a short delay). Without this change we would fail immediately.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

None ^

### DIFF
--- a/superset/utils/slack.py
+++ b/superset/utils/slack.py
@@ -22,6 +22,7 @@ from typing import Optional
 from flask import current_app
 from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
+from slack_sdk.http_retry.builtin_handlers import RateLimitErrorRetryHandler
 
 from superset import feature_flag_manager
 from superset.exceptions import SupersetException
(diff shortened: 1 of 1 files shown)


Vitor-Avila (approved):
LGTM! This is in accordance with https://tools.slack.dev/python-slack-sdk/webhook/#retryhandler

Vitor-Avila:
Hey @Usiel could you please rebase your branch? I believe that `pre-commit` issue might be fixed in `master` now. Thank you!

eschutho (approved):
Thanks @Usiel!

Usiel:
Thanks for the reviews! Rebased on `master` 🤞

PR #32514, opened 2025-03-05
refactor(Select): Migrate Select component to Ant Design 5

msyavuz:
### SUMMARY

Migrate Select component to Ant Design 5. Simplifying the Select component in the process is a secondary objective.

Before:
image
After:
image

Before:
image
After:
image

Before:
image
After:
image

### TESTING INSTRUCTIONS

Run the testing suite and do visual testing.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
@michael-s-molina and @geido put a lot of work into consolidating numerous Select modules into one Super-Duper-Select. I would check with either/both of them regarding any simplification efforts, just in case there are use cases around the intricacies. Of course if AntD now includes the functionality we built around it previously, or there are unnecessary style layers, then yes, let's go as "vanilla" as we can!

superset-frontend/src/components/Datasource/DatasourceEditor.test.jsx
```diff
@@ -197,7 +197,7 @@ describe('DatasourceEditor', () => {
 });
 
 describe('DatasourceEditor RTL', () => {
-  jest.setTimeout(15000); // Extend timeout to 15s for this test
+  jest.setTimeout(20000); // Extend timeout to 20s for this test
```
geido, line 200: What made this necessary? Is there a performance degradation of some sort?

superset-frontend/src/components/Select/Select.tsx
```diff
@@ -264,8 +276,12 @@ const Select = forwardRef(
             }
             return [
               SELECT_ALL_VALUE,
-              ...selectAllEligible.map(opt => opt.value),
-            ] as AntdLabeledValue[];
+              ...selectAllEligible
+                .map(opt => opt.value)
+                .filter(
+                  (val): val is RawValue => val !== null && val !== undefined,
```
geido, line 282: Can we double check this? I see that here the expected value is `RawValue` but I think I remember that in some cases `null` values could be acceptable in the Select. @michael-s-molina you might have a better memory about this if you can help? 

See also following changes where this same logic is being applied.

superset-frontend/src/components/Select/styles.tsx
```diff
@@ -18,8 +18,7 @@
  */
 import { styled } from '@superset-ui/core';
 import Icons from 'src/components/Icons';
-import { Spin, Tag } from 'antd';
-import AntdSelect from 'antd/lib/select';
+import { Spin, Tag, Select } from 'antd-v5';
```
geido, line 21: We can't be pulling directly from Ant Design. Can we refer to the existing src/components/{component} (`Tag` has one, `Spin` would need a wrapper but should be easy to include in this PR?)

superset-frontend/src/components/Select/styles.tsx
```diff
@@ -39,35 +38,41 @@ export const StyledContainer = styled.div<{ headerPosition: string }>`
...
-    // This is fixed in version 4.16
-    .ant-select-arrow .anticon:not(.ant-select-suffix) {
-      pointer-events: none;
+    .antd5-select-selection-item::after {
+      height: 0;
+      display: block !important;
     }
     .select-all {
+      border-radius: 0;
       border-bottom: 1px solid ${theme.colors.grayscale.light3};
```
geido, line 61: Can we re-check all usages of colors for the Select here and everywhere this is implemented? Can we go back to vanilla?

superset-frontend/src/components/Select/utils.tsx
```diff
@@ -17,20 +17,40 @@
  * under the License.
  */
 import { ensureIsArray, t } from '@superset-ui/core';
-import AntdSelect, { LabeledValue as AntdLabeledValue } from 'antd/lib/select';
+import { Select } from 'antd-v5';
```
geido, line 20: Should `Select.Option` be exported from the `Select` component itself in our wrapper?

superset-frontend/src/components/Select/utils.tsx
```diff
@@ -104,12 +124,16 @@ export const sortSelectedFirstHelper = (
     | RawValue[]
     | AntdLabeledValue
     | AntdLabeledValue[]
-    | undefined,
-) =>
-  selectValue && a.value !== undefined && b.value !== undefined
+    | undefined
+    | null,
```
geido, line 128: I think this was not here on purpose. Referring to my previous comment as well.

superset-frontend/src/explore/components/controls/FilterControl/AdhocFilterEditPopoverSimpleTabContent/AdhocFilterEditPopoverSimpleTabContent.test.tsx
```diff
@@ -544,12 +544,20 @@ describe('AdhocFilterEditPopoverSimpleTabContent Advanced data Type Test', () =>
...
+    userEvent.click(operatorValueField);
 
     await act(async () => {
       userEvent.type(operatorValueField, '{enter}');
     });
 
-    expect(screen.getByText('EQUALS')).toBeInTheDocument();
+    expect(
+      await screen.findByText('Equal to (=)', {
+        selector: '.ant-select-selection-item',
```
geido, line 559: Shouldn't this be using the `antd5` prefix?

superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/EncryptedField.tsx
```diff
@@ -67,19 +67,21 @@ export const EncryptedField = ({
...
-          >
-            <AntdSelect.Option value={CredentialInfoOptions.JsonUpload}>
-              {t('Upload JSON file')}
-            </AntdSelect.Option>
-
-            <AntdSelect.Option value={CredentialInfoOptions.CopyPaste}>
-              {t('Copy and Paste JSON credentials')}
-            </AntdSelect.Option>
-          </AntdSelect>
+            css={{ width: '100%' }}
```
geido, line 72: Let's use the string annotation instead of object for consistency with the rest of the codebase

superset-frontend/src/components/Datasource/DatasourceEditor.test.jsx
msyavuz: Apparently yes, without these two tests fail on ci. Will look further into this

superset-frontend/src/components/Select/Select.tsx
msyavuz: There is a type mismatch between v4 and v5 select value types and option types that needs reconciling with our implementation. In this case i can remove the null check but in others mentioned i will try to resolve it in a better way.

superset-frontend/src/components/Select/styles.tsx
msyavuz: This is one of the two usages of color with Ant Design's Select. This one is basically the divider between `Select all` and other options. I will check the other one to see if we can get rid of it.

superset-frontend/src/components/Select/utils.tsx
msyavuz: Yes, done!

superset-frontend/src/explore/components/controls/FilterControl/AdhocFilterEditPopoverSimpleTabContent/AdhocFilterEditPopoverSimpleTabContent.test.tsx
msyavuz: No, unit tests don't get the antd5 prefix

superset-frontend/src/features/databases/DatabaseModal/DatabaseConnectionForm/EncryptedField.tsx
msyavuz: Done!

superset-frontend/src/components/Select/AsyncSelect.tsx
```diff
@@ -219,7 +222,9 @@ const AsyncSelect = forwardRef(
       const missingValues: SelectOptionsType = ensureIsArray(selectValue)
         .filter(opt => !hasOption(getValue(opt), selectOptions))
         .map(opt =>
-          isLabeledValue(opt) ? opt : { value: opt, label: String(opt) },
+          isLabeledValue(opt)
+            ? { value: opt.value, label: opt.label }
```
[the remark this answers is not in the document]
msyavuz: We don't want to pass other props to the options such as number ids that cause error on dom parsing in tests.

superset-frontend/src/components/Select/Select.test.tsx
```diff
@@ -110,6 +110,13 @@ const findSelectValue = () =>
 const findAllSelectValues = () =>
   waitFor(() => [...getElementsByClassName('.ant-select-selection-item')]);
 
+// Tag sets classname manually.
+const findTagSelectValue = () =>
+  waitFor(() => getElementByClassName('.antd5-select-selection-item'));
+
+const findAllTagSelectValues = () =>
```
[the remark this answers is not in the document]
msyavuz: CustomTag sets the classname manually to `antd5-select-selection-item` but tests don't get antd5 prefix. So there will be a discrepency either in the browser dom or in tests. I went with the latter.

superset-frontend/src/components/Select/styles.tsx
msyavuz: Done!

superset-frontend/src/components/Select/Select.tsx
```diff
@@ -635,10 +656,9 @@ const Select = forwardRef(
           onSelect={handleOnSelect}
           onClear={handleClear}
           placeholder={placeholder}
-          showSearch={shouldShowSearch}
-          showArrow
           tokenSeparators={tokenSeparators}
           value={selectValue}
+          virtual={false}
```
geido, line 661: Should we be able to control this instead of forcing to false?

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -84,9 +84,9 @@ describe('Charts list', () => {
     it.only('should show the newly added dashboards in a tooltip', () => {
```
geido, line 84: Should this stay at `only` here? Not sure where this is coming from

msyavuz: Left this as only for now to focus on tests crashing because of Select.

superset-frontend/src/components/Select/Select.tsx
msyavuz: It is by default false. I added this just to make sure it was not antd virtualization causing the crashes. I am not sure about letting usages control `virtual` as our Select implementation is a bit complex and virtual might cause some issues

geido: Oh ok let's remove it then and keep the default

msyavuz:
I had to do a ugly workaround for a bug in rc-table which sets removeButton role to "tab". I opened up the issue and the pr for fixing it but i don't know when Ant Design will use the fixed version and when we will upgrade our Ant Design version so here's the link to the issue to track status:
https://github.com/react-component/tabs/issues/809

superset-frontend/src/components/Select/Select.tsx
```diff
@@ -19,44 +19,46 @@
...
+  isEqual as utilsIsEqual,
+  handleFilterOptionHelper,
+  dropDownRenderHelper,
+  sortSelectedFirstHelper,
   getOption,
   isObject,
-  isEqual as utilsIsEqual,
+  getSuffixIcon,
+  sortComparatorWithSearchHelper,
+  VIRTUAL_THRESHOLD,
```
geido, line 61: Should this part of constants? And maybe `SELECT_ALL_VALUE` too?

superset-frontend/src/explore/components/controls/SelectAsyncControl/index.tsx
```diff
@@ -97,10 +100,12 @@ const SelectAsyncControl = ({
         endpoint: dataEndpoint,
       })
         .then(response => {
-          const data = mutator
-            ? mutator(response.json, value)
-            : response.json.result;
-          setOptions(data);
+          if (value) {
```
geido, line 103: @msyavuz not sure if this got missed but it seems we might still ignoring `null` values with this check which should be valid.

geido (approved):
A few nits but looking good overall!

Issue #32351, opened 2025-02-21
[SIP-154] Optional dataset folders


betodealmeida:
### Motivation

It's not uncommon for core datasets to have hundreds or even thousands (!) of columns/metrics. Exploring these datasets is cumbersome, forcing users to know their names beforehand, so they can search for them in the chart builder UI. Even with small datasets, knowing columns and/or metrics are important can be unclear.

### Proposed Change

To address this problem this SIP proposes a worfklow in which metrics and columns in a dataset can be organized into folders (figure 1). The folders are surfaced in the chart builder UI, allowing for quicker and more confident exploration of the metrics and columns (figure 2).

![Image](https://github.com/user-attachments/assets/be4705d8-d6fd-44b3-a5f0-1761328af912)

![Image](https://github.com/user-attachments/assets/18db3839-aa17-4b32-b3d3-2f57e6214e01)

A few clarifications:

- Folders **can** be nested.
- Metrics and columns can only belong to a single folder. If there is a need for having multiple folders with the same metric (say, for different teams), the recommended flow would be to duplicate the dataset and build a different organization in the new dataset.

### New or Changed Public Interfaces

1. New feature flag is needed to enable the functionality
2. UI needs to be updated (see screenshots above)
3. Dataset export needs to be updated to include folders, and import to honor them

No new APIs are needed, but the dataset CRUD API will be extended to support folders.

No new models will be needed, as the structure will be stored as JSON in the dataset model, under a new column called `folders`.

### New dependencies

No new dependencies.

### Migration Plan and Compatibility

The feature is optional, and old datasets wlil work as today (without any custom folders), so there's no need for migrations other than adding a nullable column.

### Rejected Alternatives

An alternative would be to create new datasets with a subset of the metric and columns of a given dataset, but that could lead to the excessive proliferation of datasets, and different datasets could fall out of sync.

rusackas:
This is a super cool advancement! Let's put it up for [discuss], but you have my +1 when the vote comes!

michael-s-molina:
Thank you for the SIP @betodealmeida. Some questions:

- How to bulk add metrics/columns to a folder? This is important given that datasets might contain hundreds of metrics/columns and adding one by one would be really painful.
- Is it possible to edit a folder name or delete the folder? If yes, can you update the screenshots?
- How are metrics/columns that don't belong to any folder displayed in Explore?
- Does search works for folder names?
- What's "Reset all folders to default"? Is it similar to cancel?

[...]

PR #32520, opened 2025-03-05
feat: dataset folders (backend)

betodealmeida:
### SUMMARY

Tentative implementation of the backend for https://github.com/apache/superset/issues/32351.

I ended up doing the simplest thing that works, based on these requirements:

1. We want to reuse existing dataset APIs (there will be no new APIs to manage folders).
2. Order of folders is important, as well as the order of elements within folders.
3. Support for nested folders should be present from the start (new requirement).

I started the implementation with a `Folder` model, and mapped relationships:

- `Folder n:1 SqlaTable`
- `SqlMetric n:1 Folder`
- `TableColumn n:1 Folder`

One problem with this approach is that order is important, so I had to keep track of the order of each folder and each element inside a folder. This required additional columns, and complex bookkeeping when a dataset was updated — did the positions change? Were metrics removed, columns added, folders renamed? Moving the last element in a folder to the first position, for example, would require updating the position of all elements inside the folder.

Additionally, representing nested folders would require additional relationships to be tracked.

I ditched that approach, and instead opted to serialize the folder structure to a new JSON column in the `SqlaTable` model called `folders`. When doing a `GET` request to `/api/v1/dataset/` or `/api/v1/explore/` the response now includes UUIDs for metrics and columns, and has the new attribute `folders`:

```python
{
  ...
  "metrics": [
    {
      "metric_name": "count",
      "uuid": "uuid2",
      ...
    },
    ...
  ],
  "columns": [
    {
      "column_name": "country",
      "uuid": "uuid5",
      ...
    },
    {
      "column_name": "column-not-in-any-folder",
      "uuid": "uuid6",
      ...
    },
    ...
  ],
  "folders": [
    {
      "uuid": "uuid1",
      "name": "My metrics",
      "children": [
        {
          "uuid": "uuid2"
        },
      ],
    },
    {
      "uuid": "uuid3",
      "name": "My columns",
      "children": [
        {
          "uuid": "uuid4",
          "name": "Dimensions",
          "children": [
            {
              "uuid": "uuid5"
            },
          ],
        },
      ],
    },
  ]
}
```

With this solution the frontend can easily build the UI from this response. Note that the payload only includes custom folders, and the `metrics` and `columns` attribute in the response are unmodified (meaning a metric that is present in a folder will still show up under `metrics`). It's up to the frontend to build the existing "Columns" and "Metrics" sections by removing any elements that are present in custom folders. This way we can build the feature progressively by first enhancing the API, and later adding the custom UI.

To organize the metrics and columns into folders, as well as create new folders, the user must edit the dataset (dataset creation doesn't show metrics nor columns). After creating folders and adding metrics/columns to them the user can save the dataset. The `PUT` request will then send a payload that can also be enhanced with the `folders` attribute.

There is no model for folders, since it offered little to no value. Instead, the client simply declare the folder name, UUID, and an optional description, as well as the tree structure. The UUID is used for external bookkeeping, for example, for systems where semantics are defined outside of Superset and periodically synced via the API.

Depends on https://github.com/apache/superset/pull/32680.

N/A

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Pointing out something related here, is that the tree component is likely going to be composed of react nodes (not just the description). Today we have [from memory, components names may not match] MetricLabel, ColumnLabel, CalculatedColumnLabel, and these require props beyond what's in folders (description to make a InfoBubbleTooltip, sql_expression, data type, ...). Assuming we want for a rich tree with the full labels (at least on the left panel in explore), this means we'll still have to lookup the related objects from the API.

Thinking about future frontend development, we'll need some sort of `assembleTreeDataForComponent(apiPayload)` method, and that it'll have to build rich react nodes. One the other side, we'll need something that takes the `AntdTree.treeData` and prepares the POST payload required by the API, this one should be fairly simple I think.

superset/datasets/schemas.py
```diff
@@ -88,6 +88,18 @@ class DatasetMetricsPutSchema(Schema):
...
+class FolderSchema(Schema):
+    uuid = fields.UUID()
+    type = fields.String(
+        required=False,
+        validate=OneOf(["metric", "column", "folder"]),
+    )
+    name = fields.String(required=True, validate=Length(1, 250))
+    description = fields.String(allow_none=True, validate=Length(0, 1000))
+    # folder can contain metrics, columns, and subfolders:
+    children = fields.List(fields.Nested(lambda: FolderSchema()), allow_none=True)
```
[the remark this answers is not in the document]
betodealmeida: We can add this to `validate_folders`

superset/commands/dataset/update.py
```diff
@@ -189,3 +204,60 @@ def _get_duplicates(data: list[dict[str, Any]], key: str) -> list[str]:
...
+    """
+    if not is_feature_enabled("DATASET_FOLDERS"):
+        raise ValidationError("Dataset folders are not enabled")
+
+    existing = {
+        "metric": {metric.uuid: metric.metric_name for metric in metrics},
+        "column": {column.uuid: column.column_name for column in columns},
+    }
+
+    queue: list[tuple[FolderSchema, list[str]]] = [(folder, []) for folder in folders]
```
geido, line 229: Nice!

superset/migrations/versions/2025-03-03_20-52_94e7a3499973_add_folder_table.py
```diff
@@ -0,0 +1,42 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "94e7a3499973"
+down_revision = "74ad1125881c"
+
+
+def upgrade():
+    op.add_column(
+        "tables",
+        sa.Column("folders", JSON, nullable=True),
```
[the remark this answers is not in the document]
sadpandajoe: we now have util functions for adding and dropping columns that we'll want to use.

superset/datasets/schemas.py
```diff
@@ -88,6 +88,18 @@ class DatasetMetricsPutSchema(Schema):
     uuid = fields.UUID(allow_none=True)
 
 
+class FolderSchema(Schema):
+    uuid = fields.UUID()
```
eschutho, line 92: @betodealmeida does this UUID need to be generated on the client if it's required here?

tests/integration_tests/datasets/api_tests.py
```diff
@@ -1563,6 +1563,92 @@ def test_update_dataset_sqlalchemy_error(self, mock_dao_update):
...
+                    "children": [
+                        {
+                            "type": "folder",
+                            "uuid": "b5330233-e323-4157-b767-98b16f00ca93",
+                            "name": "Dimensions",
+                            "children": [
+                                {
+                                    "type": "column",
+                                    "uuid": dataset.columns[1].uuid,
+                                    "name": dataset.columns[1].column_name,
```
eschutho, line 1601: Not a big deal, but when using the api, it seems unnecessary to need the uuid and the name? We should be able to identify it by just the type and uuid?

superset/datasets/schemas.py
betodealmeida: Right, the client needs to generate a UUID when creating folders. They are only used by tools that want to keep track of external folders and Superset folders.

tests/integration_tests/datasets/api_tests.py
betodealmeida: Right, I considered making them optional.

I included them in the `GET` API because it makes it easier to understand the API response, even thought they are redundant. But I think we could make them optional in the `PUT`, and infer types/names from the existing metrics/columns.

michael-s-molina:
Thanks for the PR @betodealmeida. Some thoughts:

> Note that for metrics and columns type and name are redundant (since they're still present in the response payload under metrics and columns, respectively). But to make the API clearer I uncluded them in the response, even if technically only the UUID is needed.

This might be problematic for us given that we have a dataset with 30k metrics and 7k columns and this decision would significantly increase the payload size.

I think we'll need a very compact API response and a very efficient algorithm when changing the order of things considering the volume of metrics/dimensions. cc @justinpark @villebro

mistercrunch:
> given that we have a dataset with 30k metrics and 7k columns

Seems important to **not** store the "unclassified" items in the "folders" serialized payload. Not sure how it's behaving currently as I haven't deeply reviewed the PR, but seems like a good idea to only serialize/store items that have been actively put inside a folder.

betodealmeida:
> > Are you planning to store thousands of columns/metrics inside folders?
> 
> Once the feature is there, we'll probably organize our thousands of metrics/columns in folders by domain. In other words, all metrics/columns will probably reside in one folder.

OK, let me modify it to have only UUIDs.

villebro:
On Michael's point about having thousands of metrics and columns, I think it touches on an important design requirement to always aim for being able to handle considerable scale, even if it's not obvious at first. Some examples:
- We take full use of indexes in the data model, and typically have all logical models in dedicated tables with link tables for efficient RDBMS handling.
- In migrations we've established patterns for paginating updates.
- For listings we paginate whenever possible (both on frontend and API endpoints).
- For the FE bundle, we minimize the bundle size maximally, and split everything into chunks whenever possible.
- We always defer chart rendering to only the visible charts.
- We've invested extensively into reducing FE overrendering.
- For rendering chart queries, we've replaced `sqlparse` with `sqlglot` due to wasteful handling of the query AST.
- etc

In isolation, all of these would likely have seemed like overkill initially, but for each, their absence proved highly damaging as things grew. The point I'm trying to make is I think we need to always consider how something works when taken to the limit, even if initial use cases don't require it.

As Airbnb's requirement would likely be solved using the already battle tested pattern of having dedicated entity tables with efficient indexing + paginated APIs, I feel we should design this from the start with high scalability in mind, so as to make sure we don't have to do significant refactoring to the feature later on.

betodealmeida:
I agree with making it scalable and targeting, and I'm happy to make the API response/request payloads more compact.

But in regards to pagination, I don't think the current dataset API supports pagination for metrics and columns? As far as I can tell even if you have 10k metrics they are all included in the response, and the frontend does the pagination/search, is that correct @michael-s-molina?

michael-s-molina:
> As far as I can tell even if you have 10k metrics they are all included in the response, and the frontend does the pagination/search, is that correct @michael-s-molina?

Yes, the metrics are all included in the response and there's no pagination made by the frontend. It just renders everything without pagination and virtualization. Apart from that, another huge problem of that module is that it sends ALL columns/metrics to server side when saving instead of just sending the delta which is really inefficient. To put this is perspective, it takes more than 30 seconds to load that screen for our biggest dataset and more than a minute to make changes.

mistercrunch:
I think we should decouple the issues around handling datasets with lots of columns from this PR, though it seems it needs attention.

villebro:
> I think we should decouple the issues around handling datasets with lots of columns from this PR, though it seems it needs attention.

Agreed, I don't think changing the current dataset API for related columns and metrics is in scope for this PR. But as we know this is would probably cause a similar problem in folders, I feel like we should try to design this feature to not have the same scalability issue.

michael-s-molina:
> I think we should decouple the issues around handling datasets with lots of columns from this PR, though it seems it needs attention.

@mistercrunch I agree that the previous implementation of dataset editor can be tackled in a separate PR. What I think should be tackled in this PR is to make sure the new feature supports thousands of columns/metrics. In other words, the algorithms and strategies used need to consider this requirement. What we can't do is to make the problem even worse.

betodealmeida:
@michael-s-molina I've made the API more compact, requiring only the UUIDs for metrics and columns.

I can also change the API so that the request payload only needs to contain changes. For example, if we have these folders:

```
folder1 = ["a", "b", "c"]
folder2 = ["d"]
folder3 = ["e", "f"]
folder4 = ["g"]
```

Let's say the user:

1. Moves "f" from "folder3" to "folder1"
2. Deletes "folder2"

Then the payload would be (in pseudocode):

```
folder1 = ["a", "b", "c", "f"]
folder3 = ["e"]
folder4 = []
```

Instead of sending the whole new state. Do you think that would be enough for your use case?

michael-s-molina:
> folder1 = ["a", "b", "c"]
> folder1 = ["a", "b", "c", "f"] <-- added "f"

For example, if we added "f" to folder1, we shouldn't send all other columns in the request. If the folder contains 5k columns, this means that we're sending 4,999 unnecessary columns in the payload. We should probably only send the deltas.

betodealmeida:
> > folder1 = ["a", "b", "c"]
> > folder1 = ["a", "b", "c", "f"] <-- added "f"
> 
> For example, if we added "f" to folder1, we shouldn't send all other columns in the request. If the folder contains 5k columns, this means that we're sending 4,999 unnecessary columns in the payload. We should probably only send the deltas.
> 
> 
> 

But then how would we indicate that a metric or column was removed from a folder?

michael-s-molina:
> But then how would we indicate that a metric or column was removed from a folder?

The delta contains this type of information. It contains only the folders that had any change and the type of change made. Something like:

```
"changed_folders": [
   "uuid": "uuid1",
   "added": ["colum_a", "column_b"],
   "removed": ["colum_c"],
]
```

This is just an example, we can make the payload even more compact.

michael-s-molina:
It's the same principle of Git when only the changes are sent.

betodealmeida:
Thanks, I know how git works.

What I'm asking is, we now need to define a DSL for the payload, to represent deltas. This makes the flow much more complicated for 95-99% of the cases, so I'd much rather have the simple API for the common use cases, and later implement a delta API (maybe using PATCH) for cases like this.

The new API could then be also used for columns and metrics.

michael-s-molina:
>  I'd much rather have the simple API for the common use cases, and later implement a delta API (maybe using PATCH) for cases like this.

My concern is that by using a simple API, this feature will not work or be highly inefficient for companies with big datasets which is our case. How separating the API will work in practice? Are you going to launch the feature just to a subset of companies? If at the end we'll need to support big datasets, why paying the cost twice?

betodealmeida:
> My concern is that by using a simple API, this feature will not work or be highly inefficient for companies with big datasets which is our case. How separating the API will work in practice? Are you going to launch the feature just to a subset of companies? If at the end we'll need to support big datasets, why pay the cost twice?

Well, what I'm advocating is that we solve the problem we have today — where updating a dataset with tens of thousands of columsn takes too long — in a separate PR. The API is already slow, so I don't see the point in optimizing the folder updates when we would still be sending all the columns and all the metrics when we do an update.

Here's my proposal: it seems like we need a new `PATCH` API for dataset updates. We should add the endpoint, and use a standardized DSL for deltas, like [JSON Patch](https://en.wikipedia.org/wiki/JSON_Patch), which has Python and npm implementations. It would be used for updateing columns, metrics, and folders., and the frontend would be updated to use it, since it's important to be responsive in an interactive environment like the browser.

But we should keep the original `PUT` API for updates that exists today (and with the changes in this PR) because (1) it preserves backwards compatibility, and because (2) it's useful for non-interactive clients where response time is irrelevant. This way, someone who wants to sync datasets from dbt or Cube could use the much simpler `PUT` API, where the whole payload can be sent at once.

So what I'm asking is: let's merge this PR, and I'll work on the `PATCH` endpoint that will solve all the problems.

michael-s-molina:
@betodealmeida I think the important part of the discussion is that we're adding a new feature to Superset and that feature needs to work with our real data which includes datasets with thousands of columns/metrics. Like @villebro said above:

> it touches on an important design requirement to always aim for being able to handle considerable scale, even if it's not obvious at first.
> The point I'm trying to make is I think we need to always consider how something works when taken to the limit, even if initial use cases don't require it.

I suggested the delta approach as a possible solution but you might come up with another one that also works and that would be great too.

michael-s-molina:
> So what I'm asking is: let's merge this PR, and I'll work on the PATCH endpoint that will solve all the problems.

@betodealmeida I'm ok with this approach if by the time we launch the folders feature in an official release (5.1?) it uses the PATCH endpoint 👍🏼 

I appreciate your work on the feature and I think it will bring a lot of value. I know my comments make your job harder but I'm just trying to make sure the principles that @villebro mentioned above are followed and that my org can use it.

mistercrunch:
Resolved the merge conflict to push this forward

mistercrunch:
Fixed gnarly merge conflict while carrying master through https://github.com/apache/superset/pull/32534 and https://github.com/apache/superset/pull/32520 🤞 🤞

mistercrunch:
NOTE: ran into something a little funky where both this PR and #32680 were running the same db migration with different names. Removed the one here as a duplicate of the other.

mistercrunch:
I think I'm mostly through. Great code and TIL about unpacking with `*[]`. So new that both AI and I were a bit confused about it :)

eschutho (approved):

superset-frontend/package-lock.json
```diff
@@ -50863,7 +50863,7 @@
       "version": "0.20.3",
       "license": "Apache-2.0",
       "dependencies": {
-        "@types/react-redux": "^7.1.10",
+        "@types/react-redux": "^7.1.34",
```
mistercrunch, line 50866: a [desired AFAIK] artifact from `docker-compose up` running `npm i`, I think we should merge this along with any PR

PR #32538, opened 2025-03-07
fix(migrations): Handle comparator None in old time comparison migration

Antonio-RiveroMartnez:
### SUMMARY
The time comparison migration doesn't handle slices that have comparator set as None, causing the migration to fail. This PR fixes that behavior

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/versions/2024-05-10_18-02_f84fde59123a_update_charts_with_old_time_comparison.py
+++ b/superset/migrations/versions/2024-05-10_18-02_f84fde59123a_update_charts_with_old_time_comparison.py
@@ -85,7 +85,7 @@ def upgrade_comparison_params(slice_params: dict[str, Any]) -> dict[str, Any]:
     # Adjust adhoc_custom
     if "adhoc_custom" in params and params["adhoc_custom"]:
         adhoc = params["adhoc_custom"][0]  # As there's always only one element
-        if adhoc["comparator"] != "No filter":
+        if adhoc["comparator"] and adhoc["comparator"] != "No filter":
             # Set start_date_offset in params, not in adhoc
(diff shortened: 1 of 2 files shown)


Vitor-Avila (approved):
LGTM!

superset/migrations/versions/2024-05-10_18-02_f84fde59123a_update_charts_with_old_time_comparison.py
```diff
@@ -85,7 +85,7 @@ def upgrade_comparison_params(slice_params: dict[str, Any]) -> dict[str, Any]:
     # Adjust adhoc_custom
     if "adhoc_custom" in params and params["adhoc_custom"]:
         adhoc = params["adhoc_custom"][0]  # As there's always only one element
-        if adhoc["comparator"] != "No filter":
```
betodealmeida, line 88: I'm confused, why would this fail if the value is `None`?

```python
>>> None != "No filter"
True
```

Antonio-RiveroMartnez: it would enter this conditional and `start_date_offset.strftime("%Y-%m-%d")` would fail since get_since_until(adhoc["comparator"]) would return None as the start_date

betodealmeida: Ah, I see, the problem is down the line. Makes sense, thanks!

betodealmeida (approved):

PR #32546, opened 2025-03-07
feat: `OAuth2StoreTokenCommand`

betodealmeida:
### SUMMARY

Add a command to store OAuth2 tokens, to ensure we have a proper `@transaction` and we're not doing a implicit commit.

N/A

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/databases/api.py
+++ b/superset/databases/api.py
@@ -19,7 +19,7 @@
 from __future__ import annotations
 
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime
 from io import BytesIO
 from typing import Any, cast
 from zipfile import is_zipfile, ZipFile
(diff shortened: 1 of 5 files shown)


superset/databases/api.py
```diff
@@ -1433,51 +1433,15 @@ def oauth2(self) -> FlaskResponse:
...
+        command.run()
 
-        if "error" in parameters:
-            raise OAuth2Error(parameters["error"])
-
-        # note that when decoding the state we will perform JWT validation, preventing a
-        # malicious payload that would insert a bogus database token, or delete an
-        # existing one.
         state = decode_oauth2_state(parameters["state"])
+        tab_id = state["tab_id"]
```
Vitor-Avila, line 1440: I see we're currently calling `decode_oauth2_state` both here at the API layer and also at the command. Should the `tab_id` be returned by the command directly? Not super opinionated here, just wondering your thoughts.

Vitor-Avila, line 1440: I see we're currently calling `decode_oauth2_state` both here at the API layer and also at the command. Should the `tab_id` be returned by the command directly? Not super opinionated here, just wondering your thoughts.

Vitor-Avila (approved):
LGTM! Left non-blocking comments

superset/databases/api.py
betodealmeida: Yeah, I agree this is not ideal. But I made the command return the created model instance, and unfortunately it doesn't have `tab_id` associated with it. I thought it was better to keep the interface simple, and have a bit of duplicity here.

Vitor-Avila: that make sense. thank you!

PR #32548, opened 2025-03-07
feat(explore): Integrate dataset panel with Folders feature

kgabryje:
### SUMMARY

First Stage of [Folders SIP](https://github.com/apache/superset/issues/32351) Implementation

This PR introduces dataset folder handling in Explore’s Dataset panel.

#### Key Features:
1. Supports flat and nested folders, as well as folders mixed with columns and metrics.
2. Removes the previous 50-item display limit and associated “show more” buttons, as the list is now virtualized.
3. Ensures backward compatibility: if no folders are defined, columns and metrics appear in default “Metrics” and “Columns” folders, maintaining the existing behavior.

#### Related Work

This builds on @betodealmeida’s [backend PR](https://github.com/apache/superset/pull/32520). While UI support for setting folders is not yet implemented, folders can be configured via the API.

Screenshot 2025-03-07 at 17 54 32

### TESTING INSTRUCTIONS
1. Open Explore
2. The metrics and columns should be displayed exactly as before
3. Send a PUT request to /api/v1/dataset to create folders (see https://github.com/apache/superset/pull/32520)
5. Verify that the folders and their content are rendered in Explore correctly

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


kgabryje:
Converting it to draft as @betodealmeida's PR is not merged yet, but the code is reviewable

superset-frontend/src/explore/components/DatasourcePanel/transformDatasourceFolders.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    mockColumns,
+    folderConfig,
+  );
+
+  expect(result[0].id).toBe('empty-folder');
+  expect(result[0].name).toBe('Empty Folder');
+  expect(result[0].items).toHaveLength(0);
+});
+
+test('handles non-existent metric and column UUIDs in folder config', () => {
```
betodealmeida, line 167: It's great that you added this test!

betodealmeida (approved):
This looks great!

superset-frontend/src/explore/components/DatasourcePanel/transformDatasourceFolders.ts
```diff
@@ -0,0 +1,155 @@
...
+          }
+        } else if (child.type === 'column') {
+          const column = columnsMap.get(child.uuid);
+          if (column) {
+            folder.items.push(column);
+            assignedColumnUuids.add(column.uuid);
+          }
+        }
+      });
+    }
```
justinpark, line 84: Instead of keeping track of `assignedMetricUuids` separately, I think we can assign them from the existing metricsMap to folder items, and after that, delete them. This way, the metricsMap will only contain unassigned metrics, allowing us to skip this additional process.

```suggestion
      datasourceFolder.children.forEach(child => {
        if (child.type === 'folder') {
          folder.subFolders!.push(
            processFolder(child as DatasourceFolder, folder.id),
          );
        } else if (child.type === 'metric') {
          const metric = metricsMap.get(child.uuid);
          if (metric) {
            folder.items.push(metric);
            metricsMap.delete(metric.uuid);
          }
        } else if (child.type === 'column') {
          const column = columnsMap.get(child.uuid);
          if (column) {
            folder.items.push(column);
            columnsMap.delete(column.uuid);
          }
        }
      });
    }
```

superset-frontend/src/explore/components/DatasourcePanel/DatasourceItems.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }, [folders]);
+
+  const { flattenedItems, folderMap } = useMemo(
+    () => flattenFolderStructure(folderStructure),
+    [folderStructure],
+  );
+
+  const handleToggleCollapse = useCallback((folderId: string) => {
+    setFolderStructure(prevFolders => {
+      const updatedFolders = cloneDeep(prevFolders);
```
justinpark, line 104: In the case of Airbnb, when there are over 50k items, using cloneDeep can significantly **impact performance negatively**. Instead of modifying flatterenItems, it's much more efficient to store the collapsed folder IDs in a map when clicked, pass that along with itemData, and then determine whether the collapsed ID is included in the DatasourcePanelItem.

superset-frontend/src/explore/components/DatasourcePanel/DatasourcePanelItem.tsx
```diff
@@ -123,146 +76,95 @@ const SectionHeaderButton = styled.button`
...
-          `}
-        >
-          <div
-            className="field-length"
-            css={css`
-              flex-shrink: 0;
-            `}
-          >
-            {isColumnSection
-              ? t(`Showing %s of %s`, columnSlice?.length, totalColumns)
```
justinpark, line 227: Removing the display of total counts would be a significant regression for Airbnb users. I request that at least the count for items not included in the folder be restored.

kgabryje: @justinpark If I understand correctly, this line was displayed because we limited the initially displayed items to 50. I removed that because I don't think it had any value given that we use virtualization.
Could you explain what do you mean by `the count for items not included in the folder`? Would a simple items count for each folder be sufficient for your users?

justinpark: We discussed separately offline, and for the folder items, we agreed to display the number of items included in the folder.

PR #32549, opened 2025-03-08
fix(dashboard): Support bigint value in native filters

justinpark:
### SUMMARY

When the dataset values of the native filter include bigint values, an error occurs in the filter selection rendering.
This happens because the value is converted to a BigNumber class implemented by json-bigint and passed as a React child. To resolve this issue, this commit enables an option to convert the values to native int(`useNativeBigInt`), which allows them to be passed as compatible bigint to the children, thus solving the problem.

Before:

Screenshot 2025-03-07 at 2 20 49 PM

After:

Screenshot 2025-03-07 at 4 58 54 PM

### TESTING INSTRUCTIONS

Create a dataset from the following SQL
```
select 1234567890123456789 as bigIntVal
```
Choose the dataset in the filter, and then select the input box to check if the option is rendered correctly

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/time-format/formatters/finestTemporalGrain.ts
+++ b/superset-frontend/packages/superset-ui-core/src/time-format/formatters/finestTemporalGrain.ts
@@ -48,30 +48,30 @@ export default function finestTemporalGrain(
   } = useLocalTime ? localTimeUtils : utcUtils;
 
   let formatFunc = formatYear;
-  try {
-    values.forEach((value: any) => {
-      if (formatFunc === formatYear && isNotFirstMonth(value)) {
-        formatFunc = formatMonth;
-      }
-      if (formatFunc === formatMonth && isNotFirstDayOfMonth(value)) {
-        formatFunc = formatDay;
-      }
-      if (formatFunc === formatDay && hasHour(value)) {
-        formatFunc = formatHour;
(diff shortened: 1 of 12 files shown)


superset-frontend/packages/superset-ui-core/src/time-format/formatters/finestTemporalGrain.ts
```diff
@@ -48,26 +48,30 @@ export default function finestTemporalGrain(
...
-    }
-    if (formatFunc === formatMinute && hasSecond(value)) {
-      formatFunc = formatSecond;
-    }
-    if (formatFunc === formatSecond && hasMillisecond(value)) {
-      formatFunc = formatMillisecond;
-    }
-  });
+  try {
+    values.forEach((value: any) => {
```
michael-s-molina, line 52: Why do we need a try/catch to assign `formatFunc`?

justinpark: revert the try-catch block and add the bigint type check to clarify the issue (Cannot convert a BigInt value to a number in Date function)

michael-s-molina (approved):

PR #32553, opened 2025-03-09
fix(comp/async-ace-editor): proper import of `ace-builds`

hainenber:
fix(comp/async-ace-editor): proper import of `ace-builds`

### SUMMARY

Fixed a borked Storybook component whilst overall functionality is kept, I believe. I. still think someone with deeper knowledge on the Editor can check for impacts.

From what I checked, there is no `acequire` reference in the [source code](https://github.com/ajaxorg/ace-builds/blob/v1.37.5/src-min-noconflict/ace.js).

image

Before
image

After
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/AsyncAceEditor/index.tsx
+++ b/superset-frontend/src/components/AsyncAceEditor/index.tsx
@@ -124,7 +124,7 @@ export default function AsyncAceEditor(
       { default: ReactAceEditor },
       { config },
       { default: cssWorkerUrl },
-      { acequire },
+      { require: acequire },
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):
This looks ok to me, especially if you inspected the payload of `acequirePromise` and found the type expectation match now. Though I'm wondering if there's a way to confirm things by importing the right typescript types from the package (?) Wondering if the package or some other related package exposes types here.

PR #32571, opened 2025-03-10
fix(no-restricted-imports): Fix overrides and include no-fa-icons-usage

geido:
### SUMMARY
This PR does the following:
- Fixes the `no-restricted-imports` rule that was broken
- Adjusts the overrides so that the restricted imports trigger in the right places
- Includes a new `no-fa-icons-usage` to stop the usage of `fa` icons

We should merge this PR now to stop the bleeding. For this reason we'll skip all rules that are not respected marking the with a TODO, such as:

`TODO: Use src/components/Icons`
`TODO: Remove fa-icons`
`TODO: Remove antd`
`TODO: DO not import theme directly`

N.A.

### TESTING INSTRUCTIONS
CI should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
I think you can just use the `no-restricted-imports` feature if you want to keep it simpler.

rusackas (approved):
LGTM! I wonder how far off we are from deleting font-awesome.

rusackas:
Looks like it's finding lots of stuff... hooray! Better to set it to Warn for now, or go through and fix all the things? Hoping we don't have to sweep them under the rug with eslint-ignore :P

superset-frontend/.eslintrc.js
```diff
@@ -292,6 +386,7 @@ module.exports = {
   ],
   rules: {
     'theme-colors/no-literal-colors': 'error',
+    'icons/no-fa-icons-usage': 'warn',
```
mistercrunch, line 389: Personally would much rather have a proper `error`, combinated with `# NOQA` type statements for all of the entries we need to fix. Stops the bleeding in its track, and it's pretty easy to `grep` for the related NOQA statements.

geido:
> I think you can just use the `no-restricted-imports` feature if you want to keep it simpler.

For reference, this seemed to be broken in master. The PR should fix it.

michael-s-molina (approved):

PR #32572, opened 2025-03-10
fix: Log table retention policy

michael-s-molina:
### SUMMARY
This PR fixes a problem with the `log` table retention policy. Currently, there's no way to limit the number of records retained in that table which leads to problems in Superset when the table contains millions of rows. This table is queried from Superset's Welcome page to get recent modified items and not being able to set a retention policy is a critical error. 

To keep backward compatibility, this PR does not change the current retention policy which is to preserve all records. Adding a default retention policy would be a good practice that we can do for 6.0.

### TESTING INSTRUCTIONS
Configure the celery task called `prune_logs` and check that the records are deleted according to the `retention_days` configuration.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/logs/prune.py
+++ b/superset/commands/logs/prune.py
@@ -1,0 +1,109 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import logging
+import time
+from datetime import datetime, timedelta
+
+import sqlalchemy as sa
+
+from superset import db
+from superset.commands.base import BaseCommand
+from superset.models.core import Log
+
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=consider-using-transaction
+class LogPruneCommand(BaseCommand):
+    """
+    Command to prune the logs table by deleting rows older than the specified retention period.
+
+    This command deletes records from the `Log` table that have not been changed within the
+    specified number of days. It helps in maintaining the database by removing outdated entries
+    and freeing up space.
+
+    Attributes:
+        retention_period_days (int): The number of days for which records should be retained.
+                                     Records older than this period will be deleted.
+    """  # noqa: E501
+
+    def __init__(self, retention_period_days: int):
+        """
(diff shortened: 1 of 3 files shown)


justinpark (approved):

landryb:
@michael-s-molina a bit puzzled, doing a pip install from master branch here doesn't install the new `superset/commands/logs` subdir, is there something needed in the build system/python goo so that the `prune.py` file gets installed ?

without it, trigerring a job on a celery worker blows with:
```
Apr 09 10:07:02 demo celery[4084022]:   File "/srv/apps/superset/venv/lib/python3.11/site-packages/superset/tasks/scheduler.py", line 28, in <module>
Apr 09 10:07:02 demo celery[4084022]:     from superset.commands.logs.prune import LogPruneCommand
Apr 09 10:07:02 demo celery[4084022]: ModuleNotFoundError: No module named 'superset.commands.logs'
```

michael-s-molina:
> is there something needed in the build system/python goo so that the prune.py file gets installed ?

Nothing that I'm aware of.

landryb:
weird, probably something bogus on my side then, or a wrong invocation via ansible.. but that's reproducible in a new empty virtualenv:
```
$ ansible mygeorchestra -i host -u root -m pip -a 'name="git+https://github.com/apache/superset@6b7394e78998#egg=apache-superset" virtualenv=/usr/venv virtualenv_site_packages=true'
...
...
        "Building wheels for collected packages: apache-superset",
        "  Building wheel for apache-superset (pyproject.toml): started",
        "  Building wheel for apache-superset (pyproject.toml): finished with status 'done'",
        "  Created wheel for apache-superset: filename=apache_superset-0.0.0.dev0-py3-none-any.whl size=4617711 sha256=9c6c1087e3a9ce1b309c8632691511c3ab1a543100794ab3fdf2965f6eceee88",
        "  Stored in directory: /tmp/pip-ephem-wheel-cache-1h96rixi/wheels/a7/4e/65/283b7f2fcde75ded0fb99b186d00bf35378e26be358d2ca9ab",
        "Successfully built apache-superset",
```
and on the target venv where the resulting wheel is installed, the dir is missing:
```
[09/04 13:28] root@demo.fluela:/srv/apps/superset $ls -d /usr/venv/lib/python3.11/site-packages/superset/commands/logs
ls: cannot access '/usr/venv/lib/python3.11/site-packages/superset/commands/logs': No such file or directory
```

landryb:
and same thing, doing a `pip wheel .` from a clone builds a wheel that doesn't contain the file:
```
/data/src/georchestra/superset-core $pip wheel --wheel-dir=/tmp/wheel .
/data/src/georchestra/superset-core $unzip -l /tmp/wheel/apache_superset-0.0.0.dev0-py3-none-any.whl |grep /prune
```
the last command yields nothing (and ofc all the other files are here)

michael-s-molina:
@landryb I downloaded the [5.0.0rc2 Pypi](https://pypi.org/project/apache-superset/5.0.0rc2/) package and that folder is indeed not there 🤔 Could you open an issue for this and tag me? I'll add it to the [5.0.0 board](https://github.com/orgs/apache/projects/463).

michael-s-molina:
@landryb I discovered why the folder was missing. It was because the folder was missing the `__init__.py` file. I provided a fix in https://github.com/apache/superset/pull/33059.

PR #32585, opened 2025-03-11
chore(reports): add task for slack channels warm-up

Usiel:
### SUMMARY

Adds a celery task to warm up the Slack channel cache and allows admins to schedule the same task (with example). This is helpful for cases where retrieval of the full channel list takes minutes.

#### Changes to the `get_channels` function

The Slack conversations API is not necessarily faster when using the `types` filter: For example, when we filter for private channels, the cursor will still iterate over all public channels and the number of requests remains the same, only the data transmitted is less.

With this in mind, I believe it is more efficient for us to cache only a single value for `get_channels()` and avoid having multiple cache values for each possible value of `types`. Doing the filtering in `get_channels_with_search` on demand doesn't add much cost.

Additionally, the `limit` param is also not relevant for the cache key and is inlined.

... and, of course, this change makes the celery task more sensible :)

#### Warning

There is a minor API change in this PR: Previously, we followed Slack's [conversations.list](https://api.slack.com/methods/conversations.list) convention, where empty an `types` param means that only public channels are returned. After this PR we will return all channel types when `types` is empty or null. To me that feels more logical, but if people feel we should stick to previous approach I'm OK with changing this.

### TESTING INSTRUCTIONS

Pre:
- Set `SLACK_API_TOKEN`
- Enable `ALERT_REPORTS`

#### Test Celery Task

```
celery --app=superset.tasks.celery_app:app call slack.cache_channels
```

**Expected**: Task is launched and result is cached.

#### Test Beat Schedule Example

1. Uncomment the beat schedule entry for `slack.cache_channels`
2. [Optional]: Adapt schedule to trigger every minute (or set time)

**Expected**: Task is scheduled and result is cached.

#### Test REST API

Try for `TYPES={"public_channel", "private_channel", "", "public_channel, private_channel"}`

```
curl -X 'GET' \
  'http://localhost:8088/api/v1/report/slack_channels/?q=(types%3A!(<TYPES>))' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <TOKEN>'
```

**Expected**: Cached value is used, filtered and returned.

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/32480
- [ ] Required feature flags:
- [ ] Changes UI

cc @Vitor-Avila since this builds on your recent work in https://github.com/apache/superset/pull/32529

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -1494,6 +1494,7 @@ EMAIL_REPORTS_CTA = "Explore in Superset"
 # Slack API token for the superset reports, either string or callable
 SLACK_API_TOKEN: Callable[[], str] | str | None = None
 SLACK_PROXY = None
+SLACK_CACHE_TIMEOUT = int(timedelta(days=1).total_seconds())
 
 # The webdriver to use for generating reports. Use one of the following
 # firefox
--- a/superset/config.py
+++ b/superset/config.py
@@ -1015,6 +1015,7 @@ class CeleryConfig:  # pylint: disable=too-few-public-methods
         "superset.tasks.scheduler",
         "superset.tasks.thumbnails",
         "superset.tasks.cache",
+        "superset.tasks.slack",
     )
     result_backend = "db+sqlite:///celery_results.sqlite"
     worker_prefetch_multiplier = 1
@@ -1045,6 +1046,11 @@ class CeleryConfig:  # pylint: disable=too-few-public-methods
         #     "task": "prune_logs",
         #     "schedule": crontab(minute="*", hour="*"),
         #     "kwargs": {"retention_period_days": 180},
+        # },
+        # Uncomment to enable Slack channel cache warm-up
+        # "slack.cache_channels": {
+        #     "task": "slack.cache_channels",
+        #     "schedule": crontab(minute="0", hour="*"),
         # },
     }
 
--- a/superset/tasks/slack.py
+++ b/superset/tasks/slack.py
@@ -26,4 +26,10 @@ logger = logging.getLogger(__name__)
 
 @celery_app.task(name="slack.cache_channels")
 def cache_channels() -> None:
-    get_channels(force=True, cache_timeout=current_app.config["SLACK_CACHE_TIMEOUT"])
+    try:
+        get_channels(
+            force=True, cache_timeout=current_app.config["SLACK_CACHE_TIMEOUT"]
+        )
+    except Exception as ex:
+        logger.exception("An error occurred while caching Slack channels: %s", ex)
+        raise
--- a/superset/tasks/slack.py
+++ b/superset/tasks/slack.py
@@ -16,6 +16,8 @@
 # under the License.
 import logging
 
+from flask import current_app
+
 from superset.extensions import celery_app
 from superset.utils.slack import get_channels
 
@@ -24,4 +26,4 @@ logger = logging.getLogger(__name__)
 
 @celery_app.task(name="slack.cache_channels")
 def cache_channels() -> None:
-    get_channels(force=True)
+    get_channels(force=True, cache_timeout=current_app.config["SLACK_CACHE_TIMEOUT"])
--- a/superset/tasks/slack.py
+++ b/superset/tasks/slack.py
@@ -1,0 +1,27 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import logging
+
+from superset.extensions import celery_app
+from superset.utils.slack import get_channels
+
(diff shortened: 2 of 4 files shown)


superset/utils/slack.py
```diff
@@ -99,18 +99,25 @@ def get_channels_with_search(
...
-            limit=limit,
-            extra_params=extra_params,
             force=force,
             cache_timeout=86400,
         )
     except (SlackClientError, SlackApiError) as ex:
         raise SupersetException(f"Failed to list channels: {ex}") from ex
 
+    if types and not len(types) == len(SlackChannelTypes):
+        conditions: list[Callable[[SlackChannelSchema], bool]] = []
```
Usiel, line 111: The intention here was to make this more easily pluggable in the future if we add more types.

superset/tasks/slack.py
```diff
@@ -0,0 +1,27 @@
...
+
+from superset.extensions import celery_app
+from superset.utils.slack import get_channels
+
+logger = logging.getLogger(__name__)
+
+
+@celery_app.task(name="slack.cache_channels")
+def cache_channels() -> None:
+    get_channels(force=True)
```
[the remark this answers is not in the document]
Usiel: The Slack client handles both connection and (because we added it) rate-limit errors.

superset/utils/slack.py
```diff
@@ -99,18 +99,25 @@ def get_channels_with_search(
...
     """
-    extra_params = {}
-    extra_params["types"] = ",".join(types) if types else None
     try:
         channels = get_channels(
-            limit=limit,
-            extra_params=extra_params,
             force=force,
             cache_timeout=86400,
         )
```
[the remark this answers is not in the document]
Usiel: The parameters are from utils.cache.memoized_func

superset/config.py
```diff
@@ -1046,6 +1047,11 @@ class CeleryConfig:  # pylint: disable=too-few-public-methods
         #     "schedule": crontab(minute="*", hour="*"),
         #     "kwargs": {"retention_period_days": 180},
         # },
+        # Uncomment to enable Slack channel cache warm-up
+        # "slack.cache_channels": {
+        #     "task": "slack.cache_channels",
+        #     "schedule": crontab(minute="0", hour="*"),
+        # },
```
Vitor-Avila, line 1054: Nice!

superset/tasks/slack.py
Vitor-Avila, line 27: Could we enforce calling it with a `cache_timeout` as well? Just to prevent creating cache items that don't have an expiration (since it's a scheduled task it would always refresh on schedule, but there's an edge case of a user disabling the task and then it would be there until someone refreshes through the UI which has a `cache_timeout`).

I was actually wondering if we should have a new config like `SLACK_CACHE_TIMEOUT` (since this is hardcoded in the code, keeping it in `config.py` might make more sense). This was supposed to be part of my PR, but just thought about it after having it merged so was planning to do a follow up in the future. If we have a config, we could use it here + in the API config.

Vitor-Avila: Also, since this is calling `get_channels()` directly, could we do it in a `try/except` block and log exceptions? Raising an error wouldn't make much sense as this is async, but I think logging is always a good thing.

superset/utils/slack.py
Vitor-Avila: Nice!

superset/utils/slack.py
```diff
@@ -71,11 +71,12 @@ def get_channels(limit: int, extra_params: dict[str, Any]) -> list[SlackChannelS
     client = get_slack_client()
     channel_schema = SlackChannelSchema()
     channels: list[SlackChannelSchema] = []
+    extra_params = {"types": ",".join(SlackChannelTypes)}
     cursor = None
 
     while True:
         response = client.conversations_list(
-            limit=limit, cursor=cursor, exclude_archived=True, **extra_params
+            limit=999, cursor=cursor, exclude_archived=True, **extra_params
```
sadpandajoe, line 79: why are we hardcoding this now?

Usiel: I inlined the `limit` because it never governed the amount of channels returned by the function. The function always returns all channels. The `limit` parameter here only governs the amount of channels returned by each Slack API response. 

Going for the maximum (999) allowed by Slack is most efficient, imo. Lower values are more likely to result in rate-limiting (which is based on requests, not rows returned or similar).

Usiel: If we see a need we could add paging to our `/slack_channels` endpoint (and `get_channels_with_search(...)`). I don't know about that though, our only usage is to get all channels to give the user a choice for the report notification setting.

superset/tasks/slack.py
Usiel: Makes sense! Done and done.

Vitor-Avila (approved):
LGTM!

superset/tasks/slack.py
```diff
@@ -0,0 +1,35 @@
...
+from superset.utils.slack import get_channels
+
+logger = logging.getLogger(__name__)
+
+
+@celery_app.task(name="slack.cache_channels")
+def cache_channels() -> None:
+    try:
+        get_channels(
+            force=True, cache_timeout=current_app.config["SLACK_CACHE_TIMEOUT"]
```
eschutho, line 31: @dpgaspar can you confirm if this task would work properly for fetching the slack token (when it is a callable) for multi-tenant applications? Do we need to pass in some user context to this function?

dpgaspar: it should just work

rusackas:
Looks good to merge?

Vitor-Avila:
@rusackas my only question here is relating to the API change. While this API endpoint is solely used by the Slack integration (I can't think why a user would create any workflow on top of a Superset API endpoint that just lists Slack channels) I wonder if we need to mention this in `UPDATING.md` and if it has to be considered a breaking change or not.

PR #32589, opened 2025-03-11
docs: Update pypi.mdx

felipegarciaign:
sometimes, better is to use this:

superset run -p 8088 -h 0.0.0.0 --with-threads --reload --debugger

### DIFF
--- a/docs/docs/installation/pypi.mdx
+++ b/docs/docs/installation/pypi.mdx
@@ -157,6 +157,8 @@ superset init
 # To start a development web server on port 8088, use -p to bind to another port
 superset run -p 8088 --with-threads --reload --debugger
 ```
+### sometimes better is to use:
+#  superset run -p 8088 -h 0.0.0.0 --with-threads --reload --debugger
 
 If everything worked, you should be able to navigate to `hostname:port` in your browser (e.g.
 locally by default at `localhost:8088`) and login using the username and password you created.


sfirke:
Hello, thanks for the suggestion! Can you explain why someone would want to use that `-h` flag, so that readers know?

sfirke (changes requested):
Needs to explain what `-h  0.0.0.0` accomplishes and when/why a user should run that command.

docs/docs/installation/pypi.mdx
felipegarciaign, line 1: In the context of Apache Superset (VM deployed with Python & url HTTP), the -h 0.0.0.0 command is used to bind the Superset server to all available network interfaces on the host machine. Installin on a VM your navigator never reach Superset welcome page html:

What it does:
Binding to all interfaces: By specifying 0.0.0.0, the server listens for incoming connections on all network interfaces of the machine, not just localhost (127.0.0.1).

Accessibility: This makes the Superset instance accessible from other devices on the same network, rather than being restricted to the local machine.

When/Why to use it:
Development and Testing: If you're developing or testing Superset and need to access it from another device (e.g., a tablet or another computer on the same network), this command is useful.

Collaborative Work: When working in a team, you might want to share access to your Superset instance with colleagues on the same network.

Demonstrations: For presenting dashboards or features to others in a local environment.

However, for production environments, it's important to secure the server properly (e.g., using firewalls, HTTPS, and authentication) to prevent unauthorized access.

hainenber: I believe this should be added into the PR and not only to reply to Sam's ask.

hainenber:
This and #32736 is pretty much duplicate but I'll close this one since the other one has more details.

PR #32352, opened 2025-02-21
fix(dev/ci): pre-commit fixes galore 

rusackas:
### SUMMARY

When running `pre-commit run --all-files` I was getting a TON of console noise, TS errors, lint errors, etc. This should tidy things up significantly, fixing all the errors, improving the eslint process for docs, and more.

It also updates a couple of newer pre-commit hooks/linters to newer version, and makes changes accordingly.

One migration caught a hypothetical security issue, and that was addressed as well.

Frontend pre-commit jobs (prettier/eslint) now write fixes when run locally, so you can re-run and pass, but do not do this on CI, so they appropriately fail.

### TESTING INSTRUCTIONS

Let's see if CI passes!

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #32596, opened 2025-03-11
chore: fix precommit for eslint

mistercrunch:
Trying to fix prettier and eslint in pre-commit so it only runs on altered files.

@rusackas this works for me, but is largely a rollback of https://github.com/apache/superset/pull/32352 in some key areas. Wondering what's wrong here.

Changes:
- skipping dups in CI (eslint and tsc checks run in another GHA)
- using the prettier action repo as opposed to a custom script
- eslint back on using `scripts/eslint.sh`

If this doesn't work in some cases I can try and address the core issues.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -56,8 +56,7 @@ jobs:
       - name: pre-commit
         run: |
           set +e  # Don't exit immediately on failure
-          # Skip auto-fixing in CI to ensure changes are committed locally
-          export SKIP_FIX=1
+          export SKIP=eslint-frontend,type-checking-frontend
           pre-commit run --all-files
           PRE_COMMIT_EXIT_CODE=$?
           git diff --quiet --exit-code
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -70,7 +70,7 @@ repos:
       files: ^superset-frontend/.*\.(js|jsx|ts|tsx)$
     - id: eslint-docs
       name: eslint (docs)
-      entry: bash -c 'cd docs && FILES=$(echo "$@" | sed "s|docs/||g") && yarn eslint --ext .js,.jsx,.ts,.tsx --quiet $([ -z "$SKIP_FIX" ] && echo "--fix") $FILES'
+      entry: bash -c 'cd docs && FILES=$(echo "$@" | sed "s|docs/||g") && yarn eslint --fix --ext .js,.jsx,.ts,.tsx --quiet $FILES'
       language: system
       pass_filenames: true
       files: ^docs/.*\.(js|jsx|ts|tsx)$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,40 +52,35 @@ repos:
       - id: trailing-whitespace
         exclude: ^.*\.(snap)
         args: ["--markdown-linebreak-ext=md"]
-  - repo: local
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v4.0.0-alpha.8 # Use the sha or tag you want to point at
     hooks:
       - id: prettier
-        name: prettier
-        entry: bash -c 'cd superset-frontend && if [ -z "$SKIP_FIX" ]; then npm run format -- --log-level error; else npm run prettier-check; fi'
-        language: system
-        files: '^superset-frontend/(src|spec|cypress-base|plugins|packages|.storybook)/.+\.(js|jsx|ts|tsx|css|less|scss|sass)$|^superset-frontend/package\.json$'
-        pass_filenames: true
-      - id: eslint-frontend
-        name: eslint (frontend)
-        entry: bash -c 'cd superset-frontend && if [ -z "$SKIP_FIX" ]; then npm run lint-fix; else npm run eslint; fi'
-        language: system
-        pass_filenames: true
-        files: ^superset-frontend/.*\.(js|jsx|ts|tsx)$
-# TODO - re-assess using this script-based linting rather than the npm command. Both approaches have their merits
-#       - id: eslint
-#         name: eslint
-#         entry: ./scripts/eslint.sh
-#         language: script
-#         pass_filenames: true
-#         files: ^superset-frontend\/.*\.(js|jsx|ts|tsx)$
-      - id: eslint-docs
-        name: eslint (docs)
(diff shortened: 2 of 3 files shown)


mistercrunch:
@rusackas since you mentioned that this didn't work for you in the past, wondering if you can pull/test this and report back. Also remembering some issues with a user that was using Linux, where it's possible that `$@` and this snippet `"${@//$frontend_dir\//}"` in `scripts/eslint.sh` doesn't behave the same way as in OSX.

I don't have a linux VM laying around to test... Would love if someone can test on Linux.

rusackas:
I'll test and report back asap. Would love to get this speedy again.

.pre-commit-config.yaml
```diff
@@ -52,40 +52,35 @@ repos:
       - id: trailing-whitespace
         exclude: ^.*\.(snap)
         args: ["--markdown-linebreak-ext=md"]
-  - repo: local
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v4.0.0-alpha.8 # Use the sha or tag you want to point at
```
rusackas, line 56: We're using 3.3.3 on the repo. I wonder if that'll lead to trouble.

rusackas: Oh wait, I see it uses prettier 3.3.3 below... but this repo is archived and claims it was broken by newer versions of Prettier. Not sure what to make of that yet.

.pre-commit-config.yaml
```diff
@@ -52,40 +52,35 @@ repos:
...
+    hooks:
+    - id: eslint-frontend
+      name: eslint (frontend)
+      entry: ./scripts/eslint.sh
+      language: system
+      pass_filenames: true
+      files: ^superset-frontend/.*\.(js|jsx|ts|tsx)$
+    - id: eslint-docs
+      name: eslint (docs)
+      entry: bash -c 'cd docs && FILES=$(echo "$@" | sed "s|docs/||g") && yarn eslint --ext .js,.jsx,.ts,.tsx --quiet $([ -z "$SKIP_FIX" ] && echo "--fix") $FILES'
```
rusackas, line 73: I think we're getting rid of SKIP_FIX?

.github/workflows/pre-commit.yml
```diff
@@ -56,8 +56,7 @@ jobs:
       - name: pre-commit
         run: |
           set +e  # Don't exit immediately on failure
-          # Skip auto-fixing in CI to ensure changes are committed locally
-          export SKIP_FIX=1
+          export SKIP=eslint-frontend,type-checking-frontend
```
rusackas, line 59: Is this used anywhere?

.pre-commit-config.yaml
mistercrunch: Yeah that's the ref o the GHA repo. Looks like the latest -> https://github.com/pre-commit/mirrors-prettier/tags

rusackas: Wasn't sure how much to worry about "This repository was archived by the owner on Apr 11, 2024. It is now read-only." - I guess it might be there forever, but someone could go sweeping up one day.

rusackas:
> @rusackas this works for me, but is largely a rollback of https://github.com/apache/superset/pull/32352 in some key areas. Wondering what's wrong here.

Fine by me! It seems to just roll back the parts I wasn't really hoping to touch much in the first place :D 

The things I set out to do which still seem to be in effect here are:
• Run `eslint` with the ` --quiet` flag
• Set up/ensure Node 20
• Bump `mypy` to 1.15.0

With this change, we lose two things, which maybe we can fix here or in subsequent PR(s):
• Reverts typescript checking (which led to 95% of my last PR's contents). I'm sure we can add a new hook for that. It was catching these since `npm run type` slides in as a side-run task of `npm run lint` :) 
• `Prettier` is currently fixing busted files when running the hooks locally, but `eslint` no longer is. I wonder if we should always run with the `--fix` flag, or if that needs to be somehow dynamically disabled on CI and only run locally  ¯\\\_(ツ)_/¯

.pre-commit-config.yaml
mistercrunch: Yeah seems it'd be good hygiene to get rid of this GHA-as-a-submodule and simply `git push` ourselves in our action. I don't think we need an unmaintained reference to a simple GHA here given what we use it for is fairly simple.

.github/workflows/pre-commit.yml
mistercrunch: I think SKIP_FIX was introduced in your previous PR and isn't referenced anywhere else in the repo after this PR.

`SKIP` is used for pre-commit steps that are already handled in other GHAs, in this case those 2 are handled in the `superset-frontend` GHA.

mistercrunch:
> Reverts typescript checking

From my understanding it's still here -> https://github.com/apache/superset/pull/32596/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R77-R83

> Prettier is currently fixing busted files when running the hooks locally, but eslint no longer is

Can add `--fix` in there, super easy. It's probably better that way. Personally before I would run a custom command to try and do it, but it's probably a better behavior to always `--fix`, let me add it to the PR here.

.github/workflows/pre-commit.yml
rusackas: Ahh, ok, I wasn't understanding what SKIP did, but SKIP_FIX being removed here means we can remove it in `pre-commit-config.yml` as well, then.

mistercrunch:
Here, added --fix -> https://github.com/apache/superset/pull/32596/commits/ac7fd7e42e22351b5823dee4f31494931cce7cb4

.pre-commit-config.yaml
rusackas: Ahh, what I meant is that if SKIP_FIX is no longer in `pre-commit.yaml` as of this PR, I think we can remove it from this line.

rusackas:
Thanks for `--fix`ing. 

Once the last remnant of `SKIP_FIX` is gone, you've got my stamp.

I'll tackle TS checking in a separate PR (it'd be nice, but not blocking this)

Up to you if you want to ditch that archived GHA inclusion or not for this PR, or handle that separately. More of a TODO than a blocker.

mistercrunch:
Ok removed the last `SKIP_FIX`  reference that was there for the docs I think. `pre-commit` now always tries to fix what it can, which I think is what everyone wants all the time.

rusackas (approved):

PR #32608, opened 2025-03-11
fix(welcome): perf on distinct recent activities

justinpark:
### SUMMARY
Currently, the `recent_activity` query for the log table groups by the dashboard ID and slice ID to extract a distinct list from the entire log table, which leads to performance issues.
(In the case of Airbnb, more than 1 million logs are generated each day, and grouping by dashboard and slice ID, even with indexing, significantly impacts database performance as shown in the following log) 
```
| 3793593 | superset         | 100.117.121.20:41904  | superset_production | Query       |  1726 s| Sending data                                                  | SELECT anon_1.dashboard_id AS anon_1_dashboard_id, anon_1.slice_id AS anon_1_slice_id, anon_1.action AS anon_1_action, anon_1.dttm AS anon_1_dttm, dashboards.slug AS dashboard_slug, dashboards.dashboard_title AS dashboards_dashboard_title, slices.slice_name AS slices_slice_name
FROM (SELECT logs.dashboard_id AS dashboard_id, logs.slice_id AS slice_id, logs.action AS action, max(logs.dttm) AS dttm
FROM logs
WHERE logs.action IN ('explore', 'dashboard') AND logs.user_id = 13295 AND logs.dttm > '2024-03-06 18:00:12.201653' AND (logs.dashboard_id IS NOT NULL OR logs.slice_id IS NOT NULL) GROUP BY logs.dashboard_id, logs.slice_id, logs.action) AS anon_1 LEFT OUTER JOIN dashboards ON dashboards.id = anon_1.dashboard_id LEFT OUTER JOIN slices ON slices.id = anon_1.slice_id
WHERE dashboards.dashboard_title != '' OR slices.slice_name != '' ORDER BY anon_1.dttm DESC
 LIMIT 0, 6 |
```

To resolve this issue, it would be appropriate to create a materialized activity statistics view through a daily/hourly batch job. However, functionally, the main purpose of recent activity is to display only a few of the most recently visited items. Therefore, we improved performance by changing the approach to fetch the latest (including duplicates) log entries (by `distinct: false`) and extract a distinct list using LRU cache in the frontend side.

### TESTING INSTRUCTIONS
specs (no visual changes)
Screenshot 2025-03-11 at 4 21 02 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/views/CRUD/utils.tsx
+++ b/superset-frontend/src/views/CRUD/utils.tsx
@@ -231,7 +231,7 @@ export const getRecentActivityObjs = (
     return getFilteredChartsandDashboards(addDangerToast, filters).then(
       ({ other }) => {
         res.other = other;
-        res.viewed = distinctRes.entries().reverse();
+        res.viewed = distinctRes.values().reverse();
         return res;
       },
     );
--- a/superset-frontend/src/views/CRUD/utils.tsx
+++ b/superset-frontend/src/views/CRUD/utils.tsx
@@ -26,6 +26,7 @@ import {
   SupersetTheme,
   getClientErrorObject,
   t,
+  lruCache,
 } from '@superset-ui/core';
 import Chart from 'src/types/Chart';
 import { intersection } from 'lodash';
@@ -34,7 +35,7 @@ import { FetchDataConfig, FilterValue } from 'src/components/ListView';
 import SupersetText from 'src/utils/textUtils';
 import { findPermission } from 'src/utils/findPermission';
 import { User } from 'src/types/bootstrapTypes';
-import { WelcomeTable } from 'src/features/home/types';
+import { RecentActivity, WelcomeTable } from 'src/features/home/types';
 import { Dashboard, Filter, TableTab } from './types';
 
 // Modifies the rison encoding slightly to match the backend's rison encoding/decoding. Applies globally.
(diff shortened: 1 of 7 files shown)


superset-frontend/src/views/CRUD/utils.tsx
```diff
@@ -223,10 +224,14 @@ export const getRecentActivityObjs = (
...
     const res: any = {};
+    const distinctRes = lruCache<RecentActivity>(6);
+    recentsRes.json.result.reverse().forEach((record: RecentActivity) => {
+      distinctRes.set(record.item_url, record);
+    });
     return getFilteredChartsandDashboards(addDangerToast, filters).then(
       ({ other }) => {
         res.other = other;
-        res.viewed = recentsRes.json.result;
+        res.viewed = distinctRes.values().reverse();
```
michael-s-molina, line 234: Is it necessary to reverse twice or can this be optimized?

michael-s-molina (commented):
Thank you for the PR @justinpark.

`LogDAO.get_recent_activity` is only called from the `recent_activity` endpoint at `superset/views/log/api.py`. Should we remove the the whole distinct block? Given that this endpoint is not under v1, it's not a breaking change.

```
def get_recent_activity(
   actions: list[str],
   distinct: bool,
   page: int,
   page_size: int,
) -> list[dict[str, Any]]:
   ...
   if distinct:
      ...
```

cc @sadpandajoe @villebro

justinpark:
> Given that this endpoint is not under v1, it's not a breaking change.

Its endpoint is `/api/v1/log/recent_activity/` and the distinct param is included in the schema so it will be a break change.

justinpark:
> > Its endpoint is /api/v1/log/recent_activity/ and the distinct param is included in the schema so it will be a break change.
> 
> Oh, you're correct!

Therefore, it improves the frontend to fetch in an efficient way and keeps the API specification as-is.

michael-s-molina (approved):

PR #32649, opened 2025-03-13
feat: Agentic Query In Dashboard

Abhishek-kumar-ISM:
### SUMMARY
This PR is to add LLM agentic query search button at the top of the dashboard if the feature flag is set to `True`

SIP: https://github.com/apache/superset/issues/32650

I have added a search button at the top of the dashboard, allowing users to ask any query related to the dashboard, charts, anomalies, summaries, etc. When the button is clicked, it will perform an LLM agent query and return the query response in a modal.

By default, the search button will not be visible. To enable it, set the `SHOW_DASHBOARD_AGENT_QUERY` feature flag to True and update the OpenAI API key in superset_config. If you are using a locally hosted LLM, update the base URL instead.
```python
DASHBOARD_AGENTIC_QUERY_CONFIG = {
    "model": "gpt-4o",
    "temperature": 0.8,
    "max_tokens": None,
    "timeout": None,
    "max_retries": 2,
    "api_key": "-",
    "base_url": "",
}
```
#### Use cases:

1. Users can perform natural language query-based searches on the dashboard.
2. Users can ask specific questions about any chart using natural language.
3. Users can request a summary of a chart or the entire dashboard by asking the search button to do so.
4. Feature Flag: Users have an option to make feature flag as `False` if they don't want to use the feature.

#### When Feature Flag is `False` i.e default i.e no change exactly like original superset
107 99 237 72_9000_superset_dashboard_1__native_filters_key=6maYj4nzD9I

#### When Feature Flag is `True`
ss5
ss3
ss4

#### CLI logs
ss1
ss2


mistercrunch:
I think this is an interesting proof of concept, but to me it raises more questions than it provides answers. Raising some of the questions here:

- seems we're deciding to tightly couple with langchain/openai here, bringing new dependencies to the core project for something that's being a feature flag, is that the path forward? what about other LLMs? do we want to bundle langchain and commit to it as the bridge to LLMs? are there other toolkits we should consider before committing to langchain?
- the design is a bit rough (injecting a prompt over the chart title), potentially could be many more LLM "touchpoints" in the app, what's the pattern to expose AI features in the UX? What's the better pattern to trigger and expose this particular feature?
- sharing data/metadata with external providers: I can see how different environments may or may not be cool with sharing either metadata and/or data here. FF would have to come with heavy disclaimers. "THIS FEATURE FLAG WILL ALLOW USERS TO SHARE SENSITIVE DATA WITH THE PROVIDER ... "
- the prompt is good, but there's no such thing as a "perfect prompt", so should the prompt be configurable too? maybe that'd need to be exposed as a configuration parameter as well?
- wider context: do we need an env level prompt preamble "This instance of Superset is set up for CompanyX, we care a lot about ..."
- no vector db: here we're passing some of the current context for these types of features in the pass we might need a vector db - do we need to know whether/how this fits in as a foundational part of the wider "LLM interaction framework" as it evolves?

My intuition around much potential AI features - given that there are a hundred ways to implement them depending on prefs / environment constraints - and that design patterns (both UX and backend-wise) are still evolving rapidly - is that they probably belong in extensions, at least for the time being.

Abhishek-kumar-samsung:
1. Yes i have tightly coupled with langchain for agent creation but i can implement that agent without langchain i.e using for loop of thought, action, action input and observation, and move prompt template to environment ( I was waiting for some green signal to work on it ). Regarding tight coupling with openai - no its not coupled - one can run their custom llm in local and and configure 'base_url' in superset_config. I am only using openai library. Even i am not using open ai token while testing(i am using locally running llama3.1-8b)

2. Mostly all llm configurations i have kept in superset_config, and i thought based on usecases we can have separate prompts in backend that's why i didn't kept prompt in env(if want i can do that)

3. No, user need not to share data to external providers, i have kept 'base_url' in superset_config. If anyone runs their own llm maybe using TGI(then they can configure local llm path and no data will go outside). If user is using openai token then its external call.

4. Yes prompt i also thought to move to env, this i can do.

5. This maybe specific user can configure when we move prompt to env. But i didnt thought its benefit so i didn't did.

6. Yes RAG approach in deciding which chart data call tool should be called can be there. Earlier i didn't added because it will start a new vector db docker component(I thought that if i add new container service, no one will see the PR) Currently its not there because i have left to llm to decide which chart data tool to call by using chart titles.

7. Yes async support is already supported, i am showing superset provided progress bar till response is not fetched.

Abhishek-kumar-samsung:
@mistercrunch I have addressed your queries above,
Please have a look.

Abhishek-kumar-ISM:
@rusackas @mistercrunch 
can you pls update the status ?

mistercrunch:
I'll give you my take, but curious to hear from others too. Here are a few structured-with-AI thoughts:

---

### 1. If in Superset, LLM config should be modular and first-class

What we need here goes way beyond a single config block in `superset_config`. The idea of one global LLM setup doesn't scale—different features may need different models, token limits, temps, prompt chains, etc. Think OpenAI for one use case, Anthropic or local Llama for another.

Feels like LLM configs should be treated more like database connections: centrally defined and then mapped to features. We'd need:
- a registry of LLM configs (maybe via an admin UI or file),
- a mapping system that assigns features to specific LLMs,
- prompt templates that are overridable, probably Jinja-based, and customizable per org/use case.

---

### 2. Feature flags aren’t the right control plane

The current pattern of scattering feature flags for every AI experiment isn't going to scale. In theory, FFs were designed for progressive rollouts of “forever features,” but in practice they've already become a de facto config framework.

What we probably need is something more intentional: a centralized `AI_FEATURE_CONFIG` layer that lets you:
- toggle individual AI-powered features,
- assign them to specific LLM configs,
- override/extend prompts,
- control behavior more cleanly and scalably.

---

### 3. Superset already has a huge product surface

We have a lot of ground to cover already, and anything we land here we’re likely stuck supporting for the next decade. Every new AI integration adds cognitive and maintenance load in a fast-moving space with fragile dependencies and shifting UX norms.

That’s a real risk, especially when the ecosystem and design patterns are changing daily

---

### 4. The right place for this is the extension framework

This is exactly the kind of thing that should live in our (in-dev) extension framework. That would let us:
- avoid overloading core Superset,
- validate the modular design needed for the AI wave,
- and create space for orgs to experiment and customize safely.

If Superset becomes more like VSCode—with rich, modular extensions—this type of feature becomes additive instead of burdensome.

---

### 5. The way I think about it: **MCP is the path forward**

MCP is the cleanest way to support AI in a scalable, deterministic way. Instead of trying to bolt LLMs *into* Superset, we focus on exposing a solid API that external agents can use.

The flow becomes:

> “Hey LLM, I’m looking at the ‘Sales’ dashboard in Superset—what’s up with international orders in Canada?”

And the LLM uses MCP calls to introspect metadata, generate SQL, request charts, etc. Superset stays composable, auditable, and predictable—and the fuzzy magic stays outside the boundary.

---

### 6. A backend-up abstraction could work well

I’d be happy to help shape a backend-first approach - ideally through an extension:
- A `superset-extension-langchain` could expose LLM configs and register new REST endpoints,
- Extensions in the frontend can use those endpoints to build UX,
- And we define the safe, pluggable backend pieces: RBAC, prompt templating, async support, etc.

That gives us a clean, layered system—one that avoids LangChain/LLM lock-in, and gives orgs real flexibility.

---

### 7. AI testing is still shaky

Worth calling out: we don’t have great patterns yet for testing these kinds of flows. Even small prompt tweaks or model changes can create huge diffs in output. The more this stuff ends up in core, the more brittle our test matrix becomes.

---

**TL;DR**: This is a super cool POC, and the use case is legit—but we’re not ready to enshrine this in core. Let’s use this as a proof point for the extension framework, build modular LLM abstractions, and anchor our long-term AI strategy around MCP. That path gives us power *and* flexibility, without locking us into early decisions we’ll regret.

mistercrunch:
Side note: given the SIP on MCP https://github.com/apache/superset/issues/33870, wondering if some of the use cases could be satisfied through MCP. Sure it's not the same experience with having a chart analysis fully done right-in-context in the UI with a nice popover embedded straight in the right place, but from my early testing of what `superset-mcp` (community contributed https://github.com/aptro/superset-mcp) can do already through our REST API, it feels like it should be able to fetch any chart metadata, run some queries, perform analysis, and tell you about it.

Anyhow, ultimately I think both RAG and MCP approaches are useful and valid, but MCP is so much more "our lane" to support.

mistercrunch:
Worked with gpt as to what a foundational, backend-first extension could look like. Somehow feels like tightly coupling with `langgraph` (at least at this time) is a reasonable approach, more future-proof than langchain from my understanding. Anyhow, I think there are some good ideas in here:

# 🧠 Superset Extension: `superset-extension-langgraph`

A modular LLM framework for Superset, built on **LangGraph**, exposing secure and configurable APIs that the Superset UI (or external agents) can call.

---

## ✅ 1. Choose LangGraph as the Orchestration Engine

- LangGraph gives us:
  - explicit state handling and branching logic
  - retries, failure handling, async support
  - composability (define base graphs, let users override parts)
- Still compatible with LangChain tools and models.
- Better suited for long-lived, stateful, production-ready agent workflows.

---

## 📦 2. Define `LLMConfiguration` Objects

Analogous to a DB connection — stored in metadata DB, RBAC-controlled, API-exposed.

```python
LLMConfiguration {
  id: UUID,
  name: str,
  provider: str,  # "openai", "anthropic", "llama-cpp", etc.
  base_url: Optional[str],
  model: str,
  temperature: float,
  max_tokens: int,
  headers: Dict[str, str],
  default_prompt_context: str,
  metadata: JSON,
  created_by / modified_by: FK,
  permissions: [perm_id]
}
```

Can be encrypted/stored securely like database credentials.

---

## 🔐 3. Define RBAC Permissions

Use Superset's existing RBAC model to control access:

- `can_use_llm_config[llm_config_name]`
- `can_run_llm_graph[graph_name]`
- Optional: `can_use_ai_on_dashboard`, `can_run_sql_via_ai`, etc.

These permissions get attached to LLM config objects or graph identifiers. Admins decide which roles/users can use which LLM.

---

## ⚙️ 4. Support for Graph/Chain Configs

- Define default LangGraph graphs (`dashboard_summary_graph`, `chart_question_graph`, etc.)
- Let operators:
  - inject pre-prompts or override sections
  - disable/enable certain tools
  - alter text templates
- Graphs can be registered via:
  - Python config (`graph_registry.py`)
  - YAML files (`graph_config.yaml`)
  - CLI/API in future

---

## 📡 5. New REST API Endpoints

Expose endpoints to:

- list available graphs per user
- execute a specific graph with a selected LLM config
- return **typed, schema-validated responses**

### Example: POST `/api/v1/llm/execute`

```json
{
  "llm_config_id": "123",
  "graph": "dashboard_summary",
  "input": {
    "dashboard_id": 22,
    "question": "What’s happening with international orders in Canada?"
  }
}
```

### Response

```json
{
  "type": "dashboard_summary_response",
  "summary": "...",
  "related_charts": [123, 456],
  "sql_suggestions": [...]
}
```

Use `pydantic` + `JSONSchema` for validation and frontend integration.

---

## 🧩 6. Frontend Integration

- Add generic `callLLM(configId, graphName, inputPayload)` helper
- Register UI hooks per graph (e.g., “Summarize Dashboard” button)
- Build a `<LLMPanel>` React component:
  - takes a graph ID
  - handles input/output schema
  - renders result, handles loading/errors

---

## 🔍 7. Observability & Debugging

Track and optionally store:

- user ID, graph name, config used
- input + output (opt-in redaction)
- token usage, latency, retries
- success/failure state

Store in `llm_graph_invocations` table or stream to logs/tracing tools (e.g., LangSmith, OpenTelemetry).

---

## 🔬 8. Local Dev & Testing

- CLI tool: `superset llm test-graph dashboard_summary --input '{...}'`
- Include mock data, unit test scaffolds
- Allow extension devs to quickly test prompt variations + payload shape

---

## 💡 Summary

This architecture:

- Keeps Superset modular, secure, and flexible
- Supports multiple backends and AI providers
- Scales to many AI-powered features across the app
- Aligns with the long-term vision of Superset as a platform (like VSCode)
- Complements Superset’s **MCP** strategy, treating LLMs as clients rather than dependencies

Want a working scaffold with stubs for the backend config, graph loader, and REST layer? Let me know.

rusackas:
Converting this to draft for now, since it's unlikely to merge in the current situation. MCP work and Extensions work are well underway, but I think this serves as an example of what we want to DO with those tools, rather than a solution that should be merged in its current state.

rusackas:
Closing due to need for a rebase, general inactivity, but also (mostly) due to the fact that we have Superset MCP in place now, and are working on Extensions architecture that will serve these purposes and more in a very agnostic/flexible way.

PR #32658, opened 2025-03-13
fix(sync perms): Avoid UnboundLocalError during perm sync for DBs that don't support catalogs

Vitor-Avila:
### SUMMARY
The `SyncPermissionsCommand` could result in an `UnboundLocalError` in case the DB connection doesn't support catalogs (and therefore the `catalog` value is `None`). This would happen in case the connection display name has been updated + the connection has at least one dataset associated with it.

No UI changes.

### TESTING INSTRUCTIONS
1. Create a connection to a DB engine that don't support catalogs.
2. Create a dataset from it.
3. Create a chart from this dataset as well, just for good measure.
4. Edit the connection, updating its display name.

Test coverage added too.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/sync_permissions.py
+++ b/superset/commands/database/sync_permissions.py
@@ -249,11 +249,11 @@ class SyncPermissionsCommand(BaseCommand):
         self, catalog: str | None, schemas: Iterable[str]
     ) -> None:
         # rename existing catalog permission
+        new_catalog_perm_name = security_manager.get_catalog_perm(
+            self.db_connection.name,
+            catalog,
+        )
         if catalog:
-            new_catalog_perm_name = security_manager.get_catalog_perm(
-                self.db_connection.name,
-                catalog,
-            )
             new_catalog_vm = add_vm(db.session, security_manager, new_catalog_perm_name)
             perm = security_manager.get_catalog_perm(
                 self.old_db_connection_name,
--- a/tests/unit_tests/commands/databases/update_test.py
+++ b/tests/unit_tests/commands/databases/update_test.py
@@ -335,6 +335,54 @@ def test_rename_without_catalog(
     assert schema2_pvm.view_menu.name == f"[{database_without_catalog.name}].[schema2]"
 
 
+def test_rename_without_catalog_with_assets(
+    mocker: MockerFixture,
+    database_without_catalog: MockerFixture,
+) -> None:
+    """
+    Test that permissions are renamed correctly when the DB connection does not support
+    catalogs, and it has assets associated with it.
+    """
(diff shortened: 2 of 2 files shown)


superset/commands/database/sync_permissions.py
```diff
@@ -249,11 +249,11 @@ def _rename_database_in_permissions(
         self, catalog: str | None, schemas: Iterable[str]
     ) -> None:
         # rename existing catalog permission
+        new_catalog_perm_name = security_manager.get_catalog_perm(
+            self.db_connection.name,
+            catalog,
```
mistercrunch, line 254: wait, seems the before-the-pr approach is safer/checking if catalog exists (?)

Vitor-Avila: @mistercrunch `get_catalog_perm()` supports getting called with a catalog set to `None`, and it returns `None`:

``` python
def get_catalog_perm(
    self,
    database: str,
    catalog: Optional[str] = None,
) -> Optional[str]:
    """
    Return the database specific catalog permission.

    :param database: The Superset database or database name
    :param catalog: The database catalog name
    :return: The database specific schema permission
    """
    if catalog is None:
        return None

    return f"[{database}].[{catalog}]"
```

This is the expected `catalog_perm` value for datasets/charts related with a DB connection that does not support `multi_catalog`.

We need `new_catalog_perm_name` to be set regardless if `catalog == None` or not, because then down there (outside of this `if catalog:` block) we're doing:

``` python
...
# rename permissions on datasets and charts
for dataset in DatabaseDAO.get_datasets(
    self.db_connection_id,
    catalog=catalog,
    schema=schema,
):
    dataset.catalog_perm = new_catalog_perm_name
    dataset.schema_perm = new_schema_perm_name
    for chart in DatasetDAO.get_related_objects(dataset.id)["charts"]:
        chart.catalog_perm = new_catalog_perm_name
        chart.schema_perm = new_schema_perm_name
...
```

Currently when the code gets there we get an `UnboundLocalError` exception as `new_catalog_perm_name` would only be declared/have a value if `catalog != None`.

Vitor-Avila: We could make this:
``` python
new_catalog_perm_name = None
if catalog:
    new_catalog_perm_name = security_manager.get_catalog_perm(
        self.db_connection.name,
        catalog,
    )
```

But since `get_catalog_perm()` gracefully handles that, I thought it makes more sense to keep it this way.

Vitor-Avila: happy to change it tho

mistercrunch (approved):
Thanks for the clarification! LGTM

PR #32665, opened 2025-03-14
fix: Time Comparison Feature Reverts Metric Labels to Metric Keys in Table Charts

### LINKED ISSUE #32621: With Time Comparison enabled, the metric name ignores the label and uses the metric key instead
## Description
With the new addition of the Time Comparison feature in the Table chart, the metric’s label gets reverted to the metric key when enabled. This shows the metric not-so-friendly name to the UI.

## How to reproduce the bug
1. Add custom metrics to the dataset. 
2. Define the metric key and label.
3. Create a new table chart
4. Add a time-range filter.
5. Enable the time comparison feature, and choose the **Inherit range from time filter**.

## Expected results
The metric name is displayed correctly.

Image

## Actual results
The metric name ignores the label and uses the metric key instead.

Image

## Known Workarounds
Add the metric from the chart builder instead, through the Custom SQL field.

fardin-developer:
Fixes: #32621
FEATURE_CHART_PLUGINS_EXPERIMENTAL=true

### SUMMARY
This PR addresses an issue where, with Time Comparison enabled in the Table Chart, the metric name ignores the defined label and instead uses the metric key. This results in an unfriendly display name in the UI.

###Changes
- Introduced originalLabel in the DataColumnMeta interface to retain the correct label.
- Updated processComparisonColumns to store originalLabel before modifying the metric label.
- Ensured renderGroupingHeaders retrieves and displays the originalLabel correctly in grouped headers.

#### Before
Screenshot 2025-03-14 at 11 33 05 AM

#### After
Screenshot 2025-03-14 at 12 29 16 PM

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
@@ -608,9 +608,10 @@ export default function TableChart<D extends DataRecord = DataRecord>(
       // Retrieve the originalLabel from the first column in this group
       const originalLabel = columnsMeta[value[0]]?.originalLabel || key;
 
-      if (!columnsMeta[columnIndex]?.originalLabel) {
+      if (!columnsMeta[value[0]]?.originalLabel) {
+        // Use a proper logging function if available
         console.debug(
-          `originalLabel not found for column at index ${columnIndex}, using key as fallback`,
+          `originalLabel not found for column at index ${value[0]}, using key as fallback`,
         );
       }
 
--- a/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
@@ -977,7 +977,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
         ),
         Footer: totals ? (
           i === 0 ? (
-            <th>
+            <th key={`footer-summary-${i}`}>
               <div
                 css={css`
                   display: flex;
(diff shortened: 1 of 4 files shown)


superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -605,6 +605,15 @@ export default function TableChart<D extends DataRecord = DataRecord>(
...
       const startPosition = value[0];
       const colSpan = value.length;
+      // Retrieve the originalLabel from the first column in this group
+      const originalLabel = columnsMeta[value[0]]?.originalLabel || key;
+
+      if (!columnsMeta[value[0]]?.originalLabel) {
+        // Use a proper logging function if available
+        console.debug(
+          `originalLabel not found for column at index ${value[0]}, using key as fallback`,
+        );
```
geido, line 615:
```suggestion
```

fardin-developer: removed

Antonio-RiveroMartnez (approved):
Adding extra processing using the label + the comparison ones like we do for the key and not include a new property is also an option, but I have not strong preference and these changes look like a less risky fix. Thanks for taking care of this!

fardin-developer:
> This looks good @fardin-developer. Can you please add a test for this change?

Thank you, I’ll include a test with the PR.

eschutho:
Thanks @fardin-developer, I tested on the ephemeral and it looks great!

fardin-developer:
> Thanks @fardin-developer, I tested on the ephemeral and it looks great!

Glad to hear it worked out! Thanks for the update.

PR #32683, opened 2025-03-14
fix: coerce datetime conversion errors

betodealmeida:
### SUMMARY

The `normalize_dttm_col` function is raising 500s when the data can't be properly converted, but the presence of bad values in an analytical database should not be reason to raise a 500 — it should be at most a 400. This PR changes the behavior of the parser so it's more lenient, returning a `NaT` (not a time).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/core.py
+++ b/superset/utils/core.py
@@ -1682,18 +1682,26 @@ def normalize_dttm_col(
                     utc=False,
                     unit=unit,
                     origin="unix",
-                    errors="raise",
+                    errors="coerce",
                     exact=False,
                 )
             else:
                 # Column has already been formatted as a timestamp.
-                df[_col.col_label] = dttm_series.apply(pd.Timestamp)
+                try:
+                    df[_col.col_label] = dttm_series.apply(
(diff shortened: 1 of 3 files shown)


superset/utils/core.py
```diff
@@ -1682,18 +1682,26 @@ def normalize_dttm_col(
...
                 )
             else:
                 # Column has already been formatted as a timestamp.
-                df[_col.col_label] = dttm_series.apply(pd.Timestamp)
+                try:
+                    df[_col.col_label] = dttm_series.apply(
+                        lambda x: pd.Timestamp(x) if pd.notna(x) else pd.NaT
+                    )
+                except ValueError:
+                    logger.warning(
```
betodealmeida, line 1695: Alternatively we could raise a `ValidationError` here and below, returning a 422 error.

rusackas:
Looks like you might have been intending to write more description, based on the trailing comma?

Antonio-RiveroMartnez (approved):
lgtm

eschutho (approved):

Issue #5602, opened 2018-08-10
[SIP-0] Superset Improvement Proposals


betodealmeida:
# Superset Improvement Proposals
## Purpose
The purpose of a Superset Improvement Proposal (SIP) is to introduce any major change into Apache Superset (incubating). This is required in order to balance the need to support new features, uses cases, while avoiding accidentally introducing half thought-out interfaces that cause needless problems when changed.
## What is considered a major change that needs a SIP?
Any of the following should be considered a major change:

- Any major new feature, subsystem, or piece of functionality
- Any change that impacts the public interfaces of the project

What are the "public interfaces" of the project? All of the following are public interfaces that people build around:

- Visualization types
- Form data for saved dashboards and charts
- REST endpoints
- Data passed between backend and frontend
- Configuration
- Command line tools and arguments

## What should be included in a SIP?
A SIP should contain the following sections:

- Motivation: describe the problem to be solved.
- Proposed Change: describe the new thing you want to do. This may be fairly extensive and have large subsections of its own. Or it may be a few sentences, depending on the scope of the change.
- New or Changed Public Interfaces: impact to any of the "compatibility commitments" described above. We want to call these out in particular so everyone thinks about them.
- New dependencies: describe any third-party libraries that the feature will require, from PyPI or NPM. In particular, make sure their license is compatible with the [Apache License v2.0] (https://www.apache.org/licenses/LICENSE-2.0).
- Migration Plan and Compatibility: if this feature requires additional support for seamless upgrades describe how that will work. In particular, it’s important to mention if:
    - The feature requires a database migration;
    - Dashboards and charts that are saved or bookmarked will still work after the change;
    - The feature will coexist with similar functionality for some period of time, allowing for a deprecation period.
- Rejected Alternatives: What are the other alternatives you considered and why are they worse? The goal of this section is to help people understand why this is the best solution now, and also to prevent churn in the future when old alternatives are reconsidered.

## Who should initiate the SIP?
Anyone can initiate a SIP, but preferably someone with the intention of implementing it.

## Process
1. [Create an issue](https://github.com/apache/incubator-superset/issues) with the prefix “[SIP]” in the title. The issue will be tagged as “sip” by a committer, and the title will be updated with the current SIP number.
[...]

PR #32692, opened 2025-03-16
feat(superset-frontend): Migrate from npm to pnpm and implement pnpm workspaces

morpheus-lee:
feat superset-frontend: Migrate from npm to pnpm and implement pnpm workspaces

This pull request introduces a significant change by migrating the project's package manager from npm to pnpm and adopting pnpm workspaces for managing the monorepo.

**Key Changes:**

* **Replaced npm with pnpm:** The project now utilizes pnpm for faster and more efficient dependency management.
* **Implemented pnpm workspaces:** pnpm workspaces have been set.
* **Updated root `package.json`:** The root `package.json` has been modified to include the `pnpm-workspace.yaml` configuration.
* **Updated package `package.json` files:** All individual package `package.json` files within the monorepo have been updated to reflect the pnpm environment.
* **Adjusted `jest.config.js`:** The `jest.config.js` file(s) have been updated to accommodate the changes introduced by the pnpm migration, ensuring that tests continue to run correctly.

This migration is aimed at improving the overall development experience and efficiency of the project. Please review the changes carefully and let me know if you have any questions or concerns.


rusackas:
I'll do a little more reading here :)

I'd love to be rid of lerna... I'll see if this is the right candidate, bringing it up in discussion with others, particularly around the future of the extensions architecture and its release mechanisms

hainenber:
Plus one to Evan's point. Tbh, I think there would be less friction/effort migrating Lerna usage here with npm or something more native before going with pnpm or other package managers in future PR.

mistercrunch:
This is just an impression, tell me if that makes sense, but it seems the tool ecosystem in JS/TS has been consolidating. Where small tools like `lerna`  and `pnpm` emerge to take on specific issues, but ultimately get rolled in the larger tools on a longer timeframe. In many cases we've embraced the emerging ecosystem only to trim them out as they get rolled in.

In this particular case, I'm unclear if `npm` has features related to what  `lerna`  and `pnpm`  are taking on, or even a roadmap in that direction, but it's worth checking. Having to juggle with all the libs can be confusing as one or two committers will get intimate with a solution while others are confused with the constant shifts in the foundation...

mistercrunch:
Hey, thinking about it, seems this would require a SIP. Let's take the time to make sure we do the analysis upfront and validate everyone is on board.

Here's the template -> https://github.com/apache/superset/issues/5602

rusackas:
We're trying to figure out our ideal solution for monorepo publishing package. It looks like `pnpm` + `changesets` might scratch all the itches. The other contender is `nx` as we're not super happy with `lerna`. More research to be done here, but I don't currently see pnpm as a blocker.

image

rusackas:
I guess the other open question is whether or not we'd want to move to Yarn 4. It provides great support for dealing with package CVEs in its `overrides` section, which is sure nice for a large dependency tree for us.

Long story short, I'm not scared of this move, but it might not be the right move for the long haul, and might be SIP-worthy.

mistercrunch:
Would require consensus through a SIP, so putting a block label for now. Ideally this PR is set into DRAFT mode for now.

rusackas:
Per @mistercrunch's earlier comment, converting this to draft until a SIP is created and approved. Feel free to set it back to 'ready for review' once there's consensus on the approach\!

rusackas:
See also this PR, stuck in a similar state:
https://github.com/apache/superset/pull/35302

In that one, I think I want to try using turborepo rather than nx, but just haven't circled back to it.

rusackas:
Closing this one, for now. package-manager migration needs a SIP first per our Dec 2025 discussion, and it's badly conflicting after sitting untouched since March. Happy to see it resurface once there's a SIP and consensus on pnpm vs. alternatives. I had a POC PR as well, and already closed that too, fwiw. We'll still get to all this, but it's not our #1 priority to focus on build tooling at this moment.

PR #32694, opened 2025-03-16
fix(tests): search bug in tests

pa1ch:


pa1ch:
PR should be in fork, sorry

PR #32696, opened 2025-03-17
fix(spreadsheet uploads): make file extension comparisons case-insensitive

### LINKED ISSUE #32654: Can't upload spreadsheets with capitalized extensions .XLSX and .CSV
### Bug description

Running 5.0.0rc1, try to upload a spreadsheet with a capitalized extension and it causes an error:

Image

It's a bummer that I get sent files like this but they are still valid files.

### Superset version

master / latest-dev

### Python version

I don't know

### Node version

I don't know

### Browser

Chrome

### Additional context

I think the fix goes here: https://github.com/apache/superset/blob/master/superset-frontend/src/features/databases/UploadDataModel/index.tsx#L186-L187

Have that comparison happen with both the new file and the allow list first converted to lowercase. I can't test it out to send a PR right now, anyone is welcome to pick this up.

### Checklist

sfirke:
### SUMMARY
If `.xlsx` is an allowed upload type, files ending in `.XLSX` should be allowed as well.

### TESTING INSTRUCTIONS
Tried it out locally.

### ADDITIONAL INFORMATION

- [x] Has associated issue: fixes #32654

### DIFF
--- a/superset-frontend/src/features/databases/UploadDataModel/index.tsx
+++ b/superset-frontend/src/features/databases/UploadDataModel/index.tsx
@@ -183,8 +183,9 @@ export const validateUploadFileExtension = (
     return false;
   }
 
-  const fileType = extensionMatch[1];
-  return allowedExtensions.includes(fileType);
+  const fileType = extensionMatch[1].toLowerCase();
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

PR #32699, opened 2025-03-17
fix: Signature of Celery pruner jobs

michael-s-molina:
### SUMMARY
This PR changes the signatures of the pruner jobs to accept key-word arguments and also binds the task to the current task instance to allows the task to access its own state and request information. Previously, the `kwargs` passed in the configuration were not accessible when running the task.

### TESTING INSTRUCTIONS
CI should be sufficient.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/logs/prune.py
+++ b/superset/commands/logs/prune.py
@@ -69,7 +69,7 @@ class LogPruneCommand(BaseCommand):
 
         total_rows = len(ids_to_delete)
 
-        logger.info("Total rows to be deleted: %s", total_rows)
+        logger.info("Total rows to be deleted: %s", f"{total_rows:,}")
 
         next_logging_threshold = 1
 
@@ -92,7 +92,7 @@ class LogPruneCommand(BaseCommand):
             if percentage_complete >= next_logging_threshold:
                 logger.info(
                     "Deleted %s rows from the logs table older than %s days (%d%% complete)",  # noqa: E501
-                    total_deleted,
+                    f"{total_deleted:,}",
                     self.retention_period_days,
                     percentage_complete,
                 )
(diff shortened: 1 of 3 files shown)


superset/commands/logs/prune.py
```diff
@@ -69,7 +69,7 @@ def run(self) -> None:
 
         total_rows = len(ids_to_delete)
 
-        logger.info("Total rows to be deleted: %s", total_rows)
+        logger.info("Total rows to be deleted: %s", f"{total_rows:,}")
```
michael-s-molina, line 72: bycatch: Just making logs more readable.

michael-s-molina:
cc @Usiel

superset/commands/logs/prune.py
villebro, line 72: optional thought: it could be a good idea to have a `prettify_number` or similar to centralize this logic, with a few unit tests to verify/demonstrate how the value is formatted.

villebro (approved):
LGTM with a few code quality considerations.

superset/tasks/scheduler.py
```diff
@@ -124,8 +124,10 @@ def prune_log() -> None:
         logger.exception("An exception occurred while pruning report schedule logs")
 
 
-@celery_app.task(name="prune_query")
-def prune_query(retention_period_days: Optional[int] = None) -> None:
+@celery_app.task(name="prune_query", bind=True)
+def prune_query(
+    self: Task, retention_period_days: Optional[int] = None, **kwargs: Any
```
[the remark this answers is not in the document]
michael-s-molina: Make sense. Done!

Usiel:
> cc @Usiel

Thanks for the ping. Both versions (pre fix and post fix) work for me, in the sense that `retention_period_days` is set to the value we pass in. Is this issue specific to a broker or celery version?

michael-s-molina:
> Is this issue specific to a broker or celery version?

I previously tested with 5.4.0 too.

PR #32701, opened 2025-03-17
fix: boolean filters in Explore

betodealmeida:
### SUMMARY

When filtering boolean columns in the chart builder we're using the operator `==` for `IS TRUE` and `IS FALSE`, instead of the native operators:

```typescript
  [Operators.IsTrue]: { display: t('Is true'), operation: '==' },
  [Operators.IsFalse]: { display: t('Is false'), operation: '==' },
```

There are two problems with this approach. First, the rendered filtered looks ugly, showing `value = TRUE` instead of `value IS TRUE`:

Screenshot 2025-03-17 at 10 44 33 AM

Screenshot 2025-03-17 at 10 44 43 AM

Second, the filter in the request payload looks like this:

```json
{"col": "value", "op": "==", "val": true}
```

When the SQLAlchemy query is generated, the filter will be built like this:

```python
where_clause_and.append(sqla_col == eq)
```

Which is iffy and error prone, specially because many databases don't support native boolean types, using integers for them. This could potentially result in broken comparisons between a boolean and an integer, for example.

It's much better to map the `IS TRUE` and `IS FALSE` to the `IS TRUE` and `IS FALSE` operators, since then the SQLAlchemy filter is generated like this:

```python
where_clause_and.append(sqla_col.is_(True))
```

This PR fixes the problem by mapping `IS TRUE` and `IS FALSE` to the correct operators (and cleaning up some code). With these changes, the UI looks correct:

Screenshot 2025-03-17 at 11 11 26 AM

And the request payload:

```json
{"col": "value", "op": "IS TRUE"}
```

See discussion above.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/FilterControl/AdhocFilter/index.js
+++ b/superset-frontend/src/explore/components/controls/FilterControl/AdhocFilter/index.js
@@ -18,7 +18,7 @@
  */
 import {
   CUSTOM_OPERATORS,
-  Operators,
+  DISABLE_INPUT_OPERATORS,
   OPERATOR_ENUM_TO_OPERATOR_TYPE,
 } from 'src/explore/constants';
 import { translateToSql } from '../utils/translateToSQL';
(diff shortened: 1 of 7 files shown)


superset-frontend/src/explore/components/controls/FilterControl/AdhocFilter/index.js
```diff
@@ -103,34 +93,30 @@ export default class AdhocFilter {
...
+      ) {
+        return this.subject;
       }
+
       if (this.operator && this.subject && this.clause) {
         if (Array.isArray(this.comparator)) {
-          if (this.comparator.length > 0) {
-            // A non-empty array of values ('IN' or 'NOT IN' clauses)
-            return true;
-          }
```
betodealmeida, line 124: If we don't return `true` here we always return `false` at the end of function, so we can simplify this to a single line. Ditto below.

superset-frontend/src/explore/constants.ts
```diff
@@ -83,8 +83,8 @@ export const OPERATOR_ENUM_TO_OPERATOR_TYPE: {
     display: t('use latest_partition template'),
     operation: 'LATEST PARTITION',
   },
-  [Operators.IsTrue]: { display: t('Is true'), operation: '==' },
-  [Operators.IsFalse]: { display: t('Is false'), operation: '==' },
+  [Operators.IsTrue]: { display: t('Is true'), operation: 'IS TRUE' },
+  [Operators.IsFalse]: { display: t('Is false'), operation: 'IS FALSE' },
```
betodealmeida, line 87: This is the main change in the PR, everything else is fixing the code to work properly.

sadpandajoe (approved):

PR #32707, opened 2025-03-17
feat(stack by dimension): add a stack by dimension dropdown list

### LINKED ISSUE #32496: Feature request: stacked grouped mixed chart
Business requirement specification based on https://github.com/apache/superset/discussions/28181

### Request

We need a bar chart that supports both **grouping and stacking** simultaneously. The X-axis should handle **arbitrary data or time series**. Additionally, the chart must support a **mixed format** (e.g., bar + line overlay) and include **trend lines**.  

### Example Use Case:  
We currently track the ratio of **User Stories delivered in Sprint N** to **Bugs opened in Sprint N** (split by P1 and P2/P3). An efficiency line (#User Stories / #Bugs Open) helps visualize team performance, while a **trend line for P1 bugs** highlights patterns over time.  

This is critical for maintaining product quality as we scale. As technical debt grows, bug influx can impact efficiency. Tracking these metrics ensures we maintain high standards while keeping the product competitive.

### Google Sheet implementation 
Example of implementation using Google Sheet: https://docs.google.com/spreadsheets/d/1G3rntcOYkf9Uwz4yBHDFOsfgV87RT0UMh15HbxTXlh4/edit?usp=sharing

The chart groups **User Stories (blue bars)** and **Bugs (red + yellow bars)** by Agile sprints. At the same time, **Bugs are stacked** to show priority levels (**red for Priority 1, yellow for Priority 2/3**).  

The **green line** represents efficiency across sprints, while the **light red line** is an automatically calculated **trend line for Priority 1 Bugs**.

Image

jpchev:
### SUMMARY
feature for bar chart, to stack by dimension, see
https://github.com/apache/superset/issues/32496

basically, this is a new option to let users select a dimension to 'stack by', see the video here

https://github.com/user-attachments/assets/72f04dac-860f-4e12-990d-3bf947836f9b

the dimension to 'stack by' can be selected between the dimensions defined by the user, in the _Customize_ tab

fixes: #32496


villebro:
@jpchev please add a comprehensive description + screens to this PR so reviewers can first validate the design, and then look at the implementation details. I know these may be covered by the linked GH Issue, but the discussion is pretty lengthy. I think the feature is great, but I feel we need to be very careful of how this is implemented to ensure we don't pile on technical/design debt that will be difficult to clean up later.

rusackas:
This is looking cool!

Do you need to add the "Stack by" dimension to the "Dimensions" input? If so, this works fine as-is.

If you _don't,_ I'd be still be happy with:
• Renaming "Dimension" to "Split By"
• Moving the "Stack By" to the main control panel
• Killing the checkbox

... but if that entails a lot of work (which it might) then I still like this improvement :)

jpchev:
> This is looking cool!
> 
> Do you need to add the "Stack by" dimension to the "Dimensions" input? If so, this works fine as-is.
> 
> If you _don't,_ I'd be still be happy with: • Renaming "Dimension" to "Split By" • Moving the "Stack By" to the main control panel • Killing the checkbox
> 
> ... but if that entails a lot of work (which it might) then I still like this improvement :)

I can implement the suggested changes, but I'm having a hard time with the compiler, which doesn't recognize the 'stack' field for the type `SeriesOption` from `echarts`

maybe someone can help me ?

jpchev:
@villebro I added a description for this PR

michael-s-molina:
How is this going to interact with the Stack control that already exists?

sadpandajoe:
I'm a little bit confused about this since not everything is stacked according to your video. Is this more of a group by dimension or something?

villebro:
@jpchev I reviewed the video and in general I think this is an elegant and backwards compatible solution to the problem 👍 I will review in more detail in the coming days.

jpchev:
> How is this going to interact with the Stack control that already exists?

as far as I understand, the new feature has no interaction with the existing stack control

jpchev:
> I'm a little bit confused about this since not everything is stacked according to your video. Is this more of a group by dimension or something?

yes, this is a sort of 'group by', but it's called 'stack' in the Apache Echarts vocabulary (see https://echarts.apache.org/handbook/en/how-to/chart-types/bar/stacked-bar/). 
Actually the feature propagates a 'stack' parameter in the underlying Echarts component

michael-s-molina:
Wouldn't be better to offer a control to define how things are stacked AFTER you set `Stacked Style` to `Stack`? Right now, it find really confusing that we have 2 controls for stacking.

michael-s-molina:
> I don't know, what is the existing stack doing?

It's stacking the bar under a common name. That's why all bars appear in the same stacking group. It seems you're trying to offer a way to modify this logic and that's why I think they are related.

jpchev:
> > I don't know, what is the existing stack doing?
> 
> It's stacking the bars under a common name. That's why all bars appear in the same stacking group. It seems you're trying to offer a way to modify this logic and that's why I think they are related.

yes, they are probably related and I think that the existing stack feature should become a particular case of a broader feature which should let user stack by a label in each series. This label would work as a group by criteria, as it is implemented in Apache Echarts

b4s36t4:
Hi, @jpchev. I'm trying to implement the stacked chart as you did above for the below dataset.

```
id,sprint,stories,p0_issues,p1_p2_issues,effieciency
1,2023#07,24,1,5,4
7,2023#06,25,4,10,1.785714286
2,2023#05,26,4,12,1.625
3,2023#04,26,8,20,0.7878787879
4,2023#03,22,10,30,0.55
5,2023#02,23,9,40,0.4693877551
6,2023#01,21,15,28,0.488372093
```
Can you help me understand by giving a demo for this, I'm not able to build the chart for the dataset as you did. 
You can hit me `@b4s36t4` in discord or telegram we can take things over there. Thanks for the help!

jpchev:
> Hi, @jpchev. I'm trying to implement the stacked chart as you did above for the below dataset.
> 
> ```
> id,sprint,stories,p0_issues,p1_p2_issues,effieciency
> 1,2023#07,24,1,5,4
> 7,2023#06,25,4,10,1.785714286
> 2,2023#05,26,4,12,1.625
> 3,2023#04,26,8,20,0.7878787879
> 4,2023#03,22,10,30,0.55
> 5,2023#02,23,9,40,0.4693877551
> 6,2023#01,21,15,28,0.488372093
> ```
> 
> Can you help me understand by giving a demo for this, I'm not able to build the chart for the dataset as you did. You can hit me `@b4s36t4` in discord or telegram we can take things over there. Thanks for the help!

hello, what do you want to stack together? Is your 'stories' column the count of user stories opened by month?
Do you want to stack together the number of p0_issues and p1_p2_issues by month?
If yes, you don't need my feature, you just choose the sprint column as x axis and choose the existing stack feature to stack all data together

michael-s-molina:
@sadpandajoe You requested my review but I didn't see any change since my suggestion to implement this feature as a sub-control of the existing Stacked Style control.

rusackas:
Spinning up an ephemeral environment so we can all play around with this and figure out if it's the behavior we want. 

I see from the video that adding a Stack dimension automatically checks the Stack checkbox (awesome), but I what I might _expect_ is something like:

* Stack dimension input is not shown until the checkbox is checked
* When you check the box and _don't_ pick a dimension, it acts like it does today (no change in behavior)
* When you select a Stack By dimension, the new logic is applied. I'm a little confused by the actual logic here like @sadpandajoe is, but I'll play with it on the ephemeral and see if I can make heads or tails of it.

Also, for what it's worth, someone was just asking me about this exact feature, so I'm very excited about this getting merged once we all agree on the details :D

michael-s-molina:
> Stack dimension input is not shown until the checkbox is checked

@rusackas I don't think a checkbox is necessary here. The way I can see this working would be to offer a control to change the stack group after you set the `Stacked Style` control to `Stack`. The stack group is what ECharts use to group the dimensions. In Superset, this value is defined [here](https://github.com/apache/superset/blob/b6df88a1343d4e0bf2a066747856f3df93e9755b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts#L306).

rusackas:
> @rusackas I don't think a checkbox is necessary here.

Doh! 🤦 You're right. I forgot about the Select powering stacked mode... I thought the checkbox was there all along. So basically, would it make sense to do this? (cc @kasiazjc)
Pasted_Image_4_7_25__3_04 PM

Also, just kicking the tires here, I'm a little bit unsure of this working correctly. On the ephemeral/test workspace, I set up [this chart](http://35.93.186.138:8080/explore/?form_data_key=dPlqmlkSov4&slice_id=121). As you'll see, if you "stack by" with the "status" column, you get a stack of countries, and if you "stack by" the "country" column, you get a stack of statuses per country. It feels like the logic is inverted. 
Pasted_Image_4_7_25__3_08 PM

jpchev:
> Also, just kicking the tires here, I'm a little bit unsure of this working correctly. On the ephemeral/test workspace, I set up >[this chart](http://35.93.186.138:8080/explore/?form_data_key=dPlqmlkSov4&slice_id=121). As you'll see, if you "stack by" >with the "status" column, you get a stack of countries, and if you "stack by" the "country" column, you get a stack of >statuses per country. It feels like the logic is inverted.

Isn't "stack by" intended as "group by" in SQL ?

If you stack by status, you get one vertical stack per each status (a "group" per each status), where each segment in the vertical stack is a country (the other dimension). 
If you stack by country, you get one vertical stack per each country (a "group" per each country), where each segment in the vertical stack is a status (the other dimension).

That's the behaviour of the "stack" parameter in Apache Echarts: "_The value of stack explained what series will be stacked up together_" 

see https://echarts.apache.org/handbook/en/how-to/chart-types/bar/stacked-bar/

rusackas:
> > Also, just kicking the tires here, I'm a little bit unsure of this working correctly. On the ephemeral/test workspace, I set up >[this chart](http://35.93.186.138:8080/explore/?form_data_key=dPlqmlkSov4&slice_id=121). As you'll see, if you "stack by" >with the "status" column, you get a stack of countries, and if you "stack by" the "country" column, you get a stack of >statuses per country. It feels like the logic is inverted.
> 
> Isn't "stack by" intended as "group by" in SQL ?
> 
> If you stack by status, you get one vertical stack per each status (a "group" per each status), where each segment in the vertical stack is a country (the other dimension). If you stack by country, you get one vertical stack per each country (a "group" per each country), where each segment in the vertical stack is a status (the other dimension).
> 
> That's the behaviour of the "stack" parameter in Apache Echarts: "_The value of stack explained what series will be stacked up together_"
> 
> see https://echarts.apache.org/handbook/en/how-to/chart-types/bar/stacked-bar/

Ahh... I see your logic. I guess this is a bit of confusion on my part, but I can't help but wonder if I'm alone here. 

When you say "stack by," and select "country" I was expecting to see a stack of countries. In the current code, it's what I'd expect for a "split by" or "group by" control. It's a little weird to me that you select "stack by" and that's not what's stacked up!

Curious if @villebro / @mistercrunch / @kasiazjc have opinions on the matter. I'd be OK merging this as-is, if others are happy with it. The confusion may be all mine.

michael-s-molina:
> I'd be OK merging this as-is, if others are happy with it. The confusion may be all mine.

For me, we would need to remove the checkbox, to avoid creating confusion with the Stacked Style control, and show the additional control to change the stack group once Stacked Style = Stacked.

kgabryje:
I agree with @michael-s-molina and @rusackas - it feels more intuitive for `Dimension to stack by` control to appear only when `Stacked style` is set to `Stack`. Overall a great feature though!

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
```diff
@@ -317,6 +318,52 @@ const config: ControlPanelConfig = {
       label: t('Chart Options'),
       expanded: true,
       controlSetRows: [
+        [
+          {
+            name: 'stackbydimension',
```
kgabryje, line 323: nit: can we use camelCase for the name?

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
```diff
@@ -317,6 +318,52 @@ const config: ControlPanelConfig = {
...
+              renderTrigger: true,
+              description: t(
+                'Stack in groups, where each group corresponds to a dimension',
+              ),
+            },
+          },
+        ],
+        [
+          {
+            name: 'stackdimension',
```
kgabryje, line 337: same here

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
jpchev: sure

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
jpchev: ok

jpchev:
ok, I can implement the suggested changes.

An important point : do I need to implement something for the persistency in the database? Honestly I have missed that point, I don't know whether the fields in the form are all taken into account automatically when saving a chart

mistercrunch:
> do I need to implement something for the persistency in the database?

Nope, this will happen on its own. As you alter the controls and related logic in the database, all of this automatically end up in serialized JSON payload for the chart given chart. For reference, I'm pretty sure it lands in here -> https://github.com/apache/superset/blob/347c1740998bba767ce05a42a969f1fb295d1479/superset/models/slice.py#L80 . Another interesting to know is around "what happens with the missing params for older charts?". The short answer is we simply use the defaults defined for the unreferenced controls when loading the chart, so it's important to make sure that defaults won't change the charts in anyways. 

Now in some fairly rare cases where we want to change control semantics or structure, it sometimes requires a chart database migration, where it's possible to alter the `params` strucuture of existing charts. I don't think we need anything like that in this particular case.

jpchev:
@kgabryje @michael-s-molina @rusackas I've removed the checkbox and bound the visibility of the "dimension to stack by" dropdown to the value "Stack" of the existing "Stacked style" control.

Can you please have a look?

rusackas:
Added a test chart [here](http://35.167.64.30:8080/explore/?form_data_key=BtybBq-eASA&slice_id=121&save_action=saveas). Controls look a lot cleaner with the conditional rendering now, thank you! Seems to work well in both vertical and horizontal mode. I still feel the dimension added to "Stack by" feels more like a "Split by" or "Create Stacks for" but again, maybe it's just me expecting "Stack by" to be the dimension that IS stacked. I think the tooltip does help explain the situation, though.

jpchev:
@rusackas I replaced "Dimension to **stack** by" with "Dimension to **split** by", is that better?

jpchev:
is this feature considered as abandoned or not relevant anymore? @rusackas @michael-s-molina @villebro

michael-s-molina:
Thanks for the patience @jpchev but as you can imagine, we receive many review requests. I pulled your branch and it looks good.

- Can you remove the `package-lock.json` changes?
- I believe a better name for the control would be `Split stack by`.

michael-s-molina:
@jpchev You [deleted](https://github.com/apache/superset/pull/32707/commits/6c73fb71e3fbc08d5274467f290f8cda1862c68c) the `package-lock.json` file but I meant to just delete the changes.

michael-s-molina (approved):
Thank you for the improvement, for addressing the comments, and for the patience @jpchev!

jpchev:
thank you, I'm very happy and it creates huge value for me

taherajnaFA:
@jpchev i see the feature is merged to master, but i wasn't able to find it in the [chagelogs for either 6.0.0](https://github.com/apache/superset/blob/6.0.0/UPDATING.md) or 5.0.0.
We are currently running 5.0.0 and looking for this exact feature.
Can you advice on the release cycle for this feature? Will it be available in 6.0.0?

jpchev:
> @jpchev i see the feature is merged to master, but i wasn't able to find it in the [chagelogs for either 6.0.0](https://github.com/apache/superset/blob/6.0.0/UPDATING.md) or 5.0.0. We are currently running 5.0.0 and looking for this exact feature. Can you advice on the release cycle for this feature? Will it be available in 6.0.0?

this feature is merged in the 6.0.0
the changelog is [this one](https://github.com/apache/superset/blob/6.0.0/CHANGELOG/6.0.0.md)
and it's not available in 5.0.0

PR #32708, opened 2025-03-17
fix(logging): missing path in event data

justinpark:
### SUMMARY
Since the event data does not include the URL path, this commit adds the logic to retrieve the path by location API.

### TESTING INSTRUCTIONS
Specs

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/middleware/loggerMiddleware.js
+++ b/superset-frontend/src/middleware/loggerMiddleware.js
@@ -81,7 +81,7 @@ const loggerMiddleware = store => next => action => {
   const { eventName } = action.payload;
   let { eventData = {} } = action.payload;
 
-  const path = eventData.path || location?.href;
+  const path = eventData.path || window?.location?.href;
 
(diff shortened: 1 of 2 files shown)


superset-frontend/src/middleware/loggerMiddleware.js
```diff
@@ -81,7 +81,9 @@ const loggerMiddleware = store => next => action => {
   const { eventName } = action.payload;
   let { eventData = {} } = action.payload;
 
-  if (dashboardInfo?.id && eventData.path?.includes('/dashboard/')) {
+  const path = location?.href;
```
michael-s-molina, line 84: Maybe preserve the previous logic in case the event includes the path?
```suggestion
  const path = eventData.path || location?.href;
```

michael-s-molina (approved):

michael-s-molina (approved):

PR #32121, opened 2025-02-03
feat: security, user group support

dpgaspar:
### SUMMARY

This change completes phase 1 of [SIP-152](https://github.com/apache/superset/issues/32116)

- Bumps FAB to 4.6.0 that adds a security groups as a new feature https://github.com/dpgaspar/Flask-AppBuilder/pull/2305
- Makes all the necessary changes to support Groups for database,datasource,schema authorization
- Adds missing unit tests for `user_view_menu_names`

Screenshot 2025-02-28 at 10 50 29

Group with a `Gamma` role and a data access role with datasource births
Screenshot 2025-02-28 at 11 11 29

A user `gamma1` with Group1 
Screenshot 2025-02-28 at 11 12 34

`gamma1` user dashboard list
Screenshot 2025-02-28 at 11 13 28

`gamma1` filtered datasets dropdown
Screenshot 2025-02-28 at 11 16 27

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

Issue #32116, opened 2025-02-03
[Draft] [SIP-X] User Groups for Simplified User and Role Management


dpgaspar:
### Motivation

Currently, Apache Superset assigns roles directly to users, allowing for flexible but sometimes cumbersome permission management. As organizations scale, security management and role assignment become increasingly complex, particularly with large user bases. Managing roles at the individual user level leads to inefficiencies and inconsistencies, making access control difficult to maintain and audit.

User and role management challenges become particularly evident in the following situations:

- Applying permission changes to roles associated with a large user set, as these users are not explicitly grouped together.
- Assigning a new role to a large user set without a clear mechanism for bulk updates.
- Managing temporary or project-based access where users need time-limited permissions that must be revoked later. 
- Handling employee transitions.
- Synchronizing users with an Identity Provider (IdP) where roles are handled as groups.

This SIP proposes the introduction of User Groups, which will allow multiple users to be grouped together while also associating multiple roles with the group. This will streamline permission management, making it easier to manage access control across different user sets. By introducing groups, administrators can efficiently assign and modify permissions at scale, reducing administrative overhead while improving security and compliance.

### Proposed Change

Describe how the feature will be implemented, or the problem will be solved. If possible, include mocks, screenshots, or screencasts (even if from different tools).

### New or Changed Public Interfaces

Describe any new additions to the model, views or `REST` endpoints. Describe any changes to existing visualizations, dashboards and React components. Describe changes that affect the Superset CLI and how Superset is deployed.

### New dependencies

Describe any `npm`/`PyPI` packages that are required. Are they actively maintained? What are their licenses?

### Migration Plan and Compatibility

Describe any database migrations that are necessary, or updates to stored URLs.

### Rejected Alternatives

Describe alternative approaches that were considered and rejected.

villebro:
This is a really nice improvement! Some thoughts:

**Group properties**
 
- LDAP typically has human readable name (`cn`) that may change over time to complement the main constant "technical name" (`uid`). So it may be a good idea to have an additional field `label` that is displayed in the dropdown lists.
- Having an optional `description` field would be helpful for adding extra context. Text in this field can be displayed in the dropdown as subtitle, and also be used to filter values. We're doing something similar in the dataset dropdown, where the database name is displayed as the subtitle.

**Entity ownership**

[SIP-126](#28021) proposed introducing group based entity ownership. This would benefit deployments where charts/dashboards are expected to be ownedper group, rather than per person.

**Introduce viewer property**

[...]

Issue #28021, opened 2024-04-13
[SIP-126] Fine-grained access control to Superset entities


villebro:
### Motivation

For the main entities, like charts and dashboards, Superset employs an implicit access control model, where access is granted if the user has access to the underlying datasource. While this makes sense from a data access perspective (=you shouldn't be able to view a chart if you can't access the underlying table), it has the following drawbacks:

- Users that have access to lots of datasources face a lot of clutter in the form of seeing all charts and dashboards
- It's impossible to publish a dashboard *only* to user A if user B *also* has access to the same underlying datasource. This is especially problematic in a context where multiple orgs have access to the same database, but analyze the data from different perspectives. Think sales vs internal audit, where it can be very problematic if salespeople see how internal audit is tracking their activity.

In addition, other than via Roles, Superset doesn't have a concept of groups that multiple users can belong to. This makes it difficult to manage collective ownership of assets, as each asset has to be explicitly owned by users. While [SIP-51 Dashboard Level Access](https://github.com/apache/superset/issues/10408) proposed a solution for this, it utilized the Role model for non-permission based access controls, which is a deviation from what FAB Roles are originally intended for. This is also the case for [SIP-29 Row Level Security](https://github.com/apache/superset/issues/8644), where roles are being used due to the absence of pure non-permission based groups.

This user-centric access control model becomes particularly challenging in large organizations, where users routinely flow in and out of teams, or become absent, like when a user resigns. In these cases an admin usually has to step in and manually re-assign ownership of an asset to another team member. For this reason, it would be practical to be able to assign ownership of assets to a group or groups, rather to one or many users.

### Proposed Change

To introduce granular access controls to Superset entities, and clarify access controls, we propose the following changes:

#### 1. Introduce concept of Editors and Viewers

These new properties will replace the current "Owner" property and implicit access model as follows:

- Editor: similar to Owners, Editors are able to change the objects. 
- Viewer: Viewers are able to view, but not change entities.

The properties are added to the following models:
- Dashboard
- Chart
- Alerts & Reports
- Row Level Security

For Alerts & Reports and RLS we only introduce Editor for now. The list views are updated to only show entities that the user has access to, either via Editors or Viewers.

#### 2. Add a Group model

The Group, similar to the Role model, will make it possible to group multiple users together. This can be used to create arbitrary groups, e.g. "Finance", "Marketing", "USA", "Finland" etc. Typically, membership to these would be synced from some upstream source, like LDAP, and id token or similar.

[...]

PR #32718, opened 2025-03-18
feat(rls): Add Groups association to RLS

Antonio-RiveroMartnez:
### SUMMARY
We recently introduced the `Groups` [with this PR](https://github.com/apache/superset/pull/32121), these changes are adding support for them on the `RLS` rules. This way it become even more friendly and usable setting rules for users. 

##### Regular rule
Screenshot 2025-03-18 at 11 47 14 AM

##### Base rule
Screenshot 2025-03-18 at 11 47 26 AM

#### Rules being applied based on groups from direct user <-> group association and rule being linked to the group
Screenshot 2025-03-18 at 11 54 34 AM

##### RLS rule details
Screenshot 2025-03-18 at 11 58 38 AM

#####  Group details
Screenshot 2025-03-18 at 11 58 53 AM

##### Rule applied
Screenshot 2025-03-18 at 11 59 52 AM

### TESTING INSTRUCTIONS
1. Create group(s) using the existing functionality
2. Access the Row Level Security menu and associate the group(s) you created
3. Verify the clause gets into the generated query when for example creating a chart using a restricted table

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI

### DIFF
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -2103,8 +2103,19 @@ RLSFilterGroups = DBTable(
     "rls_filter_groups",
     metadata,
     Column("id", Integer, primary_key=True),
-    Column("group_id", Integer, ForeignKey("ab_group.id")),
-    Column("rls_filter_id", Integer, ForeignKey("row_level_security_filters.id")),
+    Column(
+        "group_id",
+        Integer,
+        ForeignKey("ab_group.id", name="rls_filter_groups_group_id_fkey"),
+    ),
+    Column(
+        "rls_filter_id",
+        Integer,
+        ForeignKey(
+            "row_level_security_filters.id", name="rls_filter_groups_rls_filter_id_fkey"
+        ),
+    ),
+    UniqueConstraint("group_id", "rls_filter_id", name="uq_rls_filter_group"),
 )
 
 
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -2107,6 +2107,7 @@ RLSFilterGroups = DBTable(
         "group_id",
         Integer,
         ForeignKey("ab_group.id", name="rls_filter_groups_group_id_fkey"),
+        nullable=False,
     ),
     Column(
         "rls_filter_id",
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -2099,6 +2099,14 @@ RLSFilterTables = DBTable(
     Column("rls_filter_id", Integer, ForeignKey("row_level_security_filters.id")),
 )
 
+RLSFilterGroups = DBTable(
+    "rls_filter_groups",
+    metadata,
+    Column("id", Integer, primary_key=True),
+    Column("group_id", Integer, ForeignKey("ab_group.id")),
+    Column("rls_filter_id", Integer, ForeignKey("row_level_security_filters.id")),
+)
+
 
 class RowLevelSecurityFilter(Model, AuditMixinNullable):
     """
@@ -2128,3 +2136,8 @@ class RowLevelSecurityFilter(Model, AuditMixinNullable):
         backref="row_level_security_filters",
     )
     clause = Column(utils.MediumText(), nullable=False)
+    groups = relationship(
+        security_manager.group_model,
+        secondary=RLSFilterGroups,
+        backref="row_level_security_filters",
+    )
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -2481,8 +2481,6 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             return []
 
         # pylint: disable=import-outside-toplevel
-        from flask_appbuilder.security.sqla.models import assoc_group_role
-
         from superset.connectors.sqla.models import (
             RLSFilterGroups,
             RLSFilterRoles,
@@ -2502,24 +2500,14 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
         )
         groups_from_roles_ids = [group_id for (group_id,) in groups_from_roles]
 
-        # Roles linked to the user via groups that were directly linked to the user
-        roles_from_groups = (
-            self.get_session.query(assoc_group_role.c.role_id)
-            .filter(assoc_group_role.c.group_id.in_(user_groups))
-            .all()
-        )
-        roles_from_groups_ids = [role_id for (role_id,) in roles_from_groups]
-
         all_group_ids = list(set(user_groups + groups_from_roles_ids))
-        all_role_ids = list(set(user_roles + roles_from_groups_ids))
-
         regular_filter_roles = (
             self.get_session.query(RLSFilterRoles.c.rls_filter_id)
             .join(RowLevelSecurityFilter)
             .filter(
                 RowLevelSecurityFilter.filter_type == RowLevelSecurityFilterType.REGULAR
             )
-            .filter(RLSFilterRoles.c.role_id.in_(all_role_ids))
+            .filter(RLSFilterRoles.c.role_id.in_(user_roles))
         )
         base_filter_roles = (
             self.get_session.query(RLSFilterRoles.c.rls_filter_id)
@@ -2527,7 +2515,7 @@ class SupersetSecurityManager(  # pylint: disable=too-many-public-methods
             .filter(
                 RowLevelSecurityFilter.filter_type == RowLevelSecurityFilterType.BASE
             )
-            .filter(RLSFilterRoles.c.role_id.in_(all_role_ids))
+            .filter(RLSFilterRoles.c.role_id.in_(user_roles))
         )
         filter_tables = self.get_session.query(RLSFilterTables.c.rls_filter_id).filter(
             RLSFilterTables.c.table_id == table.id
--- a/superset/security/manager.py
+++ b/superset/security/manager.py
@@ -31,6 +31,7 @@ from flask_appbuilder.security.sqla.models import (
     assoc_permissionview_role,
     assoc_user_group,
     assoc_user_role,
+    Group,
     Permission,
     PermissionView,
     Role,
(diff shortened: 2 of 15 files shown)


superset/connectors/sqla/models.py
```diff
@@ -2099,6 +2099,25 @@ def load_database(self: SqlaTable) -> None:
...
+    ),
+    Column(
+        "rls_filter_id",
+        Integer,
+        ForeignKey(
+            "row_level_security_filters.id", name="rls_filter_groups_rls_filter_id_fkey"
+        ),
+    ),
+    UniqueConstraint("group_id", "rls_filter_id", name="uq_rls_filter_group"),
+)
```
dpgaspar, line 2119: for roles we have:
```
RLSFilterRoles = DBTable(
    "rls_filter_roles",
    metadata,
    Column("id", Integer, primary_key=True),
    Column("role_id", Integer, ForeignKey("ab_role.id"), nullable=False),
    Column("rls_filter_id", Integer, ForeignKey("row_level_security_filters.id")),
)
```
do we need to remove `nullable=False` now? because we can have a roles or a group?

superset/security/manager.py
```diff
@@ -2473,32 +2481,73 @@ def get_rls_filters(self, table: "BaseDatasource") -> list[SqlaQuery]:
             return []
 
         # pylint: disable=import-outside-toplevel
+        from flask_appbuilder.security.sqla.models import assoc_group_role
```
dpgaspar, line 2484: nit: we could move this one to a top level import

superset/security/manager.py
```diff
@@ -2473,32 +2481,73 @@ def get_rls_filters(self, table: "BaseDatasource") -> list[SqlaQuery]:
...
+
         from superset.connectors.sqla.models import (
+            RLSFilterGroups,
             RLSFilterRoles,
             RLSFilterTables,
             RowLevelSecurityFilter,
         )
 
+        # Roles and Groups directly linked to the user
         user_roles = [role.id for role in self.get_user_roles(g.user)]
```
dpgaspar, line 2494: `get_user_roles(g.user)` will get all roles that are assigned to a user, directly or indirectly via groups

villebro:
@Antonio-RiveroMartnez @dpgaspar as proposed in [SIP-126](https://github.com/apache/superset/issues/28021), which was integrated into [SIP-152](https://github.com/apache/superset/issues/32116), there was talk of introducing a Subject model that could be either a Role, Group or User. As we start rolling out Groups into the various entity models, it feels introducing that new model + component now would be very beneficial, as otherwise we'll have to duplicate the components (one for roles, one for groups and later one for users). This will become particularly convoluted in the Dashboard and Chart models, if we introduce editors and viewers, in which case there would be 6 dropdowns all together. Doing that work now would avoid making big design overhauls later, and would save development time by not having to migrate to the new model when it's ultimately introduced.

Thoughts?

Antonio-RiveroMartnez:
Hey @villebro @michael-s-molina, totally agree on bringing the `Subject` model soon 🙏  but seems to me that it was considered as part of a phase 2 in [SIP-152](https://github.com/apache/superset/issues/32116)? While the `Group` entity in RLS would be still considered phase 1 since it's just a complementary element of the initial PR thus it could potentially be added now? What do you guys think? Thanks for looking into this

michael-s-molina:
I think it will be problematic to introduce another select for group on all screens and later consolidate all selects into one. Why don't we start with a single select given that this is a new feature? In other words, why introduce technical debt when we can actually make the experience simpler by reducing the number of things users need to handle? As @villebro pointed out, with the concepts of Editor and Viewer, we would be talking about 6 selects to handle access.

villebro:
@Antonio-RiveroMartnez I think it was my oversight not to add RLS as part of the Subject model overhaul (I wrote the [bullet list](https://github.com/apache/superset/issues/32116#issuecomment-2643567384) of entities that will be changed that was later merged into the final SIP description). I totally assumed RSL to be part of this work, as the current Role reference in RLS really is a subject reference.

The distinction in the phasing here IMO is that Phase 1 was all about introducing Groups, and Phase 2 is all about replacing existing subject-like references with the new Subject model. The fact that we're here essentially simulating the new Subject model via having double drop downs is IMO a strong indication that the Subject model + component should be introduced first, as otherwise this change would need to be completely redone in Phase 2.

Antonio-RiveroMartnez:
@villebro @michael-s-molina If `Groups` are meant to be introduced in `Phase 1`, shouldn’t they also work within RLS to be fully integrated? otherwise, they are partially usable. Also, as things stand, a refactor will be needed when we introduce the `Subject` entity regardless of `Groups` being part of `RLS` or not, but I see that we would be adding extra work on top of that.

I completely understand the concerns you've raised and I truly appreciate your time to review and having this discussion, I'm just trying to fully grasp the `Phase 1 vs Phase 2` approach so I can contribute in the best way possible.

Thanks.

dpgaspar:
> @Antonio-RiveroMartnez I think it was my oversight not to add RLS as part of the Subject model overhaul (I wrote the [bullet list](https://github.com/apache/superset/issues/32116#issuecomment-2643567384) of entities that will be changed that was later merged into the final SIP description). I totally assumed RLS to be part of this work, as the current Role reference in RLS really is a subject reference.
> 
> The distinction in the phasing here IMO is that Phase 1 is all about introducing Groups, and Phase 2 is all about replacing existing subject-like references with the new Subject model. The fact that we're here essentially simulating the new Subject model via double drop downs is IMO a strong indication that the Subject model + component should be introduced first, as otherwise this change would need to be completely redone in Phase 2.

@villebro @michael-s-molina I understand the concerns around aligning RLS with the Subject model. However, as @Antonio-RiveroMartnez pointed out, I also firmly believe that RLS should be introduced in Phase 1, without being tied to the Subject model yet. The core goal of Phase 1 is to introduce Groups in a backward-compatible way, and leaving them out of RLS at this stage would result in an incomplete and inconsistent implementation.

Per SIP-152, Phase 1 must leave Superset in a fully functional state with respect to API and datasource authorization, ensuring a smooth transition while we prepare for the breaking changes in Phase 2. Some refactoring will be necessary in Phase 1, but that would be the case regardless of RLS.

I don’t think adding a new dropdown to RLS is a strong enough reason to either prematurely introduce the Subject model or leave Phase 1 incomplete. Is it a perfect solution? No. But major changes rarely follow a straight, clean path.

villebro:
> Per SIP-152, Phase 1 must leave Superset in a fully functional state with respect to API and datasource authorization, ensuring a smooth transition while we prepare for the breaking changes in Phase 2. Some refactoring will be necessary in Phase 1, but that would be the case regardless of RLS.

@dpgaspar I'm probably missing something, as I assumed RLS should work as before without this change. Is there something in datasource authorization that has changed with the introduction of groups, causing issues for RLS? If this is indeed the case, then I can understand this intermediate change. But if not, I still feel this change is best suited for Phase 2 to avoid introducing this temporary workaround.

eschutho:
@michael-s-molina @villebro what do you think about us creating a subject dropdown that allows people to select either a group or a role? It will be an intermediary step closer to subjects, and then we wouldn't need to have two dropdowns.

villebro:
> @michael-s-molina @villebro what do you think about us creating a subject dropdown that allows people to select either a group or a role? It will be an intermediary step closer to subjects, and then we wouldn't need to have two dropdowns.

That would make it impossible to mix and match different subject types. I also assume it would require an extra db migration, by first adding a group reference to RLS, and later replacing that with the final subject reference (likely a new table).

dpgaspar:
> > @michael-s-molina @villebro what do you think about us creating a subject dropdown that allows people to select either a group or a role? It will be an intermediary step closer to subjects, and then we wouldn't need to have two dropdowns.
> 
> That would make it impossible to mix and match different subject types. I also assume it would require an extra db migration, by first adding a group reference to RLS, and later replacing that with the final subject reference (likely a new table).

Yes, it would probably require 2 migrations to get to our final destination (depends on the specifics of the implementation itself). But on the bright side, it would fully introduce `Groups` to Superset, that's 1 step forward to subjects, also I don't believe subjects implementation is right around the corner for multiple reasons, and I do think it's positive to get Superset to a state where `Groups` is a first class entity.
@eschutho proposal, could be a good compromise because, it would not introduce a new field to the RLS form addressing @michael-s-molina concern here: https://github.com/apache/superset/pull/32718#issuecomment-2734060623

Another alternative would be to feature flag this dropdown, not my first choice.

Our goal here is to get Phase 1 to be a stable and complete point regarding data authorization and RLS is an important feature on data authorization. Our users can't fully adopt `Groups` without RLS.

I also believe that introducing subject models and all it requires at this point is premature and would risk a refactor

I do agree and I'm fully onboard on stoping unnecessary tech debt, but I think this change is manageable and required, it's a compromise, also the next step is planed and approved on [SIP-152](https://github.com/apache/superset/issues/32116) that incorporated [SIP-126](https://github.com/apache/superset/issues/28021)

Yes, implementing this would likely require two migrations to reach our final goal, though the specifics depend on the implementation. However, on the positive side, this approach would fully introduce Groups into Superset, a step toward the broader Subjects concept. Given that a full `Subjects` implementation isn't imminent for several reasons, making Groups a first-class entity now makes sense (and was approved).

@eschutho's proposal seems like a reasonable compromise, as it avoids introducing a new field to the RLS form—directly addressing @michael-s-molina's concern [here](https://github.com/apache/superset/pull/32718#issuecomment-2734060623). Another alternative could be feature-flagging the dropdown, though that wouldn't be my first choice.

Our primary goal is to ensure Phase 1 is a stable and complete step forward in data authorization, with RLS playing a crucial role. Our users can't fully adopt `Groups` without RLS, making this change both necessary and timely.

Introducing now all the necessary backend implementation for Subjects is premature and could risk a refactor, since 

I agree that we should avoid unnecessary tech debt, but I believe this change is small contained and justified. It strikes a balance between progress and stability, especially considering that the next step has already been planned and approved in [SIP-152](https://github.com/apache/superset/issues/32116), which builds upon [SIP-126](https://github.com/apache/superset/issues/28021).

Unless there are strong objections, I think this is a reasonable path forward.

Antonio-RiveroMartnez:
Closing this PR in favor of the `Subjects` entity coming in `Phase 2`

PR #32720, opened 2025-03-18
fix(chart control): Change default of "Y Axis Title Margin"

### LINKED ISSUE #32528: Default Y Axis Title Margin of 15 often overlaps with axis value text
## Screenshot

Image

## Description

The default Y Axis Title Margin value is 15. A common first step for users is to increase it to 45 so that it stands far enough away from the Y Axis. I polled Slack users about this and others concurred.

## Design input
Default value should be something like 45.

Quatters:
### SUMMARY

Changes the default value of the "Y Axis Title Margin" from 15 to 30. It was reported that previous default often caused overlapping with axis value text.

Fixes #32528

| Before | After | 
|--------|-------|
|  |  |

### TESTING INSTRUCTIONS
1. Create Timeseries chart, e.g. Area Chart
2. Go to Customize tab
3. Check that "Y Axis Title Margin" is 30 by default

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/32528
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/dashboard/drillby.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/dashboard/drillby.test.ts
@@ -510,29 +510,29 @@ describe('Drill by modal', () => {
 
     it('Line chart', () => {
       testEchart('echarts_timeseries_line', 'Line Chart', [
-        [70, 93],
-        [70, 93],
+        [85, 93],
+        [85, 93],
       ]);
     });
 
     it('Area Chart', () => {
       testEchart('echarts_area', 'Area Chart', [
-        [70, 93],
-        [70, 93],
+        [85, 93],
+        [85, 93],
       ]);
     });
 
     it('Scatter Chart', () => {
       testEchart('echarts_timeseries_scatter', 'Scatter Chart', [
-        [70, 93],
-        [70, 93],
+        [85, 93],
+        [85, 93],
       ]);
     });
 
     it('Bar Chart', () => {
       testEchart('echarts_timeseries_bar', 'Bar Chart', [
-        [70, 94],
-        [362, 68],
+        [85, 94],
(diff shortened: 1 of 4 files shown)


mistercrunch:
Thanks for the PR! I restarted two cypress job in CI that appeared to be flaking.

mistercrunch (approved):
LGTM

Vitor-Avila:
hey @Quatters,

If had a chart created with the **Y Axis Title Margin** config set to the default value (have never touched it) would this code-change affect those too? Or would it only affect new charts?

I think this change makes sense, just trying to confirm if we need to have any migration for existing charts that might not have this property set or etc.

mistercrunch:
@Vitor-Avila should only affect new charts

Quatters:
@Vitor-Avila hi!
Existing charts wouldn't be affected.

Quatters:
@mistercrunch looks like jobs are still flaking.

Vitor-Avila:
great, thanks for confirming @mistercrunch and @Quatters 🙌 I thought so too, just wanted to be extra safe. I re-triggered the failed jobs

Quatters:
Any ideas why these failing? Does not seem it's related to changes.

mistercrunch:
My guess is there's some sort of "pixel-perfect" click in these tests that the new margin default offsets. Looking on what's failing `drillby.test.ts` and `drilltodetail.test.ts` these tests right-click into a chart to trigger events. I hate to say this, but you probably want to look for right-click events in those tests and shift them to the right a little (maybe by 15/2 pixels?).

Quatters:
@mistercrunch thanks for the hint! I fixed tests.

Vitor-Avila:
Awesome -- that did the trick! Thanks @Quatters

PR #32731, opened 2025-03-18
feat(where_in): Support returning None if filter_values return None

Vitor-Avila:
### SUMMARY
The `|where_in` Jinja2 filter currently returns `()` (as a string) in case the `filter_values()` macro doesn't return any value. This makes it impossible to use it in combination with the `default` Jinja2 filter. For example, using:
``` sql
select * from my_table
where column in {{ filter_values('column_name')|where_in|d('(NULL) or TRUE',True) }}
```
Produces:
``` sql
select * from my_table
where column in ()
```

This PR adds the `default_to_none` parameter to the `|where_in` filter, so that using:
``` sql
select * from my_table
where column in {{ filter_values('column_name')|where_in(default_to_none=True)|d('(NULL) or TRUE',True) }}
```

Results in:
``` sql
select * from my_table
where column in (NULL) or TRUE
```

image

### TESTING INSTRUCTIONS
Tests added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -549,12 +549,10 @@ class WhereInMacro:  # pylint: disable=too-few-public-methods
         ]
         joined_values = ", ".join(string_representations)
         result = (
-            f"({joined_values})"
-            if (joined_values is not None or not default_to_none)
-            else None
+            f"({joined_values})" if (joined_values or not default_to_none) else None
         )
 
-        if mark:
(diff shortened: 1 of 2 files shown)


betodealmeida (commented):
✅ Post-merge approval!

PR #32735, opened 2025-03-18
feat(embedded): Force a specific referrerPolicy for the iframe request

Vitor-Avila:
### SUMMARY
Some hosting apps might have configurations to automatically set a `referrerPolicy` value for any `iframe` element added, which would affect the `iframe` created by the SDK as it doesn't have a value set. This becomes an issue for more restrictive policies since the validation of allowed domains is done through the `Referrer` header value, and therefore including this header in the request is important.

This PR adds a new parameter to the `embedDashboard` function that allows specifying a `referrerPolicy` value. This parameter is optional (to ensure backwards compatibility) and if not provided should respect the hosting app.

No UI changes.

### TESTING INSTRUCTIONS
1. Load a dashboard in embedded mode specifying a `refererPolicy` in the `embedDashboard` method. 
2. Inspect the browser and validate the `refererPolicy` propagates to the request triggered by the `iframe`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-embedded-sdk/src/index.ts
+++ b/superset-embedded-sdk/src/index.ts
@@ -65,7 +65,7 @@ export type EmbedDashboardParams = {
   /** additional iframe sandbox attributes ex (allow-top-navigation, allow-popups-to-escape-sandbox) **/
   iframeSandboxExtras?: string[]
   /** force a specific refererPolicy to be used in the iframe request **/
-  referrerPolicy?: ReferrerPolicy | null
+  referrerPolicy?: ReferrerPolicy
 }
 
 export type Size = {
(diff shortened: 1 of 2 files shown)


superset-embedded-sdk/src/index.ts
```diff
@@ -64,6 +64,8 @@ export type EmbedDashboardParams = {
   iframeTitle?: string
   /** additional iframe sandbox attributes ex (allow-top-navigation, allow-popups-to-escape-sandbox) **/
   iframeSandboxExtras?: string[]
+  /** force a specific refererPolicy to be used in the iframe request **/
+  referrerPolicy?: ReferrerPolicy | null
```
Antonio-RiveroMartnez, line 68: nit: if already optional and null is only used as "is set?" you should be fine w/o it like all others optionals here. You wouldn't need the default = null neither.

Antonio-RiveroMartnez (approved):

superset-embedded-sdk/src/index.ts
Vitor-Avila: great catch! thanks for the suggestion. I've just implemented this change in a new commit

PR #32739, opened 2025-03-18
fix(excel export): big number truncation handling

### LINKED ISSUE #28551: Tables with long numbers are truncated when exported to excel
### Bug description

When downloading a table with long numbers with the excel options, these are truncated by replacing the last number with a 0. For example, 1100108628127863 would appear as 1100108628127860. This does not happen when exporting to CSV.

### How to reproduce the bug

Create a table where a dimension is some large number (15-16 digits) and then download as excel.

### Screenshots/recordings

This is an extract of the table in my dashboard
image

This is what it looks like in excel
image

### Superset version

3.1.3

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

CharlesNkdl:
### SUMMARY

Fixes #28551 
Better handling of excel export, convert 10**15 numbers to a string to prevent truncation from xlsx format.
Added test too

image

Excel export : 
BEFORE : 
Big number 1 is truncated 
image

AFTER : 
Big number 1 is a string but not truncated
image

### TESTING INSTRUCTIONS

Create a dataset with some big numbers, such as : 
```
SELECT NUMERIC '1100108628127863' AS big_number, NUMERIC '999999999999999' as big_number2
FROM video_game_sales
```
Try an excel export

### ADDITIONAL INFORMATION

- [X] Has associated issue:
https://github.com/apache/superset/issues/28551
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/utils/excel.py
+++ b/superset/utils/excel.py
@@ -56,10 +56,24 @@ def df_to_excel(df: pd.DataFrame, **kwargs: Any) -> Any:
 def apply_column_types(
     df: pd.DataFrame, column_types: list[GenericDataType]
 ) -> pd.DataFrame:
+    """
+    Applies the column types to the dataframe to prepare for an excel export
+
+    :param df: The dataframe to apply the column types to
+    :param column_types: The types of the columns
+    :return: The dataframe with the column types applied
+    """
     for column, column_type in zip(df.columns, column_types, strict=False):
         if column_type == GenericDataType.NUMERIC:
             try:
                 df[column] = pd.to_numeric(df[column])
+                # if the number is too large, convert it to a string
+                # Excel does not support numbers larger than 10^15
+                df[column] = df[column].apply(
+                    lambda x: str(x)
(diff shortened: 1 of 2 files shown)


mistercrunch (changes requested):
This will lead to mixed types (strings and numericals) in the payload which doesn't seem right. Also note that stuff like sorting in data table no behave the right way (comparing strings and numerical). Feels like this should be handled in the frontend by the Table component itself since it's purely cosmetic.

CharlesNkdl:
I'm sorry, i might not be able to understand, but this function is only called while preparing a dataframe which would be exported to xlsx format
Correct me if i'm wrong or i am missing something but tweaking with the frontend won't change the underlying problem of format problematics in a excel opening ?
Sorry still a rookie on contributing here !

mistercrunch:
Oh gotcha, somehow I thought this applied more widely for in generic dataframe/extract in other contexts...

mistercrunch (approved):

PR #32759, opened 2025-03-19
fix(migrations): fix foreign keys to match FAB 4.6.0 tables

Antonio-RiveroMartnez:
### SUMMARY
With the introduction of FAB 4.6.0 the `ab_permission_view_role` and `ab_user_role` tables are using `ON CASCADE` for deletions, however, this change is only applied in newly created DBs as there's no migration to incorporate such changes into superset for existing DBs.

Without these changes it's not possible to delete Roles due to related data still existing (Integrity error).

This PR is adding a migration to recreate those foreign keys with the proper deletion treatment and thus fixing the role deletion behavior.

Error:
error

Fix:
fix

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/shared/utils.py
+++ b/superset/migrations/shared/utils.py
@@ -193,12 +193,16 @@ def has_table(table_name: str) -> bool:
     return table_exists
 
 
-def drop_fks_for_table(table_name: str) -> None:
+def drop_fks_for_table(
+    table_name: str, constraint_names: Optional[list[str]] = None
+) -> None:
     """
-    Drop all foreign key constraints for a table if it exist and the database
-    is not sqlite.
+    Drop specific or all foreign key constraints for a table
+    if they exist and the database is not sqlite.
 
-    :param table_name: The table name to drop foreign key constraints for
+    :param table_name: The table name to drop foreign key constraints from
+    :param constraint_names: Optional list of specific foreign key names to drop.
+    If None is provided, all will be dropped.
     """
     connection = op.get_bind()
     inspector = Inspector.from_engine(connection)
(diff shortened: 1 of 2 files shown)


sadpandajoe (approved):
LGTM. Note: I think these operations aren't part of https://github.com/apache/superset/blob/master/superset/migrations/shared/utils.py

michael-s-molina (changes requested):
Please use the utility methods in https://github.com/apache/superset/blob/master/superset/migrations/shared/utils.py. Feel free to add a version of `drop_fks_for_table` where you can pass a list of names in case you don't want to drop all foreign keys.

superset/migrations/shared/utils.py
```diff
@@ -406,3 +406,33 @@ def drop_index(table_name: str, index_name: str) -> None:
     )
 
     op.drop_index(table_name=table_name, index_name=index_name)
+
+
+def drop_fks_for_table_by_names(table_name: str, constraint_names: list[str]) -> None:
```
michael-s-molina, line 411: I think it would be better in terms of simplicity and to reduce lines of code to modify the original drop_fks_for_table to accept an optional list of foreign key names. You can make it clear in the docs string that if a list is not supplied, all foreign keys will be deleted.

Antonio-RiveroMartnez: Sounds good, will also have to add one for creating the keys since `sqllite` doesn't like constraint AT ALL :(

Antonio-RiveroMartnez: Ok @michael-s-molina the PR was updated to use the utils now. Thanks

superset/migrations/shared/utils.py
```diff
@@ -207,12 +211,19 @@ def drop_fks_for_table(table_name: str) -> None:
...
-        for fk in foreign_keys:
+        existing_fks = {fk["name"] for fk in inspector.get_foreign_keys(table_name)}
+
+        # What to delete based on whether the list was passed
+        if constraint_names is not None:
+            constraint_names = list(set(constraint_names) & existing_fks)
+        else:
+            constraint_names = list(existing_fks)
+
+        for fk_name in constraint_names:
```
michael-s-molina, line 222:
```suggestion
        if fk_names is not None:
            fk_names = list(set(fk_names) & existing_fks)
        else:
            fk_names = list(existing_fks)

        for fk_name in fk_names:
```

michael-s-molina (approved):
Thanks for the changes. I left some final suggestions to improve naming.

PR #32767, opened 2025-03-20
feat: Add Aggregation Method for Big Number with Trendline

LevisNgigi:
### **SUMMARY**
This PR enhances the **Big Number with Trendline** visualization by introducing an **Aggregation Method** dropdown.   
With this change, users can now apply different aggregation methods to their selected metric, such as **SUM, AVG, MIN, MAX, MEDIAN, and LAST_VALUE**.  

### **BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF**
https://github.com/user-attachments/assets/3a304129-2c25-4ec1-b30e-89d798903b7a

### **TESTING INSTRUCTIONS**
To manually verify this feature:  
1. Navigate to **Charts** and select a chart using **Big Number with Trendline** as the visualization type.  
2. Select a **Temporal X-Axis** and a **Metric** for the query and run.  
3. Choose different values from the **Aggregation Method** dropdown,run the query and observe updates in the **Big Number**:  
   - **SUM** → Displays the total of all values over the selected period.  
   - **AVG** → Displays the mean (average) of all values.  
   - **MIN/MAX** → Displays the lowest/highest value in the selected range.  
   - **MEDIAN** → Displays the middle value, reducing the effect of outliers.  
   - **LAST_VALUE** → Works as before, displaying only the most recent data point.  

### **ADDITIONAL INFORMATION**

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
```diff
@@ -61,6 +61,28 @@ const xAxisSortVisibility = ({ controls }: { controls: ControlStateMapping }) =>
...
+      ['AVG', t('Average(Mean)')],
+      ['MIN', t('Minimum')],
+      ['MAX', t('Maximum')],
+      ['MEDIAN', t('Median')],
+    ],
+    description: t('Select an aggregation method to apply to the metric.'),
+    renderTrigger: true,
+    provideFormDataToProps: true,
+    mapStateToProps: ({ form_data }: ControlPanelState) => ({
+      value: form_data.aggregation,
```
betodealmeida, line 81: @korbit-ai seems correct here. What happens with existing charts?

```suggestion
      value: form_data.aggregation || 'LAST_VALUE',
```

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/transformProps.ts
```diff
@@ -107,13 +108,54 @@ export default function transformProps(
...
+      switch (aggregation) {
+        case 'SUM':
+          bigNumber = totalSum;
+          break;
+        case 'AVG':
+          bigNumber = totalSum / metricValues.length;
+          break;
+        case 'MIN':
+          bigNumber =
+            metricValues.length > 0 ? Math.min(...metricValues) : null;
```
betodealmeida, line 134:
```suggestion
            Math.min(...metricValues);
```

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/transformProps.ts
```diff
@@ -107,13 +108,54 @@ export default function transformProps(
...
+          break;
+        case 'MAX':
+          bigNumber =
+            metricValues.length > 0 ? Math.max(...metricValues) : null;
```
betodealmeida, line 138:
```suggestion
            Math.max(...metricValues);
```

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/transformProps.ts
```diff
@@ -107,13 +108,54 @@ export default function transformProps(
...
+      .map(d => d[1])
+      .filter((value): value is number => value !== null);
+
+    if (metricValues.length > 0) {
+      const sortedValues: number[] = [...metricValues].sort((a, b) => a - b);
+      const totalSum: number = metricValues.reduce(
+        (sum, value) => sum + value,
+        0,
+      );
+      const mid = Math.floor(sortedValues.length / 2);
```
betodealmeida, line 123: It's better to move this inside `case 'MEDIAN':`, so it's only computed if needed.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/transformProps.ts
```diff
@@ -107,13 +108,54 @@ export default function transformProps(
...
+          break;
+        case 'LAST_VALUE':
+        default:
+          bigNumber = sortedData.length > 0 ? (sortedData[0][1] ?? null) : null;
+
+          if (bigNumber === null) {
+            bigNumberFallback = sortedData.find(d => d[1] !== null) ?? null;
+            bigNumber = bigNumberFallback ? bigNumberFallback[1] : null;
+            timestamp = bigNumberFallback ? bigNumberFallback[0] : null;
+          }
```
betodealmeida, line 154: This should be moved to line 160, so that it also applies when `bigNumber` is null because of line 158.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/transformProps.ts
```diff
@@ -107,13 +108,54 @@ export default function transformProps(
...
+          break;
+        case 'MEDIAN':
+          bigNumber =
+            sortedValues.length % 2 === 0
+              ? ((sortedValues[mid - 1] ?? 0) + (sortedValues[mid] ?? 0)) / 2
+              : (sortedValues[mid] ?? 0);
+          break;
+        case 'LAST_VALUE':
+        default:
+          bigNumber = sortedData.length > 0 ? (sortedData[0][1] ?? null) : null;
```
betodealmeida, line 148: You already check that `metricValues` is not empty, so you can just:

```suggestion
          bigNumber = sortedData[0][1];
```

betodealmeida (changes requested):

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
```diff
@@ -61,6 +61,28 @@ const xAxisSortVisibility = ({ controls }: { controls: ControlStateMapping }) =>
...
+      ['LAST_VALUE', t('Last Value')],
+      ['SUM', t('Total(Sum)')],
+      ['AVG', t('Average(Mean)')],
+      ['MIN', t('Minimum')],
+      ['MAX', t('Maximum')],
+      ['MEDIAN', t('Median')],
+    ],
+    description: t('Select an aggregation method to apply to the metric.'),
+    renderTrigger: true,
+    provideFormDataToProps: true,
```
mistercrunch, line 79: should there be a `default: LAST_VALUE` prop here?

LevisNgigi: Hey,just added it here. value: form_data.aggregation || 'LAST_VALUE',

mistercrunch: I think this approach might not show the control with the value selected when it loads.

mistercrunch: I mean when you create a brand new chart, does it show with `LAST_VALUE` selected?

mistercrunch: Like on this line, setting the default value for the control:
https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/sharedControls.tsx#L320

LevisNgigi: okay just added that as well.

rusackas:
I was just wishing we had this yesterday, thanks for contributing it!

LevisNgigi:
> I was just wishing we had this yesterday, thanks for contributing it!

Thanks @rusackas pleasure is mine.

geido:
@LevisNgigi would you please include a video or screenshots of the feature? Thank you

LevisNgigi:
> @LevisNgigi would you please include a video or screenshots of the feature? Thank you

@geido I have added the video demo of the feature.

mistercrunch:
LGTM - @betodealmeida see if you can lift the "request changes" hold

Antonio-RiveroMartnez (dismissed):
Hey @LevisNgigi Thanks for taking care of this! I see the UI is the one doing the aggregation, we have some charts like `Pivot Table` that already have a similar control so you could refer to that since it supports more aggregation methods. However, I would prefer not having the UI computing these values, and luckily for us we already support `PostProcessing` rules as [you can see here](https://github.com/apache/superset/blob/master/superset/utils/pandas_postprocessing/aggregate.py),  these are usually used through an operator, we already have one [for Aggregation ](https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-core/src/query/types/PostProcessing.ts#L76), the `BigNumberWithTrendline` chart already [make use of several of these](https://github.com/apache/superset/blob/master/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/buildQuery.ts#L43), so you would only need to inject some logic here to dynamically build the proper object and let the backend do the rest along with the control with more methods?

eschutho:
Thanks for the PR @LevisNgigi! I would suggest making the aggregation field required and not allow the value to be cleared. Here's a video where if you clear the value it doesn't save. 

https://github.com/user-attachments/assets/0084a7c2-820c-40a4-aa47-803cfa6bb107

LevisNgigi:
> Hey @LevisNgigi Thanks for taking care of this! I see the UI is the one doing the aggregation, we have some charts like `Pivot Table` that already have a similar control so you could refer to that since it supports more aggregation methods. However, I would prefer not having the UI computing these values, and luckily for us we already support `PostProcessing` rules as [you can see here](https://github.com/apache/superset/blob/master/superset/utils/pandas_postprocessing/aggregate.py), these are usually used through an operator, we already have one [for Aggregation ](https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-core/src/query/types/PostProcessing.ts#L76), the `BigNumberWithTrendline` chart already [make use of several of these](https://github.com/apache/superset/blob/master/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/buildQuery.ts#L43), so you would only need to inject some logic here to dynamically build the proper object and let the backend do the rest along with the control with more methods?

@Antonio-RiveroMartnez Thank you for the feedback will be looking into it.

LevisNgigi:
> Thanks for the PR @LevisNgigi! I would suggest making the aggregation field required and not allow the value to be cleared. Here's a video where if you clear the value it doesn't save.
> 
>  Screen.Recording.2025-03-21.at.12.08.14.PM.mov

@eschutho Thanks. Will be making it it a required field.

eschutho:
Great, and we can keep the default as the last value. 👍

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/buildQuery.ts
```diff
@@ -47,5 +48,19 @@ export default function buildQuery(formData: QueryFormData) {
         flattenOperator(formData, baseQueryObject),
       ],
     },
+
+    {
```
Antonio-RiveroMartnez, line 52: Do we need this block here or is it OK just to include the new `aggregationOperator` in the list above afer or before the `flattenOperator` if teh value is not `LAST_VALUE`?

superset-frontend/packages/superset-ui-core/test/query/types/PostProcessing.test.ts
```diff
@@ -61,7 +61,7 @@ const AGGREGATES_OPTION: Aggregates = {
 };
 
 const AGGREGATE_RULE: PostProcessingAggregation = {
-  operation: 'aggregation',
+  operation: 'aggregate',
```
Antonio-RiveroMartnez, line 64: ❤️ oh so there was a mismatch between the UI and how the BE was expecting this operator to work eh? Thanks for fixing that

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
```diff
@@ -61,6 +61,30 @@ const xAxisSortVisibility = ({ controls }: { controls: ControlStateMapping }) =>
   ensureIsArray(controls?.groupby?.value).length === 0 &&
   ensureIsArray(controls?.metrics?.value).length === 1;
 
+export const aggregationControl = {
```
Antonio-RiveroMartnez, line 64: Could we consider adding the list of options that are supported by the backend instead of this shortened version so we could easily match those there thus we can reuse this control later down the road on Pivot charts and others if needed? Like step 1: This PR with the base shared control and using the backend, step 2 (follow up): replace any aggregation control (from Pivot table for example) to use this new control and use the backend for aggregation like in this PR.

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
```diff
@@ -61,6 +61,30 @@ const xAxisSortVisibility = ({ controls }: { controls: ControlStateMapping }) =>
...
+  name: 'aggregation',
+  config: {
+    type: 'SelectControl',
+    label: t('Aggregation Method'),
+    default: 'LAST_VALUE',
```
Antonio-RiveroMartnez, line 69: DO we need any special handling for this value since it's not a valid `Numpy` function to run aggregations with? Like, we need to make sure we don't send the operator when building the query if this is the value in formaData.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberWithTrendline/buildQuery.ts
LevisNgigi: The block is necessary because of how it affects the trendline visualization. When I tried moving it into the list above (before/after flattenOperator), the trendline disappeared from the chart after aggregation occured and only returned the big number.

superset-frontend/packages/superset-ui-core/test/query/types/PostProcessing.test.ts
LevisNgigi: Yes, the backend did not recognize 'aggregation' as a valid operator hence the need to change it to 'aggregate'.

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
LevisNgigi: I chose these specific options based on the UI design requirements in the scope of work, but I see the value in aligning with all backend-supported aggregation functions for better reusability.Wondering what is the best solution/hybrid approach for this.

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
LevisNgigi: Yes this needs special handling as it is not a valid Numpy function.In my aggregationOperator file I added this [code](https://github.com/apache/superset/blob/7b81c54a94bb26612cc2c6bbf3aa839b1f6afb95/superset-frontend/packages/superset-ui-chart-controls/src/operators/aggregateOperator.ts#L33-L35) so that aggregation does not happen for Last value.

superset-frontend/packages/superset-ui-chart-controls/src/shared-controls/customControls.tsx
LevisNgigi: added the todos, thanks.

betodealmeida (approved):
Looks great, I really like that we're leveraging the backend postprocessing for the aggregation, even if we lose the instant trigger when switching agg type.

Antonio-RiveroMartnez (approved):
Thanks for addressing all the comments and working on this!

PR #32770, opened 2025-03-20
feat: Add current_user_roles() Jinja macro

bmaquet:
### SUMMARY
This PR adds a new `current_user_roles()` Jinja macro to Superset that returns the roles of the currently logged-in user. This enables role-based data access control when integrating with semantic layers like Cube.dev, allowing organizations to implement fine-grained access control policies based on user roles.

The implementation follows the pattern of existing user context macros (`current_username()`, `current_user_id()`, etc.) and includes proper cache key generation support.

Example usage:
```sql
SELECT
'{{ 'yes' if 'Alpha' in current_user_roles() else 'no'  }}' as is_alpha_in_roles
```
```sql
SELECT '{{ current_user_roles()|tojson }}' as roles
```

Screenshot 2025-03-21 at 09 39 41
Screenshot 2025-03-21 at 09 39 51

Caching is enabled
Screenshot 2025-03-21 at 09 26 35
Screenshot 2025-03-21 at 09 26 44

### TESTING INSTRUCTIONS
1. Launch superset and add roles to the admin user
2. Launch SQL Lab and run the below query
```
SELECT
  '{{current_user_roles()|tojson }}'
```
This should return the roles
4. Create a Regular RLS with clause `__role = '{{ current_user_roles()|tojson }}'`. Attach the roles added to the admin user in step 1. and attach a dataset (e.g. `bart_lines`)
5. Create a chart on the `bart_lines` dataset. You will notice the macro is returning the roles.

### ADDITIONAL INFORMATION
- [x] Has associated issue: https://github.com/apache/superset/discussions/32769
- [ ] Required feature flags: None
- [ ] Changes UI: No
- [ ] Includes DB Migration: No
- [x] Introduces new feature or API: Yes (new Jinja macro)
- [ ] Removes existing feature or API: No


bmaquet:
I just noticed that my linked issue is actually a [discussion](https://github.com/apache/superset/discussions/32769). Should I create an issue instead? Should this be a SIP?

andy1xx8:
love this pr

andy1xx8:
> I don't think we need an Issue (unless there's an actual bug). We welcome enhancements. I also don't think this needs a SIP, as it's not changing any fundamental user experience. It might, however, be nice to get a nod of approval from the security/Jinja-minded people, e.g. @dpgaspar @mistercrunch @michael-s-molina.

this feature will help to broaden the usage off superset where many companies require to do custom RLS based on the role of users

superset/utils/core.py
```diff
@@ -1292,6 +1292,19 @@ def get_user_email() -> str | None:
         return None
 
 
+def get_user_roles() -> list[str] | None:
```
mistercrunch, line 1295: Not sure why we need this method in `core.py`. (?) Seems you could simply use this `[role.name for role in g.user.roles]` in `superset/jinja_context.py`

superset/utils/core.py
```diff
@@ -1292,6 +1292,19 @@ def get_user_email() -> str | None:
...
+    """
+    Get the roles (if defined) associated with the current user.
+
+    :returns: The roles
+    """
+
+    try:
+        return [role.name for role in g.user.roles]
```
mistercrunch, line 1303: not sure why we need super generic error handling here. If anything I'd try to mimic whatever error handling is done (or not done) in `superset/jinja_context.py`

superset/jinja_context.py
```diff
@@ -172,6 +174,20 @@ def current_user_email(self, add_to_cache_keys: bool = True) -> Optional[str]:
...
+        """
+        Return the roles of the user who is currently logged in.
+
+        :param add_to_cache_keys: Whether the value should be included in the cache key
+        :returns: List of role names as a JSON string
+        """
+
+        if user_roles := get_user_roles():
+            if add_to_cache_keys:
+                self.cache_key_wrapper(json.dumps(user_roles))
```
mistercrunch, line 187: why `json.dumps` here? Wouldn't the jinja use case be around parsing actual objects not json (?)

docs/docs/configuration/sql-templating.mdx
```diff
@@ -220,6 +220,21 @@ cache key by adding the following parameter to your Jinja code:
 {{ current_user_email(add_to_cache_keys=False) }}
 ```
 
+**Current User Roles**
+
+The `{{ current_user_roles() }}` macro returns the roles of the currently logged in user.
```
mistercrunch, line 225: Given your implementation now, it seems more like "returns a json-stringified array of roles for the logged in user".

mistercrunch (commented):
Left some comments. One thing I wanted to mention is that we should make sure that this will work in the context of "alerts and report", not sure whether/if the impersonation works there. Something to look into.

superset/utils/core.py
bmaquet: I decided to rely on the same general exception handling that is used in the [other functions to get user attributes](https://github.com/apache/superset/blob/master/superset/utils/core.py#L1251-L1292). 

In `superset/jinja_context.py`, there's no exception handling at all

villebro:
> Left some comments. One thing I wanted to mention is that we should make sure that this will work in the context of "alerts and report", not sure whether/if the impersonation works there. Something to look into.

@mistercrunch by default we execute Alerts & Reports as the primary owner of the report (if there are many owners, it chooses the one that last edited it). So this should work out of the box. See https://github.com/apache/superset/blob/d319543377f69ec6f0848e0393033283b39b5733/superset/config.py#L1458

superset/jinja_context.py
bmaquet: For two reasons:
1. without `json.dumps`, SQL cannot render correctly when calling the macro (see screenshot)
Screenshot 2025-03-20 at 16 59 36
2. when `add_to_cache_keys` is True, we add the value of `user_roles` into the cache, as a cache key. If `user_roles` was a list this would raise ` TypeError: unhashable type: 'list'`

If needed, users can still use JSON functions to parse the results

bmaquet:
> Left some comments. One thing I wanted to mention is that we should make sure that this will work in the context of "alerts and report", not sure whether/if the impersonation works there. Something to look into.

Thanks for your review @mistercrunch , really appreciate it! I've tried to answer as best as I could. Let me know if I can clarify anything further!

docs/docs/configuration/sql-templating.mdx
bmaquet: Yes I agree that's better. I've updated the doc

mistercrunch:
Let an inline comment, but I think we should expose another `json_dumps` macro in the jinja context, letting people compose to serve their needs.

superset/jinja_context.py
bmaquet: So the current implementation with `json_dumps()` already allows this:
Screenshot 2025-03-20 at 22 02 56

It also allows the other example you shared below:
Screenshot 2025-03-20 at 22 12 36
Screenshot 2025-03-20 at 22 12 49

I would definitely be happy to expose a `json_dumps()` macro but I'm not sure I understand what this other macro would solve?

bmaquet: I see your point now.
I believe we also could leverage jinja's `|tojson` to render the list as a json (this would remove the need for another `json_dump()` macro).

To re-use your example, this would work:
Screenshot 2025-03-20 at 22 42 25

The issue I'm struggling with now though, is that a python list can't be used as a key for a python dict, so this line `self.cache_key_wrapper(user_roles)` wouldn't work.

mistercrunch: Oh nice, didn't know about `to_json`. Neat.

bmaquet: Updated the function to use `json_dumps()` to store the list of roles as string into the cache. 
The function still returns a python list, it's just the key for the cache that is now a string.

When the query is run again, it's correctly fetching the results from the cache
Screenshot 2025-03-21 at 09 26 35
Screenshot 2025-03-21 at 09 26 44

bmaquet:
I've updated the logic based on @mistercrunch 's recommandations.

The function now returns an array of roles, instead of a JSON-stringified array.
This allows users to leverage arrays to run statements like 
```
SELECT
'{{ 'yes' if 'Alpha' in current_user_roles() else 'no'  }}' as is_alpha_in_roles
```
but also to get the list of roles in a properly formatted SQL string by using `|tojson`
```
SELECT '{{ current_user_roles()|tojson }}' as roles
```

The array of roles is added to the cache key by turning it into a string using `json.dumps()`. This allows the query to be cached and returned from the cache.

Screenshot 2025-03-21 at 09 26 35
Screenshot 2025-03-21 at 09 26 44

Vitor-Avila:
This looks really great! (btw TIL about the `|tojson` filter -- that's likely helpful for the `filter_values` macro too).

I totally agree there's value in having both options. We also have the `|where_in` filter that could make this into `('Admin', 'Public', 'Alpha')` if someone needs to use this in a SQL statement, for example. Not sure if you want to add this to the docs too.

I think it might be interesting to sort the roles specially before using them as a cache key, to avoid un-necessary hits to the DB just because of sorting changes.

superset/utils/core.py
mistercrunch, line 1303: how about `return sorted([role.name for role in g.user.roles])` so it's a bit more deterministic?

bmaquet: Added `sorted()`

bmaquet:
I've addressed your latest comments:
- Added `sorted()`
- Updated the docs to show how users can use `|where_in` with the new macro

mistercrunch (approved):
LGTM - thanks for the contribution!

PR #32774, opened 2025-03-20
fix(sqllab): Pass query_id as kwarg so backoff can see it

Antonio-RiveroMartnez:
### SUMMARY
The backoff method is accessing the `query_id` via the kwargs but the `get_query` method is always called with positional args instead so backoff library doesn't make the query_id visible for the method thus it fails with a KeyError('query_id').

This PR makes sure we call get_query(query_id=query_id) instead of get_query(query_id).

PS: We could also just try access to query_id from the args? but I have no strong preference.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


betodealmeida (approved):
That's weird, I seeing nothing on the `backoff` docs about having to use kwargs. But I guess it doesn't hurt.

PR #32778, opened 2025-03-20
fix(config): correct slack image url in talisman

v9dev:
### SUMMARY
Fixed the configuration file by replacing the incorrect Slack image URL. The previous URL was causing images to not load, even in development mode. This update ensures Slack images now load correctly.

N/A (Only config and documentation changes)

### TESTING INSTRUCTIONS

Verify that Slack images now load correctly in the UI.
Review docs/security/security.mdx to confirm the new guide is correctly formatted and informative.

### ADDITIONAL INFORMATION

- [ ] No UI changes
- [ ] No DB migration required
- [ ]  Improves documentation

### DIFF
--- a/docs/docs/security/security.mdx
+++ b/docs/docs/security/security.mdx
@@ -280,6 +280,50 @@ TALISMAN_CONFIG = {
     "content_security_policy": { ...
 ```
 
+#### Configuring Talisman in Superset
+
+Talisman settings in Superset can be modified using superset_config.py. If you need to adjust security policies, you can override the default configuration.
+
+Example: Overriding Talisman Configuration in superset_config.py for loading images form s3 or other external sources.
+
+```python
+TALISMAN_CONFIG = {
+    "content_security_policy": {
+        "base-uri": ["'self'"],
+        "default-src": ["'self'"],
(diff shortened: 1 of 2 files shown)


mistercrunch (changes requested):
Thanks for the improvements to the docs. Only thing here is duplicating `TALISMAN_CONFIG` in both `superset/config.py` and now in the docs, where they could diverge over time. I ideally we'd be able to reference one from the other.

Let's change `superset/config.py` to link out to the [new] docs ->

```python
# For more information on setting up Talisman, please refer to
# https://superset.apache.org/docs/configuration/networking-settings/#{link_to_proper_slug}
TALISMAN_CONFIG = {
{...}
```

v9dev:
@mistercrunch 
I've made the requested changes. Please review again. Thanks!

v9dev (commented):
change based on comment

rusackas:
Appriving CI run 🤞

mistercrunch (approved):
At first I thought we'd be able to remove the intricate config from config.py to keep things DRY, but obviously we need to keep it there, otherwise would be a breaking change

PR #32781, opened 2025-03-20
feat(Jinja): to_datetime filter

Vitor-Avila:
### SUMMARY
This PR adds a new Jinja filter that converts a string into a datetime. This allows to perform calculations/comparisons around temporal filters applied. For example:
``` sql
...
{% set from_expr = get_time_filter("dttm", strftime="%Y-%m-%d").from_expr %}
{% set to_expr = get_time_filter("dttm", strftime="%Y-%m-%d").to_expr %}
{% if (to_expr|to_datetime(format="%Y-%m-%d") - from_expr|to_datetime(format="%Y-%m-%d")).days > 100 %}
   do something
{% else %}
   do something else
{% endif %}
...
```

No UI changes.

### TESTING INSTRUCTIONS
Unit tests added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/unit_tests/jinja_context_test.py
+++ b/tests/unit_tests/jinja_context_test.py
@@ -444,7 +444,7 @@ def test_to_datetime(
     value: str | None, format: str | None, output: datetime | None
 ) -> None:
     """
-    Test the ``to_datetime`` Jinja2 filter.
+    Test the ``to_datetime`` custom filter.
     """
 
     result = (
@@ -470,7 +470,7 @@ def test_to_datetime(
 )
 def test_to_datetime_raises(value: str, format: str | None, match: str) -> None:
     """
-    Test the ``to_datetime`` Jinja2 raises with an incorrect
+    Test the ``to_datetime`` custom filter raises with an incorrect
     format.
     """
     with pytest.raises(
--- a/tests/unit_tests/jinja_context_test.py
+++ b/tests/unit_tests/jinja_context_test.py
@@ -431,52 +431,57 @@ def test_where_in_empty_list() -> None:
     assert where_in([], default_to_none=True) is None
 
 
-def test_to_datetime() -> None:
+@pytest.mark.parametrize(
+    "value,format,output",
+    [
+        ("2025-03-20 15:55:00", None, datetime(2025, 3, 20, 15, 55)),
+        (None, None, None),
+        ("2025-03-20", "%Y-%m-%d", datetime(2025, 3, 20)),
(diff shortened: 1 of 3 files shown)


betodealmeida (approved):

tests/unit_tests/jinja_context_test.py
```diff
@@ -429,6 +431,54 @@ def test_where_in_empty_list() -> None:
...
+    ):
+        to_datetime("2025-03-20")
+
+    with pytest.raises(
+        ValueError,
+        match="unconverted data remains:  15:55:00",
+    ):
+        to_datetime("2025-03-20 15:55:00", format="%Y-%m-%d")
+
+
```
villebro, line 481: I'm a big fan of `pytest.mark.parametrize`, I think it would sit well here.

villebro (commented):
LGTM, although **please add docs**!

Optional (crazy?) implementation idea - instead of making it a filter, we could even introduce a new parameter `datetime: bool` which would automatically spit out the endpoints as datetimes. Like so:
```python
get_time_filter("dttm", datetime=True").from_expr # This would be a datetime, not a string
```
It's probably a bad idea, but saying it out loud, just in case..

mistercrunch:
There's also this if we wanted -> https://github.com/hackebrot/jinja2-time

Vitor-Avila:
> LGTM, although **please add docs**!
> 
> Optional (crazy?) implementation idea - instead of making it a filter, we could even introduce a new parameter `datetime: bool` which would automatically spit out the endpoints as datetimes. Like so:
> 
> ```python
> get_time_filter("dttm", datetime=True").from_expr # This would be a datetime, not a string
> ```
> 
> It's probably a bad idea, but saying it out loud, just in case..

@villebro Great catch on the docs! I totally forgot that 🤦 Will add them now in a new commit.

I think that's actually a great idea, and would def make this: `{{ get_time_filter("dttm", strftime="%Y-%m-%d").to_expr|to_datetime(format="%Y-%m-%d") }}` a lot smaller. However, I think it's still valuable to have the filter in case someone is passing a date as a URL parameter, or need to compare a "static date" manually typed in a syntax against a dynamic filter? So I'd say having both would be ideal. 

I'll address feedback in a new commit here, would you like me to create a follow up for `get_time_filter()`? Or would you prefer to do it?

villebro:
@Vitor-Avila I agree, I think this is feature is valuable in itself, and can be addressed in a separate PR.

PR #32829, opened 2025-03-24
fix: update dataset/query catalog on DB changes

betodealmeida:
### SUMMARY

When we change the catalog in a database connection — say, changing a BigQuery project from `project-A` to `project-B` — if we don't have multi-catalog enabled we most likely want to update existing datasets and queries to also point to the new catalog. (If multi-catalog is enabled we assume that the admin wants to have datasets and queries across multiple catalogs, so we don't want to update them.)

This PR applies that logic.

N/A

### TESTING INSTRUCTIONS

Added some basic unit tests.

I also tested by creating a Postgres connection, adding a dataset, and changing the Postgres database. The dataset was updated accordingly.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/update.py
+++ b/superset/commands/database/update.py
@@ -23,7 +23,7 @@ from typing import Any
 
 from flask_appbuilder.models.sqla import Model
 
-from superset import is_feature_enabled
+from superset import db, is_feature_enabled
 from superset.commands.base import BaseCommand
 from superset.commands.database.exceptions import (
     DatabaseExistsValidationError,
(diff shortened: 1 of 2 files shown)


Vitor-Avila:
changes look good -- let me just manually test this locally real quick

superset/commands/database/update.py
```diff
@@ -159,6 +172,29 @@ def _handle_ssh_tunnel(self, database: Database) -> SSHTunnel | None:
...
+            SqlaTable,
+            Query,
+            SavedQuery,
+            TabState,
+            TableSchema,
+        ]:
+            fk = "db_id" if model == SavedQuery else "database_id"
+            predicate = {fk: database_id}
+            update = {"catalog": new_catalog}
+            db.session.query(model).filter_by(**predicate).update(update)
```
Vitor-Avila, line 196: Nice!

Vitor-Avila (approved):
Thanks for the fix!

PR #32830, opened 2025-03-24
feat(sqllab): Add cache sqllab sidebar table schema (#32796)

trentlavoie:
### SUMMARY
Add cache for table schema in sqllab. 

Motivation: Fetching schema can be a slow operation on query engines such as Trino and Athena. We have noticed the schema takes up to 300s to appear when Trino cluster is under load. Caching the schema allows users to have better interactive experience with sqllab.

Design: Added a new schema cache to cache manager so this can be managed separately from other caches. Many users including relational DB's such as Postgres or MySQL likely won't require this feature as this feature is ideal for query engines where fetching table schema is not a trivial operation. Feature is disabled by default and requires setting SCHEMA_CACHE_CONFIG in config.py to be enable.

### TESTING INSTRUCTIONS
1. Add Trino cluster as a database
2. Test to ensure there are no breaking changes (schema cache disabled, by default): Open sqllab and select a schema and table. Opening table schema takes up to 300s.
3. Enable schema cache: Add to config.py and restart:
```
SCHEMA_CACHE_CONFIG = {
    'CACHE_TYPE': 'redis',
    'CACHE_DEFAULT_TIMEOUT': 60 * 60 * 12, # 12 hr cache
    'CACHE_KEY_PREFIX': 'superset_schema_',
    'CACHE_REDIS_URL': f"redis://{REDIS}:{REDIS_PORT}/{REDIS_DB}"
}
```
4. Test schema cache enabled: open sqllab and select a schema and table. Opening schema table takes long on cache miss but subsequent requests are fast.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
[x] Required feature flags:
SCHEMA_CACHE_CONFIG = {
    'CACHE_TYPE': 'redis',
    'CACHE_DEFAULT_TIMEOUT': 60 * 60 * 12, # 12 hr cache
    'CACHE_KEY_PREFIX': 'superset_schema_',
    'CACHE_REDIS_URL': f"redis://{REDIS}:{REDIS_PORT}/{REDIS_DB}"
} 
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -785,6 +785,9 @@ CACHE_CONFIG: CacheConfig = {"CACHE_TYPE": "NullCache"}
 # Cache for datasource metadata and query results
 DATA_CACHE_CONFIG: CacheConfig = {"CACHE_TYPE": "NullCache"}
 
+# Schema caching configuration
+SCHEMA_CACHE_CONFIG: CacheConfig = {"CACHE_TYPE": "NullCache"}
+
 # Cache for dashboard filter state. `CACHE_TYPE` defaults to `SupersetMetastoreCache`
 # that stores the values in the key-value table in the Superset metastore, as it's
 # required for Superset to operate correctly, but can be replaced by any
--- a/superset/databases/utils.py
+++ b/superset/databases/utils.py
@@ -71,6 +71,13 @@ def get_table_metadata(database: Any, table: Table) -> TableMetadataResponse:
     :param table: Table instance
     :return: Dict table metadata ready for API response
     """
+    # Lazy import to prevent circular dependency
+    from superset.extensions import cache_manager
+    cache_key = f"table_metadata:{database.id}:{table.catalog or ''}:{table.schema or ''}:{table.table}"
+    cached_data = cache_manager.schema_cache.get(cache_key)
+    if cached_data:
+        return cached_data
+
     keys = []
     columns = database.get_columns(table)
     primary_key = database.get_pk_constraint(table)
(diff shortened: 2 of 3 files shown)


mistercrunch:
Wherever we use caching in the app, we try and couple the feature with:
- info/metadata about the data being served from cache
- a way for the user to force-refresh, bypassing the cache

Would be nice to have something similar here. Otherwise say I `ALTER TABLE` in SQL Lab, I might be in a dead-end around seeing my own changes reflected in the side-bar or elsewhere.

#### Examples
Screenshot 2025-03-24 at 1 48 16 PM
Screenshot 2025-03-24 at 1 48 01 PM

justinpark:
To add to Max's opinion, there is already a UI related to refreshing the table schema. 
Screenshot 2025-03-31 at 3 03 08 PM

However, this UI clears the cache on the frontend and re-calls the existing `/api/.../table_metadata` and `/api/.../table_metadata/extra`. Since this will fetch the cached metadata in the server cache mechanism, it will be necessary to implement logic that refreshes the cached data on the server by adding parameters similar to the force refresh (force=true) that applies to the existing schema and table list.

mistercrunch:
Oh right. Forgot about this button. So if that button adds a `?force=true`, and normal call doesn't, things should work.

Extra points if you add a `latest_refreshed` or similar to the payload(s) metadata, so that the frontend would be able to surface that say on hover of that button.

trentlavoie:
Makes sense. Let me work on adding a force refresh and latest_refreshed to the API

rusackas:

PR #32849, opened 2025-03-25
fix(plugin-chart-echarts): correct label position for Negative Values bar chart

### LINKED ISSUE #20517: Incorrect Value label position for Negative Values
Enabling 'Show Value' in bar chart should show the value label at the top of the bar(for +ve value) or below the bar(for -ve value), but it always shows value on top of bar. So all negative value labels end up on X-axis.

#### How to reproduce the bug

1. Create a bar chart with negative values.
2. Enable 'Show Value' option.

### Expected results

Negative values should be shown below the bar.

### Actual results

Negative values are shown on X-axis.

#### Screenshots

Here is a sample in master.
image

### Environment

- superset version: `master`

docker hub image

### Checklist

SBIN2010:
This PR fix https://github.com/apache/superset/issues/20517

correct label position for negative values in bar chart where stacked style is none

before
Снимок экрана от 2025-03-25 16-56-41
Снимок экрана от 2025-03-25 21-46-19

after 
Снимок экрана от 2025-03-25 16-57-27
Снимок экрана от 2025-03-25 21-44-59

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
I see there's another way to do this, but it has me a bit stumped...

ECharts supports a label position of "outside" that automatically handles this. If you look at `superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts` you can find the block that's

```
label: {
      show: !!showValue,
      position: isHorizontal ? 'right' : 'top',
      ...
```

If you set it to `position: 'outside'` it works beautifully for negative numbers, in both horizontal and vertical, without having to specify right or top at all. 

BUT... it gives me a typescript error that `outside` is not an allowed value. And I'm a little stymied by this... let me know if you think that's a better approach, and maybe we can sort out the typing issue.

SBIN2010:
> I see there's another way to do this, but it has me a bit stumped...
> 
> ECharts supports a label position of "outside" that automatically handles this. If you look at `superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformers.ts` you can find the block that's
> 
> ```
> label: {
>       show: !!showValue,
>       position: isHorizontal ? 'right' : 'top',
>       ...
> ```
> 
> If you set it to `position: 'outside'` it works beautifully for negative numbers, in both horizontal and vertical, without having to specify right or top at all.
> 
> BUT... it gives me a typescript error that `outside` is not an allowed value. And I'm a little stymied by this... let me know if you think that's a better approach, and maybe we can sort out the typing issue.


Thanks. At first I didn't notice the outside parameter in the echarts documentation for position. I tested it with this parameter - and everything works fine. 
And I didn't get any typescript error. 

Setting the show parameter is redundant in my opinion.

SBIN2010:
@rusackas 
I realized what confused you and decided to move the call of the transformNegativeLabelsPosition function to transformSeries.
Now it looks more clear - the block you mentioned is common for the series, and what is returned by the transformNegativeLabelsPosition function is a special case.

rusackas:
Sorry for taking a while to respond here, but would you be able to rebase this? I think it's effectively mergeable otherwise now :D

SBIN2010:
> Sorry for taking a while to respond here, but would you be able to rebase this? I think it's effectively mergeable otherwise now :D

What needs to be moved and where?

rusackas (approved):
LGTM!

PR #32850, opened 2025-03-25
fix(echarts): Sort series by name using natural comparison

Vitor-Avila:
### SUMMARY
The sort series logic applied to category names for echarts currently sorts alphabetically, so for example sorting a `month` column results in: `1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9`.

This PR implements `Intl.Collator` for the category name sorting.

This change would affect existing chats, but this is a bug fix as opposed to a breaking change IMHO (open for discussion here).

**Before**
image

**After**
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/package.json
+++ b/superset-frontend/plugins/plugin-chart-echarts/package.json
@@ -24,10 +24,11 @@
     "lib"
   ],
   "dependencies": {
+    "@types/react-redux": "^7.1.10",
     "d3-array": "^1.2.0",
-    "lodash": "^4.17.21",
     "dayjs": "^1.11.13",
-    "@types/react-redux": "^7.1.10"
+    "lodash": "^4.17.21",
+    "natural-compare-lite": "^1.4.0"
   },
   "peerDependencies": {
     "@superset-ui/chart-controls": "*",
--- a/superset-frontend/plugins/plugin-chart-echarts/package.json
+++ b/superset-frontend/plugins/plugin-chart-echarts/package.json
@@ -27,8 +27,7 @@
     "@types/react-redux": "^7.1.10",
     "d3-array": "^1.2.0",
     "dayjs": "^1.11.13",
-    "lodash": "^4.17.21",
-    "natural-compare-lite": "^1.4.0"
+    "lodash": "^4.17.21"
   },
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


mistercrunch:
Looked up the [lib](https://www.npmjs.com/package/natural-compare-lite?activeTab=code) and it's like 50 LoC, wondering if we should just copy/paste the code as opposed to bringing it the lib... What's our take on micropackages like this one. Maybe @rusackas has a take on it. 

Otherwise LGTM. Lib is very small/simple, and hasn't changed in 9 years :)

mistercrunch:
Screenshot 2025-03-25 at 12 42 16 PM

mistercrunch:
Fine either way for me... tradeoffs

superset-frontend/plugins/plugin-chart-echarts/package.json
```diff
@@ -24,10 +24,10 @@
     "lib"
   ],
   "dependencies": {
+    "@types/react-redux": "^7.1.10",
     "d3-array": "^1.2.0",
-    "lodash": "^4.17.21",
     "dayjs": "^1.11.13",
-    "@types/react-redux": "^7.1.10"
+    "lodash": "^4.17.21"
```
mistercrunch, line 30: should this be a peerDep?

Vitor-Avila: @mistercrunch hmm not sure. Since we changed the approach in the PR to use `Intl.Collator` there's no real changes in `package.json` -- it's just sorting these.

mistercrunch: oh wait, I thought it was added, but you simply re-ordered  them. Now wondering whether it should have been a peerDep in the first place (?)

Vitor-Avila: yeah, they were just re-ordered by running `pre-commit` before pushing this. I'm not too familiar with the `package.json` configs so not sure if we need these explicitly listed as a `dependency`, but I can look into updating this if needed.

mistercrunch (approved):
LGTM! Let's decouple this from this PR

PR #32863, opened 2025-03-26
fix(table-chart): Do not show comparison columns config if time_compare is set to []

Vitor-Avila:
### SUMMARY
If somehow a table chart has the `time_compare` configuration set to `[]`, the time comparison columns (`#`, `△` and `%`) are visible in the customize section, even though they are not displayed in the chart.

While we can't recreate this scenario through the UI (having `time_compare` set to `[]`) we have a handful of charts with this configuration on our environment, so handling this scenario in the code. It's possible to force this state via the API/import.

**Before**
image

**After**
image

### TESTING INSTRUCTIONS
1. Create a table chart with a non temporal dimension and a metric.
2. Set `params.time_compare` to `[]` either via the API or export/import.
3. Access the **Customize** tab.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
@@ -486,8 +486,9 @@ const config: ControlPanelConfig = {
                 return true;
               },
               mapStateToProps(explore, _, chart) {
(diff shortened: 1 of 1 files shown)


Antonio-RiveroMartnez (approved):
lgtm

PR #32867, opened 2025-03-26
feat(deckgl): make map tiles configurable

plavacquery:
### SUMMARY

Allow to use different tiles than mapbox with deckgl map. Tiles informations can be add in superset config.py

image

### TESTING INSTRUCTIONS

In config.py set the following
```py
DECKGL_BASE_MAP = [
     ['tile://https://c.tile.openstreetmap.org/{z}/{x}/{y}.png', 'OpenStreetMap' ],
]

ENABLE_CORS = True
CORS_OPTIONS: dict[Any, Any] = {
    "origins": [
        "https://c.tile.openstreetmap.org",
    ]
}

```
And in TALISMAN_DEV_CONF
add "https://c.tile.openstreetmap.org" to 'connect-src' array

### ADDITIONAL INFORMATION

- [x] Has associated issue :https://github.com/apache/superset/discussions/27475
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -2071,10 +2071,6 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         return False
 
     @classmethod
-    def parse_sql(cls, sql: str) -> list[str]:
-        return [str(s).strip(" ;") for s in sqlparse.parse(sql)]
-
-    @classmethod
     def get_impersonation_key(cls, user: User | None) -> Any:
         """
         Construct an impersonation key, by default it's the given username.
--- a/superset/db_engine_specs/kusto.py
+++ b/superset/db_engine_specs/kusto.py
@@ -153,11 +153,3 @@ class KustoKqlEngineSpec(BaseEngineSpec):  # pylint: disable=abstract-method
             return f"""datetime({dttm.isoformat(timespec="microseconds")})"""
 
         return None
-
-    @classmethod
-    def parse_sql(cls, sql: str) -> list[str]:
-        """
-        Kusto supports a single query statement, but it could include sub queries
-        and variables declared via let keyword.
-        """
-        return [sql]
--- a/superset/models/core.py
+++ b/superset/models/core.py
@@ -659,7 +659,7 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
         schema: str | None = None,
         mutator: Callable[[pd.DataFrame], None] | None = None,
     ) -> pd.DataFrame:
-        sqls = self.db_engine_spec.parse_sql(sql)
+        script = SQLScript(sql, self.db_engine_spec.engine)
         with self.get_sqla_engine(catalog=catalog, schema=schema) as engine:
             engine_url = engine.url
 
@@ -676,8 +676,11 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
         with self.get_raw_connection(catalog=catalog, schema=schema) as conn:
             cursor = conn.cursor()
             df = None
-            for i, sql_ in enumerate(sqls):
-                sql_ = self.mutate_sql_based_on_config(sql_, is_split=True)
+            for i, statement in enumerate(script.statements):
+                sql_ = self.mutate_sql_based_on_config(
+                    statement.format(),
+                    is_split=True,
+                )
                 _log_query(sql_)
                 with event_logger.log_context(
                     action="execute_sql",
@@ -686,7 +689,7 @@ class Database(Model, AuditMixinNullable, ImportExportMixin):  # pylint: disable
                 ):
                     self.db_engine_spec.execute(cursor, sql_, self)
 
-                rows = self.fetch_rows(cursor, i == len(sqls) - 1)
+                rows = self.fetch_rows(cursor, i == len(script.statements) - 1)
                 if rows is not None:
                     df = self.load_into_dataframe(cursor.description, rows)
 
--- a/tests/unit_tests/db_engine_specs/test_base.py
+++ b/tests/unit_tests/db_engine_specs/test_base.py
@@ -49,32 +49,6 @@ def test_get_text_clause_with_colon() -> None:
     assert text_clause.text == "SELECT foo FROM tbl WHERE foo = '123\\:456')"
 
 
-def test_parse_sql_single_statement() -> None:
-    """
-    `parse_sql` should properly strip leading and trailing spaces and semicolons
-    """
-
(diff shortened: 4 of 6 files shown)


superset/views/base.py
```diff
@@ -333,6 +333,7 @@ def cached_common_bootstrap_data(  # pylint: disable=unused-argument
         "d3_format": conf.get("D3_FORMAT"),
         "d3_time_format": conf.get("D3_TIME_FORMAT"),
         "currencies": conf.get("CURRENCIES"),
+        "deckgl_tiles": conf.get("DECKGL_BASE_MAP"),
```
mistercrunch, line 336: Not sure if it should be in scope for this PR, but we're looking to avoid growing the `bootstrap_data` payload, as it's sent to the frontend globally for all page loads. Since the set of defaults ones are hard-coded on the frontend, maybe we make the default in `superset/config.py` empty, and only pay the price of growing the payload in environments looking to customize the set (?)

mistercrunch:
Did a quick pass reviewing the PR, and overall seems like a great feature! 

Curious, are these tiles served for free, no API key required? If it's all free and open, it seems like a better default over mapbox maybe (?)

About the `bootstrap_data` comment I made, I think it's probably fair to simply leave the config empty and find a better/scalable solution in the future for the frontend to retrieve backend configuration flags async if/when needed.

NOTE: I fired up an ephemeral environment so anyone can take the feature for a test drive. Though I'm realizing that mapbox doesn't seem to work out-of-the-box, curious whether we can make OSM work in eph-envs & dev-envs.

mistercrunch:
Errors in CI point to having to `run npm install` from the `superset/superset-frontend` folder

mistercrunch:
This helps catching these errors as you commit -> https://superset.apache.org/docs/contributing/development/#git-hooks

justinpark:
@plavacquery Thanks for the contribution. Could you add the summary of the PR (and might be adding the screenshot of the changes)?

superset/views/base.py
plavacquery: Hi,
Indeed It's a good idea, I set the  DECKGL_BASE_MAP default to None

superset-frontend/plugins/legacy-preset-chart-deckgl/src/utilities/Shared_DeckGL.jsx
```diff
@@ -29,6 +29,28 @@ import {
...
 
+let deckgl_tiles;
+
+export const DEFAULT_DECKGL_TILES = [
+  ['mapbox://styles/mapbox/streets-v9', 'Streets'],
+  ['mapbox://styles/mapbox/dark-v9', 'Dark'],
+  ['mapbox://styles/mapbox/light-v9', 'Light'],
+  ['mapbox://styles/mapbox/satellite-streets-v9', 'Satellite Streets'],
+  ['mapbox://styles/mapbox/satellite-v9', 'Satellite'],
+  ['mapbox://styles/mapbox/outdoors-v9', 'Outdoors'],
```
rusackas, line 40: Should we be adding OpenStreetmaps (or others) if they're license compatible? The more, the merrier, I figure.

edogev:
I understand correctly that after this revision, it will be possible to add styles from mapbox "mapbox://styles/..." to the superset directly via config_superset.ru? 

If it has already been implemented. can you tell me how I can use the required style? Or is there another way to change the map language?

mistercrunch:
What is `config_superset.ru` (!?)

superset-frontend/plugins/legacy-preset-chart-deckgl/src/utilities/Shared_DeckGL.jsx
plavacquery: OSM tile that I added do not need any key. Instead you add to set up cors options to make it work.
If I underdtand I should replace mapbox tiles by osm ones or mix them in hard coded default tiles ?

rusackas: Hmm... I guess what we want to avoid here is any breaking change. Therefore, we can't shift _existing_ charts to OSM (that's breaking) but we could set the default for NEW charts to OSM, and leave MapBox enabled as an option. I'd also be open to converting the built-in Superset examples to use OSM as well, so things don't look broken when people first install Superset.

Maybe there's enough complexity to warrant a SIP. With that kind of consensus, we would have an idea of the to-do list, and also could get consensus to migrate away from MapBox to OSM if you (and others) think that's a viable option. I'm largely in favor of sensible defaults and configurable options, so long as we don't cause breaking changes, and provide migrations where necessary.

mistercrunch:
Triggering CI and generating an ephemeral environment to test the feature. Wondering if we should make this the default given that deckgl+mapbox doesn't work out of the box. Would be a much better experience for people trying out Superset to have an actual map out of the box without having to get/configure a mapbox api key. Might make sense to set this new default but make it configurable for those who have an API key and don't want openstreetmap as the default. Would require a new default and a note in `UPDATING.md` so let admins know that the default has changed and that they might want to configure the options.

mistercrunch:
mmmh, ephemeral env has an issue from when the `master` branch had build issues. Please rebase or merge the lastest `master` in your branch plz

plavacquery:
@mistercrunch So we can open a SIP that will make default osm tiles for deckgl and allow switching to mapbox tiles  when a mapbox token is set in config.py file.

And in the meantime can we merge this MR that allow user to use OSM or other internal tiles with config.py ? The CI has passed

mistercrunch:
Saw new commits so I approved CI run and fired up a fresh ephemeral environment

datasc24:
Hello ! This feature looks great ! Thanks @plavacquery !
@mistercrunch what's missing for this PR to be merged ?

mistercrunch:
The 3 points from my earlier comments - if we all agree ->
```
So moving forward and in-scope for this PR:

* open street map is the default and works out of the box when firing up Superset with default config. Ideally, when creating an ephemeral environment from this PR, or simply firing docker compose up, deckgl would now work out-of-the-box on top of OSM
* Mapbox also enabled, but requires setting up an API key to work
* configuration/documentation allowing for Mapbox only, mimicking the behavior prior-to-this-PR. Administrators can disable/hide OSM if they desire
```

point 3 requires a new comment in `./UPDATING.md`, something along the lines of "OpenStreetView has been promoted as the new default for Deck.gl visualization since it can be enabled by default without requiring an API key. If you have Mapbox set up and want to disable OpenStreeView in your environment, please follow the steps documented [here](https://superset.apache.org/docs/configuration/configuring-superset/)"

PR #32891, opened 2025-03-28
docs: pypi-installation on Ubuntu 24.04 and statsd package for event-logging

125m125:
### SUMMARY
* Adds Ubuntu 24.04 and the need for an older python version to the pypi installation instructions.
* Adds note that the statsd-package needs to be installed for the StatsdStatsLogger and how to inspect errors
  * Suggestion: Maybe instead of suppressing the exception, superset could be adjusted to still create the `StatsdStatsLogger` class but it throws the caught exception on init so the users can see the actual problem?
```
try:
    from statsd import StatsClient
    ...
except Exception as e:  # pylint: disable=broad-except
    class StatsdStatsLogger(BaseStatsLogger):
        def __init__(  # pylint: disable=super-init-not-called
            self,
            host: str = "localhost",
            port: int = 8125,
            prefix: str = "superset",
            statsd_client: Optional[StatsClient] = None,
        ) -> None:
            raise e
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docs/docs/configuration/event-logging.mdx
+++ b/docs/docs/configuration/event-logging.mdx
@@ -51,24 +51,12 @@ if desired. Most endpoints hit are logged as
 well as key events like query start and end in SQL Lab.
 
 To setup StatsD logging, it’s a matter of configuring the logger in your `superset_config.py`.
-If not already present, you need to ensure that the `statsd`-package is installed in superset's python environment.
+If not already present, you need to ensure that the `statsd`-package is installed in Superset's python environment.
 
 ```python
 from superset.stats_logger import StatsdStatsLogger
 STATS_LOGGER = StatsdStatsLogger(host='localhost', port=8125, prefix='superset')
 ```
-
-If you get the Error `ImportError: cannot import name 'StatsdStatsLogger' from 'superset.stats_logger' (/path/to/superset/env/superset/stats_logger.py)`, you can edit the `except Exception` block to print error details:
-```python
-try:
-    from statsd import StatsClient
-
-    class StatsdStatsLogger(BaseStatsLogger):
-    ...
-except Exception as e:  # pylint: disable=broad-except
-    print(e)
-```
-
 
 Note that it’s also possible to implement your own logger by deriving
 `superset.stats_logger.BaseStatsLogger`.
--- a/docs/docs/configuration/event-logging.mdx
+++ b/docs/docs/configuration/event-logging.mdx
@@ -51,11 +51,24 @@ if desired. Most endpoints hit are logged as
 well as key events like query start and end in SQL Lab.
 
 To setup StatsD logging, it’s a matter of configuring the logger in your `superset_config.py`.
+If not already present, you need to ensure that the `statsd`-package is installed in superset's python environment.
 
 ```python
 from superset.stats_logger import StatsdStatsLogger
 STATS_LOGGER = StatsdStatsLogger(host='localhost', port=8125, prefix='superset')
 ```
+
+If you get the Error `ImportError: cannot import name 'StatsdStatsLogger' from 'superset.stats_logger' (/path/to/superset/env/superset/stats_logger.py)`, you can edit the `except Exception` block to print error details:
+```python
+try:
+    from statsd import StatsClient
+
+    class StatsdStatsLogger(BaseStatsLogger):
+    ...
+except Exception as e:  # pylint: disable=broad-except
+    print(e)
+```
+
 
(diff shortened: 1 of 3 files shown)


docs/docs/installation/pypi.mdx
```diff
@@ -22,6 +22,13 @@ level dependencies.
 
 **Debian and Ubuntu**
 
+Ubuntu **24.04** uses python 3.12 per default, which currently is not supported by superset. You need to add a second python installation of 3.11 and install the required additional dependencies.
+```bash
+sudo add-apt-repository ppa:deadsnakes/ppa
```
sfirke, line 27: I was unfamiliar with this repo so looked it up and found it to be credible: https://askubuntu.com/questions/1398568/installing-python-who-is-deadsnakes-and-why-should-i-trust-them

docs/docs/configuration/event-logging.mdx
```diff
@@ -51,11 +51,24 @@ if desired. Most endpoints hit are logged as
...
 
 To setup StatsD logging, it’s a matter of configuring the logger in your `superset_config.py`.
+If not already present, you need to ensure that the `statsd`-package is installed in superset's python environment.
 
 ```python
 from superset.stats_logger import StatsdStatsLogger
 STATS_LOGGER = StatsdStatsLogger(host='localhost', port=8125, prefix='superset')
 ```
 
+If you get the Error `ImportError: cannot import name 'StatsdStatsLogger' from 'superset.stats_logger' (/path/to/superset/env/superset/stats_logger.py)`, you can edit the `except Exception` block to print error details:
```
sfirke, line 61: What file is being edited here? It's unclear to me, sorry.

docs/docs/configuration/event-logging.mdx
```diff
@@ -51,11 +51,24 @@ if desired. Most endpoints hit are logged as
 well as key events like query start and end in SQL Lab.
 
 To setup StatsD logging, it’s a matter of configuring the logger in your `superset_config.py`.
+If not already present, you need to ensure that the `statsd`-package is installed in superset's python environment.
```
sfirke, line 54: Nitpick: could you capitalize Superset throughout?

sfirke (approved):
I have two tiny things (capitalize Superset and clarify what file to edit) but this LGTM. @125m125 do you want to make those tweaks? That would be ideal but if it gets stale I'll just merge it.

sfirke:
> Suggestion: Maybe instead of suppressing the exception, superset could be adjusted to still create the StatsdStatsLogger class but it throws the caught exception on init so the users can see the actual problem?

@125m125 this is fine with me. Would you like to do that on a separate PR after this merges, or add it to this PR (in which case would you take out the docs about modifying the logger function) ?

rusackas:
Approving CI 🤞

rusackas:
CC @sadpandajoe @mistercrunch  - feel free to merge if this doesn't affect anything we're worried about ;)

docs/docs/installation/pypi.mdx
```diff
@@ -94,14 +101,9 @@ These will now be available when pip installing requirements.
...
-We highly recommend installing Superset inside of a virtual environment. Python ships with
-`virtualenv` out of the box. If you're using [pyenv](https://github.com/pyenv/pyenv), you can install [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv). Or you can install it with `pip`:
-
-```bash
-pip install virtualenv
-```
+We highly recommend installing Superset inside of a virtual environment.
 
-You can create and activate a virtual environment using:
+You can create and activate a virtual environment using the following commands. Ensure you are using a compatible version of python. You might have to explicitly use for example `python3.11` instead of `python3`.
```
mistercrunch, line 106: NOTE: we started using and recommending `uv`. 

Stepping back, it feels like this section of the docs is hard to keep up to date and hoping in the future we can say "use docker compose for dev/POCs, and use helm for k8s, if you're going on metal, we recommend you align with our helm/docker set up"

It's kind of crazy that we have to explain to people how to set up a virtualenv in the Superset doc. "Set up a python env, we don't care how you do it!"

rusackas: Did we need to revise anything on the topic for this PR, or should we merge as-is?

mistercrunch: I'm fine with it as is, just thinking eventually we may get rid of some of the docs here / simplify

Issue #32436, opened 2025-03-01
Can't sort bar charts based on value of bar


pankajsoni22:
### Bug description

The bug is that
1. Go to a bar chart (may be simple- x axis lable, y axis value)
2. Try to sort the bar in ascending and descending order
3. We can't do so.

Superset version 5.0.0

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [ ] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [ ] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

michael-s-molina:
Hi @pankajsoni22. Could you record a video showing the problem? I used a simple chart as you described but was not able to reproduce the problem.

https://github.com/user-attachments/assets/ee93cf75-f829-4f38-962b-4cf581777273

sfirke:
@pankajsoni22 can you add a video example? And does this bug exist in 4.1.x or is it a new regression in 5.0.0?

rusackas:
Closing this issue as we were unable to reproduce it, and there's been no response to requests for more details.

@pankajsoni22 - if you're still experiencing this issue on a recent version of Superset, please reopen with:
- A video showing the problem
- Specific reproduction steps
- Confirmation of which version(s) this affects

Thank you!

PR #32904, opened 2025-03-28
feat: don't pivot when we don't need to

betodealmeida:
### SUMMARY

Pivoting without dimensions is an unnecessary operation, and worse, it breaks sorting.

(This is a simpler solution than https://github.com/apache/superset/pull/27616.)

Before:

Screenshot 2025-03-28 at 4 01 38 PM

After:

Screenshot 2025-03-28 at 4 04 12 PM

Note that when we have dimensions sorting the query doesn't make much sense. Instead, we can sort by the x-axis:

Screenshot 2025-03-28 at 4 10 57 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


betodealmeida:
Still need more testing.

michael-s-molina:
Might be related to https://github.com/apache/superset/issues/32436.

PR #32968, opened 2025-04-01
fix(pivot-table): Revert "fix(Pivot Table): Fix column width to respect currency config (#31414)"

justinpark:
### SUMMARY

#31414 has an issue where both the header and the long text cells are causing the chart area to inefficiently expand due to the nowrap setting. Therefore, this commit reverts the changes first and then finding a new approach that only applies to the currency text.

Before:
Screenshot 2025-04-01 at 1 42 31 PM

After:
Screenshot 2025-04-01 at 1 41 52 PM

### TESTING INSTRUCTIONS
Create a Pivot table
Select a column with a long text value

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila (approved):
Thanks for your help here! 🙏

PR #31934, opened 2025-01-20
chore: Extensions architecture POC

michael-s-molina:
### SUMMARY
This PR introduces new packages and functionality for introducing extensions as proposed in SIP  #34162. Example extensions that were used for validating the system have been moved to the following draft PR: #34726

This PR introduces the following packages:

**superset-frontend/packages/superset-core** 
Our package with UI components, utility functions, types, and APIs used by extensions.

**superset-core**
The host APIs used by backend extensions.

**superset-cli**
CLI commands to build, bundle and hot-rebuild extensions.

```
superset-extensions build     Builds extension assets.

superset-extensions bundle    Packages the extension into a .supx file.

superset-extensions dev       Provides hot-rebuilding of assets.
```

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
The **Dataset references** extension displays relevant information for SQL authors in SQL Lab:
- Datasets referenced in a SQL Query: This helps users discover the origin of the data, especially for long SQL queries.
- Owners of each referenced dataset: This is crucial for dealing with access permissions or quickly finding out who to contact in case of problems.
- Latest available partition: This is important for understanding data completeness issues.
- Estimated row count for each dataset: This helps users grasp the implications of joins.

https://github.com/user-attachments/assets/cbabaca1-ae67-4a65-859a-bc53a7ac3a0d

### TESTING INSTRUCTIONS
1 - Set the `ENABLE_EXTENSIONS` feature flag to `True`.

2 - Run `npm ci` on `superset-frontend`.

3 - Install the development dependencies by running `pip install -r requirements/development.txt`. This will pull in the new cli package that contains the `superset-extensions` command.

4 - Bundle an extension by going into its root folder and executing `superset-extensions bundle`. This will build the extension and generate a `.supx` bundle of the extension.

5 - Upload the `.supx` file to Superset using Swagger and the `extensions` endpoint.

6 - Go to `SQL` -> `SQL Lab` and check the extension.

For development mode:

1 -  Execute steps 1, 2 and 3 above.

2 - Configure Superset's `LOCAL_EXTENSIONS` variable like:

```
LOCAL_EXTENSIONS = [
    "./extensions/dataset_references",
]
```

3 - Run in development mode by going into the extension root folder and executing `superset-extensions dev`. This will trigger hot-rebuilding of your frontend assets. Backend assets are currently reloaded by the default reloading mechanisms of Superset using `superset run --reload`.

4 - Go to `SQL` -> `SQL Lab` and check the extension.

[...]

PR #32969, opened 2025-04-01
feat(charts): chart publishing UI

ethan-liong:
### SUMMARY

This MR introduces publishing for charts. 
- adds the publish/draft icon when viewing a chart
- adds publish/draft status + filter in chart list
- adds publish to model

At a glance, it doesn't really do much besides provide a UI indicator as to whether a chart is tagged as published or not, but I figure that this might be the start of creating extra additional custom permissions on top of charts that an owner might deemed published and ready for more eyes. This is really just to get a foot in the door towards maybe creating "private" charts.

image
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [x] Required feature flags: PUBLISH_CHARTS
- [x] Changes UI
  - [ x] Migration is atomic, supports rollback & is backwards-compatible
  - [x ] Confirm DB migration upgrade and downgrade tested


michael-s-molina:
Hi @ethan-liong. Thank you for the PR. This type of change generally needs a SIP where we can discuss what are the intended use cases and implications of introducing the concept of published/unpublished charts. It's also a good idea to think about the whole feature, given that if we merge the PR as is, we'll have a new status but no immediate applicability for it.

You can check https://github.com/apache/superset/issues/5602 for the SIP template.

mistercrunch:
I'm open to a SIP here, but don't think this is controversial, especially behind a feature flag. As long as it fits the implementation for dashboards, its in many ways is even more needed for Charts, given that there's an order of magnitude more charts than there are dashboards. 

Some clarifying questions (whether it's for the SIP or PR...)
- we should clarify whether published/draft affects access or RBAC (?) From my understanding it doesn't/shouldn't, meaning draft isn't "private" - other people can access your drafts if they desire, it simply curates the list view (where I believe you see your own drafts + what others have published (?))
- does publishing a dashboard implies that the charts its composed-of are (or should) be published (?) What if you add charts *after* publishing a dashboard? Is it confusing the see a published dashboard made out of "draft" charts (?)

JZ6:
> This seems like it's a great idea, but it does seem like it might need a SIP to answer any questions that come up. I'll seed the discussion with a few: • 

We have had this feature in our deployment of superset for 2 years now, so I may be able to answer some of these questions based on how we have handled this:

Will there be a migration to convert all pre-existing charts to "Published" when this launches? 

Ideally old charts should stay as draft, but maybe add a way for admins to bulk publish multiple/all charts at once?


• If a chart is un-published, does it still appear in dashboards it was placed in? If not, how does it fail gracefully? 
Perhaps a chart in a published dashboard can not be unpublished unless the dashboard is also in draft. We also have checks the other way, a dashboard can only be published if all charts within is published


Who has the RBAC permissions to publish/unpublish a chart? The owner(s) of the chart, and admins, I presume?
We use a seperate role with special publish perms, but yes only allowing owners and admins would make sense.

mistercrunch:
Could be nice and simple if we kept it as simple as "published means the object is discoverable by others, namely in list views (namely the CRUD views and various dropdowns). Draft objects, while they don't show up in lists views, are accessible by others if they happen to have the link, or say if they happen to open a dashboard composed of draft chart"

michael-s-molina:
My 2 cents:

- The number of questions raised in this thread and the number of open questions are a clear indication of a SIP.
- I don't think we should use feature flags to think about new features as we code. Feature flags introduce conditions into the core codebase, increase complexity, and maintainability. If we don't know yet the extent of this feature, I would suggest creating a feature branch. We're doing this with [Theming](https://github.com/apache/superset/pull/31590) and [Extensions](https://github.com/apache/superset/pull/31934) for example. 
- In my opinion, feature flags should be used when we already discussed the feature over a SIP and we use a feature flag to develop what was agreed. Not to think about the feature while modifying the core codebase.
- Introducing feature flags in PRs without any SIP was the reason for many poorly designed features that we had to re-implement is the past.

villebro:
@ethan-liong as part of [SIP-152](https://github.com/apache/superset/issues/32116), we're introducing the concept of Viewers to Charts and Dashboards. This means, that once rolled out, Draft/Published is redundant, as read only access will be provided by assigning subjects as Viewers. See my comment https://github.com/apache/superset/issues/32116#issuecomment-2643567384 , more specifically this bullet:
> - We replace all owner references with the new Subject model that can be a User, Group or Role. This covers Dashboards, Charts, Datasets and Alerts & Reports.
> - We introduce the Viewer property to Dashboard and Chart entities (for Alerts & Reports having a Viewer is not very useful). During the migration we map all Dashboard RBAC roles as Viewers and remove the DASHBOARD_RBAC feature flag.
> - We map Roles that have implicit access to Dashboards and Charts to their respective Viewer property, and remove the implicit dashboard/chart access control logic. However, for Dashboards that are in Draft state we leave the Viewer property empty. Finally we remove the "published" property from the Dashboard model, and remove the Draft/Published pill (good riddance!).

JZ6:
> @ethan-liong as part of [SIP-152](https://github.com/apache/superset/issues/32116), we're introducing the concept of Viewers to Charts and Dashboards. This means, that once rolled out, Draft/Published is redundant, as read only access will be provided by assigning subjects as Viewers. See my comment [#32116 (comment)](https://github.com/apache/superset/issues/32116#issuecomment-2643567384) , more specifically these bullets:
> 
> > * We replace all owner references with the new Subject model that can be a User, Group or Role. This covers Dashboards, Charts, Datasets and Alerts & Reports.
> > * We introduce the Viewer property to Dashboard and Chart entities (for Alerts & Reports having a Viewer is not very useful). During the migration we map all Dashboard RBAC roles as Viewers and remove the DASHBOARD_RBAC feature flag.
> > * We map Roles that have implicit access to Dashboards and Charts to their respective Viewer property, and remove the implicit dashboard/chart access control logic. However, for Dashboards that are in Draft state we leave the Viewer property empty. Finally we remove the "published" property from the Dashboard model, and remove the Draft/Published pill (good riddance!).

awesome! do you know which version of superset this is expected to be released in? thanks!

mistercrunch:
> This means, that once rolled out, Draft/Published is redundant

I think it's probably still relevant, it's really about the owner of the chart/dashboard saying "ok I want for this dashboard I created to be discoverable by others". That user may not be an Admin so they don't really have the rights to grant access to others...

rusackas:
This PR has been on hold pending SIP discussions. Converting to draft for now. The Views concept in SIP-152 may address some of these needs - feel free to re-engage when there's clarity on the path forward\!

rusackas:
Closing this for now since it hasn't seen any traction. Feel free to reopen or ping us if you want to rekindle the conversation. Thanks!

PR #32975, opened 2025-04-02
feat(charts): add subtitle option and metric customization controls

LevisNgigi:
feat(charts): add subtitle option and metric customization controls

### SUMMARY

This PR adds two key improvements to visualization controls:

1. Added subtitle functionality to big number charts visualizations.Moved subheader input previously in BigNumber to subtitle under customize tab.

2. Enhanced the explore view with controls that allow users to:
   - Rename comparison metrics (main, #, and delta values) in both big number with time comparison chart and table chart.
   - Toggle visibility of specific metric icon and columns
   - Persist these customizations when the visualization is saved

https://github.com/user-attachments/assets/bcfb363b-5d10-40a3-a71d-04a8d2daa066

### TESTING INSTRUCTIONS

1. **Big Number with Trendline**:
   - Create a new chart using the "Big Number with Trendline" visualization
   - Verify the subtitle option appears in the customize section.
   - Add text to the subtitle and confirm it displays correctly.
   -
2. **Big Number **:
   - Create a new chart using the "Big Number" visualization
   - Verify the subtitle option appears in the customize section.
   - Add text to the subtitle and confirm it displays correctly.

3. **Big Number with Time Comparisons** (requires CHART_PLUGINS_EXPERIMENTAL flag):
   - Enable the CHART_PLUGINS_EXPERIMENTAL feature flag
   - Create a chart using the "Big Number with Time Comparisons" visualization
   - Verify the subtitle option appears and functions correctly

4. **Metric Customization**:
   - Create a chart using "Table" and create a time comparison.
   - In the explore view, locate the new controls for renaming comparison metrics
   - Test renaming the main, #, and delta metrics
   - Test toggling visibility of specific columns
   - Save the chart and verify the customizations persist when reopened
  
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ x] Changes UI


superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -229,6 +229,37 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
...
+      const container = this.createTemporaryContainer();
+      document.body.append(container);
+      fontSize = computeMaxFontSize({
+        text: subtitle,
+        maxWidth: width * 0.9,
+        maxHeight,
+        className: 'subtitle-line',
+        container,
+      });
+      container.remove();
```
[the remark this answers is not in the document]
LevisNgigi: makes sense agreed.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -368,7 +407,12 @@ export default styled(BigNumberVis)`
 
     .subheader-line {
       line-height: 1em;
-      padding-bottom: 0;
+      padding-bottom: 0.3em;
+    }
+
+    .subtitle-line {
+    line-height: 1em;
+    padding-top: 0.3em;
```
geido, line 415: Can we indent these?

superset-frontend/src/explore/components/controls/ColumnConfigControl/ColumnConfigItem.tsx
```diff
@@ -73,18 +75,53 @@ export default memo(function ColumnConfigItem({
...
         }}
       >
-        <ColumnTypeLabel type={column.type} />
-        {column.name}
-        {/* TODO: Remove fa-icon */}
-        {/* eslint-disable-next-line icons/no-fa-icons-usage */}
-        <i
-          className="fa fa-caret-right"
+        <div
+          css={{
```
geido, line 79: Can we use the `css` with template literal approach for inline styles?

geido (commented):
Some first-pass comments. This seems great so far!

kgabryje:
@LevisNgigi Thanks for the contribution, it's looking good!
'm not sure about is reverting to the metric name if the label is left empty and display type icon is unchecked.
Also, if we use a saved metric with a label, the metric name displayed in time comparison when display type is unchecked is the metric's name (not "user friendly") instead of the label (see COUNT(*) metric displayed as count).

1 more thing - I think that the metric item is unnecessarily indented in the Customize columns control (see on the recording).



https://github.com/user-attachments/assets/7351d67e-07b2-4d9d-860c-fbb152645e90

kgabryje:
Oh and 1 more thing, not sure if related to your PR - the changes in Customize columns applied to the columns generated by time comparison (e.g. # count__1 month ago) have no effect on the table

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -873,6 +892,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
               className,
               value == null ? 'dt-is-null' : '',
               isActiveFilterValue(key, value) ? ' dt-is-active-filter' : '',
+              'test-c7w8t3',
```
kgabryje, line 895:
```suggestion
```

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -795,6 +813,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
             white-space: ${value instanceof Date ? 'nowrap' : undefined};
             position: relative;
             background: ${backgroundColor || undefined};
+            padding-left: ${column.isChildColumn ? '20px' : '8px'};
```
kgabryje, line 816: Can we use theme.gridUnits here?

superset-frontend/src/explore/components/controls/ColumnConfigControl/ColumnConfigItem.tsx
```diff
@@ -35,6 +36,57 @@ export default memo(function ColumnConfigItem({
...
+    position: 'absolute',
+    right: 3 * gridUnit,
+    top: 3 * gridUnit,
+    transform: 'translateY(-50%)',
+    gap: gridUnit,
+    color: colors.grayscale.light1,
+  });
+
+  const caretIconStyle = css({
+    fontSize: '14px',
```
kgabryje, line 85: Can we use font sizes from theme?

LevisNgigi:
> @LevisNgigi Thanks for the contribution, it's looking good! 'm not sure about is reverting to the metric name if the label is left empty and display type icon is unchecked. Also, if we use a saved metric with a label, the metric name displayed in time comparison when display type is unchecked is the metric's name (not "user friendly") instead of the label (see COUNT(*) metric displayed as count).
> 
> 1 more thing - I think that the metric item is unnecessarily indented in the Customize columns control (see on the recording).
> 
>  Screen.Recording.2025-04-09.at.13.49.11.mov

@kgabryje About the not user friendly being displayed, I fixed that in my last commit such that if label is left empty(unchecked) and no custom column name,column name remains empty(as suggested in design) instead of displaying metric name.

kgabryje:
Thanks for addressing the comments!
I noticed 3 problems in the table chart:
1. There's an additional empty cell in the group header
2. Instead of "Main" we display the metric label
3. When I try to collapse the group, it disappears and the only way to restore it is to rerun the query or refresh the page

I checked and those problems are not present on current master.

https://github.com/user-attachments/assets/7ccc450b-16c2-4d32-9b1b-06356b2a4c73

LevisNgigi:
> > @kgabryje For number 2, I was to remove text Main from the main column as it does not stand out too much.
> 
> Ahh I see so I suppose that empty cell is intentional? Maybe we could add the "time ago" suffix to the group header then for the distinction between the metric and the time comparison values, like `COUNT(*) 1 month ago`? And should we have a border between the metric header and `#`?
> 
> image

@kgabryje I have restored Main and all three issue previously mentioned were fixed.

kgabryje:
@LevisNgigi Thanks! Unless @geido has more comments, I think we're good to go

kgabryje (approved):

LevisNgigi:
> @LevisNgigi Thanks! Unless @geido has more comments, I think we're good to go

@kgabryje Thank you for the feedback as well.

LevisNgigi:
> Debugging something funky that happening with this PR, somehow it failed when it merged on `master` and digging in the last commit's run here https://github.com/apache/superset/actions/runs/14382934496/job/40331066514, it appears that it skipped the python-related test, presumably because of an issue with our logic that checks which files have been touched by the PR and triggers/skips based on whether it sees the backend/frontend getting touched.
> 
> I'm guessing the issue is around the logic checking for files touched by the PR, and specific to [maybe] having a merge-based workflow instead of a rebase-type workflow.
> 
> Regardless, tests fail here and it's a bit tricky to sort this out, could use some help here to fix `master`

@mistercrunch noticed the error and saw it was a CI issue after checking the logs.Another check for a recently merged pr was running concurrently and not sure if it might have caused that?Something I can do to help?

PR #32989, opened 2025-04-02
fix: use Input & TextArea instead of native <input> and <textarea>

mistercrunch:
Moving to using AntD so that the components are themed properly


superset-frontend/src/components/ImportModal/index.tsx
```diff
@@ -65,8 +66,8 @@ const StyledInputContainer = styled.div`
     }
   }
 
-  input,
-  textarea {
+  Input,
+  TextArea {
```
geido, line 70: I think this and the ones below were a greedy find and replace?

superset-frontend/src/features/annotations/AnnotationModal.tsx
```diff
@@ -70,7 +71,7 @@ const AnnotationContainer = styled.div`
   }
 
   textarea,
-  input[type='text'] {
+  Input {
```
geido, line 74: It seems to be intentional at this point? I might be surprised it this actually works :)

mistercrunch: ooops, used "Aider" here and it went rogue... styles not required anymore...

geido (approved):
A few nits. Stamping.

superset-frontend/src/SqlLab/components/ResultSet/index.tsx
```diff
@@ -406,8 +407,7 @@ const ResultSet = ({
             )}
           </ResultSetButtons>
           {search && (
-            <input
-              type="text"
+            <Input
               onChange={changeSearch}
               value={searchText}
               className="form-control input-sm"
```
geido, line 413: Wondering if we still need these class names. This might come from bootstrap

superset-frontend/src/components/Chart/DrillDetail/DrillDetailMenuItems.test.tsx
```diff
@@ -396,7 +396,7 @@ test('context menu for supported chart, dimensions, filter B', async () => {
   await expectDrillToDetailByDimension(filterB);
 });
 
-test('context menu for supported chart, dimensions, all filters', async () => {
+test.skip('context menu for supported chart, dimensions, all filters', async () => {
```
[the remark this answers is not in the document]
mistercrunch: oh good catch

findmaster969 (approved):

PR #20143, opened 2022-05-20
feat: Adds the "Select all" option to the Select component

michael-s-molina:
### SUMMARY
This PR aims to add the "Select all" feature to the Select component. Adding this feature to a synchronous Select is really simple. The problem arises when adding "Select all" to an asynchronous, paginated, searchable, and keep selected on top Select. We have many technical challenges and design considerations as you can see [here](https://ux.stackexchange.com/questions/43937/how-should-the-select-all-select-when-paginating-elements). Some of the requirements we need to address:
- Select all items when only some pages are loaded on the client-side. Selecting only the visible items was discarded because, in our context, the majority of times the user is actually selecting all values. We also don't want to force the user to load all items before being able to select them.
- Load new items as selected when "Selected all" is active 
- Unselect a set of items when "Select all" is active and not all pages have been loaded on the client-side
- Keep the order of selected items during user interactions
- Automatically "Select all" when all items have been selected and not all pages have been loaded on the client-side. One example here is when the user opens the Select, one page is loaded, the user clicks on "Select all", unselects one item, and selects the same item again. At this point, we should go back to the "Select all" mode even though we only have one page loaded.
- Handle the payload size. We shouldn't send all selected items to the server-side because we don't know how many can exist, which could result in huge payload size. What we need to do is to come up with a payload that represents the user selection in an efficient way. One example would be if a user selects all items and unselects a subset, we could send a value representing the "Select all" alongside the subset of unselected items.
- Write tests that are able to scroll the component, trigger new page loads, and test the above scenarios.
- Expand the storybook to account for the new requirements.
- We also need to split the component into two (sync and async). It's very clear that the behaviors are different and we can significantly reduce code complexity if we treat them as different components.
- Add support to the new payload format to the API endpoints

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
https://user-images.githubusercontent.com/70410625/169524505-ac812e6c-29ab-4671-a198-03600a02b94d.mov

### TESTING INSTRUCTIONS
TODO

### ADDITIONAL INFORMATION
- [x] Has associated issue: 
- https://github.com/apache/superset/issues/19426
- https://github.com/apache/superset/pull/19979
- [ ] Required feature flags:
- [x] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
[...]

PR #32996, opened 2025-04-03
feat(rls-frontend): Add new buttons to select/deselect all datasets/roles in RLS

bmaquet:
### SUMMARY
This PR introduces "Select All" and "Deselect All" buttons to the Table & Roles elements of the Row Level Security (RLS) configuration interface. This enhancement aims to improve user experience, particularly when managing a large number of tables or roles, by allowing bulk selection and deselection instead of requiring individual clicks for each item. The implementation adds two distinct buttons to trigger these actions, updating the selection state accordingly.

#### BEFORE
Screenshot 2025-04-03 at 12 59 20

#### AFTER
Screenshot 2025-04-03 at 13 02 44

### TESTING INSTRUCTIONS
1. Navigate to the Row Level Security configuration section.
2. Click on Datasets or Roles boxes
3. Observe the presence of the new "Select All" and "Deselect All" buttons.
4. If starting with no items selected, click "Select All". Verify that all relevant items (e.g., tables, roles) are now selected.
5. Click "Deselect All". Verify that all items are now deselected.
6. Manually select a subset of items.
7. Click "Deselect All". Verify all are deselected.
8. Manually select a subset of items.
9. Click "Select All". Verify all are selected.

### ADDITIONAL INFORMATION
- [ ] Has associated issue: *(Fill in if applicable, e.g., `Fixes #123`)*
- [ ] Required feature flags: *(Fill in if applicable)*
- [x] Changes UI
- [x] Introduces new feature or API *(Implicitly, via UI interaction)*

### DIFF
--- a/superset-frontend/src/features/rls/RowLevelSecurityModal.tsx
+++ b/superset-frontend/src/features/rls/RowLevelSecurityModal.tsx
@@ -343,8 +343,8 @@ function RowLevelSecurityModal(props: RowLevelSecurityModalProps) {
     [],
   );
 
-  const tableDropdownRender = useCallback(
-    menu => (
+  const createDropdownRender = useCallback(
+    (entityType: 'tables' | 'roles') => (menu: React.ReactNode) => (
       <>
         <StyledDropdownContainer>
           <StyledDropdownButton
@@ -355,16 +355,16 @@ function RowLevelSecurityModal(props: RowLevelSecurityModalProps) {
                 page_size: -1,
               });
               SupersetClient.get({
-                endpoint: `/api/v1/rowlevelsecurity/related/tables?q=${query}`,
+                endpoint: `/api/v1/rowlevelsecurity/related/${entityType}?q=${query}`,
               }).then(response => {
-                const allTables = response.json.result.map(
+                const allEntities = response.json.result.map(
                   (item: { value: number; text: string }) => ({
                     key: item.value,
                     label: item.text,
                     value: item.value,
                   }),
                 );
-                updateRuleState('tables', allTables || []);
+                updateRuleState(entityType, allEntities || []);
               });
             }}
           >
@@ -372,7 +372,7 @@ function RowLevelSecurityModal(props: RowLevelSecurityModalProps) {
           </StyledDropdownButton>
           <StyledDropdownButton
             type="button"
-            onClick={() => updateRuleState('tables', [])}
+            onClick={() => updateRuleState(entityType, [])}
           >
             {t('Deselect all')}
           </StyledDropdownButton>
@@ -380,7 +380,16 @@ function RowLevelSecurityModal(props: RowLevelSecurityModalProps) {
         {menu}
       </>
     ),
-    [],
+    [updateRuleState],
+  );
+
+  const tableDropdownRender = useMemo(
+    () => createDropdownRender('tables'),
+    [createDropdownRender],
+  );
+  const roleDropdownRender = useMemo(
+    () => createDropdownRender('roles'),
+    [createDropdownRender],
   );
 
   return (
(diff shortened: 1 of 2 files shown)


superset-frontend/src/features/rls/RowLevelSecurityModal.tsx
```diff
@@ -102,6 +102,28 @@ const StyledTextArea = styled(TextArea)`
   margin-top: ${({ theme }) => theme.gridUnit}px;
 `;
 
+const StyledDropdownContainer = styled.div`
```
rusackas, line 105: This PR adds a lot of styled elements. Do you think that in support of Theming efforts underway, we could either:
a) Just go with "vanilla" AntD components and minimize style overrides
b) Move these styles into the root component behind a new prop so that they're reusable rather than bespoke for this one instance?

bmaquet: Thanks for the comment @rusackas !
I wasn't aware of the new theming efforts. I can definitely do either of these. Which one would you suggest?

bmaquet: "Vanilla" AntD would look like this

Screenshot 2025-04-03 at 17 16 37
Screenshot 2025-04-03 at 17 16 46

rusackas:
@michael-s-molina @geido actually, didn't we have a prop for the Select component to add a "Select All" option, rather than adding these buttons?

michael-s-molina:
> @michael-s-molina @geido actually, didn't we have a prop for the Select component to add a "Select All" option, rather than adding these buttons?

Yes, it's the [allowSelectAll](https://github.com/apache/superset/blob/63c8bbf3eba965809edda27bf5a6b6a03a2462f4/superset-frontend/src/components/Select/Select.tsx#L94).

bmaquet:
> > @michael-s-molina @geido actually, didn't we have a prop for the Select component to add a "Select All" option, rather than adding these buttons?
> 
> Yes, it's the [allowSelectAll](https://github.com/apache/superset/blob/63c8bbf3eba965809edda27bf5a6b6a03a2462f4/superset-frontend/src/components/Select/Select.tsx#L94) which is available only to the sync select. Implementing Select All for asynchronous selects has many challenges as you can see in #20143.

Thanks for the context about challenges with AsyncSelect @michael-s-molina

What would be the suggested design decision here then? Go ahead with custom buttons for this use case without implementing Select All for the AsyncSelect component, or close the PR entirely and come back to it if/when the AsyncSelect challenges can be overcome?

mistercrunch:
Oh! about theming, there's been a fair amount of work on the `template_less` branch moving `<Select>` to antd v5 and should probably hold off here. Sorry this is pretty unusual for us as we rarely have massive feature branches like that one.

Would recommend either re-opening the PR against `template_less` or holding off until it's merge.

Since I'm here, and about design, note that the Select component has an `allowClear` prop, and that common implementation for "select all" is through an `(all)` special first element. Tagging @kasiazjc for design input

Sorry about the confusion - but let's try to avoid merge conflicts.

bmaquet:
> Oh! about theming, there's been a fair amount of work on the `template_less` branch moving `<Select>` to antd v5 and should probably hold off here. Sorry this is pretty unusual for us as we rarely have massive feature branches like that one.
> 
> Would recommend either re-opening the PR against `template_less` or holding off until it's merged.
> 
> ~Since I'm here, and about design, note that the Select component has an `allowClear` prop, and that common implementation for "select all" is through an `(all)` special first element. Tagging @kasiazjc for design input~ Oh I see now some of this has been covered with more context...
> 
> Sorry about the confusion - but let's try to avoid merge conflicts. As far as ETA I think we're almost there for merging the theming / antd-v5 branch, waiting for v5.0 to be officially voted through to merge into master

Thanks for the extra context. I'll close this PR and wait until the theming PR is merged to revisit.

PR #33002, opened 2025-04-04
feat: add bulk certify functionality to charts and dashboards

vedantprajapati:
### SUMMARY
This PR introduces the ability to bulk certify charts and dashboards. This feature allows users to apply certification information (certified by, certification details) to multiple resources simultaneously. Our company uses certification as a way to validate charts and dashboards before publishing. Being able to certify in bulk is very useful when users create many similar components (especially when you have thousands of charts and dashboards).

image
image
image

### TESTING INSTRUCTIONS
Select multiple charts/dashboards using bulk select and certify them. users can only certify if they have the canEdit permission.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI


JZ6:
awesome!

sadpandajoe:
@kasiazjc to take a look at the ui change

mistercrunch:
Thanks for the PR - useful feature! And it's great to see that expanding the bulk-update functionality is pretty straightforward. 

Thinking about UX/design - if/once we add a few extra bulk-action feature, it might make sense to refactor the buttons into an `Action` dropdown, but I think for now it's ok as is with just 3 buttons. Tagging @kasiazjc (designer) for input/feedback.

vedantprajapati:
@mistercrunch good point. On that note, at GEOTAB, we also have a separate bulk update owners feature as well. This was sought after from our users as a way to transfer ownership or to share their data for collaborating. We have them separate though because the bulk update owners has a few more checks we do on our end.

mistercrunch:
Right. Parameterized bulk actions probably brings us into slightly more complex design territories. Would involve dynamic controls based on the action that is selected. Could be good to have a mock of what that would look like.

Though from my perspective the current UX is "ok for now".

vedantprajapati:
> Could be good to have a mock of what that would look like.

Sure, let me see if i can make a separate PR for some other bulk functionalities we have

JZ6:
will this pr be merged in?

vedantprajapati:
are we good to merge this one for now then?

> Though from my perspective the current UX (as of this PR) is "ok for now".

kasiazjc:
> are we good to merge this one for now then?
> 
> > Though from my perspective the current UX (as of this PR) is "ok for now".

thanks @vedantprajapati and sorry for the long wait!  

I think the perfect case scenario would be to combine all actions in a dropdown as we actually have 4 actions @mistercrunch:
- delete
- certify
- export
- tag (which is not really visible) 

I would suggest using [this](https://ant.design/~demos/dropdown-demo-loading) component, the one with the caret down 
image

Primary action would be export and all of the other ones would be in the dropdown menu (certify, tag, delete) and it would be displayed next to "deselect all" instead of all of the buttons. 

Thoughts?

vedantprajapati:
@kasiazjc okay, so for the scope of this ticket we'll just have those 4 moved to a dropdown with default being export. if @mistercrunch is okay with it, I can look into implementing this! it will definitely come in handy in the future with https://github.com/apache/superset/pull/33040

mistercrunch:
SGTM!

vedantprajapati:
@mistercrunch I've made the change, tests broke because it's a UI change. Please verify and then the tests can be fixed. on another note, how can i fix https://github.com/apache/superset/actions/runs/15742296412/job/44370616138?pr=33002#step:4:4753

superset-frontend/src/components/ListView/ListView.tsx
```diff
@@ -20,12 +20,12 @@ import { t, styled } from '@superset-ui/core';
 import { useCallback, useEffect, useRef, useState, ReactNode } from 'react';
 import Alert from 'src/components/Alert';
 import cx from 'classnames';
-import Button from 'src/components/Button';
 import { Icons } from 'src/components/Icons';
 import IndeterminateCheckbox from 'src/components/IndeterminateCheckbox';
 import Pagination from 'src/components/Pagination';
 import TableCollection from 'src/components/TableCollection';
 import BulkTagModal from 'src/features/tags/BulkTagModal';
+import { Dropdown, Menu } from 'antd';
```
rusackas: I believe these should be imported from src/components, and not directly from AntD (this allows us to customize/expand/update/migrate the component as needed, application-wide).

rusackas:
Just noting that this seems to need a rebase... hope it's not too bad :)

superset-frontend/src/components/ListView/ListView.tsx
vedantprajapati: done :)

vedantprajapati:
> > are we good to merge this one for now then?
> > > Though from my perspective the current UX (as of this PR) is "ok for now".
> 
> thanks @vedantprajapati and sorry for the long wait!
> 
> I think the perfect case scenario would be to combine all actions in a dropdown as we actually have 4 actions @mistercrunch:
> 
> * delete
> * certify
> * export
> * tag (which is not really visible)
> 
> I would suggest using [this](https://ant.design/~demos/dropdown-demo-loading) component, the one with the caret down image
> 
> Primary action would be export and all of the other ones would be in the dropdown menu (certify, tag, delete) and it would be displayed next to "deselect all" instead of all of the buttons.
> 
> Thoughts?

Hey @kasiazjc @mistercrunch ,
I've added the component in superset now! let me know if this works for you and I can proceed with making the test fixes.

image

mistercrunch:
Fired up an ephemeral environment and tagging @yousoph and @kasiazjc for user review/acceptance. Still need a code review here, let me see if I can find the time to review today.

superset-frontend/packages/superset-ui-core/src/components/Dropdown/index.tsx
```diff
@@ -110,8 +110,61 @@ export const NoAnimationDropdown = (props: NoAnimationDropdownProps) => {
   );
 };
 
+const CustomDropdownButton = (props: ButtonProps & DropdownProps) => {
```
mistercrunch: Mmmh, not sure about this section - we're really trying to go vanilla on antd as much as possible. Are we not affecting all other Dropdowns here? I'd have to spend more time to really understand what's going on here. Digging a tiny bit into it, are we re-implementing packages/superset-ui-core/src/components/DropdownButton/index.tsx ?

mistercrunch: > are we re-implementing packages/superset-ui-core/src/components/DropdownButton/index.tsx 

I'm thinking that if that's the case, there's some meta questions about discoverability of existing components and developer docs. We really need to be cautious around making sure that our design-system and component-lib is used properly somehow.

Side note, but while working on the theming branch, I noticed that we have a number of instances of places in the codebase that recreates/restyle buttons for instance. I can totally see that it's easy to merge something like this, and realize at some point we have 3-4 flavors of dropdown buttons ...

vedantprajapati: I am unable to have the dropdown loading component (https://ant.design/components/dropdown) that @kasiazjc suggested without adding this customization. otherwise if we wanted to use what we already have, then we can have the dropdown just not have that arrow icon on the right and still provide options on hover.

superset-frontend/packages/superset-ui-core/src/components/Dropdown/index.tsx
```diff
@@ -110,8 +110,61 @@ export const NoAnimationDropdown = (props: NoAnimationDropdownProps) => {
...
+
+    // Button-specific props
+    children,
+    buttonStyle,
+    disabled,
+    ...rest
+  } = props;
+
+  return (
+    <Space.Compact
```
mistercrunch: as a note, I think we'll try to avoid doing things like this in component library, the spacing / layout of the component should be left to the caller, unless it's super clear that we always need spacing to be a certain way. If anything we should create or use things like a `ButtonGroup` or things like `ButtonSpace` if/when things get repetitive.

Keeping antd as vanilla as possible allows for working with the assumption that you can assume things are "generic antd", so or most of the antd docs, or AI-generate code based on AI's understanding of antd is accurate. "Oddly mutated components from antd" is probably confusing to everyone, people and AIs

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -117,10 +117,11 @@ describe('Charts list', () => {
       cy.get('input[type="checkbox"]:checked').should('have.length', 26);
       cy.getBySel('bulk-select-copy').contains('25 Selected');
       cy.getBySel('bulk-select-action')
-        .should('have.length', 2)
+        .should('have.length', 3)
```
sadpandajoe: While you're in here, I think it's probably better to move some of these into a component test. Right now these e2e tests don't really show that bulk actions are working.

sadpandajoe:
I think the button dropdown looks off, maybe it's not aligned?
Screenshot 2025-07-09 at 1 15 52 PM

superset-frontend/packages/superset-ui-core/src/components/Dropdown/index.tsx
sadpandajoe: I think to be more vanilla, maybe not having the hover would be ok? Looking at sql lab we have something similar with the save button and it's more of a click than a hover. This tends to also make the arrow more aligned as it's a caret than a triangle.

Screenshot 2025-07-09 at 1 16 04 PM

vedantprajapati: > I think to be more vanilla, maybe not having the hover would be ok? Looking at sql lab we have something similar with the save button and it's more of a click than a hover. This tends to also make the arrow more aligned as it's a caret than a triangle.

@sadpandajoe

Sure we can do that too, however similarly, it would just create another unnecessary component. Allowing the base dropdown to support these options makes a bit more sense to me and it works out of the box as long as we include the correct props. 

We're kind of just recreating a seperate component for this save carat and the run buttons in sqllab
https://github.com/apache/superset/blob/master/superset-frontend/src/SqlLab/components/SaveDatasetActionButton/index.tsx 


By default the button can look like any of the following. If we're going with base styles, the third one looks the best suited:
image

Let me see if i can reduce the intrusiveness of my code in the dropdown component

mistercrunch: Mmmmh, so `superset-frontend/packages/superset-ui-core/src/components/DropdownButton/index.tsx` doesn't work as expected? I just looked at it in storybook and it seems broken there. 

But I'd say we shouldn't use `antd.Dropdown.Button` directly ideally as it would bypass our own `<Button />` wrapper and potentially look different and diverge from other buttons in the app.

I'd say ideally `superset-frontend/packages/superset-ui-core/src/components/DropdownButton/index.tsx` should be fixed up to use our own `Button` instead of `Dropdown.Button` (assuming that's possible), and consistently using it across the codebase. 

Sorry to put you through this but we've been doing a lot of component library refactor as we upgraded from antd v4 to v5, and many components are not what they should be, or aren't used properly in the app. Trying to avoid adding yet another layer in there.

vedantprajapati: speak of the devil I was looking into just that 😆  I think i got the DropdownButton component to work. let me touch up the spacing and hopefully this will be better. fyi this doesnt make any change to the core superset ui components

image

mistercrunch: nice! these little adjustments on the component lib is how we get back on track! Guessing `DropdownButton` should align with our `Button` component which is a pretty mutated version of the native antd, with its own props like `buttonStyle`, `buttonSize`, ... There's probably more simplification needed on that component to realign it to antd-v5, but we'll save that for another PR.

mistercrunch: If you can fix the storybook for `DropdownButton` to not crash when hovering it you get extra points

vedantprajapati:
> I think the button dropdown looks off, maybe it's not aligned? Screenshot 2025-07-09 at 1 15 52 PM

image

Fixed

yousoph:
Thanks for the contribution! 

I took a look at the ephemeral env from a couple days back and I noticed that the certification doesn't always update right away after it's been applied 

Here I had selected all but 2 dashboards bulk selected + certified but only one of them updates right away (before I refresh the page). I tried it with 2 charts and the page looks like it's updating but the certification icon doesn't appear. 
If I refresh the page, everything works as expected but the initial state is a bit confusing
image

vedantprajapati:
> I took a look at the ephemeral env from a couple days back and I noticed that the certification doesn't always update right away after it's been applied


Hey @yousoph I'm unable to replicate this. the certification icon appears automatically when i do it.

image

image

vedantprajapati:
> @mistercrunch Ephemeral environment spinning up at http://35.90.255.222:8080. Credentials are 'admin'/'admin'. Please allow several minutes for bootstrapping and startup.

is it possible to respin up the environment with the latest changes?

vedantprajapati:
> Thanks for the contribution!
> 
> I took a look at the ephemeral env from a couple days back and I noticed that the certification doesn't always update right away after it's been applied

After looking into it a little more. it seems like I didn't notice that the refreshData() and onHide() are called before the API
  requests actually complete which is why it only completely updates on browser refresh. let me push a fix @yousoph

yousoph:
Looks like the certification state is updating correctly now :)

vedantprajapati:
> Looks like the certification state is updating correctly now :)

let me just fix up the tests and build now

mistercrunch:
Looks like CI is giving you some troubles, let us know if you need help!

vedantprajapati17:
> Looks like CI is giving you some troubles, let us know if you need help!

@mistercrunch seems like i got the frontend tests to work, but somethings up with e2e. Can I get some help here?

eschutho:
Hi @vedantprajapati I just restarted the E2E tests for you.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -94,12 +94,57 @@ describe('Charts list', () => {
     });
   });
 
+  describe('list mode', () => {
+    before(() => {
+      visitChartList();
+      setGridMode('list');
+    });
+
+    it('should load rows in list mode', () => {
```
sadpandajoe: These should not be in e2e tests. There are no e2e actions, we're just checking if headers are showing up.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -94,12 +94,57 @@ describe('Charts list', () => {
...
+      cy.getBySel('sort-header').eq(1).contains('Name');
+      cy.getBySel('sort-header').eq(2).contains('Type');
+      cy.getBySel('sort-header').eq(3).contains('Dataset');
+      cy.getBySel('sort-header').eq(4).contains('On dashboards');
+      cy.getBySel('sort-header').eq(5).contains('Owners');
+      cy.getBySel('sort-header').eq(6).contains('Last modified');
+      cy.getBySel('sort-header').eq(7).contains('Actions');
+    });
+
+    it('should bulk select in list mode', () => {
```
sadpandajoe: Same as above statement, you'll probably get more value moving these over to component tests instead of e2e tests.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -94,12 +94,57 @@ describe('Charts list', () => {
...
+      cy.getBySel('bulk-select-action').should('not.exist');
+    });
+  });
   describe('card mode', () => {
     before(() => {
       visitChartList();
       setGridMode('card');
     });
 
+    it('should load rows in card mode', () => {
```
sadpandajoe: Same with this one, we are just loading things on the page but there is no e2e action.

superset-frontend/cypress-base/cypress/e2e/chart_list/list.test.ts
```diff
@@ -94,12 +94,57 @@ describe('Charts list', () => {
...
+      cy.getBySel('listview-table').should('not.exist');
+      cy.getBySel('styled-card').should('have.length', 25);
+    });
+
+    it('should bulk select in card mode', () => {
```
sadpandajoe: Same as the other tests.

superset-frontend/cypress-base/cypress/e2e/dashboard_list/list.test.ts
```diff
@@ -94,12 +94,15 @@ describe('Dashboards list', () => {
         .should('be.checked')
         .should('have.length', 6);
       cy.getBySel('bulk-select-copy').contains('5 Selected');
-      cy.getBySel('bulk-select-action')
-        .should('have.length', 2)
-        .then($btns => {
-          expect($btns).to.contain('Delete');
-          expect($btns).to.contain('Export');
-        });
+
```
sadpandajoe: I'd convert all these over to a component test.

superset-frontend/cypress-base/cypress/e2e/dashboard_list/list.test.ts
```diff
@@ -197,7 +215,19 @@ describe('Dashboards list', () => {
       cy.getBySel('table-row').eq(1).contains('4 - Sample dashboard');
       cy.get('[data-test="table-row"] input[type="checkbox"]').eq(0).click();
       cy.get('[data-test="table-row"] input[type="checkbox"]').eq(1).click();
-      cy.getBySel('bulk-select-action').eq(0).contains('Delete').click();
+
+      // Try to access Delete action (may not be available depending on permissions)
+      cy.getBySel('bulk-select-action').trigger('mouseenter');
+      // Check if Delete is available in the dropdown
+      cy.get('body').then($body => {
+        if ($body.text().includes('Delete')) {
```
sadpandajoe: Should not use if statements in tests since it may never hit one of the statements.

superset-frontend/cypress-base/cypress/e2e/dashboard_list/list.test.ts
```diff
@@ -182,7 +188,19 @@ describe('Dashboards list', () => {
 
       cy.getBySel('styled-card').eq(0).contains('1 - Sample dashboard').click();
       cy.getBySel('styled-card').eq(1).contains('2 - Sample dashboard').click();
-      cy.getBySel('bulk-select-action').eq(0).contains('Delete').click();
+
+      // Try to access Delete action (may not be available depending on permissions)
+      cy.getBySel('bulk-select-action').trigger('mouseenter');
+      // Check if Delete is available in the dropdown
+      cy.get('body').then($body => {
+        if ($body.text().includes('Delete')) {
```
sadpandajoe: Should not use if statements in tests since it may never hit one of the statements.

sadpandajoe (changes requested):
Remove unnecessary e2e tests. Move them to either component or unit tests if needed.

rusackas:
Thanks for sticking with this @vedantprajapati. @sadpandajoe's note still stands — let's move those e2e tests to component/unit tests. It's gone conflicting, so a rebase plus a green CI build would mean we can take another look.

rusackas:
@vedantprajapati happy to rebase this, but the standing ask is still to move those Cypress e2e cases over to component/unit tests — can you handle that part? Then we'll take another look with green CI.

rusackas:
Closing this out. Asked twice (June 17 and June 24) for a rebase plus moving the Cypress e2e cases to component/unit tests, and it's been quiet since. The bulk-certify feature itself is solid, the UX @kasiazjc landed on and the testing from @yousoph and @mistercrunch back that up, so if you want to pick it back up, a fresh PR off current `master` with those e2e cases converted would get another look. Thanks for sticking with this as long as you did, @vedantprajapati.

PR #33007, opened 2025-04-05
fix: Ignore 404 errors in fetchFavoriteStatus for deleted dashboards

UncoveredBread:
Fixes #32533. Prevents the "There was an issue fetching the favorite status of this dashboard" toast when loading a dashboard after deleting another. Added a check to silently ignore 404 responses from `/api/v1/dashboard/<id>/favorite_status`. Reproduced with `latest-dev`; couldn’t test locally due to system constraints—please verify.

### DIFF
--- a/superset/examples/helpers.py
+++ b/superset/examples/helpers.py
@@ -14,6 +14,34 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+"""Helpers for loading Superset example datasets.
+
+All Superset example data files (CSV, JSON, etc.) are fetched via the
+jsDelivr CDN instead of raw.githubusercontent.com to avoid GitHub API
+rate limits (60 anonymous requests/hour/IP).
+
+jsDelivr is a multi‑CDN front for public GitHub repos and supports
+arbitrary paths including nested folders. It doesn’t use the GitHub REST API
+and advertises unlimited bandwidth for open-source use.
+
+Example URL::
+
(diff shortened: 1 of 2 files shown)


Vitor-Avila:
I'm curious to get thoughts from @geido and @kgabryje here. If the issue happens because when accessing dashboard B after deleting A we're still trying to get the favorite status from A, should we identify why the dashboard state is still trying to fetch data from a previously accessed dashboard?

I understand the idea for this fix here and if that's the case it should def prevent the error, but I'm wondering if there could be other metadata in the `dashboardState` that could be getting pulled from an incorrect dashboard?

superset-frontend/src/dashboard/actions/dashboardState.js
```diff
@@ -98,23 +98,25 @@ export function toggleFaveStar(isStarred) {
   return { type: TOGGLE_FAVE_STAR, isStarred };
 }
 
-export function fetchFaveStar(id) {
-  return function fetchFaveStarThunk(dispatch) {
+export function fetchFavoriteStatus(dashboardId) {
+  return dispatch => {
     return SupersetClient.get({
-      endpoint: `/api/v1/dashboard/favorite_status/?q=${rison.encode([id])}`,
+      endpoint: `/api/v1/dashboard/${dashboardId}/favorite_status`,
```
geido, line 104: This does not seem right

geido (changes requested):
I think this is potentially breaking the functionality instead of fixing it. I agree with @Vitor-Avila we need to fix the root cause. Thank you!

PR #32959, opened 2025-04-01
refactor(Collapse): Upgrade Collapse to Antd5

DamianPendrak:
### SUMMARY

Fixes #32542
##### Before
Screenshot 2025-04-04 at 13 57 16
##### After
Screenshot 2025-04-04 at 13 55 34

##### Before
Screenshot 2025-04-01 at 08 31 40
##### After
Screenshot 2025-04-04 at 14 04 12

##### Before
Screenshot 2025-04-01 at 08 32 44
##### After
Screenshot 2025-04-04 at 14 04 52

##### Before
Screenshot 2025-04-04 at 14 10 27
##### After
Screenshot 2025-04-04 at 14 06 57

##### Before
Screenshot 2025-04-04 at 14 14 13
##### After
Screenshot 2025-04-04 at 14 16 38

##### Before
Screenshot 2025-04-04 at 17 26 26
##### After
Screenshot 2025-04-04 at 14 18 53

##### Before
Screenshot 2025-04-04 at 16 45 43
##### After
Screenshot 2025-04-04 at 16 54 04

##### Before
Screenshot 2025-04-04 at 17 28 43
##### After
Screenshot 2025-04-04 at 17 01 19

##### Before
Screenshot 2025-04-04 at 17 30 25
##### After
Screenshot 2025-04-04 at 17 02 27

##### Before
Screenshot 2025-04-04 at 17 40 21
##### After
Screenshot 2025-04-04 at 17 38 37

##### Before
Screenshot 2025-04-04 at 17 31 19
##### After
Screenshot 2025-04-04 at 17 33 43

### ADDITIONAL INFORMATION

- [x] Has associated issue: #32542
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #33038, opened 2025-04-08
fix: upgrade tabs in database modal

DamianPendrak:
This is post merge fix for #32959 

### SUMMARY

The Tabs component had custom styles and did not use items prop to display the items, which caused Collapse component to not occupy the full width

#### Before
image

#### After
Screenshot 2025-04-08 at 15 45 39

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
I think in this context we'd like for the `<Collapse />` component to **not** have rounded corners. This is a fairly common use case (I think we have other collapse-in-modal in other places) and the component should either offer a prop like `roundedCorners={false}`, or a `modalMode={true}`. I'm not sure if there are other stylistic adjustment we want to make in that mode. @kasiazjc (designer) can inform whether that's just a rounded corner issue or if we want to expand (pun-intended!) on this mode.

DamianPendrak:
Added the modalMode. It removes rounded corners and side borders

Screenshot 2025-04-09 at 17 07 00

kasiazjc:
Hi, thanks for working on this, and thanks @mistercrunch for driving the changes! 

One thing that I can see is that the caret is not centered to the title of the section, but this might be component problem. Is there anything we can do with it? 

image

One more thing, that I might not be a part of the PR is that the tabs do not have paddings, and they should have some on the left: 
image

DamianPendrak:
I can remove the margin from the text elements:
Screenshot 2025-04-09 at 17 42 30

or center it vertically, but only relative to the whole label. That would require some custom styles for the Collapse component, which can be harder to maintain.
Screenshot 2025-04-09 at 17 42 55

superset-frontend/src/features/alerts/components/ValidatedPanelHeader.tsx
```diff
@@ -38,18 +39,29 @@ const ValidatedPanelHeader = ({
...
-        <span>{t(title)}</span>
+    <div>
+      <Space size="small" data-test={testId}>
+        <h5
+          style={{
+            fontSize: theme.fontSizeLG,
+          }}
+        >
+          {t(title)}
+        </h5>
```
msyavuz, line 50: I think this is a good opportunity to use our Typography component. Something like `<Title level={5}>`

superset-frontend/src/features/databases/DatabaseModal/index.tsx
```diff
@@ -1882,137 +1880,166 @@ const DatabaseModal: FunctionComponent<DatabaseModalProps> = ({
...
-                    value: target.value,
-                  })
-                }
-                conf={conf}
-                testConnection={testConnection}
-                testInProgress={testInProgress}
-              >
-                <SSHTunnelSwitchComponent
-                  dbModel={dbModel}
+        items={[
```
msyavuz, line 1888: Just a heads up this will have merge conflicts with https://github.com/apache/superset/pull/33057

msyavuz (commented):
Thanks for working on this! You have some ci issues to fix and there will be some merge conflicts but it looks good otherwise.

kasiazjc:
> I can remove the margin from the text elements: Screenshot 2025-04-09 at 17 42 30
> 
> or center it vertically, but only relative to the whole label. That would require some custom styles for the Collapse component, which can be harder to maintain. Screenshot 2025-04-09 at 17 42 55

Thanks! I think the top version looks good if this makes better sense to implement

mistercrunch:
Please rebase and install/run pre-commit as it appears there are issues at that level. https://superset.apache.org/docs/contributing/development/#git-hooks

msyavuz (approved):
I added the modalMode to FiltersConfigForm as well. Looks good to me.

PR #33040, opened 2025-04-08
feat: add bulk update owner functionality to charts and dashboards

vedantprajapati:
### SUMMARY
This PR introduces the ability to bulk certify charts and dashboards. This feature allows users to update the owners to multiple resources simultaneously. Our company uses this as a way to transfer ownership and collaborating on charts and dashboards. Being able to update owners in bulk is very useful when someone is leaving the company and needs to have someone else take over their responsibilities.

image
image
image

### TESTING INSTRUCTIONS
Select multiple charts/dashboards using bulk select and certify them. users can only update owners if they have the canEdit permission.

### ADDITIONAL INFORMATION

Relates to: 
https://github.com/apache/superset/pull/33002
https://github.com/apache/superset/pull/33002#issuecomment-2784577792 

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI


vedantprajapati:
cc @mistercrunch

rusackas:
If you select four charts/dashboards with four different owners, does the UI here _overwrite_ the existing owners? Or display each of those owners (relevant to 1/4 assets) in the UI? Or does it allow you to select and _upsert_ owners?

rusackas:
CC @kasiazjc - I wonder how many buttons it's sane to add before we need a dropdown or something?

Relaying a comment from @michael-s-molina that at some point we should open a holistic SIP about how we handle expanding bulk actions, which is inevitable in the future as we add themes and who knows what else.

vedantprajapati:
> If you select four charts/dashboards with four different owners, does the UI here _overwrite_ the existing owners? Or display each of those owners (relevant to 1/4 assets) in the UI? Or does it allow you to select and _upsert_ owners?

it calls a put request for all of them. You'd have to be an owner and it would overwrite the existing owners with what you define. 

```
   Promise.all(
      selected.map(item => {
        const url = `/api/v1/${resourceName}/${item.id}`;
        const payload = {
          owners: owners.map(owner => owner.id),
        };
        return SupersetClient.put({
          url,
          headers: { 'Content-Type': 'application/json' },
          jsonPayload: payload,
        });
      }),
    )
    ```

JZ6:
> CC @kasiazjc - I wonder how many buttons it's sane to add before we need a dropdown or something?
> 
> Relaying a comment from @michael-s-molina that at some point we should open a holistic SIP about how we handle expanding bulk actions, which is inevitable in the future as we add themes and who knows what else.

For our version of Superset we grouped everything into 3 buttons. Bulk Publish, Bulk Edit, Bulk certify.

For the sake of this pr we called this button bulk update owners, but this button can easily be evolved to become bulk edit that allows users to bulk edit all properties of a dashboard or chart at the same time

vedantprajapati:
> CC @kasiazjc - I wonder how many buttons it's sane to add before we need a dropdown or something?
> 
> Relaying a comment from @michael-s-molina that at some point we should open a holistic SIP about how we handle expanding bulk actions, which is inevitable in the future as we add themes and who knows what else.

It would be great to support more bulk functionality.

Geotab also has bulk publish for charts/dashboards haha. 
We're working on this PR for publishing charts here https://github.com/apache/superset/pull/32969

vedantprajapati:
> CC @kasiazjc - I wonder how many buttons it's sane to add before we need a dropdown or something?
> 
> Relaying a comment from @michael-s-molina that at some point we should open a holistic SIP about how we handle expanding bulk actions, which is inevitable in the future as we add themes and who knows what else.

This is a screenshot of our instance. I wouldnt say it looks messy to have just the 3.

image

JZ6:
we can honestly combine certify and update together since they are both updating the chart property anyway

michael-s-molina:
Thank you for the PR and discussion @vedantprajapati. 

Bulk editing is a highly requested feature at Airbnb and I completely understand its value. I do think though that we should think about it holistically as multiple attributes of an asset can be bulk edited. I don't think adding specific actions for each attribute is sustainable given the number of attributes an asset can have. For example, these are some of the attributes that can be bulk edited:
- Owners (Editors and Viewers later)
- Certified
- Color scheme
- Cache timeout
- Theme
- etc

We also need to think how things will work when the values differ between selected assets. For example, let's say that we have the following:

Dashboard 1
- Owners: A, B, C

Dashboard 2
- Owners: B, D, F

Dashboard 3
- Owners: F, X, Y

How would bulk editing work? In the following screenshot, what would be the initial value of the Owners select? Do you start with an empty select and only allow adding new owners? Do override everything for all dashboards? What if you need to just add/remove one owner?

Screenshot 2025-04-08 at 16 45 30

vedantprajapati:
> How would bulk editing work? In the following screenshot, what would be the initial value of the Owners select? Do you start with an empty select and only allow adding new owners? Do you override everything for all dashboards? What if you need to just add/remove one owner?

It simply overwrites. We can look into appending owners maybe?

vedantprajapati:
> > We also need to think how things will work when the values differ between selected assets.
> 
> To keep it simple, I'd vote for a one-operation-per-assignment approach, so for owners the user would check a set of dashboard and assign owners to them, one operation per set of owners.
> 
> Note that it could be great to have multiple actions here `Add Owners`, `Set Owners`, and potentially `Remove Owners`.
> 
> Most CRUD framework offer this kind of functionality to some degree. Would be great if we can reuse Actions across entities too where that makes sense. Owner-management being a common use case here, where in theory could be backed by reusable stuff in the frontend/backend, like Mixins on `Model` and `RestApi` objects.

so for now, I can rename this button to Set Owners then?

michael-s-molina:
> It simply overwrites.

@vedantprajapati That would not work for us as the most common operation is to add or remove a single user. We can't override all the other owners that exist in each dashboard.

JZ6:
> > It simply overwrites.
> 
> @vedantprajapati That would not work for us as the most common operation is to add or remove a single user. We can't override all the other owners that exist in each dashboard.

@mistercrunch has a good suggestion here https://github.com/apache/superset/pull/33040#issuecomment-2787521952

once this bulk update framework is merged in, we can add more options like add a checkbox for "append only mode", or a new field called "bulk remove these users from owner"

If it becomes too messy then it may even make sense to split bulk update owners to its own button.

There is a lot of potential here

michael-s-molina:
> There is a lot of potential here

I agree. I propose we treat this PR as a feature branch where we discuss and implement the necessary refactors as @mistercrunch suggested and validate/test all use cases using an ephemeral environment before merging the PR. 

@villebro will probably have some thoughts as well as the author of [[SIP-126] Fine-grained access control to Superset entities](https://github.com/apache/superset/issues/28021).

vedantprajapati:
> > There is a lot of potential here
> 
> I agree. I propose we treat this PR as a feature branch where we discuss and implement the necessary refactors as @mistercrunch suggested and validate/test all use cases using an ephemeral environment before merging the PR.
> 
> @villebro will probably have some thoughts as well as the author of [[SIP-126] Fine-grained access control to Superset entities](https://github.com/apache/superset/issues/28021).

what should we do with https://github.com/apache/superset/pull/33002 then?

michael-s-molina:
> what should we do with https://github.com/apache/superset/pull/33002 then?

I think we should wait for this one to use the same pattern.

vedantprajapati:
> > There is a lot of potential here
> 
> I agree. I propose we treat this PR as a feature branch where we discuss and implement the necessary refactors as @mistercrunch suggested and validate/test all use cases using an ephemeral environment before merging the PR.
> 
> @villebro will probably have some thoughts as well as the author of [[SIP-126] Fine-grained access control to Superset entities](https://github.com/apache/superset/issues/28021).

any thoughts? @villebro @mistercrunch

rusackas:
Is anyone working on testing this? It's been untouched for about half a year, so I'm inclined to mark it as draft, but would love to see some forward momentum if I'm able to help facilitate anything here.

rusackas:
Marking this as draft for now, but hoping we can drive it to completion. Thanks again @vedantprajapati! 

Per the thread, we want to make sure this is aligned with SIP-126 (fine-grained access control) and validated on an ephemeral env before merge, but it's currently conflicting. Could you rebase and let us know if you can pick the refactor discussion back up? Leaving as draft in hopes that it picks up momentum.

rusackas:
@vedantprajapati this is a handy feature, but it's waiting on the bigger questions, aligning with SIP-126 access control and michael-s-molina's call for a holistic bulk-actions design, more than on a rebase. I'll close for now; worth reopening once that direction's settled.

PR #33043, opened 2025-04-08
feat(Select): Select all and Deselect all that works on visible items while searching

### LINKED ISSUE #33362: "Select all" option of select filters is hardcoded
### Bug description

"Select all" option is hardcoded. The option is not localized. There is a comment with TODO in Select.tsx but [similar issue](https://github.com/apache/superset/issues/30160) was closed. There are any plans to fix it?

### Screenshots/recordings

_No response_

### Superset version

4.1.2

### Python version

I don't know

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

msyavuz:
### SUMMARY

This pr aims to change our sync Select component to have `Select all` and `Deselect all` buttons that works on visible options. With this change users can filter what they want to select by searching for common patterns without clicking each item individually.

fixes: #33362

Some current screenshots from storybook:

image
image
image

Demo that showcases the functionality:

[select_demo.webm](https://github.com/user-attachments/assets/aab33a57-80d1-409b-b3ea-82ccd391080c)

### TESTING INSTRUCTIONS

Run the testing suite

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
+++ b/superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
@@ -328,12 +328,18 @@ test('Should schema select display options', async () => {
   });
   expect(select).toBeInTheDocument();
   userEvent.click(select);
-  expect(
-    await screen.findByRole('option', { name: 'public' }),
-  ).toBeInTheDocument();
-  expect(
-    await screen.findByRole('option', { name: 'information_schema' }),
-  ).toBeInTheDocument();
+  await waitFor(() => {
+    expect(screen.queryByText('Loading...')).not.toBeInTheDocument();
+  });
+  const publicOption = await screen.findByText('public', {
+    selector: '.ant-select-item-option-content',
+  });
+  expect(publicOption).toBeInTheDocument();
+
+  const infoSchemaOption = await screen.findByText('information_schema', {
+    selector: '.ant-select-item-option-content',
+  });
+  expect(infoSchemaOption).toBeInTheDocument();
 });
 
 test('Sends the correct db when changing the database', async () => {
--- a/superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
+++ b/superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
@@ -331,13 +331,11 @@ test('Should schema select display options', async () => {
   await waitFor(() => {
     expect(screen.queryByText('Loading...')).not.toBeInTheDocument();
   });
-  const publicOption = await screen.findByText('public', {
-    selector: '.ant-select-item-option-content',
-  });
+  const publicOption = await screen.findByRole('option', { name: 'public' });
   expect(publicOption).toBeInTheDocument();
 
-  const infoSchemaOption = await screen.findByText('information_schema', {
-    selector: '.ant-select-item-option-content',
+  const infoSchemaOption = await screen.findByRole('option', {
+    name: 'information_schema',
   });
   expect(infoSchemaOption).toBeInTheDocument();
 });
--- a/superset-frontend/src/components/Select/Select.test.tsx
+++ b/superset-frontend/src/components/Select/Select.test.tsx
@@ -26,7 +26,6 @@ import {
   within,
 } from 'spec/helpers/testing-library';
 import Select from 'src/components/Select/Select';
-import { SELECT_ALL_VALUE } from './utils';
 
 type Option = {
   label: string;
@@ -77,9 +76,6 @@ const defaultProps = {
   showSearch: true,
 };
 
-const selectAllOptionLabel = (numOptions: number) =>
-  `${String(SELECT_ALL_VALUE)} (${numOptions})`;
-
 const getElementByClassName = (className: string) =>
   document.querySelector(className)! as HTMLElement;
 
@@ -87,6 +83,9 @@ const getElementsByClassName = (className: string) =>
   document.querySelectorAll(className)! as NodeListOf<HTMLElement>;
 
 const getSelect = () => screen.getByRole('combobox', { name: ARIA_LABEL });
+
+const selectAllButtonText = (length: number) => `Select all (${length})`;
+const deselectAllButtonText = (length: number) => `Deselect all (${length})`;
 
 const findSelectOption = (text: string) =>
   waitFor(() =>
@@ -109,11 +108,6 @@ const findSelectValue = () =>
 
 const findAllSelectValues = () =>
   waitFor(() => [...getElementsByClassName('.ant-select-selection-item')]);
-
-const findAllCheckedValues = () =>
-  waitFor(() => [
-    ...getElementsByClassName('.ant-select-item-option-selected'),
-  ]);
 
 const clearAll = () => userEvent.click(screen.getByLabelText('close-circle'));
 
@@ -255,33 +249,22 @@ test('should sort selected to the top when in multi mode', async () => {
 
   await open();
   userEvent.click(await findSelectOption(labels[2]));
-  expect(
-    await matchOrder([selectAllOptionLabel(originalLabels.length), ...labels]),
-  ).toBe(true);
+  expect(await matchOrder(labels)).toBe(true);
 
   await reopen();
   labels = labels.splice(2, 1).concat(labels);
-  expect(
-    await matchOrder([selectAllOptionLabel(originalLabels.length), ...labels]),
-  ).toBe(true);
+  expect(await matchOrder(labels)).toBe(true);
 
   await open();
   userEvent.click(await findSelectOption(labels[5]));
   await reopen();
   labels = [labels.splice(0, 1)[0], labels.splice(4, 1)[0]].concat(labels);
-  expect(
-    await matchOrder([selectAllOptionLabel(originalLabels.length), ...labels]),
-  ).toBe(true);
+  expect(await matchOrder(labels)).toBe(true);
 
   // should revert to original order
   clearAll();
   await reopen();
-  expect(
-    await matchOrder([
-      selectAllOptionLabel(originalLabels.length),
-      ...originalLabels,
-    ]),
-  ).toBe(true);
+  expect(await matchOrder(originalLabels)).toBe(true);
 });
 
 test('searches for label or value', async () => {
@@ -634,15 +617,19 @@ test('finds an element with a numeric value and does not duplicate the options',
 test('render "Select all" for multi select', async () => {
   render(<Select {...defaultProps} mode="multiple" options={OPTIONS} />);
   await open();
-  const options = await findAllSelectOptions();
-  expect(options[0]).toHaveTextContent(selectAllOptionLabel(OPTIONS.length));
+  expect(
+    await screen.findByText(selectAllButtonText(OPTIONS.length)),
+  ).toBeInTheDocument();
 });
 
 test('does not render "Select all" for single select', async () => {
   render(<Select {...defaultProps} options={OPTIONS} mode="single" />);
   await open();
   expect(
-    screen.queryByText(selectAllOptionLabel(OPTIONS.length)),
+    screen.queryByText(selectAllButtonText(OPTIONS.length)),
+  ).not.toBeInTheDocument();
+  expect(
+    screen.queryByText(selectAllButtonText(OPTIONS.length)),
   ).not.toBeInTheDocument();
 });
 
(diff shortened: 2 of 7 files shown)


michael-s-molina:
@msyavuz Could you add a recording so we understand better your proposal?

geido:
We have involved @kasiazjc in the conversation as the current proposal might not work well with grouped options. We'll follow up.

michael-s-molina (changes requested):
I'm just requesting changes to make sure to review this once everything is ready. We might need to get alignment on what features will change and what features will not be supported anymore.

@msyavuz I know this is still a draft but it will be really important to have a very detailed PR description describing the changes. This component is heavily used in Superset and I'm concerned with possible feature removals.

michael-s-molina (changes requested):
I'm just requesting changes to make sure to review this once everything is ready. We might need to get alignment on what features will change and what features will not be supported anymore.

@msyavuz I know this is still a draft but it will be really important to have a very detailed PR description describing the changes. This component is heavily used in Superset and I'm concerned with possible features being removed.

EnxDev:
@msyavuz Great job! I was wondering if the text entered in the input field is supposed to disappear once an item is selected.

https://github.com/user-attachments/assets/d3598657-1b7f-42c1-987a-dc87810a9820

superset-frontend/src/components/Select/Select.tsx
```diff
@@ -239,6 +237,23 @@ const Select = forwardRef(
...
+    useEffect(() => {
+      const selectedValues = ensureIsArray(selectValue);
+      const selectable = visibleOptions.filter(
+        option =>
+          (!option.disabled || hasOption(option.value, selectedValues)) &&
+          !option.isNewOption,
+      ).length;
+      const deselectable = visibleOptions.filter(
+        option => !option.disabled && hasOption(option.value, selectedValues),
+      ).length;
```
EnxDev, line 249: Could it be done with just a single loop?

superset-frontend/src/components/Select/styles.tsx
```diff
@@ -138,3 +136,19 @@ export const StyledErrorMessage = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
 `;
+
+export const StyledBulkActionsContainer = styled(Space)`
+  ${({ theme }) => `
+    padding: ${theme.gridUnit * 2}px;
+    display: flex;
+    justify-content: center;
```
EnxDev, line 144: Should be styled(Flex) ?

msyavuz:
> Great job! I was wondering if the text entered in the input field is supposed to disappear once an item is selected.
Thanks! The behaviour of that is same in master so i don't think so.

superset-frontend/src/components/Select/styles.tsx
msyavuz: We don't wrap the flex component from ant design yet in master, i didn't want to use directly from ant design.

superset-frontend/src/components/Select/Select.tsx
msyavuz: Yes, good catch! `.reduce` is a more elegant solution.

superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
```diff
@@ -328,12 +328,18 @@ test('Should schema select display options', async () => {
...
-    await screen.findByRole('option', { name: 'public' }),
-  ).toBeInTheDocument();
-  expect(
-    await screen.findByRole('option', { name: 'information_schema' }),
-  ).toBeInTheDocument();
+  await waitFor(() => {
+    expect(screen.queryByText('Loading...')).not.toBeInTheDocument();
+  });
+  const publicOption = await screen.findByText('public', {
+    selector: '.ant-select-item-option-content',
```
geido, line 335: Let's look into this as I believe class names should be last resort

superset-frontend/src/components/Select/Select.test.tsx
```diff
@@ -699,7 +662,7 @@ test('selects all values', async () => {
     />,
   );
   await open();
-  userEvent.click(await findSelectOption(selectAllOptionLabel(OPTIONS.length)));
+  userEvent.click(await screen.findByText(selectAllButtonText(OPTIONS.length)));
```
geido, line 665: Can we use a role instead?

superset-frontend/src/components/Select/Select.tsx
```diff
@@ -308,37 +310,32 @@ const Select = forwardRef(
...
     const handleOnDeselect: SelectProps['onDeselect'] = (value, option) => {
       if (Array.isArray(selectValue)) {
-        if (getValue(value) === getValue(SELECT_ALL_VALUE)) {
-          clear();
-        } else {
-          let array = selectValue as AntdLabeledValue[];
-          array = array.filter(
-            element => getValue(element) !== getValue(value),
+        let array = selectValue as AntdLabeledValue[];
+        array = array.filter(element => getValue(element) !== getValue(value));
```
geido, line 314: I think we do not need this to be re-assignable after these changes

geido (commented):
First pass

superset-frontend/src/components/Select/Select.tsx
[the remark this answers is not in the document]
msyavuz: It's not needed here, removed!

EnxDev:
> > Great job! I was wondering if the text entered in the input field is supposed to disappear once an item is selected.
> 
> Thanks! The behaviour of that is same in master so i don't think so.

@michael-s-molina what do you think about this? Do you think it's a bug? If so, maybe we could take note of it and consider changing it in the future, thanks.

superset-frontend/src/components/Select/Select.tsx
```diff
@@ -308,37 +310,32 @@ const Select = forwardRef(
...
 
+    const handleFilterOption = (search: string, option: AntdLabeledValue) =>
+      handleFilterOptionHelper(search, option, optionFilterProps, filterOption);
+
     const handleOnSearch = debounce((search: string) => {
       const searchValue = search.trim();
+      setIsSearching(!!searchValue);
+
+      let updatedOptions = selectOptions;
+
```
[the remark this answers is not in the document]
msyavuz: This is necessary because we are reassigning it

michael-s-molina:
> @michael-s-molina what do you think about this? Do you think it's a bug? If so, maybe we could take note of it and consider changing it in the future, thanks.

This was by design. Many times you're looking for values that have a prefix like `marketing_` and wish to select multiple values. We preserve the search to allow users to make their selections, and remove the search once the dropdown is closed.

EnxDev:
> > @michael-s-molina what do you think about this? Do you think it's a bug? If so, maybe we could take note of it and consider changing it in the future, thanks.
> 
> This was by design. Many times you're looking for values that have a prefix like `marketing_` and wish to select multiple values. We preserve the search to allow users to make their selections, and remove the search once the dropdown is closed.

When you click the 'Select All' button after typing, shouldn't the text in the search field disappear?
There shouldn't be any more values left to select, right?

michael-s-molina:
> When you click the 'Select All' button after typing, shouldn't the text in the search field disappear?

Correct.

EnxDev:
> > When you click the 'Select All' button after typing, shouldn't the text in the search field disappear?
> 
> Correct.

So, there are two things to keep in mind:

1. If all elements are selected by  "Select All' button", the text in the search field should disappear.
2. The "Deselect all" button doesn’t work in this case.  @msyavuz 

(The video was recorded on ephemeral environment)

https://github.com/user-attachments/assets/c239fed2-356e-4f75-bfb2-62a0dbeeeaf8

michael-s-molina:
> I don't think it is a good idea to clear input text this way.For me as a user having a stable dropdown is more important when using select all/deselect all to filter options. Clearing input when user selects all items punishes users when they make a mistake by making them search and deselect/select again with no apparent upside.

That's a good point @msyavuz. I agree that we should only clear the text when closing the dropdown.

superset-frontend/src/components/DatabaseSelector/DatabaseSelector.test.tsx
msyavuz: Done!

superset-frontend/src/components/Select/Select.test.tsx
msyavuz: In this case we are also checking for the number displayed on buttons

superset-frontend/src/components/Select/Select.tsx
msyavuz: Fixed!

geido:
I am not sure if this is caused by this PR but there is a weird behavior. Sometimes while searching the "No data" remains open but this dropdown has a creatable option so it should not do that. Also, the number of items to select when that happens is wrong


https://github.com/user-attachments/assets/6645aad4-3d99-4cd6-bceb-8280a3cb1d36

geido:
@michael-s-molina do you have more comments for this PR?

michael-s-molina:
> @michael-s-molina do you have more comments for this PR?

I'll wait for your reviews first and once you think it's good, I'll do a full review/test. @msyavuz Let me know when that is the case, or in other words, when you think this PR is ready to be merged.

msyavuz:
@michael-s-molina I think that is the case now. I would appreciate the review/test.

michael-s-molina:
1 - Can we reduce the actions padding or increase?

Screenshot 2025-04-16 at 16 36 29

2 - The Select all action is not counting new options.

Screenshot 2025-04-16 at 16 37 07

3 - I was also able to reproduce the No data when typing fast.

Screenshot 2025-04-16 at 16 44 15

msyavuz:
> 1 - Can we reduce the actions padding?
> Screenshot 2025-04-16 at 16 36 29
> 
> 2 - The Select all action is not counting and selecting new options.
> Screenshot 2025-04-16 at 16 37 07
> 
> 3 - I was also able to reproduce the No data when typing fast.
> Screenshot 2025-04-16 at 16 44 15

1) We can, done!
2) That was by design to prevent accidental new options when user is trying to filter with select/deselect all.
3) I think this is the debounce?

What concerned me in the screenshots is the Deselect all button not being disabled when no options are selected. I remember fixing this, did you test on branch latest?

kasiazjc:
> > 2 - The Select all action is not counting and selecting new options.
> 
> @kasiazjc I don't think this should be the case as we have no indication that the user intends to actually create that new option or just filtering at that point, but I am curious to hear your thoughts on this.

I agree with what @msyavuz said above too, I also think the primary expected action when typing in the field is filtering the values and not really creating a new filter in this case. So that's why we opted for selecting [all] of the existing values instead of selecting everything you see in the dropdown. 

We also wanted to add indicators for existing vs new values in the dropdown, to make it easier to understand, but ant-d does not support that currently. Even though this is the case, selecting existing ones is more intuitive/expected than creating a new value (that could have been random search phrase) in addition to select all action. Creating a new value should feel intentional, for example as right now - with hitting enter.

michael-s-molina:
> I agree with what @msyavuz said above too, I also think the primary expected action when typing in the field is filtering the values and not really creating a new filter in this case. So that's why we opted for selecting [all] of the existing values instead of selecting everything you see in the dropdown.

> We also wanted to add indicators for existing vs new values in the dropdown, to make it easier to understand, but ant-d does not support that currently.

@kasiazjc This makes sense. It would be better though if we had some indication for new values. Did we consider showing a text suffix for new items like `(+) John Doe` or `(new) John Doe` that is displayed only when the dropdown is open?

> What concerned me in the screenshots is the Deselect all button not being disabled when no options are selected. I remember fixing this, did you test on branch latest?

@msyavuz I used the latest [test environment](https://github.com/apache/superset/pull/33043#issuecomment-2809377937) given that I didn't see any commit after.

michael-s-molina:
@msyavuz I noticed you removed some tests for previous behaviors that don't exist anymore but there are no new tests. Some possible tests would include:
- Deselect counting, enable/disable, etc. 
- Select all counting with new items
- Select all counting with search
- Maybe we can try to add a test for the 'No data' bug too?

michael-s-molina (approved):
@msyavuz I'll be out until Tuesday and I don't want to block the progress in this PR. I'm approving to remove the request changes status. I would appreciate if we can add the test cases (select all counting with search) and also fix the 'No data' issue before merging.

Thank you for the amazing work you've done in this PR and for addressing the suggestions in a very constructive way. I'm glad you joined the committers pool :smiley:

@geido @kasiazjc I'll leave the final review and tests with you. Thanks for the discussions!

geido (approved):
LGTM! Since @msyavuz is working on migrating the Select on the theming branch and he will need to re-apply the No data fix and these updates in there, we will move the fix in a follow-up PR on top of the theming branch. Thanks all!

PR #33046, opened 2025-04-08
feat: add a title prop to the dashboard link in CRUD LIST view

### LINKED ISSUE #23929: Long dashboard titles are cut
## Screenshot

image

## Description

When the dashboard title is long, it is not fully shown in Dashboards tab, so users have to zoom out.

## Design input
It would be great if users could see the full dashboard titles no matter the zoom in their browser.

mistercrunch:
When dashboard titles are long, they get trunctated with `...` ellipsis stuff. This solution simply uses the html <a title=""/> prop so the link can be hovered and the full title viewed by the user. More details in the issue

closes https://github.com/apache/superset/issues/23929#issuecomment-2786912016

Screenshot 2025-04-08 at 1 18 33 PM

### DIFF
--- a/superset-frontend/src/pages/DashboardList/index.tsx
+++ b/superset-frontend/src/pages/DashboardList/index.tsx
@@ -329,7 +329,7 @@ function DashboardList(props: DashboardListProps) {
             },
           },
         }: any) => (
-          <Link to={url}>
+          <Link to={url} title={dashboardTitle}>
             {certifiedBy && (
               <>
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM

mistercrunch:
Quick note - might be worth doing the same for other key props in other list views, should be easy for anyone to take on. Could create issues as "good starter task" - whatever the label is for that.

PR #33048, opened 2025-04-08
fix: improve error type on parse error

justinpark:
### SUMMARY
When _sql_parse_ encounters a **SupersetParseError**, the logged _error_type_ is recorded as `QUERY_SECURITY_ACCESS_ERROR`, which is ambiguous. This commit updates the _error_type_ to `INVALID_SQL_ERROR` for clarity.

Before:
Screenshot 2025-04-08 at 4 41 12 PM

After:
Screenshot 2025-04-08 at 4 42 38 PM

### TESTING INSTRUCTIONS
Run a invalid sql that causes the sql_parse error like `SELECT * FROM table LIMIT 100'`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -272,7 +272,7 @@ class ParsedQuery:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
             raise SupersetSecurityException(
                 SupersetError(
-                    error_type=SupersetErrorType.QUERY_SECURITY_ACCESS_ERROR,
+                    error_type=SupersetErrorType.INVALID_SQL_ERROR,
                     message=__(
                         "You may have an error in your SQL statement. {message}"
                     ).format(message=ex.error.message),


michael-s-molina:
@betodealmeida You probably has more context on why this was a security exception previously so I'll wait for your review.

superset/sql_parse.py
```diff
@@ -272,7 +272,7 @@ def _extract_tables_from_sql(self) -> set[Table]:
             logger.warning("Unable to parse SQL (%s): %s", self._dialect, self.sql)
             raise SupersetSecurityException(
                 SupersetError(
-                    error_type=SupersetErrorType.QUERY_SECURITY_ACCESS_ERROR,
+                    error_type=SupersetErrorType.INVALID_SQL_ERROR,
```
betodealmeida, line 275: Thanks! The intent here was that we often need to extract tables in order to see if a query can run, but obviously we do for other reasons. Potentially we could add a try/except in the security manager to capture this error and convert it to a security error, but I don't think it's necessary.

betodealmeida (approved):

PR #33054, opened 2025-04-09
feat(Native Filters): Exclude Filter Values

amaannawab923:
### SUMMARY

This pr modifies the existing Inverse Selection feature , and makes it dynamic ,  Now if the Inverse Selection is enabled
it will show up a Select besides the filter which has 2 values , Is and Is not 
where 
IS: IN
IS NOT : NOT IN
and the user can switch between these .

It is backward compatible and will be true for  Existing Native Filters with Inverse Selection

[screen-capture (2).webm](https://github.com/user-attachments/assets/2a87efa5-f4c2-4756-8214-64648d3a3870)

It works in both Vertical and horizontal filter bar

Screenshot 2025-04-15 at 1 43 07 AM
Screenshot 2025-04-15 at 1 43 21 AM

Current Native Filter Control panel untouched with just inverse Selection Checkbox , which is being used to operate the is and is not Select for dynamic control 
Screenshot 2025-04-15 at 1 44 24 AM

### TESTING INSTRUCTIONS

To test this functionality:
Edit a dashboard and open the native filters panel.
Click Add/Edit Filters and choose the dataset and column you want to filter on.
Under Filter Settings, scroll down and enable the checkbox:
✅ Inverse Selection
Save the filter.

Now in the dashboard view, you will see a new select dropdown besides the native filter with 2 options 
is and is not where
IS: IN
IS NOT : NOT IN

Select a few values and have not in selected
Click Apply and verify that the charts exclude the selected values using the NOT IN logic.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/DropdownContainer/index.tsx
```diff
@@ -179,6 +179,39 @@ const DropdownContainer = forwardRef(
...
+        const { children } = container;
+        const childrenArray = Array.from(children);
+
+        const currentWidths = childrenArray.map(child => child.getBoundingClientRect().width);
+
+        // Update state with new widths
+        setItemsWidth(currentWidths);
+      }
+    };
+
```
geido, line 214: I am curious as of why all of this logic came into existence? What is this trying to solve?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -151,20 +169,21 @@ const HorizontalOverflowFormItem = VerticalFormItem;
...
-          FormItem: HorizontalOverflowFormItem,
+          FormItem : HorizontalOverflowFormItem,
           FilterControlTitleBox: HorizontalOverflowFilterControlTitleBox,
           FilterControlTitle: HorizontalOverflowFilterControlTitle,
         };
       }
       return {
         FilterControlContainer: HorizontalFilterControlContainer,
-        FormItem: HorizontalFormItem,
+        FormItem: (props: any) => <HorizontalFormItem {...props} showExcludeSelection={showExcludeSelection} />,
```
geido, line 186: Is there any reasons to pass showExcludeSelection explicitly? It seems this isn't necessary for all other settings, what makes this special?

superset-frontend/src/filters/components/Select/SelectFilterPlugin.test.tsx
```diff
@@ -71,15 +85,17 @@ describe('SelectFilterPlugin', () => {
...
-      // @ts-ignore
-      <SelectFilterPlugin
-        // @ts-ignore
-        {...transformProps({
-          ...selectMultipleProps,
-          formData: { ...selectMultipleProps.formData, ...props },
-        })}
-        setDataMask={setDataMask}
-      />,
+      <Provider store={store}>
```
geido, line 88: I don't think you need to pass the provider directly here? You can use the `useRedux` prop

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -322,10 +383,35 @@ export default function PluginFilterSelect(props: PluginFilterSelectProps) {
...
+          >
+            <Checkbox
+              dataTestAttribute="exclude-filter-checkbox"
+              checked={excludeFilterValues}
+              onChange={handleExcludeCheckboxChange}
+            />
+            <span data-test="exclude-filter-label">
+              {filterBarOrientation === FilterBarOrientation.Horizontal ? t('Exclude Values')  : t('Exclude Filter Values')}
+            </span>
+            <Tooltip
```
geido, line 405: I think the `IconTooltip` could work here?

superset-frontend/src/components/DropdownContainer/index.tsx
amaannawab923: Earlier the native filter width was constant at gridUnit * 41 so we needed to calculate the overflowing elements only when a native filter was added or removed... And the UseLayoutEffect hook doesnt run whenever an existing Native filter adds in or loses the Exclude filter values checkbox and also the item widths arent updated whenever we add in or remove the dynamic exclude checkbox

What happens when we remove this code ? 
Screenshot 2025-04-10 at 9 36 14 PM

I have added 6 filters which looks perfectly fine because the overflowing are being calculated properly but what happens if i add exclude filters checkbox on 2 of those without these resize observers 
Screenshot 2025-04-10 at 9 36 39 PM
The item widths arent updated , the overflowing elements arent recalculated and the 4th one overflows even when it should have gone inside the dropdown 


When i add the resize observers back , the calculations re run and the 4th one is pushed inside the dropdown , which is expected

Screenshot 2025-04-10 at 9 39 00 PM

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
amaannawab923: This particular setting is when the FormItem is part of horizontal list but not in overflowing dropdown and not in Vertical list 

In this case the Width of this native filter is frozen to gridUnit * 41 px but we need space for the exclude filters checkbox as well so this prop is used as Styled Components props 

`

  .ant-form-item-control {
    width: ${({ theme, showExcludeSelection }) =>
    // Requires twice the space 
      showExcludeSelection ? theme.gridUnit * 41 * 2 : theme.gridUnit * 41}px;
  }

  .ant-form-item-control-input-content {
  // both the elements needs to be in flex display
    ${({ showExcludeSelection }) => showExcludeSelection && 'display: flex;'}
  }

  .select-container {
    ${({ showExcludeSelection, theme }) =>
    // Since the select container takes 100% width by default it tries to takes the entire space of ant-form-item-control
    // hence we need to tell it to take same width as you did when you didnt have the exlude checkbox
      showExcludeSelection &&
      `
      width: ${theme.gridUnit * 41}px;
    `}
  }

  .exclude-filter {
  // and since its not below the Native filter but on the right side of it , we do not need a margin top 
    margin-top: 0px;
  }`

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -41,14 +43,21 @@ import { FilterBarOrientation } from 'src/dashboard/types';
 import { PluginFilterSelectProps, SelectValue } from './types';
 import { FilterPluginStyle, StatusMessage, StyledFormItem } from '../common';
 import { getDataRecordFormatter, getSelectExtraFormData } from '../../utils';
+import Checkbox from 'src/components/Checkbox/Checkbox';
+import { Tooltip } from 'src/components/Tooltip';
+import { Icons } from 'src/components/Icons';
+import { useSelector } from 'react-redux';
+import { RootState } from 'src/views/store';
+
+const EXCLUDE_FILTER_VALUES_TOOLTIP = (filterName: string) => `Check this box to exclude the selected ${filterName} values from the results instead of filtering them`
```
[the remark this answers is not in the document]
amaannawab923: Done ... Please check if its ok

superset-frontend/src/filters/components/Select/SelectFilterPlugin.test.tsx
amaannawab923: Done

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
amaannawab923: Screenshot 2025-04-10 at 10 55 26 PM

Tried using it .... But its a different svg altogether that doesnt match the shared designs ... The infotooltip svg has a single path whereas the one that is required by designs belongs to the chart control panel which has 2 paths and is able to take the correct color

michael-s-molina (changes requested):
Thank you for the PR @amaannawab923. This feature already exists with the `Inverse selection` option.


https://github.com/user-attachments/assets/0856d5ae-6a48-47af-bf63-1d8d6f023a49

We also explicitly add filter configurations to the modal instead of directly to the filter bar to avoid the complexity for the user. The inverse selection is just one option but you could also have `Dynamically search all filter values` or `Select filter value as default` as quick options which would totally pollute an already complex area. That's why the configurations are always a secondary thing.

amaannawab923:
> Thank you for the PR @amaannawab923. This feature already exists with the `Inverse selection` option.
> 
>  Screen.Recording.2025-04-10.at.17.05.47.mov 
> We also explicitly add filter configurations to the modal instead of directly to the filter bar to avoid the complexity for the user. The inverse selection is just one option but you could also have `Dynamically search all filter values` or `Select filter value as default` as quick options which would totally pollute an already complex area. That's why the configurations are always a secondary thing.

Thanks for the quick reply @michael-s-molina  , but these changes were a requirement from preset and were hence implemented

geido:
> Thank you for the PR @amaannawab923. This feature already exists with the `Inverse selection` option.
> 
> 
> https://github.com/user-attachments/assets/0856d5ae-6a48-47af-bf63-1d8d6f023a49
> 
> We also explicitly add filter configurations to the modal instead of directly to the filter bar to avoid the complexity for the user. The inverse selection is just one option but you could also have `Dynamically search all filter values` or `Select filter value as default` as quick options which would totally pollute an already complex area. That's why the configurations are always a secondary thing.

@michael-s-molina the requirement is to be able to exclude or include filter values dynamically at selection. The inverse selection forces the user to decide the configuration upfront, often requiring to create two filters of the same column to achieve the flexibility they need, which is inconvenient. This change aims to reduce that friction. @kasiazjc feel free to provide more UX context.

superset-frontend/src/components/DropdownContainer/index.tsx
```diff
@@ -179,6 +179,46 @@ const DropdownContainer = forwardRef(
...
+      const container = current?.children.item(0);
+      if (!container) return;
+
+      const childrenArray = Array.from(container.children);
+
+      const resizeObserver = new ResizeObserver(() => {
+        recalculateItemWidths();
+      });
+
+      for (const child of childrenArray) {
```
Antonio-RiveroMartnez, line 192: nit: childrenArray.map(...)

superset-frontend/src/components/DropdownContainer/index.tsx
```diff
@@ -179,6 +179,46 @@ const DropdownContainer = forwardRef(
...
+        resizeObserver.observe(child);
+      }
+
+      // eslint-disable-next-line consistent-return
+      return () => {
+        for (const child of childrenArray) {
```
Antonio-RiveroMartnez, line 198: same nit from above.

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -142,7 +144,24 @@ const HorizontalFormItem = styled(StyledFormItem)`
...
+
+  .ant-form-item-control-input-content {
+    ${({ showExcludeSelection }) => showExcludeSelection && 'display: flex;'}
+  }
+
+  .select-container {
+    ${({ showExcludeSelection, theme }) =>
+      showExcludeSelection &&
+      `
+      width: ${theme.gridUnit * 41}px;
```
Antonio-RiveroMartnez, line 159: I see `theme.gridUnit * 41` multiple times, should this be a const somewhere?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -164,7 +184,12 @@ const useFilterControlDisplay = (
       }
       return {
         FilterControlContainer: HorizontalFilterControlContainer,
-        FormItem: HorizontalFormItem,
+        FormItem: (props: any) => (
```
Antonio-RiveroMartnez, line 187: do we know the type of these props? so we avoid having `any`

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -77,6 +94,33 @@ function reducer(draft: DataMask, action: DataMaskAction) {
   }
 }
 
+const CheckBoxControlWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 8px;
```
Antonio-RiveroMartnez, line 100: Do we have these sizes in our theme?

geido:
@kasiazjc I am testing the functionality and I am agreeing with @michael-s-molinaon the part where this seems a duplicate of inverse selection as currently there are no clear boundaries between the two. Should the "dynamic" option be a sub-option of the "inverse selection"?

Another option would be to make "inverse selection" dynamic and set to true by default for backward compatibility. Curious to hear more feedback.

geido:
Should we leverage the Select option to render custom elements with the Select itself? 

Screenshot 2025-04-11 at 11 36 05

kasiazjc:
> Should we leverage the Select option to render custom elements within the Select itself? The checkbox could be part of the Dropdown in the Select to avoid taking unnecessary space on the filters.
> 
> Screenshot 2025-04-11 at 11 36 05

@geido I think this is a good catch in terms of creating a new values + in terms of inverse selection I think showing a dynamic control could be a solution in here. So we would keep our normal inverse selection filter and then, when the filter is checked, you would see a second checkbox with something like "show dynamic inverse selection in the dashboard" 

@michael-s-molina good point that it's a duplicate, but where it came from we found that users sometimes would prefer using excluding in terms of including values and it's not easily accessible - and only to admins. This doesn't really affect things like security etc, so we thought that it's a good idea to bring it to dashboard filters section. 

In terms of design - I can see how checkbox can take up a lot of space, so here is the second proposal that I had before in the designs 

image

amaannawab923:
> Should we leverage the Select option to render custom elements within the Select itself? The checkbox could be part of the Dropdown in the Select to avoid taking unnecessary space on the filters.
> 
> Screenshot 2025-04-11 at 11 36 05

This one makes much more sense , it doesnt occupy space and doesnt bring major ui changes which might rather break 

As mentioned by @kgabryje  , Inverse selection and this dynamic checkbox are both redundant .... so we rename inverse selection to dynamic inverse selection which makes this checkbox appear of "Exclude filter values" & we do not keep the new config of "Show dynamic exclusion selection field" to avoid redundancy
So the ideal flow can be 

1. Enable Dynamic Inverse selection which makes the checkbox of exclude filter values appear
2. Keep true by default 
3.Toggling it will basically go back & forth to NOT IN and IN operator 

This flow will be straightforward and wont confuse end users ...Awaiting for feedback @geido  @kasiazjc

geido:
> > Should we leverage the Select option to render custom elements within the Select itself? The checkbox could be part of the Dropdown in the Select to avoid taking unnecessary space on the filters.
> 
> This one makes much more sense , it doesnt occupy space and doesnt bring major ui changes which might rather break
> 
> As mentioned by @kgabryje , Inverse selection and this dynamic checkbox are both redundant .... so we rename inverse selection to dynamic inverse selection which makes this checkbox appear of "Exclude filter values" & we do not keep the new config of "Show dynamic exclusion selection field" to avoid redundancy So the ideal flow can be
> 
> * Enable Dynamic Inverse selection which makes the checkbox of exclude filter values appear
> * Keep true by default
> * Toggling it will basically go back & forth to NOT IN and IN operator
> 
> This flow will be straightforward and wont confuse end users ...Awaiting for feedback @geido @kasiazjc

I also feel this is the most straightforward solution that should make everyone happy. I'll be waiting for @michael-s-molina to have a look here and bring his opinion as well on the matter.

amaannawab923:
> > we found that users sometimes would prefer using excluding in terms of including values and it's not easily accessible
> 
> I agree the user need is pretty clear.
> 
> I think @amaannawab923 proposal makes sense:
> 
> > Enable Dynamic Inverse selection which makes the checkbox of exclude filter values appear
> > Keep true by default
> > Toggling it will basically go back & forth to NOT IN and IN operator
> 
> In this context, my observations would be:
> 
> * Do we really need the `Enable Dynamic Inverse selection` or should we always present the checkbox in the filter so users can dynamically change between IN / NOT IN?
> * We also need to think about how to migrate existing filters that currently have `Inverse selection` checked.


Do we really `Enable Dynamic Inverse selection` or should we always present the checkbox .... thats something to discuss about , if it always appears within the native filter , it wont be much of an issue i guess , since its not interrupting the ui at all 

About migrating the existing filter , we can presume that someone has enabled inverse selection because they want to have the NOT IN operator ... in those cases ... We can use the inverse selection field from the form data of existing native filter to keep the Exclude filter values as True... as the existing native filters with inverse selection will work the way they expect 

but point number 1 needs to be discussed , do we really need it configurable or are we ok with it appearing always 

awaiting for feedback @michael-s-molina  @geido

kasiazjc:
> > > we found that users sometimes would prefer using excluding in terms of including values and it's not easily accessible
> > 
> > 
> > I agree the user need is pretty clear.
> > I think @amaannawab923 proposal makes sense:
> > > Enable Dynamic Inverse selection which makes the checkbox of exclude filter values appear
> > > Keep true by default
> > > Toggling it will basically go back & forth to NOT IN and IN operator
> > 
> > 
> > In this context, my observations would be:
> > 
> > * Do we really need the `Enable Dynamic Inverse selection` or should we always present the checkbox in the filter so users can dynamically change between IN / NOT IN?
> > * We also need to think about how to migrate existing filters that currently have `Inverse selection` checked.
> 
> Do we really `Enable Dynamic Inverse selection` or should we always present the checkbox .... thats something to discuss about , if it always appears within the native filter , it wont be much of an issue i guess , since its not interrupting the ui at all
> 
> About migrating the existing filter , we can presume that someone has enabled inverse selection because they want to have the NOT IN operator ... in those cases ... We can use the inverse selection field from the form data of existing native filter to keep the Exclude filter values as True... as the existing native filters with inverse selection will work the way they expect
> 
> but point number 1 needs to be discussed , do we really need it configurable or are we ok with it appearing always
> 
> awaiting for feedback @michael-s-molina @geido

I think maybe we could show this (check) by default, but user can still disable it in the settings if they want the simple filter without the possibility of choosing between? If they want the simpler solution @michael-s-molina

michael-s-molina:
@kasiazjc I like your solution. The only change I would do is to remove the `show dynamic inverse selection option in the dashboard` and always enable that feature. It seems nice to always have a quick way of changing between IN and NOT IN in the filters and I want to avoid more complexity that needed. We just need to make sure that if `Inverse selection` is selected, the NOT IN is displayed by default.

You all probably already know this but I just want to say anyway. The difference between the `Inverse selection` checkbox in the edit modal and the quick action to alternate between in/not in is that the former is saved and shared between all dashboard users and the latter is just a transient thing that is not saved.

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -118,9 +120,23 @@ const VerticalFormItem = styled(StyledFormItem)`
       }
     }
   }
+
+  .ant-form-item-control-input-content {
+    ${({ inverseSelection }) => inverseSelection && 'display: flex;'}
```
geido, line 125: Can we avoid adding custom styles here and use a Flex component inside the SelectFilterPlugin itself?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -142,7 +158,24 @@ const HorizontalFormItem = styled(StyledFormItem)`
   }
 
   .ant-form-item-control {
-    width: ${({ theme }) => theme.gridUnit * 41}px;
+    width: ${({ theme, inverseSelection }) =>
+      inverseSelection ? theme.gridUnit * 41 * 1.6 : theme.gridUnit * 41}px;
+  }
+
+  .ant-form-item-control-input-content {
+    ${({ inverseSelection }) => inverseSelection && 'display: flex;'}
```
geido, line 166: Same as my comment above, can we use the Ant Design built-in components such as `Space` and `Flex`?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -142,7 +158,24 @@ const HorizontalFormItem = styled(StyledFormItem)`
   }
 
   .ant-form-item-control {
-    width: ${({ theme }) => theme.gridUnit * 41}px;
+    width: ${({ theme, inverseSelection }) =>
+      inverseSelection ? theme.gridUnit * 41 * 1.6 : theme.gridUnit * 41}px;
```
geido, line 162: width should not use `gridUnit` which should be only meant for spacing. Can we use a fixed value here? Also, do we really need to set a specific width?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
```diff
@@ -142,7 +158,24 @@ const HorizontalFormItem = styled(StyledFormItem)`
...
+
+  .ant-form-item-control-input-content {
+    ${({ inverseSelection }) => inverseSelection && 'display: flex;'}
+  }
+
+  .select-container {
+    ${({ inverseSelection, theme }) =>
+      inverseSelection &&
+      `
+      width: ${theme.gridUnit * 41}px;
```
geido, line 173: No `gridUnit` for widths, please. I know this was like this before but it is likely a mistake.

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -288,18 +309,54 @@ export default function PluginFilterSelect(props: PluginFilterSelectProps) {
...
       <StyledFormItem
         validateStatus={filterState.validateStatus}
         extra={formItemExtra}
       >
+        {inverseSelection && (
+          <SelectWrapper>
+            <Select
+              value={`${excludeFilterValues}`}
+              options={[
+                { value: 'true', label: 'is not' },
```
geido, line 347: Should labels be localized?

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
amaannawab923: That is likely difficult due to the structure of the Native filter

The FilterControl.tsx has the label  and the SelectFilterPlugin has the Select dropdown 
    <div>
            <FormItem
              label={label}
              htmlFor={filter.id}
              required={filter?.controlValues?.enableEmptyFilter}
              validateStatus={validateStatus}
            >
              <OutPortal node={portalNode} />
            </FormItem>
          </div>
          
          The filter control is rendering these Selects through portals .... I would have placed everything here in FilterControl only
          
          <div>
            <FormItem
              label={label}
              htmlFor={filter.id}
              required={filter?.controlValues?.enableEmptyFilter}
              validateStatus={validateStatus}
            >
              <OutPortal node={portalNode} /> <-
              <ExcludeFilterValues /> 
            </FormItem>
          </div>
          But cant since the SelectFilterPlugin is getting rendered through these portals and i dont have access to these states
          
          and i cant access the parent container css from the SelectFilterPlugin .... Because it will be difficult to differentiate
          
          and also i had to add here because we need less space in case of VerticalFormItem and HorizontalOverFlowitem & we have a bit more space in case of horizontal list ... and i can only have the context from here 
          if we have Horizontal , HorizontalOverflow and VerticalForm item 
  Trying to do it from SelectFilterPlugin might require a lot of hacks and unclean code

amaannawab923: I managed to remove most of the custom styles and used Space inside of SelectFilterPlugin @geido

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterControl.tsx
amaannawab923: This is removed now ... Used Space from antd in SelectFilter @geido

michael-s-molina (approved):
Removing the Request Changes status as I'll be out for the next days.

Thanks for the work @amaannawab923! @geido @kasiazjc I'll leave the final review and tests with you. Thanks for the discussions!

geido (approved):

PR #33070, opened 2025-04-09
chore: use create table util

eschutho:
### SUMMARY

Using the create_table util from https://github.com/apache/superset/pull/31303
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2015-09-21_17-30_4e6a06bad7a8_init.py
```diff
@@ -22,17 +22,19 @@
 
 """
 
+import sqlalchemy as sa
+from alembic import op
```
sadpandajoe, line 26: You can probably get rid of this import by using `drop_table` as part of the utils: https://github.com/apache/superset/blob/5537c6fcd12ad6d4e932832245441a69178bd39b/superset/migrations/shared/utils.py#L250

superset/migrations/versions/2016-07-25_17-48_ad82a75afd82_add_query_model.py
```diff
@@ -22,16 +22,18 @@
 
 """
 
+import sqlalchemy as sa
+from alembic import op
```
sadpandajoe, line 26: I see we have an add columns, which we should move over to `add_columns`: https://github.com/apache/superset/blob/5537c6fcd12ad6d4e932832245441a69178bd39b/superset/migrations/shared/utils.py#L305C5-L305C16

There is also a `create_index`: https://github.com/apache/superset/blob/5537c6fcd12ad6d4e932832245441a69178bd39b/superset/migrations/shared/utils.py#L365

And a `drop_columns`: https://github.com/apache/superset/blob/5537c6fcd12ad6d4e932832245441a69178bd39b/superset/migrations/shared/utils.py#L335

sadpandajoe (commented):
If we're only concerned about the table creations then I think this move looks good, but if we're concerned about columns and indexes too, then we should use those util functions

eschutho:
> If we're only concerned about the table creations then I think this move looks good, but if we're concerned about columns and indexes too, then we should use those util functions

I was planning on doing the others in a separate pr

superset/migrations/migration_utils.py
```diff
@@ -15,21 +15,32 @@
...
+    # won't abort the outer transaction.
+    try:
+        with op.get_context().autocommit_block():
+            with op.batch_alter_table(
+                table_name, naming_convention=naming_convention
+            ) as batch_op:
+                batch_op.create_unique_constraint(index_id, uix_columns)
+    except Exception as e:
+        # Likely the constraint already exists, or another DDL error occurred.
+        logger.warning("Failed to create unique constraint %s: %s", index_id, e)
```
eschutho, line 43: this was failing during testing due to a constraint that already existing when downgrading.

eschutho:
github isn't pushing up my latest changes, so I'm going to reopen this as a new PR.

PR #33072, opened 2025-04-10
chore: use create table util 

eschutho:
### SUMMARY

Using the create_table util from https://github.com/apache/superset/pull/31303
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/app.py
```diff
@@ -17,8 +17,7 @@
 
 import logging
 import os
-from typing import cast, Iterable, Optional
-from wsgiref.types import StartResponse, WSGIApplication, WSGIEnvironment
```
eschutho, line 21: wsgiref.types does not exist on Python 3.10

superset/migrations/migration_utils.py
```diff
@@ -15,21 +15,32 @@
...
+    # won't abort the outer transaction.
+    try:
+        with op.get_context().autocommit_block():
+            with op.batch_alter_table(
+                table_name, naming_convention=naming_convention
+            ) as batch_op:
+                batch_op.create_unique_constraint(index_id, uix_columns)
+    except Exception as e:
+        # Likely the constraint already exists, or another DDL error occurred.
+        logger.warning("Failed to create unique constraint %s: %s", index_id, e)
```
eschutho, line 43: this was failing during testing due to a constraint that already existing when downgrading.

eschutho:
I pushed up another change. The last commit should be 3162bb3f594fe7b3f018cc86806cb01998b2d79d, but github is being slow.

superset/app.py
eschutho: fixing from https://github.com/apache/superset/commit/09b92e7d08bc96b6006938c00d9f8ce7a108102b

superset/migrations/migration_utils.py
michael-s-molina: @eschutho Can we check if the constraint already exists like we do for the other functions?

```
def create_unique_constraint(
    op: Operations, index_id: str, table_name: str, uix_columns: list[str]
) -> None:
    # Get the database connection and inspector
    conn = op.get_bind()
    inspector = inspect(conn)

    # Check if the unique constraint already exists
    existing_constraints = inspector.get_unique_constraints(table_name)
    for constraint in existing_constraints:
        if constraint["name"] == index_id:
            # Constraint already exists, no need to create it
            return

    # Create the unique constraint if it doesn't exist
    with op.batch_alter_table(
        table_name, naming_convention=naming_convention
    ) as batch_op:
        batch_op.create_unique_constraint(index_id, uix_columns)
```

mistercrunch (approved):
LGTM - seeing some migrations from 2015 .... would be great to eventually collapse migrations that are more then 2-3 years old. Doubt anyone is upgrading from that old of versions, and even if/when they are, they could follow multi-step upgrades (first jump from pre 1.0 to say 3.0 using the 3.0 version, and then proceed from there)

PR #33080, opened 2025-04-10
fix: block MySQL system variables (@@) in SQL queries

sha174n:
### SUMMARY

This PR adds protection against MySQL system variables (@@) in SQL queries, which can be used to extract sensitive system information.

### TESTING INSTRUCTIONS

- Set up a local instance of Superset with the modified code
- In SQL Lab, attempt to execute a query containing a MySQL system variable like SELECT @@hostname
- Verify that the query is blocked and an error message appears stating that MySQL system variables are not allowed

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/config.py
+++ b/superset/config.py
@@ -1393,6 +1393,11 @@ DISALLOWED_SQL_FUNCTIONS: dict[str, set[str]] = {
 def SQL_QUERY_MUTATOR(  # pylint: disable=invalid-name,unused-argument  # noqa: N802
     sql: str, **kwargs: Any
 ) -> str:
+    import re
+
+    # Check if the query contains any @@ variables
+    if re.search(r"@@\w+", sql):
+        raise Exception("MySQL system variables (@@) are blocked for security reasons")
     return sql
 
 


rusackas:
cc @mistercrunch @dpgaspar @betodealmeida

superset/config.py
```diff
@@ -1393,6 +1393,11 @@ def engine_context_manager(  # pylint: disable=unused-argument
 def SQL_QUERY_MUTATOR(  # pylint: disable=invalid-name,unused-argument  # noqa: N802
     sql: str, **kwargs: Any
 ) -> str:
+    import re
+
+    # Check if the query contains any @@ variables
+    if re.search(r"@@\w+", sql):
```
michael-s-molina, line 1399: I'll wait for @dpgaspar and @betodealmeida inputs but this function is a global function which affects queries to all database types. If this is a MySQL restriction, it probably needs to be inside the `MySqlEngineSpec`.

mistercrunch: It's not a good idea to implement this here, this is a no-op configuration that's meant as a hook to be used for customization in your own local environment. Anyone who has defined `SQL_QUERY_MUTATOR` for whatever purposes would bypass this.

But you can totally do this in your own local `superset_config.py` as a thing that is specific to your local configuration.

Now the best way to handle access to this is for the database administrator to GRANT or REVOKE the privileges related to this particular feature of MySQL to the service account used to access Superset. More on the topic here https://chatgpt.com/share/67f82669-4918-8010-897f-b26e5cba32a1

mistercrunch (changes requested):
Copying what I wrote inline in the PR -> The best way to handle access to this is for the database administrator to GRANT or REVOKE the privileges related to this particular feature of MySQL to the service account used to access Superset. More on the topic here https://chatgpt.com/share/67f82669-4918-8010-897f-b26e5cba32a1

Setting as "request changes", but really I think it's a bit of a dead end. Solution is to REVOKE from the database user roles.

mistercrunch:
Closing given there's a better, more appropriate solution. 

Why this doesn't work as is:
- @@ variables are a feature, a DBA with a proper account might use SQL Lab to query `SELECT @@version`, this is can be perceived as a feature for many users. Changing the behavior could be seen as a regression
- implemented as a default config hook. Anything that's database-engine-specific belongs in `superset/db_engine_specs/`
- if Superset wanted to expose such functionality, it should be configurable, probably through RBAC/perms, though I think the service account approach is preferable

PR #33081, opened 2025-04-10
refactor(Components): Use named imports - P.1

geido:
### SUMMARY
This PR aims to bring consistency in the usage of named imports vs default. After gathering feedback from multiple parties, we believe named imports is the way to go for the following reasons:

- It improves developer experience by making sure the intended names are used.
- It will simplify the process of moving components to a dedicated package in the future via a barrel file.

This is P.1 going through half of the components. I will open follow-up PRs to avoid that this becomes too large.

## What's included
- Usage of named imports for all `src/components/*`.
- Better split of types from presentational layer.
- Export of commonly used types that can be imported via `src/components `OR directly from the dedicated types file of each component (most people prefer via `src/components` for simplicity of access).

## What's not included
- Some components such as `src/components/Chart` seem like they do not belong to atomic / presentational components and won't be touched by this PR.
- The `src/components/Icons `is a special component and we are discussing whether it would make sense to host it in a dedicated package (similar to what Ant Design does with their icons). For that reason, the `Icons` component has not been touched in this PR.

## Some opinionated decisions
- The barrel file only exports specific types and not all of them to make sure the codebase is not polluted by types that are not intended to be used outside of the component.
- The barrel file exports specific components / functions from the component for the same reason.
- Imports of types have been changed across the codebase to go through the barrel file with the `type` specification. Most people have opted for this solution. There is a slight risk of name collisions but it should not be a common case as long as the types are named correctly.

N.A.

### TESTING INSTRUCTIONS
CI should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/EmptyState/index.tsx
```diff
@@ -36,7 +35,9 @@ import EmptySqlChartImage from 'src/assets/images/empty_sql_chart.svg';
 import EmptyQueryImage from 'src/assets/images/empty-query.svg';
 import EmptyTableImage from 'src/assets/images/empty-table.svg';
 import EmptyImage from 'src/assets/images/empty.svg';
+import { Button } from '../Button';
```
kgabryje, line 38: There are multiple places where we use relative paths instead of `src/components`. Is that intentional or are those auto generated by your IDE?

kgabryje (approved):
Great work, it feels good to see the imports/exports consolidated like this! Left 1 non-blocking comment, can be addressed in a follow-up

superset-frontend/src/components/EmptyState/index.tsx
geido: There have been some issues with circular dependencies caused by pulling components from the barrel file within components inside components and I figured we should just use relative imports. We might revisit this when we move to a dedicated package and split things a bit better. This has grown too large already.

michael-s-molina:
Thank you!

PR #33084, opened 2025-04-10
fix: enhance disallowed SQL functions list for improved security

sha174n:
### SUMMARY

This PR expands the list of disallowed SQL functions in the DISALLOWED_SQL_FUNCTIONS configuration to improve security when using PostgreSQL, MySQL, SQLite, MS SQL Server and Clickhouse. It adds several functions that could potentially expose sensitive system information when used in SQL queries.

### TESTING INSTRUCTIONS

- Set up a local instance of Superset with the modified code
- Manual testing in SQL Lab by attempting to execute queries using the newly disallowed functions
- Verify that the query is blocked and an error message appears

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Glancing at the list (and the failing tests) I worry this might be *too* restrictive. CC @mistercrunch @dpgaspar

michael-s-molina:
I think we need to be really careful about the potential of breaking things in this change. +1 to @rusackas's comment.

dpgaspar:
> Glancing at the list (and the failing tests) I worry this might be _too_ restrictive. CC @mistercrunch @dpgaspar

don't think failing tests are related to this: https://github.com/apache/superset/pull/30134/files#r2038270948

mistercrunch:
Guessing `master` was broken at the time of branch creation, please rebase.

About the feature, I'm not sure whether/how we should be in the business of disabling SQL functionality for security purposes. I think where I'd draw the line is in areas where a SQL feature can compromise the integrity of the OS and Superset instance itself. I don't see any issue with `current_schema` for instance, it's a useful feature AFAIC, not a security concern - unless I'm missing somehting.

Also noting that much of these database features can be disabled at the database user account level. Seems good DBA hygiene is key here, not BI tool stewardship.

Now. Here we're modifying a local config setting, so it's designed to be managed/altered at the environment level. You can do this on your envrionment today. It can certainly be good to prevent certain things that we know are bad. "secure by default" is great, though I wouldn't want to provide the impression that it's Superset's responsibility to secure your database or give the impression of a false sense of safety from it. It's not scalable for us (and I imagine for other BI tools) to be on top of all the potential security issues different database engines may expose.

In any case, whatever we decide here, those are breaking changes and would need an entry in UPDATING.md

michael-s-molina:
> In any case, whatever we decide here, those are breaking changes and would need an entry in UPDATING.md

If there are breaking changes, they need to go into 6.0.

dpgaspar:
> > In any case, whatever we decide here, those are breaking changes and would need an entry in UPDATING.md
> 
> If there are breaking changes, they need to go into 6.0.

The problem here is that, many of these function simply can't be turned off at the database level, regarding Superset's responsibility I agree, but let's talk

sha174n:
Right now we're showing the following message:
"SQL statement contains disallowed function(s): {'inet_client_addr', 'query_to_xml', 'table_to_xml', 'table_to_xml_and_xmlschema', 'version', 'inet_server_addr', 'query_to_xml_and_xmlschema', 'database_to_xml'}".

I agree a more descriptive message would be a good idea, definitely.

mistercrunch:
Definitely should only show the term used that is not allowed. Otherwise LGTM apart from the failing unit tests, maybe rebase would help here.

michael-s-molina (changes requested):
- Can we make the PR title more accurate? It's not only MySQL that is being affected.
- Can we add a note in `UPDATING.md` about the breaking changes?
- This can only be merged during a breaking window. I added this PR to the [6.0 project board](https://github.com/orgs/apache/projects/462).

rusackas:
@sha174n I think we're near an appropriate time to merge this in. Are you able to add the note to UPDATING.md, or do you need help with that?

betodealmeida (approved):

mistercrunch:
@michael-s-molina looks like progress has been made since you requested changes. Mind taking another look?

michael-s-molina (approved):
@mistercrunch I'll approve the PR to unblock but can we please add something to `UPDATING.md` before merging given that this PR contain breaking changes? I'm assuming that this is part of the current 6.0 breaking window.

mistercrunch:
+1 let's add a note to `UPDATING.md`

dpgaspar (approved):

mistercrunch:
@sha174n can you add a line in UPDATING.md

```
* [33084](https://github.com/apache/superset/pull/33084) The DISALLOWED_SQL_FUNCTIONS configuration now includes additional potentially sensitive database functions across PostgreSQL, MySQL, SQLite, MS SQL Server, and ClickHouse. Existing queries using these functions may now be blocked. Review your SQL Lab queries and dashboards if you encounter "disallowed function" errors after upgrading
```

PR #33092, opened 2025-04-10
fix: CI file change detector to handle large PRs

mistercrunch:
Improve our change detector. The way it works currently, we do a simple call to the github API to fetch the list of files that have changed in the PR, so that we can see if we should trigger cetain jobs. Now it appears that the particular endpoint we use returns a max of 30 results by call, unless we specify the `?per_page=100` argument.

Now I'm a bit lazy and didn't want to implement paging, so instead I'm resorting to fetching 100, and if/when we see 100, we assume that we should trigger all checks.

Why not paging?
- i'm lazy, and we should probably use a proper github client instead of re-implementing paging, and that'd require a bit more time/effort
- triggering everything if we have 100 files is fine, maybe super occasionally we'll trigger a bit more work than necessary (large frontend-only refactor, large backend-only refactor, ...) but I think that's fine

### DIFF
--- a/scripts/change_detector.py
+++ b/scripts/change_detector.py
@@ -63,7 +63,10 @@ def fetch_files_github_api(url: str):  # type: ignore
 
 def fetch_changed_files_pr(repo: str, pr_number: str) -> List[str]:
     """Fetches files changed in a PR using the GitHub API."""
-    url = f"https://api.github.com/repos/{repo}/pulls/{pr_number}/files"
+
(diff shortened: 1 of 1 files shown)


eschutho (approved):

mistercrunch:
Blocked by errors in master, might try to tackle if no one else is on it already...

sadpandajoe (approved):

PR #33096, opened 2025-04-11
feat(Native Filters): Configure creatable filter behavior

geido:
### SUMMARY
This PR introduces a new option for Native Filters to enable / disable the creatable behavior of values filters.

### AFTER

https://github.com/user-attachments/assets/3d05a8c7-5595-48aa-9a0a-7e8cc0cb93fb

### TESTING INSTRUCTIONS
- Open a Dashboard and set a value filter
- The creatable option should be true by default and it should be possible to create new values
- Disabling the option should disallow the creation of new values

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/dashboard/shared_dashboard_functions.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/dashboard/shared_dashboard_functions.ts
@@ -55,6 +55,7 @@ export function prepareDashboardFilters(
         controlValues: {
           enableEmptyFilter: false,
           defaultToFirstItem: false,
+          creatable: true,
           multiSelect: true,
           searchAllOptions: false,
           inverseSelection: false,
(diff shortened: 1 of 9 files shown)


kgabryje (approved):
LGTM

mistercrunch (approved):

michael-s-molina:
@geido @kgabryje What's the difference between this option and "Dynamically search all filter values"?

eschutho:
> @geido @kgabryje What's the difference between this option and "Dynamically search all filter values"?

@michael-s-molina this feature would disallow adding a filter for a value that does not exist in the results. As opposed to searching for a field that isn't in the first 1k results. So for example, if you have values "yes" and "no", it wouldn't allow someone to search for "maybe".

PR #33099, opened 2025-04-11
feat: add metric name for big number chart types #33013

### LINKED ISSUE #32985: Feat: Show Metric for Big Number chart types
## Description

For Big Number, Big Number with Trendline and Big Number with Time Comparison (enable Feature Flag CHART_PLUGINS_EXPERIMENTAL to see this), we want to have an option to show the metric name.

## Design input

- In Big Number and Big Number with Trendline, the metric name should be placed above the big number.
- Spacings should follow the patterns of the Big Number elements (Big Number and subtitle).
- The base size of the metric name should be 24px and should change dynamically as it scales (as in the default Big Number).

Image

- For Big Number with Time Comparison, the metric name should be above the Big Number, and the spacing between the two should be as shown in the pricture..
- The base size of the metric name should be 24px and should change dynamically as it scales (as in the default Big Number).

Image

- In the Customize section add an option to show / hide the Metric name and set its font-size, with options similar to Subheader Font Size as available in Big Number.

→ if unchecked metric font size is not shown 
→ if checked metric font size is shown

Image

Image

## Notes

Make sure that friendly metric names are respected as in https://github.com/apache/superset/pull/32665

fardin-developer:
Fixes: #32985

## SUMMARY
This PR introduces support for displaying the **Metric name** in **Big Number** chart types, including Big Number and Big Number with Time Comparison. The metric name appears above the main number, styled consistently with existing elements, and includes customization options in the control panel.

### CHANGES
- Added metric name display to:
  - Big Number
  - Big Number with Time Comparison
- Positioning and spacing follow existing patterns of Big Number and subtitle.
- Metric name font size defaults to 24px and scales responsively with the chart.
- Added toggle in Customize section to show/hide metric name.
- Added input to control metric name font size (similar to Subheader Font Size).

### Before

Screenshot 2025-04-07 at 1 28 27 AM
Screenshot 2025-04-07 at 1 32 45 AM

##
### After
Screenshot 2025-04-07 at 12 34 36 AM
Screenshot 2025-04-07 at 12 35 58 AM

https://github.com/user-attachments/assets/a95ef5a7-28c7-4985-81e5-ada59c931b34

### ADDITIONAL NOTES
- Ensured the feature behaves correctly across chart types and screen sizes.
- Maintained dynamic scaling and visual hierarchy.
- Verified with RTL support and Storybook examples.

### CHECKLIST
- [x] Displayed metric name above the number for all Big Number chart types.
- [x] Used responsive font scaling with base size 24px.
- [x] Added toggle and font size input to Customize controls.
- [x] Verified RTL compatibility.


michael-s-molina:
@fardin-developer Is it possible to hide the metric name? There are cases where they are really extensive or the chart's title is pretty clear.

@rusackas also asked how does it handle really long metric names.

fardin-developer:
> @fardin-developer Is it possible to hide the metric name? There are cases where they are really extensive or the chart's title is pretty clear.
> 
> @rusackas also asked how does it handle really long metric names.

Thanks for your feedback. I have added a checkbox to show metric name, by default it will be hidden.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
```diff
@@ -26,7 +26,12 @@ import {
   SimpleAdhocFilter,
   ensureIsArray,
 } from '@superset-ui/core';
-import { getComparisonFontSize, getHeaderFontSize } from './utils';
+import {
+  getComparisonFontSize,
+  getHeaderFontSize,
+  getMetricNameFontSize,
+} from './utils';
+import { getFriendlyMetricName } from '../utils';
```
Antonio-RiveroMartnez, line 34: The `friendlyName` is not called like that in the table chart, could we be consistent? https://github.com/apache/superset/pull/32665 also, not sure but, is this something we can extract and reuse both in table and BN?

fardin-developer: Thanks @Antonio-RiveroMartnez . I’ve removed that way of getting the friendlyName and simplified it directly within transformProps — we no longer need that function. Also, yes, I agree it makes sense to extract and reuse it in both Table Chart and Big Number. 

I’m planning to create a separate issue for that and will try to merge both implementations accordingly. Do you think that approach would work?

Antonio-RiveroMartnez: Awesome! Thanks @fardin-developer , sounds like a plan.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
```diff
@@ -91,11 +96,16 @@ export default function transformProps(chartProps: ChartProps) {
     percentDifferenceFormat,
     subtitle = '',
     subtitleFontSize,
-    columnConfig,
+    columnConfig = {},
   } = formData;
   const { data: dataA = [] } = queriesData[0];
   const data = dataA;
   const metricName = metric ? getMetricLabel(metric) : '';
+  const metricKey = metric ? getMetricLabel(metric) : '';
```
Antonio-RiveroMartnez, line 104: Same with this block, could this be some sort of util we can reuse in the 3 chart types?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/controlPanel.ts
```diff
@@ -70,6 +71,28 @@ const config: ControlPanelConfig = {
         ],
         [subtitleControl],
         [subtitleFontSize],
+        [
```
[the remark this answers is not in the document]
fardin-developer: resolved it.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/transformProps.ts
fardin-developer: Thanks for the update! I’ve moved the originalLabel logic into a separate getOriginalLabel function and updated it to address issues where the metric key was incorrectly displayed in some cases. The new implementation now works as expected.

geido:
Should this be this "huge"? Also, when it is this big, it gets cut out on the top. I have tried scrolling to the top but it is still cut.

Screenshot 2025-04-17 at 18 20 39

Screenshot 2025-04-17 at 18 22 50

fardin-developer:
> Should this be this "huge"? Also, when it is this big, it gets cut out on the top. I have tried scrolling to the top but it is still cut.
> 
> Screenshot 2025-04-17 at 18 20 39
> 
> Screenshot 2025-04-17 at 18 22 50
Let me check this issue. 
I have not got this issue on my side. I also tried inspecting it on chrome too, still.

fardin-developer:
https://github.com/user-attachments/assets/3dc545a0-b4a4-4512-9db7-f5029eb3302a

Hey @geido,
I've fixed the last issue. Now we have two options: we can either keep the current font size or reduce it. If we choose to reduce the font size, we'll also need to update the test case file accordingly. Am I allowed to make changes to the test case file? If yes, I’ll go ahead and update it. Otherwise, we can stick with the existing font size and enable a scrollable option instead.

fardin-developer:
Reopening PR — It was closed accidentally during a force-push.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/sharedControls.ts
```diff
@@ -21,106 +21,68 @@
...
+      clearable: false,
+      default: defaultValue,
+      options,
+    },
+  };
+}
+
+export const headerFontSize = makeFontSizeControl(
+  'header_font_size',
+  'Big Number Font Size',
```
geido, line 61: I think all of these labels need to be localized for transation?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/utils.ts
```diff
@@ -40,20 +49,28 @@ const getFontSizeMapping = (
...
 
 const comparisonFontSizesMapping = getFontSizeMapping(
   subheaderProportionValues,
-  comparisonFontSizes,
+  sharedFontSizes,
 );
 
+export const getMetricNameFontSize = (proportionValue: number) =>
+  metricNameFontSizesMapping[proportionValue] ??
+  sharedFontSizes[sharedFontSizes.length - 1];
```
geido, line 68: Is this logic correct here of always defaulting to the bigger font size?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -92,6 +95,38 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
...
+    const fontSize = computeMaxFontSize({
+      text,
+      maxWidth: width,
+      maxHeight,
+      className: 'metric-name',
+      container,
+    });
+    container.remove();
+
+    if (!showMetricName || !metricName) return null;
```
geido, line 115: Isn't this the same as the check above?

geido, line 115: Isn't this the same as the check above?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/sharedControls.ts
```diff
@@ -21,106 +21,68 @@
...
+  0.3,
```
geido, line 62: The default originally was` 0.4`. Is this change wanted?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberPeriodOverPeriod/utils.ts
fardin-developer: Yes, the logic is correct as implemented. It only defaults to the larger font size when proportionValue isn’t found in the mapping—which should be rare with controlled inputs. This fallback approach is consistent with the previous logic used in getHeaderFontSize, maintaining the same design principle of reliability.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
fardin-developer: Thank you, by mistake it was added for twice, I am remvoing the second one.

LevisNgigi:
I’ve rebased PR [#33099](https://github.com/apache/superset/pull/33099) on top of my [branch](https://github.com/apache/superset/pull/33196) did some manual testing, and everything looks good so far.There are a few merge conflicts that will need resolving, but I haven’t found any backward compatibility issues—form data, class names, and control panels remain intact with no breaking changes.

fardin-developer:
Reopening PR

geido:
A few last points:

- Can we have the timestamp on top of the metric name with Big Number with Trendline?
- I still see the behavior as shown in the video


https://github.com/user-attachments/assets/db6fd18b-d4ab-43fe-8dd2-6fd126c0cfd1

Antonio-RiveroMartnez (approved):
Tested on the ephemeral and LGTM! Thanks for taking care of this

https://github.com/user-attachments/assets/4973372e-d011-4680-9ee4-c40837a09770

fardin-developer:
> Tested on the ephemeral and LGTM! Thanks for taking care of this
> 
>  Screen.Recording.2025-05-07.at.4.43.02.PM.mov

Thank you @Antonio-RiveroMartnez .

PR #33100, opened 2025-04-11
fix(OAuth2): Update connection should not fail if connection is missing OAuth2 token

Vitor-Avila:
### SUMMARY
The `UpdateDatabaseCommand` command calls the `SyncPermissionsCommand` to reflect new datasources into Roles (for data access). The update itself might purge all OAuth2 tokens, or the user performing the change might not have a valid token yet, so the update should not fail in case the token is missing.

No UI changes.

### TESTING INSTRUCTIONS
Test coverage added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/sync_permissions.py
+++ b/superset/commands/database/sync_permissions.py
@@ -28,6 +28,7 @@ from superset.commands.database.exceptions import (
     DatabaseConnectionFailedError,
     DatabaseConnectionSyncPermissionsError,
     DatabaseNotFoundError,
+    MissingOAuth2TokenError,
     UserNotFoundInSessionError,
 )
 from superset.commands.database.utils import (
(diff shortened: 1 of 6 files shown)


superset/commands/database/sync_permissions.py
```diff
@@ -115,6 +116,11 @@ def validate(self) -> None:
             try:
                 alive = ping(engine)
             except Exception as err:
+                if (
+                    self.db_connection.is_oauth2_enabled()
+                    and self.db_connection.db_engine_spec.needs_oauth2(err)
+                ):
```
betodealmeida, line 122: In the near future we should DRY this code, we use it a lot.

betodealmeida (approved):

superset/commands/database/sync_permissions.py
Vitor-Avila: That's a good callout. I'll merge this one for now but I'll keep this in mind.

PR #32115, opened 2025-02-03
fix(Scope): Correct issue where filters appear out of scope when sort is unchecked.

LevisNgigi:
### SUMMARY
This fixes a bug where updating a native filter will mark filters as out of scope even though they aren't.

Fixes #32014 and #32123

Fix(Scope): Correct issue where filters appear out of scope when sort is unchecked

### TESTING INSTRUCTIONS
1.Go to the video games dashboard
2.Edit the Genre filter by saying sort value filters by count
3.Save the filter
4.Edit the filter again and uncheck filter values and save
5.When we weren't sorted all filters were in scope, after we sorted filters are still in scope, now when you uncheck, sort filters should still be in scope.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #32014 
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #33106, opened 2025-04-11
fix(dashboard): invalid active tab state

justinpark:
### SUMMARY
The bug where filters outside the current scope are displayed was newly introduced during the process of identifying filters applied to active tabs in [#32115](https://github.com/apache/superset/pull/32115/files#diff-1bbe96474254d89b21e3b373e1c9e8a222ee892518595a69a809c5f035e5bd4eR126). The root cause occurs from the issue that when tabs are structured with multiple levels and the scope is applied to a lower tab, changing the top-level tab leads to an issue where tabs that are no longer visible are not removed from the active tabs state.
For example, in the following tab structure, when the top-level tab is changed, the previous active tabs retain the sub-tabs of the previous tab.

```
        ROOT
    |         |
   TAB-A     TAB-B
    |         |
   TAB-SUB-A     TAB-SUB-B

active_tabs: ['ROOT', 'TAB-A', 'TAB-SUB-A']
// setActiveTab from TAB-A to TAB-B
active_tabs: ['ROOT', 'TAB-B', 'TAB-SUB-A', 'TAB-SUB-B']
```

To address this issue, this commit added a logic to remove the inactive tabs from the active tabs when the top-level tab is switched. Furthermore, to restore the state of previously selected sub-tabs when the tabs return to their original state, this commit modified the logic to store them in the inactive tabs state and recover the sub-tabs associated with the changed top-level tab.

Before:

https://github.com/user-attachments/assets/6348574e-3177-4fae-b7ad-39780dd8ae72

After:

https://github.com/user-attachments/assets/a19bc36e-6381-4c01-8152-26b859f47b83

### TESTING INSTRUCTIONS
Follow the above screencapture steps

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/actions/dashboardState.js
+++ b/superset-frontend/src/dashboard/actions/dashboardState.js
@@ -665,18 +665,21 @@ export function setActiveTab(tabId, prevTabId) {
     const { present: currentLayout } = dashboardLayout;
     const restoredTabs = [];
     const queue = [tabId];
+    const visited = new Set();
     while (queue.length > 0) {
       const seek = queue.shift();
-      const found =
-        prevInactiveTabs?.filter(inactiveTabId =>
-          currentLayout[inactiveTabId]?.parents
-            .filter(id => id.startsWith('TAB-'))
-            .slice(-1)
-            .includes(seek),
-        ) ?? [];
-
-      restoredTabs.push(...found);
-      queue.push(...found);
+      if (!visited.has(seek)) {
+        visited.add(seek);
+        const found =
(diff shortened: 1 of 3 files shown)


justinpark:
cc: @LevisNgigi

LevisNgigi:
> cc: @LevisNgigi

@justinpark This is well noted. Thanks for fixing this.

kgabryje (approved):
lgtm!

PR #33111, opened 2025-04-12
fix: Remove non-functional subheader_font_size control in Big Number with Trendline

### LINKED ISSUE #32776: Subheader Size control on Big Number with Trendline does nothing.
## Screenshot

Image

## Description

The above highlighted control does nothing. We should either remove it (easy!) or perhaps add a Subheader input like the regular Big Number chart so that it DOES provide value (probably harder, but not rocket surgery).

## Design input
[describe any input/collaboration you'd like from designers, and
tag accordingly. For design review, add the
label `design:review`. If this includes a design proposal,
include the label `design:suggest`]

UncoveredBread:
Fixes #32776

Removed the non-functional `subheader_font_size` control from Big Number with Trendline in `plugin-chart-echarts/src/BigNumber/sharedControls.ts`, as it had no effect on the subheader. Updated related tests in `superset-ui-core/test/time-comparison/` by removing `subheader_font_size` references.

Verification
- Manually tested: Confirmed the "Subheader Font Size" dropdown is gone from the "Customize" tab of a Big Number with Trendline chart, and the chart renders correctly (e.g., displays the number and trendline).
- Note: I couldn’t run `npm run test` due to an `ENOENT: no such file or directory` error when running from the root directory. I believe the tests should pass after removing `subheader_font_size` references, but I’d appreciate assistance in running them correctly or confirmation via CI.


LevisNgigi:
Hey,subheader font size is functional and works well with the subheader that appears below the big number.It appears when you set value in "Comparison Period Lag" and also if you add a comparison suffix as in the image below.
subheader2

mistercrunch:
Adding a `hold!` label since the premise for the issue/fix are flawed.

superset-frontend/src/features/databases/UploadDataModel/index.tsx
```diff
@@ -185,10 +185,15 @@ export const validateUploadFileExtension = (
   }
 
   const fileType = extensionMatch[1].toLowerCase();
+<<<<<<< HEAD
```
rusackas, line 188: We probably shouldn't see this conflict here.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/sharedControls.ts
```diff
@@ -55,36 +55,4 @@ export const headerFontSize: CustomControlItem = {
   },
 };
 
-export const subheaderFontSize: CustomControlItem = {
```
rusackas, line 58: Ahh... removing it from SharedControls makes sense, but doesn't this remove it from the regular BigNumber, too? That one _does_ have the Subheader, and _should_ have the control.

superset-frontend/src/dashboard/actions/dashboardState.js
```diff
@@ -98,23 +98,25 @@ export function toggleFaveStar(isStarred) {
   return { type: TOGGLE_FAVE_STAR, isStarred };
 }
 
-export function fetchFaveStar(id) {
-  return function fetchFaveStarThunk(dispatch) {
+export function fetchFavoriteStatus(dashboardId) {
```
rusackas, line 101: Not sure what this change has to do with the PR. Maybe a rebase went wrong?

rusackas:
It seems the subheader now works as expected, as does the size control. I'll close this, but holler if I/we are missing something! Thank you for the PR, and happy to reopen if there's more to be done!

PR #33147, opened 2025-04-16
fix(theming): Fix various ci issues

msyavuz:
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/explore/utils.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/explore/utils.ts
@@ -75,7 +75,6 @@ export function saveChartToDashboard(chartName: string, dashboardName: string) {
   interceptDashboardGet();
   interceptUpdate();
   interceptExploreGet();
-  interceptChartDashboardsGet();
 
   cy.getBySel('query-save-button')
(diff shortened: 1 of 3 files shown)


superset-frontend/cypress-base/cypress/e2e/explore/utils.ts
```diff
@@ -91,7 +86,6 @@ export function saveChartToDashboard(chartName: string, dashboardName: string) {
           '.antd5-select-selection-search-input[aria-label="Select a dashboard"]',
         )
         .type(dashboardName, { force: true });
-      cy.wait('@chartDashboards');
```
geido, line 94: What is the reason this call does not need to be fired now? Is it actually pulling the data correctly?

geido (approved):

superset-frontend/cypress-base/cypress/e2e/explore/utils.ts
msyavuz: Yes, it actually pulls the data correctly without and this causes it to fail

PR #33159, opened 2025-04-17
refactor(Table): Use Table instead of html <table> in CollectionTable

msyavuz:
### SUMMARY

Refactor DatasourceEditor and CollectionTable components to use our Table component instead of html table.

Before:
image
image

After:
image
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/Datasource/CollectionTable.tsx
```diff
@@ -331,109 +276,114 @@ export default class CRUDCollection extends PureComponent<
...
+                {' '}
+                <InfoTooltipWithTrigger
+                  label={t('description')}
+                  tooltip={tooltip}
+                  placement="top"
+                />
+              </>
+            )}
+          </>
+        ),
```
EnxDev, line 319: here as well

superset-frontend/src/components/Datasource/CollectionTable.tsx
```diff
@@ -331,109 +276,114 @@ export default class CRUDCollection extends PureComponent<
...
+        render: (text: any, record: CollectionItem) =>
```
EnxDev, line 320: here as well

msyavuz: Same thing here, this could be anything and best way to handle this is through generics which we can't do.

superset-frontend/src/components/Datasource/CollectionTable.tsx
```diff
@@ -16,87 +16,30 @@
...
-      border-top: 0;
-      padding: 0;
-    }
-  `}
-`;
+import Table, {
+  ColumnsType,
+  SortOrder,
+  SorterResult,
+  TablePaginationConfig,
```
geido, line 40: These need to be imported as type for proper tree shaking

superset-frontend/src/components/Datasource/CollectionTable.tsx
```diff
@@ -113,6 +56,32 @@ const StyledButtonWrapper = styled.span`
   `}
 `;
 
+type CollectionItem = { id: string | number; [key: string]: any };
```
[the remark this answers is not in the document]
msyavuz: I need generics for this type but the only component that uses this is jsx so generics don't do anything. But we should definitely consider replacing these with proper types when DatasourceEditor is typescript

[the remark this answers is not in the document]
mistercrunch: Yeah I think that's fine. The original intent when I built this a long time ago was for it to become the basis for all CRUD in the app, but ended up going a different way. If anything in the future, we'd remove this base component and go more towards what other CRUDs are using, if that makes sense. For now we'll flag as "needs refactor" and can do minimal edits to get theming through.

mistercrunch (approved):
LGTM

EnxDev:
I believe this bug was introduced, could we fix it in a separate PR?

https://github.com/user-attachments/assets/78eb123f-6018-44ba-b03a-14fa46d3d7e4

PR #33180, opened 2025-04-18
feat(ci): flag chart metadata changes via labeler

geido:
### SUMMARY

A chart's `controlPanel` declares the `form_data` keys that get persisted on every slice using it, and its `transformProps` reads those keys back at render time. Changing either can silently break charts and dashboards that were saved against the previous shape, and nothing surfaces that risk during review.

This adds a `risk:chart-metadata` entry to `.github/labeler.yml`, so the `actions/labeler` run we already have flags any PR touching chart metadata — alongside the `risk:db-migration` and `risk:ci-script` warnings already in that file.

Coverage spans chart plugins, native filters, chart customizations and TimeTable, which all register through the same plugin API, plus the shared control panel sections. Native filter config is persisted into dashboard `json_metadata`, so it carries the same risk.

| Glob | Matches |
| --- | --- |
| `superset-frontend/**/{controlPanel,transformProps}.{ts,tsx}` | 125 files — chart plugins at any nesting depth, native filters, chart customizations |
| `superset-frontend/**/config/{controlPanel,transformProps}/index.ts` | TimeTable's config barrels |
| `superset-frontend/src/explore/controlPanels/**` | shared control panel sections |

131 files in total. The globs match metadata files by exact name, so unit tests like `test/transformProps.test.ts` don't trigger the label — the one exception is `explore/controlPanels/Separator.test.ts`, which sits in a directory that is otherwise entirely metadata.

`risk:chart-metadata` doesn't exist on the repo yet. `actions/labeler` will auto-create it, but as an undescribed grey label, so it's worth creating properly first:

```
gh label create "risk:chart-metadata" --repo apache/superset --color e67e22 \
  --description "PR changes a chart's controlPanel/transformProps; needs backward-compatibility review"
```

N/A — CI config only, no UI change.

### TESTING INSTRUCTIONS

`actions/labeler` runs on `pull_request_target`, so the new entry only takes effect once this is on `master` — it can't label this PR itself. To verify the globs without merging, run them through `minimatch`, which is what the pinned action uses internally:

```bash
npm i minimatch js-yaml
node -e "
const {Minimatch}=require('minimatch'), yaml=require('js-yaml'), fs=require('fs');
const globs=yaml.load(fs.readFileSync('.github/labeler.yml','utf8'))['risk:chart-metadata'][0]['changed-files'][0]['any-glob-to-any-file'];
const ms=globs.map(g=>new Minimatch(g,{dot:true}));
require('child_process').execSync('git ls-files').toString().trim().split('\n')
  .filter(f=>ms.some(m=>m.match(f))).forEach(f=>console.log(f));
"
```

Expect 131 files, no `.test.` files other than `Separator.test.ts`. Spot checks:

- `superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/Regular/Line/controlPanel.tsx` → labeled
- `superset-frontend/src/filters/components/Select/controlPanel.ts` → labeled
- `superset-frontend/plugins/plugin-chart-country-map/test/transformProps.test.ts` → not labeled
- `superset-frontend/src/explore/components/ControlPanelsContainer.tsx` → not labeled

After merge, open any PR touching a plugin's `controlPanel` and confirm the label lands.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/metadata-backward-compatibility.yml
+++ b/.github/workflows/metadata-backward-compatibility.yml
@@ -34,7 +34,7 @@ jobs:
             CHANGED_FILES=$(git diff --name-only origin/master...HEAD)
             echo "Changed files: $CHANGED_FILES"
 
-            if echo "$CHANGED_FILES" | grep -qE '^superset-frontend/plugins/.*/transformProps'; then
+            if echo "$CHANGED_FILES" | grep -qE '^superset-frontend/plugins/.*/(transformProps|controlPanel)'; then
                 echo "BACKWARD_COMPATIBILITY_CHECK_FAILED=true" >> $GITHUB_ENV
             else
                 echo "BACKWARD_COMPATIBILITY_CHECK_FAILED=false" >> $GITHUB_ENV
@@ -50,9 +50,9 @@ jobs:
                 if (!validationLabelPresent) {
                     core.setFailed(
                         [
-                            "🚨 We have identified potential changes to the chart metadata in `transformProps`.",
-                            "⚠️ This could affect backward compatibility.",
-                            "✅ To proceed, please review your changes and add the label:",
+                            "🚨 Identified potential changes to the chart metadata in `transformProps` or `controlPanel`.",
+                            "⚠️ Such changes may affect backward compatibility and cause charts dependent on previous metadata to break.",
+                            "✅ To proceed, please review your changes to confirm backward-compatibility and add the label:",
                             "`validation:backward-compatible`",
                         ].join("\n")
                     );
--- a/.github/workflows/metadata-backward-compatibility.yml
+++ b/.github/workflows/metadata-backward-compatibility.yml
@@ -6,7 +6,7 @@ on:
       - "master"
       - "[0-9].[0-9]*"
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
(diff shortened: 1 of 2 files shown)


kgabryje (approved):
Not an expert on github workflows so I'm not diving into the code, but I like the idea as a temporary solution for ensuring backward compatibility

.github/workflows/metadata-backward-compatibility.yml
```diff
@@ -0,0 +1,64 @@
...
+            submodules: recursive
+
+      - name: "Set up Git"
+        run: |
+          git fetch origin master
+
+      - name: "Check for changes in critical files"
+        id: check_changes
+        run: |
+            CHANGED_FILES=$(git diff --name-only origin/master...HEAD)
```
msyavuz, line 34: I think this runs diff against master but the base might be different?

msyavuz (approved):
Looks good to me with a small nit

rusackas:
We probably don't need a whole new action for this. I think you could just add a few filenames to [this file](https://github.com/apache/superset/blob/master/.github/labeler.yml).

geido:
> We might not need a whole new action for this. You could just add a few filenames to [this file](https://github.com/apache/superset/blob/master/.github/labeler.yml). However, that would apply to all branches, I believe, so I'm not sure if that solves the issue for you.

This is great. Let me change it!

rusackas:
Should we set this to Draft?

rusackas:
@geido did you want to keep this open, or shall we kill it off?

rusackas:
@geido kind of assuming this isn't coming back to life... I'll probably close it the next time it hits my stale draft radar :P

rusackas:
@geido I think the labeler.yml route we discussed supersedes this and it's been quiet since, so I'll close it. Ping me if you want to revive the workflow approach instead.

geido:
> We might not need a whole new action for this. You could just add a few filenames to [this file](https://github.com/apache/superset/blob/master/.github/labeler.yml).

Finally circling back on this — you were right, it's a handful of globs and the workflow is gone. Rebased on master.

Couple of things worth your eye: I widened it past `plugins/`, since native filters and chart customizations register through the same plugin API from `src/` and their config gets persisted into dashboard `json_metadata`. And it only flags now, no CI gate — shout if you'd rather it had teeth, `no-hold-label.yml` is already the pattern for that. Happy to rename the label to a `viz:` one too if that sits better with the TODO at the top of that file.

msyavuz (approved):

PR #33192, opened 2025-04-21
feat(i18n): Frontend add zh_TW Option

bestlong:
UI locale add zh_TW option and update Traditional Chinese translation


rusackas:
In general this looks fine to me. Does the correct flag already exist in the dropdown?

Adding @zhaoyongjie in case he (or someone he knows) might be able to help validate the translations.

rusackas (approved):

rusackas:
Well.... Claude AI seems to mostly agree with the translations, so let's just merge this thing :D Thanks1

Issue #25786, opened 2023-10-29
Cannot save a query with Parameters as a Dataset (Jinja template default values)


qleroy:
#### How to reproduce the bug

1. Go to the SQL lab and open a new query, select the default examples database 
2. Click on the three-dots menu and Parameters.
3. Enter the following JSON to specify defaults values for Jinja templates ([relevant doc](https://superset.apache.org/docs/installation/sql-templating/#jinja-templates) )
```json
{
  "my_table": "persons"
}
```
4. Enter the following text as the query
```sql
SELECT * FROM {{ my_table }}
```
Observe that you can execute the query, the context you specified in the Parameters is correctly injected because the results appear as expected.
6. Save the query as a dataset and the Fatal error is raised.

### Expected results

We should be able to save the Jinja-templated query as a new virtual dataset.

### Actual results

A fatal error is raised.

See below the logs of the superset_app container :
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/flask_appbuilder/api/__init__.py", line 110, in wraps
    return f(self, *args, **kwargs)
  File "/app/superset/views/base_api.py", line 127, in wraps
    raise ex
  File "/app/superset/views/base_api.py", line 121, in wraps
    duration, response = time_function(f, self, *args, **kwargs)
  File "/app/superset/utils/core.py", line 1526, in time_function
    response = func(*args, **kwargs)
  File "/app/superset/utils/log.py", line 255, in wrapper
    value = f(*args, **kwargs)
  File "/app/superset/views/base_api.py", line 93, in wraps
    return f(self, *args, **kwargs)
  File "/app/superset/datasets/api.py", line 315, in post
    new_model = CreateDatasetCommand(item).run()
  File "/app/superset/datasets/commands/create.py", line 50, in run
    dataset.fetch_metadata(commit=False)
  File "/app/superset/connectors/sqla/models.py", line 1247, in fetch_metadata
    new_columns = self.external_metadata()
  File "/app/superset/connectors/sqla/models.py", line 715, in external_metadata
    return get_virtual_table_metadata(dataset=self)
  File "/app/superset/connectors/sqla/utils.py", line 143, in get_virtual_table_metadata
    raise SupersetGenericDBErrorException(message=str(ex)) from ex
superset.exceptions.SupersetGenericDBErrorException: syntax error at or near "{"
LINE 1: SELECT * FROM {{ my_table }}
```
#### Screenshots

[...]

PR #33195, opened 2025-04-21
fix(sqllab): save datasets with template parameters

ethan-l-geotab:
### SUMMARY

ENABLE_TEMPLATE_PROCESSING=true

Looking to solve this: https://github.com/apache/superset/issues/25786.
Trying to allow for datasets with stuff in the parameters to save without having a fatal error.
All it does is add the template parameters at the creation (post) of the dataset
### TESTING INSTRUCTIONS
TAKEN FROM THE ISSUE:
1. Go to the SQL lab and open a new query, select the default examples database
2. Click on the three-dots menu and Parameters.
3. Enter the following JSON to specify defaults values for Jinja templates ([relevant doc](https://superset.apache.org/docs/installation/sql-templating/#jinja-templates) )
{
  "my_table": "persons"
}
4. Enter the following text as the query
SELECT * FROM {{ my_table }}
Observe that you can execute the query, the context you specified in the Parameters is correctly injected because the results appear as expected.
5. Save the query as a dataset and it should save correctly instead of giving fatal error.

### ADDITIONAL INFORMATION
- [x] Has associated issue: https://github.com/apache/superset/issues/25786
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila:
Should this be behind a checkbox in the save dataset modal/dialog? I'm thinking if users setting a `_filters` config just to emulate dashboard filters might not want these to persist to the dataset 🤔 a checkbox that's disabled by default (to match previous behavior) and can be enabled if desired could help maintaining flexibility.

rusackas:
CC @Vitor-Avila as well. I need to ramp up on my jinja intricacies!

ethan-l-geotab:
Thanks for the answers, @Vitor-Avila 
I made something like this:
image

Vitor-Avila:
that's awesome, @ethan-l-geotab! Thank you so much 🙏 

Do you think it would be possible to create some tests, perhaps in `SaveDatasetModal.test.tsx`? It would be great if we could test the element is rendered properly, and the payload sent based on the checkbox state.

ethan-l-geotab:
@Vitor-Avila. I added tests in `SaveDatasetModal.test.tsx`! 

1. to check the render of the checkbox (only render if the flag is enabled)
2. to check if "createDatasource" has been called with template parameters based on the checkbox status

Vitor-Avila:
ooof forgot to enable the FF in the ephemeral -- will have to redo that

Vitor-Avila:
Couldn't make the FF work so I tested locally -- thank you so much for this feature!

Vitor-Avila (approved):
LGTM!

qleroy:
Awesome ! Thank you @ethan-l-geotab  !

PR #33172, opened 2025-04-17
fix: subheader should show as subtitle

eschutho:
### SUMMARY

On some recent changes to the subheader, we lost the subtitle. This maps the subheader to the subtitle for bigNumber and bigNumber with trendline. 

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

After:
(showing the form data as subheader and subtitle)
Screenshot 2025-04-17 at 4 23 01 PM
Screenshot 2025-04-17 at 4 22 18 PM
Screenshot 2025-04-17 at 4 21 18 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #33196, opened 2025-04-21
fix(chart): Restore subheader used in bignumber with trendline

LevisNgigi:
### SUMMARY
We lost the subheader in the big number with trendline as we tried to map it to subtitle.This aims at restoring the subheader being used in big number with trendline.
related to this [pr](https://github.com/apache/superset/pull/33172)

BEFORE
before

AFTER
after

### TESTING INSTRUCTIONS
1)Create or open a chart using  Big Number with Trendline chart configured with subheader/comparison suffix such as Wow,MoM etc
2)Confirm that the chart displays the subheader  as expected.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


geido:
Can you please link the original PR that introduced the issue for traceability? It would also help adding a bit more to the description for others to validate the fix. Thanks for your help!

LevisNgigi:
> Can you please link the original PR that introduced the issue for traceability? It would also help adding a bit more to the description for others to validate the fix. Thanks for your help!

I have updated accordingly,Thank you.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -188,6 +188,47 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
     );
   }
 
+  renderSubheader(maxHeight: number) {
```
eschutho, line 191: @LevisNgigi is this component only used for the metrics comparisons? If so, can we rename it something more specific to that?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -188,6 +188,47 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
...
+
+    const NO_DATA_OR_HASNT_LANDED = t(
+      'No data after filtering or data is NULL for the latest time record',
+    );
+    const NO_DATA = t(
+      'Try applying different filters or ensuring your datasource has data',
+    );
+    let text = subheader;
+    if (bigNumber === null) {
+      text = bigNumberFallback ? NO_DATA : NO_DATA_OR_HASNT_LANDED;
```
eschutho, line 203: I'm also thinking that if we were to use this component only for metrics comparisons, then we can move this fallback text to the subtitle instead. How does that sound?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
LevisNgigi: Yes it is used for metric comparison to display the subhearder that appears below the big number.

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
LevisNgigi: That sounds fine, it can be cleaner that way I think.

eschutho: I don't believe so. I mapped that subhead to the subtitle in transformProps, but let's test it out to be doubly sure!

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -188,19 +188,21 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
...
       const container = this.createTemporaryContainer();
       document.body.append(container);
       try {
         fontSize = computeMaxFontSize({
-          text: subtitle,
+          text,
           maxWidth: width * 0.9,
           maxHeight,
-          className: 'subtitle-line',
+          className: 'subheader-line',
```
eschutho, line 205:
```suggestion
          className: 'metrics-comparison-line',
```
Can we also rename this?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -209,19 +211,65 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
 
       return (
         <div
-          className="subtitle-line"
+          className="subheader-line"
```
eschutho, line 214: here too?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -209,19 +211,65 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
...
+        maxHeight,
+        className: 'subtitle-line',
+        container,
+      });
+      container.remove();
+
+      return (
+        <>
+          <div
+            className="subtitle-line subheader-line"
```
eschutho, line 259: what does the `subheader-line` class do?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
```diff
@@ -294,6 +343,11 @@ class BigNumberVis extends PureComponent<BigNumberVizProps> {
             {this.renderHeader(
               Math.ceil(headerFontSize * (1 - PROPORTION.TRENDLINE) * height),
             )}
+            {this.rendermetricComparisonSummary(
+              Math.ceil(
+                subheaderFontSize * (1 - PROPORTION.TRENDLINE) * height,
```
eschutho, line 348: can we rename this to metricComparisonFontSize?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
eschutho: I see, this is for backwards compatibility. 👍

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
LevisNgigi: I restored the classname to how it was originally so that it does not break any css customizations that users will have?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
LevisNgigi: We can,this means the subheader font size in the UI needs updating [here](https://github.com/apache/superset/blob/master/superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/sharedControls.ts#L91) well?

superset-frontend/plugins/plugin-chart-echarts/src/BigNumber/BigNumberViz.tsx
LevisNgigi: Yes some css customizations were targetting the subheader-line and without it the customizations did not apply.

Antonio-RiveroMartnez:
Hey! The suffix is working as expected, however, when you set a subtitle, and you remove it, it's not clearing it up:

https://github.com/user-attachments/assets/a628313e-2e1f-4bfa-97aa-b4dd8a81a900

Antonio-RiveroMartnez (approved):
Ok, subheader and subtitle working as expected

https://github.com/user-attachments/assets/bad2ecb5-780b-45f5-a50b-32e8f46c0fc3

Vitor-Avila:
Performed below manual tests on my end:
1. Spin up Superset from sha/commit `164a07e2be7854520070a7d845eec449405c4840` (last one before the PR that introduced `subtitle` - [ref](https://github.com/apache/superset/commit/164a07e2be7854520070a7d845eec449405c4840)).
2. Created 5 Big Numbers (one for each subheader font size) and added them to a dashboard.
3. Did the same with the Big Number and Trendline chart type.
4. Checked out to commit `d75ff9e784ca571e8498febb3b492874118312c8` (the one introducing subtitles - [ref](https://github.com/apache/superset/commit/d75ff9e784ca571e8498febb3b492874118312c8)).
5. Validated that the subheader text have disappeared for the Big Numbers. Big Numbers with Trendline were still ok.
6. Created 5 more Big Numbers using the new schema (with `subtitle`) and added to the same dashboard.
7. Checked out to commit `41402617970368f0b14ff4e7acc1d271d5bcdd89` (the one that fixes the missing headers - [ref](https://github.com/apache/superset/commit/41402617970368f0b14ff4e7acc1d271d5bcdd89)). Validated that the Big Numbers are now showing the headers. Big Number with Trendline no longer show the subtitle. Also, the subtitle font sizes would show correct in the dashboard, but when accessing the charts in Explore they would always show up as **Small** (default value).
8. Checked out to this PR. Validated that:
 - Subheaders for Big Numbers with Trendlines are back. ✅ 
 - Their font size is showing correctly both in charts and dashboards. ✅ 
 - The subtitle for Big Numbers appear duplicated: 🔴 
image
 - Accessing the chart in Explore does not show the subtitle content (the subtitle size is also not correct): 🔴 
image

Vitor-Avila:
@LevisNgigi my last comment was longer just to highlight the testing scenarios and flows, but to summarize, these are the remaining issues:
* Big Numbers created prior to [the `subtitle` PR](https://github.com/apache/superset/pull/32975) show the subheader text duplicated.
* Accessing the chart directly in Explore does not carry over the `subtitle` config (both text and size).


To easily repro this:
1. Spin up Superset in dev mode.
2. Check out to commit `164a07e2be7854520070a7d845eec449405c4840`.
3. Create a dashboard including one Big Number for each subheader size available. For example:
image
4. Checkout to this branch.
5. Refresh the dashboard:
image
6. Accessing a chart from this dashboard does not show the subtitle config:
image

Let me know if you have any questions!

Vitor-Avila:
@LevisNgigi glad to help and excited we're making progress here! 🙌

hmm could we make the migration dynamic based on the viz type? If it's a Big Number migrate `subheader` to `subtitle` and `subheaderFontSize` to `subtitleFontSize`, if Big Number with Trendline don't do it? I believe we can differentiate the two via their params, right? Ideally this migration could happen on the fly by the frontend.

Let me know if that makes sense/is possible -- I'm not super close to our frontend logic so it's totally possible that this wouldn't work/isn't possible.

superset-frontend/src/explore/components/ExploreViewContainer/index.jsx
```diff
@@ -768,6 +779,30 @@ function mapStateToProps(state) {
...
+    form_data.viz_type === 'big_number_total' &&
+    slice?.form_data?.subheader_font_size &&
+    (!controls.subtitle_font_size?.value ||
+      controls.subtitle_font_size.value === '')
+  ) {
+    controls.subtitle_font_size = {
+      ...controls.subtitle_font_size,
+      value: slice.form_data.subheader_font_size,
+    };
+  }
```
Vitor-Avila, line 802: I'm suggesting this change for two reasons:
- It moves the second change (`controls.subtitle_font_size =  slice.form_data.subheader_font_size`) as part of the first `if` block. This is mostly because if the chart has a `subheader` config and does not have a `subtitle` config, then `subheader_font_size` should be the source of truth. For safety, just checking if `slice?.form_data?.subheader_font_size` exists.
- In the current implementation, `!controls.subtitle_font_size?.value` is evaluated. I believe `controls.subtitle_font_size` always gets a default value, so that second `if` would never really execute and instead the old `subheader` ends up with the default size.


```suggestion
  if (
    form_data.viz_type === 'big_number_total' &&
    slice?.form_data?.subheader &&
    (!controls.subtitle?.value || controls.subtitle.value === '')
  ) {
    controls.subtitle = {
      ...controls.subtitle,
      value: slice.form_data.subheader,
    };
    if (slice?.form_data?.subheader_font_size) {
      controls.subtitle_font_size = {
        ...controls.subtitle_font_size,
        value: slice.form_data.subheader_font_size,
      };
    }
  }
```

Vitor-Avila:
@LevisNgigi thank you so much for taking the time to implement these fixes! 🙏 I tested the flow again and:
- Big Numbers with Trendline seem to be working properly. ✅ 
- Existing Big Numbers are no longer showing both `subheader` and `subtitle` in the dashboard view. Just one is displayed. ✅ 
- Accessing these Big Numbers in Explore shows a single entry as well. ✅ 
- Accessing these Big Numbers in Explore defaults the subtitle size to the default value (**Small**) as opposed to respecting the old value from subheader. 🔴 

I suggested a potential fix for this remaining issue, but I didn't want to commit it as I want your thoughts on it first. If you're comfortable with it, feel free to push it to this branch and I can do one final round of tests. 🙌

Vitor-Avila (approved):
LGTM!

PR #33204, opened 2025-04-22
feat: custom limit sqllab

Abhishek-kumar-ISM:
Earlier in SQL lab we had limited options to use as **limit of number of rows in sql results**. These options were only [10,100,1000, 10000, 100000]

Earlier it used to look like : 
107 99 237 104_8013_sqllab_ (3)

I have added an option to put your custom limit while fetching sql query results as below : 
107 99 237 104_8013_sqllab_
107 99 237 104_8013_sqllab_ (1)
107 99 237 104_8013_sqllab_ (2)

One can put their custom limit in input text and hit enter, and that limit will be applied while fetching number of rows in sql results.

**Use cases:**
Actually earlier once for me requirement came to fetch 700 results, so i had two options to complete the task i.e (put limit 700 in query itself, but in that case this limit feature does not makes any sense, it becomes useless), and second way is fetch 1000 results, then download csv and then remove 300 rows(but this results in lot of overhead)
-> To resolve above drawbacks this feature of custom limit is helpfull.


Abhishek-kumar-samsung:
> LGTM, but I wonder if we should replace the existing component with a Select (configured to let you add custom values) rather than adding an input to the current component (which looks like a bit of a design system outlier). @geido or @kasiazjc might have additional opinions here.

In select too, i would have implemented in same fashion because rest options in dropdown are just clickable and this one is customized to take input, thats why i didn't changed existing implementation of dropdown from Menu to Select, just added a new Input field as a new menu item in current Menu.

Abhishek-kumar-samsung:
@rusackas i did few more changes,
earlier on clicking enter the dropdown was not closing automatically, so i moved input to Menu.Item
And then testcases was having issue because number of items in menu increased, so then did some changes in menu items.

Abhishek-kumar-samsung:
Pls check

Abhishek-kumar-samsung:
@rusackas @geido @kasiazjc @michael-s-molina @justinpark 
Its been couple of days without any reply, so just thought to followup.
Can you pls review the PR.

michael-s-molina:
Thanks for the PR @Abhishek-kumar-samsung.
- The menu with input is not a common pattern used in Superset. For consistency, a Select would be more appropriate as @rusackas mentioned.
- The max limit is calculated from [SQL_MAX_ROW](https://github.com/apache/superset/blob/c029b532d47d6a611535e6a6c846c959a4fbd707/superset/config.py#L970) and we can't allow that limit to be exceeded. This means that the Select would need to reject inputs that go beyond `SQL_MAX_ROW` or have negative values.
- I'm not sure about the usefulness of this feature given that you can simply apply a `LIMIT` to the query which should be something users of SQL Lab will know. Keep in mind that Superset displays a warning when the values are limited by the query. 

Screenshot 2025-04-29 at 08 46 26
<br><br>
Maybe @justinpark @kasiazjc @rusackas have some thoughts about this.

Abhishek-kumar-samsung:
Okay i will convert Menu to Select and also put maxRow condition

Abhishek-kumar-ISM:
I have updated by changing menu to select as was suggested, pls review

sss1

mistercrunch:
Noting that I believe there is logic to use the `LIMIT` from the SQL-provided code (may not work with `ROWNUM` or `TOP` databases) if it's within the config-allowed range. I haven't looked at the code in years, so things might have changed, but wanted to note that. My expectation is that it'll use the user-provided LIMIT as long as it's smaller than the allowed limit. 

If adding a new control/setting, we should make sure it also obeys to the same rule, always enforcing the config-defined hard limit. This is important as someone could crash a pod on the backend by overflowing it with much more data than it can handle.

Abhishek-kumar-ISM:
I have kept dropdown items less than max row, and also kept this line : 
if (limit_val > maxRow) limit_val = maxRow;

so i think limits will be satisfied.

rusackas:
Hi @Abhishek-kumar-ISM 👋

This PR has merge conflicts and has been inactive for about 5 months. Converting to draft until resolved.

When you have time, please rebase on `master` to resolve the conflicts.

If you're no longer able to work on this, let us know and we can close it or have someone else pick it up.

Thanks for your contribution!

rusackas:
Closing this stale draft (idle several months). The custom SQL Lab row-limit idea is reasonable, but the change needs a rebase against the reworked SQL Lab UI and discussion vs. the existing configurable limit options. Please reopen a fresh PR if you'd like to pursue it. Thanks for the contribution!

PR #33214, opened 2025-04-22
fix(export): Full CSV/Excel exports respecting SQL_MAX_ROW config

Vitor-Avila:
### SUMMARY
Superset provides two CSV/Excel export options:
* **Export to .CSV/Excel**: Downloads the chart data (respecting the chart's row limit).
* **Export to full .CSV/Excel**: This option is behind the `ALLOW_FULL_CSV_EXPORT` FF (disabled by default) and only available to Table charts. It would download all data from the chart query (up to the `SQL_MAX_ROW` config).

The full download was not working properly and instead of getting the `SQL_MAX_ROW` limit, it was getting a default limit of `5000`.

This is a follow up to https://github.com/apache/superset/pull/31241 and https://github.com/apache/superset/pull/31720, to fix this flow so that the full download respects the `SQL_MAX_ROW` coinfig.

No UI changes.

Before this change, **Export to full .CSV/Excel** results in a file with <= 5k rows. Now, it results in a file with <= `SQL_MAX_ROW` rows.

### TESTING INSTRUCTIONS
Updated one frontend test. For manual testing:
1. Enable the `ALLOW_FULL_CSV_EXPORT` FF.
2. Set `SQL_MAX_ROW` to `100000`.
3. Create a table chart using the `flights` dataset with the row limit set to `15000` and add it to a dashboard.
4. Access the dashboard.
5. Click on the chart menu, then **Download > Export to .CSV**. Validate the file includes 15k rows of data.
6. Click on the chart menu, then **Download > Export to full .CSV**. Validate the file includes ~55k rows of data.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/gridComponents/Chart.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/Chart.jsx
@@ -150,6 +150,9 @@ const Chart = props => {
   const emitCrossFilters = useSelector(
     state => !!state.dashboardInfo.crossFiltersEnabled,
   );
+  const maxRows = useSelector(
+    state => state.dashboardInfo.common.conf.SQL_MAX_ROW,
+  );
   const datasource = useSelector(
     state =>
       (chart &&
@@ -360,9 +363,7 @@ const Chart = props => {
         is_cached: props.isCached,
       });
       exportChart({
-        formData: isFullCSV
-          ? { ...formData, row_limit: props.maxRows }
-          : formData,
+        formData: isFullCSV ? { ...formData, row_limit: maxRows } : formData,
         resultType: isPivot ? 'post_processed' : 'full',
         resultFormat: format,
         force: true,
--- a/superset-frontend/src/dashboard/components/gridComponents/Chart.test.jsx
+++ b/superset-frontend/src/dashboard/components/gridComponents/Chart.test.jsx
@@ -34,7 +34,7 @@ const props = {
   height: 100,
   updateSliceName() {},
   // from redux
-  maxRows: 666,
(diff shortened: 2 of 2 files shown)


kgabryje (approved):
LGTM

rusackas:
If I understand correctly, the FF currently lets you grab the entire table regardless of SQL_MAX_ROW. On the one hand, I appreciate this fix, as such a feature could nuke the DB and melt your computer. 

On the other hand, this is a departure from behavior some people have come to expect. It's _kind of_ a breaking change, but that's debatable... this almost feels like a bug fix given the above. Nevertheless, I think there should be a note in UPDATING.md about it, and if you want bonus points, maybe we could surface a tooltip or something that propagates SQL_MAX_ROW so the user knows how many rows they can expect.

Vitor-Avila:
@rusackas I think this was already the case before. @kgabryje recently refactored the dashboard loading for performance improvements (https://github.com/apache/superset/pull/31241), and if you check the `superset-frontend/src/dashboard/containers/Chart.jsx` file in there `maxRows` was already set it this way:
```
...
maxRows: common.conf.SQL_MAX_ROW,
...
```

I thought about that too (we don't heavily expose this FF in Preset so I wasn't sure) but I think it always had this "precautionary" measure.

rusackas:
If this doesn't really change the existing behavior, then I'm fine with it as is, merge away! If it does change the behavior, we should _probably_ note it in UPDATING.md so folks are made ware.

Issue #33162, opened 2025-04-17
Marshmallow 4.0.0 breaks superset init


MattBodey:
### Bug description

On a fresh install of superset, since marshmallow 4.0.0 released https://pypi.org/project/marshmallow/, superset fails to init due to an error validating parameters 

It is possible to work around this by manually installing marshmallow at 3.26.1

pip3 install marshmallow==3.26.1

It looks like the version is being set in requirements/development.txt, so I'm unsure as to where the version update is occurring 

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Not applicable

### Additional context

 File "/home/ec2-user/.local/lib/python3.9/site-packages/superset/charts/schemas.py", line 621, in ChartDataProphetOptionsSchema
    periods = fields.Integer(
  File "/home/ec2-user/.local/lib/python3.9/site-packages/marshmallow/fields.py", line 977, in __init__
    super().__init__(as_string=as_string, **kwargs)
  File "/home/ec2-user/.local/lib/python3.9/site-packages/marshmallow/fields.py", line 924, in __init__
    super().__init__(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'min'

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [x] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
Is there somewhere we need to pin the dependency? It looks like Superset already installs 3.26.1. I'm not sure about the details of the failure, but I know we also use `marshmallow-sqlalchemy` which doesn't seem to support marshmallow 4.0 yet if I'm understanding correctly.

ColinMaudry:
I followed the [PyPi install instructions](https://superset.apache.org/docs/installation/pypi/#os-dependencies), and running `pip install marshmallow==3.26.1` just before `superset db upgrade` did the trick.

amotl:
Hi there,

we are running a little integration test suite against Apache Superset and CrateDB, in order to validate both work together well. Two days ago, CI apparently started tripping on it.

```python
TypeError: Field.__init__() got an unexpected keyword argument 'minLength'
```

Do you think we need to adjust a few details, or is it just about [marshmallow-4.0.0](https://pypi.org/project/marshmallow/4.0.0/), which has been released two days ago, as suggested above?

With kind regards,
[...]

PR #33216, opened 2025-04-23
fix: Downgrade to marshmallow<4

amotl:
## Problem
With Marshmallow 4, just released, Superset 4 trips like:
```python
TypeError: Field.__init__() got an unexpected keyword argument 'minLength'
```

## Solution
Pin dependency to `marshmallow<4`.

## References
- GH-33162

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,8 @@ dependencies = [
     "jsonpath-ng>=1.6.1, <2",
     "Mako>=1.2.2",
     "markdown>=3.0",
+    # marshmallow>=4 has issues: https://github.com/apache/superset/issues/33162
+    "marshmallow<4",
     "msgpack>=1.0.0, <1.1",
     "nh3>=0.2.11, <0.3",
     "numpy==1.23.5",


pyproject.toml
```diff
@@ -64,6 +64,7 @@ dependencies = [
     "jsonpath-ng>=1.6.1, <2",
     "Mako>=1.2.2",
     "markdown>=3.0",
+    "marshmallow<4",
```
mistercrunch, line 67: Let's always put a comment above when setting an upper bound that's not based on pure semver assumptions, something like "marshmallow>=4 has shows issues related to `'minLength'`", ...

amotl: Thank you. I didn't see such comments on other spots, but I've just added it, effectively linking to the corresponding issue on GitHub.

mistercrunch (approved):
LGTM - curious about your internal release process that lead to that issue, do you run your own `uv pip compile --upgrade` process internally? Or maybe you have your own snyk/dependabot-type setup in place and upgrade certain things?

In any case, would be great to centralize the dependency management and upgrades in the main repo as much as possible. Could certainly use help with the process if you wanted to get involved in dep management.

amotl:
@mistercrunch: Thanks for merging.

> curious about your internal release process that lead to that issue

You can inspect the testing rig which flagged this regression through its nightly CI runs over here. HTH, otherwise feel free to ask.

- https://github.com/crate/cratedb-examples/tree/main/application/apache-superset
- https://github.com/crate/cratedb-examples/actions/workflows/application-apache-superset.yml

amotl:
Hi again. This still needs a release, in order to resolve the issue completely for average downstream users.

- https://github.com/apache/superset/issues/33162#issuecomment-2957291330

/cc @mistercrunch, @rusackas

amotl:
> In any case, would be great to centralize the dependency management and upgrades in the main repo as much as possible. Could certainly use help with the process if you wanted to get involved in dep management.

It looks like this might still be an unresolved issue? Maybe I can free up some time for that, as I guess it might be a one-time action? On the other hand, I strongly believe Superset has its intriciacies, so I certainly would need to learn a significant amount of your current philosophy and about your most prominent pain points in order to possibly improve things thereafter, if at all.

mistercrunch:
This could be pull in our various active LTS branches though release managers are already pretty busy with everything and LTS-done-right is difficult, and I believe it's an area where only committers can do certain things (people without write access to the repo simply cannot push a release through for good reasons). Tagging the folks that have been doing release management over the past few releases here: @michael-s-molina @sadpandajoe 

On your side though, wondering if your build process can build upon Superset's pinned dependencies. Say if you use `pip compile` or `uv pip compile` for you project, you may be able to add a constraint on our pinned deps in your build process. Notice the `-c` in this script that compiles our deps here -> https://github.com/apache/superset/blob/master/scripts/uv-pip-compile.sh#L28 as you may be able to do something similar. Now it may conflict with with core app's dependencies, but at least you would integrate our pinned requirements along yours, and have to resolve them pre-build.

Superset is/was designed to be in a fairly standalone mode, meaning we're assuming your deploying the Superset app, not another app that happens to be integrating Superset. The way I've seen Superset be a dependency of a wider project is typically by creating a wrapper/shell around it while bringing the Superset repo as a git submodule inside your project or repo. Typically the wrapper will have CI/cloud-provisioning things like terraform/docker and whatever is needed to integrate Superset in your organization's typical build process and requirements. Clearly there's a lot of different constraints for different organizations (everyone has different reqs on how they deploy services internally), and it's been difficult to share the various building blocks and reusable components that caters to such a wide variety of ways of doing things ... Even things like the Helm chart or Dockerifle often needs to be forked for things to work. Similar with terraform, usually there are more things in a typical terraform package that are related to your org's specificities as there are things related to "how Superset should be deployed".

Anyhow, open to helping out around it and offering the buildings blocks that are useful to most. 

One thing I'd like to see if "reference implementations" OUTSIDE our repos showing various ways around how this can be done in different places...

sadpandajoe:
I think I had already cherried this as part of 4.1.3rc3, so it should be there already.

zhangyijun:
Lastest 5.0.0 version release, this bug is still exists. pip install will install **marshmallow-4.0.0. CAN NOT** work anything untill mannually run pip install -U marshmallow==3.26.1.

betodealmeida:
Pinned it here: https://github.com/apache/superset/pull/33978

PR #33222, opened 2025-04-23
fix: Fixed numeric filtering issue in Superset

### LINKED ISSUE #33206: Incorrect filtering for numeric values with different decimal precision
### Summary

There is an issue with filtering numeric values in Superset when values have different numbers of digits after the decimal point.

### Steps to Reproduce

1. Add a filter for a numeric field that contains both integer and decimal values (e.g., 21, 21.8, 25.35, etc.).

Image

2. First select an integer value (for example, 21) in the filter, and then fractional values.

Image

3. The query returns only the rows where the value is exactly 21 (i.e., without any decimal places), excluding 21.8, 25.35, etc.

Image

Image

4. Now, if you select a decimal value first (e.g., 21.8 or 25.35), the filtering works as expected, returning all matching rows regardless of decimal precision.

Image

Image

### Expected Behavior

Filtering by numeric value should match all equivalent numeric representations (e.g., 21, 21.8, 25.35) regardless of how many decimal places are used.

### Actual Behavior

Filtering is inconsistent depending on whether an integer or decimal value is selected first. Selecting an integer leads to an overly strict match that excludes decimal representations of the same value.

### Databases Affected

This issue has been observed across all three databases tested:

- ClickHouse  
- Impala  
- PostgreSQL  

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

I don't know

### Browser

Not applicable

### Additional context

_No response_

### Checklist

Sayan199:
### SUMMARY

Fixes #33206

### TESTING INSTRUCTIONS

In order to test this, we need to 
1. Add a filter for a numeric field that contains both integer and decimal values (e.g., 21, 21.8, 25.35, etc.).
2. First select an integer value (for example, 21) in the filter, and then fractional values.
This should now be able to filter and give results that do match with 21 but also with 21.8 and 25.35

### ADDITIONAL INFORMATION

- [ ] Has associated issue: #33206
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Looks like there are a bunch of CI tasks failing, but we should probably start by running `pre-commit` to fix the linting/formatting issues, and see if anything else gets unblocked. 

```
pre-commit install
```
```
pre-commit run --all-files
```

michael-s-molina:
Possible duplicate of https://github.com/apache/superset/pull/33230

rusackas:
@Sayan199 curious if you're able to fix the CI issues, or if you think this is indeed a duplicate of https://github.com/apache/superset/pull/33230 (where you're also welcome to review or add thoughts)

rusackas:
Thanks for the contribution. After reviewing the diff more carefully, there's a structural bug that would need to be addressed: the `if target_generic_type == utils.GenericDataType.BOOLEAN` block and final `return value` at the bottom of `handle_single_value` were accidentally dedented out of that nested function into the outer `filter_values_handler` scope, which would break boolean filter handling.

The possible duplicate (#33230) was also closed without merging, so that concern is moot, but the underlying fix here needs another look before it would be safe to merge.

rusackas:
Marking as draft until it an get further attention.

rusackas:
Thanks for the contribution, @Sayan199. As noted in review, the diff has a structural bug — the BOOLEAN branch and the final return value in handle_single_value got dedented out of the intended block — and it's been idle since. Closing this draft; the underlying issue #33206 stays open and we'll pick it back up there (ideally with a regression test). Feel free to reopen a fresh PR if you'd like to continue.

PR #33234, opened 2025-04-24
fix(standalone): Ensure correct URL param value for standalone mode

Vitor-Avila:
### SUMMARY
In order to generate a chart screenshot via async mode, the celery worker accesses `https://${SUPERSET_DOMAIN}/superset/slice/${ID}/?standalone=true`. We have logic in place to add the `form_data` to the query parameters, and also persist the `standalone` parameter. The latter was done by getting the parameter name from an `Enum`.

The logic was currently not explicitly accessing the Enum's `.value`, which would not work properly for older Python versions (for example, `3.10.x`).

This PR should enforce we're using it's value.

No UI changes.

### TESTING INSTRUCTIONS
Run Superset with Python `3.10.x` and validate that downloading a chart screenshot via the API works properly.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/views/core.py
+++ b/superset/views/core.py
@@ -135,7 +135,7 @@ class Superset(BaseSupersetView):
         endpoint_params = {"form_data": f"{form_data}"}
 
         if ReservedUrlParameters.is_standalone_mode():
-            endpoint_params[ReservedUrlParameters.STANDALONE] = "true"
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM!

PR #33242, opened 2025-04-25
feat(Native Filters): Time Filter Plugin Redesign

amaannawab923:
### SUMMARY

This is currently a draft pr which will introduce a new more beautiful version of Time Filter Plugin 
Screenshot 2025-04-25 at 5 59 30 PM

This is a long way to go & a work in progress

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Thank you for taking a stab at this! I added some folks who have been thinking on it quite a bit.

superset-frontend/src/explore/components/controls/DateFilterControl/DateFilterLabel.tsx
```diff
@@ -277,78 +278,7 @@ export default function DateFilterLabel(props: DateFilterControlProps) {
...
-          disabled={!validTimeRange}
-          key="apply"
-          onClick={onSave}
-          data-test={DateFilterTestKey.ApplyButton}
-        >
-          {t('APPLY')}
-        </Button>
-      </div>
-    </ContentStyleWrapper>
+    <DateFilter onApply={e => console.log(e)} onCancel={onHide} />
```
eschutho, line 281: Nit-console

eschutho:
Thanks for this contribution @amaannawab923! The main concern in the sip was that we should stay close to the antd library. @michael-s-molina and @geido does this design line up with antd? I did a quick look through and saw a fair amount of css so not sure what your thoughts are.

geido:
> Thanks for this contribution @amaannawab923! The main concern in the sip was that we should stay close to the antd library. @michael-s-molina and @geido does this design line up with antd? I did a quick look through and saw a fair amount of css so not sure what your thoughts are.

This seems to be a semi-custom implementation. Curious @amaannawab923 what drove the customization decision VS going with what Ant Design provides and build on top of that?

amaannawab923:
> > Thanks for this contribution @amaannawab923! The main concern in the sip was that we should stay close to the antd library. @michael-s-molina and @geido does this design line up with antd? I did a quick look through and saw a fair amount of css so not sure what your thoughts are.
> 
> This seems to be a semi-custom implementation. Curious @amaannawab923 what drove the customization decision VS going with what Ant Design provides and build on top of that?

Thanks for the review @eschutho  @geido  ,  i tried to address all of the concerns that were part of sip discussion , and i tried to address all of the concerns mentioned in the discussion , even in the earlier version of time filter plugin, we were using basic antd components like Select, Icons & Buttons and building on top of that & i have kept the same consistency here with 
import { Tabs, Button, Input, Radio } from 'antd';
these components from ant design... 
I tried to take control over all of the css as per the discussion to address the End user concerns of simplified ui & this custom Implementation doesnt use any extra libraries ... Still relies on ant design for its core components  but with custom css on top of ant design i am able to control the component better & make it more beautiful

rusackas:
I don't love all the myriad tags, perhaps we could organize these a little better, to have Previous/Current/etc tags, and then pick Day/Month/Year/etc from there? There's a high cognitive load to find the range you want in all those buttons.

amaannawab923:
@rusackas  I agree , I tried my best to bring all of the time ranges of last previous and current and at times it really makes it very confusing that do we even require so many time categories in the first place … if I could get some better design idea from @kasiazjc or a better consensus as to what we should keep then we can perhaps agree on a better design & approach

rusackas:
Hi @amaannawab923! This was a promising start on the time filter redesign. Are you still interested in continuing this work? The PR has some conflicts and CI issues that would need attention. Let us know!

PR #33247, opened 2025-04-25
feat(deckgl): add point radius controls for GeoJSON layer

### LINKED ISSUE #31176: Impossible to set circle size less then 100 meters in deck.gl geojson chart
Hello!
The issue is in impossibility to change size of point circles in a deck.gl geojson chart. Although the interface allows setting the size manually with any figure, if it is less than 100, all point circles disappear.
Thus, it's only possible to build enormous circles, that sometimes hide the necessary information below them. Moreover, there are no options in interface to switch static size (meters) to a dynamic one (pixels), depending on a map scale.

image (1)
image (2)
image (3)

I've also tried to set the size manually by editing a geojson itself. I used the folowing options in "properties" dict to fit it, but didn't succeed. 

```
[geojson building in PostGIS]...

'properties', jsonb_build_object(
        'guid', guid,
        'strokeColor', '#0000FF',
        
        'pointRadius', 30,
        'getRadius', 30,
        'radius', 30,
        'getPointRadius', 30,
        'size', 30,
        'getSize', 30,
        'marker-size', 30,
        'radiusPixels', 30,
        'getRadiusPixels', 30,
        'point_size', 30,
        'circle-radius', 30,
        'point_size_pixels', 30,
        'getPointSize', 30,
        'pointRadiusUnits', 'pixels',
        'radiusUnits', 'pixels'
    )
```

So, there are two questions:
1) Is it possible to set the correct size now (any in (0;100) pixels, dynamically scaling with map)?
2) Is it possible to modify interface menu for deck.gl geojson charts to allow user setting proper radii directly there?

Thank you!

P.S.: Superset version: 3.1.2; also tried in 4.1.1 with the same result.

rusackas:
### SUMMARY

Adds new controls to configure point radius for GeoJSON features, addressing issue #31176 where users couldn't set circle sizes below 100 meters or use pixel-based sizing.

**New controls:**
- **Point Radius**: Set the base radius value (default: 10)
- **Point Radius Units**: Choose between pixels, meters, or common units (default: pixels for intuitive screen-space sizing)
- **Point Radius Scale**: Multiplier for uniform scaling (default: 1)

The controls map directly to deck.gl's GeoJsonLayer API (`getPointRadius`, `pointRadiusUnits`, `pointRadiusScale`).

**Key improvements over original draft:**
- Default units changed to `pixels` (more intuitive for most users)
- Added `renderTrigger: true` to all controls for live preview
- Added helpful descriptions explaining how the controls interact
- Changed `point_radius_scale` default from 100 to 1 (so Point Radius values work intuitively)
- Dropped the unrelated "unintended commit" from the branch

Before: Only `Point Radius Scale` control exists, defaults to null, and only works in meters - making small circles impossible.

After: Users can set explicit radius values in pixels, meters, or common units with a scale multiplier.

### TESTING INSTRUCTIONS
1. Create a GeoJSON chart with point features
2. In the chart controls, find the new "Point Radius", "Point Radius Scale", and "Point Radius Units" controls
3. Verify:
   - Setting Point Radius to small values (1, 5, 10) with Units = "Pixels" renders appropriately sized circles
   - Changing Units to "Meters" makes circles scale with zoom
   - Point Radius Scale multiplies the base radius as expected

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #31176
- [ ] Required feature flags:
- [x] Changes UI
- [ ] Includes DB Migration

🤖 Generated with [Claude Code](https://claude.com/claude-code)


superset-frontend/plugins/preset-chart-deckgl/src/layers/Geojson/controlPanel.ts
```diff
@@ -343,15 +345,56 @@ const config: ControlPanelConfig = {
...
+          {
+            name: 'point_radius_units',
+            config: {
+              type: 'SelectControl',
+              label: t('Point Radius Units'),
+              description: t(
+                'The unit for point radius. Use "pixels" for consistent ' +
+                  'screen-space sizing regardless of zoom level.',
+              ),
+              default: 'pixels',
```
rusackas, line 400: This is now defaulted to meters, i think

rusackas: Aha! @sadpandajoe we were thinking of the Path one, which is different.
https://github.com/apache/superset/pull/37248

In this case, we DO want pixels, so markers stay the same size when you zoom the map.

superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Geojson/Geojson.tsx
```diff
@@ -321,7 +321,9 @@ export const getLayer: GetLayerType<GeoJsonLayer> = function ({
       getFillColor(feature, filterState?.value),
     getLineColor,
     getLineWidth: fd.line_width || 1,
-    pointRadiusScale: fd.point_radius_scale,
+    getPointRadius: fd.point_radius ?? 10,
+    pointRadiusUnits: fd.point_radius_units ?? 'pixels',
```
sadpandajoe: **Backward-compatibility break for existing charts**

These fallback defaults differ from deck.gl's built-in defaults, which changes rendering for every existing GeoJSON chart that never stored these new fields:

| Property | deck.gl default | Fallback here | Impact |
|----------|----------------|---------------|--------|
| `getPointRadius` | `1` | `10` | 10x larger base radius |
| `pointRadiusUnits` | `'meters'` | `'pixels'` | Completely different sizing model |

**Scenario**: An existing chart saved `point_radius_scale = 200` (from the old `[0, 100, 200, 300, 500]` choices) but has no `point_radius` or `point_radius_units` (since those controls didn't exist yet).
- **Before this PR**: `getPointRadius=1 (deck.gl default), units='meters', scale=200` → 200-meter circles
- **After this PR**: `getPointRadius=10, units='pixels', scale=200` → 2000-pixel circles

**Suggested fix** — use deck.gl's defaults as the fallbacks here, and let the *control panel* defaults handle the improved UX for *new* charts:

```typescript
getPointRadius: fd.point_radius ?? 1,
pointRadiusUnits: fd.point_radius_units ?? 'meters',
pointRadiusScale: fd.point_radius_scale ?? 1,
```

New charts will still get `point_radius=10, units='pixels', scale=1` from the control panel defaults. Existing charts without these fields get safe deck.gl defaults.

**Suggested test** to guard against this regression:

```typescript
describe('getLayer backward compatibility', () => {
  it('preserves rendering for existing charts without new point radius fields', () => {
    // Simulate form data from an existing chart that only has point_radius_scale
    const legacyFormData = {
      ...baseFormData,
      point_radius_scale: 200,
      // point_radius and point_radius_units intentionally absent
    };

    const layer = getLayer({ formData: legacyFormData, ...otherArgs });
    const props = layer.props;

    // Should match deck.gl defaults, NOT the new control panel defaults
    expect(props.getPointRadius).toBe(1);           // deck.gl default, not 10
    expect(props.pointRadiusUnits).toBe('meters');   // deck.gl default, not 'pixels'
    expect(props.pointRadiusScale).toBe(200);        // user's saved value preserved
  });

  it('uses control panel defaults for new charts', () => {
    const newChartFormData = {
      ...baseFormData,
      point_radius: 10,
      point_radius_units: 'pixels',
      point_radius_scale: 1,
    };

    const layer = getLayer({ formData: newChartFormData, ...otherArgs });
    const props = layer.props;

    expect(props.getPointRadius).toBe(10);
    expect(props.pointRadiusUnits).toBe('pixels');
    expect(props.pointRadiusScale).toBe(1);
  });
});
```

sadpandajoe (commented):
Overall the feature looks great — clean controls, good descriptions, and solid mapping to deck.gl's API. One backward-compatibility concern that should be addressed before merge.

superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Geojson/Geojson.tsx
rusackas: Great catch! You're absolutely right about the backward compatibility issue.

Fixed in 7705ffcfe0 - the runtime fallbacks now use deck.gl's defaults (`getPointRadius=1`, `pointRadiusUnits='meters'`) instead of the control panel defaults. This way:

- **Existing charts** (without new fields) → get deck.gl defaults → rendering unchanged  
- **New charts** → get control panel defaults (`point_radius=10`, `units='pixels'`) → improved UX

Added a comment in the code explaining the reasoning.

Regarding the suggested test - that's a good idea for ensuring this doesn't regress. I'll leave that as a follow-up since this PR doesn't currently have test coverage for the layer props, but it's worth adding.

rusackas: Tests added.

rusackas:
@sadpandajoe Thanks for the detailed backward compatibility analysis and test suggestions! The tests you proposed have been implemented in commit d43d40da.

The tests verify:
1. **Legacy charts** (with only `point_radius_scale` set) receive deck.gl defaults (`getPointRadius=1`, `pointRadiusUnits='meters'`) - preserving existing rendering
2. **New charts** (with all three fields) correctly use the control panel values

Also pushed a fix for the pre-commit failure (ee1a8ffa) - the `react/jsx-no-duplicate-props` rule in oxlint doesn't support configuration options, so removed the `ignoreCase` parameter.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rdyatlov:
Hi @rusackas and All!

Thank you very much for the new amendments!

I've also checked: everything works fine! Just for your information, there is a problem with a statical description to a point on the map (from properties list in my json): I can't move it below/above the point and Cyrillic text is not visible.
But maybe, I'd better open a new issue for that.

Could you please let me know, by when are you going to merge the new changes to a Master branch?

Thank you!

Best regards,
Roman

enlohhy:
Hi! I’d like to help take this over and get it merged. I’ve reviewed the current state, the backward-compatibility follow-up, and the latest checks. I’m going to sync it against current master and see what, if anything, is still needed to move it forward.

rusackas:
@rdyatlov glad it's working for you! The label positioning + Cyrillic rendering sounds unrelated to this PR, so yeah, a separate issue would be the way to go. No firm merge date yet, still need a green CI and another set of eyes on it, but I'll keep it moving.

rusackas:
@enlohhy appreciate the offer! This one's still mine and pretty much up to date with master, so no takeover needed. It's basically just waiting on a review/approval at this point... if you want to help, a review would be hugely welcome :)

rusackas:
Added controlPanel coverage for the three new controls (point_radius, point_radius_scale, point_radius_units) - defaults, validators, choices - so the new `controlPanel.ts` lines are exercised. That's the `core-packages-ts` project-coverage drop codecov was flagging. Running CI 🤞

superset-frontend/plugins/preset-chart-deckgl/src/layers/Geojson/controlPanel.ts
```diff
@@ -352,15 +354,56 @@ const config: ControlPanelConfig = {
...
```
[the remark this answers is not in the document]
rusackas: Good catch, and you're right that the runtime `?? 1` fallback in `Geojson.tsx` never fires on the dashboard path because `applyDefaultFormData` fills the control default first. The tradeoff is deliberate though: defaulting new charts to pixels/10 was the whole point here, and there's no reliable signal in `applyDefaultFormData` to tell a brand-new chart from a legacy one missing the field. I'd rather take the slight visual shift on old GeoJSON charts than ship undefined-by-default controls. Open to a migration if folks feel strongly.

superset-frontend/plugins/preset-chart-deckgl/src/layers/Geojson/Geojson.tsx
```diff
@@ -328,7 +328,11 @@ export const getLayer: GetLayerType<GeoJsonLayer> = function ({
       getFillColor(feature, filterState?.value),
     getLineColor,
     getLineWidth: fd.line_width || 1,
-    pointRadiusScale: fd.point_radius_scale,
+    // Use deck.gl defaults as fallbacks for backward compatibility with existing charts.
+    // New charts will get control panel defaults (point_radius=10, units='pixels', scale=1).
+    getPointRadius: fd.point_radius ?? 1,
+    pointRadiusUnits: fd.point_radius_units ?? 'meters',
+    pointRadiusScale: fd.point_radius_scale ?? 1,
```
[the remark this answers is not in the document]
rusackas: Good catch, free-form input can definitely sneak a string through. Wrapped both in a NaN-safe `Number()` coercion (with the deck.gl defaults as fallback) and added a test for the string case.

rusackas:
Not sure where all my reviewers are... but adding a new one :D

superset-frontend/plugins/preset-chart-deckgl/src/layers/Geojson/Geojson.tsx
```diff
@@ -328,7 +335,11 @@ export const getLayer: GetLayerType<GeoJsonLayer> = function ({
       getFillColor(feature, filterState?.value),
     getLineColor,
     getLineWidth: fd.line_width || 1,
-    pointRadiusScale: fd.point_radius_scale,
+    // Use deck.gl defaults as fallbacks for backward compatibility with existing charts.
+    // New charts will get control panel defaults (point_radius=10, units='pixels', scale=1).
+    getPointRadius: toNumber(fd.point_radius, 1),
+    pointRadiusUnits: fd.point_radius_units ?? 'meters',
```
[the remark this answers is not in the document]
rusackas: The `meters` fallback is deliberate, not a slip. Existing charts have no `point_radius_units`, so this branch keeps them on deck.gl's old `meters` behavior; new charts pick up `pixels` from the control panel default instead. Flipping the fallback to `pixels` would silently resize every legacy GeoJSON chart, which is exactly what we fixed earlier in this PR.

aminghadersohi (commented):
## Review — `feat(deckgl): add point radius controls for GeoJSON layer`

Reviewed at HEAD `792295676ebddd42dfc73ca369123c079e6a36ac`.

The overall approach is sound — three coordinated controls with explicit backward-compat fallbacks and string-coercion for free-form `SelectControl` values. Test coverage for the intended cases (new charts, string coercion, absent fields) is solid. One logic error needs fixing before this merges.

---

### HIGH — `toNumber(null, 1)` returns `0`, not `1`

**`Geojson.tsx:270–272` (definition) · `Geojson.tsx:342` (exercised)**

```ts
const toNumber = (value: unknown, fallback: number) => {
  const num = Number(value);
  return Number.isFinite(num) ? num : fallback;
};
```

`Number(null) === 0` and `Number.isFinite(0) === true`, so the fallback is never reached when `value` is `null`. The previous `point_radius_scale` control had `default: null`, so any chart where the user never explicitly set that field has `null` persisted in its saved `form_data`. Those charts now receive `pointRadiusScale: 0` — zeroing the rendered radius of every circle point. If deck.gl previously resolved `pointRadiusScale: null` to its documented default of `1` (e.g. via internal nullish coalescing), this is a silent regression that makes all GeoJSON circle features invisible in existing charts.

**Fix** — guard `null`/`undefined` before calling `Number`:

```ts
const toNumber = (value: unknown, fallback: number) => {
  const num = Number(value ?? fallback);
  return Number.isFinite(num) ? num : fallback;
};
```

`null ?? 1 → 1`, so `toNumber(null, 1) → 1`. The `undefined` case already works. An intentional `0` (which was one of the old preset choices) still returns `0`. ✓

A test that exercises `point_radius_scale: null` should accompany the fix — the existing backward-compat test at `Geojson.test.tsx:321` uses `200`, so it does not catch this case.

---

### MEDIUM — control-panel default vs. runtime fallback asymmetry for `point_radius_units`

**`controlPanel.ts:392` · `Geojson.tsx:341`**

```ts
// controlPanel.ts:392 — shown in the explore UI
default: 'pixels',

// Geojson.tsx:341 — used when field is absent from saved form_data
pointRadiusUnits: fd.point_radius_units ?? 'meters',
```

Old charts rendered in a **dashboard** (raw saved params, no control defaults applied) get `'meters'`; the same chart opened in **explore** gets `'pixels'` from the control default — two different visual outputs from the same saved chart. The inline comment at line 338 explains the deck.gl-default strategy but doesn't address this split-context inconsistency. Either align the control default to `'meters'` so both paths agree, or note the asymmetry explicitly in the comment.

---

### NIT — `validateInteger` vs. `legacyValidateInteger` in the same file

**`controlPanel.ts:367` vs. `:177`, `:290`**

`label_size` and `icon_size` (same `SelectControl freeForm: true` pattern) use `legacyValidateInteger`; the new `point_radius` uses `validateInteger`. If the intent is to modernise, the two existing controls should be updated to match; otherwise `validateInteger` should be `legacyValidateInteger`.

---

### NIT — `toNumber` defined inside `getLayer`

**`Geojson.tsx:270`**

Pure function with no closure dependencies; move it to module scope to avoid reallocating it on every `getLayer` call.

aminghadersohi (approved):
approving with a review comment

rusackas:
@aminghadersohi great catch on the `null` case, that one would have quietly zeroed out legacy charts. Fixed in 7e7f38df92 by coercing through the fallback first (`Number(value ?? fallback)`), so `null`/`undefined` land on the default while an explicit `0` is preserved, plus a test for it. Also hoisted `toNumber` to module scope.

On the units asymmetry: that split is deliberate, same reasoning as the backward-compat thread above. Old charts have no `point_radius_units` so both paths should resolve to `meters`... the dashboard path does; the explore default pulling `pixels` is the one inconsistency. I went back and forth and would rather new charts default to pixels than make legacy charts in explore drift, but I take your point that it is two outputs from one saved chart. Happy to align the control default to `meters` if you feel strongly.

Left the `validateInteger` vs `legacyValidateInteger` nit alone for now, since modernizing the two neighboring controls feels like its own change.

PR #33250, opened 2025-04-25
fix(DB update): Gracefully handle querry error during DB update

Vitor-Avila:
### SUMMARY
Some DB engines require running a SQL query to get the default catalog. This wouldn't work in case the connection is broken (for example, expired credentials). This PR gracefully handles this scenario to unblock the connection update.

No UI changes.

### TESTING INSTRUCTIONS
Added test.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/update.py
+++ b/superset/commands/database/update.py
@@ -80,9 +80,19 @@ class UpdateDatabaseCommand(BaseCommand):
             # existing personal tokens.
             self._handle_oauth2()
 
-        # build new DB
+        # Some DBs require running a query to get the default catalog.
+        # In these cases, if the current connection is broken then
+        # `get_default_catalog` would raise an exception. We need to
+        # gracefully handle that so that the connection can be fixed.
         original_database_name = self._model.database_name
-        original_catalog = self._model.get_default_catalog()
+        force_update: bool = False
+        try:
+            original_catalog = self._model.get_default_catalog()
+        except Exception:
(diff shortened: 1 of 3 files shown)


superset/commands/database/update.py
```diff
@@ -80,9 +80,19 @@ def run(self) -> Model:
...
+        # Some DBs require running a query to get the default catalog.
+        # In these cases, if the current connection is broken then
+        # `get_default_catalog` would raise an exception. We need to
+        # gracefully handle that so that the connection can be fixed.
         original_database_name = self._model.database_name
-        original_catalog = self._model.get_default_catalog()
+        force_update: bool = False
+        try:
+            original_catalog = self._model.get_default_catalog()
+        except Exception:
```
geido, line 91: Do we want to log a warning here? Would that be useful in our logs?

geido (approved):

superset/commands/database/update.py
Vitor-Avila: I think it's not needed, as this failure would be expected in case the "current" credentials are not working properly, and the new credentials should have been validated already through the `/validate_parameters` check.

For safety, we're then later updating the catalog on existing assets in case the connection does not allow multi-catalog, so I believe we're good.

Should be a quick change to add it if you think it's best, tho. :)

PR #33255, opened 2025-04-28
feat(Login): Migrate login page to frontend

msyavuz:
### SUMMARY

This PR tries to move the Login fab view to the frontend spa.

AuthDB Login:

Before:
image

After:
image

### TESTING INSTRUCTIONS

Follow the instructions on setting up auth methods in flask-appbuilder to see different auth methods working. The configuration should be something like this for different auth methods:

```
AUTH_TYPE = AUTH_OAUTH
OAUTH_PROVIDERS = [
    {
        "name": "google",
        "icon": "fa-google",
        "token_key": "access_token",
        "remote_app": {
            "client_id": GOOGLE_KEY,
            "client_secret": GOOGLE_SECRET,
            "api_base_url": "https://www.googleapis.com/oauth2/v2/",
            "client_kwargs": {"scope": "email profile"},
            "request_token_url": None,
            "access_token_url": "https://accounts.google.com/o/oauth2/token",
            "authorize_url": "https://accounts.google.com/o/oauth2/auth",
            "project_id": "symbolic-path-413013",
            "auth_uri": "https://accounts.google.com/o/oauth2/auth",
            "token_uri": "https://oauth2.googleapis.com/token",
            "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
            "redirect_uris": ["http://localhost:8088/oauth-authorized/google"],
            "javascript_origins": ["http://localhost:8088"],
            "jwks_uri": "https://www.googleapis.com/oauth2/v3/certs",
        },
    },
]

AUTH_TYPE = AUTH_OID
OPENID_PROVIDERS = [
    {"name": "Yahoo", "url": "https://me.yahoo.com"},
    {"name": "AOL", "url": "http://openid.aol.com/<username>"},
    {"name": "Flickr", "url": "http://www.flickr.com/<username>"},
    {"name": "MyOpenID", "url": "https://www.myopenid.com"},
]

AUTH_TYPE = AUTH_LDAP
AUTH_LDAP_SERVER = "ldap://ldap.example.org"
AUTH_LDAP_USE_TLS = False

# registration configs
AUTH_USER_REGISTRATION = True  # allow users who are not already in the FAB DB
AUTH_USER_REGISTRATION_ROLE = (
    "Public"  # this role will be given in addition to any AUTH_ROLES_MAPPING
)
AUTH_LDAP_FIRSTNAME_FIELD = "givenName"
AUTH_LDAP_LASTNAME_FIELD = "sn"
AUTH_LDAP_EMAIL_FIELD = (
    "mail"  # if null in LDAP, email is set to: "{username}@email.notfound"
)

# search configs
AUTH_LDAP_SEARCH = "ou=users,dc=example,dc=org"  # the LDAP search base
AUTH_LDAP_UID_FIELD = "uid"  # the username field
AUTH_LDAP_BIND_USER = (
    "uid=admin,dc=example,dc=org"  # the special bind username for search
)
AUTH_LDAP_BIND_PASSWORD = "admin_password"  # the special bind password for search
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/pages/Login/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { css } from '@emotion/react';
```
EnxDev, line 20: should be better to import this from `@superset-ui/core`

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { SupersetClient, styled, t } from '@superset-ui/core';
+import { Button, Card, Flex, Form, Input } from 'src/components';
```
EnxDev, line 22:
```suggestion
import { Button, Card, Flex, Form, Input, Typography  } from 'src/components';
```

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { Icons } from 'src/components/Icons';
+import Typography from 'src/components/Typography';
```
EnxDev, line 24: should be imported from `src/components`

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+              onFinish={onFinish}
+            >
+              <Form.Item<LoginType>
+                label={<StyledLabel>{t('Username:')}</StyledLabel>}
+                name="username"
+                rules={[
+                  { required: true, message: t('Please enter your username') },
+                ]}
+              >
+                <Input prefix={<Icons.UserOutlined size={1} />} />
```
EnxDev, line 139: Should it be one of these: 'xs', 's', 'm', 'l', 'xl', 'xxl'?

EnxDev (commented):
Thanks!

msyavuz:
> Yay!!!! Is this the magic moment we've all been waiting for that kills the little gray close buttons that show up before the React frontend is built/loaded?

Well, i think those will show up anyways until the frontend is loaded. Sorry!

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+        {authType === AuthType.AuthOID && (
+          <Flex justify="center" vertical gap="middle">
+            <Form layout="vertical" requiredMark="optional" form={form}>
+              {providers.map((provider: OIDProvider) => (
+                <Form.Item<LoginForm>>
+                  <Button
+                    href={`/login/${provider.name}`}
+                    block
+                    iconPosition="start"
+                    icon={AuthIconMap[provider.name]}
```
geido, line 115: Can new providers be introduced? What happens if the icon isn't available?

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+                  { required: true, message: t('Please enter your password') },
+                ]}
+              >
+                <Input.Password
+                  prefix={<Icons.KeyOutlined iconSize="l" />}
+                  data-test="password-input"
+                />
+              </Form.Item>
+              <Form.Item label={null}>
+                <Flex
```
geido, line 179: You have a `LoginContainer` that is the same as this. I prefer inline, so maybe we can get rid of `LoginContainer`?

mistercrunch:
YES! should the login box be `vertical-algin: center;`?

superset-frontend/src/pages/Login/index.tsx
geido: It seems that there is a pattern in naming convention for social icons on Ant Design. We can make a util function that gets the provider name, capitalizes it and adds either Outlined or Filled (to check with Kasia) and if it does not exist can fallback to a button or a fallback icon.

mistercrunch:
While at it, wondering if we can reduce the width of the login card, and maybe remove the icon at the top-right, the word "Log in" speaks for itself, no need for an icon here. @kasiazjc may have some input here too.

alexandrusoare:
Great work @msyavuz ! The page is looking good, I just have a few nits here and there.

- When you hover over the login and register buttons there's a slight difference, the register text gets underlined, maybe we can get rid of that: 
image
image

- I think we should either reduce a bit the height of the login card or to give a little bit more padding/margin to buttons on the top so that we can have some better proportions, let me know what you think: 
image

- Maybe center the inputs and labels or at least place them on the left
image

superset/views/alerts.py
```diff
@@ -43,6 +44,8 @@ def list(self) -> FlaskResponse:
     @has_access
     @permission_name("read")
     def log(self, pk: int) -> FlaskResponse:  # pylint: disable=unused-argument
+        from superset import is_feature_enabled
```
dpgaspar, line 47: nit: is it possible to move this import to the module level?

dpgaspar (commented):
Tested AUTH_DB and AUTH_OAUTH, 
using OAUTH and `AUTH_USER_REGISTRATION = True` login breaks, it seems `RECAPTCHA_PUBLIC_KEY` is now required, although this is a good measure it was not a required config before.
This does not seem specifically related with this PR, but a possible issue

msyavuz:
> using OAUTH and AUTH_USER_REGISTRATION = True login breaks, it seems RECAPTCHA_PUBLIC_KEY is now required, although this is a good measure it was not a required config before.


@dpgaspar I think it was required when registration is enabled but it was implicit, Recaptcha would be broken and you couldn't register. I am not sure what was the intended flow with fab as https://flask-appbuilder.readthedocs.io/en/latest/user_registration.html doesn't mention oauth.

alexandrusoare (approved):

PR #32112, opened 2025-02-02
refactor(Icons): Replaces custom icons with Ant Design 5 icons

EnxDev:
### SUMMARY

This PR improves the overall standardization and maintainability of the project by replacing custom icons with Ant Design 5 icons. 
It removes nearly all FontAwesome icons, streamlining the icon set to ensure consistency. 
The Icon component has been enhanced to support new features and improved performance. 

Additionally, missing icons have been added, and standardized sizing has been applied across various icons, including those used in buttons, to ensure a cohesive UI. 

Finally, it eliminates direct imports of icons from Ant Design, simplifying the import structure and improving code readability.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

- Before
Screenshot 2025-03-07 095627

- After
Screenshot 2025-03-07 095923

Tags :

- before
before
- after
after

Filter: 

- before
before
- after
after
- before
before-h
- after
after-h
- before
before-filter
- after
after-filter
- before
before-panel
- after
after-panel

Tags: 
- before
before
- after
after

Table: 
before
after
before-2
after-2
before
after
before
after

Trash:
before
after
before
after
before
after
before
after
before
after
before
after
before
after
before
after

Link: 

- before
before
- after
after
### TESTING INSTRUCTIONS

1. CI should pass
2. All custom icons replaced with Ant Design 5 icons, should be consistent in color, size, and relevance.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #33256, opened 2025-04-28
refactor(InfoTooltipWithTrigger): Replace support for fa icons with antd5 icons

EnxDev:
### ✨ What’s changed

- Defined and centralized tooltip icon variants `(info, warning, notice, error, question)` in a structured way to ensure consistent logic and styling across the codebase ♻️
- Replaced old FontAwesome (fa) icons with modern Ant Design 5 (antd5) icons inside InfoTooltipWithTrigger 🔥
- Updated icon styling to use inline styles (color, fontSize) based on Ant Design theme tokens 🎨
- Removed unused bsStyle prop (no longer needed—relying on theme tokens now) 🧹
- Updated tests to reflect the new icons and variant behavior 🔍

### 🛠 Why

- To modernize the component and eliminate deprecated FontAwesome usage
- To align with the latest design system based on AntD 5 icons (This can be considered a continuation of [#32112](https://github.com/apache/superset/pull/32112))
- To improve consistency across the codebase by defining tooltip variants in a single place
- To ensure better maintainability and compatibility with future design and theming updates

- Before
option-control-before
- After
option-control-after

https://github.com/user-attachments/assets/921b09fc-a88d-48e0-a866-30a1586f7800

- Error variant
variant-error

- Info variant
variant-info

- Notice variant
variant-notice

- Question variant
variant-question

- Warning variant
variant-warning

### TESTING INSTRUCTIONS

- All tests should pass
- No visual regressions expected — icons still match old behavior but now using AntD 5.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
```diff
@@ -16,56 +16,60 @@
...
-  icon = 'info-circle',
+  icon = <ExclamationCircleOutlined />,
   className = 'text-muted',
   placement = 'right',
-  iconsStyle = {},
-}: InfoTooltipWithTriggerProps) {
-  const iconClass = `fa fa-${icon} ${className} ${
-    bsStyle ? `text-${bsStyle}` : ''
-  }`;
+  iconsStyle = { fontSize: 12 },
```
geido, line 42: I don't think we need this prop at all. The `fontSize` appear to be always the same? Also, if we really need it, it would make more sense to apply the styles to the `icon` and not indirectly through the props.

superset-frontend/src/explore/components/controls/AnnotationLayerControl/index.tsx
```diff
@@ -193,7 +193,7 @@ class AnnotationLayerControl extends PureComponent<Props, PopoverState> {
       return (
         <InfoTooltipWithTrigger
           label="validation-errors"
-          bsStyle="danger"
+          iconsStyle={{ color: theme.colorError }}
```
geido, line 196: I think the previous approach of standardizing the component via "danger", "warning", etc. was better than this. We should rethink the approach here. Just standardize the possible statuses and use the appropriate Icon directly within the component, otherwise this just becomes a wrapper for icons and it does not make a lot of sense.

superset-frontend/src/explore/components/controls/DndColumnSelectControl/Option.tsx
```diff
@@ -72,9 +72,13 @@ export default function Option({
       <Label data-test="control-label">{children}</Label>
       {(!!datasourceWarningMessage || isExtra) && (
         <StyledInfoTooltipWithTrigger
-          icon="exclamation-triangle"
+          icon={
```
geido, line 75: Same comment as above. Let's standardize with "danger", "warning" etc. If we do not need to customize the icon, then we can remove the prop entirely and just standardize the component.

EnxDev: I can standardize the color and font-size: for example, I could have three types:

- default with grayscale color and 12 px
- error with error color and 12 px
- warning with warning color and 12 px

We would need to know which icon you want, right? 

For example, I could have ExclamationCircleOutlined in yellow or grayscale, or I could have 
InfoCircleOutlined in yellow or red.

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
EnxDev: I would like to allow flexibility in choosing the icon size.
If I include the 3 icons that can be used inside the InfoTooltipWithTrigger component, I will allow the user to choose the icon (from the three available) through props, as well as the size, which can be expressed in px or sizes like s, m, l, etc. 
At that point, I will directly apply the style inside the component to adjust the size.

superset-frontend/src/explore/components/controls/AnnotationLayerControl/index.tsx
EnxDev: I agree, I will provide three types:

1. base -> with grayscale color
2. warning  -> with warning color
3. error -> with error color

I will allow the selection from three icons:

1. info circle
2. exclamation circle
3. exclamation triangle

I will add sizes for the icon sizes:

'xs' | 's' | 'm' | 'l' | 'xl' | 'xxl'

Any other suggestions?

geido: Why do you need to give the ability to choose the icon? I think `warning ` will always be `ExclamationCircle`, "info" will always be `InfoCircle`, etc.?

superset-frontend/src/explore/components/controls/DndColumnSelectControl/Option.tsx
geido: I think colors should always be the same based on the type "info", "warning", "error". Font sizes might be different if you think you want to give that option and we can standardize them too with 'xs' | 's' | 'm' | 'l' | 'xl' | 'xxl' as you suggested above.

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
geido: See my comments below

superset-frontend/src/explore/components/controls/AnnotationLayerControl/index.tsx
EnxDev: Because it seems to me that I saw, for example, a yellow exclamation circle and a yellow exclamation triangle

EnxDev: According to the philosophy that was there before, with the bsStyle prop you could decide the color, which was a CSS class, and with the icon prop you chose the icon you wanted to use.

The component is also designed to work without the icon.

I had also thought about using only the AntD component, because this one seemed like just a wrapper, but I didn’t want to do too much rework.

geido: > yellow exclamation circle and a yellow exclamation triangle
Mhhh this seems like it should not be the case. Maybe it was wrong implementation decision. I'll leave to you to decide the level of standardization based on what we have in the codebase / what we can reasonably refactor. Thanks

EnxDev: Ok, I'll check if there is a standard and whether it has been applied.
If it hasn’t been applied, I will apply it; otherwise, I will keep the previous logic.

The main goal of this PR is to remove FontAwesome icons in favor of AntD icons, remove Bootstrap class and custom CSS classes, remove deprecated props, and update the RTL tests. Thanks!

EnxDev: I have verified that, in this case, there wasn’t a clear standard being followed, so @kasiazjc and I defined one together (see the PR description) and I updated the purpose of this PR accordingly.

In SqlLab, `InfoTooltipWithTrigger` was being used from the `@superset-ui/chart-controls` package in two instances. I replaced it with the `IconTooltip` component.

I applied the defined standard to the pages where the `InfoTooltipWithTrigger` component was being used.

For now, I’ve decided to keep the `InfoTooltipWithTrigger` component in the `@superset-ui/chart-controls` package to avoid an additional refactor.

We can revisit this later, perhaps when the components are fully moved into the shared package.

superset-frontend/src/explore/components/controls/DndColumnSelectControl/Option.tsx
EnxDev: continues [here](https://github.com/apache/superset/pull/33256#discussion_r2065981930)

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
EnxDev: [continues here](https://github.com/apache/superset/pull/33256#discussion_r2065981930)

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
```diff
@@ -16,65 +16,101 @@
...
   );
+
   if (!tooltip) {
     return iconEl;
   }
+
   return (
     <Tooltip
-      id={`${kebabCase(label)}-tooltip`}
+      id={`${kebabCase(label) || Math.floor(Math.random() * 10000)}-tooltip`}
```
msyavuz, line 109: I don't think we should do this. If it is necessary to default to random id, there should be a useId hook for use.

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
```diff
@@ -16,65 +16,101 @@
...
+      error: { color: theme.colorError, icon: <CloseCircleOutlined /> },
+    }),
+    [theme],
+  );
+
+  const variant = type ? infoTooltipWithTriggerVariants[type] : null;
+
+  const iconCss = css`
+    color: ${variant?.color ?? theme.colorIcon};
+    font-size: ${themeObject.getFontSize(iconSize)};
```
msyavuz, line 70: Can't we do theme.fontSize instead of importing themeObject directly?

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
```diff
@@ -16,65 +16,101 @@
...
-      onKeyPress={
-        onClick &&
-        (event => {
-          if (event.key === 'Enter' || event.key === ' ') {
-            onClick();
-          }
-        })
-      }
-    />
+      onKeyDown={handleKeyDown}
```
msyavuz, line 97: Can't we just use a button instead?

EnxDev: I’ve kept the old implementation for now. We can consider replacing it with Ant Design’s <Button> component if achieving the same result doesn’t require too many custom styles. I’d avoid using the native <button>, as I’m currently removing all of them from the codebase. thanks!

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
EnxDev: Is theme.fontSize equal to 14px, right?

superset-frontend/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
EnxDev: I'll keep it in mind, thanks!

msyavuz: I think we should still look to avoid doing this. Is this necessary?

EnxDev: Do you mean that you'd like to remove the id prop?

EnxDev: If you meant to remove the id prop, I noticed it's a common practice throughout the codebase, so I’d suggest keeping it for now.

kasiazjc (approved):
lgtm

PR #33269, opened 2025-04-28
fix(echarts): rename time series shifted colnames

justinpark:
### SUMMARY
With the migration to eChart, the logic for colnames in the legacy time-series has changed. When applying a time shift, the metric name is displayed along with the formula, making it difficult to review the column items.

|legacy chart|echart|
|--|--|
|Screenshot_2025-04-28_at_4_21_16 PM|Screenshot 2025-04-28 at 4 25 22 PM|

This commit expands the existing rename post-processing logic applied to actual values, enabling it to be similarly applied when multiple metrics are selected and when the comparison type is not actual values.

||Before|After|legacy chart|
|--|--|--|--|
|1 metric|Screenshot 2025-04-28 at 4 25 22 PM|Screenshot 2025-04-28 at 4 23 11 PM|Screenshot_2025-04-28_at_4_54_12 PM|
|multi metrics|Screenshot 2025-04-28 at 4 25 02 PM|Screenshot 2025-04-28 at 4 23 31 PM|Screenshot_2025-04-28_at_4_21_51 PM|

### TESTING INSTRUCTIONS
Create a bar chart and apply time shift
Select any value other than "actual values" in calculation type

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/test/operators/renameOperator.test.ts
+++ b/superset-frontend/packages/superset-ui-chart-controls/test/operators/renameOperator.test.ts
@@ -186,7 +186,7 @@ test('should add renameOperator if exist "actual value" time comparison', () => 
   });
 });
 
-test('should add renameOperator if exist derived time comparison', () => {
+test('should add renameOperator if derived time comparison exists', () => {
   expect(
     renameOperator(
       {
@@ -211,7 +211,7 @@ test('should add renameOperator if exist derived time comparison', () => {
   });
 });
 
-test('should add renameOperator if exists multiple metrics', () => {
+test('should add renameOperator if multiple metrics exist', () => {
   expect(
     renameOperator(
       {
--- a/superset-frontend/packages/superset-ui-chart-controls/test/operators/renameOperator.test.ts
+++ b/superset-frontend/packages/superset-ui-chart-controls/test/operators/renameOperator.test.ts
@@ -109,7 +109,7 @@ test('should add renameOperator if exists derived metrics', () => {
     ).toEqual({
       operation: 'rename',
       options: {
-        columns: { 'difference__count(*)__count(*)__1 year ago': '1 year ago' },
+        columns: { [`${type}__count(*)__count(*)__1 year ago`]: '1 year ago' },
         inplace: true,
         level: 0,
       },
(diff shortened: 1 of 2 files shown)


michael-s-molina:
Thanks for working on this @justinpark! Would be possible to always keep the `dimension, metric, offset` order? Example:

```
count
count, 1 week ago

airbnb
airbnb, 1 week ago

airbnb
airbnb, count, 1 week ago
airbnb, sum, 1 week ago
```

justinpark:
> Thanks for working on this @justinpark! Would be possible to always keep the `dimension, metric, offset` order? Example:
> 
> ```
> count
> count, 1 week ago
> 
> airbnb
> airbnb, 1 week ago
> 
> airbnb
> airbnb, count, 1 week ago
> airbnb, sum, 1 week ago
> ```

To change the order in the same manner as above, we need to reverse the following flatten logic(level1, level2 vs level2, level1), which seems likely to affect the overall results derived using pandas. I'm considering applying this during the next breaking window period (so we should apply the change in the next version)

https://github.com/apache/superset/blob/13f77a74166c25a20ad2c25bf58db6a6b7cafae2/superset/utils/pandas_postprocessing/flatten.py#L78-L91

michael-s-molina:
> To change the order in the same manner as above, we need to reverse the following flatten logic(level1, level2 vs level2, level1)

Ah, makes sense! Thanks for referencing the code.

michael-s-molina (approved):
LGTM. @villebro @betodealmeida @sadpandajoe it would be good to have a second pair of 👀 given the possible impact of the changes.

superset-frontend/packages/superset-ui-chart-controls/test/operators/renameOperator.test.ts
```diff
@@ -170,6 +186,61 @@ test('should add renameOperator if exist "actual value" time comparison', () =>
...
+        'ratio__count(*)__count(*)__1 year ago': '1 year ago',
+        'ratio__count(*)__count(*)__1 year later': '1 year later',
+      },
+      inplace: true,
+      level: 0,
+    },
+  });
+});
+
+test('should add renameOperator if exists multiple metrics', () => {
```
rusackas, line 214:
```suggestion
test('should add renameOperator if multiple metrics exist', () => {
```

rusackas (approved):
Teensy grammare nit suggestions, but in general LGTM!

PR #33271, opened 2025-04-29
fix: Exclude Filter Values

amaannawab923:
### SUMMARY
This PR addresses four primary bugs:
1 is Introduced due to pr  https://github.com/apache/superset/pull/33054  , 1 is a ui bug & 2 of them are legacy bugs which exist on master

1. Incorrect Data Mask Dispatch on Initial Render
A bug was introduced in [PR #33054](https://github.com/apache/superset/pull/33054), where the data mask was being updated on the initial render—regardless of whether the dropdown value had changed. This is incorrect behavior.
On initial render, when filterState.value is undefined, an invalid ad-hoc filter is attached, causing the chart to incorrectly display "No Data Found" even when data exists:
Before: 

Screenshot 2025-05-01 at 3 37 11 PM

Solution
The fix involves tracking the "Exclude Filter Values" checkbox state using a ref. We now only dispatch dispatchDataMask when the checkbox is toggled, preventing unnecessary updates during initial render.

Screenshot 2025-05-01 at 3 40 18 PM

2. Select All & Deselect All of Select getting cutoff 
Since Select All & Deselect All was introduced post Exclude Filter Values , and the is & is not dropdown shrinks down the Native filter , the ui of Select All & Deselect All started getting cutoff 

Before: 

Screenshot 2025-05-01 at 5 25 00 PM

The Solution: 
The options Select All & Deselect All are arranged in a column format whenever there is less space

After: 

Screenshot 2025-05-01 at 5 22 36 PM
Screenshot 2025-05-01 at 5 22 57 PM

3 .Incorrect Overwrite of dataMask When Default Filter Value is Set
When a native filter’s configuration is updated, the action SET_DATA_MASK_FOR_FILTER_CHANGES_COMPLETE is triggered. This action replaces the corresponding filter object in the Redux store with the new dataMask object in the reducer. While this is generally fine, it leads to issues when the filter is configured to "Select first value by default."
In such cases, the action replaces the filter’s filterState with an empty object, effectively breaking the filter and causing a crash.

  Before Fix:
  
   
  [screen-capture (4).webm](https://github.com/user-attachments/assets/0cb23376-c9fd-4e4c-b832-3e534fc58cf7)

Solution:
If the filter configuration change does not alter the target column and the “Select first value by default” option remains enabled (e.g., changes like enabling "Allow creation of new values" or "Inverse Selection"), we now preserve the existing filterState and extraFormData.
This ensures the native filter remains stable and doesn’t crash unnecessarily.

After Fix:

[screen-capture (6).webm](https://github.com/user-attachments/assets/3d9d04b3-619b-48c8-ad8c-5d9c74b49a69)

4 . Native Filter Freeze Due to Infinite Loop
This issue, likely a legacy bug, was caused by an infinite loop triggered within a useEffect hook. The loop stemmed from the way the local useImmerReducer was initialized with a default value. Specifically, the hook used filterState.value as both a dependency and a trigger to update the dataMask.

The problem:

Updating the dataMask relied on filterState.value.

That update then triggered a change in filterState.value.

Since filterState.value was a dependency of the useEffect, it caused the effect to run again—creating a continuous loop.

Before  Fix: 

[screen-capture (8).webm](https://github.com/user-attachments/assets/a65bd92d-611a-4e45-a8a3-7bc0d77ffbd8)

Solution:
To avoid this loop, we now rely on formData?.defaultValue for the initial render and value setup instead of filterState.value. This breaks the cycle and results in a stable, predictable behavior.

After Fix:

[screen-capture (10).webm](https://github.com/user-attachments/assets/e3708db3-fbb5-462c-97bc-b423ef2301db)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Code LGTM, but do you think it's worth adding an RTL test to harden it?

amaannawab923:
Screenshot 2025-04-30 at 1 06 54 AM
@geido  There is a bug on the master branch in which if the Select First by default is checked, then the DataMask filter State is erased by the method 
      dispatch(setDataMaskForFilterChangesComplete(filterChanges, oldFilters));
Causing re renders & would impact other functionailities related to Native filters including inverse selection... 
I have tried out on the Master branch 
Steps to reproduce
Create a Dashboard with a chart first
1. Place the Chart on the grid
2. Create a native filter with configuration , Select First value by default & Filter Value is required
3. Save this 

This method   dispatch(setDataMaskForFilterChangesComplete(filterChanges, oldFilters)); will erase the filter State & extra form data in case of default Value or first value is required causing it to become invalid for a while & cause re renders

amaannawab923:
This happens only for new filters on first time creation, I am trying to resolve it so it wont impact native filter features in future

amaannawab923:
@geido  , @rusackas There is a bug on master branch which will break the native filter features in future 
      dispatch(setDataMaskForFilterChangesComplete(filterChanges, oldFilters));
This method erases the DataMask extra form data & filterState on default Value selected .... 
The impact
If you create a native filter with a Default value as first selected
& then modify the native filter & enable some new feature 
The native filter will crash 
Steps to reproduce:
1. Create a Dashboard with chart , In this case Top Timezones
2. Now create a native filter with following configuration: 
Filter Value is required
Select first filter value by default 
Can select multiple values 
Disable/Unselect Allow creation of new values
Screenshot 2025-04-30 at 1 40 29 AM

3. Create the filter & now it will work properly ... with default filter
Screenshot 2025-04-30 at 1 43 20 AM

4. Now modify the filter & enable Allow Creation of New Values & save it

Screenshot 2025-04-30 at 1 44 16 AM

5. The chart will crash since this method dispatch(setDataMaskForFilterChangesComplete(filterChanges, oldFilters)); will erase the filterState & extraFormData which was not required in this case

amaannawab923:
I thought this was due to Exclude Filter values feature since i found out during testing but figured out this always happens  & its a bug on the master branch ....
cc: @geido @rusackas

amaannawab923:
@geido  Fixed the filter over write bug which was reported above

geido:
@amaannawab923 could you please include in the PR description the approach to fix the bugs and what bugs has this fixed? Thank you!

geido:
Echoing @evans suggestion on adding tests to make this mergiable and to avoid that the same regressions will enter the codebase in the future.

A few issues from manual testing:

We might need to tweak the "Select All" - "Deselect All" UI here as it cuts this off in small sizes. cc @msyavuz for help.

Screenshot 2025-04-30 at 18 50 59

Not sure if it possible to keep the Select only in one line to avoid this UI behavior. I think we do it correctly in horizontal mode so hopefully it should be easy to implement in vertical mode too

Screenshot 2025-04-30 at 18 55 06

This bug seems a bit more serious. I have created a filter with a default value, saved and used the filter normally. I then updated the default value with a different value and saved but the now the filter is empty and shows no default. What is even more concerning is that my browser completely freezes when doing that operation.

michael-s-molina:
Thanks for the PR @amaannawab923.

> This pr fixes a bug introduced as part of the previous pr of Exclude Filter Values

It's good to think about PR descriptions as historical documents that someone will read in the future to understand the reasons for a change. For that reason, it's a good practice to always add as much context as possible to the PR description, given that people reading the PR might not have any previous context about what you're working on. Linking the previous PR that caused the regression is also helpful especially when dealing with releases.

Could you improve the PR description by linking the previous PR and filling the BEFORE/AFTER SCREENSHOTS and TESTING INSTRUCTIONS sections?

amaannawab923:
@geido  The freezing ui is some infinite re render present on master branch , I discovered it too after seeing your comment , I reverted all the exclude filter values code & its not related to Exclude Filter Values code ... Still trying to figure out what is causing the infinite re render

amaannawab923:
Yes @michael-s-molina , I will write the pr with all instructions in detail ... Was busy with manual testing & covering all cases .... Will make sure to add all the details

michael-s-molina:
> I will write the pr with all instructions in detail ... Was busy with manual testing & covering all cases .... Will make sure to add all the details

@amaannawab923 Thank you. If you're still working on the PR, a good tip is to mark it as Draft 😉

amaannawab923:
@michael-s-molina  @rusackas  @geido  Please review the pr, Resolved the bugs , Attatched all necessary videos , screenshots , links & explaination of bugs & their resolution 
Thank you !

geido:
@amaannawab923 

It says a filter value is required even though I did not mark it as required but I only set a default value in config. On top of that, it does not care whether I add values or not. It is stuck like this.

Screenshot 2025-05-02 at 16 57 37

Even when I remove the default value option from the config, I still get the error message. The only difference now is that it allows me to apply the filter.

Screenshot 2025-05-02 at 17 00 33

In general I am not sure why that "value is required" is showing up. It breaks the horizontal layout too.

Also, I still see this happening in vertical mode. Can we fix it to one line like horizontal mode?

Screenshot 2025-05-02 at 17 03 29

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -288,58 +320,78 @@ export default function PluginFilterSelect(props: PluginFilterSelectProps) {
...
+          : null;
+        if (firstItem?.[0] !== undefined) {
+          updateDataMask(firstItem);
+        }
+      } else if (formData?.defaultValue) {
+        // Set to defaultValue if available
+        updateDataMask(formData.defaultValue);
+      }
+      return;
+    }
```
geido, line 349: @amaannawab923 I am curious how these logics came into existence. It seems they are pretty much a duplicate of what's after? Should we reconsider if this is actually needed or DRY it up if it really is?

amaannawab923: True , this repetitive , Removed the repetitive code

amaannawab923:
> @amaannawab923
> 
> It says a filter value is required even though I did not mark it as required but I only set a default value in config. On top of that, it does not care whether I add values or not. It is stuck like this.
> 
> Screenshot 2025-05-02 at 16 57 37
> 
> Even when I remove the default value option from the config, I still get the error message. The only difference now is that it allows me to apply the filter.
> 
> Screenshot 2025-05-02 at 17 00 33
> 
> In general I am not sure why that "value is required" is showing up. It breaks the horizontal layout too.
> 
> Also, I still see this happening in vertical mode. Can we fix it to one line like horizontal mode?
> 
> Screenshot 2025-05-02 at 17 03 29

Resolved this , The validation state was being manipulated by Filter config form's control

amaannawab923:
Screenshot 2025-05-06 at 3 03 03 PM

The above test case which was failing "Value is required" is passed now & not reproducible anymore in the ephmeral environment @geido 
cc: @kgabryje  @Antonio-RiveroMartnez 

> @amaannawab923
> 
> It says a filter value is required even though I did not mark it as required but I only set a default value in config. On top of that, it does not care whether I add values or not. It is stuck like this.
> 
> Screenshot 2025-05-02 at 16 57 37
> 
> Even when I remove the default value option from the config, I still get the error message. The only difference now is that it allows me to apply the filter.
> 
> Screenshot 2025-05-02 at 17 00 33
> 
> In general I am not sure why that "value is required" is showing up. It breaks the horizontal layout too.
> 
> Also, I still see this happening in vertical mode. Can we fix it to one line like horizontal mode?
> 
> Screenshot 2025-05-02 at 17 03 29

Antonio-RiveroMartnez (approved):
lgtm

superset-frontend/src/dataMask/reducer.ts
```diff
@@ -106,10 +114,27 @@ function updateDataMaskForFilterChanges(
...
+    const existingFilter = draftDataMask[filter.id] as FilterWithExtaFromData;
+
+    // Check if targets are equal
+    const areTargetsEqual = isEqual(existingFilter?.targets, filter?.targets);
+
+    // Preserve state only if filter exists, has enableEmptyFilter=true and targets match
+    const shouldPreserveState =
+      existingFilter &&
+      areTargetsEqual &&
+      (filter.controlValues?.enableEmptyFilter === true ||
```
kgabryje, line 126:
```suggestion
      (filter.controlValues?.enableEmptyFilter ||
```

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -158,6 +166,30 @@ export default function PluginFilterSelect(props: PluginFilterSelectProps) {
       : filterState?.excludeFilterValues,
   );
 
+  const prevExcludeFilterValues = useRef(excludeFilterValues);
+
+  const hasOnlyOrientationChanged = useRef(false);
+
+  useEffect(() => {
+    // Get previous orientation for this specific filter
+    const previousOrientation = orientationMap.get(formData.nativeFilterId);
```
kgabryje, line 175: why do we need it to be a map? Can't we simply preserve the previous value with a ref? We also have a helper hook `usePrevious` for cases like this - `const prevOrientation = usePrevious(filterBarOrientation)` (prevOrientation will be undefined on the first render)

kgabryje (dismissed):

superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
amaannawab923: Doesnt work , this particular SelectFilterPlugin is having hard refresh & losing all of its values when used with refs , the approach used in this file uses a closure formed outside this component which is not deleted even on Complete remount

As suggested i did try using usePrevious & the function ran on change of orientation of filter bar which isnt expected 

Attatched are the screenshots when used usePrevious

Orientation changed from Vertical to Horizontal with usePrevious : 

Screenshot 2025-05-06 at 4 17 16 PM


Orientation changed from  Horizontal  to vertical with usePrevious : 
Now function ran count increased to 2
Screenshot 2025-05-06 at 4 17 34 PM

kgabryje: thanks for explanation!

kgabryje (approved):

Issue #30816, opened 2024-11-01
[SIP] A simpler chart API for dashboards


betodealmeida:
# [SIP] A simpler chart API for dashboards

# Motivation

Historically, the Apache Superset dashboard has been built on top its interactive, no-code, chart builder (”Explore”). While it makes sense to reuse the chart API endpoints for fetching the data for charts in dashboards, in order to make development faster, at the end of the day these are two different workflows: one is an interactive chart *creator*, the other is a mostly passive chart *consumer*.

This distinction has left considerable technical debt behind. In order for a dashboard to load a given chart it needs to perform a `POST` request to the the chart API, submitting a request payload that includes the `query` and `form_data` needed by the endpoint. This data is redundant, since  it’s already available to the backend (in fact, it needs to be fetched from the backend before the `POST`) and attached to the chart ID (which is also passed in the request). This causes the following problems:

1. Dashboards need to load the full data for all of its charts, instead of just their IDs. This can cause noticeable latency (in our case we’ve observed a latency of ~400ms for larger dashboards, and of course this could potentially be much higher).
2. For each chart, a `POST` request is done sending back the data that was just provided by the backend. Not only this is unnecessary, but it opens an attack vector where users can modify the form data in order to fetch unauthorized data — for example, in an embedded dashboard. While this should never be possible by design, there were multiple CVE reports exploiting this channel.
3. While the `POST` requests still benefit from the Superset cache (if one is configured), they won’t benefit from other caches along the way (reverse proxies, browser cache). Replacing some of these with `GET` requests will provide for better dashboard performance and fewer costs, since the caching can be done upstream, as long as response headers are correctly returned by Superset.

# Proposed change

Currently, in order to retrieve the data needed to render a chart with ID **74**, a dashboard with ID **8** will issue a request like this:

```bash
POST https://superset/api/v1/chart/data?form_data={"slice_id":74}&dashboard_id=8
```

The request payload is not small:

```json
{
  "datasource": {
    "id": 19,
    "type": "table"
  },
  "force": false,
  "queries": [
    {
      "filters": [
        {
          "col": "time_start",
          "op": "TEMPORAL_RANGE",
          "val": "No filter"
        }
      ],
      "extras": {
        "having": "",
        "where": ""
      },
      "applied_time_extras": {},
      "columns": [
        "first_time_developer"
      ],
      "metrics": [
        "count"
      ],
      "orderby": [
        [
          "count",
          false
        ]
      ],
      "annotation_layers": [],
[...]

PR #33285, opened 2025-04-29
fix: Adjust viz migrations to also migrate the queries object

### LINKED ISSUE #33152: Viz migrations are not migrating the query object
### Bug description

Viz migrations only migrate the form_data and not the query object. This is problematic as backend processes such as cache warmup or alerts & reports use the saved query object to generate the queries.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

luizotavio32:
### SUMMARY

This PR introduces migration code designed to convert the query objects of legacy charts. This migration is crucial for ensuring the proper functionality of the warm_up_cache endpoint, as well as the Alerts & Reports features.

To streamline the generation of query objects, we have implemented the `_build_query()` method in `processors.py` within each class responsible for legacy chart migrations. This method effectively translates the TypeScript logic from `buildQuery.ts` into Python. Additionally, auxiliary functions originally found in `buildQuery.ts`  have been translated and placed into a new Python file called `query_functions.py`.

For testing the migration process, a script was developed to compare query objects generated by TypeScript with those produced by Python. This script was validated against over 50,000 charts across all legacy types and achieved an accuracy rate of 100%.

Fixes https://github.com/apache/superset/issues/33152

### TESTING INSTRUCTIONS

Run superset db upgrade and superset db downgrade CLI commands.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
@mistercrunch @michael-s-molina this is pertinent to what we were discussing in the Release Strategy sync today.

michael-s-molina:
@rusackas Yes, @luizotavio32 (Airbnb) is working on this currently and I'm reviewing the work.

mistercrunch:
Great. Not sure if this is the right place for this conversation, but I'm wondering why `query_context` was needed on the backend in the first place. I'm guessing it's to support Alerts & Reports use cases, more specifically:

- cache-warmup "reports", where we want the backend to force-refresh the cache for a given dashboard
- Rich "Table" and "Pivot Table" reports, where we generate html tables with data in them instead of a pure screenshot
- anything else (?)

Now clearly the issue here is that the `form_data` -> `query_context`, while it's deterministic, the logic lives only in the frontend, meaning there's some clear issues around backend database migrations ...

Some related ideas:
- forcing some reimplementation of logic in the backend as part of migrations ... that's pretty rough, I think that's the general idea for this PR. This is fairly brittle and hard to test. How do we know that the logic is sound/perfect? Probably need to fake a migration and inspect the charts ... Not ideal.
- sandboxing JS code in-python, forcing the execution of `buildQueryContext(form_data)` (or whatever it's called) in the backend
- "invalidating" query_context upon migrations, forcing someone with a browser, or a headless browser to pick up the work to regenerate the `query_context` post-migration. This is kind of nasty and sub-ideal, but technically could provide a solution. In theory, some specific endpoints could be developed to support this, think `/api/v1/headless/regenerate_query_context?chart_id={chart_id}`. Have to call an async job with a headless browser ...
- similar to above, except we flag alerts & reports as disabled, and show a `[re-enable]` button in the UI that uses frontend code to regenerate and push a fresh `query_context` to the backend ...
- for force-refreshing the cache, in theory we could have a similar `/api/v1/headless/force_refresh_cache?chart_id={chart_id}`

I'm not advocating for these ideas but wanted to itemize and share to see if it sparks other ideas.

michael-s-molina:
@mistercrunch I forgot to mention but we also noticed that the query_context evolved over the years without schemas and versioning, which resulted in many versions of the query_context for the same chart type. In other words, an ECharts Line chart created some time ago does not have the same schema as one created recently. One of the objectives of this PR is to also align these schemas for the migrated charts.

betodealmeida:
I started working on a node.js service that converts the form data to the query object using the viz logic, I can try to finish that work in the next week or two. It would also allow us to implement https://github.com/apache/superset/issues/30816.

michael-s-molina (approved):

Issue #33294, opened 2025-04-30
Virtual Dataset get_filters ignores IS NULL and IS NOT NULL filters


Prokos:
### Bug description

Using the `get_filters` jinja macro inside of a virtual dataset does not get a filter if the operator is set to IS NULL or IS NOT NULL. Because of this it also does not remove them when `remove_filter` is set to `true`.

IS NULL and IS NOT NULL filters are the only SIMPLE filters that do not have a `comparator`. They fail this check.

Proposed fix: https://github.com/apache/superset/pull/33296

### Superset version

4.1.2

### Python version

I don't know

### Node version

I don't know

### Browser

Not applicable

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
Thanks for opening the PR to tackle this!

Vitor-Avila:
Closing this one as https://github.com/apache/superset/pull/33296 was merged

PR #33296, opened 2025-04-30
fix(template_processing): get_filters now works for IS_NULL and IS_NOT_NULL operators

Prokos:
### SUMMARY
The macro `get_filters` within the Jinja context can be used to implement filtering in custom locations in your virtual dataset SQL. This works great, except for when filtering with the IS_NULL or IS_NOT_NULL operator. These are ignored, cannot be used, and will break certain implementations that rely on remove_filter to be functional (like the one I've outlined below).

The reason the code breaks for these operators is because they are unique in the sense that they don't have a comparator. Adjusting the code to not need a comparator for these operator types fixes the issue.

### TESTING INSTRUCTIONS
1. Create a virtual dataset on BigQuery with the following SQL 
- likely reproducible with less - but it's vital that both column selection/grouping and filtering is handled
- not a BigQuery specific bug, thats just the database I use
```
{% set finalColumns = columns if columns is defined else ['__ALL__'] %}

SELECT


  sum(measure) as measure,
  
  {% for col in finalColumns %}

  	{% if col == 'col1' or col == '__ALL__' or (col is mapping and 'sqlExpression' in col and col.sqlExpression == 'col1') %}
  		col1
  		{%- if col == '__ALL__' -%},{%- endif -%}
  		{%- if not loop.last -%},{%- endif -%}
  	{% endif %}
  	
  	{% if col == 'col2' or col == '__ALL__' or (col is mapping and 'sqlExpression' in col and col.sqlExpression == 'col2') %}
  		col2
  		{%- if col == '__ALL__' -%},{%- endif -%}
  		{%- if not loop.last -%},{%- endif -%}
  	{% endif %}
  	
  {% endfor %}
  
FROM UNNEST([
  STRUCT(1 as measure, 'value1A' as col1, 'value2A' as col2),
  STRUCT(2 as measure, 'value1A' as col1, 'value2B' as col2),
  STRUCT(5 as measure, 'value1B' as col1, 'value2B' as col2),
  STRUCT(15 as measure, NULL as col1, NULL as col2)
])

-- FILTERS

{% for filter in get_filters('col1', remove_filter=True) -%}
    {%- if loop.first %} WHERE 1 = 1 {% endif %}
    AND
    {% if filter.get('op') == 'IN' -%}
        col1 IN {{ filter.get('val')|where_in }}
    {% elif filter.get('op') == 'NOT IN' -%}
        col1 NOT IN {{ filter.get('val')|where_in }}
    {% elif filter.get('op') == 'ILIKE' -%}
        LOWER(col1) LIKE LOWER({{ filter.get('val') }})
    {% elif filter.get('op') == 'IS NULL' -%}
        col1 IS NULL
    {% elif filter.get('op') == 'IS NOT NULL' -%}
        col1 IS NOT NULL
    {% else -%}
        col1 {{ filter.get('op') }} {{ filter.get('val') }}
    {%- endif -%}
{%- endfor %}

{% for filter in get_filters('col2', remove_filter=True) -%}
    {% if loop.first %} WHERE 1 = 1 {% endif %}
    AND
    {% if filter.get('op') == 'IN' -%}
        col2 IN {{ filter.get('val')|where_in }}
    {% elif filter.get('op') == 'NOT IN' -%}
        col2 NOT IN {{ filter.get('val')|where_in }}
    {% elif filter.get('op') == 'ILIKE' -%}
        LOWER(col2) LIKE LOWER({{ filter.get('val') }})
    {% elif filter.get('op') == 'IS NULL' -%}
        col2 IS NULL
    {% elif filter.get('op') == 'IS NOT NULL' -%}
        col2 IS NOT NULL
    {% else -%}
        col2 {{ filter.get('op') }} {{ filter.get('val') }}
    {%- endif -%}
{%- endfor %}

-- GROUP BY
  
{% set ns = namespace(groupByEmitted=false) %}

{% for col in finalColumns %}

	{% if col == 'col1' or col == '__ALL__' or (col is mapping and 'sqlExpression' in col and col.sqlExpression == 'col1') %}
		{%- if not ns.groupByEmitted %}
			GROUP BY
			{% set ns.groupByEmitted = true %}
		{% endif -%}
		col1
		{%- if col == '__ALL__' -%},{%- endif -%}
		{%- if not loop.last -%},{%- endif -%}
	{% endif %}
	
	{% if col == 'col2' or col == '__ALL__' or (col is mapping and 'sqlExpression' in col and col.sqlExpression == 'col2') %}
		{%- if not ns.groupByEmitted %}
			GROUP BY
			{% set ns.groupByEmitted = true %}
		{% endif -%}
		col2
		{%- if not loop.last -%},{%- endif -%}
	{% endif %}

{% endfor %}
  
``` 

3. Click "Create Chart"
4. Set a filter on one of the columns but don't select it as a dimension:
image

5. On master, this breaks due to the following resulting SQL:

```
SELECT `measure` AS `measure`, `col2` AS `col2` 
FROM (
  SELECT
    sum(measure) as measure,
    col2
    
  FROM UNNEST([
    STRUCT(1 as measure, 'value1A' as col1, 'value2A' as col2),
    STRUCT(2 as measure, 'value1A' as col1, 'value2B' as col2),
    STRUCT(5 as measure, 'value1B' as col1, 'value2B' as col2),
    STRUCT(15 as measure, NULL as col1, NULL as col2)
  ])
	
  GROUP BY		
		col2
) AS `virtual_table` 
WHERE `col1` IS NULL
 LIMIT 1000
```

7. With this fix implemented, this works and the resulting SQL looks like this:

```
SELECT `measure` AS `measure`, `col2` AS `col2` 
FROM (

SELECT

  sum(measure) as measure,
  col2
  
FROM UNNEST([
  STRUCT(1 as measure, 'value1A' as col1, 'value2A' as col2),
  STRUCT(2 as measure, 'value1A' as col1, 'value2B' as col2),
  STRUCT(5 as measure, 'value1B' as col1, 'value2B' as col2),
  STRUCT(15 as measure, NULL as col1, NULL as col2)
])

 WHERE 1 = 1 
  AND
  col1 IS NULL
	
	GROUP BY	
		col2

) AS `virtual_table`
 LIMIT 1000
```

### ADDITIONAL INFORMATION
- [X] Fixes https://github.com/apache/superset/issues/33294
- [X] Required feature flags: ENABLE_TEMPLATE_PROCESSING = True

### DIFF
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -380,9 +380,8 @@ class ExtraCache:
                 and flt.get("subject") == column
                 and (
                     val
-                    or
                     # IS_NULL and IS_NOT_NULL operators do not have a value
-                    op
+                    or op
                     in (
                         FilterOperator.IS_NULL.value,
                         FilterOperator.IS_NOT_NULL.value,
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -378,7 +378,15 @@ class ExtraCache:
                 flt.get("expressionType") == "SIMPLE"
                 and flt.get("clause") == "WHERE"
                 and flt.get("subject") == column
-                and val
+                and (
+                    val
+                    or
+                    # IS_NULL and IS_NOT_NULL operators do not have a value
+                    op in (
+                        FilterOperator.IS_NULL.value,
+                        FilterOperator.IS_NOT_NULL.value,
+                    )
+                )
             ):
                 if remove_filter:
                     if column not in self.removed_filters:
--- a/superset/jinja_context.py
+++ b/superset/jinja_context.py
@@ -382,7 +382,8 @@ class ExtraCache:
                     val
                     or
                     # IS_NULL and IS_NOT_NULL operators do not have a value
-                    op in (
+                    op
+                    in (
                         FilterOperator.IS_NULL.value,
                         FilterOperator.IS_NOT_NULL.value,
                     )
--- a/tests/unit_tests/jinja_context_test.py
+++ b/tests/unit_tests/jinja_context_test.py
@@ -214,6 +214,36 @@ def test_get_filters_adhoc_filters() -> None:
         assert cache.applied_filters == ["name"]
 
 
+def test_get_filters_is_null_operator() -> None:
+    """
+    Test the ``get_filters`` macro with a IS_NULL operator,
+    which doesn't have a comparator
+    """
+    with app.test_request_context(
+        data={
+            "form_data": json.dumps(
+                {
+                    "adhoc_filters": [
+                        {
+                            "clause": "WHERE",
+                            "expressionType": "SIMPLE",
+                            "operator": "IS NULL",
+                            "subject": "name",
+                            "comparator": None,
+                        }
+                    ],
+                }
+            )
+        }
+    ):
+        cache = ExtraCache()
+        assert cache.get_filters("name", remove_filter=True) == [
+            {"op": "IS NULL", "col": "name", "val": None}
+        ]
+        assert cache.removed_filters == ["name"]
+        assert cache.applied_filters == ["name"]
+
+
 def test_get_filters_remove_not_present() -> None:
     """
     Test the ``get_filters`` macro without a match and ``remove_filter`` set to True.


Prokos:
@sfirke Pre-commit now runs successfully locally :)

superset/jinja_context.py
```diff
@@ -378,7 +378,15 @@ def get_filters(self, column: str, remove_filter: bool = False) -> list[Filter]:
...
-                and val
+                and (
+                    val
+                    or
+                    # IS_NULL and IS_NOT_NULL operators do not have a value
+                    op in (
+                        FilterOperator.IS_NULL.value,
+                        FilterOperator.IS_NOT_NULL.value,
+                    )
+                )
```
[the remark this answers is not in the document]
Prokos: Not me talking to an AI: after the ruff-format fix maybe the final logic is a bit too expanded but I'll leave it to a reviewer to decide on what formatting should be used. At least now it's passing the test.

sfirke:
I tagged a few committers who I believe are familiar with the Jinja internals. Hopefully one of them can take a quick look and we can get this merged!

superset/jinja_context.py
```diff
@@ -378,7 +378,16 @@ def get_filters(self, column: str, remove_filter: bool = False) -> list[Filter]:
...
+                and (
+                    val
+                    or
+                    # IS_NULL and IS_NOT_NULL operators do not have a value
+                    op
+                    in (
+                        FilterOperator.IS_NULL.value,
+                        FilterOperator.IS_NOT_NULL.value,
+                    )
+                )
```
Vitor-Avila, line 390: hmm what do you think about this? I noticed the convo with Korbit too.
``` python
and (
    val
    # IS_NULL and IS_NOT_NULL operators do not have a value
    or op in (
        FilterOperator.IS_NULL.value,
        FilterOperator.IS_NOT_NULL.value,
    )
)
```

Prokos: Nice!

Prokos:
@Vitor-Avila Thanks for having a look!

I added a test, that I confirmed fails without my change, and is successful with my change.

I've also implemented your formatting suggestion. (Note that pre-commit hooks forced the `in` on a separate line still, so all I did was move the `or` which I think is still an improvement).

Vitor-Avila (approved):
LGTM! Thank you so much for fixing this flow!

PR #33341, opened 2025-05-02
chore(explore): Add format sql and view in SQL Lab option in View Query

justinpark:
### SUMMARY
The "View Query" feature was previously used to display SQL in a formatted manner. However, the prettier formatting was reverted in [this pull request](https://github.com/apache/superset/pull/30350) due to unintended changes in the original SQL. To minimize the impact while retaining the formatting option, this commit introduces a "Format SQL" feature in View Query modal and an option to show the original SQL for comparison. Additionally, this commit adds a "View in SQL Lab" button, allowing users to conveniently connect to SQL Lab directly from the dashboard, which is particularly useful for users.

Before:
Screenshot 2025-05-01 at 5 20 53 PM

After:

https://github.com/user-attachments/assets/504b1e61-0dad-4c1a-8c57-d6c9b3c47842

### TESTING INSTRUCTIONS

Click "View Query" in the chart or explore and then check the toolbars above.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/ViewQuery.test.tsx
+++ b/superset-frontend/src/explore/components/controls/ViewQuery.test.tsx
@@ -1,0 +1,145 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import {
+  screen,
+  render,
+  fireEvent,
+  waitFor,
+} from 'spec/helpers/testing-library';
+import fetchMock from 'fetch-mock';
+import copyTextToClipboard from 'src/utils/copy';
+import ViewQuery, { ViewQueryProps } from './ViewQuery';
+
+const mockHistoryPush = jest.fn();
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useHistory: () => ({
+    push: mockHistoryPush,
+  }),
+}));
+
+jest.mock('src/utils/copy', () =>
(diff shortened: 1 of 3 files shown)


michael-s-molina:
@justinpark The `View in SQL Lab` button is a great addition!

About the "Format SQL" feature, I think a formatted SQL is better for users when visualizing a query. Given that https://github.com/apache/superset/pull/30350 states:

> When running a query we pass it through sqlglot to prettify the SQL, but sqlglot is doing some unwanted changes when the dialect is set to generic.

and the fact that the dialect will not be generic most of the time, I wonder if we should always format the SQL when visualizing the query. To be clear, I'm not proposing reverting @betodealmeida's changes but to always format the SQL when clicking on `View query`. 

If the dialect is generic, then you could always show the non-formatted version of the query and warning saying that the SQL can't be formatted for that database type.

Pinging @betodealmeida, who probably has more context, to review too.

justinpark:
> and the fact that the dialect will not be generic most of the time, I wonder if we should always format the SQL when visualizing the query. To be clear, I'm not proposing reverting betodealmeida's changes but to always format the SQL when clicking on View query.

I like the idea too. Since the formatted SQL in this section doesn't modify the original SQL, it should be okay.
Additionally, I plan to display a "Show Original" button so that users can revert back to the original SQL if needed, allowing access to both versions.

justinpark:
Here is the updated design.

Screenshot 2025-05-02 at 1 29 25 PM

superset-frontend/src/explore/components/controls/ViewQuery.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+afterEach(() => {
+  jest.resetAllMocks();
+  fetchMock.restore();
+});
+
+const getFormatSwitch = () =>
+  screen.getByRole('switch', { name: 'Show original SQL' });
+
+it('renders the component with Formatted SQL and buttons', async () => {
```
michael-s-molina, line 74: Can you use `test` instead of `it`?

michael-s-molina (approved):

sfirke:
@justinpark I am testing this feature now. The slider for formatted vs. original is greyed out and nothing happens when I click on it. Is this intended behavior, maybe because my data warehouse is Microsoft SQL Server (MSSQL) and it lacks some features? As a user I am trying to understand this.
image

justinpark:
>  I am testing this feature now. The slider for formatted vs. original is greyed out and nothing happens when I click on it. Is this intended behavior, maybe because my data warehouse is Microsoft SQL Server (MSSQL) and it lacks some features? As a user I am trying to understand this. 

@sfirke I just saw your question now. To answer: the "format SQL" feature uses the formatting function provided by sqlglot. Currently, sqlglot does not have a dedicated dialect for MSSQL, so formatting is based on T-SQL. However, some syntax is not correctly recognized in this process, which causes parsing errors and may result in the SQL not being formatted/rendered properly.

PR #33354, opened 2025-05-04
fix: loading examples from raw.githubusercontent.com fails with 429 errors

mistercrunch:
Recently we've been getting `HTTP Error 429: Too Many Requests` in CI while loading examples.

Historically we fetched the sample files straight from `raw.githubusercontent.com`.  Anonymous requests are limited to **60 per hour per IP** on GitHub, so our busy CI matrix started to hit 429s.

We now point to **jsDelivr** – a multi‑CDN cache in front of every public GitHub repo.  It doesn’t consume the GitHub API quota and advertises unlimited bandwidth for OSS.

### DIFF
--- a/superset/examples/helpers.py
+++ b/superset/examples/helpers.py
@@ -14,6 +14,40 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+"""Helpers for loading Superset example datasets.
+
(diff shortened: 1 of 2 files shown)


geido (approved):

IdlePhysicist:
@mistercrunch @geido is there a plan to cut a release with this fix in the near term future?

PR #33356, opened 2025-05-05
fix(embedded): handle SUPERSET_APP_ROOT in embedded dashboard URLs

irodriguez-nebustream:
### SUMMARY

This PR adds support for `SUPERSET_APP_ROOT` subpath routing to the embedded dashboard SDK. When Superset is configured with a subpath (e.g., `SUPERSET_APP_ROOT="/prefix"`), the embedded dashboard SDK now correctly constructs the iframe URL to include this path prefix.

**Problem:**
When embedding dashboards in applications where Superset is deployed with a subpath routing configuration, the SDK would attempt to load the dashboard at `/embedded/{id}` instead of the correct path `/prefix/embedded/{id}`, resulting in 404 errors.

**Solution:**
- Added an optional `appRootPath` parameter to the `EmbedDashboardParams` type
- Modified the `embedDashboard` function to include the application root path in the iframe source URL
- Normalized the app root path to ensure consistent handling (removing trailing slashes)

This change ensures the SDK produces iframe URLs that match Superset's routing configuration, eliminating 404 errors and making the embedded dashboard feature compatible with subpath deployments.

N/A - This is an SDK fix that affects routing behavior

### TESTING INSTRUCTIONS

1. Configure a Superset instance with `SUPERSET_APP_ROOT="/prefix"` in the config
2. Use the updated SDK to embed a dashboard:
   ```typescript
   embedDashboard({
     id: "your-dashboard-id",
     supersetDomain: "http://localhost:8088",
     mountPoint: document.getElementById("dashboard-container"),
     fetchGuestToken: () => fetchGuestTokenFromYourBackend()
   });
   ```
3. Verify that:
   - The iframe loads successfully without 404 errors
   - The iframe source URL is constructed as `/prefix/embedded/{id}`
   - The embedded dashboard renders correctly
   - API calls within the iframe (like `/prefix/api/v1/me/roles/`) work properly

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #XXXX (replace with actual issue number)
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration
- [ ] Introduces new feature or API ✓ (adds optional `appRootPath` parameter to SDK)

**Notes:**
- This change is backward compatible - existing implementations without `appRootPath` will continue to work as before
- The `appRootPath` parameter is optional and defaults to an empty string to maintain backward compatibility
- Related to Superset's use of `process.env.SUPERSET_APP_ROOT` for subpath routing configuration


irodriguez-nebustream:
I notice this happen a few times on my builds too. I get that 429 as well `urllib.error.HTTPError: HTTP Error 429: Too Many Requests`

morngaraf:
@irodriguez-nebustream 
Thank you for opening a PR to fix the issue. I noticed it as well, but ended up monkey patching during deployment on my side for the time-being.

In my first attempt I tried to replace only `/api/v1/me/roles/` with the expected path, but it was not enough: React routes, e.g. `<Route path="/embedded/:uuid/" component={EmbeddedRoute} />`, are also broken. When I "fixed" the routes, dashboard started loading, but all API calls the dashboard itself is making (like `/api/v1/dashboard` etc.) were still using the wrong path.

I suspect that the following changes have to be done to fix all of the above:
```ts
import getBootstrapData, { applicationRoot } from 'src/utils/getBootstrapData';

...

const EmbeddedApp = () => (
  <Router basename={applicationRoot()}>
    {/* todo (embedded) remove this line after uuids are deployed */}
    <Route path="/dashboard/:idOrSlug/embedded/" component={EmbeddedRoute} />
    <Route path="/embedded/:uuid/" component={EmbeddedRoute} />
  </Router>
);

...

function setupGuestClient(guestToken: string) {
  setupClient({
    appRoot: applicationRoot(),
    guestToken,
    guestTokenHeaderName: bootstrapData.config?.GUEST_TOKEN_HEADER_NAME,
    unauthorizedHandler: guestUnauthorizedHandler,
  });
}
```


`<Router basename={applicationRoot()}>` is taken from `App.tsx`, 
`setupClient({ appRoot: applicationRoot() });` is used in `preamble.ts`.

Since `makeApi` by default uses singleton of `SupersetClient`, which is configured by `setupClient`, we should not have to modify the actual '/api/v1/me/roles/' endpoint with these changes, and all other "broken" api calls should also be fixed automatically.

Anyways, these are simply my guesses and I'm quite new to Superset. Maybe later this month I or one of my colleagues will have time to verify the theories locally, and we would be happy to help moving this forward.

rusackas:
I think this will need a rebase to get the CI to pass (it's stuck on a docker fix that was contributed after this PR opened).

irodriguez-nebustream:
@Vitor-Avila @rusackas Let me try implementing the other things that @morngaraf mentioned. I think its worth looking into and I can thoroughly test locally.

irodriguez-nebustream:
image
Okay, I can 100% confirm that the changes  that @morngaraf mentioned worked locally. I was able to get the embedded section working and I built the image, tagged it locally then spun up a container and called it from my webapp. I can add a screenshot of the network call if that helps.

rusackas:
running CI... ping me if it looks good 🤞

irodriguez-nebustream:
@rusackas I think everything passed. Except for that skipped check.

superset-frontend/src/embedded/index.tsx
```diff
@@ -41,6 +41,7 @@ setupPlugins();
 
 const debugMode = process.env.WEBPACK_MODE === 'development';
 const bootstrapData = getBootstrapData();
+const appRootPath = process.env.SUPERSET_APP_ROOT || '';
```
mistercrunch, line 44: mmmh, wondering why we go through `process.env` here, seems we should only use `applicationRoot()` ideally always outside of module scope...

superset-frontend/src/embedded/index.tsx
```diff
@@ -156,7 +157,7 @@ function guestUnauthorizedHandler() {
 function start() {
   const getMeWithRole = makeApi<void, { result: UserWithPermissionsAndRoles }>({
     method: 'GET',
-    endpoint: '/api/v1/me/roles/',
+    endpoint: `${appRootPath}/api/v1/me/roles/`,
```
mistercrunch, line 160: why not `applicationRoot()` here?

morngaraf: i think we should leave the original `'/api/v1/me/roles/'` endpoint value here:
now with `appRoot` being passed to `setupClient` this `makeApi` call should work properly without explicit app root path in the endpoint.

irodriguez-nebustream: I updated the code with the changes. Everything works as expected.

superset-frontend/src/embedded/index.tsx
irodriguez-nebustream: Agreed, this was not needed. I made the changes.

irodriguez-nebustream:
@mistercrunch @rusackas I went ahead an updated the code to match what the comments and suggested changes that were mentioned. I verified the changes locally and everything works the same. Attached a new screenshot 
image

morngaraf (approved):

rusackas:
Approving CI to run again 🤞

mistercrunch (approved):
LGTM

Issue #31626, opened 2024-12-26
Table chart pagination is not working since version 4.1.0


Yisel-Ulloa:
### Bug description

Pagination is not working correctly. When interacting with the dropdown menu to change the number of records to display, the number of records does not update.

If I set the 'server page length' to the value of 200, the number of records updates correctly based on the selection in the dropdown menu. However, if I leave the value at 10, the information does not update and continues to show only those 10 records, even when I select a higher number or the option to show all records from the selector.

I appreciate your help.

https://github.com/user-attachments/assets/16697e76-c86a-4f0f-a1d8-d00ee64e5114

You can see in the images that it always shows me 10 records even if I have a larger number selected.
![Captura de pantalla 2024-12-26 152851](https://github.com/user-attachments/assets/67635e08-be6a-4c96-9782-524a2bd880d5)
![Captura de pantalla 2024-12-26 152905](https://github.com/user-attachments/assets/06f3fd04-c14f-4c7c-89c9-236c32e831a1)

### Superset version

4.1.1

### Python version

3.10

### Node version

Not applicable

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
Sorry to ask the obvious question, but is the table scrollable (i.e. you SEE the first 10, but there are more that you can scroll to)?

I'm unable to reproduce this issue when I try it, and the change in the little page number buttons at the bottom of your screenshots also indicate it's working.

KlyapkoV:
In a Table-type graph, if you select **Server pagination** and set the value to m, then when you select n in **Show n entries** only m records will be displayed on the page if n > m.

The dataset contains more than 10 rows.

![Image](https://github.com/user-attachments/assets/d13a9e05-a480-4f68-a93e-82a8f1e34d96)

![Image](https://github.com/user-attachments/assets/c1fb84dc-ea46-40e1-a15b-1d3802ca3f48)

amaannawab923:
The mentioned issue has been fixed with the pr https://github.com/apache/superset/pull/33357 @rusackas

rusackas:
This should have been closed by https://github.com/apache/superset/pull/33357 (but it didn't have the "fixes #31626" text that automagically does so :D

[...]

PR #33357, opened 2025-05-05
feat(Table Chart): Row limit Increase , Backend Sorting , Backend Search , Excel/CSV Improvements

amaannawab923:
###  🚀 Row Limit Increase
This PR introduces a more flexible and user-aware row limit configuration for the Table visualization in Apache Superset. Previously, the row limit was globally enforced at 100,000 rows via the SQL_MAX_ROW setting. This behavior has now been refined to allow significantly larger datasets when server-side pagination is enabled.

###  🔧  New Config Variables:
Two new configuration options have been introduced in config.py to distinguish row limits based on whether server-side pagination is enabled:

### TABLE_VIZ_MAX_ROW_CLIENT = 100000
→ Maximum number of rows allowed without server-side pagination (frontend capped).

### TABLE_VIZ_MAX_ROW_SERVER = 500000
→ Maximum number of rows allowed with server-side pagination enabled.

### 🎯  Behavior Changes:
If server pagination is disabled, users will continue to be capped at 100,000 rows.

Screenshot 2025-05-06 at 12 10 46 AM

Screenshot 2025-05-06 at 12 10 58 AM

If server pagination is enabled, the row limit increases to 500,000 rows, allowing users to query and paginate through much larger datasets.

Screenshot 2025-05-06 at 12 11 20 AM

### ⚠️  Frontend UX Improvements:
A new warning tooltip is shown on the row limit control when a user attempts to fetch more than 100k rows without enabling server pagination. It advises:

“Server pagination must be enabled for row limits over 100,000.”

If a user attempts to exceed 500,000 rows, the tooltip now shows:

“Value cannot exceed 500,000 rows.”

Below are the test cases & expectations of this functionality : 

https://www.notion.so/Table-Performance-Improvements-Test-Cases-1ea89b69c3c48038aec6ddb7be2d7b98

### ↕️ Backend Sorting for Table Charts
This PR introduces backend-driven sorting for Table charts, improving performance and scalability when handling large datasets.

### 📊 Affected Scope:
Applies only to Table visualizations

Sorting is pushed to the database layer when server-side pagination is enabled

Screenshot 2025-05-06 at 12 14 30 AM

### ⚙️ Conditional Behavior:
✅ With server pagination enabled:
Sorting is now performed at the backend by injecting an ORDER BY clause in the SQL query. This ensures accurate, efficient sorting even with large datasets (up to 500k rows), and avoids excessive browser-side memory usage.

❌ With server pagination disabled:
Sorting behavior remains frontend-based as before. This is suitable for smaller datasets (under 100k rows), where client-side sorting is performant and simple.

### 📁 Excel/CSV Export Improvements
This PR significantly enhances the export functionality for Excel and CSV downloads from Table charts, making it more intuitive and aligned with user expectations.

### 🛠️ Key Changes:
Simplified Export Format:
Previously, exports were bundled as a .zip file containing two separate Excel files — one for metadata and another for actual data. This extra step caused confusion and was unnecessary. Now, the export returns a single .csv file (or .xlsx if needed) with the complete dataset.

Full Dataset Export (Not Just Current Page):

Before: Only the records shown on the current page (e.g., 10 rows) were exported.

Now: The export includes all records up to the row limit — ensuring that users get the entire dataset, whether that's 100k rows (client mode) or up to 500k (with server pagination enabled).

### ✅ Affects:
Both Excel and CSV export options under the Table visualization.

### 🔍 Backend Search Support for Table Charts
This PR introduces backend search for Table visualizations, enabling efficient, scalable text filtering directly within the database query layer — a major step forward for handling large datasets and improving search accuracy.

### 🧠 Search Logic & SQL Integration:
A new operator, TEXT_SEARCH, has been added to the SQLAlchemy filter operators.

This operator constructs a WHERE clause using a smart pattern with ILIKE for case-insensitive partial matching:

column ILIKE '{pattern}%' OR column ILIKE '% {pattern}%'

This allows matching across both first and last words, e.g., searching “Anderson” will find both John Anderson and Mary Anderson.

### 🧪 Technical Behavior:
Applies only when server-side pagination is enabled

Targets only string-based columns (non-string columns are cast to string when necessary)

Appends the search logic directly into the SQL query using SQLAlchemy

### 🖥️ UI Enhancements:
A new search column dropdown has been added next to the search input

This dropdown lists all string columns of the dataset

When the user types into the search box:

The selected column (from the dropdown) and the input value are combined to generate the appropriate TEXT_SEARCH SQL clause

The dropdown allows users to dynamically change which column the backend search applies to

Screenshot 2025-05-06 at 12 16 44 AM
Screenshot 2025-05-06 at 12 17 30 AM

### 🛠️  Fixes the issue

https://github.com/apache/superset/issues/31626

This issue bascially described that the page doesn't respect the entries per page even when the sql query returns correct number of rows , this was because the transformProps didnt get the correct Page Size , it always used server page length even when it was outdated

Before fix :

[screen-capture (13).webm](https://github.com/user-attachments/assets/e558dc6a-0bf3-4778-932d-183d3c84988a)

After fix: 

[screen-capture (14).webm](https://github.com/user-attachments/assets/a92cf686-5bcc-4b49-80cd-2df2ffce9aa3)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


amaannawab923:
Hey @Antonio-RiveroMartnez  , Can you please review this pr whenever available

superset/utils/core.py
```diff
@@ -1764,6 +1765,31 @@ def apply_max_row_limit(
...
 
+def apply_max_row_limit_table(
+    limit: int,
+    form_data: dict[str, Any] | None = None,
+) -> int:
+    """
+    Override row limit for table viz type based on server pagination setting
+
+    :param limit: requested row limit
+    :param max_limit: Maximum allowed row limit for table viz
```
Antonio-RiveroMartnez, line 1776: This is not an param of the method

superset/utils/core.py
```diff
@@ -1764,6 +1765,31 @@ def apply_max_row_limit(
...
+    :param form_data: Form data containing server_pagination setting
+    :return: Capped row limit
+    """
+    max_limit = None
```
Antonio-RiveroMartnez, line 1780: nit: If you initialize `max_limit = current_app.config["TABLE_VIZ_MAX_ROW_CLIENT"]` then you won't need the else below

superset/common/query_object_factory.py
```diff
@@ -105,6 +113,22 @@ def _process_row_limit(
         )
         return apply_max_row_limit(row_limit or default_row_limit)
 
+    def _process_row_limit_table(
```
Antonio-RiveroMartnez, line 116: Why not just extending the existing _process_row_limit so it can receive the optional form_data and pass it along instead of creating this specific method for tables?

superset/utils/core.py
```diff
@@ -1764,6 +1765,31 @@ def apply_max_row_limit(
     return max_limit
 
 
+def apply_max_row_limit_table(
```
Antonio-RiveroMartnez, line 1768: Same as with `_process_row_limit`, why not extending the current one instead of creating a table specific to handle optional param?

superset/config.py
```diff
@@ -969,6 +969,13 @@ class D3TimeFormat(TypedDict, total=False):
 # Maximum number of rows returned for any analytical database query
 SQL_MAX_ROW = 100000
 
+# Maximum number of rows that can be returned for table without Server pagination
+TABLE_VIZ_MAX_ROW_CLIENT = 100000
```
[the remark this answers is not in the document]
amaannawab923: Makes sense , since both of them have same values... We can eliminate one

superset/common/query_object_factory.py
amaannawab923: Makes sense , i was just too afraid to break some existing unit test cases that might be using _process_row_limit ... Let me do that

superset/utils/core.py
amaannawab923: Right

superset/models/helpers.py
```diff
@@ -1888,12 +1888,26 @@ def get_sqla_query(  # pylint: disable=too-many-arguments,too-many-locals,too-ma
...
+                            # Convert to string and handle None case
+                            search_value = str(eq) if eq is not None else ""
+                            if (
+                                search_value
+                            ):  # Only add clause if search value is not empty
+                                pattern = search_value.lower()
+                                where_clause_and.append(
+                                    sa.or_(
+                                        sqla_col.ilike(f"{pattern}%"),
+                                        sqla_col.ilike(f"% {pattern}%"),
```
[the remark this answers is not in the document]
Antonio-RiveroMartnez: I see, in that case, extensions like https://www.postgresql.org/docs/current/pgtrgm.html should be preferred which would have to leverage some new indexes for those fields you want to search. That's probably out of the scope of this PR itself, but using hardcoded leading spaces IMO is not something we should do

betodealmeida:
Taking a look now, but a few initial comments:

1. In general, any PR that changes the UI/UX significantly should go [through a discussion](https://github.com/apache/superset/issues/5602). For uncontroversial changes the process is usually quick, since you can request "lazy consensus" and people don't have to vote. I think we're OK in this one, just letting you know for the future.
2. Splitting the PR into 4 would've been better, even if it takes more time/work upfront. It makes reviews quicker and it's simpler to revert features if something goes wrong.
3. I'm a bit concerned about the performance of the `LIKE` search, and wondering if we should have optional ways of optimizing it, but let me look at the PR in details before we start worrying about this.

superset/models/helpers.py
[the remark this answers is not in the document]
betodealmeida: There also the question of UX here. I would expect that searching for `son` would match `Anderson`, since that's the behavior we have in other input fields in Superset:

Screenshot 2025-05-06 at 11 32 37 AM

So instead of adding a new operator here I think it's better to just use `ILIKE`.

superset-frontend/packages/superset-ui-core/src/query/types/Operator.ts
```diff
@@ -33,6 +33,7 @@ const BINARY_OPERATORS = [
   'NOT LIKE',
   'REGEX',
   'TEMPORAL_RANGE',
+  'TEXT_SEARCH',
```
betodealmeida, line 36: Let's just use `ILIKE` for consistency with how search works across the app, see my first comment.

superset/utils/core.py
betodealmeida: Or just do a single statement:

```python
max_limit = (
    current_app.config["TABLE_VIZ_MAX_ROW_SERVER"]
    if form_data and form_data.get("server_pagination")
    else current_app.config["TABLE_VIZ_MAX_ROW_CLIENT"]
)
```

superset/models/helpers.py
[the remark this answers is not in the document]
betodealmeida: @Antonio-RiveroMartnez keep in mind we're fitering the analytics database here, so most of them won't have indexes that can be easily leveraged.

superset/common/query_object_factory.py
betodealmeida: Also, instead of passing the whole `form_data` we can just pass `server_pagination`

superset/common/query_context_factory.py
```diff
@@ -70,7 +70,7 @@ def create(  # pylint: disable=too-many-arguments
                 datasource_model_instance,
                 form_data,
                 self._query_object_factory.create(
-                    result_type, datasource=datasource, **query_obj
+                    result_type, datasource=datasource, **query_obj, form_data=form_data
```
betodealmeida, line 73:
```suggestion
                    result_type, datasource=datasource, form_data=form_data, **query_obj
```

superset/models/helpers.py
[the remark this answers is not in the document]
amaannawab923: @betodealmeida  i agree , that we should maintain consistency from a ux point of view but filtering results on every possible substring of the search value absolutely breaks the table in case of analytics tables &  gives random results.. 
Here i am searching for all real_name that includes ad 

Screenshot 2025-05-06 at 9 53 25 PM

& the search operation begins to feel random  since Searches are supposed to be based on prefixes
An example would be linkedin search below since it has the search by name case 

Screenshot 2025-05-06 at 9 56 17 PM

amaannawab923:
Resolved all the comments @betodealmeida  @Antonio-RiveroMartnez 
Removed the Text search operator & refactored the process row limit function as per comments & removed redundant implementation

Please re review whenever available, Thanks !

superset/utils/core.py
```diff
@@ -1741,24 +1741,29 @@ def parse_boolean_string(bool_str: str | None) -> bool:
 
 def apply_max_row_limit(
     limit: int,
-    max_limit: int | None = None,
```
Antonio-RiveroMartnez, line 1744: Does this mean this method is now ignoring a custom `max_limit` and instead using the config max limits? because that's it not what it was doing before, meaning, this could be a braking change

superset/utils/core.py
```diff
@@ -1741,24 +1741,29 @@ def parse_boolean_string(bool_str: str | None) -> bool:
...
     :return: Capped row limit
 
-    >>> apply_max_row_limit(100000, 10)
-    10
-    >>> apply_max_row_limit(10, 100000)
-    10
-    >>> apply_max_row_limit(0, 10000)
-    10000
-    """
-    if max_limit is None:
```
Antonio-RiveroMartnez, line 1760: We used to use the config values only if no max_limit was provided, why the change to always using the configs now?

superset/utils/core.py
```diff
@@ -1741,24 +1741,29 @@ def parse_boolean_string(bool_str: str | None) -> bool:
 
 def apply_max_row_limit(
     limit: int,
-    max_limit: int | None = None,
+    server_pagination: bool = False,
```
Antonio-RiveroMartnez, line 1744: could this allow None too? so we're not forced to provide it

superset/common/query_object_factory.py
```diff
@@ -96,14 +102,27 @@ def _process_extras(
         return extras
 
     def _process_row_limit(
-        self, row_limit: int | None, result_type: ChartDataResultType
+        self,
+        row_limit: int | None,
+        result_type: ChartDataResultType,
+        server_pagination: bool = False,
```
Antonio-RiveroMartnez, line 108: Could this be None too?

superset/utils/core.py
amaannawab923: The have checked the function usage, The function is not using this custom max_limit anywhere in the code ... across all of its uage .... Hence i ended up removing it since it was not being used ....

superset/utils/core.py
amaannawab923: Because the function is no where using the max_limit from anywhere... I can add it again ... i only removed it because the argument wasn't getting used

Usages 

superset/viz.py
row_limit = apply_max_row_limit(row_limit). # No second arg for max_limit

superset/datasource/api.py
row_limit = apply_max_row_limit(app.config["FILTER_SELECT_ROW_LIMIT"]) # No second arg for max_limit

superset/sqllab/sqllab_execution_context.py
limit = apply_max_row_limit(query_params.get("queryLimit") or 0) # No second arg for max_limit

superset/utils/core.py
amaannawab923: Sure

superset/common/query_object_factory.py
amaannawab923: Sure

superset/utils/core.py
Antonio-RiveroMartnez: Oh ok ok, yeah, I'm OK with deleting if not being used

Antonio-RiveroMartnez (commented):
LGTM, I have seen the functionality and the comments were addressed

superset/config.py
```diff
@@ -969,6 +969,10 @@ class D3TimeFormat(TypedDict, total=False):
 # Maximum number of rows returned for any analytical database query
 SQL_MAX_ROW = 100000
 
+# Maximum number of rows for any query without Server Pagination in Table Viz type
```
betodealmeida, line 972:
```suggestion
# Maximum number of rows for any query with Server Pagination in Table Viz type
```

superset/common/query_object_factory.py
```diff
@@ -96,14 +102,27 @@ def _process_extras(
         return extras
 
     def _process_row_limit(
-        self, row_limit: int | None, result_type: ChartDataResultType
+        self,
+        row_limit: int | None,
+        result_type: ChartDataResultType,
+        server_pagination: bool | None = None,
```
betodealmeida, line 108: ❤️

betodealmeida (approved):
Looks great! Left a minor correction of a typo.

superset-frontend/plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -127,7 +168,8 @@ export default typedMemo(function DataTable<D extends object>({
     ...initialState_,
     // zero length means all pages, the `usePagination` plugin does not
     // understand pageSize = 0
-    sortBy: sortByRef.current,
+    // sortBy: sortByRef.current,
+    sortBy: serverPagination ? sortByFromParent : sortByRef.current,
```
amaannawab923, line 172: sortByFromParent is the sortBy state from the persistent filterState of the table chart which we need on initial Render as the table loses all of its variables on complete re render... So getting the sortBy state from redux helps maintain consistency

superset-frontend/plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -198,10 +246,46 @@ export default typedMemo(function DataTable<D extends object>({
...
+          };
+
+          handleSortByChange([sortByWithColumnKey]);
+        }
+      } else {
+        handleSortByChange([]);
+      }
+    }
+  }, [sortBy]);
+
```
amaannawab923, line 288: useEffect for updating the sortBy state in redux of the table viz & triggering query to get the latest data based on sortBy

superset-frontend/plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -381,16 +466,31 @@ export default typedMemo(function DataTable<D extends object>({
...
+                {serverPagination && (
+                  <div className="search-select-container">
+                    <span className="search-by-label">Search by: </span>
+                    <SearchSelectDropdown
+                      searchOptions={searchOptions}
+                      value={serverPaginationData?.searchColumn || ''}
+                      onChange={onSearchColChange}
+                    />
+                  </div>
+                )}
```
amaannawab923, line 479: The search select dropdown component to have dynamic search column

superset-frontend/plugins/plugin-chart-table/src/DataTable/components/GlobalFilter.tsx
```diff
@@ -66,17 +92,37 @@ export default (memo as <T>(fn: T) => T)(function GlobalFilter<
...
+  // Preserve focus during server-side filtering to maintain a better user experience
+  useEffect(() => {
+    if (
+      serverPagination &&
+      isSearchFocused.get(id) &&
+      document.activeElement !== inputRef.current
+    ) {
+      inputRef.current?.focus();
+    }
+  }, [value, serverPagination]);
```
amaannawab923, line 104: There was a problem that i faced with losing the focus in the search field ... as the user types in search field it triggers a complete re render & the search field focus is lost... Hence we are preserving the focus with the help of a map

superset-frontend/plugins/plugin-chart-table/src/DataTable/components/GlobalFilter.tsx
```diff
@@ -66,17 +92,37 @@ export default (memo as <T>(fn: T) => T)(function GlobalFilter<
...
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const target = e.target as HTMLInputElement;
+    e.preventDefault();
+    isSearchFocused.set(id, true);
+    setValue(target.value);
+  };
+
+  const handleBlur = () => {
+    isSearchFocused.set(id, false);
+  };
```
amaannawab923, line 115: Updating the map to inform that the field has lost focus ... and there is no need to focus it on initial render

rusackas:
Merging!

PR #33364, opened 2025-05-05
fix(deckgl): fix deckgl multiple layers chart filter and viewport

### LINKED ISSUE #13731: Visualization Type : deck.gl Multiple Layers - Query filters are not executing
I am not able to use Query->Filters while using the visualization type deck.gl Multiple Layers. 
To be more specific the filtera are not getting applied after adding the values too.
Superset version 0.36.0
superset

syedbarimanjan:
### SUMMARY
Filters werent applying to multiple layers deck.gl chart because they werent being passed down to the sublayers properly. Fixes #13731 

### Before
[430684705-7217504e-9ae2-43f4-821d-a03e9fd77a0b.webm](https://github.com/user-attachments/assets/aebbc52a-11e7-46a7-8cbf-72f38cadb235)

### After
[Screencast_20250501_203725.webm](https://github.com/user-attachments/assets/063f681d-661a-40b8-a139-155a261e473e)

### DIFF
--- a/superset/viz.py
+++ b/superset/viz.py
@@ -1532,7 +1532,31 @@ class DeckGLMultiLayer(BaseViz):
 
         slice_ids = self.form_data.get("deck_slices")
         slices = db.session.query(Slice).filter(Slice.id.in_(slice_ids)).all()
+
+        features: dict[str, list[Any]] = {}
+
+        for slc in slices:
+            form_data = slc.form_data
+
+            form_data["extra_filters"] = self.form_data.get("extra_filters", [])
+            form_data["extra_form_data"] = self.form_data.get("extra_form_data", {})
+            form_data["adhoc_filters"] = self.form_data.get("adhoc_filters")
+
+            viz_type_name = form_data.get("viz_type")
+            viz_class = viz_types.get(viz_type_name)
+            if not viz_class:
+                continue  # skip unknown viz types
+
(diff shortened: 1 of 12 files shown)


superset/viz.py
```diff
@@ -1532,7 +1532,31 @@ def get_data(self, df: pd.DataFrame) -> VizData:
 
         slice_ids = self.form_data.get("deck_slices")
         slices = db.session.query(Slice).filter(Slice.id.in_(slice_ids)).all()
+
+        features: dict[str, list[Any]] = {}
```
mistercrunch, line 1536: noting that we're trying to deprecate `viz.py` in favor of `/api/v1/chart/data`, though might be out-of-scope for this PR. I think some visualization like the deck.gl ones will need to migrate to the chart/data endpoint eventually. Pilling on more stuff onto viz.py might make it harder to migrate.

mistercrunch: Given this, it might inform doing this in JS if that's possible, and possibly hitting `/api/chart/data` N times (for each chart), and joining the data client-side.

rusackas:
Approving CI to run again 🤞

mistercrunch (approved):
All this complexity will need to get handled on the frontend once we deprecate `viz.py`. Looks like it should be deprecated already...

PR #33377, opened 2025-05-06
fix(native-filters): add column required validation for filter_select

arthi-arumugam-git:
### What does this PR do?

Adds a required field validation to ensure that users select a column when configuring a native filter of type `filter_select`.

### Bug Fix Summary

Previously, users could create a native filter of type `Value` (i.e., `filter_select`) without choosing a column. This caused runtime issues and invalid filters.

This PR fixes it by enforcing required validation on the column field in the filter configuration panel.

### Screenshot

image

### Test Plan

- Go to “Add and edit filters”
- Set filter type = “Value”
- Do not select column to validation error appears
- Select column to error disappears and filter can be saved

 Bug reproduced and resolved successfully.


mistercrunch:
PR isn't right, shows 215 commits. Please start from a recent commit on `master` as a base

rusackas:
Looks like a rebase went wrong here. You can open a fresh PR or try the rebase again.

rusackas:
This PR hasn't been updated in a long time and is in need of a rebase. We'll mark it as draft, but please set it as 'ready for review' when it's been brought up to speed.

PR #33392, opened 2025-05-08
fix(chart list): Facepile shows correct users when saving chart properties

### LINKED ISSUE #31592: In chart management, editing any chart properties may result in incorrect updates to the owner badge list
### Bug description

In chart management, editing any chart properties may result in incorrect updates to the owner badge list，It seems that the latest version 4.0 also has issues。

### Screenshots/recordings

de430b9ff19d4ddecc1fa6d4f6611cc
2e9eab87c2887e2f221e1ee2869efec

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

superset version： 3.1.3

### Checklist

ethan-l-geotab:
### SUMMARY

Looking to fix: https://github.com/apache/superset/issues/31592.
It's reproducible on all the test-env up environments. Add "Superset Admin" as user on a chart and then save shows the undefined undefined.

I just update the redux with the owner list back in it's original "shape".

first half is the broken, second half is working
fix-chart-list

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/31592
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/PropertiesModal/index.tsx
+++ b/superset-frontend/src/explore/components/PropertiesModal/index.tsx
@@ -192,33 +192,15 @@ function PropertiesModal({
     }
 
     try {
-      const res = await SupersetClient.put({
+      let res = await SupersetClient.put({
         endpoint: `/api/v1/chart/${slice.slice_id}`,
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify(payload),
       });
-      // update the redux state
-      // convert the tags to the format used in the selectOwners back to original redux format
-      const selectedOwnersArray = ensureIsArray(selectedOwners);
-      const newOwners = selectedOwnersArray.map((owner: any) => {
-        const nameParts = owner.label.split(' ');
-        const first_name = nameParts[0] || '';
-        const last_name =
-          nameParts.length > 1 ? nameParts.slice(1).join(' ') : '';
-        return {
-          id: owner.value,
-          first_name,
-          last_name,
-        };
+      res = await SupersetClient.get({
+        endpoint: `/api/v1/chart/${slice.slice_id}`,
       });
-      const updatedChart = {
-        ...payload,
-        ...res.json.result,
-        tags,
-        id: slice.slice_id,
-        owners: newOwners,
-      };
-      onSave(updatedChart);
+      onSave(res.json.result);
       addSuccessToast(t('Chart properties updated'));
       onHide();
     } catch (res) {
--- a/superset-frontend/src/explore/components/PropertiesModal/index.tsx
+++ b/superset-frontend/src/explore/components/PropertiesModal/index.tsx
@@ -192,13 +192,14 @@ function PropertiesModal({
     }
 
     try {
+      const chartEndpoint = `/api/v1/chart/${slice.slice_id}`;
       let res = await SupersetClient.put({
-        endpoint: `/api/v1/chart/${slice.slice_id}`,
+        endpoint: chartEndpoint,
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify(payload),
       });
       res = await SupersetClient.get({
-        endpoint: `/api/v1/chart/${slice.slice_id}`,
+        endpoint: chartEndpoint,
       });
       onSave(res.json.result);
       addSuccessToast(t('Chart properties updated'));
--- a/superset-frontend/src/explore/components/PropertiesModal/index.tsx
+++ b/superset-frontend/src/explore/components/PropertiesModal/index.tsx
@@ -201,11 +201,14 @@ function PropertiesModal({
       // convert the tags to the format used in the selectOwners back to original redux format
       const selectedOwnersArray = ensureIsArray(selectedOwners);
       const newOwners = selectedOwnersArray.map((owner: any) => {
-        const [first_name, ...last_name_parts] = owner.label.split(' ');
+        const nameParts = owner.label.split(' ');
+        const first_name = nameParts[0] || '';
+        const last_name =
+          nameParts.length > 1 ? nameParts.slice(1).join(' ') : '';
         return {
           id: owner.value,
           first_name,
-          last_name: last_name_parts.join(' '),
+          last_name,
         };
       });
       const updatedChart = {
--- a/superset-frontend/src/explore/components/PropertiesModal/index.tsx
+++ b/superset-frontend/src/explore/components/PropertiesModal/index.tsx
@@ -198,12 +198,22 @@ function PropertiesModal({
         body: JSON.stringify(payload),
       });
       // update the redux state
+      // convert the tags to the format used in the selectOwners back to original redux format
+      const selectedOwnersArray = ensureIsArray(selectedOwners);
+      const newOwners = selectedOwnersArray.map((owner: any) => {
+        const [first_name, ...last_name_parts] = owner.label.split(' ');
+        return {
+          id: owner.value,
+          first_name,
+          last_name: last_name_parts.join(' '),
+        };
+      });
       const updatedChart = {
         ...payload,
         ...res.json.result,
         tags,
         id: slice.slice_id,
-        owners: selectedOwners,
+        owners: newOwners,
       };
       onSave(updatedChart);
       addSuccessToast(t('Chart properties updated'));
--- a/superset-frontend/src/explore/components/PropertiesModal/PropertiesModal.test.tsx
+++ b/superset-frontend/src/explore/components/PropertiesModal/PropertiesModal.test.tsx
@@ -86,6 +86,12 @@ fetchMock.get('glob:*/api/v1/chart/318*', {
           type: 1,
         },
       ],
+      show_title: 'Show Slice',
+      certification_details: 'Test certification details',
+      certified_by: 'Test certified by',
+      description: 'Test description',
+      cache_timeout: '1000',
+      slice_name: 'Test chart new name',
     },
     show_columns: [
       'owners.id',


ethan-l-geotab:
/korbit-review.

superset-frontend/src/explore/components/PropertiesModal/index.tsx
```diff
@@ -185,12 +185,24 @@ function PropertiesModal({
...
+      const newOwners = selectedOwnersArray.map((owner: any) => {
+        const nameParts = owner.label.split(' ');
+        const first_name = nameParts[0] || '';
+        const last_name = nameParts.length > 1 ? nameParts.slice(1).join(' ') : '';
+        return {
+          id: owner.value,
+          first_name,
+          last_name,
+        };
+      });
```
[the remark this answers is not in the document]
ethan-l-geotab: I think it's a pretty specific case where this one component handles owners in a specific way which when trying to update redux, the "shape" of the object is different, so i have to translate it back to the same shape which originally existed.

superset-frontend/src/explore/components/PropertiesModal/index.tsx
```diff
@@ -185,12 +185,24 @@ function PropertiesModal({
         body: JSON.stringify(payload),
       });
       // update the redux state
+      // convert the tags to the format used in the selectOwners back to original redux format
+      const selectedOwnersArray = ensureIsArray(selectedOwners);
+      const newOwners = selectedOwnersArray.map((owner: any) => {
+        const nameParts = owner.label.split(' ');
+        const first_name = nameParts[0] || '';
+        const last_name = nameParts.length > 1 ? nameParts.slice(1).join(' ') : '';
```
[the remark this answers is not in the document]
rusackas: This is probably OK, but I'm _slightly_ nervous about things like middle names. This would mean if we have Michael J. Fox sign up, it'd be "Michael" "J. Fox" instead of "Michael J." "Fox."

Is there other first/last name logic currently in place, or is this a new concept?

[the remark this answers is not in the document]
ethan-l-geotab: I thought about that too actually. "Updating" only happens around user ID's. As for display, I'm on the same line as you where it's "probably okay". 

There is no real existing "full_name" split. I tried checking out how dashboards/dataset list handles it, but they're both different :(. 

I dont think there is a true way of doing it unless we retrieve it from the backend.

I have tried doing something like this in the backend to "pass ownership data" along with the put request 
and then just removing the modification to the owner field and using just the returned owner from the put request, but I'm not sure that's the best design either...
image
image
 
I've also tried seeing if I could sort of "cache" the load options and see if it was split there when im getting the eligible users to add as owners, but that won't work either because it's only `item.text` and makes no differentiation of first and last names.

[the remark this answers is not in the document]
rusackas: > There is no real existing "full_name" split. I tried checking out how dashboards/dataset list handles it, but they're both different :(.

This is what I was worried about. If there's any logic to be DRY'd up (moved to a util or something) that would be preferable than implementing another bit of inline logic.

ethan-l-geotab:
@rusackas Hello, I went with a much more simple approach I saw use by datasets where i just make 1 get request to the resource and then update from the resource from the put request.

superset-frontend/src/explore/components/PropertiesModal/index.tsx
```diff
@@ -179,20 +179,15 @@ function PropertiesModal({
...
-        ...res.json.result,
-        tags,
-        id: slice.slice_id,
-        owners: selectedOwners,
-      };
-      onSave(updatedChart);
+      res = await SupersetClient.get({
+        endpoint: `/api/v1/chart/${slice.slice_id}`,
+      });
+      onSave(res.json.result);
```
ethan-l-geotab: @rusackas so i was thinking about this again, and noticed that for datasets, there are actually multiple get requests. I was wondering if it was just better to return the data inside the put request as a response, or do it in a second get request.

ethan-l-geotab:
Hey, can i get any updates on this?

JZ6:
@rusackas we are running into this issue too, can this pr be merged, thanks

superset-frontend/src/explore/components/PropertiesModal/index.tsx
rusackas: I don't have anything against an extra call, as long as it's not something that's going to happen a million times, like in a CRUD view (for obvious performance reasons). 

@michael-s-molina @justinpark @mistercrunch can provide a better second opinion here, perhaps.

rusackas:
Hey! Sorry I went AWOL on the thread here :) I think we can likely merge this, pending a second opinion on the extra API call (probably not an issue). The main blocker at the moment is that there's a rebase needed, which I hope isn't too painful with the changes that have taken place in the tests.

superset-frontend/src/explore/components/PropertiesModal/index.tsx
mistercrunch: shouldn't the PUT return the new object?

mistercrunch: @korbit-ai what do you think? REST api best practices on a PUT, return the new committed object?

mistercrunch: or @Copilot WDYT?

mistercrunch: how do you summon these things!? Which we had our buddy @claude in this repo ...

JZ6:
> Hey! Sorry I went AWOL on the thread here :) I think we can likely merge this, pending a second opinion on the extra API call (probably not an issue). The main blocker at the moment is that there's a rebase needed, which I hope isn't too painful with the changes that have taken place in the tests.

should be rebased now

superset-frontend/src/explore/components/PropertiesModal/index.tsx
rusackas: Hmmm... I thought @korbit-ai should do it. Maybe @copilot or @korbit or @dosu. C'mon bots!

rusackas: @ethan-l-geotab I see that CI is looking good again after your rebase, thanks! I think this part of the thread is the crux of whether we merge it now or not. Is getting/using the response from the PUT call doable? If so, that does seem like the more elegant approach, but if not, I have no personal objection to fixing the bug with the current approach.

rusackas:
Looks technically mergeable. [This](https://github.com/apache/superset/pull/33392#discussion_r2288913534) (or the [bot version](https://github.com/apache/superset/pull/33392#discussion_r2263820617) thereof) seems to be the last standing question.

superset-frontend/src/explore/components/PropertiesModal/index.tsx
ethan-l-geotab: @rusackas It's doable. I'm just worried theres other part of the code that "require" the specific return from the put request.

superset-frontend/src/explore/components/PropertiesModal/index.tsx
```diff
@@ -192,20 +192,15 @@ function PropertiesModal({
...
-      const updatedChart = {
-        ...payload,
-        ...res.json.result,
-        tags,
-        id: slice.slice_id,
-        owners: selectedOwners,
-      };
-      onSave(updatedChart);
+      res = await SupersetClient.get({
+        endpoint: `/api/v1/chart/${slice.slice_id}`,
```
mistercrunch: DRY up that endpoint top-of try block

superset-frontend/src/explore/components/PropertiesModal/index.tsx
mistercrunch: are you sure the PUT doesn't actually returns the payload already? looking at https://superset.apache.org/docs/api it seems like it might

ethan-l-geotab: Kinda.
put returns:
image
get returns:
image

I think cuz schema is int instead of relationship. I just am not sure if theres other parts of the code that expects the owner list to be returned as a list of ints.

mistercrunch: related: https://chatgpt.com/share/68a90a13-5390-8010-b149-6a87d0014de3

Personally would say: got for it, make it such that all 3 PUT/POST/GET all return the same schema as per best practices. If something break, a test should fail or we'll know soon enough.

Might be good to grep/search or add Claude Code or AI to do a deep search in the repo to confirm.

mistercrunch: Higher level solution I've been meaning to push forward: the superset client should be dynamically generated from the openapi spec, and be fully typed. It would mean that typescript would then be able to catch this kind of stuff at compile time. Meaning as you change the openapi spec to return this consistent schema across PUT/POST/GET, it would regenerate a fully typed client, and `tsc` would tell you exactly which calls break. Fun project if someone wants to take this on!

superset-frontend/src/explore/components/PropertiesModal/index.tsx
ethan-l-geotab: Hey, sorry I'm not sure I'm catching the meaning of this.

rusackas: DRY = "don't repeat yourself" - i.e. set a variable for the endpoint and use it twice rather than repeating the string :D

mistercrunch:
Need help figuring out how to get the PUT to return the same object as the GET?

EnxDev:
Hi @ethan-l-geotab, let me know if you need any help with completing this PR

ethan-l-geotab:
Sorry, got distracted with a lot of other stuff. Yes please, any help would be nice, could someone let me know what i have to do to modify whatever schemas  or anything else i'd have to do?

legal-s:
Hi, 

Have you got news for merge this pull request ? 

We need this correction.

Thank you

rusackas (approved):
Seems like there are more reasons to merge this than not to, at this point. In it goes!

PR #33410, opened 2025-05-12
chore: rip antd-v4 from the app!

mistercrunch:
Finally getting to a place where we can strip all references to antd v4 from the repo.


mistercrunch:
Target branch `template_less` changed and it's not reflected in the PR, re-opening as a new PR

PR #33411, opened 2025-05-12
chore: rip antd-v4 from the app!

mistercrunch:
Finally getting to a place where we can strip all references to antd v4 from the repo.


mistercrunch:
Hitting some significant issues here around a couple libraries:
- `react-js-cron@2.1.2`: the version we use uses antd-v4 internally, and trying to bump it to a future version that supports antd-v5 creates some issues
- `geostyler@12.0.2`: also has some intricate issues related to antd-v4

Unclear what the solution is, but would love to get some help as I got dead-ended trying to push this through.

rusackas:
This is the best kind of PR ❤️

.github/workflows/superset-e2e.yml
```diff
@@ -73,6 +73,7 @@ jobs:
         with:
           persist-credentials: false
           submodules: recursive
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
```
mistercrunch, line 76: nastiest github bug I've ever seen, `actions/checkout` somehow sometimes decides to checkout random commit -> https://github.com/actions/checkout/issues/299#issuecomment-2881779075

superset-frontend/plugins/plugin-chart-cartodiagram/src/util/controlPanelUtil.tsx
```diff
@@ -57,7 +57,7 @@ export const getLayerConfig = (controlPanel: ControlPanelConfig) => {
  */
 export const selectedChartMutator = (
   response: Record<string, any>,
-  value: SelectValue | undefined,
+  value: any,
```
kgabryje, line 60: Can we use a type here?

msyavuz (commented):
Nice! Maybe we can also get rid of the antd5 prefix while we are at it? Repo wide sed for antd5 → antd?

mistercrunch:
> get rid of the antd5 prefix while we are at it?

yup let me add a commit, though I'm wondering if it would make sense to namespace it elsewhere, like `superant-` or similar to namespace things, and potentially allow app-makers (presumably also using antd in the host app) to target only Superset with this selector, though now I'm thinking that app builder probably want the same namespace in most cases to affect the whole app... 

Alright, pushing a commit to remove the `prefixCls` and roll with the default

superset-frontend/plugins/plugin-chart-cartodiagram/src/util/controlPanelUtil.tsx
mistercrunch: I should have had a TODO here and other similar places, but the plan is to import all antd-related components from `@superset-ui/core/components` and avoid the `antd` peerDep altogether. I'm happy fast-following with that once the SIP is out. 

The work is going to be easier then with the single flat namespace with all the components and related types.

kgabryje (approved):
Looks great!

msyavuz (approved):
Even nicer!

mistercrunch:
oh wasn't passing CI upon merge but happy to merge things!

PR #33412, opened 2025-05-12
fix: loading examples in CI returns http error "too many requests"

mistercrunch:
Recently in https://github.com/apache/superset/pull/33354 I tried to address http 429 "too many attempts that we started getting in CI. This PR adds a bit of custom logic with retries and expodential backoffs to work around the issue. Ideally we'd have another approach using a proper CDN that supports our needs or use a github repo to clone locally, but this should help relieve the issue.

Regardless of the approach, having a central function `examples.helpers.read_example_csv` is a step in the right direction as it centralizes the logic used to load the csv into a single DRY place.

### DIFF
--- a/superset/examples/bart_lines.py
+++ b/superset/examples/bart_lines.py
@@ -16,7 +16,6 @@
 # under the License.
 import logging
 
-import pandas as pd
 import polyline
 from sqlalchemy import inspect, String, Text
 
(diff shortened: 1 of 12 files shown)


msyavuz (approved):

michael-s-molina:
@mistercrunch Why do we need this follow-up? Didn't the previous one fix the issue? I'm just asking to know if I need to cherry-pick this into 5.0 like I did with the previous one.

mistercrunch:
seems it helped - went from seeing a lot of 429 in CI to much fewer, but there are still issues with it. Not sure if this fix fully fixes things either, working on yet another follow up.

PR #33420, opened 2025-05-13
feat(theming): implement a ThemeController

msyavuz:
### SUMMARY

Implement a ThemeController that handles theme injection and persistence that can be overriden for different methods of storage. This pr also includes a ThemeSelect component that loads prefers-dark-scheme by default and persists dark/light/system.

After:
image

### TESTING INSTRUCTIONS

Enable the new `THEME_ENABLE_DARK_THEME_SWITCH` feature flag.

Select from Light/Dark/System and refresh the page. Theme mode should persist.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/packages/superset-ui-chart-controls/test/components/ColumnOption.test.tsx
```diff
@@ -53,9 +53,9 @@ const defaultProps: ColumnOptionProps = {
 
 const setup = (props: Partial<ColumnOptionProps> = {}) =>
   render(
-    <ThemeProvider theme={supersetTheme}>
+    <EmotionThemeProvider theme={supersetTheme}>
```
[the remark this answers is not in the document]
msyavuz: Actually it already wraps it by default in `spec/helpers/testing-library.tsx` but using it would mean adding superset as a dependency to each package that needs it. Maybe instead we can lift that to @superset-ui/core as well for theme provider and extend it in frontend app to have useDnd, useRedux etc.? cc @mistercrunch

superset-frontend/src/features/home/Menu.test.tsx
```diff
@@ -255,6 +255,7 @@ test('should render', async () => {
     useRedux: true,
     useQueryParams: true,
     useRouter: true,
+    useTheme: true,
```
mistercrunch, line 258: wondering if `true` should be the default arg in the spec/utils render function (?) Might add extra work/overhead on the test suite but it's nice to not have to worry and be explicit about it. What do you think?

mistercrunch: oh also during my refactor I noticed some specific tests were doing their own theme provider wrapping, we should do this consistently in the `render` function

msyavuz: > wondering if true should be the default arg in the spec/utils render function (?) Might add extra work/overhead on the test suite but it's nice to not have to worry and be explicit about it. What do you think?

I thought about it but for now it is only needed in two places and overhead for the whole testing suite felt a bit too much to do this.

msyavuz: > oh also during my refactor I noticed some specific tests were doing their own theme provider wrapping, we should do this consistently in the render function

There is some funkiness around accessing spec/helpers in packages and some places it is necessary to have it like `src/SqlLab/components/ShareSqlLabQuery/ShareSqlLabQuery.test.tsx`

mistercrunch: Oh not sure if you're reporting this pre-merge or post-merge of my component relocation, but I untangled a lot of this in the PR I merged. Generally there's a new spec/helper for core in `packages/superset-ui-core/src/spec/` and one for the main app.

mistercrunch: In the new one, I didn't bother implementing a `useTheme` but could be good depending on overhead. Trying to think about measuring overhead, but maybe setting "false expect where needed" and comparing the cpu cycles with or without might be worthwhile.

superset-frontend/packages/superset-ui-core/src/components/ThemeEditor/index.tsx
```diff
@@ -63,7 +68,7 @@ const ThemeEditor: React.FC<ThemeEditorProps> = ({
     try {
       const parsedTheme = JSON.parse(jsonMetadata);
       console.log('Parsed theme:', parsedTheme);
```
geido, line 70:
```suggestion
```

EnxDev (commented):
@msyavuz  I didn't see the icon to choose the mode

Screenshot 2025-06-16 143757

msyavuz:
> @msyavuz I didn't see the icon to choose the mode
> 
> Screenshot 2025-06-16 143757

You need to have the feature flag for it to show up. Let me add it to testing instructions!

EnxDev (approved):
LGTM!

Issue #30075, opened 2024-08-30
[SIP-145] Proposal for Embeddable Charts using the same mechanism as #17187


jayakrishnankk:
### Motivation

The existing method of embedding charts is using iframe method.
What this means is that the user have to be logged-in in parallel to the superset system for the chart to display without a login screen in the embedded mode.

Similar to embedding dashboards as outlined in [SIP-75](https://github.com/apache/superset/issues/17187), superset dashboard developers would like to embed charts in a similar fashion so that the login page does not appear and a parallel login is not needed by the end user, avoiding the user to have an account created in superset to view the charts.

### Proposed Change

This SIP is aimed to implement the Embedded Charts capability using the same GuestToken method that the Embeddable Dashboards are using.

### New or Changed Public Interfaces

Added REST endpoints
```
GET    {host}/api/v1/chart/{pk}/embedded
POST   {host}/api/v1/chart/{pk}/embedded
DELETE {host}/api/v1/chart/{pk}/embedded
```
Added Views

In addition to existing view {host}/embedded/{uuid}, added the following
```
{host}/embedded/dashboard/{uuid} 
{host}/embedded/chart/{uuid}
```

NOTE: {host}/embedded/dashboard/{uuid}  works the same way as {host}/embedded/{uuid}, which should be deprecated and removed in one of the subsequent versions.

### New dependencies

None

### Migration Plan and Compatibility

Needs a migration. Please see the attached merge request to see the migration details.
A new table called `embedded_charts` will be added and a foreign key reference will be made to the `slices` table.

### Rejected Alternatives

Did not discuss alternate approaches

Michila0:
Is this done

rusackas:
> Is this done

Not even close... the proposal has been numbered, but has not yet been put up for a discussion on the ASF list, which then would lead to a vote, which then would open the door for the work to be completed/merged.

@Michila0 do you have thoughts on the proposal you'd like to add to carry the discussion forward?

Michila0:
ASF list? I like contribute with you

rusackas:
Discussion opened! Feedback welcome, and anyone volunteering to help contribute to this should chime in on the linked PR and/or reach out to @jayakrishnankk on [slack](http://bit.ly/join-superset-slack) or other means.

rusackas:
I'll open the vote on this, if everyone agrees it's ready.

villebro:
[...]

PR #33424, opened 2025-05-13
feat(Embedding): Embeddable Charts as per SIP 

amaannawab923:
### SUMMARY

This pr implements Embeddable Charts as per SIP https://github.com/apache/superset/issues/30075#issue-2498101831
The only takeaway , the solution doesnt use charts in standalone mode , instead it takes the dashboard charts, hydrates them with all the required data & tells them to take full width & height of iframe , this avoids use of standalone mode & gives the same experience as dashboard charts 

Chart embedding in action with nextjs app 

[screen-capture (23).webm](https://github.com/user-attachments/assets/557e93ee-15c2-4618-8f82-d8ee42736f17)

Embed uuid generation just like dashboard
chart_embed

Screenshot 2025-05-14 at 1 13 28 AM

Most of normal chart features of dashboard works

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
I commented on the SIP back in October, https://github.com/apache/superset/issues/30075#issuecomment-2442801379, but why not embedding a single-chart-dashboard?

There's a related idea which is allowing to embed parts of the Explore experience, potentaly allowing embedded use cases for light data exploration. Wondering if that's in-scope here (?) The idea would be to allow/limit part of the Explore experience, say I may want to hide/show different panels and/or controls... I'm guessing the idea here is more to simply embed a single chart, without interactions.

amaannawab923:
> I commented on the SIP back in October, [#30075 (comment)](https://github.com/apache/superset/issues/30075#issuecomment-2442801379), but why not embedding a single-chart-dashboard?
> 
> There's a related idea which is allowing to embed parts of the Explore experience, potentaly allowing embedded use cases for light data exploration. Wondering if that's in-scope here (?) The idea would be to allow/limit part of the Explore experience, say I may want to hide/show different panels and/or controls... I'm guessing the idea here is more to simply embed a single chart, without interactions.

This pr deals with embedding a single chart with normal dashboard interactons

What i understood from the previous comment of the sip was https://github.com/apache/superset/issues/30075#issuecomment-2442801379 we need to have in dashboard chart creation ... I have implemented the same in a private fork , we used to call it easy charts

1.User can place a chart container on the grid <- completely empty which will have a button  +add chart on it 
2. It opens up a control panel in a popover with 3 controls , Groupby , metric & dataset  & a top bar for selecting viz type
3. When the user saves the chart ... it gets saved to slices & gets absorbed by the chart container & the easy charts default attributes can also be edited via the chartlist
This allowed for flexibility for creating charts on the fly & more customisations via proper control panel if the end user wants to 

We can connect on this , if easy charts are implemented in a seperate sip , we can bring those anytime here in embedded charts

@mistercrunch

amaannawab923:
Screenshot 2025-05-15 at 2 16 01 AM

Something like this @mistercrunch , Some level of explore controls in dashboard for the chart

Or something where its metric centric & the user can select metrics & create instant charts within dashboard like below
Screenshot 2025-05-15 at 2 21 05 AM

Or something where the user can just change the query groupby just by hitting a dropdown & selecting from a list of Groupbys like Display by departments like the below 
Screenshot 2025-05-15 at 2 21 22 AM

Curious to know , whats the idea will fit best with superset so we can implement it 

@mistercrunch

mistercrunch:
Complex topic. We should probably talk.

Some thoughts related to all this:

1. it would be really great to be able to "edit chart", or "explore data" within dashboard. Idea here is that within a dashboard, if you "edit chart" or "explore chart", we pop a large Modal with what is currently in the explore page, while keeping the user in the dashboard. Here there needs to be different options depending on the user level of access (do they have access to edit/modify the chart?). If we allow them to explore the data, what controls do we expose?  Metric? Filters? Groupby? Filters? All? In the context of embedded, the person embedding needs to be able to define what it allows the user to do, if anything, within that modal.
2. it would be great to be able to embed a single chart, yes, and offer some options. Which controls do you want to expose? Do you allow the user to switch the metric? apply filters? ...
3. it would be great to allow "embedders" to externally control the chart, so that the host application can emit events inside the embedded chart, things like "group by something else", "switch the metric to this other metric", "apply this filter", ...

amaannawab923:
> Complex topic. We should probably talk.
> 
> Some thoughts related to all this:
> 
> 1. it would be really great to be able to "edit chart", or "explore data" within a dashboard. Idea here is that within a dashboard, if you "edit chart" or "explore chart", we pop a large Modal with what is currently in the explore page, while keeping the user in the dashboard. Here there needs to be different options depending on the user level of access (do they have access to edit/modify the chart?). If we allow them to explore the data, what controls do we expose?  Metric? Filters? Groupby? Filters? All? In the context of embedded, the person embedding needs to be able to define what it allows the user to do, if anything, within that modal.
> 2. it would be great to be able to embed a single chart, yes, and offer some options. Which controls do you want to expose? Do you allow the user to switch the metric? apply filters? ...
> 3. it would be great to allow "embedders" to externally control the chart, so that the host application can emit events inside the embedded chart, things like "group by something else", "switch the metric to this other metric", "apply this filter", ...

Yes , i somewhat got the idea , we can connect on this , 

https://github.com/apache/superset/issues/33448 

i have created an sip for the same , would include points from the ones mentioned above
  @mistercrunch

rusackas:
This needs a big rebase... do we still want to carry this through? The SIP is already approved, so it just needs to be completed, or the SIP may (eventually) be discarded due to inactivity.

rusackas:
The SIP's approved, so this is worth landing, but it needs a sizable rebase and the tests are red — and the design conversation @mistercrunch raised never quite resolved. Are you up for picking it back up? Let me know if you want help with this one.

rusackas:
@amaannawab923 the SIP's approved so this is worth landing, but it needs a sizable rebase, the tests are red, and @mistercrunch's design question never resolved. Are you up for picking it back up? If not I'm willing to push some commits, rebase, or somehow try to move the ball forward here if this is still on the roadmap.

gabotorresruiz (changes requested):
Hi @amaannawab923, thanks for the work here, and sorry it took this long for a committer pass. The SIP is approved and I would genuinely like to see this land, so here is an honest assessment of where the branch stands. I built and ran this branch locally to verify the points below.

This block worries me the most: as far as I can tell the guest token system never learns about charts, and I could reproduce that live on this branch. `GuestTokenResourceType` still only contains `DASHBOARD` (superset/security/guest_token.py), so `POST /api/v1/security/guest_token/` with `{"type": "chart"}` returns 400 with "Must be one of: dashboard." on a running instance. And the new embedded chart page's first data call is `GET /api/v1/explore/` (useExploreData.ts), which is protected under the Explore permission: with a valid dashboard guest token that request returns 403, while the same request as an authenticated admin returns 200. The `/embedded/<uuid>` page itself renders fine for an anonymous request, but the chart inside it has no authorized way to load data as a true guest, which is the core promise of SIP-145. Was the demo running against a browser session that was already logged in? Or am I missing where the guest token learns about chart resources? If I am right, a revived version needs a `CHART` resource type plus matching `raise_for_access` support in the security manager, with a functional test that mints a chart guest token and asserts the data endpoints honor it (and reject a token scoped to a different chart).

Two smaller things I verified on this branch:
- superset/config.py flips the `EMBEDDED_SUPERSET` default to `True`. That changes the shipped security posture of every deployment and looks like a dev leftover; it needs reverting.
- Just a small NIT for the revival: `with_chart` in superset/charts/api.py catches `ChartForbiddenError`, but nothing in that path raises it. Access control currently rests entirely on `ChartFilter` inside `ChartDAO.find_by_id`, which does work (I verified a Gamma user gets a 404 for a chart outside their access), but I would mirror the dashboard pattern and call `raise_for_access` explicitly the way `DashboardDAO.get_by_id_or_slug` does, so the two embedded surfaces stay symmetric.

On mechanics: I test-merged current master into this branch and 19 of the 30 changed files conflict, with master roughly 5,300 commits ahead. The embedded surface has also gained security hardening since (postMessage origin enforcement in #40629, a Sec-Fetch-Dest check in #40667) that the chart path must inherit. Realistically this is a fresh branch on top of master rather than a rebase, and the new endpoints, model, and migration will need backend tests, since none are included yet. On the plus side, the migration itself applied cleanly when I stood the branch up.

None of this diminishes the value of the SIP, and the frontend approach of hydrating a dashboard-style chart instead of standalone mode is a neat idea. If you are still interested, I am happy to dig in with you on the guest token design, since that is the part I would want settled before the UI work. If you no longer have bandwidth, no hard feelings at all, and per @rusackas's earlier offer we can look at moving it forward from here.

PR #33425, opened 2025-05-13
fix(table-chart): time shift is not working

justinpark:
### SUMMARY
In situations where a floating number is even and the decimal number is large, there is an issue with JSONbig parse converting it to BigNumber instead of a number. This leads to a problem in the process of checking [isMetric](https://github.com/apache/superset/blob/7f14e434c87d3c3ffe403ecf84d10f1ca7686446/superset-frontend/plugins/plugin-chart-table/src/transformProps.ts#L234), as it recognizes that the data contains items that are not of type number, resulting in incorrect output of the time shift values.

The code addresses this issue by converting the parse section to a number.

Before:
Screenshot 2025-05-13 at 1 40 13 PM

After:
Screenshot 2025-05-13 at 1 36 44 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/connection/callApi/parseResponse.ts
+++ b/superset-frontend/packages/superset-ui-core/src/connection/callApi/parseResponse.ts
@@ -59,8 +59,9 @@ export default async function parseResponse<T extends ParseMethod = 'json'>(
       response,
       json: cloneDeepWith(json, (value: any) => {
         if (
-          value?.isGreaterThan?.(Number.MAX_SAFE_INTEGER) ||
-          value?.isLessThan?.(Number.MIN_SAFE_INTEGER)
+          value?.isInteger?.() === true &&
+          (value?.isGreaterThan?.(Number.MAX_SAFE_INTEGER) ||
+            value?.isLessThan?.(Number.MIN_SAFE_INTEGER))
         ) {
           return BigInt(value);
         }
--- a/superset-frontend/packages/superset-ui-core/src/connection/callApi/parseResponse.ts
+++ b/superset-frontend/packages/superset-ui-core/src/connection/callApi/parseResponse.ts
@@ -58,16 +58,16 @@ export default async function parseResponse<T extends ParseMethod = 'json'>(
     const result: JsonResponse = {
       response,
       json: cloneDeepWith(json, (value: any) => {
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM . Wow, fix has been merged in 2021 ... maybe 1.1 will never happens.

superset-frontend/packages/superset-ui-core/src/connection/callApi/parseResponse.ts
```diff
@@ -58,17 +58,17 @@ export default async function parseResponse<T extends ParseMethod = 'json'>(
...
-        }
         if (
           value?.isGreaterThan?.(Number.MAX_SAFE_INTEGER) ||
           value?.isLessThan?.(Number.MIN_SAFE_INTEGER)
         ) {
           return BigInt(value);
         }
+        // // `json-bigint` could not handle floats well, see sidorares/json-bigint#62
+        // // TODO: clean up after json-bigint>1.0.1 is released
+        if (value?.isNaN?.() === false) {
```
michael-s-molina, line 69: @justinpark Won't decimal numbers be truncated in case they are passed to the native Javascript `BigInt` object? This can happen when `value?.isGreaterThan?.(Number.MAX_SAFE_INTEGER) || value?.isLessThan?.(Number.MIN_SAFE_INTEGER)`. In other words, do we need to include a check for type integer?
```
if (
   value?.isInteger?.()  <-- check for integer
   value?.isGreaterThan?.(Number.MAX_SAFE_INTEGER) ||
   value?.isLessThan?.(Number.MIN_SAFE_INTEGER)
) {
   return BigInt(value);
}
```

Also, if that's the case, don't we also need to handle large decimals given that `toNumber` will truncate?

justinpark: Adding an integer check to BigInt is a good idea. I have added that part.

I think the truncation of large decimals is a trade-off. For example, when a value used as bigint is truncated, it affects the accuracy of the information related to the ID, so it is appropriate to keep all decimals by bigint format. However, in the case of large decimal floating numbers, since the values are within a range that allows arithmetic operations, I believe it is a part that can be compromised to support operations in such cases, even if some decimal can be truncated.

michael-s-molina (approved):

PR #33445, opened 2025-05-15
fix: remove all `less` styling!

mistercrunch:
Pilling onto https://github.com/apache/superset/pull/33411, plan here is to get CI to run and merge onto the Theming (`template_less`) branch

- REMOVED ALL .less  and less-related stuff
- REMOVED bootstrap  as a dependency (this breaks FAB views, but we're ditching those!)
- REMOVE font-awesome  as a dependency too
- CLEANED a lot of the .html  / jinja, backend generated stuff


mistercrunch:
Screenshot 2025-05-22 at 1 17 37 PM
!!!!!!

mistercrunch:
Thanks so much for all the help so far here!

michael-s-molina:
giphy

PR #33450, opened 2025-05-15
fix(chart): update geographical info for latvia

eriks47:
### SUMMARY

Updates latvia.geojson so that it matches the current administrative subdivison of Latvia, as the previous one reflects the regions before the 2021 subdivision reform. See https://en.wikipedia.org/wiki/Administrative_divisions_of_Latvia

**Before:**
Screenshot From 2025-05-15 12-40-49
Many regions are not colored, because the data doesn't match the ISO codes in latvia.geojson.

**After:**
Screenshot From 2025-05-15 18-08-15
The regions have correct borders, and everything gets colored correctly.

### TESTING INSTRUCTIONS

```
git clone https://github.com/eriks47/superset
cd superset
docker compose up --build
```
then create a map chart, select Latvia as the country, use mock data to see the changed
borders; the data I used (has proper territorial ISO codes):
[territorial_data_iso.csv](https://github.com/user-attachments/files/20230862/territorial_data_iso.csv)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Please make your changes in the Jupyter notebook, otherwise any GeoJSON changes will be overwritten by the Notebook. Here's some info on how to do it: https://preset.io/blog/how-and-when-and-why-to-update-apache-supersets-country-maps/

eriks47:
Made the changed to the notebook, now the notebook shows the updated map of Latvia both at the added section for Latvia, and in the "all countries" section. Hopefully the changes are correct, because I don't have any experience with Jupyter Notebooks :sweat_smile:.

I tested the changes with `docker compose up --build` and everything works as expected, here is a more clear screenshots, that shows a map with new subdivisions:
Screenshot From 2025-05-17 16-52-56

eriks47:
@rusackas, Hello there, should I close the PR, looks like it is not going to get merged?

rusackas:
Oof... sorry this has dragged on. I'd LOVE to merge it. 

I do see the jupyter nobeook files, but the results from running it (the geojson) aren't checked in here on the PR. Can you add those? If you just run the notebook, you can just add/push the resulting file changes here.

If you've already moved on, I can run it and commit those changes here (if indeed I can write to your fork).

eriks47:
Well, I updated latvia.geojson, however the notebook makes changes to several geojson files, would you like those changes as well?

rusackas:
Hmm... well, we _probably_ want those... they can be included here or they can be a separate PR. I'd wager the upstream data was changed, so those would be a more general/periodic update. I'd open the separate PR now, just for geojson hygeine, but then you'd probably have to rebase the notebook, and I don't want to cause further speedbumps here :)

rusackas (approved):

rusackas:
Thanks for your contribution! Merged this, but still open to reviewing/merging the rest of the changes if you feel like providing those too... or I will eventually :)

PR #20816, opened 2022-07-22
fix: chart empty state & result panel when multiple queries are executed display incorrectly

diegomedina248:
### SUMMARY

Depending on the number of queries being executed for a given chart, we are displaying the results in panels at the bottom (explore).
This causes some issues, because the "Show Totals" and pagination features add extra queries, leading, for instance, a table to display more than one result panel, which is incorrect.

The same issue appears for the chart empty state, since having multiple queries can make the state not to appear, as expected.

We currently have a `getQueryCount` util that is used in some (but not all) places. This PR moves that value to the chart metadata and then queries it in all the necessary places to solve the issue mentioned above.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

Before:

https://user-images.githubusercontent.com/17252075/180353269-f5d599ec-d7dd-4a2e-9a3a-d792847916fc.mov

After:

https://user-images.githubusercontent.com/17252075/180353372-60e2ff11-fcd1-4fa7-8c6e-2103d6aa4faa.mov

### TESTING INSTRUCTIONS

1. Create a chart table
2. Enable Show Totals
3. Update the chart

Ensure the results panel show only a single result.
Repeat the process above, but for a Mixed Timeseries chart. Ensure you can see two results panels.

For the chart above, add some filters that would yield no results.
Update the chart and ensure you can see the empty state.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #33451, opened 2025-05-15
feat(chart): add dynamicQueryObjectCount property to Chart Metadata

DamianPendrak:
### SUMMARY
This PR implements a `dynamicQueryObjectCount` property to Chart Metadata to allow Chart to have a dynamic number of queries based on Controls. Without this, the Chart will only show 1 query even if more queries are provided.

### TESTING INSTRUCTIONS
1. Use a Chart with more than one query
2. Add `dynamicQueryObjectCount: true` to a Chart Metadata
3. See if there is a correct number of Results tabs in Explore and in "View as table" in Dashboard

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/chart/models/ChartMetadata.ts
+++ b/superset-frontend/packages/superset-ui-core/src/chart/models/ChartMetadata.ts
@@ -49,6 +49,7 @@ export interface ChartMetadataConfig {
   label?: ChartLabel | null;
   labelExplanation?: string | null;
   queryObjectCount?: number;
+  dynamicQueryObjectCount?: boolean;
   parseMethod?: ParseMethod;
   // suppressContextMenu: true hides the default context menu for the chart.
   // This is useful for viz plugins that define their own context menu.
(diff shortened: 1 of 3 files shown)


rusackas:
We already have `queryObjectCount` so I'm not sure what the use case is for this additional field. I'm by no means against it, but would love to know when/where it'll be used. Could this _replace_ the `queryObjectCount` field?

Do you have screenshots before/after that we can refer to as well?

Thanks for contributing!

DamianPendrak:
Right, there is a `queryObjectCount` but it forces you to decide what will be the max number of queries in the Chart - then the queries are limited to this value. If it's set to a big number, then there are multiple empty result tabs even though there are no queries created by the Chart:

Screenshot 2025-05-15 at 20 02 45

With the `dynamicQueryObjectCount` it will display just one tab:

Screenshot 2025-05-15 at 20 04 39

Then, when any number of queries is created, it will show all of them:

Screenshot 2025-05-15 at 20 06 14

There is the PR that introduced the `queryObjectCount`: https://github.com/apache/superset/pull/20816
It looks like there is a use case for limiting the number of queries.

geido (approved):

Issue #33449, opened 2025-05-15
Pivot table cells not rendering html


richardfogaca:
### Description

The Pivot Table chart does not render HTML tags within its data cells. While HTML in row and column headers is rendered correctly, HTML content in the data cells is escaped and displayed as literal text.

This limits the ability to, for example, embed links or apply custom styling directly within the data portion of the pivot table when the underlying data source contains HTML.

### How to reproduce the bug

- This [chart export](https://github.com/user-attachments/files/20231045/chart_export_20250515T153644.zip) can be imported into Superset to directly observe the issue. But also possible to be reproduced following the steps below:

1.  Create or import a Pivot Table chart where the dataset provides HTML content for one or more cells.
2.  Configure the chart to use these HTML-containing fields.
3.  Observe the rendering of the data cells in the pivot table.

### Expected results

The HTML tags within the data cells should be rendered by the browser. For example:
*   `Capital of <strong style="color:blue;">France</strong>` should display as: Capital of **France** (with "France" in blue and bold).
*   `The <a href="https://en.wikipedia.org/wiki/New_York_City" target="_blank">Big Apple</a>` should display as: The Big Apple (with "Big Apple" as a clickable link).

### Actual results

The HTML tags within the data cells are displayed as literal text. For example:
*   You see: `Capital of <strong style="color:blue;">France</strong>`
*   You see: `The <a href="https://en.wikipedia.org/wiki/New_York_City" target="_blank">Big Apple</a>`

### Screenshot

<img width="1480" alt="Image" src="https://github.com/user-attachments/assets/9e3d97cb-835f-4c57-bbd6-e43f72f98abb" />

### Superset version

master / latest-dev

### Python version

3.10

### Node version

18 or greater

### Browser

Chrome

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [x] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

msyavuz:
Fixed with https://github.com/apache/superset/pull/34351

PR #33452, opened 2025-05-15
fix(pivot-table): apply safeHtmlSpan for cells in the pivot table chart

richardfogaca:
### SUMMARY
The Pivot Table chart was not rendering HTML content within its cells. Instead of displaying formatted HTML (e.g., hyperlinks), it showed the raw HTML tags as plain text. This occurred even when the underlying data contained valid HTML strings and the allowRenderHtml prop was set to true.

### BEFORE SCREENSHOT
Screenshot 2025-05-15 at 11 58 30

### AFTER SCREENSHOT
Screenshot 2025-05-15 at 12 19 33

### TESTING INSTRUCTIONS
- This [chart export](https://github.com/user-attachments/files/20231051/chart_export_20250515T153644.zip) can be imported into Superset to directly observe the issue. But also possible to be reproduced following the steps below:
1. Create or import a Pivot Table chart where the dataset provides HTML content for one or more cells.
2. Configure the chart to use these HTML-containing fields.
3. Observe the rendering of the data cells in the pivot table. The attached chart export can be imported into Superset to directly observe the issue.

### ADDITIONAL INFORMATION
- [X] Has associated issue: #33449 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-pivot-table/src/react-pivottable/TableRenderers.jsx
+++ b/superset-frontend/plugins/plugin-chart-pivot-table/src/react-pivottable/TableRenderers.jsx
@@ -65,6 +65,13 @@ function displayHeaderCell(
   );
 }
 
+function renderFormattedValue(formattedValue, allowRenderHtml) {
+  if (allowRenderHtml && typeof formattedValue === 'string') {
+    return safeHtmlSpan(formattedValue);
+  }
+  return formattedValue;
+}
+
 export class TableRenderer extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


richardfogaca:
> So I'm a little confused here. I see you have the "Render Columns in HTML" checkbox checked, and it looks like the cells ARE rendering HTML... but it's just not rendering certain tags (e.g. `strong`)?

@rusackas Actually the cells do not render html properly, with the exception of `country` column (this would be a header). In the example you can see a <strong> tag not rendered and also a <em> tag, but other tags are also not rendered like <a>

rusackas:
This has conflicts, and I think there was a recent PR that tackled the same topic and got merged. Converting to draft for now, but maybe this isn't needed any more?

msyavuz:
Yeah, i think i fixed this with [this ](https://github.com/apache/superset/pull/34351). It looks identical to this one. We had a bit of duplicate work so sorry about that!

PR #33467, opened 2025-05-16
fix(NativeFilters): Apply existing values

geido:
### SUMMARY
This PR https://github.com/apache/superset/pull/33271 has introduced an issue for which existing values would not be showing in select filters when using a permalink for filters with a default value, even though they were applied correctly to the end query.

### BEFORE

https://github.com/user-attachments/assets/f44210bc-ac7e-4c05-b644-a80c10ddb082

### AFTER

https://github.com/user-attachments/assets/7733112e-01f5-455e-af22-048f76b76a50

### TESTING INSTRUCTIONS
1. Open a Dashboard and set a filter with a default value
2. Apply some other filter value and copy a permalink
3. Visit the permalink
4. The filter value should be the last applied not the default

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
+++ b/superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
@@ -335,7 +335,13 @@ export default function PluginFilterSelect(props: PluginFilterSelectProps) {
       return;
     }
 
-    // Case 2: Handle the default to first Value case
+    if (filterState.value !== undefined) {
+      // Set the filter state value if it is defined
+      updateDataMask(filterState.value);
+      return;
+    }
+
+    // Handle the default to first Value case
     if (defaultToFirstItem) {
       // Set to first item if defaultToFirstItem is true
       const firstItem: SelectValue = data[0]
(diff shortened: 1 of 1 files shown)


superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx
```diff
@@ -335,7 +335,13 @@ export default function PluginFilterSelect(props: PluginFilterSelectProps) {
       return;
     }
 
-    // Case 2: Handle the default to first Value case
+    if (filterState.value !== undefined) {
```
Antonio-RiveroMartnez, line 338: Are other nullish values something we handle differently? (like for resetting the value?)

geido: I think so this is reintroducing the logic as it was

Antonio-RiveroMartnez (approved):
lgtm

geido:
@amaannawab923 can you verify that this will work with your original changes?

PR #33473, opened 2025-05-16
feat: use sqlglot to set limit

betodealmeida:
### SUMMARY

Part of https://github.com/apache/superset/issues/26786, stacked on:

- https://github.com/apache/superset/pull/33457
- https://github.com/apache/superset/pull/33456

This PR changes the logic of setting the limit in a query to use `sqlglot`, adding the `set_limit_value(limit: int) -> None` method to the `BaseSQLStatement` class.

The method **always** sets the requested limit. It's up to the application to determine if a bigger limit should be forced or not. Before, it had some logic that determined if the limit should be updated or not depending on if the existing limit was smaller than the desired limit. This keeps the method simpler and decoupled from business logic.

One of the main advantages of using `sqlglot` here is that it abstracts all the different ways a query can be limited:

```sql
SELECT * FROM t LIMIT 10;  -- Postgres, Clickhouse, Bigquery, etc.
SELECT TOP 10 * FROM t;  -- Teradata, MS SQL, Vertica, etc.
SELECT * FROM t FETCH FIRST 10 ROWS ONLY;  -- Oracle, DB2, Apache Derby
```

These are all abstracted as a `sqlglot.exp.Limit`, and rendered correctly when the engine is specified (see unit tests).

Previously there was additional logic for Teradata that accounted for the `SAMPLE` function:

```sql
SELECT * FROM t SAMPLE n  -- returns n rows from t
```

I removed this logic because `SAMPLE` is not a limit — it limits the number of rows fetched from the table, but **before the projection is applied**. The projection could have a UDTF that generates millions of rows even if only 10 were read from the table. Because of that, I removed the logic. (The tests were also using invalid syntax, so I'm not even sure if this feature ever worked!)

### TESTING INSTRUCTIONS

Added tests, and moved old tests to the new pattern.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/db_engine_specs/kusto.py
```diff
@@ -105,7 +106,6 @@ def convert_dttm(
 
 
 class KustoKqlEngineSpec(BaseEngineSpec):  # pylint: disable=abstract-method
-    limit_method = LimitMethod.WRAP_SQL
```
betodealmeida, line 108: Kusto supports limit, no need to wrap the query.

superset/db_engine_specs/mssql.py
```diff
@@ -52,10 +52,8 @@
 class MssqlEngineSpec(BaseEngineSpec):
     engine = "mssql"
     engine_name = "Microsoft SQL Server"
-    limit_method = LimitMethod.WRAP_SQL
```
betodealmeida, line 55: MS SQL supports limit, no need to wrap the query.

superset/db_engine_specs/teradata.py
```diff
@@ -15,19 +15,16 @@
...
-from superset.db_engine_specs.base import BaseEngineSpec, LimitMethod
+from superset.db_engine_specs.base import BaseEngineSpec
 
 
 class TeradataEngineSpec(BaseEngineSpec):
     """Dialect for Teradata DB."""
 
     engine = "teradatasql"
     engine_name = "Teradata"
-    limit_method = LimitMethod.WRAP_SQL
```
betodealmeida, line 26: Ditto here.

superset/sql/parse.py
```diff
@@ -767,6 +781,33 @@ def get_limit_value(self) -> int | None:
 
         return None
 
+    def set_limit_value(self, limit: int) -> None:
```
betodealmeida, line 784: It's been fun implementing the logic in KQL, but I'm considering writing a SIP to get rid of it, since it's not SQL and double the work.

tests/unit_tests/db_engine_specs/test_teradata.py
```diff
@@ -1,43 +0,0 @@
...
-@pytest.mark.parametrize(
-    "limit,original,expected",
-    [
-        (100, "SEL TOP 1000 * FROM My_table", "SEL TOP 100 * FROM My_table"),
-        (100, "SEL TOP 1000 * FROM My_table;", "SEL TOP 100 * FROM My_table"),
-        (10000, "SEL TOP 1000 * FROM My_table;", "SEL TOP 1000 * FROM My_table"),
-        (1000, "SEL TOP 1000 * FROM My_table;", "SEL TOP 1000 * FROM My_table"),
-        (100, "SELECT TOP 1000 * FROM My_table", "SELECT TOP 100 * FROM My_table"),
-        (100, "SEL SAMPLE 1000 * FROM My_table", "SEL SAMPLE 100 * FROM My_table"),
-        (10000, "SEL SAMPLE 1000 * FROM My_table", "SEL SAMPLE 1000 * FROM My_table"),
```
betodealmeida, line 30: This is not even valid syntax... `SAMPLE` goes at the end, not in the projection!

Vitor-Avila (approved):
LGTM! Manually tested with SQLite and Snowflake too

PR #33475, opened 2025-05-16
fix(dependabot): adds required schedule to uv updates

rusackas:
### SUMMARY
Seems dependabot got more picky somewhere along the line:
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,8 @@ updates:
   - package-ecosystem: "uv"
     directory: "requirements/"
     open-pull-requests-limit: 10
+    schedule:
+      interval: "monthly"
     labels:
       - uv
       - dependabot
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,7 +28,7 @@ updates:
     directory: "requirements/"
     open-pull-requests-limit: 10
     schedule:
-      interval: "monthly"
+      interval: "weekly"
     labels:
       - uv
       - dependabot


mistercrunch (approved):
should we do weekly? My intent was to monitor progress and report back on https://github.com/dependabot/dependabot-core/pull/10040#issuecomment-2698391122 , so far I don't think dependabot is doing anything or what's required to allow for the magic to happen.

I found out looking at some of the code / PR that dependabot seems to be trying to be magical, finding your reqs files and triggering script based on their headers (I think), so if you `requirements.txt` has a comment saying `# generated using this command`, it'll try to run that.

In our case, dependabot should run `scripts/uv-pip-compile.sh` that runs atomically for all `requirements/*.in` and I don't think there's anyway to control/configure that in `.github/dependabot.yml`.

.github/dependabot.yml
```diff
@@ -27,6 +27,8 @@ updates:
   - package-ecosystem: "uv"
     directory: "requirements/"
     open-pull-requests-limit: 10
+    schedule:
+      interval: "monthly"
```
rusackas, line 31:
```suggestion
      interval: "weekly"
```

rusackas:
Weekly it is! Was just conforming to the rest of 'em. We can switch it back to monthly if it gets too noisy and we want to get hit by the wave less often.

PR #33494, opened 2025-05-16
fix(table): table ui fixes

amaannawab923:
### SUMMARY

This pr fixes 2 of the ui issues caused by the pr  https://github.com/apache/superset/pull/33357

The following are fixed: 
1. Error messages missing punctuation to separate them
2.Search field misaligned in smaller resolution

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/packages/superset-ui-core/src/validator/validateServerPagination.ts
```diff
@@ -21,10 +21,18 @@ import { t } from '../translation';
...
-    return t('Server pagination needs to be enabled for values over %s', max);
+  if (
+    Number(v) > +maxValueWithoutServerPagination &&
+    Number(v) <= maxServer &&
+    !serverPagination
+  ) {
+    return t(
+      'Server pagination needs to be enabled for values over %s',
+      maxValueWithoutServerPagination,
+    );
```
amaannawab923, line 35: We were getting multiple messages because over 500k also this validator was returning warning .... hence we now check the range that only if its in between 100k & 500k then the validator should return warning message

superset-frontend/plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -113,6 +113,10 @@ const StyledSpace = styled(Space)`
   }
 `;
 
+const StyledRow = styled.div`
+  display: flex;
+`;
```
amaannawab923, line 118: The alignment wasnt proper because flex wasnt getting applied

superset-frontend/plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -466,7 +470,11 @@ export default typedMemo(function DataTable<D extends object>({
               ) : null}
             </div>
             {searchInput ? (
-              <StyledSpace className="col-sm-6">
+              <StyledSpace
+                className={
+                  renderTimeComparisonDropdown ? 'col-sm-7' : 'col-sm-8'
+                }
```
amaannawab923, line 476: Also the Space Component had less space with sm-6 which was causing it to break on Small screen

superset-frontend/plugins/plugin-chart-table/src/consts.ts
```diff
@@ -30,3 +30,7 @@ export const PAGE_SIZE_OPTIONS = formatSelectOptions<number>([
   100,
   200,
 ]);
+
+export const SERVER_PAGE_SIZE_OPTIONS = formatSelectOptions<number>([
+  10, 20, 50, 100, 200,
```
amaannawab923, line 35: The original Page size options had an all option as well causing the query to become heavy hence made a new variable

superset-frontend/plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -397,6 +397,7 @@ const config: ControlPanelConfig = {
                     v,
                     state?.server_pagination,
                     state?.maxValueWithoutServerPagination || DEFAULT_MAX_ROW,
+                    state?.maxValue || DEFAULT_MAX_ROW_TABLE_SERVER,
```
amaannawab923, line 400: Validate Server Pagination needs the maxValue as well now to check the range

mistercrunch (approved):
LGTM

PR #33501, opened 2025-05-17
fix: optimize Explore popovers rendering

mistercrunch:
Recently noticed that in Explore, loading popovers for metrics, columns, and filters take forever to load. 10+ seconds depending on the number of columns in your dataset. Upon inspection it had to do with a call to `react-dom/server`'s `renderToStaticMarkup` which isn't meant to be used on the client, more of a server thing. The goal was to align tooltips in AceEditor (which use native html) with AntD's. For that we would pass an antd Tooltip to a server-side like component to generate html, which is super slow and inefficient, also it won't work once we land the theming branch as it leverages antdv4's .less definitions.

My solution is to generate the html directly. Not great or fancy but it works and is fast. Plan is probably to move off Ace and onto Monaco...

10 seconds to load a tooltip is not ok ...
Screenshot 2025-05-16 at 5 37 41 PM

Also throwing in dompurify in here as we html-render potentially unsafe, user-provided field into html. I think this addresses a XSS vulnerability.

#### before
Screenshot 2025-05-17 at 10 48 26 AM

#### after
Screenshot 2025-05-17 at 10 40 49 AM

Tagging @justinpark since he introduced `react-dom/server` here ->  https://github.com/apache/superset/pull/29672

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -240,6 +240,12 @@ disallow_untyped_calls = false
 disallow_untyped_defs = false
 disable_error_code = "annotation-unchecked"
 
+# TODO: remove this once cryptography is fixed, introduced in cryptography 44.0.3
+[[tool.mypy.overrides]]
+module = "cryptography.*"
+ignore_errors = true
+follow_imports = "skip"
+
 [tool.ruff]
 # Exclude a variety of commonly ignored directories.
 exclude = [
@@ -271,7 +277,6 @@ exclude = [
     "site-packages",
     "venv",
 ]
-
 
 # Same as Black.
 line-length = 88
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -240,6 +240,12 @@ disallow_untyped_calls = false
 disallow_untyped_defs = false
 disable_error_code = "annotation-unchecked"
 
+# TODO: remove this once cryptography is fixed, introduced in cryptography 44.0.3
+[[tool.mypy.overrides]]
```
mistercrunch, line 244: This is unrelated to the PR but fixes the issues in CI in `master`

msyavuz (approved):
LGTM! Maybe we should have an eslint rule for not using server stuff like renderToStaticMarkup/renderToString?

mistercrunch:
> Maybe we should have an eslint rule for not using server stuff like renderToStaticMarkup/renderToString?

Probably a one-off mistake, probably don't need a rule...

PR #33523, opened 2025-05-19
fix(dataset): prevent metric duplication error when editing SQL and adding metric

### LINKED ISSUE #33505: Editing a dataset SQL query and adding a new metric at the same time causes a metric duplication error
### Bug description

Go to Datasets, and click on any virtual dataset.

Click Edit dataset.

Update the SQL query under the Source tab (e.g., change a limit or add a new column).

Switch to the Metrics tab and add a new custom metric (e.g., COUNT(*) + 1).

Click Save.

Actual result:
An error appears:
"Error saving dataset: One or more metrics already exist"

### Screenshots/recordings

Image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

When editing the dataset, if we edit both the SQL and the metric, we would see this error, can fix it by fetching the latest metrics info/ id

Superset’s onConfirmSave() logic first sends a PUT request to save the dataset changes, including any new metrics and SQL updates. If the SQL was changed, it automatically triggers a second PUT after syncing columns. However, this second request reuses the in-memory currentDatasource.metrics, which doesn’t yet include the freshly assigned metric ids returned by the backend after the first save. As a result, those metrics still appear to be new (because they lack ids), so the backend treats them as duplicates when it sees the same metric_names being submitted again. This triggers a validation error, with the backend assuming you’re trying to create metrics that already exist.

Per discussion, the easiest fix is to fetch the metrics' id before update.

### Checklist

jqqin:
### SUMMARY

This PR is intended to solve the issue: https://github.com/apache/superset/issues/33505

When editing a dataset, if both the SQL and a new metric are changed at the same time, Superset throws a "metric already exists" validation error.

This happens because onConfirmSave():

Sends a first PUT to save the updated SQL and metrics

Then, if the SQL has changed, sends a second PUT after syncing columns — but reuses the outdated currentDatasource.metrics without refreshed IDs

As a result, the new metrics appear to lack ids, so the backend treats them as duplicates based on name.

This fix ensures that, after the first save and before the second PUT, we fetch the updated dataset and assign the correct metric IDs, preventing the duplication error.

fixes #33505 

We would get this error when both the SQL and a new metric are changed at the same time:

image

### TESTING INSTRUCTIONS

After the fix, try editing the SQL query and adding a new metric would cause no errors:

image

image

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Datasource/DatasourceModal.test.jsx
+++ b/superset-frontend/src/components/Datasource/DatasourceModal.test.jsx
@@ -34,8 +34,8 @@ import mockDatasource from 'spec/fixtures/mockDatasource';
 // Define your constants here
 const SAVE_ENDPOINT = 'glob:*/api/v1/dataset/7';
 const SAVE_PAYLOAD = { new: 'data' };
-const SAVE_DATASOURCE_ENDPOINT = 'glob:*/api/v1/dataset/7';
-const GET_DATASOURCE_ENDPOINT = SAVE_DATASOURCE_ENDPOINT;
+const SAVE_DATASOURCE_ENDPOINT = 'glob:*/api/v1/dataset/7?override_columns=*';
+const GET_DATASOURCE_ENDPOINT = 'glob:*/api/v1/dataset/7';
 const GET_DATABASE_ENDPOINT = 'glob:*/api/v1/database/?q=*';
 
 const mockedProps = {
--- a/superset-frontend/src/components/Datasource/DatasourceModal.tsx
+++ b/superset-frontend/src/components/Datasource/DatasourceModal.tsx
@@ -17,7 +17,7 @@
  * under the License.
  */
 import { FunctionComponent, useState, useRef } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
+import { useSelector } from 'react-redux';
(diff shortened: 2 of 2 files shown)


rusackas:
A rebase should solve that dependency license check that's failing on CI. Thanks for the PR!

jqqin:
@rusackas Thanks! The rebase is complete — looks like the workflows now need maintainer approval to re-trigger. Please let me know if there’s anything else I should do.

mistercrunch:
Pointing out that this component is already pretty heavy and slow, especially on larger datasets. Haven't dug in the logic in place, but wondering if this could be handled on the backend directly and avoid a round trip (?)

JZ6:
> Pointing out that this component is already pretty heavy and slow, especially on larger datasets. Haven't dug in the logic in place, but wondering if this could be handled on the backend directly and avoid a round trip (?)

thanks, let us take a look at optimizing this further

sadpandajoe:
Tested this in the ephemeral and LGTM

eschutho (approved):

PR #33550, opened 2025-05-21
feat(dashboards): Add open graph metadata for dashboards

worldlyjohn:
### SUMMARY

- include Open Graph metadata in SPA template
- supply dashboard title and description to dashboard and embedded views
- Note: Since dashboards are behind auth -- third party apps that aren't already authenticated -- will not be able to unfurl. This PR however is a pre-req for future third-party app unfurling of urls once auth can be handled.

### TESTING INSTRUCTIONS

- Open a Dashboard
- View source and confirm the presence of `og:title` tags (or paste into slack to render)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Oh nice, actual good use of server-side html generation!

Wondering a few things -> 
- does it even work for embedded? I doubt Slack will process the iframe if there's one and bubble up the meta?
- how does auth work here? I'd assume the in-slack browser wouldn't have access to the dashboard, unless maybe if it's public?

worldlyjohn:
> * how does auth work here? I'd assume the in-slack browser wouldn't have access to the dashboard, unless maybe if it's public?

@mistercrunch You're right, Slack’s crawler is not logged in, so it receives this redirect and never reaches the dashboard page. 

Upon further research, unfurling works for analytics tools like Mixpanel because they have [developed Slack integrations](https://apache-superset.slack.com/marketplace/AKMP1USA3-mixpanel) that provide an access token.

Appears a similar Preset/Superset Slack integration would need to handle auth before this scope of work is relevant 🤷‍♂️ .

Alternative approaches:
- **Wrap metadata behind a small “unfurl” view.** Create a lightweight endpoint that serves only the Open Graph tags (and optional screenshot) without loading the full application. Slack would be sent to this endpoint via the public link. Downside is this could leak possibly confidential dashboard titles to anyone with a dashboard link.
- **Provide Slack access via a token.** Use the existing guest-token mechanism (`SecurityRestApi.guest_token`) to generate a short-lived token that authorizes Slack to fetch only the metadata. The request could include `?guest_token=<token>` so the crawler receives the correct page without a redirect. Downside is a request for this link that cloaks itself as a Slack request (modifying HTTP request headers perhaps) to retrieve the dashboard title.
- **Develop a Superset Slack integration**.

wdyt?  (I likely don't have the bandwidth atm to work on these alternative paths, but this unfurling support would be a welcomed addition as we roll out Superset/Preset to our 1000 person org next month) :)

mistercrunch:
I want to highlight some of the security concerns and mitigation mechanisms. 

- Object metadata can be sensitive in some contexts: just noting that a dashboard can be titled `Next Earning Calls: Crushing it!!!!` and that's potentially sensitive, also paired with the fact that we use autonumber ids, an attacker could essentially fetch a list of dashboard / chart titles on a Superset instance
- Public instances: if/when an instance is public, this is even more sensitive - thinking of Preset here or any internet-visible instance
- Superset in-VPN: assuming the user is in some sort of VPN, Slack should be able to call and unfurl through that same VPN, maybe that's where this feature is "ok", where someone could decide that dashboard names are public within their network (?) Still somewhat dubious... Users might not know that while creating a new dashboard.
- RBAC concerns: currently Superset only shows you charts/dashboards that are accessible by you, meaning `Next Earning Calls: Crushing IT!!!!` could be a hazard even internally: knowing about the existence of a specific dashboard title could be an issue within the VPN walled garden

Given that:
- Slack app + impersonation (solving for RBAC concerns) seems required for most environment
- About creating a public/insecure endpoint that serves logged off metadata: maybe a feature flag could mitigate this configuration for instances that consider dashboard names public within their VPN. Flag should come as `false` by default and come with a clear disclaimer
- About squeezing metadata in headers, totally fine by me, but Slack won't see them, though it's a pre-req building block for the Slack app to work. Say if you build your own Slack app, you'll need that and there's no arm in adding those meta tags today.

About the Slack app:
- probably needs to be aligned with your auth mechanism? so if you use oauth, or username/password or Okta or whatevs, the Slack app might be different, probably hard to build one that works for everyone
- if anything, an open source implementation as a reference for the community would be great. Ideally lives outside of the main repo

worldlyjohn:
> it's a pre-req building block for the Slack app to work

Agree here.  In this case, I think this PR is properly scoped.
I changed the title to reflect the new scope

mistercrunch:
From my understanding I think it's safe where that page should only render if the user has access to the dashboard. What would be a security concern is if we returned metadata to an unauthenticated user to enable [unauthenticated] Slack (and everything else).

About touching `superset/embedded/view.py` I don't think the meta tags will make it through an iframe but probably doesn't hurt.

rusackas:
Hi @worldlyjohn\! This PR was looking good and the security discussion seemed resolved. It's developed conflicts since then - could you rebase when you get a chance? Would love to get this merged\!

rusackas:
@worldlyjohn this one's developed conflicts since the security discussion wrapped up. I'll see if I can help rebase it to move it forward.

rusackas (approved):
Code LGTM — small and clean, mistercrunch's access concern is settled, and the values are autoescaped by Jinja. I'll get this merged once CI's green.

rusackas:
I'll see if I can help CI to pass here...

Issue #31348, opened 2024-12-09
Negative values are being shown on the wrong side of radar chart axes


gilles-dataroots:
### Bug description

Since my company updated to the latest version of Superset begin December, I'm seeing weird visuals on radar plots. As you can see the negative values are now being plotted on the opposite side of the axis.

### Screenshots/recordings

![image](https://github.com/user-attachments/assets/d9a9aa0a-c059-4d73-8918-80551acd79f6)

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [ ] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

gilles-dataroots:
The previous version of the radar chart would dynamically make sure all values were being plotted on the axis. Is this no longer the intended goal? A hardcoded radarMetricMinValue would not allow for an index without a hard minimum...

gilles-dataroots:
That sounds promising, where can I actually find and update the radarMetricMinValue parameter?

gilles-dataroots:
I've found the option in the Customize tab of the edit chart page. Even after removing the 0 and setting the radarMetricMinValue for the metric Night to auto, the chart still displays some negative values on the wrong size of the origin as you can see below.
![image](https://github.com/user-attachments/assets/6693b162-97fc-42ef-ba7b-c20f8f5c1bee)
![image](https://github.com/user-attachments/assets/a2457bfd-c11f-4560-a538-6f4ec4df5f4c)

rusackas:
Has anyone tried to reproduce this with the ECharts examples on their site? I wonder if it's our bug or theirs.

rusackas:
This seems to happen in ECharts, too... perhaps someone would be interested in filing an issue on their repo?

rusackas:
This same negative-value placement happens in ECharts directly, so it's an upstream limitation rather than a Superset bug. Closing as not-planned - please file upstream with ECharts; setting radarMetricMinValue is the workaround in the meantime.

PR #33559, opened 2025-05-22
fix(Radar): Radar chart normalisation

amaannawab923:
### 🔧 Summary
This PR introduces a fix for for the Radar Chart in Superset, addressing a critical issue where large discrepancies in metric magnitudes made the chart visually misleading and unusable for comparison.

### 🐛 The Issue
When plotting multiple metrics with vastly different scales (e.g., 208000, 2590, 9200), the Radar Chart currently uses a faulty calculation mechanism which renders just a Polygon because each value occupies the max on its axis. This causes small values to appear indistinguishable from large ones—rendering the chart ineffective for visual comparison.

### 🎥 Before (no normalization):

Screenshot 2025-05-26 at 11 08 55 PM

As it can be seen in above screenshot 797.73 & 2.43k appear visually equivalent, even though they arent & the end user is not able to differentiate between them through the radar chart

### 🧠 Root Cause
Radar Charts plot all metric values using a common 0-to-max scale. This approach ignores the fact that some metrics may span much smaller ranges. As a result:

A metric like 2.5k appears to stretch just as far as 208k

There’s no visual cue for the viewer to recognize magnitude differences

### ✅ The Fix
This PR introduces:

A new normalisation function in Radar chart where  each metric’s values are scaled between 0 and 1
This results in equal visual weight per metric, regardless of absolute value
Safe denormalization logic for tooltips and labels
Tooltips and labels show original values, not normalized ones
Denormalization map tracks each series’ normalized-to-original value mapping

### 🧪 How It Works
The Normalize function activates per-series normalization during data transformation.
Original values are stored in a SeriesNormalizedMap structure.
Custom formatters use this map to denormalize values when displaying them in tooltips or labels.

### 🖼️ UX Enhancements
Tooltip and label values now reflect true, raw numbers even when normalized values are plotted.

Improves clarity when comparing metrics with different scales.

### 🎥 After (with normalization):

Screenshot 2025-05-26 at 11 07 12 PM

Now 797.73 appears visually appropriate in comparison to 2.43k

Metrics now appear proportionally within their own scale, improving accuracy and readability.

### 🧰 Developer Notes

Introduced new tooltip formatter and label formatter handling
Implemented safe getDenormalisedSeriesValue() accessor

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
```diff
@@ -91,6 +91,20 @@ const config: ControlPanelConfig = {
...
+          {
+            name: 'is_normalised',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Normalise Values'),
+              renderTrigger: true,
+              default: false,
+              description: t(
+                'Normalise the radar chart values between 0 & 1 to get consistent plotting.',
+              ),
```
amaannawab923, line 104: Control for toggling the Normalisation

superset-frontend/plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
```diff
@@ -210,6 +224,8 @@ const config: ControlPanelConfig = {
                   columnsPropsObject: { colnames, coltypes },
                 };
               },
+              visibility: ({ controls }) =>
+                Boolean(!controls?.is_normalised?.value),
```
amaannawab923, line 228: The column configuration allows to specify min & max value for each axis , but since normalise value will lock the min & max value to 0 & 1 respectively , the column config shouldnt be shown

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -111,11 +124,36 @@ export default function transformProps(
   const { setDataMask = () => {}, onContextMenu } = hooks;
   const colorFn = CategoricalColorNamespace.getScale(colorScheme as string);
   const numberFormatter = getNumberFormatter(numberFormat);
+  const denormalizedSeriesValues: SeriesNormalizedMap = {};
+
```
amaannawab923, line 128: We use an Object to store the denormalised values belonging to each series because after normalising we need to pass the denormalised values to the label formatter & tooltip formatter to keep the chart consistent & preserve original values

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -111,11 +124,36 @@ export default function transformProps(
...
+        denormalizedSeriesValues[seriesName],
+        normalisedValue,
+      )
+    ) {
+      return denormalizedSeriesValues[seriesName][normalisedValue];
+    }
+
+    // Fallback: return the normalised value itself as a number
+    return Number(normalisedValue);
+  };
```
amaannawab923, line 148: The  getDenormalisedSeriesValue is getter function for safely getting the denormalised value , if we dont find the value we fallback to normalised value, this shouldnt happen but we do this to prevent crashes & unexpected errors which can crash the user experience

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -212,7 +250,40 @@ export default function transformProps(
...
+        const seriesName = String(series?.name || '');
+        denormalizedSeriesValues[seriesName] = {};
+
+        return {
+          ...series,
+          value: normalizeArray(series.value as number[], 10, seriesName),
+        };
+      }
+      return series;
+    });
```
amaannawab923, line 276: We normalise the values for each series only if we have isNormalised as true

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -259,6 +330,33 @@ export default function transformProps(
...
+        seriesName,
+        String(normalizedValue),
+      );
+      tooltip += `
+        <div style="display:flex;">
+          <div>${colorDot}${metric}:</div>
+          <div style="font-weight:bold;margin-left:auto;">${originalValue}</div>
+        </div>
+      `;
+    });
```
amaannawab923, line 355: Since we need to override the denormalised values with the normalised values in tooltip we create the tooltip from scratch with the tooltip formatter in the format that echarts expect which is html string

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -267,6 +365,9 @@ export default function transformProps(
       ...getDefaultTooltip(refs),
       show: !inContextMenu,
       trigger: 'item',
+      ...(isNormalised && {
+        formatter: NormalisedTooltipFormaterr,
+      }),
```
amaannawab923, line 370: We use this normalised tooltip formatter only when normlalised is true

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -259,6 +330,33 @@
...
 
+  const NormalisedTooltipFormaterr = function (params: any) {
+    const { color } = params;
+    const seriesName = params.name || '';
+    const values = params.value;
+
+    const colorDot = `<span style="display:inline-block;margin-right:5px;border-radius:50%;width:5px;height:5px;background-color:${color}"></span>`;
+
+    // Start with series name without dot
+    let tooltip = `<div style="font-weight:bold;margin-bottom:5px;">${seriesName || 'series0'}</div>`;
```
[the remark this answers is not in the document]
amaannawab923: If no dimensions are available , Echarts falls back to series0 ... we are following the convention here

superset-frontend/plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
Antonio-RiveroMartnez: Don't we need a tooltip to inform users why the controls disappear and the implication of selecting or not the normalize?

Antonio-RiveroMartnez (commented):
Could we also add tests for this? like it hides the other controls, transforms the props etc.

superset-frontend/plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
amaannawab923: This checkbox will only apply for Radar chart , i dont think so Normalization is applicable anywhere else

michael-s-molina:
@amaannawab923 Amazing PR description by the way 👏🏼

rusackas:
Just adding a note that we have an open issue about negative values misbehaving on this chart:
https://github.com/apache/superset/issues/31348

I'm not sure if this could be a "two birds with one stone" opportunity :D

betodealmeida:
This is awesome, @amaannawab923, but I'm a bit confused by a couple things.

First, for the current behavior you say that "all metric values using a common 0-to-max scale". But we can see from the results tab that **we only have a single value for each metric**. There is no max nor min, so how do we determine where in the radial axis we should place the metric?

Second, to me the original behavior you describe is what I'd call "normalized". Since these are different metrics, if we're showing them relative to some computed value so that two different metrics with widely different ranges appear to have similar values, then they're are normalized. On the other hand, if we want to make sure that a metric with a high range is shown as bigger value than one that has a small range (like the new behavior), I'd call that "denormalized".

amaannawab923:
> This is awesome, @amaannawab923, but I'm a bit confused by a couple things.
> 
> First, for the current behavior you say that "all metric values using a common 0-to-max scale". But we can see from the results tab that **we only have a single value for each metric**. There is no max nor min, so how do we determine where in the radial axis we should place the metric?
> 
> Second, to me the original behavior you describe is what I'd call "normalized". Since these are different metrics, if we're showing them relative to some computed value so that two different metrics with widely different ranges appear to have similar values, then they're are normalized. On the other hand, if we want to make sure that a metric with a high range is shown as bigger value than one that has a small range (like the new behavior), I'd call that "denormalized".


1.Hey  @betodealmeida  , great question — here’s how the Radar Chart handles max values in the non-normalized mode when no explicit max is provided:

const maxValueInControl = columnConfig?.[metricLabel]?.radarMetricMaxValue;
const minValueInControl = columnConfig?.[metricLabel]?.radarMetricMinValue;

// Ensure that 0 is at the center of the polar coordinates
const metricValueAsMax =
  metricLabelAndMaxValueMap.get(metricLabel) === 0
    ? Number.MAX_SAFE_INTEGER
    : metricLabelAndMaxValueMap.get(metricLabel);
const max =
  maxValueInControl === null
    ? metricValueAsMax
    : maxValueInControl;
💡 What’s happening here:
If the user does not manually configure a max (radarMetricMaxValue is null or undefined), we fallback to the actual metric value returned from the query.

That value is treated as the max for that metric.

This means the data point will appear at the outer edge of the radial axis (since it lies at its own max).

If the value is 0, we use Number.MAX_SAFE_INTEGER to avoid dividing by zero and ensure the chart renders safely.

This is the old function where we are calculating the min & max but usually the max defaults to the value of metric itself hence metric goes onto stretch to the end of the Chart

2. Thanks for the thoughtful feedback! Just to clarify, by "normalized" we mean rescaling each series’ values to a [0, 1] range, based on the max in that series. This ensures fair visual comparison between metrics with very different scales.
The original behavior plotted raw values, which can mislead when metrics vary significantly in magnitude (e.g., 208k vs. 2.5k).

This is common term in data science that normalizing meaning scaling back to 0 to 1 ... Since the values on screen will still refer to the original values in Tooltip & labels , i thought it would be intuitive to the end clients who can be usually data engineers or data analyst to communicate easily , that hey this control will rescale your values back to scale of 0 to 1

amaannawab923:
> Could we also add tests for this? like it hides the other controls, transforms the props etc.

Thanks for the suggestion @Antonio-RiveroMartnez  , i have added some normalization tests for radar charts transform props , thanks for your time for reviewing the pr

amaannawab923:
> @amaannawab923 Amazing PR description by the way 👏🏼

Thanks for noticing @michael-s-molina 😊 , i try to make my prs very descriptive so that it clearly communicates , Thanks

betodealmeida:
> This is common term in data science that normalizing meaning scaling back to 0 to 1 ... Since the values on screen will still refer to the original values in Tooltip & labels , i thought it would be intuitive to the end clients who can be usually data engineers or data analyst to communicate easily , that hey this control will rescale your values back to scale of 0 to 1

I know it's a common term, but the normalization you're doing is an **implementation detail** to show the metrics denormalized — that is, so their absolute values can be compared. From the **user perspective** you're not normalizing the metrics, you're doing the opposite. Normalizing the metrics from the user perspective would be representing `coffee_price=5` and `house_price=500000000` with similar sizes.

betodealmeida:
Instead of calling this "normalize metrics", I think it would be better if we just defaulted to using the highest metric as the maximum value for each metric. So if we have 4 metrics: 1, 10, 50, 100 the first one would have a length of 1% of the radius, the second 10%, etc. Then if the user wants they can customize the maximum range of particular metric.

Right now each metric has by default a range of 0 to "auto", we just make it so "auto" means "the highest metric".

betodealmeida:
> If the user does not manually configure a max (radarMetricMaxValue is null or undefined), we fallback to the actual metric value returned from the query.

Let's change this behavior so that it falls back to the highest metric value returned from the query.

evanvolgas:
> Instead of calling this "normalize values", 

Would "scaling values" make more sense?

betodealmeida:
> Would "scaling values" make more sense?

No, because (with the new approach) the values are not being scaled. It's literally just showing big numbers big, and small numbers small. Scaling values would be if we normalized the values to somehow show a small number as big as a big number on a different metric (the price of coffee vs. price of house example I gave).

If the user wants that they can customize the chart and manually add the maximum value for each metric. It has to be manual because there's no information on how to do that; the query simply returns a single value for each metric, there's no indication on their normal range.

I don't think we want an option at all. The new behavior should be the default: if you choose multiple metrics they need to be compared to each other because there's nothing else to compare them too. You can scale them or normalize them, because each one is just a single number. You can only show their relative sizes to each other, unless the user has more context and changes the range manually.

betodealmeida:
I don't want to distract from the fact that this PR is important and fixes a bug that makes the radar visualization completely useless, but I have a strong impression that the description of this PR and some of the comments were written using AI. If that's the case, I think it's important to be transparent about the use of AI tools.

The description of the issue, for example, says:

> When plotting multiple metrics with vastly different scales (e.g., 208000, 2590, 9200), the Radar Chart currently uses a shared scale across all metrics. This causes small values to appear indistinguishable from large ones—rendering the chart ineffective for visual comparison.

This is not true at all. There's no shared scale across all metrics, **on the contrary**: the problem we see here happens because each metric has its own scale, so that they all end maxed up. We need a shared scale across all the metrics so that big numbers are shown big, and small numbers are shown small. I spent quite some time trying to understand the mismatch between the description of the problem and the proposed solution.

Please don't use AI to generate PR descriptions. If you do so, please be transparent and declare how you're using AI in the process.

betodealmeida:
And again, this PR is awesome and makes the chart usable, so I really appreciate your work on it! I just disagree with the terminology being used here. I'd be happy to loop in a product manager or designer to help us find the best UX for this visualization.

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -111,11 +124,36 @@ export default function transformProps(
...
+  ): number => {
+    if (
+      Object.prototype.hasOwnProperty.call(
+        denormalizedSeriesValues,
+        seriesName,
+      ) &&
+      Object.prototype.hasOwnProperty.call(
+        denormalizedSeriesValues[seriesName],
+        normalisedValue,
+      )
```
kgabryje, line 141: Can't we just do `denormalizedSeriesValues?.[seriesName]?.[normalisedValue]`?

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -259,6 +330,39 @@ export default function transformProps(
...
+      color: string;
+      name: string;
+      value: number[];
+    },
+  ) => {
+    const { color } = params;
+    const seriesName = params.name || '';
+    const values = params.value;
+
+    const colorDot = `<span style="display:inline-block;margin-right:5px;border-radius:50%;width:5px;height:5px;background-color:${color}"></span>`;
```
kgabryje, line 344: You could move NormalisedTooltipFormatter to a separate .tsx file and use JSX instead of strings

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
amaannawab923: Yes done

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
amaannawab923: Yes thanks ... Made a seperate tooltip component

rusackas:
Just a comment from the peanut gallery on terminology (naming things is hard). I like something more like "Equalize Axis Scales" which is more descriptive of what it's _doing_ from a UX perspective. 

A tooltip could be added next to the checkbox to the effect of "When enabled, each axis is given the same numeric range to improve visual comparison."

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -111,20 +128,46 @@ export default function transformProps(
...
+        config?.radarMetricMinValue !== 0;
+      return hasMax || hasMin;
+    }),
+  );
+
   const metricLabelAndMaxValueMap = new Map<string, number>();
   const metricLabelAndMinValueMap = new Map<string, number>();
   const columnsLabelMap = new Map<string, string[]>();
-  const transformedData: RadarSeriesDataItemOption[] = [];
+  let transformedData: RadarSeriesDataItemOption[] = [];
```
msyavuz, line 170: Is this necessary? Can't we create a new array when this is reassigned and use this?

kgabryje:
I didn't follow the discussion too closely here, but shouldn't we keep the control for changing the scales? Also, we should make sure that the existing charts are not affected, and the best way to do it would be to either set the control to false by default, or if we need it to be enabled by default, then we should create a migration for existing Radar charts.
Also, if there's a consensus about changing the name from "Normalise" to something else, could you change the variable and functions names in the code as well?

kgabryje (approved):

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -111,20 +128,46 @@ export default function transformProps(
...
+
+  const metricLabels = metrics.map(getMetricLabel);
+
   const formatter = (params: CallbackDataParams) =>
     formatLabel({
       params,
       numberFormatter,
       labelType,
+      getDenormalizedSeriesValue,
+      metricsWithCustomBounds,
```
msyavuz, line 148: I thought we have no-use-before-define configured, this is used before it was defined, we can just move definition above usage.

msyavuz (approved):
Thanks for this, maybe we should have more fitting examples for different charts. Some example data do not fit the chart use cases at all.

superset-frontend/plugins/plugin-chart-echarts/src/Radar/NormalizedTooltip.jsx
```diff
@@ -0,0 +1,79 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
amaannawab923: Done

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -29,12 +29,14 @@ import type { CallbackDataParams } from 'echarts/types/src/util/types';
 import type { RadarSeriesDataItemOption } from 'echarts/types/src/chart/radar/RadarSeries';
 import type { EChartsCoreOption } from 'echarts/core';
 import type { RadarSeriesOption } from 'echarts/charts';
+import { isNull, isUndefined } from 'lodash';
```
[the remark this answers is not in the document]
amaannawab923: Done

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
amaannawab923: Done

amaannawab923:
Please give it a final pass @betodealmeida , Thanks

superset-frontend/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
msyavuz: Echarts expects a string and renderToStaticMarkup causes a performance hit. See https://github.com/apache/superset/pull/33501. We should probably revert this.

PR #33567, opened 2025-05-22
chore(alerts & reports): increase Playwright timeout from 30 -> 60 seconds

sfirke:
### SUMMARY
Current default value for Playwright timeout on Alerts & Reports is 30 seconds, which seems unreasonably short. No value will please everyone but this seems like a step in the right direction.

### TESTING INSTRUCTIONS
Test it out in a deployment.


rusackas (approved):
LGTM!

PR #33577, opened 2025-05-23
fix(Security): Apply permissions to the AllEntities list/get_objects API endpoint

Vitor-Avila:
### SUMMARY
Currently, a user accessing the `/api/v1/tag/get_objects/` API endpoint is able to retrieve metadata from all objects -- even objects they don't have access to. This PR fix this issue by applying base filters to the charts and dashboards list returned.

It also implements a minor UI change that listing tagged content for a tag would also include this tag on the Tags columns -- I thought it was more consistent with other menus.

**Before: no permissions applied. The selected tag is also not included in the Tags column.**

image

**Now -- permissions applied. The selected tag is displayed in the Tags column.**

image

### TESTING INSTRUCTIONS
Tests added. For manual testing:
1. Create two charts (A and B) that are powered by different datasets.
2. Add the same tag to both.
3. Create a test account, and grant them access only to Chart A. 
4. Make sure this test account has access to list/read tags properly.
5. While logged in with the test account, make sure you can only see Chart A listed when accessing this tag.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/allEntities/AllEntitiesTable.test.tsx
+++ b/superset-frontend/src/features/allEntities/AllEntitiesTable.test.tsx
@@ -127,7 +127,7 @@ describe('AllEntitiesTable', () => {
     expect(screen.queryByText('Add tag to entities')).not.toBeInTheDocument();
   });
 
-  it('renders the correct tags for each object type, excluding the current tag', () => {
+  it('renders the correct tags for each object type', () => {
     render(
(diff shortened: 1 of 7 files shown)


betodealmeida (approved):
Lovely!

PR #33580, opened 2025-05-26
fix(sql_parse): correctly handle undefined Jinja macros during table extraction

wuqicyber:
### SUMMARY
This patch addresses Issue #33579, where Jinja-templated queries that invoke Superset macros
(like `filter_values()`) would fail the permission-check phase with an
`UndefinedError` (and thus `QueryIsForbiddenToAccessException`). 

Before:
image
After:
image

### TESTING INSTRUCTIONS
1. Enable Jinja templating: set `ENABLE_TEMPLATE_PROCESSING = True` in
   your `superset_config.py`.  
2. Ensure a user has only datasource access (no raw SQL access) to table like `video_game_sales`.  
3. As that user, run:
   ```sql
   WITH calculation AS (
     SELECT count(*), genre
     FROM video_game_sales
     WHERE genre IN ({{ "'" + "','".join(filter_values('genre')) + "'" }})
     GROUP BY genre
   )
   SELECT * FROM calculation;

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -972,7 +972,7 @@ def extract_tables_from_jinja_sql(sql: str, database: Database) -> set[Table]:
             node.data = "NULL"
 
     # re-render template back into a string
-    rendered_template = Template(template).render()
+    rendered_template = Template(template).render(processor._context)
 
     return (
         tables


hainenber:
Hello and thanks for your contribution! Pythonically speaking, underscore-prefixed attribute means it's for internal usage and expose the context to outside world may have unexpected security implications. 

Can you foresee a different way to tackle the problem?

Issue #33441, opened 2025-05-14
[SIP-169] Proposal for Extracting and Publishing Superset Core UI Components


mistercrunch:
### Motivation

The `src/components/` directory in Superset has grown into an unstructured mix of UI primitives, AntD wrappers, app-specific components, and deeply coupled logic. This creates several problems:

- Hard to reuse components in plugins, packages, and future SDKs. This forces `peerDependencies` on `antd`, which is suboptimal. Packages/plugins should NOT depend on the main app!
- No clear boundary between reusable components and app-specific UI.
- Difficult navigation and discoverability due to inconsistent folder structures.

We propose to extract a clean, flat, reusable set of core UI components into a dedicated package, collocated with theming utilities, to better support modular development.

### Proposed Change

We propose to:

1. **Create a new package:**

   ```
   @superset-ui/core/components
   ```

   Collocated with:

   ```
   @superset-ui/core/theme
   ```

2. **Move reusable components** (especially AntD wrappers and UI primitives) from `src/components/` into this package. 

   Criteria for inclusion:
   - Component does **not** depend on app-specific logic (Redux, routes, etc.)
   - Component **wraps or styles** base UI elements (AntD, `styled-components`, etc.)
   - Component is likely to be reused across plugins, embedded apps, or external SDKs

3. **Leave app-specific components** in `src/components/` if they:
   - Import from other parts of the application (e.g. `src/views`)
   - Are unlikely to be useful outside the main app

4. **Flat export structure:**

   All components and types are exported from a single unified namespace:
   ```ts
   import { Button, Tabs, type TabsProps } from '@superset-ui/core/components';
   ```

   We intentionally avoid nested folders like `layout/Tabs` or `inputs/Button` to keep things flat, discoverable, and easier to manage. If name collisions arise, we will handle them explicitly via namespacing. Tree specific `DataNode` becomes `TreeDataNode` for clarity and avoiding potential collisions.

5. **Initial integration via theming branch:**

   The Theming branch will lead this migration to minimize merge conflicts. Other efforts (e.g., plugin SDK) will coordinate to ensure parallel needs are addressed cleanly. We can coordinate with the Plugins / Extensions effort.

### New or Changed Public Interfaces

- New package: `@superset-ui/core/components`
- New top-level exports for all reusable components and types
- Existing imports from `src/components` will be migrated to the new namespace as part of the rollout
- Internal theming APIs from `@superset-ui/core/theme` will become shared as part of this effort

### New dependencies

[...]

PR #33593, opened 2025-05-27
chore: moving reusable components from `src/components` to `packages/superset-ui-core/src/components`

mistercrunch:
Screenshot 2025-05-29 at 2 07 15 PM

We've been talking about factoring out the reusable primitives from `src/components` to `packages/superset-ui-core/src/components` for a moment now. Followed most of what was stated here https://github.com/apache/superset/issues/33441

This might not be the final destination of where components will land. Final resting place of components might change to `/primitives` and/or outside of `core/`, but this surgery is the first big step in that direction.

Work here include:
* figuring out which components depend on the main app and should be collocated
* de-tangling components that should NOT depend on the main app but did
* sorted out imports, moving the proper imports to the right place, using peerDeps in the main package where packages are needed in both core/ and main app
* mostly unified linting rules across the packages
* [unfinished] flattening the components namespace, and generally bring all components and their types to a single namespace
* bubbling up the module tree using export * as much as possible
* improving typescript type specifications in many places
* specifying exports in @superset-ui/core as opposed to using @superset-ui/core/src  - unclear if I should move forward with this as it broke jest's mocking assumption, unclear if I push through or rollback this approach

TODO:
- fix jest tests, module resolution and mocks are broken RN and need attention
- guessing cypress e2e tests need attention


mistercrunch:
Oh wow, e2e/cypress just worked out of the box! INCREDIBLE.

On the topic of jest, it's failing around package resolution, related to have more clear/restricted `exports` in `superset-ui/core`. Wondering if I should push forward in that direction, or rollback on this particular topic. Wondering what other people think.

mistercrunch:
related -> https://github.com/apache/superset/issues/33441

.github/workflows/superset-frontend.yml
```diff
@@ -121,24 +121,6 @@ jobs:
           files: merged-output/coverage-summary.json
           slug: apache/superset
 
-  core-cover:
```
mistercrunch, line 124: NOTE: this forced 100% test coverage across `core/`. Dropped it as we're bringing more things in here. Could be nice to keep the 100% requirement on parts of it, but gets in the way of velocity. Test now covered as part of the main repo tests. We had double test coverage here before which is a bit wasteful.

geido:
Not sure if this was related to this branch, but when visiting "Query history" for the first time under SQL Lab, it was showing a huge loading icon on the left of the screen that went away and did not come back when refreshing.

superset-frontend/packages/superset-ui-core/src/components/index.ts
```diff
@@ -17,5 +17,160 @@
...
+export { AntdThemeProvider } from './AntdThemeProvider';
+export {
+  ConfirmStatusChange,
+  type ConfirmStatusChangeProps,
+} from './ConfirmStatusChange';
+export { Alert, type AlertProps } from './Alert';
+export { CertifiedBadge } from './CertifiedBadge';
+export * from './Icons';
+export * from './Timer';
+export {
```
geido, line 35: Is it now a good time to go with the approach that we wanted to implement where we are just re-exporting everything from here?

`export * from '{component}';`

geido (approved):
Approving. Hopefully we can ensure the plugin publication process is working before merging.

PR #33626, opened 2025-05-30
fix: Update dataset's last modified date from column/metric update

Vitor-Avila:
### SUMMARY
The `last_modified` value for datasets was not getting updated in case only metric(s) and/or column(s) changed. I suspect the issue started with the introduction of the `@transaction` decorator + the way we handle these updates:
- Before [SIP-99B](https://github.com/apache/superset/issues/25108), we would `commit` after these changes -- [ref](https://github.com/apache/superset/commit/8fb8199a55257ae5121300aa3bd28d14d327bc33#diff-1fabcc7af69ff8ee94b640857ad3552402286c735a5a770b9d14f204774c8f60L269-L322).
- We're currently using `bulk_insert_mappings` for the update which does not integrate well with SQLAlchemy's state management (and therefore does not trigger an `after_update`) -- [ref](https://github.com/apache/superset/blob/master/superset/daos/dataset.py#L233-L256).

**Before**

https://github.com/user-attachments/assets/9187b29a-bc63-4c6e-9ee7-8db8254cd4bd

**After**

https://github.com/user-attachments/assets/67a82bcc-0f57-4450-a540-1968c1526201

### TESTING INSTRUCTIONS
1. Apply a label to a column, and confirm the `last_modified` value for the dataset gets updated.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -49,7 +49,6 @@ from sqlalchemy import (
     String,
     Table as DBTable,
     Text,
-    update,
 )
 from sqlalchemy.engine.base import Connection
 from sqlalchemy.ext.declarative import declared_attr
@@ -2023,21 +2022,6 @@ class SqlaTable(
         security_manager.dataset_before_update(mapper, connection, target)
 
     @staticmethod
-    def update_column(  # pylint: disable=unused-argument
-        mapper: Mapper, connection: Connection, target: SqlMetric | TableColumn
-    ) -> None:
-        """
-        :param mapper: Unused.
-        :param connection: Unused.
-        :param target: The metric or column that was updated.
-        """
-        session = inspect(target).session  # pylint: disable=disallowed-name
-
-        # Forces an update to the table's changed_on value when a metric or column on the  # noqa: E501
-        # table is updated. This busts the cache key for all charts that use the table.
-        session.execute(update(SqlaTable).where(SqlaTable.id == target.table.id))
-
-    @staticmethod
     def after_insert(
         mapper: Mapper,
         connection: Connection,
@@ -2072,8 +2056,6 @@ class SqlaTable(
 sa.event.listen(SqlaTable, "before_update", SqlaTable.before_update)
 sa.event.listen(SqlaTable, "after_insert", SqlaTable.after_insert)
 sa.event.listen(SqlaTable, "after_delete", SqlaTable.after_delete)
-sa.event.listen(SqlMetric, "after_update", SqlaTable.update_column)
-sa.event.listen(TableColumn, "after_update", SqlaTable.update_column)
 
 RLSFilterRoles = DBTable(
     "rls_filter_roles",
--- a/superset/daos/dataset.py
+++ b/superset/daos/dataset.py
@@ -21,6 +21,7 @@ from datetime import datetime
 from typing import Any
 
 import dateutil.parser
+from sqlalchemy import update
 from sqlalchemy.exc import SQLAlchemyError
 
 from superset.connectors.sqla.models import SqlaTable, SqlMetric, TableColumn
@@ -184,15 +185,24 @@ class DatasetDAO(BaseDAO[SqlaTable]):
         """
 
         if item and attributes:
+            force_update: bool = False
             if "columns" in attributes:
                 cls.update_columns(
                     item,
                     attributes.pop("columns"),
                     override_columns=bool(attributes.get("override_columns")),
                 )
+                force_update = True
 
             if "metrics" in attributes:
                 cls.update_metrics(item, attributes.pop("metrics"))
+                force_update = True
+
+            # columns and metrics are stored in separate tables so if they change we
+            # force an update to the dataset too to ensure the last_updated_timestamp
+            # is updated (it's possible no other dataset attributes are changed).
+            if force_update:
+                db.session.execute(update(SqlaTable).where(SqlaTable.id == item.id))
 
         return super().update(item, attributes)
 
--- a/superset/daos/dataset.py
+++ b/superset/daos/dataset.py
@@ -21,7 +21,6 @@ from datetime import datetime
 from typing import Any
 
 import dateutil.parser
-from sqlalchemy import update
 from sqlalchemy.exc import SQLAlchemyError
 
 from superset.connectors.sqla.models import SqlaTable, SqlMetric, TableColumn
@@ -198,11 +197,8 @@ class DatasetDAO(BaseDAO[SqlaTable]):
                 cls.update_metrics(item, attributes.pop("metrics"))
                 force_update = True
 
-            # columns and metrics are stored in separate tables so if they change we
-            # force an update to the dataset too to ensure the last_updated_timestamp
-            # is updated (it's possible no other dataset attributes are changed).
             if force_update:
-                db.session.execute(update(SqlaTable).where(SqlaTable.id == item.id))
+                attributes["changed_on"] = datetime.now()
 
         return super().update(item, attributes)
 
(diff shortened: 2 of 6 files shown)


anthonyhungnguyen:
We're at Geotab and gonna get this fixed soon!

superset/daos/dataset.py
```diff
@@ -184,15 +185,24 @@ def update(
...
+                force_update = True
 
             if "metrics" in attributes:
                 cls.update_metrics(item, attributes.pop("metrics"))
+                force_update = True
+
+            # columns and metrics are stored in separate tables so if they change we
+            # force an update to the dataset too to ensure the last_updated_timestamp
+            # is updated (it's possible no other dataset attributes are changed).
+            if force_update:
```
michael-s-molina, line 204: @Vitor-Avila Why not be explicit about this operation and update the `changed_on` field on the dataset? It would avoid the whole comments and the cognitive load to understand that `changed_on` will be automatically updated once the record is changed.

Vitor-Avila: @michael-s-molina are you suggesting we always do it? 

Technically we could. I opted for this flow because we only need to worry with this in case columns and/or metrics were updated (and no direct dataset attribute was). I don't think there would be a side-effect of always doing, it's just one extra query.

betodealmeida: I think he's suggesting that instead of updating the last modified column as a side effect like you do on line 205, do something that makes the intent clear, something like:

```python
dataset = db.session.query(SqlaTable).get(item.id)
dataset.changed_on = XXX
```

michael-s-molina: Exactly @betodealmeida!

Vitor-Avila: ohhhh, I see! that makes perfect sense. Let me work on that. Thanks @betodealmeida and @michael-s-molina

Vitor-Avila: alright, I added `changed_on` to `attributes` so it's processed as any other column.

superset/daos/dataset.py
```diff
@@ -184,15 +184,21 @@ def update(
...
                     override_columns=bool(attributes.get("override_columns")),
                 )
+                force_update = True
 
             if "metrics" in attributes:
                 cls.update_metrics(item, attributes.pop("metrics"))
+                force_update = True
+
+            if force_update:
+                attributes["changed_on"] = datetime.now()
```
betodealmeida: I think we want `datetime.utcnow` here. I'm in EST:

```python
>>> print(datetime.now(), datetime.utcnow())
2025-06-02 16:29:53.651504 2025-06-02 20:29:53.652092
>>>
```

```suggestion
                attributes["changed_on"] = datetime.utcnow()
```

Although I just checked the codebase and it seems like we have some columns where the default is `datetime.now()`. :(

Vitor-Avila: it seems that `.utcnow()` is duplicated, so I used `.now(tz=timezone.utc)`

Vitor-Avila: we'll need a follow up for other places tho

tests/integration_tests/datasets/api_tests.py
```diff
@@ -1188,6 +1190,10 @@ def test_update_dataset_create_column_and_metric(self):
             metric.pop("type_generic", None)
 
         data["result"]["metrics"].append(new_metric_data)
+
+        # Sleep to ensure that the changed_on is updated
```
michael-s-molina: Optional: You can use `freezegun` to control time and make tests more accurate and efficient.

```
from freezegun import freeze_time

with freeze_time("2021-01-01T00:00:00Z"):
   ...
```

superset/connectors/sqla/models.py
```diff
@@ -2072,8 +2056,6 @@ def load_database(self: SqlaTable) -> None:
 sa.event.listen(SqlaTable, "before_update", SqlaTable.before_update)
 sa.event.listen(SqlaTable, "after_insert", SqlaTable.after_insert)
 sa.event.listen(SqlaTable, "after_delete", SqlaTable.after_delete)
-sa.event.listen(SqlMetric, "after_update", SqlaTable.update_column)
```
michael-s-molina, line 2075: @Vitor-Avila The previous event-based implementation was triggered for changes from any source. This means that any time `SqlMetric` or `TableColumn` was updated, from any part of the code, the timestamp was updated. Did we check to see if these models are updated outside the DAO?

superset/daos/dataset.py
```diff
@@ -184,15 +184,21 @@ def update(
...
                     override_columns=bool(attributes.get("override_columns")),
                 )
+                force_update = True
 
             if "metrics" in attributes:
                 cls.update_metrics(item, attributes.pop("metrics"))
+                force_update = True
+
+            if force_update:
+                attributes["changed_on"] = datetime.now(tz=timezone.utc)
```
michael-s-molina, line 201: Thanks 🙏🏼

tests/integration_tests/datasets/api_tests.py
Vitor-Avila: @michael-s-molina thank you 🙏 I used that in another test, but here I intentionally want to confirm that the new value for `changed_on` is more recent than the old one, so this should be sufficient. I wanted to avoid mocking it to ensure the logic is properly applied.

superset/connectors/sqla/models.py
Vitor-Avila: @michael-s-molina the only other place I found that was modifying these is via import flows, which I manually tested and confirmed it is updating the `changed_on` as well, so I believe we're good. Other than that I found some migrations.

If we find other places in the future, happy to do a follow up tho.

tests/integration_tests/datasets/api_tests.py
michael-s-molina: You can do something like:

```
@freeze_time("2023-10-01")
def test_advance_time():
    # Check the initial frozen time
    assert datetime.now() == datetime(2023, 10, 1)
    
    # Advance time by one day
    freeze_time().tick()
    
    # Now it should reflect the new time
    assert datetime.now() == datetime(2023, 10, 2)
```

This will make your test run instantly instead of taking 3 seconds to execute.

michael-s-molina (approved):
I left a non-blocking comment. Thanks for addressing the comments.

tests/integration_tests/datasets/api_tests.py
betodealmeida: Oh yeah, we should avoid calling `sleep()` on tests. With freezegun you can do:

```python
 with freeze_time(initial_datetime) as frozen:
    do_something()
    frozen.tick(delta=timedelta(seconds=3))
    do_something_else()
```

betodealmeida (changes requested):
Let's remove the `sleep()` from the tests, it's a slippery slope — people will start copying the pattern and make the tests take even more time.

tests/integration_tests/datasets/api_tests.py
Vitor-Avila: oh, TIL -- I can work on that

Vitor-Avila: ok, this is done now! 🙌 I'll work on a follow up to remove `time.sleep()` from other places I found in the code. I'll probably combine that with a deprecation of `datetime.now()`

betodealmeida (approved):
Thanks!

Vitor-Avila:
thanks for taking the time to review this @michael-s-molina and @betodealmeida 🙏 I really appreciate the valuable feedback.

PR #33656, opened 2025-06-01
feat(chart): add toggle for percentage metric calculation mode in Table chart

LevisNgigi:
### SUMMARY
This PR introduces a new control in the Table chart visualization that allows users to configure how percentage metrics are calculated. Previously, percentage metrics were always computed based on the limited set of rows returned by the main query, which could lead to misleading results.
With this change, users can now choose between two modes:
Row limit (default) – Percentage values are calculated using only the data returned by the query, respecting the row limit.
All records – Percentage values are calculated using totals from the full dataset, ignoring the row limit. This mode triggers an auxiliary query to fetch the necessary totals, which are injected into post-processing for accurate contribution calculations.

https://github.com/user-attachments/assets/06e71bda-933b-4fc5-af68-cb13bc15158a

### TESTING INSTRUCTIONS
1) Open the Explore view for a Table chart.

2) Select a percentage metric and set a row limit (e.g., 10).

3) In the control panel, choose "Percentage metric calculation" → "All records".

4) Run the query and verify that percentage values reflect the entire dataset (not just the top 10 rows).

5) Switch back to "Row limit" and confirm that percent values update to reflect only the visible rows.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ x ] Changes UI


superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -139,22 +141,35 @@ const buildQuery: BuildQuery<TableChartFormData> = (
...
         );
+
         metrics = removeDuplicates(
           metrics.concat(percentMetrics),
           getMetricLabel,
         );
-        postProcessing = [
-          {
+
+        if (calculationMode === 'all_records') {
```
geido, line 157: Can we use a type for `calculationMode` instead of relying on strings?

superset-frontend/plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -139,22 +141,35 @@ const buildQuery: BuildQuery<TableChartFormData> = (
...
+          postProcessing.push({
+            operation: 'contribution_with_totals',
+            options: {
+              columns: percentMetricLabels,
+              rename_columns: percentMetricLabels.map(m => `%${m}`),
+              totals: (formData.extra_form_data as Record<string, any>)
+                ?.contribution_totals?.totals,
+            },
+          } as unknown as PostProcessingRule);
```
geido: What is the reason of forcing this type via `unknown`? Should we tackle the typing problem instead?

geido (commented):
Hey @LevisNgigi a few first-pass comments. Can we please add test cases too?

superset/common/query_context_processor.py
```diff
@@ -741,24 +741,77 @@ def get_data(
 
         return df.to_dict(orient="records")
 
-    def get_payload(
+    def get_payload(  # noqa: C901
```
Antonio-RiveroMartnez: Is there a way we can make the function not raise this?

superset/common/query_context_processor.py
```diff
@@ -741,24 +741,77 @@ def get_data(
...
         self,
         cache_query_context: bool | None = False,
         force_cached: bool = False,
     ) -> dict[str, Any]:
         """Returns the query results with both metadata and data"""
+        # Handle empty queries case
+        if not self._query_context.queries:
```
Antonio-RiveroMartnez: Looking at the `buildQuery` this is something to be computed by `query_object -> exe_post_processing`. I don't  think we should bring these into the context processor which is a more abstract concept. Meaning, these changes would be introducing an implementation that's very specific for your use case into something that's broadly used across the app.

Is there a way to include your new post processor without affecting the context? how is the current one working without having to alter it?

LevisNgigi: Thank you for the review,the current 'contribution' post-processing step computes percentages based on the data already in the DataFrame (i.e. totals are computed from the same dataset and no external input needed).

Therefore, no external data (like auxiliary totals) needs to be injected in the context layer. The whole computation happens inside contribution(df, ...) at post-processing time.But for this we need the totals from the dataset to be able to calculate percentage based on all records not just the row limit,so trying to look for another way to do this.

betodealmeida: I agree with @Antonio-RiveroMartnez, the logic added here is specific to a single post-processing operation, and should live elsewhere.

betodealmeida (commented):
@LevisNgigi can you show the queries that are generated in both cases?

LevisNgigi:
> @LevisNgigi can you show the queries that are generated in both cases?

@betodealmeida For Row limit it is;SELECT global_sales AS global_sales, COUNT(*) AS count 
FROM main.video_game_sales GROUP BY global_sales ORDER BY count DESC
 LIMIT 10 OFFSET 0;
 
 For All records it is this; SELECT global_sales AS global_sales, COUNT(*) AS count 
FROM main.video_game_sales GROUP BY global_sales ORDER BY count DESC
 LIMIT 10 OFFSET 0;
 
SELECT COUNT(*) AS count 
FROM main.video_game_sales
 LIMIT 50000 OFFSET 0;

amaannawab923:
Went through the pr 
I had some input for this , there might be an easier solution for this... We can just pass a flag to contribution operator that it should use total of a metric value while Calculating the percentage instead of the Usual Sum of a column 


The current implementation in contribution.py is 
```

if orientation == PostProcessingContributionOrientation.COLUMN:
        numeric_df_view = numeric_df_view / numeric_df_view.values.sum(
            axis=0, keepdims=True
        )
        contribution_df[rename_columns] = numeric_df_view
        return contribution_df
```
        
 And then we can do it like this
 
```
 
 if orientation == PostProcessingContributionOrientation.COLUMN:
        if contribution_totals:
            # Use pre-calculated totals for division
            for col in numeric_df_view.columns:
                if col in contribution_totals:
                    numeric_df_view[col] = numeric_df_view[col] / contribution_totals[col] <- Using existing contribution_total pre calculated ones
        else:
            # Use existing column sum logic
            numeric_df_view = numeric_df_view / numeric_df_view.values.sum(
                axis=0, keepdims=True
            )
        contribution_df[rename_columns] = numeric_df_view
        return contribution_df       
```
   
 i agree with @betodealmeida  & @Antonio-RiveroMartnez that since query_context_processor is the center of query context processing it shouldnt be touched but instead we can reuse the existing contribution operator .... Let me know if that makes sense...But i guess reusing existing operator will reduce the code changes by a huge scale
        
 cc: @geido  @LevisNgigi

geido:
> SELECT COUNT(*) AS count
> FROM main.video_game_sales
> LIMIT 50000 OFFSET 0;

@LevisNgigi isn't the `LIMIT 50000` there going to create a problem for the calculation of actual totals if there are more rows available?

LevisNgigi:
> > SELECT COUNT(*) AS count
> > FROM main.video_game_sales
> > LIMIT 50000 OFFSET 0;
> 
> @LevisNgigi isn't the `LIMIT 50000` there going to create a problem for the calculation of actual totals if there are more rows available?

@geido that was a great catch.I have updated so that no row limit is applied for the totals. Now it looks like this. " SELECT COUNT(*) AS count 
FROM main.video_game_sales; "

superset-frontend/plugins/plugin-chart-table/src/buildQuery.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      },
+    });
+
+    const totalsQuery = queries[1];
+    expect(totalsQuery.metrics).toEqual(['count']);
+    expect(totalsQuery.columns).toEqual([]);
+    expect(totalsQuery.post_processing).toEqual([]);
+    expect(totalsQuery.row_limit).toBe(0);
+  });
+});
```
msyavuz: I think there is already a file called buildQuery.test.ts in `plugins/plugin-chart-table/test/buildQuery.test.ts`. Also, i am not sure about mocks here. Can't we test how the function directly?

LevisNgigi: Thanks I updated accordingly.

geido (approved):

Issue #25740, opened 2023-10-23
Endpoint /api/me/roles returns 401 "Not authorized" despite valid access token


larshelge:
I am running Superset 3.0.0 with Docker Compose. Trying to access the following API endpoint:

```
/api/me/roles
```

I have created an access token from `/api/v1/security/login` for a user with the Admin, Public and Gamma roles. I make a request for the mentioned API endpoint with the appropriate header.

```
Authorization: Bearer {access-token}
Accept: application/json
```

Various endpoints return 200 and a valid JSON payload, such as `/api/v1/dashboard` and `/api/v1/chart`. However, the 
`/api/me/roles` endpoint returns 401 "Not authorized". This blocks the embedded dashboard feature. Could this be a bug, or am I doing something wrong? I sense there is something special with this endpoint and authentication as it relates to the currently authenticated user.

The `/api/me/roles` endpoint returns the roles when I log in through the UI and load the endpoint in a web browser.

### Expected results

I expect the roles to be returned from the API with 200 OK status.

### Actual results

The endpoint returns 401 "Not authorized" despite the access token seemingly being valid.

### Environment

(please complete the following information):

- Chrome latest.
- Linux Ubuntu 22.04.
- Superset 3.0.0 on Docker Compose in production mode

### Flags

```
WTF_CSRF_ENABLED = False
TALISMAN_ENABLED = False
ENABLE_CORS = True
```

cbuffevant:
Still happening in 3.0.1

siddhartha8916:
You can try a fix around by hitting a get request to http://localhost:8088/login endpoint and sending the csrf_token obtained along with username and password as formData and then get the access token to request http://localhost:8088/api/v1/me/roles api

bryanjknight:
Still happening in 3.1.1, also blocking usage of embedded dashboards.  Unfortunately the workaround above won't work because the embed SDK won't do that for me

bryanjknight:
Digging into this more, it seems `/api/v1/security/login` does not set a session cookie; hence, the `/api/v1/me` and underlying calls fail b/c there's no session cookie

EuphoriaCelestial:
> Digging into this more, it seems `/api/v1/security/login` does not set a session cookie; hence, the `/api/v1/me` and underlying calls fail b/c there's no session cookie

I am facing the same issue
Normally, how can I get the session cookie?

siddhartha8916:
[...]

PR #33673, opened 2025-06-03
fix: clarify GUEST_TOKEN_JWT_AUDIENCE usage in the SDK

schollz:
### SUMMARY

Change to the wording of the SDK. There is little documentation around `GUEST_TOKEN_JWT_AUDIENCE` and having it set is crucial to having the SDK work. 

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/25740

### DIFF
--- a/superset-embedded-sdk/README.md
+++ b/superset-embedded-sdk/README.md
@@ -117,7 +117,7 @@ Example `POST /security/guest_token` payload:
 ```
 
 Alternatively, a guest token can be created directly in your app without interacting with the Superset API.
-To do this, you should update the  `GUEST_TOKEN_JWT_SECRET` 
+To do this, you should update the `GUEST_TOKEN_JWT_SECRET`
 in the Superset [config.py](https://github.com/apache/superset/blob/master/superset/config.py). Also set the
 `GUEST_TOKEN_JWT_AUDIENCE` variable that matches what is set for the `aud` in the JSON payload:
 
--- a/superset-embedded-sdk/README.md
+++ b/superset-embedded-sdk/README.md
@@ -116,8 +116,11 @@ Example `POST /security/guest_token` payload:
 }
 ```
 
-Alternatively, a guest token can be created directly in your app with a json like the following, and then signed
-with the secret set in configuration variable `GUEST_TOKEN_JWT_SECRET` (see configuration file config.py)
+Alternatively, a guest token can be created directly in your app without interacting with the Superset API.
+To do this, you should update the  `GUEST_TOKEN_JWT_SECRET` 
+in the Superset [config.py](https://github.com/apache/superset/blob/master/superset/config.py). Also set the
+`GUEST_TOKEN_JWT_AUDIENCE` variable that matches what is set for the `aud` in the JSON payload:
+
 ```
 {
   "user": {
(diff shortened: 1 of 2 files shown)


superset-embedded-sdk/README.md
```diff
@@ -139,6 +139,14 @@ with the secret set in configuration variable `GUEST_TOKEN_JWT_SECRET` (see conf
...
 ```
 
+In this example, the configuration file includes the following setting:
+
+```python
+GUEST_TOKEN_JWT_AUDIENCE="superset"
+```
+
+If you're using a different audience value, be sure to update the GUEST_TOKEN_JWT_AUDIENCE variable accordingly.
+
```
Vitor-Avila: Do we want to talk about the default value? Checking `superset/config.py`, it's declared like this:
``` python
GUEST_TOKEN_JWT_AUDIENCE: Callable[[], str] | str | None = None
```

Then it seems in the `security_manager` we have:
``` python
@staticmethod
def _get_guest_token_jwt_audience() -> str:
    audience = current_app.config["GUEST_TOKEN_JWT_AUDIENCE"] or get_url_host()
    if callable(audience):
        audience = audience()
    return audience
```

So it seems the URL host would be the default value if not specified in the config?

schollz: I can add it, but personally I think its a little confusing. `get_url_host()` returns the value of another configuration variable - `WEBDRIVER_BASEURL` which also has its own default value. For the sake of simplicity I'd lean towards a recommendation of making clear that setting `GUEST_TOKEN_JWT_AUDIENCE` is recommended for using the SDK.

Vitor-Avila: @schollz totally agree with you that it's confusing to fallback to `get_url_host()` (I was genuinely surprised when I found that lol). In that case, what do you think about intentionally instructing users to set this config in their `superset_config.py` file? Above the JSON example we have:

> Alternatively, a guest token can be created directly in your app with a json like the following, and then signed
with the secret set in configuration variable `GUEST_TOKEN_JWT_SECRET` (see configuration file config.py)

I'm thinking we could do something like:

Alternatively, a guest token can be encoded directly in your app without interacting with the Superset API. Set up a secret via the `GUEST_TOKEN_JWT_SECRET` config and the token audience via the `GUEST_TOKEN_JWT_AUDIENCE` in your `superset_config.py` file, then sign the token using `GUEST_TOKEN_JWT_SECRET`. Example payload:
```
<example JSON>
```
The `aud` value must match the `GUEST_TOKEN_JWT_AUDIENCE` value.


Just an example (might be a bit confusing) but thinking this way it's explicit that users should set values for these two configs.

schollz: Yeah that seems like a good idea. And along those lines I'll propose a small edit to `config.py` `GUEST_TOKEN_JWT_AUDIENCE` that calls out that it will default to `WEBDRIVER_BASEURL`. The `config.py` does serve as one of the best places of documentation so I'd love to keep that well informed too.

Vitor-Avila: awesome -- than you!

rusackas: Docs site PRs are warmly welcomed, if you want to add anything there :D

schollz: Thanks, I added a bit of clarity there too.

Vitor-Avila (approved):
Thank you so much for improving our docs!

rusackas:
Just needs some pre-commit touchups to pass. 

```
pre-commit install
```

```
pre-commit run --all-files
``` 

I can try to touch up the PR by hand if needed... the CI errors tell you what to fix ;)

schollz:
Thanks @rusackas, working through it now.

PR #33679, opened 2025-06-04
fix: optimize catalog permission sync when importing dashboards

### LINKED ISSUE #32993: Catalog permission syncs severely degrades dashboard import performance
### Bug description

The addition of catalogs in Superset 4.1.0 has resulted in significantly worse dashboard import performance. We're seeing dashboard import time increase from <10 seconds to >3 minutes.

We're currently using Postgres (RDS) to serve embedded dashboards in our multi-tenanted application. As we are multi-tenanted, we have one catalog/database per tenant.

When importing dashboards, Superset now iterates through all catalog names to sync permissions. As we have hundreds of catalogs/dbs, this sync adds a couple of minutes per import. 

We need to import a default set of dashboards for each tenant (with their own unique set of datasets pointing to their individual databases). The sync permission step needs to run for each dashboard import.

This results in NxN syncs (in our case, N=200, so on the order of 40k syncs). So, the time to upload dashboards for all our tenants has exploded from 30-60 minutes to over 10 hours.

### Expected Behavior
I see there already exists a flag to disable catalog discovery - `allow_multi_catalog`. I propose that if this flag is set to false, then Superset should only sync permissions for the default catalog. 

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

arafoperata:
### SUMMARY
Following up from https://github.com/apache/superset/pull/33000. The previous PR fixed catalog permission syncing for databases like Postgres, which don't support multiple catalogs in one instance. However, permissions were still being synced for all the catalogs when creating a database connection (e.g. during a dashboard import).

This PR applies the same logic as the previous fix to `CreateDatabaseCommand`.

Fixes https://github.com/apache/superset/issues/32993.

N/A

### TESTING INSTRUCTIONS
Added unit tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/database/utils.py
+++ b/superset/commands/database/utils.py
@@ -57,10 +57,21 @@ def add_permissions(database: Database, ssh_tunnel: SSHTunnel | None) -> None:
     """
     # TODO: Migrate this to use the non-commiting add_pvm helper instead
     if database.db_engine_spec.supports_catalog:
-        catalogs = database.get_all_catalog_names(
-            cache=False,
-            ssh_tunnel=ssh_tunnel,
-        )
+        # Adding permissions to all catalogs (and all their schemas) can take a long
+        # time (minutes, while importing a chart, eg). If the database does not
+        # support cross-catalog queries (like Postgres), and the multi-catalog
+        # feature is not enabled, then we only need to add permissions to the
+        # default catalog.
+        if (
+            database.db_engine_spec.supports_cross_catalog_queries
+            or database.allow_multi_catalog
+        ):
+            catalogs = database.get_all_catalog_names(
+                        cache=False,
+                        ssh_tunnel=ssh_tunnel,
+                    )
+        else:
+            catalogs = {database.get_default_catalog()}
 
         for catalog in catalogs:
             security_manager.add_permission_view_menu(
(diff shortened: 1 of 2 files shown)


superset/commands/database/utils.py
```diff
@@ -57,10 +57,21 @@ def add_permissions(database: Database, ssh_tunnel: SSHTunnel | None) -> None:
...
+        if (
+            database.db_engine_spec.supports_cross_catalog_queries
+            or database.allow_multi_catalog
+        ):
+            catalogs = database.get_all_catalog_names(
+                cache=False,
+                ssh_tunnel=ssh_tunnel,
+            )
+        else:
+            catalogs = {database.get_default_catalog()}
```
betodealmeida, line 74: I just realized we probably want to pass the `ssh_tunnel` to `get_default_catalog`, since for some DB engine specs we need to make a connection and run a query in order to figure out the default catalog (it can't be inferred from the SQLAlchemy URI). But we can do that in the future when we need it.

betodealmeida (approved):
Awesome, thanks for fixing this!

PR #33682, opened 2025-06-04
fix: Dataset currency

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/33303. 

Even though that PR changed the `currency` column in the `sql_metrics` table to `json`, some currency configurations were stringified multiple times, so a single `json.loads()` would still return a string. Also, the client was still sending the currency configuration as a string, so the problem persisted after the migration. 

On the video below I demonstrate that:
1. The client is still stringifying the currency config before storing it via the API.
2. Due to that, the existing configuration would not persist in the UI for future edits.
3. Trying to "overwrite" the existing config (that's not visible in the UI) throws an error.
4. Once you face the error, editing the dataset again (without touching the currency config) would stringify it again.

https://github.com/user-attachments/assets/fe824719-74ad-4320-95b0-903c68351c11

This PR performs the following changes:
1. Updates the `buildPayload` logic when updating a dataset so that metrics are no longer stringified.
2. Adds a `DatasetMetricCurrencyPutSchema` to ensure new metrics are stored as JSON (API-level validation).
3. Creates a new migration to recursively migrate existing metrics to JSON.
4. Redefines `currencyFormats` at the client level. This is needed by the charts to apply currency configs from the dataset level. 

Video above can show the previous experience. I'll add a video once this is green with the current behavior.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/connectors/sqla/models.py
```diff
@@ -312,6 +312,10 @@ def explore_url(self) -> str:
     def column_formats(self) -> dict[str, str | None]:
         return {m.metric_name: m.d3format for m in self.metrics if m.d3format}
 
+    @property
+    def currency_formats(self) -> dict[str, dict[str, str | None] | None]:
+        return {m.metric_name: m.currency for m in self.metrics if m.currency}
+
```
betodealmeida: This is just sending the same thing that we're already sending. This data structure should be built by the frontend if it needs the currencies in a single dictionary.

superset-frontend/src/components/Chart/Chart.tsx
```diff
@@ -178,6 +182,17 @@ class Chart extends PureComponent<ChartProps, {}> {
...
+      // eslint-disable-next-line no-param-reassign
+      props.datasource.currencyFormats = Object.fromEntries(
+        props.datasource.metrics
+          .filter(
+            (metric): metric is Metric & { currency: Currency } =>
+              metric.currency !== undefined,
+          )
+          .map(metric => [metric.metric_name, metric.currency]),
+      ) as Record<string, Currency>;
+    }
```
Vitor-Avila: I know that modifying `chartProps` here at this stage is not the most elegant solution, but I think it's the DRYer way. If we don't do it here, then in every `transformProps` we would have to add similar logic (or call a util that does this).

Anyways, open to feedback! There could be another option I couldn't find.

superset/connectors/sqla/models.py
Vitor-Avila: ok, I managed to get rid of this! 🙌

superset-frontend/src/components/Chart/Chart.tsx
kgabryje: I think the datasource passed to SuperChart comes from Redux. Could we perhaps set the currencyFormats at hydration stage?

Vitor-Avila: that's a great suggestion! Working on that now

superset-frontend/src/explore/actions/hydrateExplore.ts
```diff
@@ -97,6 +98,14 @@ export const hydrateExplore =
...
 
     const initialDatasource = dataset;
+    initialDatasource.currency_formats = Object.fromEntries(
+      (initialDatasource.metrics ?? [])
+        .filter(metric => metric.currency !== undefined)
+        .map((metric): [string, Currency] => [
+          metric.metric_name,
+          metric.currency!,
+        ]),
+    );
```
betodealmeida: Oh, nice!

superset/migrations/versions/2025-06-06_00-39_363a9b1e8992_convert_metric_currencies_from_str_to_json.py
```diff
@@ -0,0 +1,91 @@
...
+                    logger.error(
+                        f"Error loading metric {metric.metric_name} as json: {e}"
+                    )
+                    metric.currency = {}
+                    break
+            else:
+                break
+
+
+def downgrade():
```
betodealmeida, line 75: Since this migration is done after mine, I think we want the downgrade to be a no-op, instead of storing as string. The reasoning is that after my migration we expect all currency to be JSON, the only reason they're not is because the frontend keeps stringifying. So the way I see it, the `upgrade` fixes the problem of values being strings instead of objects, and the `downgrade` shouldn't do anything.

tests/integration_tests/datasets/api_tests.py
```diff
@@ -1462,6 +1462,29 @@ def test_update_dataset_update_metric_duplicate(self):
...
+            "metrics": [
+                {
+                    "metric_name": "test",
+                    "expression": "COUNT(*)",
+                    "currency": '{"symbol": "USD", "symbolPosition": "suffix"}',
+                },
+            ]
+        }
+        rv = self.put_assert_metric(uri, data, "put")
+        assert rv.status_code == 422
```
betodealmeida, line 1484: ❤️

betodealmeida (commented):
Looks good, just a comment on the `downgrade`!

superset/migrations/versions/2025-06-06_00-39_363a9b1e8992_convert_metric_currencies_from_str_to_json.py
Vitor-Avila: that makes sense to me! I wasn't sure if that would be acceptable, but I agree with that. I'll just leave a comment there.

betodealmeida: Yeah, the reason is that this is not a proper migration, but instead custom logic to fix an issue.

betodealmeida (approved):

superset-frontend/src/components/Datasource/DatasourceModal.tsx
```diff
@@ -152,9 +152,7 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
             metric_name: metric.metric_name,
             metric_type: metric.metric_type,
             d3format: metric.d3format || null,
-            currency: !isDefined(metric.currency)
-              ? null
-              : JSON.stringify(metric.currency),
+            currency: !isDefined(metric.currency) ? null : metric.currency,
```
eschutho: can this just be 
```suggestion
            currency: metric.currency,
```
or is there a particular reason that we need to change undefined to null?

superset-frontend/src/explore/actions/hydrateExplore.ts
```diff
@@ -97,6 +98,14 @@ export const hydrateExplore =
     }
 
     const initialDatasource = dataset;
+    initialDatasource.currency_formats = Object.fromEntries(
+      (initialDatasource.metrics ?? [])
+        .filter(metric => metric.currency !== undefined)
```
eschutho: probably relates to the earlier comment- this won't pick up null values, so maybe 
```suggestion
        .filter(metric => !!metric.currency)
```
would filter out null and undefined values. (and false and 0, etc if that's what you want)

superset-frontend/src/explore/actions/hydrateExplore.ts
```diff
@@ -97,6 +98,14 @@ export const hydrateExplore =
...
+        .map((metric): [string, Currency] => [
+          metric.metric_name,
+          metric.currency!,
```
eschutho, line 106: I think the reason why typescript may have been complaining is because the value could still be null.

superset-frontend/src/hooks/apiResources/dashboards.ts
```diff
@@ -43,7 +43,18 @@ export const useDashboardCharts = (idOrSlug: string | number) =>
...
 export const useDashboardDatasets = (idOrSlug: string | number) =>
-  useApiV1Resource<Datasource[]>(`/api/v1/dashboard/${idOrSlug}/datasets`);
+  useTransformedResource(
+    useApiV1Resource<Datasource[]>(`/api/v1/dashboard/${idOrSlug}/datasets`),
+    datasets =>
+      datasets.map(dataset => ({
+        ...dataset,
+        currencyFormats: Object.fromEntries(
+          (dataset.metrics ?? [])
+            .filter(metric => metric.currency !== undefined)
```
eschutho: same here, I would also look for null

superset/migrations/versions/2025-06-06_00-39_363a9b1e8992_convert_metric_currencies_from_str_to_json.py
```diff
@@ -0,0 +1,84 @@
...
+
+def upgrade():
+    bind = op.get_bind()
+    session = db.Session(bind=bind)
+    currency_configs = session.query(SqlMetric).filter(SqlMetric.currency.isnot(None))
+    for metric in paginated_update(
+        currency_configs,
+        lambda current, total: logger.info((f"Upgrading {current}/{total} metrics")),
+    ):
+        while True:
```
eschutho, line 61: Is this while loop necessary? The only reason it would run again is if metric.currency is parsed, but then it would break anyway, and be a noop? Or am I missing something?

betodealmeida: Unfortunately it is, there were a few cases where the currency was processed multiple times by `JSON.stringify` when editing, resulting in double and triple stringified objects.

Vitor-Avila: @eschutho yeah it was getting overly stringified by the client as you edit the dataset. I demonstrated it in the video, you could end up with something like `"\"\\\"{\\\\\\\"symbolPosition\\\\\\\":\\\\\\\"prefix\\\\\\\",\\\\\\\"symbol\\\\\\\":\\\\\\\"GBP\\\\\\\"}\\\"\""` and I imagine even more "nested".

superset-frontend/src/components/Datasource/DatasourceModal.tsx
Vitor-Avila: @eschutho I'm actually not sure, here I'm just removing `JSON.stringify()` from the previous logic. I'll see if I can do some more manual testing around this.

superset/migrations/versions/2025-06-06_00-39_363a9b1e8992_convert_metric_currencies_from_str_to_json.py
eschutho: Oh, I see.. that makes sense then.

superset-frontend/src/explore/actions/hydrateExplore.ts
Vitor-Avila: this is a great catch! I manually tested and couldn't find any issues, and I think this is aligned with back when `currency_formats` was built by the backend: `{m.metric_name: m.currency_json for m in metrics if m.currency_json}`

superset-frontend/src/explore/actions/hydrateExplore.ts
Vitor-Avila: Even with the previous change, I had to keep this, unfortunately.

superset-frontend/src/hooks/apiResources/dashboards.ts
Vitor-Avila: Done!

PR #33693, opened 2025-06-04
fix: Do not convert dataset changed_on to UTC

Vitor-Avila:
### SUMMARY
In https://github.com/apache/superset/pull/33626, we decided to store the dataset's `changed_on` attribute in UTC, to ensure consistency. This introduced an issue to calculate the `changed_on_humanized` property (`now - item.changed_on`). In order to fix this calculation for new edits, we would have to update it to `now.utc() - item.changed_on` which then would break the calculation for items that were modified before this modification.

Ideally, a long-term solution would be to implement timezone support at the DB level, and then migrate existing columns to use the timezone from the server. This however is not aligned with https://github.com/apache/superset/pull/33626 original's scope, so this PR just reverts that to store the `changed_on` using `now()`.

No UI changes.

### TESTING INSTRUCTIONS
N/A.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/daos/dataset.py
+++ b/superset/daos/dataset.py
@@ -17,7 +17,7 @@
 from __future__ import annotations
 
 import logging
-from datetime import datetime, timezone
+from datetime import datetime
 from typing import Any
 
 import dateutil.parser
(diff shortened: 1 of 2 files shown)


betodealmeida (approved):
Thanks! Sorry for the extra work in the original PR!

Vitor-Avila:
> Thanks! Sorry for the extra work in the original PR!

No worries! It was def a good idea and something we should implement. It's just going to take a little bit more work :P

PR #33700, opened 2025-06-05
fix(compose): environment entries in compose*.yml override values in docker/.env

denodo-research-labs:
### SUMMARY

Docker compose files (`docker-compose.yml`, `docker-compose-non-dev.yml`, `docker-compose-image-tag.yml`) contain `environment` entries for several containers, establishing values for env variables that are also included in `docker/.env`.

But due to [the way environment variable precedence works in Docker](https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/), any env variable merely listed in the `.yml` files –even if no value is specified– will overwrite any values coming from the `env-file` entries (`docker/.env` and `docker/.env-local`).

This means that if users set, for instance, `SUPERSET_LOAD_EXAMPLES=no` in `docker/.env` or `docker/.env-local` (as explained in the documentation), this value **will be ignored** and examples will be loaded anyway, because the `docker-compose-*.yml` contain this:
v
### TESTING INSTRUCTIONS

  - Set `SUPERSET_LOAD_EXAMPLES=no` in `docker/.env-local`.
  - Start with `docker compose ... up -d`
  - Examples will be loaded (and container `superset_init` will take a long time to start, which is inconvenient).

### ADDITIONAL INFORMATION

- [X] Has associated issue: #31747 (`environment` entries were added here).
- [ ] Required feature flags:
- [ ] Changes UI

### ADDITIONAL COMMENTS

Please note the above would not happen if `docker/.env` was in fact `.env`, and this may cause some confusion.

The reason is, the `.env` file present in the docker-compose root folder will be used for replacing any environment variables **used in the `.yml` file itself**. So for instance, if we had a `.env` file with:

```
SUPERSET_LOAD_EXAMPLES=no
```
Then this line in `docker-compose.yml` would work as expected, setting the value of the variable to `no`:
```yml
environment:
  SUPERSET_LOAD_EXAMPLES: "${SUPERSET_LOAD_EXAMPLES:-yes}"
```
But there is a difference between `.env` and Docker compose's `env-file`'s specified for each of the containers. The former applies on the `.yml` file, while the latter applies exclusively _inside the container_, but taking less precedence than the former.

So the fact that those files in the `docker/` folder are called `.env` **is misleading**. Those are not `.env` files, but instead `env-file`s (yes, subtle 😄).

So, **proposal**: Would you consider renaming those files inside the `docker/` folder to something like (just a suggestion):
  * `docker/.env` -> `docker/environment`
  * `docker/.env-local` -> `docker/environment-local`

(Please note renaming these would mean also changes in the public documentation, and in the `.gitignore` file.)

If it makes sense for you, we could provide an additional PR for that.

### DIFF
--- a/docker-compose-image-tag.yml
+++ b/docker-compose-image-tag.yml
@@ -65,8 +65,6 @@ services:
       superset-init:
         condition: service_completed_successfully
     volumes: *superset-volumes
-    environment:
-      SUPERSET_LOG_LEVEL: "${SUPERSET_LOG_LEVEL:-info}"
 
   superset-init:
     image: *superset-image
@@ -86,9 +84,6 @@ services:
     volumes: *superset-volumes
     healthcheck:
       disable: true
-    environment:
-      SUPERSET_LOAD_EXAMPLES: "${SUPERSET_LOAD_EXAMPLES:-yes}"
-      SUPERSET_LOG_LEVEL: "${SUPERSET_LOG_LEVEL:-info}"
 
   superset-worker:
     image: *superset-image
@@ -111,8 +106,6 @@ services:
           "CMD-SHELL",
           "celery -A superset.tasks.celery_app:app inspect ping -d celery@$$HOSTNAME",
         ]
-    environment:
-      SUPERSET_LOG_LEVEL: "${SUPERSET_LOG_LEVEL:-info}"
 
   superset-worker-beat:
     image: *superset-image
@@ -131,8 +124,6 @@ services:
     volumes: *superset-volumes
     healthcheck:
       disable: true
-    environment:
-      SUPERSET_LOG_LEVEL: "${SUPERSET_LOG_LEVEL:-info}"
 
 volumes:
   superset_home:
--- a/docker-compose-non-dev.yml
+++ b/docker-compose-non-dev.yml
@@ -71,8 +71,6 @@ services:
       superset-init:
         condition: service_completed_successfully
     volumes: *superset-volumes
-    environment:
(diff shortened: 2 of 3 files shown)


mistercrunch (approved):
Oh wow this is super confusing. Trying to think about what's best to do to insure backwards compatibility where it makes sense and make it easy for people to figure out how to use docker-compose properly.

I think the PR as is restores the expected behavior, where people can either specify their env vars in their shell, or defaults get set in `.env-local` if it exists or `.env` if it doesn't. Only minor downside is that it's a bit hard to find env vars you may want to alter (users have to find the  .env/.env-local ref in the docker-compose and go inspect what's in there, which is ok by me).

Looking at the docs, I think it's clear enough as is today -> 
Screenshot 2025-06-06 at 2 22 08 PM

mistercrunch:
This LGTM overall, I double checked that the defaults are/were the same between what we're de-referencing here in `docker-compose.*.yml` and in the default `docker/.env` file.

mistercrunch:
Thanks for the contribution btw - lots of forensics involved here!

denodo-research-labs:
Thanks for merging.

> Only minor downside is that it's a bit hard to find env vars you may want to alter

Agreed that it is very confusing. First impulse is to use the `environment` section of the `*.yml` to _document_ the variables that should be present in the `env-file`s and give them a default value, but unfortunately this is not how it works. Also, the similarity in name between the `.env` file and the `env-file` property does not help when they are configuration steps applied so differently.

PR #33701, opened 2025-06-05
chore: update sqlglot dialect map

betodealmeida:
### SUMMARY

Update the DB engine spec -> sqlglot dialect map.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql/parse.py
+++ b/superset/sql/parse.py
@@ -58,6 +58,7 @@ SQLGLOT_DIALECTS = {
     # "databend": ???
     "databricks": Dialects.DATABRICKS,
     # "db2": ???
+    # "denodo": ???
     # "dremio": ???
     "drill": Dialects.DRILL,
     "druid": Dialects.DRUID,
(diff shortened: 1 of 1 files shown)


superset/sql/parse.py
```diff
@@ -72,14 +73,18 @@
...
     # "impala": ???
-    # "kustokql": ???
+    # "kustosql": ???
     # "kylin": ???
+    "mariadb: ": Dialects.MYSQL,
+    "motherduck": Dialects.DUCKDB,
     "mssql": Dialects.TSQL,
     "mysql": Dialects.MYSQL,
     "netezza": Dialects.POSTGRES,
+    "oceanbase": Dialects.MYSQL,
```
Vitor-Avila, line 83: I think this should be ok, googled it and it seems they're aiming for a perfect replacement for MySQL so it should be fully compatible, but who knows 🤷 Don't think it would hold the PR just mentioning

superset/sql/parse.py
```diff
@@ -72,14 +73,18 @@
...
     # "ocient": ???
     # "odelasticsearch": ???
     "oracle": Dialects.ORACLE,
+    "parseable": Dialects.POSTGRES,
```
Vitor-Avila, line 87: ditto here

Vitor-Avila (approved):

superset/sql/parse.py
betodealmeida: Yeah, that's my understanding. I think that even if these are not perfect matches it's better than using the generic parser.

PR #33703, opened 2025-06-05
fix(theming): Fix visual regressions from theming P4

EnxDev:
### SUMMARY

This pull request makes minor UI tweaks to address regressions caused by the theming project.
The goal is to restore visual consistency and ensure components match the intended design.
No functional logic has been impacted.

- Before
before-2

- After
after-2

- Before
before-1

- After
after-1

- Before
before-9

- After
after-9

- Before
before-3

- After
after-3

- Before
before-10

- After
after-10

- Before
https://github.com/user-attachments/assets/6a2de66c-1c01-47ad-b027-2e1f34cba7a3

- After
https://github.com/user-attachments/assets/0771ff27-7efe-4da6-92ed-e3e01283d9b1

- Before
before

- After
after

### TESTING INSTRUCTIONS

- All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/components/Datasource/DatasourceEditor.jsx
```diff
@@ -1245,7 +1265,7 @@ class DatasourceEditor extends PureComponent {
                           <Icons.ExportOutlined
                             iconSize="s"
                             css={theme => ({
-                              color: theme.colors.primary.dark1,
+                              color: theme.colors.primary.light5,
```
msyavuz: Can we use first level tokens instead?

superset-frontend/src/components/Datasource/DatasourceEditor.jsx
```diff
@@ -1271,9 +1291,6 @@ class DatasourceEditor extends PureComponent {
                         </Button>
                       </div>
                     }
-                    errorMessage={
-                      this.props.database?.error && this.renderSqlErrorMessage()
```
msyavuz, line 1348: Eslint should catch this but renderSqlErrorMessage is unused now. Also there is some additional stuff rendered in that function that is lost with this. Maybe we should modify that instead?

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FilterTitlePane.tsx
```diff
@@ -99,7 +99,7 @@ const FilterTitlePane: FC<Props> = ({
           alignItems: 'flex-start',
           paddingTop: theme.sizeUnit * 3,
           position: 'sticky',
-          bottom: theme.sizeUnit,
+          marginBottom: theme.sizeUnit * 3,
```
msyavuz: Is this intended?

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigModal.tsx
```diff
@@ -76,6 +76,7 @@ const StyledModalWrapper = styled(StyledModal)<{ expanded: boolean }>`
 
   .ant-modal-body {
     padding: 0px;
+    overflow: unset;
```
msyavuz: This causes an issue with filter modal where in smaller screens buttons are overflowing, similar issue might exist in other places where `overflow: unset` is used:
image

superset-frontend/src/components/Datasource/DatasourceEditor.jsx
EnxDev: sure

superset-frontend/src/components/Datasource/DatasourceEditor.jsx
EnxDev: Good catch!

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigModal.tsx
EnxDev: I had removed it from there to place it somewhere else, but it doesn’t seem to work well across all browsers as I had hoped. I’ve added it back there and in other places, and slightly reduced the width of the left section; it seems to be working better now.

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FilterTitlePane.tsx
msyavuz: I don't think it is. It causes buttons to not be sticky.
Before:
image

After:
image

superset-frontend/src/components/Datasource/DatasourceEditor.test.jsx
```diff
@@ -132,7 +132,7 @@ describe('DatasourceEditor', () => {
     await userEvent.type(inputDtmFormat, 'test');
     await userEvent.type(inputCertifiedBy, 'test');
     await userEvent.type(inputCertDetails, 'test');
-  });
+  }, 60000); // 60 seconds timeout to avoid timeouts
```
msyavuz: 60 seconds is a bit much, if we decide to increase timeout let's do it in smaller increments

superset-frontend/src/components/Datasource/Field.tsx
```diff
@@ -72,10 +72,20 @@ export default function Field<V>({
   const extra = !compact && description ? description : undefined;
   const infoTooltip =
     compact && description ? (
-      <Tooltip id="field-descr" placement="right" title={description}>
-        <Icons.InfoCircleFilled
+      <Tooltip
+        css={css`
+          color: ${theme.colors.grayscale.base};
```
msyavuz: Can we move all of `colors.` to appropriate first level tokens? We will need to refactor these in the near future anyways.

superset-frontend/src/components/Datasource/DatasourceEditor.test.jsx
EnxDev: I added a 60-second timeout, exactly like it was done in the test right after the one called "can delete columns".
I believe this is just a temporary fix to avoid blocking the tests. The entire test likely needs to be reviewed; for example, some await statements that have no effect should probably be removed.

superset-frontend/src/components/Datasource/Field.tsx
EnxDev: sure

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FilterTitlePane.tsx
EnxDev: I thought you were referring to the * 3, Good catch!

mistercrunch (approved):
MERGE AWAY!

msyavuz (approved):

PR #33710, opened 2025-06-06
fix: Migrate charts with empty query_context

luizotavio32:
### SUMMARY
This PR implements a fix for the legacy charts migration. When a chart has a `null` `query_context` , a new `query_context` will be generated using the migrated `form_data` . Additionally, the query backup will be saved as `null` in the backup field `queries_bak` , ensuring that if the chart is downgraded, the `query_context` will revert to `null` .

This change was necessary because executing `warm_up_cache` for a migrated chart with a `null` `query_context` results in the error: `Chart's query context does not exist`. As a consequence, the chart fails to be cached.

### TESTING INSTRUCTIONS
Run superset db upgrade and superset db downgrade CLI commands.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/shared/migrate_viz/base.py
+++ b/superset/migrations/shared/migrate_viz/base.py
@@ -160,6 +160,7 @@ class MigrateViz:
             backup = {FORM_DATA_BAK_FIELD_NAME: form_data_bak}
 
             query_context = try_load_json(slc.query_context)
+            queries_bak = None
 
             if query_context:
                 if "form_data" in query_context:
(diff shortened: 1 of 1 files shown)


michael-s-molina (approved):

rusackas:
Merge? 
mailchimp-button

mistercrunch:
Was this a blocker for 5.0? Perhaps the only block RN!?

michael-s-molina:
> Was this a blocker for 5.0? Perhaps the only block RN!?

This was one of the blockers from our side. I'm not sure if it's the only blocker because people didn't vote on RC3 yet 🤷🏼‍♂️

PR #33715, opened 2025-06-06
chore: delete remaining Enzyme tests

mistercrunch:
Enzyme is long deprecated and we've long moved on to RTL. As discussed with some core committers, we believe it's a net positive to drop all remaining Enzyme tests even if that means lowering test coverage.

Now we benefit from less complexity in the repo, and can focus on test coverage in areas that matter most, without having to deal with the cognitive load of having multiple testing libs, and maintain upgradability to things like React 18 without having to drag Enzyme into that world.

@villebro @michael-s-molina @kgabryje @rusackas 

As we've been pretty forward-leaning in adopting new patterns, we need to be equality forward-leaning in deprecating old patterns. In many ways, wrapping up long-standing migration like deprecating less, bootstrap, enzyme, antd-v4 and other things is how we enable ourselves to adopt new and exciting, modern and vetted patterns and libraries.


villebro (approved):
Love it! There were surprisingly few Enzyme tests left, so I don't think this will have any significant impact on stability. 🚢 it!

PR #33717, opened 2025-06-07
fix(explore): add gap to the "Cached" button

Quatters:
### SUMMARY

This PR adds a cosmetic gap to the "Cached" button and replaces the Font Awesome icon with an SVG equivalent.

| Before | After |
| - | - |
|  |  |

### TESTING INSTRUCTIONS

1. Go to the explore page of any chart.
2. Update the chart a few times until the data is cached.
3. Check the changed button in the right upper corner.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


rusackas:
Much appreciated! I think we can just use AntD's `SyncOutlined` rather than adding a custom icon/svg, though.

Quatters:
> Much appreciated! I think we can just use AntD's `SyncOutlined` rather than adding a custom icon/svg, though.

I thought it looks too thin in this place. Or is it ok?

image

superset-frontend/src/components/CachedLabel/index.tsx
```diff
@@ -44,14 +45,14 @@ const CacheLabel: FC<CacheLabelProps> = ({
     >
       <Label
         className={`${className}`}
+        css={{ gap: '2px' }}
```
msyavuz:
```suggestion
        css={theme => css`
          gap: ${theme.gridUnit * 0.5}px;
        `}
```

msyavuz (commented):
Thanks for this! This part of the code is changed in [this](https://github.com/apache/superset/pull/31590) theming pr as well but i think sync icon looks better than what we had there. Also i have a small nit of convention for using emotion for styles. I think we can merge it after that is changed.

superset-frontend/src/components/CachedLabel/index.tsx
Quatters: Done!

msyavuz (approved):

rusackas:
Thank you for this! It's been bothering me forever :D

PR #33720, opened 2025-06-08
fix(serialization): allow unicode in json serialization

### LINKED ISSUE #19388: nested column cannot display correctly

### LINKED ISSUE #22904: Russian (Cyrillic) text is improperly displayed when using array_agg in the SQL query

Quatters:
### SUMMARY

PR changes the output of `json.dumps` to show unicode characters as they are, instead of using escape sequences.

#### Before:
Снимок экрана 08 06 2025@2x

#### After:
Снимок экрана 08 06 2025@2x

### ADDITIONAL INFORMATION

- [x] Has associated issue: 
Fixes #19388 
Fixes #22904
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
This seems like a win to me, though seeing `test-mysql` fail in CI (assuming it's related to this change) shows that this change might have some upgrade implications.

If there are indeed change management issues, it may require:
- adding a line in `UPDATING.md` about required checks/changes around people upgrading: something like "if you use mysql as your metadata database, you may need to configure MySQL with this specific flag {...}". Also making this clear in the docs around "cofiguring mysql"
- double checking that things that have been serialized with `ensure_ascii=True` can be deserialized with `ensure_ascii=False` to make sure we don't need a database migration to re-encode what's in the metadata db. (GPT thinks it is NOT a problem ...)

related GPT thread https://chatgpt.com/share/684b5999-8d50-8010-ab53-d62e10188b50

rusackas:
re-running that test, just in case.

mistercrunch:
Related: https://chatgpt.com/share/68507758-62e8-8010-8f63-a73f6c1eed55

Seems the connection string AND [maybe] the database settings need to be altered to support this for MySQL specifically (?)

If we can make CI pass, add a note in `UPDATING.md`, and document that it's a requirement for MySQL somewhere I'm fine with the change.

Other option would be to set `ensure_ascii=False` only for analytics database use cases, and keep it True for metadata database interaction, though that's probably confusing and maybe harder to maintain as we'd have different logic based on context...

rusackas:
Copilot's suggestions, FWIW:

1. Migration Failure Fix
Update the migration script to handle the foreign key constraint before dropping the column. Modify the script to drop the foreign key first and then the column:

```python
from alembic import op
import sqlalchemy as sa

def upgrade():
    # Drop the foreign key constraint
    op.drop_constraint('slices_ibfk_1', 'slices', type_='foreignkey')
    # Drop the column
    op.drop_column('slices', 'druid_datasource_id')

def downgrade():
    # Recreate the column (if needed during downgrade)
    op.add_column('slices', sa.Column('druid_datasource_id', sa.Integer(), nullable=True))
    # Recreate the foreign key constraint
    op.create_foreign_key('slices_ibfk_1', 'slices', 'druid_datasource', ['druid_datasource_id'], ['id'])
```

2. Test Failure Fix
The OperationalError indicates that the position_json column does not support certain Unicode characters. Ensure the column uses a character set that supports all Unicode characters (e.g., utf8mb4 for MySQL).

  Steps:

  * Update the database schema:
    ```sql
    ALTER TABLE dashboards MODIFY COLUMN position_json TEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
    ```
   * Ensure the database connection is configured to use utf8mb4: Modify the connection string in your configuration:
     ```python
    SQLALCHEMY_DATABASE_URI = "mysql+pymysql://<user>:<password>@<host>/<db_name>?charset=utf8mb4"
    ```

Quatters:
Done here

mistercrunch:
I don't think we need a db migration, except maybe for MySQL (unclear). In theory the json/simplejson implementation is compatible across the `ensure_ascii` setting. Now the issue seems to be that MySQL, depending on its setup and `CHARACTER_SET` default setting, chokes when getting unicode characters.

UPDATING.md
```diff
@@ -23,6 +23,9 @@ This file documents any backwards-incompatible changes in Superset and
 assists people when migrating to a new version.
 
 ## Next
+- [33720](https://github.com/apache/superset/pull/33720) If you're using MySQL as your metadata database for Superset, please
```
mistercrunch: @michael-s-molina, I believe your Superset instance @Airbnb uses MySQL behind the scene. How do you feel about this `UPDATING.md` notice? Should we test this forward before merging? Would love to get your blessing on this approach.

mistercrunch:
@Quatters , thank for doing all the work here. This seems viable. I added a comment to check in with folks using MySQL as their metadata db to get their input/blessing on this approach.

Note that this PR fixes user-visible issues around unicode handling with analytics databases.

UPDATING.md
michael-s-molina: > How do you feel about this UPDATING.md notice?

I'm not sure if this would be considered a breaking change...

>  Should we test this forward before merging?

Definitely. We need to check what the update process looks like.

michael-s-molina (changes requested):
Requesting changes just to make sure we test this before merging. I'll ask @justinpark's help as he worked on related tasks.

UPDATING.md
michael-s-molina: @villebro @sadpandajoe @rusackas Would this be a breaking change in your opinion?

mistercrunch: yes, from my understanding you'd have to change the character set in your MySQL instance, and I'm unclear if switching the global setting is sufficient, or whether that global setting gets materialized into columns in which case you may have to change it for every column (?)

Quatters:
Hi @mistercrunch, @michael-s-molina, any updates on this?

michael-s-molina:
> Hi @mistercrunch, @michael-s-molina, any updates on this?

We were OOO due to the holidays but I added this item to our team's backlog. Hoping to test this in the next couple weeks.

rusackas:
Resolved the conflict and currently re-running CI. Hopefully we can get this merged, pronto!

michael-s-molina:
@Quatters @mistercrunch @rusackas I executed the following migration to convert our existing MySQL database to be adherent to the changes in this PR.

Step 1: Backup Your Database First

```
mysqldump -u your_username -p your_database_name > backup_$(date +%Y%m%d_%H%M%S).sql
```

Step 2: Check Current State

```
-- Check current database character set
SHOW CREATE DATABASE your_database_name;

-- Check all tables and their character sets
SELECT TABLE_NAME, TABLE_COLLATION 
FROM INFORMATION_SCHEMA.TABLES 
WHERE TABLE_SCHEMA = 'your_database_name';

-- Check all columns with character sets
SELECT TABLE_NAME, COLUMN_NAME, CHARACTER_SET_NAME, COLLATION_NAME, DATA_TYPE
FROM INFORMATION_SCHEMA.COLUMNS 
WHERE TABLE_SCHEMA = 'your_database_name' 
AND CHARACTER_SET_NAME IS NOT NULL
ORDER BY TABLE_NAME, COLUMN_NAME;
```

Step 3: Change Database Default

```
-- Change the database default character set
ALTER DATABASE your_database_name 
DEFAULT CHARACTER SET utf8mb4 
COLLATE utf8mb4_0900_ai_ci;
```

Step 4: Generate Conversion Commands for All Tables

```
-- Generate ALTER TABLE commands for all tables
SELECT CONCAT(
    'ALTER TABLE `', TABLE_NAME, 
    '` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;'
) AS conversion_command
FROM INFORMATION_SCHEMA.TABLES 
WHERE TABLE_SCHEMA = 'your_database_name' 
AND TABLE_TYPE = 'BASE TABLE'
ORDER BY TABLE_NAME;
```

Step 5: Optional - Check Table Sizes Before Conversion

```
SELECT 
    TABLE_NAME,
    TABLE_ROWS,
    ROUND(((DATA_LENGTH + INDEX_LENGTH) / 1024 / 1024), 2) AS 'Size_MB',
    ROUND(AVG_ROW_LENGTH) AS 'Avg_Row_Length'
FROM INFORMATION_SCHEMA.TABLES 
WHERE TABLE_SCHEMA = 'your_database_name' 
AND TABLE_TYPE = 'BASE TABLE'
ORDER BY (DATA_LENGTH + INDEX_LENGTH);
```

Step 6: Prepare for Conversion

```
-- Disable foreign key checks temporarily (if you have foreign keys)
SET FOREIGN_KEY_CHECKS = 0;

-- Optional: Increase timeouts for large tables
SET SESSION innodb_lock_wait_timeout = 300;
SET SESSION lock_wait_timeout = 300;
```

Step 7: Convert All Tables
Replace your_database_name with your actual database name, then run the output from Step 4. Here's the template:

```
-- Example conversion commands (replace with your actual table names)
ALTER TABLE `table1` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
ALTER TABLE `table2` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
ALTER TABLE `table3` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
-- ... (continue for all tables)
```

Step 8: Re-enable Constraints

```
-- Re-enable foreign key checks
SET FOREIGN_KEY_CHECKS = 1;
```

Step 9: Verify the Conversion

```
-- Verify database default
SHOW CREATE DATABASE your_database_name;

-- Verify all tables are converted
SELECT TABLE_NAME, TABLE_COLLATION 
FROM INFORMATION_SCHEMA.TABLES 
WHERE TABLE_SCHEMA = 'your_database_name'
AND TABLE_COLLATION != 'utf8mb4_0900_ai_ci';

-- Should return no rows if all tables are converted

-- Verify all character columns are converted
SELECT TABLE_NAME, COLUMN_NAME, CHARACTER_SET_NAME, COLLATION_NAME
FROM INFORMATION_SCHEMA.COLUMNS 
WHERE TABLE_SCHEMA = 'your_database_name' 
AND CHARACTER_SET_NAME IS NOT NULL
AND CHARACTER_SET_NAME != 'utf8mb4';

-- Should return no rows if all columns are converted
```

Some observations:
- You can't simply change the database default character set. If you do that, the setting will only be applicable to new tables.
- This means that you need to convert all tables and their data to the new setting.
- The process does not take very long time (100k rows in 17 seconds) but does require downtime. @Quatters this needs to be explicitly called in `UPDATING.md`.
- We also need to mention the required change for the connection string in `UPDATING.md` as admins will not read the Configuring Superset docs.
- After the migration, characters like 大盘 are rendered correctly but emojis like 😀🌟 are rendered as ????????

sadpandajoe:
I guess the question here is, if i'm using a mysql database and I don't run the migrations, will this break superset for me? If it does, probably better to just create a migration file and have it detect if the db is mysql and if it is, then run it, else just pass.

Quatters:
I think automatic migration as @sadpandajoe suggests is better even if it's not strictly required, to be sure that all users starting from version X applied this migration. This seems more consistent and less problems in the future.

Therefore I plan to 
1. Create a migration file that detects whether MySQL is being used and applies the actions suggested by @michael-s-molina.
2. Write some tests to ensure the migration works correctly.
3. Add a note in UPDATING.md stating that applying the migration requires downtime.

rusackas:
Looks like this needs conflict resolution and perhaps some migration work. I'll set it to draft for now.

rusackas:
Anyone interested in brushing this up and getting it through? I'm closing out some old drafts, but this one looks like it had significant effort, so I'm a little hesitant in this case.

rusackas:
Rebased and cleaned up on the author's behalf. The branch had 11 commits (including several "Try to set collation" iterations and merge commits) — squashed down to a single clean commit with the net changes:

- `superset/utils/json.py`: `ensure_ascii=False` added to `simplejson.dumps`
- `tests/unit_tests/utils/json_tests.py`: unicode test added
- `.github/workflows/bashlib.sh`: MySQL charset/collation updated to `utf8mb4`/`utf8mb4_0900_ai_ci`

All pre-commit checks pass. Marking ready for review.

rusackas (commented):
Looks like `ensure_ascii=False` trips an existing guard. `test_position_json_emoji_is_ascii_safe_in_db` now fails (lands in position_json un-escaped), which is what's failing `test-sqlite/postgres/mysql`. That test was added on purpose for an emoji-truncation bug, so we've got a real conflict to settle, not just a flake. 

And @michael-s-molina's change request was waiting on the MySQL migration + UPDATING.md note we'd agreed to, which I don't see here yet. 

@Quatters can you take a look at the emoji test, and is the migration still on your plate?

rusackas:
@Quatters the `ensure_ascii=False` change seems like a good fix for the unicode display issue, but `test_position_json_emoji_is_ascii_safe_in_db` (tests/integration_tests/dashboards/test_update_emoji.py:92) asserts the opposite, that the stored string stays ASCII-safe, so it's red on test-mysql/postgres/sqlite. That test was added on purpose for the emoji-truncation bug, so we need to settle the conflict rather than just delete it.

And @michael-s-molina's change request was waiting on the MySQL migration + UPDATING.md note we'd agreed to, which I don't see here yet. Is that still on your plate, or do you want a hand with the rebase to get CI re-running?

rusackas:
Thanks for chasing this down @Quatters, the diagnosis here was spot on. Since it'd gone quiet, I took a crack at it on #41533 with a narrower approach: instead of flipping `ensure_ascii=False` globally (which tripped the `position_json` emoji guard from #39737 and would've needed a MySQL charset migration), it only changes the result-set `stringify` path so array/struct values render unicode, leaving the metadata escaping intact. Kept you on as co-author. I'll close this in favor of that one, but holler if you'd rather carry it forward yourself :)

PR #33722, opened 2025-06-08
chore: move backend under superset-backend

villebro:
### SUMMARY
WIP

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Nice to see this! Wondering if you should take one step further and move to `superset-backend/app/`, assuming will be a need for things like `superset-backend/extension-sdk/`, `superset-backend/data-access-layer/`, all depending on something like `superset-backend/core/` (?)

Oh also... now that there will be no `pyproject.toml` in the root (currently that makes it clear that it's a python project primarily and that you can `pip install apache-superset`), I'm wondering if we should have a `Code Orientation` section at the bottom of the readme giving pointers to important places across the monorepo.

PR #33723, opened 2025-06-09
fix(Timeseries): Temporal x axis defensive check for missing data

msyavuz:
### SUMMARY

There is an intermittent issue with temporal x-axis that causes charts to display NaN or epoch time instead of correct formats. There is a possible diversion from the happy path where coltype got from backend is corrupted that can cause this. This pr aims to fix that by having defensive checks in frontend to display correct formats. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -210,6 +210,7 @@ export default function transformProps(
   const colorScale = CategoricalColorNamespace.getScale(colorScheme as string);
   const rebasedData = rebaseForecastDatum(data, verboseMap);
   let xAxisLabel = getXAxisLabel(chartProps.rawFormData) as string;
+
   if (
     isPhysicalColumn(chartProps.rawFormData?.x_axis) &&
     isDefined(verboseMap[xAxisLabel])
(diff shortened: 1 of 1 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -257,7 +258,22 @@ export default function transformProps(
...
-  const xAxisDataType = dataTypes?.[xAxisLabel] ?? dataTypes?.[xAxisOrig];
+  let xAxisDataType = dataTypes?.[xAxisLabel] ?? dataTypes?.[xAxisOrig];
+
+  // Defensive check for xAxisDataType in case it is not provided by the backend.
+  if (xAxisDataType === undefined && data.length > 0) {
+    const firstValue = data[0][xAxisLabel];
+    if (
+      firstValue !== null &&
+      !Number.isNaN(new Date(firstValue as string | number).getTime())
+    ) {
```
amaannawab923, line 269: This check does require us to manually check if the first value is a data type or not from the x Axis.  
But the datasource has a list of all time columns under `granularitySqla`, so we can bring a boolean from `transformProps.ts` — basically `isXAxisTemporal`.

```ts
let timeColumnsSet;
let isXAxisTemporal = false;

if (Array.isArray(datasource.granularitySqla) && datasource.granularitySqla.length > 0) {
  timeColumnsSet = new Set(datasource.granularitySqla.flat());
}

if (timeColumnsSet && timeColumnsSet.has(formData?.xAxis)) {
  isXAxisTemporal = true;
}
```

The code will be as above, and we can just use the boolean like this:

```ts
if (xAxisDataType === undefined && isXAxisTemporal) {
  xAxisDataType = GenericDataType.Temporal;
}
```

I’ve tried this in the past — this approach works without fail, since `datasource?.granularitySqla` will always have relevant temporal columns.

Screenshot 2025-06-09 at 11 14 14 PM

Attatched screenshot shows datasource & its time columns under granularitysqla

msyavuz: This is an elegant solution! But since i can't reproduce the problem, i went with the simpler one that should always exist. I can do a refactor to something similar to yours after we confirm this works.

PR #33763, opened 2025-06-12
fix: select star

betodealmeida:
### SUMMARY

In databases that support catalogs `select_star` is building a query with the fully qualified table name:

```sql
SELECT * FROM my_catalog.my_schema.my_table
```

The problem is that in databases that support catalog but don't support cross-catalog queries (Hive, Postgres) the query fails, because it should simply be:

```sql
SELECT * FROM my_schema.my_table
```

This PR fixes it by making `select_star` take `supports_cross_catalog_queries` in consideration.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Vitor-Avila (approved):
LGTM!

PR #33764, opened 2025-06-13
fix: use risingwave as the sqlalchemy_uri_placeholder prefix for RisingWave engine

hzxa21:
### SUMMARY

Prior to this PR, when user choose RisingWave as the db engine, it will display `postgresql://...`  as the sqlalchemy uri placeholder. However, this may cause issues when browsing the dataset because there are unsupported functionalities. Based on the official [doc](https://superset.apache.org/docs/configuration/databases/#risingwave), `risingwave:://...` should be used so that [sqlalchemy-risingwave](https://pypi.org/project/sqlalchemy-risingwave/#history) will be chosen to be the driver.

This PR modifies the sqlalchemy_uri_placeholder for RisingWave engine and adds optional dependency for risingwave in `pyproject.toml`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas (approved):
LGTM, but cc @betodealmeida to make sure I'm not missing some detail(s). Looks like the docs are up to date, which is nice :D

PR #33769, opened 2025-06-13
feat(deck-gl): Enable individual deck.gl layer selection in FilterScope tree

richardfogaca:
### SUMMARY
Implemented layer-specific filtering for deck.gl multilayer charts, allowing users to apply dashboard filters to individual layers instead of all layers at once.

  What was added:
  - Individual layer selection in FilterScope tree - layers now appear as sub-nodes under multilayer charts
  - Layer-specific filter application - filters only affect selected layers, not the entire chart
  - Backend support for processing filters per layer based on selection
  - Enhanced UI with layer names and visual indicators

  Key changes:
  - FilterScope tree displays deck.gl layers as expandable sub-nodes with check icons
  - Filter system tracks which layers should receive which filters
  - Frontend chart component applies filters conditionally per layer
  - Backend processes each layer's data with only its applicable filters
  - Maintains backward compatibility with existing chart-level filtering

Screenshot 2025-06-11 at 17 20 47

### TESTING INSTRUCTIONS
1. Create test setup:
    - Add a deck.gl multilayer chart to a dashboard
    - Add a native filter
  2. Test layer selection:
    - Go to filter's "Scoping" tab
    - Expand the multilayer chart - you should see individual layers listed
    - Select specific layers (not the whole chart)
  3. Test filtering behavior:
    - Apply filter values
    - Verify only selected layers are filtered
    - Verify non-selected layers remain unchanged
  4. Test backward compatibility:
    - Select entire charts (existing behavior) - should work as before
    - Test existing dashboards - no regression expected

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X] Changes UI


superset-frontend/plugins/legacy-preset-chart-deckgl/src/Multi/Multi.tsx
```diff
@@ -111,29 +111,49 @@ const DeckMulti = (props: DeckMultiProps) => {
...
             ...(subslice.formData?.adhoc_filters || []),
-            ...(formData.extra_form_data?.adhoc_filters || []),
           ];
 
+          if (layerFilterScope) {
+            const filterDataMapping = formData.filter_data_mapping || {};
+
+            Object.entries(layerFilterScope).forEach(
+              ([filterId, filterScope]: [string, any]) => {
+                if (!filterScope || filterScope.includes(correctLayerIndex)) {
```
geido: `ensureIsArray` util in superset ui core could help drying up this a bit

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/ScopingTree.tsx
```diff
@@ -18,14 +18,22 @@
...
 import { Tree } from 'src/components';
 import { DASHBOARD_ROOT_ID } from 'src/dashboard/util/constants';
 import { Tooltip } from 'src/components/Tooltip';
 import { Icons } from 'src/components/Icons';
+import { Layout } from 'src/dashboard/types';
 import { useFilterScopeTree } from './state';
 import { findFilterScope, getTreeCheckedItems } from './utils';
 
+const StyledTree = styled(Tree)`
+  .antd5-tree-title {
```
geido: When rebasing with master all class names should now use the standard `ant` prefix. `antd5` prefix is gone as the migration has been finalized.

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/ScopingTree.tsx
```diff
@@ -40,8 +48,19 @@ const buildTreeLeafTitle = (
...
   hasTooltip: boolean,
   tooltipTitle?: string,
+  isDeckMultiChart?: boolean,
 ) => {
   let title = <span>{label}</span>;
+
+  if (isDeckMultiChart) {
+    title = (
+      <span style={{ display: 'inline-flex', alignItems: 'center' }}>
+        <Icons.CheckSquareOutlined style={{ marginRight: 4, fontSize: 16 }} />
```
geido: Icons have standard sizes that you can leverage in the props. Also, let's use the `css` prop with template literals over `style`

superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FilterScope/utils.ts
```diff
@@ -37,6 +37,52 @@ export const getNodeTitle = (node: LayoutItem) =>
   node?.id?.toString?.() ??
   '';
 
+export const generateChartSubNodes = (
+  chartId: number,
+  chart: any,
```
geido: I think we might be able to find existing type definitions in the codebase to re-use. Let's avoid `any` whenever possible

superset-frontend/src/dashboard/util/charts/getFormDataWithExtraFilters.ts
```diff
@@ -125,20 +141,49 @@ export default function getFormDataWithExtraFilters({
...
+    const aggregatedFormData = getExtraFormData(
+      dataMask,
+      filterIdsAppliedOnChart,
+    );
     extraData = {
-      extra_form_data: getExtraFormData(dataMask, filterIdsAppliedOnChart),
+      extra_form_data: aggregatedFormData,
     };
+
+    if (chart.form_data?.viz_type === 'deck_multi') {
```
geido: I saw we are using this check in a few places and I am wondering if this is the right way. I am worried that we are tying logics that are meant to be generic to specific plugins now. Is there a better way to apply the required logics so that do not feel so tied to the `deck_multi` viz type only?

geido:
Left a first-pass comment. Thanks for all the great work here!

superset-frontend/src/dashboard/util/charts/getFormDataWithExtraFilters.ts
richardfogaca: @geido Thanks for the comments. You're absolutely right that the hardcoded deck_multi checks create tight coupling, which isn't ideal architecturally. However, I think this is justified in the short term for these reasons:

1. deck_multi is genuinely unique - it's the only composite chart that combines multiple individual charts into one visualization, requiring layer-specific filtering that no other chart type needs.
2. The filtering requirements are complex - deck_multi needs special data structures like filter_data_mapping and layer_filter_scope to handle individual layer filtering, which don't apply to regular charts.
3. Limited scope - only one chart type currently needs this behavior, so building a complex generic solution might be over-engineering for now.

The checks are limited to specific filter processing functions and don't broadly impact the dashboard. While not perfect from a plugin architecture perspective, it's a pragmatic solution for deck_multi's unique composite nature.

For future composite charts, we should consider a plugin-driven filter strategy where charts can register their own filtering logic instead of hardcoding checks in core files.

What do you think?

rebecagbenevides:
Tested and I didn't find any issues related to this PR - QA passed

geido (approved):

PR #33781, opened 2025-06-16
feat(Dashboard): Row limit warning in dashboards

msyavuz:
### SUMMARY

Display row limit reached warning on dashboard view for charts. This functionality existed in explore view but same logic can be applied to dashboards as well.

image

### TESTING INSTRUCTIONS

Create a chart with row limits and add it to the dashboard.
Warning icon with tooltip should appear.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -527,6 +527,9 @@ const SliceHeaderControls = (
           id={`slice_${slice.slice_id}-controls`}
           aria-label="More Options"
           aria-haspopup="true"
+          css={theme => css`
+            padding: ${theme.gridUnit * 1.5}px;
```
geido: Nit: Can we avoid the 1.5 here? Maybe 2 or just gridUnit?

superset-frontend/src/dashboard/components/SliceHeader/index.tsx
```diff
@@ -176,6 +183,17 @@ const SliceHeader = forwardRef<HTMLDivElement, SliceHeaderProps>(
...
+    const chart = useSelector<RootState, Chart>(
+      state => state.charts[slice.slice_id],
+    );
+
+    const theme = useTheme();
+
+    const firstQueryResponse = chart.queriesResponse?.[0];
+
+    const rowLimit = Number(formData.row_limit || 0);
+    const sqlRowCount = Number(firstQueryResponse?.sql_rowcount || 0);
```
geido: Is falling back to 0 safe here? If no row limit is set and there are no results from the query, wouldn't this show the warning incorrectly?

superset-frontend/src/dashboard/components/SliceHeader/index.tsx
```diff
@@ -182,6 +183,17 @@ const SliceHeader = forwardRef<HTMLDivElement, SliceHeaderProps>(
       ({ dashboardInfo }) => dashboardInfo.crossFiltersEnabled,
     );
 
+    const chart = useSelector<RootState, Chart>(
+      state => state.charts[slice.slice_id],
```
kgabryje: We need the chart object only for the first query response right? If so, can we change this selector to `state.charts[id].queriesResponse?.[0]`?
Selecting the whole chart object will trigger an unnecessary rerender every time any field in the chart changes, and we only care about the queries response here

msyavuz: Good catch!

EnxDev (approved):
LGTM!

rebecagbenevides:
Passed in QA!

PR #33786, opened 2025-06-17
fix(theming): Fix visual regressions from theming P5

EnxDev:
### SUMMARY
This pull request applies small UI adjustments to correct visual issues introduced during the theming overhaul.
The focus is on reestablishing a consistent look and aligning components with the original design specifications.
There are no changes to application behavior or functionality.

@kasiazjc  I also replaced `font-weight 600` with `font-weight 500` everywhere

- Before
filters-before

- After
filters-after

- Before
before-1

- After
after-1

- Before
before-2

- After
after-2

- Before
before-3

- After
after-3

- Before
before-binocular

- After
after-binocular

- Before
label-modal-before

- After
label-modal-after

- increased size from 14px to 16px
Screenshot 2025-06-17 113736

### TESTING INSTRUCTIONS

- All tests should pass

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-chart-controls/src/components/ControlSubSectionHeader.tsx
+++ b/superset-frontend/packages/superset-ui-chart-controls/src/components/ControlSubSectionHeader.tsx
@@ -20,7 +20,7 @@ import { styled, css } from '@superset-ui/core';
 
 export const ControlSubSectionHeader = styled.div`
   ${({ theme }) => css`
-    font-weight: ${theme.fontWeightStrong};
+    font-weight: ${theme.fontWeightMedium};
     margin-bottom: ${theme.sizeUnit}px;
     font-size: ${theme.fontSizeSM}px;
   `}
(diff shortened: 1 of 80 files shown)


superset-frontend/packages/superset-ui-chart-controls/src/components/ControlSubSectionHeader.tsx
```diff
@@ -20,7 +20,7 @@ import { styled, css } from '@superset-ui/core';
 
 export const ControlSubSectionHeader = styled.div`
   ${({ theme }) => css`
-    font-weight: ${theme.fontWeightStrong};
+    font-weight: ${theme.fontWeightMedium};
```
mistercrunch: isn't `fontWeightMedium` the default? Can we simply remove the styling or are we cascading a stronger font-weight?

Less styling is better!

EnxDev: Yes, I believe it would be better to remove the custom styles. In the meantime, though, these fixes seem necessary in the short term. Could we plan to remove the custom styles in a separate PR?

EnxDev: I restored fontWeightStrong, and I set fontWeightStrong to 500 as the default

mistercrunch (approved):
LGTM!

PR #33790, opened 2025-06-17
refactor: rename docker-compose files and update references

polRk:
### SUMMARY

This pull request updates references to Docker Compose configuration files across multiple workflows, documentation, and scripts. The changes standardize the naming of these files from `docker-compose*.yml` to `compose*.yml` for consistency.

### ADDITIONAL INFORMATION
https://github.com/compose-spec/compose-spec/blob/main/spec.md


mistercrunch:
I'm not against this renaming, though it's unclear to me if it's a net positive. Tradeoff is things are a bit more intuitive to newcomers and creating a bit of confusion for current devs.

polRk:
> I'm not against this renaming, though it's unclear to me if it's a net positive. Tradeoff is things are a bit more intuitive to newcomers and creating a bit of confusion for current devs.

Thank you for your feedback!

The main reason for renaming docker-compose.yml to compose.yml is that there are now various solutions, beyond Docker, that support running applications based on the Compose specification. Having "docker" in the file name can mislead some developers, particularly those who don't have access to Docker, into thinking that Docker is a mandatory requirement for running applications, which is not necessarily the case. 

Renaming makes the configuration more universal and accessible to a broader range of developers, regardless of the tools they are using.

mistercrunch:
Oh that makes more sense. What are the alternatives? Any pointers to the latest/greatest packages? Articles to point to about their advantages over `docker compose` ?

polRk:
> Oh that makes more sense. What are the alternatives? Any pointers to the latest/greatest packages? Articles to point to about their advantages over `docker compose` ?

Docker Compose indeed works well with the Compose specification and detects compose.y(a)ml files by default. However, there are other tools that also support this specification, including:

- Docker Compose
- Podman Compose
- Kompose
- Nerdctl
- Okteto Stacks
- Docker Cloud Integrations

You can find the full list and additional information on the official Compose specification page: [Compose-spec README](https://github.com/compose-spec/compose-spec/blob/main/README.md#implementations).

In my changes, I'm not suggesting replacing `docker compose ...`. All documentation remains the same and everything continues to work as before.

mistercrunch (approved):
Approving but will let a few days before merging to let other committers time to chime in.

rusackas:
Nobody's chimed in. Merge?

jbat:
This creates some confusion on the 4.1.2 official release as it appears this doc change is live already.

As per the [Quickstart](https://superset.apache.org/docs/quickstart) 
```
git checkout tags/4.1.2
docker compose -f compose-image-tag.yml up
superset/compose-image-tag.yml: no such file or directory
```
4.1.2 release has the yml file still as `docker-compose-image-tag.yml`

If we are sticking with this compose-image-tag.yml filename change
we may need 2 cmds mentioned in the Quickstart for pre and post namechange

mistercrunch:
Oh right, the real longer term solution is to publish the versioned docs. I know @rusackas was working on that a little while back.

In the meantime, it'd be great if someone could add a note about this. Here it changes depending on what version you have checked out.

Could also revert...

polRk:
@jbat Thanks for catching this! You’re right — the filename change in the docs can be confusing for anyone following the Quickstart on the 4.1.2 release, since that version still uses docker-compose-image-tag.yml. The main branch now uses compose-image-tag.yml, so the docs and code are out of sync for older releases.

Until versioned documentation is available, I can add a note to the Quickstart to clarify which filename to use depending on the version. Would that work for you, or would you prefer to revert the doc changes for now? Happy to help either way!

Let me know which approach you prefer — I’m happy to submit a PR with the clarification!

sfirke:
It will also appear as `docker-compose-image-tag.yml` to anyone who sets their repo to the 5.0.0 tag: https://github.com/apache/superset/blob/5.0/docs/docs/quickstart.mdx.

We could revert the renaming PR and hold it for the next major release, that way it would hit the repo right before 6.0.0 come out and be included in 6.0.0? That would minimize the amount of time there's a difference between how you run master vs. latest. Though even then we'd want a note about the file name being different for versions <=5.x

mistercrunch:
So I'd say two options:
- revert until we have versioned docs in place
- clarify the docs with "prior to version N do this, post version N do this"

sfirke:
I guess let's revert until we get versioned docs? New issues keep coming in about this.

hainenber:
For now, my vote is on the revert. More folks will feel baffled reading the Quickstart guide.

hainenber:
yikes, GH doesn't allow creating revert due to newly changes. Had to manually create the PR [here](https://github.com/apache/superset/pull/33883), committer folks.

PR #33821, opened 2025-06-18
fix: Set time filter's isExtra to false when saving as new chart

Vitor-Avila:
### SUMMARY
When accessing a chart from a dashboard, filters are carried over. If a temporal filter was carried over, saving the chart as a new item would correctly reset the filter to `No filter`, however it would preserve `isExtra=true`, so it can't be overwritten.

This PR fixes this flow.

**Before**

https://github.com/user-attachments/assets/3883a69e-6118-4d3b-b1a4-1adf593fd9ca

**After**

https://github.com/user-attachments/assets/738c3b0a-08e7-45b3-ba76-2ba26ebbec10

### TESTING INSTRUCTIONS
Frontend test added. For manual testing:

1. Create a chart.
2. Save it to a dashboard that has a time range filter.
3. Apply a time range filter value to the dashboard, and access the chart.
4. Use the Save as option to duplicate the chart.
5. Validate the new chart does not have any extra filters.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/actions/saveModalActions.test.ts
+++ b/superset-frontend/src/explore/actions/saveModalActions.test.ts
@@ -689,4 +689,40 @@ describe('getSlicePayload', () => {
 
     expect(hasTemporalRange).toBe(true);
   });
+
+  test('should reset isExtra flag to false for temporal filter when saving as a new chart', () => {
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

PR #33830, opened 2025-06-19
fix: Consider last data point for Big Number comparison lag

Vitor-Avila:
### SUMMARY
https://github.com/apache/superset/pull/32767 introduced the ability to select an aggregation for the big number displayed in a Big Number with Trendline chart. 

The comparison lag feature was relying on the big number, so when the chart is set to aggregate the calculation was comparing the aggregated value with the specified period lag. 

This PR fixes the logic so that the comparison always use the last data point.

**Before**

image

**After**

image

### TESTING INSTRUCTIONS
Frontend test added. For manual testing.
1. Create a big number with trendline with an aggregation.
2. Implement a comparison lag and validate the number.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


geido (approved):

rusackas:
Think it's time to remove the EXPERIMENTAL CHARTS categorization for that?

PR #33834, opened 2025-06-19
fix(native filters): Make the Apply button available after click on Clear All

Vitor-Avila:
### SUMMARY
Clicking on **Clear All** removes all filter selections, but the **Apply** button is still disabled and can't be applied.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
@@ -40,7 +40,7 @@ import {
   styled,
 } from '@superset-ui/core';
 import { useHistory } from 'react-router-dom';
-import { updateDataMask, clearDataMask } from 'src/dataMask/actions';
+import { updateDataMask } from 'src/dataMask/actions';
 import { useImmer } from 'use-immer';
 import { isEmpty, isEqual, debounce } from 'lodash';
 import { getInitialDataMask } from 'src/dataMask/reducer';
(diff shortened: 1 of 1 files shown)


superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
```diff
@@ -256,25 +256,17 @@ const FilterBar: FC<FiltersBarProps> = ({
...
           if (draft[id].filterState?.value !== undefined) {
             draft[id].filterState!.value = undefined;
           }
+          draft[id].extraFormData = {};
         });
       }
     });
-    if (dispatchAllowed) {
-      clearDataMaskIds.forEach(id => dispatch(clearDataMask(id)));
-    }
```
Vitor-Avila, line 277: I believe the idea here was to actually have `let dispatchAllowed = true;`, then set it to `false` if any filter is required, and finally `dispatch(clearDataMask(id))` for all filters -- this would automatically trigger the clear all to charts (without the need to click on the **Apply** button).

I talked with @yousoph about this and it might be a confusing experience, as no other filter interaction is "automatic" -- every filter change requires applying them, so I just removed this logic (it wasn't really working as `dispatchAllowed` was never set to `true`). 

If we feel like that's the flow we want to support, we can just re-add this syntax and set `dispatchAllowed` to initially `true`.

Vitor-Avila:
The original issue described in the PR is solved and can be validated in this dashboard: http://54.245.182.133:8080/superset/dashboard/9

1. Select some filter values.
2. Apply them.
3. Click on Clear All, and confirm the Apply is available.
4. Clicking on Apply updates all charts (without any filter aplied).

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
Vitor-Avila: From more recent conversations, it might make more sense to actually introduce a new button (Like **Reset**) that would restore all default values and then automatically apply.

betodealmeida (approved):

superset-frontend/src/dashboard/components/nativeFilters/FilterBar/index.tsx
```diff
@@ -256,25 +256,17 @@ const FilterBar: FC<FiltersBarProps> = ({
...
       if (dataMaskSelected[id]) {
-        if (filter.controlValues?.enableEmptyFilter) {
-          dispatchAllowed = false;
-        }
-        clearDataMaskIds.push(id);
         setDataMaskSelected(draft => {
           if (draft[id].filterState?.value !== undefined) {
             draft[id].filterState!.value = undefined;
           }
+          draft[id].extraFormData = {};
```
Vitor-Avila, line 266: Clearing up `extraFormData` here would then allow `isApplyDisabled` to return` false`.

Vitor-Avila:
I'll work on a follow up PR to fix some remaining issues around the Clear All button

PR #33837, opened 2025-06-20
feat(helm): Allow extra PostgreSQL parameters when using externally-managed PostgreSQL (postgresql.enabled:false)

nathanjrobertson:
### SUMMARY
When using a self hosted PostgreSQL (eg. cloud vendor hosted), they often require extra parameters for the SQLAlchemy URI string. In my case, my vendor requires "sslmode=require". However, I respect that other users might have other parameters they need to pass, so I haven't hardcoded an "SSL" flag, but instead provided a generic connection parameters option.

The new `supersetNode.connections.db_conn_params` option is commented out by default (ie. defaults to `nil`), which makes this change backward compatible. The commented out value provided in `values.yaml` shows how to enable SSL (which I expect is the most common use case for this feature).

### TESTING INSTRUCTIONS

1. Setup PostgreSQL server with SSL support enabled
2. In `values.yaml`, set `supersetNode.connections.db_conn_params: "sslmode=require"`
3. Install Helm chart.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sfirke (approved):
Thanks for the contribution! I am not a code owner on the Helm chart so can't merge it but it looks uncontroversial to me. The hook requires the helm chart version to be bumped with every change, so can you please do that to make it pass CI?

nathanjrobertson:
> Thanks for the contribution! I am not a code owner on the Helm chart so can't merge it but it looks uncontroversial to me. The hook requires the helm chart version to be bumped with every change, so can you please do that to make it pass CI?

Yep. I'll take a look at it and make the changes later in the week. Thanks for the feedback.

mistercrunch:
OMG a CH supporter. AFAIC this gets you an elevated level of support and attention on this repo!   Screenshot 2025-06-23 at 1 38 48 PM

hainenber:
IMO, the Helm attribute can be named something like `supersetNode.connections.postgresql.db_conn_params` as this is a bit tool-specific and I think we can reserve the generic term for other future uses. My 2cent.

nathanjrobertson:
> IMO, the Helm attribute can be named something like `supersetNode.connections.postgresql.db_conn_params` as this is a bit tool-specific and I think we can reserve the generic term for other future uses. My 2cent.

Connection parameters actually aren't proprietary to PostgreSQL. [MySQL and MariaDB use them as well](https://docs.sqlalchemy.org/en/20/dialects/mysql.html#supported-versions-and-features), as does [the Oracle driver](https://docs.sqlalchemy.org/en/20/dialects/oracle.html#dialect-oracle-oracledb-connect), [Microsoft SQL Server](https://docs.sqlalchemy.org/en/20/dialects/mssql.html#temporary-table-resource-reset-for-connection-pooling), even [SQLite](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#uri-connections) has them. Example for MySQL:

```
engine = create_engine(
    "mysql+pymysql://user:pass@some_mariadb/dbname?charset=utf8mb4"
)
```

The only thing is that makes this PR PostgreSQL specific is that the pre-existing template variable is already hardcoded to be PostgreSQL specific:

https://github.com/apache/superset/blob/f802474b07bc7a2765e57fcd6612d2b4454f5c62/helm/superset/templates/_helpers.tpl#L99

The URL scheme in the chart template is hardcoded to `postgresql+psycopg2://`, meaning the chart currently is hardcoded to be PostgreSQL. Removing the PostgreSQL hardcoding (adding support for, and testing, other databases) is something for a different PR. The purpose of this PR is to add parameter support, and when the hardcoding is removed, this new variable would then apply to all databases supported by SQLAlchemy.

nathanjrobertson:
> Thanks for the contribution! I am not a code owner on the Helm chart so can't merge it but it looks uncontroversial to me. The hook requires the helm chart version to be bumped with every change, so can you please do that to make it pass CI?

I've just pushed an extra commit to this PR which bumps the chart version. Thanks.

nathanjrobertson:
I've fixed the pre-commit CI error. Could somebody please re-trigger the CI tests to run? Thanks.

After that is confirmed, it should be ready for review and potential merging.

rusackas (approved):
LGTM, but I'll defer to @mistercrunch or @villebro to merge this one.

mistercrunch:
I'm checking out of reviewing anything related to the helm chart personally, we should clarify if the goal is to have a running base implementation, or try-to-make-it-work-for-everyone-type deal. Personally would vouch for putting this in another repo and have people who care about the helm chart collab there.

nathanjrobertson:
> I'm checking out of reviewing anything related to the helm chart personally, we should clarify if the goal is to have a running base implementation, or try-to-make-it-work-for-everyone-type deal. Personally would vouch for putting this in another repo and have people who care about the helm chart collab there.

I'd advocate against moving to a separate repository. Superset is pretty involved to get running, and the Helm chart makes it reasonably easy to get it working. Don't underestimate how much this charts lowers the barrier to entry for new users.

Having an officially supported chart [with excellent documentation](https://superset.apache.org/docs/installation/kubernetes/) is a big asset to the project. Personally, it was a major driver to me evaluating Superset in the first place - I didn't need to spend weeks figuring out how to deploy and maintain it. Somebody who has a deep understanding of the product has done 80% of the work, producing a far better default configuration that I would be able to do in the time I could allocate to it.

And docker-compose isn't the same. That's for dev environments.

mistercrunch:
> Having an officially supported chart is a big asset to the project

Totally agree. Ideally people using the chart actively can help maintain it. One thing I've found in devops is that trying to write a complex, multi-services abstraction (pick you own database, pick your own caching layer, pick your own message queue, ...), with hundreds of config keys is extremely challenging. Like the abstraction itself (say the helm chart of k8s operator) ends up being much more code than any given implementation would need.

The main dillema around the helm chart here as a contributor put it, it's to decide whether it's just "a reference implementation that works well and is bulletproof" VS "an abstraction that allows you to configure your system in all sorts of ways". I found the former would be much easier for us to maintain and provide guarantees around.

About the config system, I'm working on something that will make Superset much more configurable, where every config key that's static (serializable to json, not a python object) will be available through env vars (prefixed by `SUPERSET__`) or will support using the metadata database for storing them (defined in a `DATABASE_MANAGED_CONFIG_KEYS = []` or similar), which will be fully dynamic/runtime. Should help quite a bit.

rusackas:
@mistercrunch merge if you're cozy with this implementation, I'll leave the call to you :D

nathanjrobertson:
> decide whether it's just "a reference implementation that works well and is bulletproof" VS "an abstraction that allows you to configure your system in all sorts of ways"

I understand what you're saying. I think this PR is trivial enough that it is compatible with both. The existing functionality assumes PostgreSQL, but explicitly allows you to turn off the built-in Postgres instance and use your own external one. Without SSL support (added by this PR) the existing functionality doesn't work at all on kubernetes vendors who require SSL to be switched on to be able to connect to their managed database solution (eg DigitalOcean, I'd imagine other vendors have the same sane default). Hence, the current implementation doesn't work well in a best practice industry standard environment.

This PR brings it one step closer to supporting other databases (removing the hard coding of PostgreSQL in the database URI is the next step). So, if that's the goal, this PR gets us one step closer.

Summary - this PR fixes a current issue with supported functionality if the goal is a reference bulletproof implementation, and a step in the right direction if the goal is something broader.

villebro:
I'm not clear on why we'd want this change. IMO it adds unnecessary complexity to do something that can be done in local CI/CD pipelines. For instance, you could lint your helm repo to error out if a SQLA connection string doesn't have the required URL params. I agree that this change is not problematic in itself, especially if this is universally useful, but over time these types of customizations tend to add up, and add to maintenance burden unnecessarily.

nathanjrobertson:
> I'm not clear on why we'd want this change. IMO it adds unnecessary complexity to do something that can be done in local CI/CD pipelines. For instance, you could lint your helm repo to error out if a SQLA connection string doesn't have the required URL params. I agree that this change is not problematic in itself, especially if this is universally useful, but over time these types of customizations tend to add up, and add to maintenance burden unnecessarily.

The reason this change is needed is because all managed Kubernetes vendors highly recommend TLS/SSL be switched on, and a number of them enforce it as being required to be on. A snapshot of some large managed kubernetes vendors:

* [Managed Azure Kubernetes connecting to their managed PostgreSQL won't work](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/how-to-connect-tls-ssl) (as TLS/SSL is required and enforced)

* TLS/SSL is [required and enforced between DigitalOcean Kubernetes and their managed PostgreSQL](https://docs.digitalocean.com/products/databases/postgresql/how-to/secure/#increase-the-ssl-mode-verification-level)

* On Google Kubernetes, it is not enforced but highly recommended - ["We recommend that you enforce all connections to use SSL/TLS."](https://cloud.google.com/sql/docs/postgres/configure-ssl-instance)

* On AWS Kubernetes, ["by default, RDS for PostgreSQL uses and expects all clients to connect using SSL/TLS, but you can also require it"](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Concepts.General.SSL.html). Basically, highly recommended, not enforced by default, with a customer configurable switch to enable requiring it. 

Basically **all four** highly recommend this, and two require and enforce it. The current chart can't do that without this PR.

Helm charts are designed to capture the most common usage patterns and make complicated Kubernetes deployments simpler. Managed databases are common and best practice. You shouldn't use a managed database without TLS/SSL turned on, and a number of vendors have made the sensible decision and required TLS/SSL such that you can't use it without TLS/SSL being turned on.

If the functionality to use an external PostgreSQL instance wasn't a feature of the helm chart, then yes, it would add complexity to the chart. But the functionality for connecting to an external (commonly managed) database is already explicitly there, and it should be - it's a very common pattern that makes life simpler (which is what Helm is all about).

Without this PR you can't meet best practice on **any** Kubernetes / managed database pair, and in at least two cases it just won't work at all. The current chart provides functionality that can only be used in an insecure way, and prevents you from running in a secure fashion. And this issue can be fixed with the trivial change this PR provides (it's 12 lines of code).

rusackas:
As a non-helm-concerned person, I'm not sure what to do with this... but it seems pretty stuck. Maybe this is something we can hash out in the next Town Hall sync?

If so, feel free to add it to the agenda and join the zoom call!
https://docs.google.com/document/d/1WNtQhrJLCN59tc6FDI2iq4W-aoS6CISCo6S7hHgGMpM/edit?tab=t.0#heading=h.54tpwybrogat

rusackas:
Moving this to draft mode, while it awaits a rebase and a decision on if/how we want to proceed with this. If there's no further discussion it may be closed as inactive eventually.

nathanjrobertson:
I'll close this. We're not using this anymore so it doesn't really bother me that much. It'd be a sensible change to include (given managed databases in public clouds largely won't work without it), but I'll sleep ok with it not being merged.

PR #33851, opened 2025-06-20
feat(theming): improving theme docs and configuration

mistercrunch:
Fixing/improving how the theme is set, feature flag for ThemeEditor and related docs.

- renamed `THEME_OVERRIDES` (legacy) to `THEME`
- deleted unused feature flag leftover from the development process
- re-hooked `superset_config.THEME` to make it's way into the app on app-load

Screenshot 2025-06-20 at 5 02 29 PM
Screenshot 2025-06-20 at 5 01 55 PM
Screenshot 2025-06-20 at 5 00 51 PM

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
+++ b/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
@@ -52,7 +52,6 @@ export enum FeatureFlag {
   SqlValidatorsByEngine = 'SQL_VALIDATORS_BY_ENGINE',
   SshTunneling = 'SSH_TUNNELING',
   TaggingSystem = 'TAGGING_SYSTEM',
-  ThemeEnableDarkThemeSwitch = 'THEME_ENABLE_DARK_THEME_SWITCH',
   ThemeAllowThemeEditorBeta = 'THEME_ALLOW_THEME_EDITOR_BETA',
   Thumbnails = 'THUMBNAILS',
   UseAnalogousColors = 'USE_ANALOGOUS_COLORS',
--- a/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
+++ b/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
@@ -52,6 +52,7 @@ export enum FeatureFlag {
   SqlValidatorsByEngine = 'SQL_VALIDATORS_BY_ENGINE',
   SshTunneling = 'SSH_TUNNELING',
   TaggingSystem = 'TAGGING_SYSTEM',
+  ThemeEnableDarkThemeSwitch = 'THEME_ENABLE_DARK_THEME_SWITCH',
   ThemeAllowThemeEditorBeta = 'THEME_ALLOW_THEME_EDITOR_BETA',
   Thumbnails = 'THUMBNAILS',
   UseAnalogousColors = 'USE_ANALOGOUS_COLORS',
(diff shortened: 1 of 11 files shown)


superset-frontend/src/features/home/RightMenu.tsx
```diff
@@ -500,15 +494,6 @@ const RightMenu = ({
             <ThemeEditor theme={themeEditorTheme} setTheme={setTheme} />
           </span>
         )}
-        {isFeatureEnabled(FeatureFlag.ThemeEnableDarkThemeSwitch) && (
-          <span>
-            <ThemeSelect
```
geido: @mistercrunch isn't this the component that controls the option for the user for default / dark / system settings mode? I don't think this is a leftover. This was meant to be here.

superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
```diff
@@ -52,7 +52,6 @@ export enum FeatureFlag {
   SqlValidatorsByEngine = 'SQL_VALIDATORS_BY_ENGINE',
   SshTunneling = 'SSH_TUNNELING',
   TaggingSystem = 'TAGGING_SYSTEM',
-  ThemeEnableDarkThemeSwitch = 'THEME_ENABLE_DARK_THEME_SWITCH',
```
geido: Just confirmed. This is meant to be here. Not a leftover. Please, check https://github.com/apache/superset/pull/33420/files

geido (changes requested):

superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
mistercrunch: Gotcha. Felt redundant as switching to dark can be achieved in a few extra clicks. I don't mind keeping it.

superset/config.py
```diff
@@ -551,19 +551,26 @@ class D3TimeFormat(TypedDict, total=False):
...
     # Set to True to to enable factory resent CLI command
     "ENABLE_FACTORY_RESET_COMMAND": False,
     # Whether Superset should use Slack avatars for users.
     # If on, you'll want to add "https://avatars.slack-edge.com" to the list of allowed
     # domains in your TALISMAN_CONFIG
     "SLACK_ENABLE_AVATARS": False,
-    # Adds a switch to the navbar to enable/disable the dark theme
-    # This is used for development to expose what is dynamic (css-in-js) vs
-    # what is managed by `.less` files.
-    "DARK_THEME_SWITCH": False,
```
mistercrunch, line 565: I think that was a leftover/miswired which got me confused as it wasn't referenced elsewhere, and the two I'm adding now were missing references on the backend...

superset/config.py
```diff
@@ -662,11 +669,17 @@ class D3TimeFormat(TypedDict, total=False):
...
 # You can use the AntDesign theme editor to generate a theme structure
 # https://ant.design/theme-editor
-THEME_OVERRIDES: dict[str, Any] = {}
+# To expose a JSON theme editor modal that can be triggered from the navbar
+# set the `ENABLE_THEME_EDITOR` feature flag to True.
+#
+# To set up the dark theme:
+# THEME = {"algorithm": "dark"}
+
+THEME: dict[str, Any] = {}
```
mistercrunch, line 682: simple renaming here as overrides doesn't make sense anymore

superset/config.py
geido: I believe this is meant to be the preferred way of switching theme modes and works independently of the editor. We might want to keep the editor FF false by default and this default True?

mistercrunch: @geido should be good now. I got confused because the feature flags object wasn't aligned. We could also consolidate the 2 flags into a single flag and a single menu.

geido (approved):

PR #33852, opened 2025-06-21
fix: Update file naming to skip ID in chart and dashboard exports

### LINKED ISSUE #33784: [SIP]  Proposal for naming the assets in export
### Motivation

Currently when I export a dashboard and related assets like charts, the dashboard and chart files names seem to have a random number as the suffix of the filename. This number can change even if the actual chart or dashboard is the same. This constant changing of the asset file names creates a problem when storing the assets in a version control system like Git.

### Proposed Change

Name the exported assets in the following manner: <asset-name>_<asset-uuid>.yaml

abhinav-1305:
### SUMMARY

Added `skip_id=True` to `get_filename()` calls in dashboard/chart exports to remove ID suffixes from filenames.

### ADDITIONAL INFORMATION

- [x] Has associated issue: resolves #33784 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/chart/export.py
+++ b/superset/commands/chart/export.py
@@ -47,7 +47,7 @@ class ExportChartsCommand(ExportModelsCommand):
 
     @staticmethod
     def _file_name(model: Slice) -> str:
-        file_name = get_filename(model.slice_name, model.id)
+        file_name = get_filename(model.slice_name, model.id, skip_id=True)
         return f"charts/{file_name}.yaml"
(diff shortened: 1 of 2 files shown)


betodealmeida:
I mentioned in https://github.com/apache/superset/issues/33784, but the problem with this approach is that chart and dashboard names are not unique. So if you have an export with two charts with the same name there would be only one file.

PR #33853, opened 2025-06-21
build(deps): remove legacy browser polyfills

hainenber:
build(deps): remove legacy browser polyfills

### SUMMARY

Remove polyfills for `fetch` and `abortcontroller` that are no longer relevant thanks to built-in support modern browsers.

image

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/jest.config.js
+++ b/superset-frontend/jest.config.js
@@ -32,7 +32,7 @@ module.exports = {
       '<rootDir>/node_modules/@superset-ui/$1/src/$2',
     '^@superset-ui/([^/]+)$': '<rootDir>/node_modules/@superset-ui/$1/src',
   },
-  testEnvironment: 'jsdom',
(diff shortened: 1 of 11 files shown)


mistercrunch (approved):
LGTM, but leaving open in case there's a reviewer closer to the matter. There's probably more cleanup possible up and around this. 

Related to cleanup, I think we might have some unused deps and/or devDeps laying around. Would be really neat if we could take one an analysis of each package thinking "do we really still need this?" for each lib.

hainenber:
Yes, there are plenty of deps plus dev-deps we can revisit to assess their necessity. Spoiler alert: more upcoming PRs from me on this aspect ;)

mistercrunch:
> Spoiler alert: 

Quite the opposite of a spoiler :) you'll be unspoiling here :)

PR #33864, opened 2025-06-23
fix: #33862 – Security menu missing on MySQL due to case mismatch

### LINKED ISSUE #33862: Bug: “Security” menu disappears in Superset 4 on MySQL
### Bug description

# Bug: “Security” menu disappears in Superset 4 on MySQL

|  | **Value** |
| --- | --- |
| **Superset version** | 4.0.x (verified on 4.0.0 & 4.1.0-SNAPSHOT) |
| **Metadata DB** | MySQL
| **Regression?** | Yes – Superset 3 shows the menu correctly |
| **Impact** | *Admin* users lose access to **Security → Users / Roles / RLS / Action Log*** |

---

## Steps to Reproduce

1. **Build a local Superset 4 Docker image**  

2. **Start a brand-new MySQL container**  

3. **Run Superset 4 pointing at that MySQL instance**  

4. **Log into the Superset UI as the admin user**  
   *open the Settings navigation menu and look for the **Security** category.*

**Expected result:** *Security* section is visible for Admin.  
**Actual result:** *Security* section is missing.

(Repeat the same steps with PostgreSQL as the metadata DB and notice the menu appears, confirming the regression.)

---

## Findings (database state)

| Database | Entries in `ab_view_menu` that match `%ecurity%` **after first init** |
| --- | --- |
| **MySQL** | Only one row: `"security"` (all lowercase) |
| **PostgreSQL** | Two separate rows: `"Security"` **and** `"security"` |

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

I don't know

### Node version

I don't know

### Browser

Chrome

### Additional context

_No response_

### Checklist

caerux:
### Summary
Align view permission names with their menu label so Superset creates the correct “Security” row on MySQL.

### Fix
* superset/views/roles.py – `class_permission_name = "Security"`
* superset/views/users_list.py – `class_permission_name = "Security"`

### Impact
Restores the Security menu on fresh MySQL installs without affecting
PostgreSQL or existing deployments.

Resolves #33862

### DIFF
--- a/superset/views/roles.py
+++ b/superset/views/roles.py
@@ -25,7 +25,7 @@ from .base import BaseSupersetView
 
 class RolesListView(BaseSupersetView):
     route_base = "/"
-    class_permission_name = "security"
+    class_permission_name = "Security"
 
     @expose("/roles/")
     @has_access
--- a/superset/views/users_list.py
+++ b/superset/views/users_list.py
@@ -25,7 +25,7 @@ from .base import BaseSupersetView
 
 class UsersListView(BaseSupersetView):
     route_base = "/"
-    class_permission_name = "security"
+    class_permission_name = "Security"
 
     @expose("/users/")
     @has_access


mistercrunch:
Not sure, but I'm wondering if this can break things for people that have attached roles to it (?)

Quick grep shows more low-cap 'security':
```
16:52 $  git grep 'class_permission_name.*".ecurity"'
superset/views/groups.py:    class_permission_name = "security"
superset/views/logs.py:    class_permission_name = "security"
superset/views/roles.py:    class_permission_name = "security"
superset/views/users_list.py:    class_permission_name = "security"
```

Putting an `hold!` label on the PR until we can resolve this. From my understanding we'd need to align across the board and probably have a db migration script to reassign current/existing perms in the backend.

sadpandajoe:
I think permissions are stored in a database, so if we update the name without a migration, it could break for others. I think @michael-s-molina ran into something like this but could be wrong.

michael-s-molina:
> I think permissions are stored in a database, so if we update the name without a migration, it could break for others. I think @michael-s-molina ran into something like this but could be wrong.

Yes, these types of changes without proper migrations will break existing deployments. We ran into one of these when upgrading to 4.1.

mistercrunch:
Do we have reusable constructs / methods to easily generate permission renames as part of db migrations?

rusackas:
Hi @caerux 👋

This PR has been on hold for about 6 months. Per the discussion, this change needs a DB migration to avoid breaking existing deployments that have roles attached to the current permission names.

@mistercrunch asked if there are reusable constructs/methods to easily generate permission renames as part of DB migrations - that might be a good starting point.

Are you still interested in working on this? If so, adding a migration would unblock the PR. If you need help or guidance on the migration approach, feel free to ask!

Thanks for your contribution!

rusackas:
I accidentally picked up the same piece of work over on https://github.com/apache/superset/pull/40527

Happy to close this in favor of that, or vice versa. Review(s) appreciated if you want to help merge the new one.

rusackas:
Thanks for digging into this @caerux, the diagnosis is right. As I mentioned I picked up the work over in #40527 (still open), so I'll close this in favor of that one, but ping me if you think there's value here I'm missing and we'll reopen this. Reviews on #40527 very welcome if you'd like to help land it. Thank you once again!

PR #33876, opened 2025-06-24
update doc(quickstart.mdx): changed the file name

sripathroy:
File Name is modified

### SUMMARY
The yml file name is wrong in the existing docs. So modified accordingly


sfirke:
Thanks for making this PR! It was just fixed in this https://github.com/apache/superset/pull/33883 which reverted a commit that had introduced the discrepancy. I think we are all set now but please do make another PR if you see future things like this that need updating 🙏

Issue #33888, opened 2025-06-24
Superset 5.0.0 Helm Chart Not Available


brandon-kaplan:
### Bug description

Superset 5.0.0 Helm Chart Not Available

### Superset version

master / latest-dev

### Python version

3.10

### Node version

16

### Browser

Chrome

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [x] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
This is largely a volunteer effort, and tends to lag. CC @villebro here as he tends to manage this.

gpchandran:
Hi @villebro - any good news with helm chart support Superset 5.x ? I can contribute if you can help with the steps/changes required

brandon-kaplan:
> Hi @villebro - any good news with helm chart support Superset 5.x ? I can contribute if you can help with the steps/changes required

@gpchandran Just waiting on https://github.com/apache/superset/pull/33889

PR #33889, opened 2025-06-24
chore(helm): bump app version to 5.0.0

brandon-kaplan:
### SUMMARY

chore(helm): update app version and bump chart version for 5.0.0 release

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes https://github.com/apache/superset/issues/33888
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
@villebro as I think helm chart upgrade requires special handling

sfirke:
Just curious: when would we bump the minor version of the chart e.g., to 0.15.0? I get that the chart itself isn't changing much, but the end result is a big change with the major version upgrade to Superset.

nathanjrobertson:
> Just curious: when would we bump the minor version of the chart e.g., to 0.15.0? I get that the chart itself isn't changing much, but the end result is a big change with the major version upgrade to Superset.

I would have thought you'd want a major version update in the chart for a major version update in the app. Despite the YAML in the chart itself not changing much, the end result of what that YAML does is hugely different.

Those that are running this chart in a Continuous Deployment scenario (eg. HelmRelease on FluxCD) tend to have rules around automated chart version updates (eg. `superset<1.0.0`). There is an implicit assumption that large, breaking changes will cause a major version update, and minor versions (due to no breaking changes) should be ok to automatically deploy using a CD tool. A new major version of the app is a large breaking change (notwithstanding the chart itself being unchanged).

sfirke:
@nathanjrobertson that makes sense to me. I wonder if we should start now bumping the major version of the chart upon each major Superset version bump. This chart could then be 5.0.0 (or 1.0.0 but IMO going straight to 5 is clearer).

Again I know nothing about Helm charts or their versioning, I'm just brainstorming and will defer to others.

chgl:
Whether the `appVersion` and the chart `version` should be the same is usually up to taste. Personally, I think it's better to keep them separate as it allows releasing changes to just the chart template independently of the slower-paced superset releases.

I'd recommend bumping the chart to v0.15.0 as it indicates a potentially breaking change due to the semantics behind a `0.` semver version. Going further, we could also start including artifacthub changelog annotations to indicate breaking changes and required steps for updating: https://artifacthub.io/docs/topics/annotations/helm/.

nathanjrobertson:
I think the important thing about version numbers is documenting what they mean. Particularly in the context of where they'll be used by various automation frameworks.

I'm fine with 0.15.0, just add some documentation to the `README.md`. In my view, the "0.15" gets incremented to "0.16" when a breaking change in either the app or the chart happens., whereas the "0.15.0" becomes "0.15.1" when non breaking chart change or minor release of the app occurs. Basically, most people in an automation context will want automated updates of "safe" changes, but will want manual control of major and breaking changes.

helm/superset/Chart.yaml
```diff
@@ -29,7 +29,7 @@ maintainers:
   - name: craig-rueda
     email: craig@craigrueda.com
     url: https://github.com/craig-rueda
-version: 0.14.2
```
nathanjrobertson: I think it would be good to add a comment here stating the convention of the chart version. For example:

```
# The version of this chart (not Superset itself - that's appVersion). The convention for the chart version X.Y.Z is:
# X: Remains "0"
# Y: Is incremented if appVersion has a major or minor upgrade in version (eg. 4.0 to 4.1, or 4.1 to 5.0), or there is a breaking change in the chart itself.
# Z: "0" if Y is incremented, otherwise incremented for any non-breaking change to the chart or a patch release of the appVersion (eg. 4.1.0 to 4.1.1).
```

If somebody disagrees with my definition and wants something else, that's fine. But having an explicit definition is what we need.

brandon-kaplan: Good with the definition, thanks. I do think this belongs in the readme vs a comment, I'll update the PR.

rusackas:
Looks like pre-commit caught something... here's the usual blurb if it helps:

```
pre-commit install
```

```
pre-commit run --all-files
```

helm/superset/Chart.yaml
nathanjrobertson: I'd put a small comment in `values.yaml` as well. The penalty for somebody inadvertently changing the version number and not adhering to the convention in an environment where automation deploys updated versions (eg. CD in Kubernetes environments) is high. Something like:

```
# For version X.Y.Z, X=0, Y is incremented for Superset version upgrades and breaking changes in the chart, Z for patch releases and other non-breaking changes. See <<link_to_docs>> for details.
```

brandon-kaplan:
> Looks like pre-commit caught something... here's the usual blurb if it helps:
> 
> Superset uses Git pre-commit hooks courtesy of [pre-commit](https://pre-commit.com/). To install run the following:
> 
> ```
> pip3 install -r requirements/development.txt
> pre-commit install
> ```
> 
> A series of checks will now run when you make a git commit.
> 
> Alternatively it is possible to run pre-commit by running pre-commit manually:
> 
> ```
> pre-commit run --all-files
> ```

@rusackas should be good to go now, thanks

chgl (approved):

brandon-kaplan:
> @villebro as I think helm chart upgrade requires special handling

@mistercrunch any other code owners able to review here? @villebro seems MIA.

villebro (approved):

PR #33895, opened 2025-06-24
fix: intermittent retrieval error under high concurrency

### LINKED ISSUE #33828: Intermittent Dataset Retrieval Errors When Making Parallel API Requests
### Bug description

**Describe the bug**
This issue occurs specifically with the /dataset method.  It does not consistently return the correct dataset when called concurrently with Rison encoded filter looking for the "table_name". 
The problem manifests about 5% of the time, indicating it may be a race condition or a resource contention issue.

**To Reproduce**
1. Make multiple parallel API requests to the /dataset endpoint, each querying for different "table_name" values.
2. Compare the returned datasets with the expected results based on the initial queries.

**Additional context**

I'm developing an application that generates dashboards "on the fly" where each chart is basically fed by a dataset (in a 1:1 relationship). Since the dataset's table_name is a unique ID, I'm using it to retrieve the dataset through the API, giving it a fixed UUID based on the user/dashboard.

I am making approximately 12 concurrent requests to retrieve around 70 datasets, which are updated frequently.

What I've tried:
- When the error occurs, a subsequent call with the same query returns the correct dataset, suggesting that the issue may not be with the dataset itself but rather with how the API handles concurrent requests.
- Serial calls (one after the other) do not produce errors, indicating that the problem is likely related to concurrency.
- Using 'mitmweb' to monitor traffic revealed that the error originates from the Superset response.

**Superset version**
The issue has been observed in both Superset v5.0.0rc3 and v4.1.2, suggesting it may not be version-specific.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

MIKEX818:
Fixes #33828.

Rewrote the filter method to support high-concurrency dataset queries. Also:
	•	Fixed incomplete uppercase letter matching in regex
	•	Resolved undefined variables and missing request params
	•	Added correctness metric for debugging

### DIFF
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -17,18 +17,22 @@
 # pylint: disable=too-many-lines
 from __future__ import annotations
 
+import copy
 import logging
 from datetime import datetime
 from io import BytesIO
+from typing import Any, Callable, Dict
 from typing import Any, Callable
 from zipfile import is_zipfile, ZipFile
 
 from flask import request, Response, send_file
 from flask_appbuilder.api import expose, protect, rison, safe
 from flask_appbuilder.api.schemas import get_item_schema
+from flask_appbuilder.models.filters import Filters
 from flask_appbuilder.const import (
     API_RESULT_RES_KEY,
     API_SELECT_COLUMNS_RIS_KEY,
+    API_FILTERS_RIS_KEY,
 )
 from flask_appbuilder.models.sqla.interface import SQLAInterface
 from flask_babel import ngettext
@@ -298,6 +302,32 @@ class DatasetRestApi(BaseSupersetModelRestApi):
 
     list_outer_default_load = True
     show_outer_default_load = True
+
+    def _handle_filters_args(self, rison_args: Dict[str, Any]) -> Filters:
+        """
+        Handle filters arguments passed to the API endpoint.
+
+        Parses and applies filtering criteria provided as Rison-encoded arguments
+        to construct a Filters instance. This method ensures that each request 
+        uses an isolated Filters instance to avoid shared state issues 
+        in concurrent or asynchronous environments.
+
+        Example input:
+            rison_args = {
+                "filters": [
+                    {"col": "table_name", "opr": "eq", "value": "some_table"}
+                ]
+            }
+       :param rison_args: A dictionary of arguments parsed from the API request'sAdd commentMore actions
+                          Rison-encoded `q` parameter.
+        :returns: A Filters instance containing the applied filters.
+        """
+        filters = self.datamodel.get_filters(
+            search_columns=self.search_columns, search_filters=self.search_filters
+        )
+        # self._filters.clear_filters()
+        filters.rest_add_filters(rison_args.get(API_FILTERS_RIS_KEY, []))
+        return filters.get_joined_filters(self._base_filters)    
 
     @expose("/", methods=("POST",))
     @protect()
--- a/superset/datasets/api.py
+++ b/superset/datasets/api.py
@@ -26,7 +26,12 @@ from zipfile import is_zipfile, ZipFile
 from flask import request, Response, send_file
 from flask_appbuilder.api import expose, protect, rison as parse_rison, safe
 from flask_appbuilder.api.schemas import get_item_schema
-from flask_appbuilder.const import API_RESULT_RES_KEY, API_SELECT_COLUMNS_RIS_KEY
+from flask_appbuilder.const import (
+    API_FILTERS_RIS_KEY,
+    API_RESULT_RES_KEY,
+    API_SELECT_COLUMNS_RIS_KEY,
+)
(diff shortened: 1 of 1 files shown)


superset/datasets/api.py
```diff
@@ -17,18 +17,22 @@
 # pylint: disable=too-many-lines
 from __future__ import annotations
 
+import copy
```
hainenber: This import is not being used anywhere. Should be removed
```suggestion
```

superset/datasets/api.py
```diff
@@ -299,6 +303,32 @@ class DatasetRestApi(BaseSupersetModelRestApi):
...
+        uses an isolated Filters instance to avoid shared state issues 
+        in concurrent or asynchronous environments.
+
+        Example input:
+            rison_args = {
+                "filters": [
+                    {"col": "table_name", "opr": "eq", "value": "some_table"}
+                ]
+            }
+       :param rison_args: A dictionary of arguments parsed from the API request'sAdd commentMore actions
```
hainenber: nit: indent this sentence to be on par with remaining comments
```suggestion
        :param rison_args: A dictionary of arguments parsed from the API request'sAdd commentMore actions
```

superset/datasets/api.py
```diff
@@ -17,18 +17,22 @@
 # pylint: disable=too-many-lines
 from __future__ import annotations
 
+import copy
 import logging
 from datetime import datetime
 from io import BytesIO
+from typing import Any, Callable, Dict
```
hainenber: extraneous import
```suggestion
```

superset/datasets/api.py
```diff
@@ -17,18 +17,22 @@
...
 from typing import Any, Callable
```
hainenber, line 23: `Dict` type should be imported here
```suggestion
from typing import Any, Callable, Dict
```

superset/datasets/api.py
```diff
@@ -299,6 +303,32 @@ class DatasetRestApi(BaseSupersetModelRestApi):
     list_outer_default_load = True
     show_outer_default_load = True
 
+    def _handle_filters_args(self, rison_args: Dict[str, Any]) -> Filters:
```
sadpandajoe: I see the new function here, is it being called anywhere?

hainenber: It's likely an override for [existing FAB method](https://github.com/dpgaspar/Flask-AppBuilder/blob/e2f7fdecffa770eba55083879c9eaf84bf1ecde3/flask_appbuilder/api/__init__.py#L1975).

Vitor-Avila:
hey @dpgaspar this fix seems to be overwriting a FAB method. I wonder if you think this fix should live in FAB instead? Unsure if applicable but thought it would be good to get your thoughts. Thank you!

LisaHusband:
I think it is not appropriate to address this issue within App Builder, since it is based on Flask, and Flask’s native framework lacks robust support for high-concurrency scenarios. As a lightweight web framework, Flask is rarely used in applications requiring high-frequency asynchronous access — whereas Superset, which handles large datasets and frequent user interactions, is exactly such a case.

Therefore, for this problem, one recommended approach is to patch Superset directly. Alternatively, a more fundamental solution would be to either refactor Flask and Flask App Builder to better support concurrency, or migrate to a framework like Django, which natively handles high-concurrency environments more effectively.

Otherwise, even if the filter-related issue is resolved, it remains difficult to ensure the correctness of other parameters under concurrent access patterns — and performance issues may persist.

EnxDev:
Hey @MIKEX818, are you planning to continue working on this contribution? Also, please feel free to let me know if I can help in any way.

If it makes sense, it might also be helpful to add some tests to help prevent regressions where possible.

rusackas:
Seems we have a conflict to resolve, along with the open review comments. Would love to see this one through!

rusackas:
Rebased, and CI is looking better!

The open question @Vitor-Avila raised still stands: this overrides a FAB method, so should the fix live upstream in Flask-AppBuilder rather than here? @dpgaspar, what's your opinion? Patch it in Superset for now, or push it into FAB?

rusackas:
Thanks for chasing this one down, your diagnosis of the shared `Filters` state was spot on. The same fix merged to `master` via #39685, which moved the per-request `_handle_filters_args` override up into `base_api.py` so it covers every REST endpoint, not just datasets, and closed #33828. So this is covered now, I think, but let me know if you (dis)agree before I close it out.

rusackas:
Going ahead and closing this out as covered by #39685, which moved the same per-request `_handle_filters_args` override into `base_api.py` so it applies to every endpoint, not just datasets. Your diagnosis of the shared `Filters` state was exactly right... the repo-wide version just landed first. Thanks for chasing it down, and happy to reopen if I've got this wrong.

PR #33933, opened 2025-06-26
fix(dashboard): Fix subitem selection on dashboard download menu

tahvane1:
### SUMMARY

Dashboard download menu show both item selected on hover (as id is undefined for both). This is not probably correct way to fix as it is not still the key used but at least it works.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [X ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/menu/DownloadMenuItems/DownloadAsImage.tsx
+++ b/superset-frontend/src/dashboard/components/menu/DownloadMenuItems/DownloadAsImage.tsx
@@ -27,6 +27,7 @@ export default function DownloadAsImage({
   text,
   logEvent,
   dashboardTitle,
+  ...props
 }: {
   text: string;
   dashboardTitle: string;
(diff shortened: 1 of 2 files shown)


sadpandajoe:
@tahvane1 since i see there is a changes UI checked, can you post either before or after screenshots or a video of what the new change looks like?

tahvane1:
Sorry. This is not my normal environment and it is not that easy anymore. Try to remember to do that next time. I just stumbled to it while investigating another issue. This happens only when dashboard screenshot endpoint is not enabled. Before change both items are highlighted no matter which one is hovered over. After change only the hovered one. But like I already told this is not probably best way to fix as this only creates temporary unique key and still doesn't use the one given in code.

rusackas (approved):
Seems OK to me, but would have loved screenshots, or more details about what we should be expecting to see here. 

Approving, but adding tests or getting a second opinion here would be helpful. I'll spin up a test env for anyone that can reproduce the "after" state here at least.

rusackas:
Seems fine! Thanks for the fix!

Issue #24490, opened 2023-06-22
Trivy scan showed some vulnerability in superset image, is these vulnerabilities are still exist? If yes, how to resolve


nagarajmmu:
Hi Team

I wanted use superset for my application, I am planning to deploy Superset using helm in Kubernetes.
Before installing superset, I have scanned superset images by Trivy tool, Trivy listed below vulnerability, please let me know, how to resolve these vulnerebility to use superset. I am listing only CRITICAL and HIGH severity vulnerability.

Image Name: apache/superset
CRITICAL:
cve-2019-1010022
cve-2019-8457
HIGH:
CVE-2022-3715
CVE-2018-12934
CVE-2018-18483
CVE-2021-3530
CVE-2021-3549
CVE-2021-3826
CVE-2021-45078
CVE-2022-42916
CVE-2022-43551
CVE-2022-1304
CVE-2018-20796
CVE-2019-1010023
CVE-2019-9192
CVE-2018-6829
CVE-2021-33560
CVE-2018-5709
CVE-2017-17740
CVE-2022-27385
CVE-2022-29458
CVE-2017-11164
CVE-2017-7245
CVE-2017-7246
CVE-2019-20838
CVE-2011-4116
CVE-2020-16156
CVE-2021-36690
CVE-2022-35737
CVE-2022-1304
CVE-2022-29458
CVE-2008-4609
CVE-2013-7445
CVE-2019-12456
CVE-2019-19070
CVE-2019-19378
CVE-2019-19449
CVE-2019-19814
CVE-2020-11725
CVE-2020-12362
CVE-2021-26934
CVE-2021-32078
CVE-2021-3714
CVE-2021-3847
CVE-2021-3864
CVE-2021-39686
CVE-2021-4204
CVE-2022-0400
CVE-2022-0500
CVE-2022-1247
CVE-2022-2196
CVE-2022-25265
CVE-2022-2961
CVE-2022-3424
CVE-2022-3566
CVE-2022-3567
CVE-2022-39189
CVE-2022-4379
CVE-2022-43945
CVE-2022-45884
CVE-2022-45885
CVE-2022-45886
CVE-2022-45919
CVE-2023-23559

Image Name: jwilder/dockerize:latest
HIGH:
CVE-2017-16544
CVE-2017-15650

Imaage Name: postgres:14
CRITICAL:
CVE-2019-1010022
cve-2019-8457
HIGH:
CVE-2022-3715
CVE-2018-20796
CVE-2019-1010023
[...]

PR #33936, opened 2025-06-26
docs(security): add Q&A related to CVE scans to FAQ

sfirke:
### SUMMARY
We get enough posts in Slack and GitHub of CVE lists that look alarming (e.g., https://github.com/apache/superset/issues/24490) that I thought I'd add something to the FAQ. That way we can just point people to that answer.

(I also fixed a formatting bug in a different docs file)

### DIFF
--- a/docs/docs/faq.mdx
+++ b/docs/docs/faq.mdx
@@ -278,7 +278,7 @@ Alternatively, you can programmatically take regular exports of the assets as a 
 
 ## I ran a security scan of Superset and it showed dozens of "high" and "critical" vulnerabilities! Can you release a version of Superset without these?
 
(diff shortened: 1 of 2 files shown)


rusackas (approved):
Thank you!!

hainenber:
nitpicky but from your Bluetooth example, I think you're referring to vulnerabilities when scanning Superset's *container image* rather then PyPi distribution. 

Plus, I wouldn't say these OS vulnerabilities wouldn't hurt Superset, threat actors are no joke these days. Let's just say it isn't relevant to our immediate remediation scope and await for upstream folks to release security patches

sfirke:
good points @hainenber, thanks for reviewing! I've shortened it along the lines you suggested.

hainenber:
LPerfectTM :D

hainenber (approved):

PR #33939, opened 2025-06-26
fix(dremio): apply same fix as for drill to solve alias ambiguity

mistercrunch:
Applying the same fix for Dremio as was applied in drill recently here https://github.com/apache/superset/pull/32857


rusackas (approved):
LGTM!

giftig (approved):
Just woke up, can't read, ignore my comment.

PR #33946, opened 2025-06-27
feat: Remove charts from welcome page for users without access to edit charts

martimors:
As discussed [here](https://github.com/apache/superset/discussions/23802#discussion-5127077), it would be nice if users who cannot create or edit charts would not see these on the Welcome page.

This tries to follow the same pattern as was already there for users who can or cannot view saved queries.

**Any chance we can shoehorn this into the 5.x releases? I don't think it is a breaking change but it could be argued.

Here's a screen recording where I show the role without the permission assigned to a user, then log in as that user and see that the charts are not shown on the home page.

Then, towards the end, I show that if the user knows the URL they can still list charts, which I think is fine.

https://github.com/user-attachments/assets/b6385ab1-de2e-46d3-9824-e42cfccb8363

### SUMMARY

I decided to go for the `Chart` `can_edit` permission here, although there are several similar ones I could have picked, or a combination. In my view, this is good enough. We cannot use `Chart` `can_read` here obviously, because this is required for users to see the chart in dashboards.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Closes https://github.com/apache/superset/discussions/23802#discussion-5127077
- [ ] Required feature flags:
- [x] Changes UI


pbrostean:
@martimors any updates on this? Waiting badly for this feature..

martimors:
> @martimors any updates on this? Waiting badly for this feature..

Don't ask me, I haven't heard anything from the maintainers since submitting this almost half a year ago. If this is still relevant I don't mind picking it up again at some point, but I'd rather wait for some feedback and avoid wasting my time.

pbrostean:
> > @martimors any updates on this? Waiting badly for this feature..
> 
> Don't ask me, I haven't heard anything from the maintainers since submitting this almost half a year ago. If this is still relevant I don't mind picking it up again at some point, but I'd rather wait for some feedback and avoid wasting my time.

Fair point.

I suspect the automated checks need to be completed before anyone will look at it properly. There are still tests missing for your changes. Can you take a look at this or should I take a look?

martimors:
I've written tests, which all pass.

martimors:
> > > @martimors any updates on this? Waiting badly for this feature..
> > 
> > Don't ask me, I haven't heard anything from the maintainers since submitting this almost half a year ago. If this is still relevant I don't mind picking it up again at some point, but I'd rather wait for some feedback and avoid wasting my time.
> 
> Fair point.
> 
> I suspect the automated checks need to be completed before anyone will look at it properly. There are still tests missing for your changes. Can you take a look at this or should I take a look?
> 

I missed parts of your reply, I'm on mobile, my apologies. Nothing would make me happier than sharing the contribution with anyone willing to, feel free to branch or anything else from my work, it's public domain. I will more than probably not be able to pick this up this year anyways!

pbrostean:
@mistercrunch @sadpandajoe Could you give a quick update on whether you have scheduled a review for this pull request or whether there are still outstanding issues?

I tried to look at the outstanding issue with the Playwright tests, which are still marked as pending. However, I think the GitHub actions have hung up here. Please let me know if I can push this forward or what the hold-up is.

rusackas (commented):
Thanks for picking this back up @martimors, the feature itself makes sense. CI's red though, and I think it's the new tests: they call fetchMock.calls(...) but the existing ones use fetchMock.callHistory.calls(...). We're on fetch-mock v12 where the bare .calls() was dropped, so those three Without chart role tests should fail. Mind switching them to callHistory.calls and getting lint-frontend + sharded-jest green? Happy to take another look once it's passing.

martimors:
@cpnielsen Could you have a look if relevant?

superset-frontend/src/views/CRUD/utils.tsx
```diff
@@ -250,20 +260,25 @@ export const getRecentActivityObjs = (
...
-    );
+    return getFilteredChartsandDashboards(
+      addDangerToast,
+      filters,
+      undefined,
+      undefined,
+      canWriteChart,
+    ).then(({ other }) => {
+      res.other = other;
+      res.viewed = distinctRes.values().reverse();
```
sadpandajoe, line 279: I think this still lets chart cards appear on Home for users without `Chart.can_write`: `/recent_activity` can return `item_type: 'slice'` rows, and we copy them into `Viewed` here regardless of `canWriteChart`. The Recents panel also still loads edited charts through `getEditedObjects` when the Edited tab is opened. Can we filter chart/slice records from the Recents data path when `canWriteChart` is false and add a regression test with a recent slice entry?

rusackas:
Happy to try to move this along with a few commits if others don't have time.

cpnielsen:
> Happy to try to move this along with a few commits if others don't have time.

Please feel free. Me and @martimors are a little hung up at the moment, so unlikely to progress this in the near future.

rusackas:
First, I can't rebase this anyway, since it's from your `master` branch, so the guardrails won't even let me. 

But also, on second thought, thinking through the feature and permissions here a bit more. 

If a user can't edit a chart, why shouldn't they be allowed to SEE it on the welcome page? They still have permission to view it, so why not see it? 

You mentioned they don't see saved queries if they don't have view/read permission on them, which makes sense, but if they have permission to view the chart, why not _let_ them, and even fiddle with the control panels, but just not _save_ it? Isn't that more true to what those permissions are for?

superset-frontend/src/pages/Home/Home.test.tsx
```diff
@@ -137,7 +140,23 @@ const mockedPropsWithoutSqlRole = {
   ...mockedProps,
   user: {
     ...mockedProps.user,
-    roles: {},
+    roles: {
```
sadpandajoe, line 143: This fixture now satisfies `WelcomeProps`, so the three existing `@ts-expect-error` directives below are unused and `lint-frontend` fails with TS2578 even after fixing the fetch-mock calls. Can we remove those stale directives with this fixture change?

superset-frontend/src/pages/Home/Home.test.tsx
```diff
@@ -220,6 +239,31 @@ test('Without sql role - calls api methods in parallel on page load', async () =
...
+  */
+  // @ts-ignore-next-line
+  await renderWelcome(mockedPropsWithoutChartWriteRole);
+  expect(await screen.findByText('Dashboards')).toBeInTheDocument();
+});
+
+test('Without chart role - renders all panels on the page on page load', async () => {
+  // @ts-ignore-next-line
+  await renderWelcome(mockedPropsWithoutChartWriteRole);
+  const panels = await screen.findAllByText(/Dashboards|Recents|Saved queries/);
```
sadpandajoe, line 254: This test still passes if the Charts panel renders, because the regex never matches `Charts` and only counts the other three headings. Can we assert that `queryByText('Charts')` is absent so the permission-gating behavior is actually covered?

sadpandajoe (commented):
See inline comments.

rusackas:
Closing this out. The design question I raised a few weeks back never got answered: if a user can still view a chart, why hide it from the welcome page just because they can't edit it? GitHub also shows conflicts with `master` now. Thanks for the work here @martimors, happy to see this reopened once there's a clear answer on the permissions model.

PR #33958, opened 2025-06-27
fix: Dashboard native filter fixes

Vitor-Avila:
### SUMMARY
This PR fixes a few different things around native filters:
1. Proper red highlighting for required filters.
2. Do not show the Apply button when selecting the same option, as that would remove the filter.
3. Clicking on Apply after Clear All when there's a filter configured to select the first option by default (but it's not required) throws an error. Then, selecting a value would auto-fire the filter change.

This was co-authored with Claude. Feedback always encouraged and appreciated!

**Proper red highlighting for required filters**
Before: 

https://github.com/user-attachments/assets/9c9c3604-14ac-4890-a91b-988081c089b1 

After:

https://github.com/user-attachments/assets/ede9a0a9-e56b-41c4-a2b9-6a26cda24b65

**Do not show the Apply button when selecting the same option, as that would remove the filter**
Before:

https://github.com/user-attachments/assets/f47a523b-435c-40c2-88eb-2c22b5f7bc26

After:

https://github.com/user-attachments/assets/c53c8521-01f9-423f-ab03-256bdb1b84c9

**Clicking on Apply after Clear All when there's a filter configured to select the first option by default (but it's not required) throws an error. Then, selecting a value would auto-fire the filter change.**
Before:

https://github.com/user-attachments/assets/46bc27ae-7f53-4323-96ae-8732b52b55c4

After:

https://github.com/user-attachments/assets/d1770f36-d4aa-4349-8f13-5784004a1490

### TESTING INSTRUCTIONS
Some test coverage added. For manual testing, it's possible to replicate the videos and confirm the expected behavior happens.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
@@ -95,6 +95,8 @@ const FilterValue: FC<FilterControlProps> = ({
   orientation = FilterBarOrientation.Vertical,
   overflow = false,
   validateStatus,
+  clearAllTrigger,
+  onClearAllComplete,
 }) => {
   const { id, targets, filterType, adhoc_filters, time_range } = filter;
   const metadata = getChartMetadataRegistry().get(filterType);
@@ -273,6 +275,8 @@ const FilterValue: FC<FilterControlProps> = ({
       setFocusedFilter,
       unsetFocusedFilter,
       setFilterActive,
+      clearAllTrigger,
+      onClearAllComplete,
     }),
     [
       setDataMask,
@@ -281,13 +285,15 @@ const FilterValue: FC<FilterControlProps> = ({
       unsetHoveredFilter,
       setFocusedFilter,
       unsetFocusedFilter,
+      clearAllTrigger,
+      onClearAllComplete,
     ],
   );
 
   const filterState = useMemo(
     () => ({
-      validateStatus,
       ...filter.dataMask?.filterState,
+      validateStatus,
     }),
     [filter.dataMask?.filterState, validateStatus],
   );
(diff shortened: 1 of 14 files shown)


superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterControls/FilterValue.tsx
```diff
@@ -281,13 +285,15 @@ const FilterValue: FC<FilterControlProps> = ({
...
+      clearAllTrigger,
+      onClearAllComplete,
     ],
   );
 
   const filterState = useMemo(
     () => ({
-      validateStatus,
       ...filter.dataMask?.filterState,
+      validateStatus,
```
Vitor-Avila, line 296: This is the fix for missing red highlight indicator for required filters. At first I was confused on why this would not happen before, and apparently this became evident with the AntD migration to v5. RCA from Claude:
> Ant Design v5 moved to full CSS-in-JS architecture and removed CSS files. The custom `FormItem` styled component was missing error state styling that was previously
  provided by Ant Design's CSS files.

mistercrunch:
@sadpandajoe saw you added the `review-checkpoint` label, wondering if part of the review process should also trigger a `test-env:up` when deemed useful for future review. Also would be great if you can QA/test-env this PR when you have a moment!

sadpandajoe:
> @sadpandajoe saw you added the `review-checkpoint` label, wondering if part of the review process should also trigger a `test-env:up` when deemed useful for future review. Also would be great if you can QA/test-env this PR when you have a moment!

@mistercrunch the `review:checkpoint` isn't something QA has tested, this is just the last PR we have looked at an assigned during the daily PR review meeting. So we know which ones may not have any reviewers yet and need one for feedback and merge.

betodealmeida (approved):
This is awesome, lots of improvement! <3

kgabryje (approved):
Thanks for the improvement! The code looks good and it works well.
Given how the native filters are prone to regressions, it'd be great to do some more thorough end-to-end testing and maybe add more test cases to the cypress e2e tests. Not a blocker for this PR though!

msyavuz (approved):

PR #33959, opened 2025-06-27
docs: add FAQ entry around how dashboard filters are broadcasted

mistercrunch:
FAQ entry + note in context -> 

Screenshot 2025-06-27 at 12 43 53 PM

### DIFF
--- a/docs/docs/faq.mdx
+++ b/docs/docs/faq.mdx
@@ -283,3 +283,28 @@ You are talking about dependency CVEs: identified vulnerabilities in software th
 We address these dependency CVEs as best we can by regularly updating our dependencies to newer versions. We use bots to assist with that and cheerfully welcome pull requests from humans that fix dependency CVEs.
 
 The Superset [security team](https://superset.apache.org/docs/security/#reporting-security-vulnerabilities) focuses primarily on vulnerabilities _in Superset itself_. See our [CVEs page](https://superset.apache.org/docs/security/cves) for a list of past Superset CVEs.
+
+## How are dashboard filters broadcasted?
+
+In Superset, **filters on a dashboard are broadcasted to other charts by default** based on matching **column names**, even across different datasets.
+
+### Default behavior
+
+When you add a filter to a dashboard (e.g. selecting `country = Canada`), it will **automatically apply to all charts** that contain a column named `country`, regardless of which dataset they use. This default setting is labeled **"All panels"** in the filter’s configuration modal.
+
+If two datasets both have a `country` column, the filter will be applied to both. Superset assumes the shared column name reflects the same semantic dimension.
+
+### Things to watch out for
+
+- Superset only matches by **column name**, not content. If one dataset uses `country = Canada` and another uses `country = CA`, the filter might produce **invalid or empty results**.
+- This can lead to confusion or broken dashboards unless the datasets are aligned and filterable by the same values.
+- When building dashboards that mix datasets, consider how your dimensions are named and whether their values match.
+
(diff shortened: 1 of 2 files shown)


docs/docs/faq.mdx
```diff
@@ -275,3 +275,28 @@ No. Currently, there is no way to recover a deleted Superset dashboard/chart/dat
 Hence, it is recommended to take periodic backups of the metadata database. For recovery, you can launch a recovery instance of a Superset server with the backed-up copy of the DB attached and use the Export Dashboard button in the Superset UI (or the `superset export-dashboards` CLI command). Then, take the .zip file and import it into the current Superset instance.
 
 Alternatively, you can programmatically take regular exports of the assets as a backup.
+
+### How are dashboard filters broadcasted?
+
+In Superset, **filters on a dashboard are broadcasted to other charts by default** based on matching **column names**, even across different datasets.
```
Vitor-Avila, line 289: I don't think this is the default behavior (tho it's possible)?

By default, the filter would be automatically scoped to all charts **using the filter's dataset**. The user can manually go to the scoping rules and scope the filter to charts using another dataset if needed.

mistercrunch: From my understanding it's the way it works, by default it's "Apply to all panels", and different charts can be targeted. For the filter to "carry through", the column / dimension name needs to match, and it cannot be rewired for say a field called `country` to wire to one say `country_name`. But pretty sure it's the only way currently. It's pretty logical and generally users will expect such behavior. My guess is most dashboards are made out of charts that are either on the same dataset, or "neighbor datasets" that share a certain number of dimension where dimension that have the same name do align in terms of dimension member. Say two datasets `country`, one with `Canada` and another with `CA` on the same dashboard is pretty unusual and probably bad naming/practices in terms of data preparation,

mistercrunch: In many ways you confusion (or mine if you prove me wrong) makes the case for the FAQ entry. Also thinking that we could really use a clarification `NOTE` in the UI, maybe right under the `Scoping` tab here.  
Screenshot 2025-06-27 at 12 15 09 PM

Vitor-Avila: @mistercrunch here's a video (it's a bit long just to show end-to-end experience)

https://github.com/user-attachments/assets/1b7f5c5b-087f-43f4-976c-ba86b64f52c3

mistercrunch: oh I see, so by default it's not set to "All panels" but cross dataset filter would/will work if you check it?

Vitor-Avila: Exactly! If I manually update scoping the filter would work as long as column name is the same across datasets. 

There could be another edge case here: the test I did was to create a new filter in a dashboard that's already using two distinct datasets. I'm not sure what happens if I add the first chart, then create the filter (which would have the default "All panels" select) and then later add another chart from another dataset.

msyavuz: @mistercrunch Do we intend to change this default? I think it makes sense to have the filter not apply across datasets unless you extend the scope.

mistercrunch: AFAiC this PR is on hold until we decide whether we want to change defaults vs. document the current behavior. I didn't want to document something we might change shortly. Curious what @yousoph thinks.

Turns out what I wrote here *is not* the current behavior ... though maybe it should be, idk.

rusackas: I slapped a `hold` label on for good measure, but remove it whenever's clever.

PR #33960, opened 2025-06-28
fix: prevent cyclic dependencies in dashboard filters (#33908)

### LINKED ISSUE #33908: Dashboard filters allow creating bi-directional dependencies in an unexpected way with value filters
### Bug description

When configuring dashboard filters with the "Values are dependent on other filters" option, the system prevents creating bi-directional dependencies upfront by showing a "Cyclic dependency detected" message. However, it allows creating the same dependency indirectly after saving the initial filter configuration.

1. If Filter A is set to depend on Filter B, the system blocks setting Filter B to depend on Filter A immediately, showing a "Cyclic dependency detected" message.
2. However, if Filter A is saved with the dependency on Filter B, the system later allows editing Filter B to create the reverse dependency, effectively forming a bi-directional (cyclic) dependency.

This behavior is inconsistent and might lead to unexpected filter behavior. It appears to contradict the intended prevention of cyclic dependencies.
How to reproduce the bug

1. Create a dashboard filter (e.g., Filter A) with a dependency on another filter (e.g., Filter B) by enabling the Values are dependent on other filters option.
2. Save the filter configuration.
3. Open the dashboard filter configuration again, select Filter B, and set it to depend on Filter A.
4. Save the changes.

## Expected results

The system should prevent bi-directional (cyclic) dependencies for all filter types, not just date filters. Users should not be able to create a cyclic dependency by editing the filters after saving the initial configuration.

## Actual results

The platform allows creating cyclic dependencies indirectly after the initial configuration, even though it prevents it upfront. This behavior is inconsistent and may lead to unexpected filter behavior.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.11

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

saiaika:
## Description
Fixes #33908 - Prevents users from creating bi-directional (cyclic) dependencies between dashboard filters, including indirect cycles that could be created after saving initial configurations.

## Problem
The existing validation only checked for cyclic dependencies during immediate UI interactions, but users could still create indirect cycles by:
1. Setting up Filter A → depends on Filter B
2. Saving the configuration  
3. Later editing Filter B → to depend on Filter A
4. Saving again (this would create a cycle but wasn't blocked)

## Solution
- Added comprehensive validation in the `handleSave` method of `FiltersConfigModal.tsx`
- The validation now runs before every save operation and blocks saving if any cyclic dependency is detected
- Added proper error handling with user feedback when cycles are found
- Ensures all filter types are protected from creating bidirectional dependencies

## Changes Made
- **FiltersConfigModal.tsx**: Added `validateDependencies()` call in `handleSave` method with error checking
- **FiltersConfigModal.test.tsx**: Added comprehensive tests covering both immediate and indirect cyclic dependency scenarios
- **ISSUE_33908_FIX_SUMMARY.md**: Detailed documentation of the fix and verification

## Test Coverage
- ✅ Test immediate cyclic dependency blocking (existing functionality preserved)
- ✅ Test indirect cyclic dependency blocking (main issue fix)
- ✅ Test error message display when cycles are detected
- ✅ Test successful save when no cycles exist
- ✅ Verify focus is set to problematic filter when cycle is found

## Technical Details
The fix adds save-time validation by:
1. Calling `validateDependencies()` before saving
2. Checking form fields for dependency validation errors
3. Blocking save operation if cycles are detected
4. Providing user feedback by focusing on the problematic filter

## Verification
This fix has been verified to address the exact scenario described in issue #33908:
- ✅ Indirect cyclic dependencies are now blocked at save time
- ✅ Users receive clear feedback about the cyclic dependency
- ✅ The problematic filter is automatically focused for easy correction
- ✅ All existing functionality remains intact
- ✅ Works for all filter types (select, range, time, etc.)

## Testing
Run the following to test the changes:
```bash
cd superset-frontend
npm test -- FiltersConfigModal.test.tsx

### DIFF
--- a/ISSUE_33908_FIX_SUMMARY.md
+++ b/ISSUE_33908_FIX_SUMMARY.md
@@ -1,0 +1,241 @@
+# Apache Superset Issue #33908 Fix Summary
+
+## Issue Description
+**Problem**: Users could create bi-directional (cyclic) dependencies between dashboard filters indirectly after saving initial configurations. The system blocked immediate cyclic dependencies but allowed indirect cycles to be saved.
+
+**Original Behavior**: 
+- ✅ Immediate validation worked (blocked direct cyclic dependencies in UI)
+- ❌ Save-time validation was missing (allowed indirect cyclic dependencies)
+- ❌ Only applied to some filter types
+- ❌ Users could create A→B dependency, save, then add B→A dependency and save again
+
+## Root Cause Analysis
+
+### Frontend Validation Logic Location
+The cyclic dependency validation is implemented entirely on the frontend:
+
+**Core Files Examined:**
+1. **`superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/utils.ts`**
+   - Contains `hasCircularDependency()` function - the core validation algorithm
+   - Uses recursive depth-first search to detect cycles
+   - Works with dependency maps to traverse relationships
+
+2. **`superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigModal.tsx`**
+   - Main modal component managing filter configuration
+   - Contains `handleSave()` function - **THE CRITICAL MISSING PIECE**
+   - Contains `validateDependencies()` function that calls the core validation
+
+3. **`superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/FiltersConfigForm.tsx`**
+   - Individual filter configuration form
+   - Contains dependency UI via `DependencyList` component (lines 924-946)
+   - Calls `validateDependencies()` when dependencies change in real-time
+
+4. **`superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/FiltersConfigForm/DependencyList.tsx`**
+   - UI component for managing filter dependencies
+   - Handles adding/removing dependencies
+   - Triggers validation on changes
+
+### Validation Flow
+**Immediate Validation (Working):**
+```
+User changes dependencies → DependencyList.onDependenciesChange → validateDependencies() → UI shows error
+```
+
+**Save-time Validation (MISSING - Fixed):**
+```
+User clicks Save → handleSave() → [NO VALIDATION] → Save proceeds
+```
+
+## The Fix
+
+### What Was Changed
+**File: `FiltersConfigModal.tsx` - `handleSave()` function**
+
+**Before (Lines 460-487):**
+```typescript
+const handleSave = async () => {
+  const values: NativeFiltersForm | null = await validateForm(
+    form,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ISSUE_33908_FIX_SUMMARY.md
Comment (sadpandajoe) [inline ISSUE_33908_FIX_SUMMARY.md:1]: Why do we have this file?

Comment (msyavuz) [reply ISSUE_33908_FIX_SUMMARY.md:1]: Wondering the same thing, what is this file?

**Review by msyavuz — CHANGES_REQUESTED** [review]

**Comment by saiaika** [comment]
DELETED ISSUE_33908_FIX_SUMMARY.md

**Comment by msyavuz** [comment]
@saiaika Can you join our slack channel?

File: superset-frontend/src/dashboard/components/nativeFilters/FiltersConfigModal/utils.ts [hunk 1]
```diff
@@ -49,6 +50,39 @@ export const hasCircularDependency = (
...
+  if (errorField) {
+    return {
+      hasError: true,
+      filterId: errorField.name[FILTER_ID_INDEX] as string,
+      errors: errorField.errors,
+    };
+  }
+
+  return null;
+};
```
msyavuz:
```suggestion
interface FieldError {
  name: (string | number)[];
  errors: string[];
  warnings: string[];
}

/**
 * Helper function to find dependency validation errors in form fields.
 * This function improves code maintainability by centralizing the logic
 * for detecting cyclic dependency errors and extracting the problematic filter ID.
 *
 * @param fields - Array of form field error objects from Ant Design Form
 * @returns Object containing error status and filter ID, or null if no dependency errors
 */
export const findDependencyError = (fields: FieldError[]) => {
  const FILTER_ID_INDEX = 1;

  const errorField = fields.find(({ name, errors }) => {
    const [path, subpath, id] = name;
    return (
      path === 'filters' &&
      subpath === 'dependencies' &&
      typeof id === 'string' &&
      errors?.length > 0
    );
  });

  if (!errorField) {
    return null;
  }

  const filterId = errorField.name[FILTER_ID_INDEX];

  return {
    hasError: true,
    filterId: filterId as string,
    errors: errorField.errors,
  };
};

```

saiaika:
> @saiaika Can you join our slack channel?

Im in your slack channel

msyavuz:
@saiaika You put `i don't have` on the survey form field for slack handle. I don't have any idea who you are in the channel. Would you kindly send me a message there?

saiaika:
> @saiaika You put `i don't have` on the survey form field for slack handle. I don't have any idea who you are in the channel. Would you kindly send me a message there?

I filled out that form earlier and later joined Slack channel

PR #33963, opened 2025-06-28
fix(handlebars): remove serverPaginationControlSetRow from control pa…

### LINKED ISSUE #29657: Handlebars chart - pagination support
### Bug description

When enabling "SERVER PAGINATION" option in Handlebars chart, these is no impact on the chart generated.

Expected result: Pagination list and "show entries" option menu, same as for Table chart.

Each time we click on different page, I would expect to viewer re-render with updated 'data' variable.

### How to reproduce the bug

Any dataset with more than 10 rows is not properly paginated in Handlebars chart

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

LisaHusband:
### SUMMARY
fix(handlebars): remove serverPaginationControlSetRow from control panel config

This PR removes the `serverPaginationControlSetRow` from the Handlebars chart control panel configuration to fix pagination UI issues when "SERVER PAGINATION" option is enabled but has no effect.

Before:
图片
After:
图片

### TESTING INSTRUCTIONS
1. Run Superset and open a Handlebars chart with a dataset exceeding 10 rows.
2. Verify that the pagination controls related to server pagination do not appear.
3. Confirm the chart data renders correctly without pagination UI interference.

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #29657  
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration

### DIFF
--- a/superset-frontend/src/features/userInfo/UserInfoModal.tsx
+++ b/superset-frontend/src/features/userInfo/UserInfoModal.tsx
@@ -55,7 +55,7 @@ function UserInfoModal({
       addSuccessToast(
         isEditMode
           ? t('The user was updated successfully')
-          : t('The password reset was successfull'),
+          : t('The password reset was successful'),
       );
       onSave();
     } catch (error) {
--- a/superset-frontend/src/pages/UserInfo/index.tsx
+++ b/superset-frontend/src/pages/UserInfo/index.tsx
@@ -66,7 +66,6 @@ const StyledLayout = styled.div`
       padding-left: ${theme.sizeUnit * 4}px;
       position: relative;
       height: ${theme.sizeUnit * 15}px;
-      background-color: ${theme.colors.grayscale.light5};
     }
   `}
 `;


msyavuz (changes requested):
Hey, @LisaHusband, thanks for tackling this. Can't we simply remove `serverPaginationControlSetRow` and related stuff from `plugins/plugin-chart-handlebars/src/plugin/controlPanel.tsx` and from `controls` instead.

LisaHusband:
No, we can't, because other charts rely on this control and it works properly for them. It's just that in the chart involved in this issue, it doesn't function as expected. Therefore, the adjustment should be limited to this specific chart only. 😄

msyavuz:
Removing the control from the mentioned location should not have an effect on other charts

LisaHusband:
> Removing the control from the mentioned location should not have an effect on other charts

I made a mistake in checking the file path and thought it was the upper-level class file. 🥲

PR #33966, opened 2025-06-29
fix(plugin-chart-echarts): last month doesn't appear on X axis for time series line chart

### LINKED ISSUE #25099: Last date/time label missing on chart X axis
Hi, I managed to run a superset docker, connected to my own database and plotted some charts successfully.
However, it didn't show up the last day in X axis, has anybody faced the same problem, please help.
Thanks for your time.

#### Screenshots

superset

### Environment
Docker

Sameerali0:
### SUMMARY

The last month label wasnt visible for the time series line chart so I turned the showMaxLabel config to true.
Screenshot_20250629_100030

Fixes: https://github.com/apache/superset/issues/25099, https://github.com/apache/superset/issues/33905

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #33905
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -505,6 +505,7 @@ export default function transformProps(
       formatter: xAxisFormatter,
       rotate: xAxisLabelRotation,
       interval: xAxisLabelInterval,
+      showMaxLabel: true,
     },
     minorTick: { show: minorTicks },
     minInterval:


Sameerali0:
@msyavuz

msyavuz (commented):
Can we write a test to check if last month is shown

sfirke:
@Sameerali0 @msyavuz FYI here is the original bug report with more detail: https://github.com/apache/superset/issues/25099 that history links to a prior fix that had to be reverted because the result was worse than the original bug. Hopefully that gives some useful info to check for and avoid regressions with this fix. (And if this gets merged you can also mark that issue closed)

msyavuz:
@sfirke Oh, thanks for bringing that up. @Sameerali0 can we make sure the linked regression is not happening?

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -505,6 +505,7 @@ export default function transformProps(
       formatter: xAxisFormatter,
       rotate: xAxisLabelRotation,
       interval: xAxisLabelInterval,
+      showMaxLabel: true,
```
michael-s-molina, line 508: We should probably also set `showMinLabel: true` for consistency. 

Could you also replicate these configs in `MixedTimeseries`?

PR #33974, opened 2025-06-30
fix(i18n): Update Japanese translations

aikawa-ohno:
The Japanese translation has been updated.

### DIFF
--- a/superset/translations/ja/LC_MESSAGES/messages.po
+++ b/superset/translations/ja/LC_MESSAGES/messages.po
@@ -17,7 +17,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2025-04-29 12:34+0330\n"
+"POT-Creation-Date: 2025-06-30 15:30+0900\n"
 "PO-Revision-Date: 2024-05-14 13:30+0900\n"
 "Last-Translator: Yuri Umezaki <bungoume@gmail.com>\n"
 "Language: ja\n"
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):
Blindly approving since I don't speak japanese, but I'm assuming this is a significant step forward. 

Now I'm thinking it'd be good to keep a list of github username that we can call upon for each language, so we can tag them for reviews. 

In the meantime we'll generally have to assume that most updates are incrementally better than the previous state.

rusackas:
Claude likes this diff:

Looking at these Japanese translation updates, they generally appear to be improvements in consistency and accuracy. Here are the key patterns I notice:
Positive Changes:

Terminology Consistency:

"グラフ" (graph) → "チャート" (chart) - Better consistency with modern UI terminology
"電子メールレポートがアクティブです" → "電子メールレポートが有効です" - More natural Japanese


Translation Quality Improvements:

"1ヶ月前" instead of just "月" for "1 month ago" - More precise
"3次元" instead of "3日" for "3D" - Correct translation
"10/90 パーセンタイル" with proper numbers instead of "9/91"


Removing "fuzzy" Markers: Many translations had their #, fuzzy markers removed, indicating they've been reviewed and finalized.
Better Formality:

Adding "。" (periods) to error messages for proper sentence structure
More consistent use of polite forms



Potential Concerns:

Question Mark Changes:

"?" → "？" (full-width question mark) - This is actually correct for Japanese typography


Some entries show blank translations:
msgid "Email is required"
msgstr ""
This should probably have a Japanese translation.
Technical terms: Some technical terms like "catalog", "schema" are kept in katakana, which is appropriate.

Overall, these changes represent a positive update that improves consistency, accuracy, and naturalness of the Japanese translations. The systematic replacement of "グラフ" with "チャート" and other terminology standardizations show this was a well-planned localization improvement.

aikawa-ohno:
@mistercrunch @rusackas 
Thank you for your comments and merging.

PR #33603, opened 2025-05-28
feat(deckgl): add support for OpenStreetMap as our new default and make "tile-providers" more configurable

plavacquery:
### SUMMARY
This PR introduces support for OpenStreetMap as the tile provider behind deckgl charts. It also changes the default tile-provider to be OpenStreetMap, since Mapbox isn't as open and requires configuration + an API key.  This PR makes tile-providers configurable, so it's possible for those who have an API key and prefer Mapbox to match the default as of before-this-PR.

Changing defaults here is a "breaking change" in the sense that the default is changing, but you can reproduce the previous behavior with a bit of documented configuration.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
image

### TESTING INSTRUCTIONS

In config.py set the following
```py
DECKGL_BASE_MAP = [
     ['tile://https://c.tile.openstreetmap.org/{z}/{x}/{y}.png', 'OpenStreetMap' ],
]

ENABLE_CORS = True
CORS_OPTIONS: dict[Any, Any] = {
    "origins": [
        "https://c.tile.openstreetmap.org",
    ]
}

```
And in TALISMAN_DEV_CONF
add "https://c.tile.openstreetmap.org" to 'connect-src' array

### ADDITIONAL INFORMATION

- [x] Has associated issue :https://github.com/apache/superset/discussions/27475. Fixes https://github.com/apache/superset/issues/30763
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34017, opened 2025-07-01
feat(deckgl): add new color controls with color breakpoints

DamianPendrak:
### SUMMARY
Add new Controls that handle fixed colors, palette, and color breakpoints based on selection. Implement new color breakpoints Control to set colors of the Deck.gl Charts based on the Metric value.

Screenshot 2025-07-01 at 20 33 12

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Very cool! Looks like it needs a rebase, but I'll spin up a test environment in the meantime.

kgabryje:
Looks very cool!
A few questions:
1. Changing the color breakpoints and scheme types prompts to run a new query. Could we make it so that the changes are reflected immediately?
2. Could you add tooltips to the new controls?
3. After I added the breakpoints, the values that don't match any breakpoint take the color of the first breakpoint, which could be confusing (screenshot - the value 2415 and the values below 1000 use the color of 1000-1500 breakpoint). Maybe we should add a default color control?

image

superset-frontend/plugins/legacy-preset-chart-deckgl/src/utils.ts
```diff
@@ -190,3 +192,19 @@ export function getBuckets(
 
   return buckets;
 }
+
+export function getColorBreakpointsBuckets(fd: QueryFormData) {
```
kgabryje: nit: maybe this function could accepts only the breakpoints as the argument instead of the whole form data object? That way we might avoid some rerenders caused by passing form data to a dependency array in useEffect

superset-frontend/plugins/legacy-preset-chart-deckgl/src/utilities/utils.ts
```diff
@@ -16,8 +16,33 @@
...
+export type ColorSchemeType =
+  (typeof COLOR_SCHEME_TYPES)[keyof typeof COLOR_SCHEME_TYPES];
+
 /* eslint camelcase: 0 */
 
 export function formatSelectOptions(options: (string | number)[]) {
   return options.map(opt => [opt, opt.toString()]);
 }
+
+export const getSelectedColorSchemeType = (
```
kgabryje: Do we need this function? Can't we just call formData.color_scheme_type directly?

superset-frontend/src/explore/components/controls/ColorBreakpointsControl/ColorBreakpointOption.tsx
```diff
@@ -0,0 +1,66 @@
...
+import ColorBreakpointPopoverTrigger from './ColorBreakpointPopoverTrigger';
+import { DragContainer } from '../OptionControls';
+import Option from '../DndColumnSelectControl/Option';
+
+const BreakpointColorPreview = styled.div`
+  width: ${({ theme }) => theme.sizeUnit * 3}px;
+  height: ${({ theme }) => theme.sizeUnit * 3}px;
+  border-radius: ${({ theme }) => theme.sizeUnit / 2}px;
+  background: ${(props: { color: string }) => props.color};
+  margin-right: ${({ theme }) => theme.sizeUnit * 1.5}px;
```
kgabryje: Can we use the full size units? Whatever looks better, theme.sizeUnit or theme.sizeUnit *2

DamianPendrak:
> Very cool! Looks like it needs a rebase, but I'll spin up a test environment in the meantime.

Thanks @rusackas! It does, but the bump of the deck.gl dependencies in #33603 broke the deck.gl charts for me on local env in the master branch. Is it just me, or should I revert the deck.gl packages versions bump on this branch?

Screenshot 2025-07-09 at 15 56 44

amaannawab923:
Same here , it breaks on local setup @DamianPendrak , https://github.com/apache/superset/pull/33603 ... tried to fix it while maintaining the changes from the open street maps ... but it keeps on breaking

superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Hex/Hex.tsx
```diff
@@ -84,6 +84,9 @@ export const getLayer: GetLayerType<HexagonLayer> = function ({
       ? (p: number[]) => getColorForBreakpoints(aggFunc, p, colorBreakpoints)
       : aggFunc;
 
+  console.log('colorBreakpoints', colorBreakpoints);
+  console.log('colorRange', colorRange);
```
kgabryje: Could you remove the console logs?

DamianPendrak: Done

kgabryje (approved):
Works great! LGTM

mistercrunch:
Looks like this PR has bad merge conflicts and reverted things that were set up in https://github.com/apache/superset/pull/33603#issuecomment-3074408905.

DamianPendrak:
@mistercrunch yes, I didn't manage to bring back the revert before merging to master, as I explained in https://github.com/apache/superset/pull/34123#issuecomment-3068851131

mistercrunch:
Oh so this was this intentional then? Not sure if the reviewer/merger knew we were reverting a PR as part of this one. 

I created https://github.com/apache/superset/pull/34176  thinking this was a mistake, but unclear if / what is the path forward. I'm not sure why these 2 PRs had to be mutually exclusive but you all should work on a fix, happy to review.

DamianPendrak:
I'm sorry, I didn't communicate that here, my mistake. I was going to bring back the reverted PR, but it got merged before passing QA. The #33603 breaks all deck.gl Charts on local environment, but I thought it breaks on ephemeral too, so I reverted it

mistercrunch:
Oh I wasn't clear on https://github.com/apache/superset/pull/33603 breaking deck.gl - are we sure that's the case?

DamianPendrak:
Yes, only breaking local environment with Docker is confirmed. Not sure about ephemeral env, but @amaannawab923 says it should work on ephemeral as building with `npm run build` works

mistercrunch:
Good to know, do we know more about what's happening? I think I see this in my console:
```
Uncaught TypeError: Cannot read properties of null (reading 'stack')
    at eval (overlay.js:608:17)
```
But there are so many other messages in the damn console that I can't tell if that's the actual/exact issue. We'll take on cleaning up the console, but in the meantime we should get to the bottom of this....

So somehow `npm run build` works and `npm run dev` leads to that msg?

amaannawab923:
@mistercrunch 
The Reason for this is , somehow the node modules are getting deleted on local environment , some problem with versioning and it tries to find those dependancies within node modules in local dev environment and when it doesnt , it throws out the error "Module not found"
I tried patching it multiple times with missing methods and dependancies , Even going inside Node modules folder and editing the source code to make it work but it doesnt work out

amaannawab923:
But somehow if i do npm run build and use bundled code , i stopped getting those errors and thats how i was able to implement this dark theme for deck gl maps 
Screenshot 2025-07-09 at 1 23 54 PM

PR #34018, opened 2025-07-01
fix: Update spacing on echart legends

sadpandajoe:
### SUMMARY

The legend looks a bit more squished on master now than it was before with 4.X and 5.X. This PR will get it back to that size.

Before:
Screenshot 2025-07-01 at 12 34 44 PM

After:
Screenshot 2025-07-01 at 12 49 04 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/packages/superset-ui-core/src/utils/html.tsx
```diff
@@ -35,6 +35,13 @@ const xssFilter = new FilterXSS({
       'width',
       'muted',
     ],
+    table: ['width', 'border', 'align', 'valign', 'style'],
+    tr: ['rowspan', 'align', 'valign', 'style'],
+    td: ['width', 'rowspan', 'colspan', 'align', 'valign', 'style'],
+    th: ['width', 'rowspan', 'colspan', 'align', 'valign', 'style'],
+    tbody: ['align', 'valign', 'style'],
+    thead: ['align', 'valign', 'style'],
+    tfoot: ['align', 'valign', 'style'],
```
msyavuz, line 44: Not sure about allowing style tags here, can't we change the styles directly?
https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html#style-attribute-that-breaks-up-an-expression

sadpandajoe: I think we tried to add the styles here: https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-core/src/utils/tooltip.ts, but when we sanitized it, it then broke the spacing and formatting. If we think what we did with theming could change how we redo tooltips, we should probably implement it the correct way.

msyavuz: Yeah, those styles do nothing. I don't think theming can change how we handle tooltips, maybe setting `classname` might be the better way to do this?

betodealmeida (approved):
LGTM

eschutho (approved):

PR #34028, opened 2025-07-02
chore: refactor react-syntax-highlither to handle dark themes

mistercrunch:
It appears `react-syntax-highlither` has been used directly without wrappers in the past, making it hard to style / theme it properly as it's used in the app.

In this PR:
- write a simple wrapper for `react-syntax-highlither` that's theme-aware
- refactor all usage of `react-syntax-highlither` in the main app to use this new component
- basic styling that switches based on whether the theme is dark or not (doesn't use the intricate brand colors for now)
- basic unit tests and storybook for the new wrapper

Screenshot 2025-07-02 at 12 37 26 PM
Screenshot 2025-07-02 at 11 44 23 AM


msyavuz (approved):
Looks good to me. Unrelated to this pr and this also exists in master but there is a z-index issue with the tooltip of Copy button:
image

PR #33189, opened 2025-04-19
fix: localization of humanize naturaltime

ceenayekta:
### SUMMARY
This pull request introduces localization support to the humanize library used within Apache Superset. The change enables better handling of localized text and formats, improving the internationalization capabilities of the platform. The implementation includes updates to integrate localization logic into the existing humanize library usage, ensuring consistency across supported languages.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
Spanish:
image
French:
image

### TESTING INSTRUCTIONS
1. Pull the branch humanize_lib_localization and start the development server.
2. Verify that text formatting and human-readable labels (e.g., dates, durations) are properly localized based on the configured language.
3. Test the behavior across different supported locales to ensure functionality and consistency.
4. Check for any regressions in areas where the humanize library is used.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [x] Introduces new feature or API
- [ ] Removes existing feature or API

-----
Feel free to modify the description further to include any additional details or specifications related to your changes.

PR #34051, opened 2025-07-03
fix(translations): Fix translation of time-related strings like "7 seconds ago", "a minute ago", etc

### LINKED ISSUE #7792: Superset seems to be ignoring few frontend translations
Superset internationalization does not seem to be working on few components at the frontend.
The 'Add new slice' menu @ superset/assets/src/addSlice/AddSliceContainer.jsx is one example, even if the @superset-ui module has been called.

`import { t } from '@superset-ui/translation';`
`t('Choose a datasource')`

### Expected results

All UI elements should be translated.

### Actual results

Few frontend elements seem to ignore the translation, i couldn't reproduce the issue on backend rendered components like tables.

#### How to reproduce the bug

Download the current main branch, or 0.32.0rc2, haven't tested on other releases.

### Environment

- superset version: 0.32.0rc2
- python version: 3.6
- node.js version: 12.0

### Checklist

- [*] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [*] I have reproduced the issue with at least the latest released version of superset.
- [*] I have checked the issue tracker for the same issue and I haven't found one similar.

PolinaFam:
### SUMMARY
**Issue Description**
Switching the language to English does not fully apply, leaving some time-related strings in the previous language. All untranslated time-related strings are created by dayjs and humanize libraries.

**Solution**
Fix dayjs strings: move all dayjs locale imports from `useLocale.ts` hook to `preamble.ts`. Without them, `dayjs.locale()` calls have no effect, leaving dates/times in the default language.
Fix humanized strings: translate `changed_on_humanized` and `created_on_humanized` using current locale from flask_babel.

fixes: #33630
fixes: #7792

screenshot3
screenshot4

### TESTING INSTRUCTIONS
1. Add next lines to config file:
```py
BABEL_DEFAULT_LOCALE` = "ru"
LANGUAGES = {
  "ru": {"flag": "ru", "name": "Русский"},
  "en": {"flag": "us", "name": "English"}
}
```
2. Start the application.
3. Switch the interface language to English.
4. Verify the translations.

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #33630, #7792
- [ ] Required feature flags:
- [ ] Changes UI


superset/models/helpers.py
```diff
@@ -546,10 +546,28 @@ def changed_on_utc(self) -> str:
 
     @property
     def changed_on_humanized(self) -> str:
+        locale = str(get_locale() or 'en')
+        try:
+            humanize.i18n.activate(locale)
+            result = humanize.naturaltime(datetime.now() - self.changed_on)
+            humanize.i18n.deactivate()
+            return result
+        except:
```
mistercrunch: Are we expecting this to fail? In what cases? Not sure if swallowing anything/everything is best here, often if we know something might fail we'd try to swallow just that one particular exception, and maybe `logging.exception(e)` to handle the rest of the pessimisticism, so it's unexcpected, handled - but **logged** somewhere. 

Otherwise you may just never know that say your particular locale isn't handled by `humanize`.

superset/models/helpers.py
```diff
@@ -546,10 +546,28 @@ def changed_on_utc(self) -> str:
 
     @property
     def changed_on_humanized(self) -> str:
+        locale = str(get_locale() or 'en')
```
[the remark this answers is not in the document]
PolinaFam: Fixed

superset/models/helpers.py
```diff
@@ -546,10 +546,28 @@ def changed_on_utc(self) -> str:
...
+        try:
+            humanize.i18n.activate(locale)
+            result = humanize.naturaltime(datetime.now() - self.created_on)
+            humanize.i18n.deactivate()
+            return result
+        except:
+            pass
+        # Fallback to English
         return humanize.naturaltime(datetime.now() - self.created_on)
```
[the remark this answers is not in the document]
PolinaFam: Fixed

superset/models/helpers.py
PolinaFam: There is only one case to fail is missing locale in `humanize`. Added logging for missing locale and handle default humanize `en` locale as humanize.i18n.activate('en') will throw exception.

superset-frontend/src/explore/components/controls/DateFilterControl/tests/CustomFrame.test.tsx
```diff
@@ -26,6 +26,7 @@ import {
   waitFor,
 } from 'spec/helpers/testing-library';
 import { CustomFrame } from '../components';
+import 'dayjs/locale/fr';
```
mistercrunch: Why only fr here? Mistake/leftover?

superset-frontend/src/preamble.ts
```diff
@@ -26,6 +26,20 @@ import setupFormatters from './setup/setupFormatters';
 import setupDashboardComponents from './setup/setupDashboardComponents';
 import { User } from './types/bootstrapTypes';
 import getBootstrapData, { applicationRoot } from './utils/getBootstrapData';
+import 'dayjs/locale/en';
```
mistercrunch: What's the reason for moving this from `superset-frontend/src/hooks/useLocale.ts` to this location? I'm almost wondering if we should try to hot-load this using a dynamic import in webpack so there's no penaltly on always loading all languages and bringing this all in package/memory in the browser. Ideally webpack would prepare one bundle for each language and only load the language that's needed for the current user ...

It was already this way, but maybe that's an opportunity to improve things here.

superset/models/helpers.py
```diff
@@ -544,13 +545,28 @@ def changed_on_utc(self) -> str:
         # Convert naive datetime to UTC
         return self.changed_on.astimezone(pytz.utc).strftime("%Y-%m-%dT%H:%M:%S.%f%z")
 
+    def _format_time_humanized(self, timestamp: datetime) -> str:
```
mistercrunch, line 544: Seems this improvement should be backend only (?) I'm wondering why we need to touch the frontend at all.

superset/models/helpers.py
```diff
@@ -103,6 +103,7 @@
 VIRTUAL_TABLE_ALIAS = "virtual_table"
 SERIES_LIMIT_SUBQ_ALIAS = "series_limit"
 ADVANCED_DATA_TYPES = config["ADVANCED_DATA_TYPES"]
+DEFAULT_LOCALE = "en"
```
mistercrunch: Not sure if this is the right spot for this constant ... Seems we use `get_locale()` in a least one other place (superset/views/base.py).

Oh actually I just found `BABEL_DEFAULT_LOCALE` in https://github.com/apache/superset/blob/master/superset/config.py#L378 , you should use that instead.

superset-frontend/src/preamble.ts
PolinaFam: String `dayjs.locale(bootstrapData.common.locale);` in `preamble.ts` has no effect without these imports

superset/models/helpers.py
PolinaFam: The problem with time-related strings is connected not only with `humanize` but also with `dayjs`. It is described in PR Summary

superset-frontend/src/explore/components/controls/DateFilterControl/tests/CustomFrame.test.tsx
PolinaFam: Test `should translate Date Picker` requires `fr` locale

superset/models/helpers.py
mistercrunch: btw I'm unclear whether `BABEL_DEFAULT_LOCALE` works as expected currently, if I could find time I'd try it, but given my semi-deep-dive into translation, I'm not sure if the stars align properly when setting this config flag.

mistercrunch: But if you need to latch someplace for a default locale, that's probably the right hook ultimately.

PolinaFam: Actually there is no need to use `BABEL_DEFAULT_LOCALE` or `DEFAULT_LOCALE` as `get_locale()` will return default locale if it wasn't changed. Updated PR according to this discussion.

PolinaFam:
Hi @mistercrunch
Could you help me to clarify what is the problem with `test-postgres` job?

mistercrunch (approved):
LGTM

mistercrunch (changes requested):
[oops, approved the PR by mistake] adding a request-changes on top ...

About the frontend changes, can we have preamble simply import useLocale? About the test that requires `fr`, i'm thinking it should work without this import as long as tests import preamble (?)

PolinaFam:
@mistercrunch 
Do you suggest to return all `dayjs/locale/...` imports in `useLocale.ts` and add `import './hooks/useLocale` in `preamble.ts`?

mistercrunch:
Yes exactly!

mistercrunch (approved):
LGTM though didn't QA personally and based on pure code review.

PolinaFam:
@mistercrunch 
There's another PR with overlapping changes (#33189). I reviewed both and noticed that my PR also fixes dayjs issue but missed fixes in `sql_lab.py` file.

rusackas:
Not sure which one we should accept first, if it's a venn-diagram. If we merge the other, would you want to touch up yours? Or would you like to adopt/credit some of their changes now? I'm open to suggestion 😅

mistercrunch:
Other one has CI issues, this one is good to go. Mergin'!

PR #34057, opened 2025-07-03
chore: move auth e2e tests to component tests

sadpandajoe:
### SUMMARY

Looking at some of the e2e in auth they should be a component test versus an e2e test as we are using it to check form elements. This PR removes these tests from e2e and moves them to component tests to hopefully speed up the e2e pipeline.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch (approved):
LGTM

msyavuz (approved):
Makes sense to speed up the pipeline, my intention with making those e2e were to check if fab views or frontend is loaded for login page and regression of that is highly unlikely.

sadpandajoe:
> Makes sense to speed up the pipeline, my intention with making those e2e were to check if fab views or frontend is loaded for login page and regression of that is highly unlikely.

I can see that, but if the fields for logging in is broken the assumption there is that all tests would be broken since we wouldn't be able to log in. But just in case if things are broken, it'd be easier to catch in the component as people may just want to skip the tests instead of investigating.

PR #34067, opened 2025-07-04
style(Button): Vertically align icons across all buttons

EnxDev:
### SUMMARY

This PR improves visual consistency across the UI:

- Vertically aligns icons within buttons
- Fixes layout issues in the Alerts & Reports UI
- Adjusts the background color on the Home page
- Updates the icon color in toast notifications for better visibility
- Adjusts the size for BigNumberChartTile icon inside chart section to ensure better visual alignment

- Before
Screenshot 2025-07-11 172943

image (1)

- After
Screenshot 2025-07-11 171634
Screenshot 2025-07-11 171631

- Before
Screenshot 2025-07-11 154021

- After
Screenshot 2025-07-11 154107

- Before
Screenshot 2025-07-11 160338

- After
Screenshot 2025-07-18 223711

- Before
Screenshot 2025-07-04 101453

- After
Screenshot 2025-07-04 101621

- Before
Screenshot 2025-07-04 101311

- After
Screenshot 2025-07-04 101430

- Before
Screenshot 2025-07-03 194838

- After
Screenshot 2025-07-03 195234

- Before
Screenshot 2025-07-04 103953

- After
Screenshot 2025-07-04 111118

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


mistercrunch:
Mmmh, does antd gets this wrong, or is it because we're not using antd right?

I noticed antd's button expects an `icon` prop, but often we pass the icon as a children instead. Is that the case that what we're addressing here?

If it's the case I'd much rather fix our use of `<Button icon={} />` throughout the codebase than patching it with styles in the wrapper.

From my understanding, if/when using button with the icon prop, antd should do all the right things in terms of alignement with standard sizing and all. Guessing it may pass-in/override the icon sizing based on the button's size attributes, also probably should get the color of the icon to match the text and all. I've seen places where we `theme.colors.warning.light4` on the icon to try and match the text's color and that's just stuff antd will do properly when used right ....

superset-frontend/src/pages/ChartList/index.tsx
```diff
@@ -774,10 +774,17 @@ function ChartList(props: ChartListProps) {
   if (canCreate) {
     subMenuButtons.push({
       name: (
-        <>
+        <span
```
mistercrunch: Here it seems that we could either change the `SubMenu` components to receive actual buttons directly (instead of data structures that get turned into buttons (not sure why someone created this indirection?)), or you could extend it to expect receiving a `icon` key in as part of the object.

Seems passing in an array of buttons is more reasonable, but looks like SubMenu is used in many places:
```
12:06 $ git grep "<SubMenu"
superset-frontend/packages/superset-ui-core/src/ui-overrides/types.ts:  'home.submenu': ComponentType<SubMenuProps>;
superset-frontend/src/features/home/ActivityTable.tsx:      <SubMenu
superset-frontend/src/features/home/ChartTable.tsx:      <SubMenu
superset-frontend/src/features/home/DashboardTable.tsx:      <SubMenu
superset-frontend/src/features/home/LanguagePicker.tsx:    <SubMenu
superset-frontend/src/features/home/SavedQueries.tsx:      <SubMenu
superset-frontend/src/features/home/SubMenu.test.tsx:      <SubMenu {...props} />
superset-frontend/src/features/home/SubMenu.tsx:const SubMenuComponent: FunctionComponent<SubMenuProps> = props => {
superset-frontend/src/features/home/SubMenu.tsx:              <SubMenu
superset-frontend/src/pages/ActionLog/index.tsx:      <SubMenu name={t('Action Logs')} buttons={subMenuButtons} />
superset-frontend/src/pages/AlertReportList/index.tsx:      <SubMenu
superset-frontend/src/pages/AnnotationLayerList/index.tsx:      <SubMenu name={t('Annotation layers')} buttons={subMenuButtons} />
superset-frontend/src/pages/AnnotationList/index.tsx:      <SubMenu
superset-frontend/src/pages/ChartList/index.tsx:      <SubMenu name={t('Charts')} buttons={subMenuButtons} />
superset-frontend/src/pages/CssTemplateList/index.tsx:      <SubMenu {...menuData} />
superset-frontend/src/pages/DashboardList/index.tsx:      <SubMenu name={t('Dashboards')} buttons={subMenuButtons} />
superset-frontend/src/pages/DatabaseList/index.tsx:      <SubMenu {...menuData} />
superset-frontend/src/pages/DatasetList/index.tsx:      <SubMenu {...menuData} />
superset-frontend/src/pages/ExecutionLogList/index.tsx:      <SubMenu
superset-frontend/src/pages/GroupsList/index.tsx:      <SubMenu name={t('List Groups')} buttons={subMenuButtons} />
superset-frontend/src/pages/Home/index.tsx:        <SubMenu {...menuData} />
superset-frontend/src/pages/QueryHistoryList/index.tsx:      <SubMenu {...menuData} />
superset-frontend/src/pages/RolesList/index.tsx:      <SubMenu name={t('List Roles')} buttons={subMenuButtons} />
superset-frontend/src/pages/RowLevelSecurityList/index.tsx:      <SubMenu name={t('Row Level Security')} buttons={subMenuButtons} />
superset-frontend/src/pages/SavedQueryList/index.tsx:      <SubMenu {...menuData} />
superset-frontend/src/pages/Tags/index.tsx:      <SubMenu name={t('Tags')} buttons={subMenuButtons} />
superset-frontend/src/pages/UserInfo/index.tsx:      <SubMenu buttons={SubMenuButtons} />
superset-frontend/src/pages/UserRegistrations/index.tsx:      <SubMenu name={t('User registrations')} />
superset-frontend/src/pages/UsersList/index.tsx:      <SubMenu name={t('List Users')} buttons={subMenuButtons} />
```

mistercrunch: Personally I'm in favor of refactoring/fixing over adding yet another layer on top.

mistercrunch:
about box-shadow, I'd vote for keeping Superset vanilla antd, and if we decide to change the default style of buttons to not have box-shadow, I'd do it in the default's theme definition, no in the component.

By unsetting `box-shadow` in the Button component here, I think you're preventing anyone for having any shadow while they'd be trying to add a shadow by specifying they want shadow in their theme.

EnxDev:
> about box-shadow, I'd vote for keeping Superset vanilla antd, and if we decide to change the default style of buttons to not have box-shadow, I'd do it in the default's theme definition, no in the component.
> 
> By unsetting `box-shadow` in the Button component here, I think you're preventing anyone for having any shadow while they'd be trying to add a shadow by specifying they want shadow in their theme.

Thanks for your comment, I agree with you that the best approach is to define it within the theme definition. Personally, in this case, I vote to remove the box-shadow from the button because I don't like this inconsistency.

Screenshot 2025-07-07 160046

I'm wondering whether this should apply only to the primary variant.

@kasiazjc  what do you think about that?

superset-frontend/src/pages/ChartList/index.tsx
EnxDev: Aware of the upcoming release, I noticed that the change could introduce regressions. For this reason, I decided to go with a temporary, suboptimal solution.

I agree with you that the icon property can be used, except in cases where a tooltip is involved, given the current structure of the array of objects.

For now, I think it's fine to keep the current approach (an array of config objects with icon key), unless you think it's worth doing a refactor and switching to an array of <Button /> components instead?

EnxDev: **Regarding the icon alignment:** 
In this case, the icon has a font-size of 14px, while the text to its right has a font-size of 12px, which causes a slight misalignment.

By adding display: flex to the <span> that wraps the icon, the icon becomes vertically centered.
In the screenshot below, the button on the right uses display: flex on the icon container span, which has the `.ant-btn-icon class` 

Screenshot 2025-07-07 173848

geido: The whole SubMenu component has a few debatable implementation choices currently. The same indirection is there for `tabs`. This `name` property accepting a node is also weird as `name` does not really feel like the right property. 

I think what we really want as a bare minimum right now is to avoid more fragmented custom styles at the implementation level. I see that inside `SubMenu` the `Button` component accepts an `icon` prop as well. Wouldn't that cover this use case and if we still have a problem then maybe it is something to fix at the `Button` component level instead?

kasiazjc:
> > about box-shadow, I'd vote for keeping Superset vanilla antd, and if we decide to change the default style of buttons to not have box-shadow, I'd do it in the default's theme definition, no in the component.
> > By unsetting `box-shadow` in the Button component here, I think you're preventing anyone for having any shadow while they'd be trying to add a shadow by specifying they want shadow in their theme.
> 
> Thanks for your comment, I agree with you that the best approach is to define it within the theme definition. Personally, in this case, I vote to remove the box-shadow from the button because I don't like this inconsistency.
> 
> Screenshot 2025-07-07 160046
> 
> I'm wondering whether this should apply only to the primary variant.
> 
> @kasiazjc what do you think about that?

I agree with you @EnxDev! And I think it should apply to all for consistency

superset-frontend/src/pages/ChartList/index.tsx
EnxDev: If we want, we can keep the current implementation by simply renaming the name prop (I agree with you — it was a bit confusing at first). When there’s no tooltip, we can use the icon prop to render the icons.

Regarding the icon alignment relative to the text, there's a slight mismatch due to the different line-height. We can leave it as is, unless we want to apply additional styles like display: flex, which could break some vertical alignments. In that case, we could preserve the current font-size and line-height.

If it’s not already supported for the button, I think we could consider introducing props like prependIcon and appendIcon.

superset-frontend/src/pages/AnnotationLayerList/index.tsx
```diff
@@ -288,10 +284,10 @@ function AnnotationLayersList({
     image: 'filter-results.svg',
     buttonAction: () => handleAnnotationLayerEdit(null),
     buttonText: (
-      <>
+      <span className="ant-btn-icon">
```
mistercrunch: Here maybe it would be better to have `EmptyState` receive a full `Button` object, so that the caller can craft the button properly, using the `icon` prop on the button. Seems like the pattern of passing props to a parent component for a children component within it isn't great, seems better to simply pass in the component (?)

msyavuz: That might lead to some code duplication, since the Button would be same everywhere. We can maybe pass another prop like buttonIcon instead?

EnxDev:
> > > about box-shadow, I'd vote for keeping Superset vanilla antd, and if we decide to change the default style of buttons to not have box-shadow, I'd do it in the default's theme definition, no in the component.
> > > By unsetting `box-shadow` in the Button component here, I think you're preventing anyone for having any shadow while they'd be trying to add a shadow by specifying they want shadow in their theme.
> > 
> > 
> > Thanks for your comment, I agree with you that the best approach is to define it within the theme definition. Personally, in this case, I vote to remove the box-shadow from the button because I don't like this inconsistency.
> > Screenshot 2025-07-07 160046
> > I'm wondering whether this should apply only to the primary variant.
> > @kasiazjc what do you think about that?
> 
> I agree with you @EnxDev! And I think it should apply to all for consistency

For now, while we decide which button variants to keep, I'm leaving this change in place.

superset-frontend/src/pages/AnnotationLayerList/index.tsx
EnxDev: I believe the `EmptyState ` component could benefit from a refactor. 

I've noticed that in many cases, a `<span>` with role="button" is passed into the description just to render a button. This doesn't follow our best practices and should be avoided.

In general, the choice between passing a `Button `component directly versus using individual props depends on the level of modularity we want. 

One possible approach could be to replace the current `button` implementation with a prop like `actions`, which would render a dedicated `Actions section` and allow us to insert any content we need. 

Alternatively, we could keep the current structure and simply add props for the icon and its position (e.g., start or end). However, my concern is that we might end up introducing too many props just to replicate the same flexibility we’d get by passing a `Button` component directly.

Whether we opt for flexibility or constraints really depends on `our needs`. 

Personally, I find that passing a complete `Button` component is more readable and powerful, since it gives access to all the available props. 
On the other hand, if our use case is relatively static, a controlled, prop-based would allow us to apply centralized changes without having to modify each usage individually.

That said, I believe the `description` section should definitely remain flexible, as we already have use cases that require custom content. 

The button section, however, could stay more constrained; though I would lean towards introducing an `actions` prop to allow content to be passed more explicitly.

mistercrunch: > I believe the EmptyState  component could benefit from a refactor.

I did some refactoring as part of the theming branch work on that component, but there's still a lot to untangle in that area. It looked like a manifesto on how to misuse react in every way possible.

superset-frontend/packages/superset-ui-core/src/components/Button/index.tsx
```diff
@@ -128,10 +132,13 @@ export function Button(props: ButtonProps) {
         minWidth: cta ? theme.sizeUnit * 36 : undefined,
         minHeight: cta ? theme.sizeUnit * 8 : undefined,
         marginLeft: 0,
+        '& > .ant-btn-icon': {
+          display: 'inline-flex',
```
mistercrunch: still not big on adding any more styles to `Button`, if anything we should be removing styles here. Are you sure this doesn't mess up any of the hundreds of buttons across the app? What if I don't want my button to be inline-flex?

superset-frontend/packages/superset-ui-core/src/components/Button/index.tsx
```diff
@@ -128,10 +132,13 @@ export function Button(props: ButtonProps) {
...
+        },
         '& + .superset-button': {
           marginLeft: theme.sizeUnit * 2,
         },
-        '& > span > :first-of-type': {
+        '& svg': {
```
mistercrunch: any way we can just use `Button.icon` everywhere instead? Does this / could this mess up other buttons?

superset-frontend/src/pages/AnnotationLayerList/index.tsx
mistercrunch: For the sake of moving on, I'd go with @msyavuz and introduce a new `buttonIcon` prop to this component, passing it on to the button instantiation in the component

superset-frontend/src/explore/components/controls/VizTypeControl/constants.tsx
```diff
@@ -33,7 +33,7 @@ export const FEATURED_CHARTS: VizMeta[] = [
   { name: VizType.Table, icon: <Icons.TableOutlined iconSize="l" /> },
   {
     name: VizType.BigNumberTotal,
-    icon: <Icons.BigNumberChartTile />,
+    icon: <Icons.BigNumberChartTile iconSize="l" viewBox="0 0 16 14" />,
```
mistercrunch, line 36: NOTE: not sure how this works exactly, but assuming this icon ends up in an `Button.icon`, I think antd's internal might adjust the icon size it receives based on the button size props. Just pointing out it could be possibly overridden there (?) maybe only if it's not already set (?) Not sure how it all works but curious to clarify or check it works as expected.

EnxDev: The viewBox defines how much virtual space the inner drawing occupies, and it automatically scales the content to fit the visible dimensions (width / height) of the SVG.

In our case, the font-size determines how large the SVG container should be (which sets its width and height).

So, if we keep the default viewBox of "0 0 24 24" and apply a font-size of 16px, we end up with a scale factor of 16 / 24 = 0.666... (about 66.7%), meaning the drawing will not fill 100% of the box.

On the other hand, setting the viewBox to "0 0 16 16" results in 16 / 16 = 1 (100%), so the drawing will scale to fill the entire box.

- 100%
Screenshot 2025-07-17 110713

- 66% 
Screenshot 2025-07-17 112259

superset-frontend/packages/superset-ui-core/src/components/Button/index.tsx
EnxDev: We should use the same font-size for icons and text. See  #52353

https://github.com/apache/superset/pull/34067#discussion_r2194801626

Screenshot 2025-07-17 114029

translated: Remove the 2px font-size difference between the "pure icon button" and the icon + text button, so that the centering and alignment issues can be more completely fixed.

- Icon 14 px font-size 12 px
Screenshot 2025-07-17 114322

- Icon 14 px font-size 14 px

Screenshot 2025-07-17 114349

- Icon: 14px, font-size: 12px, text line-height: 1 instead of 1.5715

Screenshot 2025-07-17 123137

They had several issues, related PRs: 

#51381
#51588
#52132
#52353

In general, it’s not recommended to apply display: inline-flex in this case, we should change the `button line-height`

I was waiting for a response here https://github.com/apache/superset/pull/34067#discussion_r2194801626 regarding this topic

superset-frontend/src/pages/AnnotationLayerList/index.tsx
EnxDev: Ok, if we don't need to specify where the icon should be (at the start or end), I can add only the `buttonIcon` prop and then
I'll do the EmptyState component refactor in a separate PR.

superset-frontend/src/pages/ChartList/index.tsx
EnxDev: >  If it’s not already supported for the button, I think we could consider introducing props like prependIcon and appendIcon.

PS: For completeness, I noticed that the `Ant Design Button component` uses the iconPosition prop for this with start\end values

superset-frontend/packages/superset-ui-core/src/components/Button/index.tsx
EnxDev: Sure, we just needed to decide to adjust the font size to avoid it or change the line-height from `line-height: 1.5715;` to line-height: 0 (for the text)

Ant Design encountered issues when using inline-flex due to the way different fonts handle baselines and how flex containers manage multiline content:
Different fonts have varying baseline alignments. When inline-flex was used for the Button component (initially in version 5.21.6), buttons with custom fonts; such as Mukta; experienced icons that didn't align vertically correctly.

The result: icons looked shifted upward, breaking the visual consistency `(same us cases)`
 
Flexbox containers align items based on flex alignment rules, not text baseline.

With different fonts, the icon (as inline SVG or element) did not match the text baseline, leading to misaligned icons.

**Why inline-flex caused issues?** 

- Flexbox containers align items based on flex alignment rules, not text baseline.
- With different fonts, the icon (as inline SVG or element) did not match the text baseline, leading to misaligned icons.

Additionally, switching from inline-block to inline-flex caused layout regressions for buttons containing multiple lines of text or in tighter layouts:

Previously multilined buttons started being forced onto a single line, resulting in content overflow in grids or table contexts.

This behavior further contributed to UI inconsistencies across different button use cases .

So They reverted display back to inline-block for .ant-btn in version 5.22.1

msyavuz (approved):
LGTM

geido (approved):

PR #34079, opened 2025-07-06
chore: Improve performance to load the chart properties modal

Vitor-Avila:
### SUMMARY

Currently, the chart properties modal fires two requests to `/api/v1/chart/${ID}`: one to load tags and another to load owners. Additionally, the request returns all model columns, including the `json_metadata` value for all dashboards that contain this chart, which can considerably affect performance.

 This PR improves the performance by reducing to one API call + selecting only the columns neeed by the modal.

**Before**
image

**After**
image

### TESTING INSTRUCTIONS
1. Access the Charts list.
2. Open the browser's Developer Tools and access the Network tab.
3. Hover over any chart and click on the pencil icon to open the properties modal.
4. Confirm that a single API call is fired.
5. Confirm the API call includes the `select_columns` argument.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/explore/components/PropertiesModal/PropertiesModal.test.tsx
```diff
@@ -422,3 +419,28 @@ test('"Certification details" should not be empty when saved', async () => {
...
+    typeof isFeatureEnabled
+  >;
+  mockIsFeatureEnabled.mockImplementation(
+    flag => flag === FeatureFlag.TaggingSystem,
+  );
+
+  const props = createProps();
+  renderModal(props);
+
+  await waitFor(() => {
```
geido: Nit. Can we prefer `findBy` for the following ones?

superset-frontend/src/explore/components/PropertiesModal/index.tsx
```diff
@@ -114,6 +123,12 @@ function PropertiesModal({
             label: `${owner.first_name} ${owner.last_name}`,
           })),
         );
+        if (isFeatureEnabled(FeatureFlag.TaggingSystem)) {
+          const customTags = chart.tags?.filter(
+            (tag: TagType) => tag.type === 1,
```
geido, line 128: Makes me wonder what type `1` is? Is this documented in the code somewhere?

geido (approved):

superset-frontend/src/explore/components/PropertiesModal/PropertiesModal.test.tsx
Vitor-Avila: Sure, I'll take a look

superset-frontend/src/explore/components/PropertiesModal/index.tsx
Vitor-Avila: Yup. We have different type of tags -- the type is stored as an enum implemented here: https://github.com/apache/superset/blob/master/superset/tags/models.py#L61-L77

This filter is needed so that the UI only shows custom tags (user-created tags) and not the system tags.

superset-frontend/src/explore/components/PropertiesModal/PropertiesModal.test.tsx
Vitor-Avila: forgot to say that this is done now

PR #34080, opened 2025-07-06
fix: Support metric currency as dict during import

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/33303 and https://github.com/apache/superset/pull/33682. The import schema was still expecting the metric currency as a string.

No UI changes.

### TESTING INSTRUCTIONS
Added test coverage for the import flow.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/datasets/schemas.py
+++ b/superset/datasets/schemas.py
@@ -247,12 +247,15 @@ class ImportMetricCurrencySchema(Schema):
 class ImportV1MetricSchema(Schema):
     # pylint: disable=unused-argument
     @pre_load
-    def fix_extra(self, data: dict[str, Any], **kwargs: Any) -> dict[str, Any]:
+    def fix_fields(self, data: dict[str, Any], **kwargs: Any) -> dict[str, Any]:
         """
-        Fix for extra initially being exported as a string.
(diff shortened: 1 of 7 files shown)


superset/datasets/schemas.py
```diff
@@ -257,7 +262,7 @@ def fix_extra(self, data: dict[str, Any], **kwargs: Any) -> dict[str, Any]:
     expression = fields.String(required=True)
     description = fields.String(allow_none=True)
     d3format = fields.String(allow_none=True)
-    currency = fields.String(allow_none=True, required=False)
+    currency = fields.Nested(ImportMetricCurrencySchema, allow_none=True)
```
betodealmeida, line 268: Can we add `pre_load` logic to handle cases where currency is still a string? Similar to `fix_extra` in line 250 above.

Vitor-Avila: ohh, great catch!

betodealmeida (approved):
Awesome! Thanks for the backwards compat fix!

Issue #32625, opened 2025-03-12
[SIP-156] Proposal for a Rearchitected Apache Superset Documentation Portal


rusackas:
## [SIP-156] Proposal for a Rearchitected Apache Superset Documentation Portal

_[Formerly: [SIP-156] Proposal for a new Apache Superset Developer Portal]_
Title updated to address scope and direction decisions

### Motivation

Apache Superset is evolving to support a extension-based architecture as part of [SIP-151](https://github.com/apache/superset/issues/31932). With this change, there is a need for comprehensive and well-structured documentation tailored to developers building extensions, integrations, and customizations. The current documentation is fragmented across multiple sources, making it difficult for contributors and developers to find relevant information efficiently. It gives no distinction between, or single entry to, documentation created for users, developers, and administrators.

### Proposed Change

This SIP proposes the creation of a **Superset Developer Portal**, which will serve as a single, authoritative source of documentation for the Superset SDK and extension ecosystem. Additionally, it will lay the foundation for an eventual overhaul of Superset’s broader documentation, restructuring content into dedicated sections for Developers, Users, Contributors, and Installation/Deployment.

A key goal of this initiative is to **grow the Superset community** by making it easier for both Superset and Superset extension developers to contribute, and by empowering users and administrators to easily find all the documentation they need to successfully run and use Superset.

This SIP aims to align on the requirements and select the best tooling for implementing a modern, centralized Developer Portal for Apache Superset. After doing some tool comparison (see grid below), the tool we’re proposing is Docusaurus. While by stroke of luck that happens to be the tool we’re using on our current docs site, it also seems to be the only one to meet the broad requirements we have. 

Key requirements of the Superset Developer Portal include:

- Providing well-organized, **text and code-snippet-based documentation**, inspired by the [VS Code extension SDK doc](https://code.visualstudio.com/api)s.
- Serving as a **single repository** for documentation, consolidating:
  - GitHub Wiki
  - Current documentation site
  - Various README files
  - _…and more!_
- Supporting **MDX** (Markdown with JSX) to embed interactive components in documentation, reducing reliance on screenshots.
- Allowing embedding or replacement of **React Storybook**, enabling developers to interact with UI components directly.
- Integration with **automated testing (E2E tests)** to capture and embed static screenshots, ensuring documentation stays up to date.
- Implement **multi-instance versioning** to allow users to reference the correct documentation relative to:
  - Superset core versions
  - Monorepo package versions (e.g., components, themes, SDK, etc.)
[...]

PR #34093, opened 2025-07-07
chore: clean up more flask/jinja html views

mistercrunch:
Grepping through the code base, it appears all these jinja/html files are unused at this time. This should be safe as it is right now.

~~Now I'd love to go deeper and remove all of `superset/templates/appbuilder/` as well, I think it should be safe as long as we don't use any of the FAB-provided CRUD, do we?~~

I also would love to remove all of FAB's `ModelView`-related stuff. In the past that auto-generated API endpoints around CRUD operations, and I'm not sure how to rule out whether we're using any of this anymore. 

I'd say quick-merge this as it should be safe (as unreferenced, superset-specific jinja templates), and I can fast-follow on trying to strip more of the ModelView-related stuff, see if anything breaks. About that, we'll have to rule out that:
- the frontend is using any of the FAB-generated REST endpoints
- we're not piggy-backing or referencing FAB-provided perms
- notify API users that some endpoints are going away (via UPDATING.md, or bumping to `api/v2`)


mistercrunch:
For the record, I tried removing the `.*ModelView` references from `superset/initialization/__init__.py`, and some menu items disappeared. I think there is some entanglement between the FAB configuration and the Menu (navbar). Working on refactoring the main menu during the theming branch work, I remember there was a lot of complications around receiving a deep menu configuration from the backend, trying to see which menu should be shown based on FAB-related settings. I think we should be able to remove a lot of that and build the menu more statically from the frontend eventually.

superset/views/core.py
```diff
@@ -897,8 +896,6 @@ def fetch_datasource_metadata(self) -> FlaskResponse:
     def welcome(self) -> FlaskResponse:
         """Personalized welcome page"""
         if not g.user or not get_user_id():
-            if conf["PUBLIC_ROLE_LIKE"]:
-                return self.render_template("superset/public_welcome.html")
```
geido, line 901: Nit. Just wondering if this could be confusing for those guest users who have been granted some specific access via PUBLIC_ROLE_LIKE and now get redirected to login suddenly. The previous custom welcome page was pretty useless anyways, so maybe this calls for a better experience to support this config option going forward.

geido (approved):

superset/views/core.py
mistercrunch: Oh, yeah, I don't think I thought that through and @sadpandajoe merged. Let's do a follow up PR to sort this out.

qleroy:
I am using `public_welcome.html`
@mistercrunch  Does this also break using `head_custom_extra.html` and `tail_js_custom_extra.html` ?

mistercrunch:
> Does this also break using head_custom_extra.html and tail_js_custom_extra.html ?

I believe so. We could reconsider but the goal is to move away from jinja/html altogether, move towards the backend really just pushing a "bootloader", and handing off controls to the frontend.

Curious to better understand you use cases. Are importing Superset as a big of a flask extension and overriding templates and such, or maybe in your fork you have extra files under `superset/templates/`

Also really unclear as to how many people have used this as an interface to alter Superset in intricate ways.

More about the direction here https://github.com/apache/superset/issues/32625

qleroy:
I override default Superset templates to inject additional JavaScript and CSS—primarily to render custom components from our Design System within Markdown/HTML blocks.

As part of the Helm deployment, I use the following bootstrap script:


```bash
#!/bin/bash

# # Download the DSFR design system
wget https://github.com/GouvernementFR/dsfr/releases/download/v1.14.0/dsfr-v1.14.0.zip
unzip dsfr-v1.14.0.zip -d /app/dsfr

# Download DSFR Charts
wget https://github.com/GouvernementFR/dsfr-chart/releases/download/v2.0.4/dsfr-chart-v2.0.4.zip
unzip dsfr-chart-v2.0.4.zip -d /app/dsfr-chart

# Copy the contents to Superset’s static assets directory
cp -r /app/dsfr/dist /app/superset/static/assets/dsfr
cp -r /app/dsfr-chart/dsfr-chart/dist/DSFRChart /app/superset/static/assets/dsfr-chart

# Clone the templates override repository
git clone https://github.com/etalab-ia/chartsgouv /app/chartsgouv

# Apply custom template overrides
cp /app/superset/templates_overrides/superset/public_welcome.html /app/superset/templates/superset/
cp /app/superset/templates_overrides/head_custom_extra.html /app/superset/templates/head_custom_extra.html
cp /app/superset/templates_overrides/tail_js_custom_extra.html /app/superset/templates/tail_js_custom_extra.html

# Replace error pages with custom versions
cp /app/superset/static/assets/local/404.html /app/superset/static/assets/404.html
cp /app/superset/static/assets/local/500.html /app/superset/static/assets/500.html
```

For example, the `head_custom_extra.html` file includes:

```html
<link
    rel="stylesheet"
    type="text/css"
    href="{{ assets_prefix }}/static/assets/dsfr/dsfr.min.css"
/>
<link
    rel="stylesheet"
    type="text/css"
    href="{{ assets_prefix }}/static/assets/dsfr/utility/utility.min.css"
/>
<link
    rel="stylesheet"
    type="text/css"
    href="{{ assets_prefix }}/static/assets/local/css/tail_css_custom_extra.css"
/>
```

mistercrunch:
Mmmh. Yeah we should assume 6.0 will break things in and around this area. Not sure exactly how to go about it, but it seems we need to offer new ways do extend Superset, or simply tell people to fork in the meantime, ideally with clear/isolated entry points that are easy to manage/maintain.

We should list out all the use cases and try to document recommend ways to achieve such goals. Listing out some things here just to brainstorm, and we can cross check with the extension SIP whether that's handled or not:
- side-loading JS packages (logging/observability/feature-injection)
- alter Menus (adding extra items in the navbar)
- ...

PR #34095, opened 2025-07-08
feat: Don't show the row limit warning for embedded dashboards by default

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/33781.

Embedded dashboards are typically exposed for external audience, and in this case you might not want to indicate to end users that the row limit was reached. 

This PR adds a new `dashboardUiConfig.showRowLimitWarning` config to the embedded SDK. By default, the warning is not visible in embedded but can be enabled through this new config.

**Dashboard loaded in Superset directly**
image

**Same dashboard in embedded mode**
image

### TESTING INSTRUCTIONS
Frontend test added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/SliceHeader/index.tsx
+++ b/superset-frontend/src/dashboard/components/SliceHeader/index.tsx
@@ -34,6 +34,7 @@ import {
   useTheme,
 } from '@superset-ui/core';
 import { useUiConfig } from 'src/components/UiConfigContext';
+import { isEmbedded } from 'src/dashboard/util/isEmbedded';
 import { Tooltip, EditableTitle, Icons } from '@superset-ui/core/components';
 import { useSelector } from 'react-redux';
 import SliceHeaderControls from 'src/dashboard/components/SliceHeaderControls';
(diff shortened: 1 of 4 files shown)


superset-frontend/src/dashboard/components/SliceHeader/index.tsx
```diff
@@ -34,6 +34,7 @@ import {
   useTheme,
 } from '@superset-ui/core';
 import { useUiConfig } from 'src/components/UiConfigContext';
+import { isEmbedded } from 'src/dashboard/util/isEmbedded';
```
Vitor-Avila, line 37: In other files I see `isEmbedded` set to `const isEmbedded = !dashboardInfo?.userId;` -- unsure what's the best way.

amaannawab923: This util is better i guess because this util detects if we are inside an iframe or not

Vitor-Avila: thanks! I thought that could be the case

msyavuz (approved):
LGTM!

PR #34108, opened 2025-07-09
fix: improve login page placement and width

mistercrunch:
Some small adjustments to make the login page look more standard.

- Width set to 400px, should be plenty
- centering the card vertically
- making sure things look decent on mobile using max-width
- autofocus on username

### before
Screenshot 2025-07-08 at 5 18 44 PM

### after
Screenshot 2025-07-08 at 5 13 20 PM
Screenshot 2025-07-08 at 5 13 31 PM
Screenshot 2025-07-08 at 5 13 43 PM

### DIFF
--- a/superset-frontend/src/pages/Login/index.tsx
+++ b/superset-frontend/src/pages/Login/index.tsx
@@ -57,8 +57,10 @@ enum AuthType {
 
 const StyledCard = styled(Card)`
   ${({ theme }) => css`
-    width: 40%;
+    max-width: 400px;
+    width: 100%;
     margin-top: ${theme.marginXL}px;
+    color: ${theme.colorBgContainer};
     background: ${theme.colorBgBase};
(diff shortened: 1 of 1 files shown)


superset-frontend/src/pages/Login/index.tsx
```diff
@@ -57,8 +57,10 @@ enum AuthType {
 
 const StyledCard = styled(Card)`
   ${({ theme }) => css`
-    width: 40%;
+    max-width: 400px;
```
rusackas, line 60: I wonder if we need a few constants in the theme for this... maybe some reusable widths for modals/forms/columns.

rusackas: Actually, another idea would be to move it to the base/root component... rather than styling it here. Or adding a prop for a particular width.

msyavuz (approved):

PR #21645, opened 2022-09-29
feat(charts): allow query mutator to update queries after splitting original sql

solanksh:
feat(charts): allow query mutator to update queries after splitting original sql

### SUMMARY
The changes allow a user to use the SQL_QUERY_MUTATOR function to add statements such as SET ROLE without breaking functionality. Currently, adding a statemen like this works fine in the SQL Editor, but Charts run individual statements by splitting a passed query and throw errors since SET ROLE does not return any data on its own. With the new MUTATE_AFTER_SPLIT variable, users can choose to apply the mutator function after the chart's query execution function splits a given query, allowing users to apply a change to all statements in a query and have Chart functionality not be interrupted.

### TESTING INSTRUCTIONS
In the config file, update the SQL_QUERY_MUTATOR function to add a statement which does not return any results on its own and is something which should be run alongside all other statements in a given query. An example of this is adding a SET ROLE statement to the beginning of a query. Create a new Chart and see if queries return errors due to no data being fetched. Now in the config file, add a MUTATE_AFTER_SPLIT variable and set it equal to True. Create a new Chart and make a query, and verify that the desired results are showing.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [X ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34111, opened 2025-07-09
fix(sqllab): ensure SQL_QUERY_MUTATOR is called when MUTATE_AFTER_SPLIT=True

### LINKED ISSUE #30169: SQL_QUERY_MUTATOR not called in SQLab MUTATE_AFTER_SPLIT = True
### Bug description

`SQL_QUERY_MUTATOR` function is not called for queries in SQLab when `MUTATE_AFTER_SPLIT = True`

Simply not called, but is called when `MUTATE_AFTER_SPLIT = False`

### How to reproduce the bug

config.py:

`MUTATE_AFTER_SPLIT = True

def SQL_QUERY_MUTATOR(  # pylint: disable=invalid-name,unused-argument
    sql: str, **kwargs
) -> str:
    print("Query mutator called") ## will not print during SQLab sessions
    return sql`

Navigate to SQLLab and do some queries.. see nothing printed..

Go to a chart and see the print message.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

LucasWolke:
### SUMMARY

Very simple fix for bug described in: #30169
Currently all queries in sqllab are mutated like this:
`query.executed_sql = database.mutate_sql_based_on_config(block)`

```
def mutate_sql_based_on_config(self, sql_: str, is_split: bool = False) -> str:
        sql_mutator = config["SQL_QUERY_MUTATOR"]
        if sql_mutator and (is_split == config["MUTATE_AFTER_SPLIT"]):
            return sql_mutator(
                sql_,
                security_manager=security_manager,
                database=self,
            )
        return sql_
```

This means `is_split` is always `False` -> We always use the sql_mutator when `MUTATE_AFTER_SPLIT` is False, and never when it's True.

The fix passes the value of `MUTATE_AFTER_SPLIT` to the `mutate_sql_based_on_config` function, so `is_split == config["MUTATE_AFTER_SPLIT"]` always evaluates to True.

This does not break current implementations and fix the bug.

### TESTING INSTRUCTIONS
In config.py:

```
MUTATE_AFTER_SPLIT = True


def SQL_QUERY_MUTATOR(  # pylint: disable=invalid-name,unused-argument
    sql: str, **kwargs
) -> str:
    print("Query mutator called") ## will now print during SQLab sessions
    return sql
```
Run query in sql lab -> Now prints message when `MUTATE_AFTER_SPLIT` = True or False

### ADDITIONAL INFORMATION
Fixes #30169
- [x] Has associated issue: #30169
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_lab.py
+++ b/superset/sql_lab.py
@@ -431,7 +431,7 @@ def execute_sql_statements(  # noqa: C901
     for statement in parsed_script.statements:
         apply_limit(query, statement)
 
-    # some databases (like BigQuery and Kusto) do not persist state across mmultiple
+    # some databases (like BigQuery and Kusto) do not persist state across multiple
     # statements if they're run separately (especially when using `NullPool`), so we run
     # the query as a single block.
     if db_engine_spec.run_multiple_statements_as_one:
@@ -475,7 +475,7 @@ def execute_sql_statements(  # noqa: C901
             db.session.commit()
 
             # Hook to allow environment-specific mutation (usually comments) to the SQL
-            query.executed_sql = database.mutate_sql_based_on_config(block)
+            query.executed_sql = database.mutate_sql_based_on_config(block, is_split=config["MUTATE_AFTER_SPLIT"])
 
             try:
                 result_set = execute_query(query, cursor, log_params)
--- a/superset/sql_lab.py
+++ b/superset/sql_lab.py
@@ -475,7 +475,9 @@ def execute_sql_statements(  # noqa: C901
             db.session.commit()
 
             # Hook to allow environment-specific mutation (usually comments) to the SQL
-            query.executed_sql = database.mutate_sql_based_on_config(block, is_split=config["MUTATE_AFTER_SPLIT"])
+            query.executed_sql = database.mutate_sql_based_on_config(
+                block, is_split=config["MUTATE_AFTER_SPLIT"]
+            )
 
             try:
                 result_set = execute_query(query, cursor, log_params)


superset/sql_lab.py
```diff
@@ -475,7 +475,9 @@ def execute_sql_statements(  # noqa: C901
             db.session.commit()
 
             # Hook to allow environment-specific mutation (usually comments) to the SQL
-            query.executed_sql = database.mutate_sql_based_on_config(block)
+            query.executed_sql = database.mutate_sql_based_on_config(
+                block, is_split=config["MUTATE_AFTER_SPLIT"]
```
mistercrunch, line 479: Not sure if I understand the logic in place, but shouldn't the `is_split` argument here be related to whether the sql-to-be-mutated is currently split (as in based on the context of where the function is called), and not be the config flag that means "I want to mutate before/after the split"?

mistercrunch: From my understanding the code calling `mutate_sql_based_on_config` should call it twice, once before it's split with `is_split=False` and once after it's split with  `is_split=True`

mistercrunch: If that's the case, it might be worth changing the function definition to *not* set a default and, forcing all calls to `mutate_sql_based_on_config` to be explicit at every call.

mistercrunch: The whole thing seems brittle and a bit of a hack. Ideally we'd be consistent here, say always call before split and make it easy for mutators to apply a split if they want, but that would break backwards compatibility ...

LucasWolke: Thanks for your assessment. To clarify my approach: the `MUTATE_AFTER_SPLIT` flag was introduced specifically to control when the query mutator is applied for chart queries, as discussed in #21645. This behavior was already established in SQL Lab, so the flag was not originally intended to affect SQL Lab queries.
From my understanding, the `is_split` parameter is also primarily meant for chart queries, to ensure backward compatibility.
It would be possible to create a separate `mutate_sql` function that always applies the mutator, just for SQL Lab (if semantic clarity is the issue with my fix), or to add a comment clarifying why the config flag is passed. However, I believe always applying the mutator after splitting the query in SQL Lab is the intended and correct approach - the "before or after splitting" logic was only ever meant for chart queries, not for SQL Lab.

mistercrunch: What I'm thinking is that `is_split` on the line right above these comments should probably hard coded to True or False, depending on whether the SQL at that time is split or not. 

Looking at where it executes -> https://github.com/apache/superset/blob/5a32777dd090c840bafbec458bba91f6b36e1dba/superset/models/core.py#L646 , it only applies the logic if the parameter matches the config. Meaning the argument is supposed to represent whether the SQL is split or not at the time of the call.

If merging the code as is, that method would always execute since `config["MUTATE_AFTER_SPLIT"] == config["MUTATE_AFTER_SPLIT"]`

LucasWolke: Yes, the whole point of my change was that (in sql lab) the method (that mutates the sql) should always be executed.
The Problem is that the is_split parameter was introduced to solve an issue regarding chart queries - and the sql lab is now broken accidentally.

LucasWolke: The sql lab shouldn't even be using the `mutate_sql_based_on_config` method, it always wants to mutate the sql.

I think the confusion comes from a refactor you did [here](https://github.com/apache/superset/commit/cfc440c56cb973bd3202ab8d7daabbbde888704a#diff-a0173518300cdfa4526b24e65fb23e7b77acc76a9b4c5983d581759028b4dbdbL267-R266).

``` diff
-    sql = SQL_QUERY_MUTATOR(
-        sql,
-        security_manager=security_manager,
-        database=database,
-    )
+    sql = database.mutate_sql_based_on_config(sql)
```

The sql lab was refactored to use `mutate_sql_based_on_config`, but that method was only written for a fix in chart queries (#21645) . In the sql lab we don't care about `is_split`.

I propose that we either create a new `mutate_sql` function that always mutates or just use the mutator like before the refactor.

Again, we don't want to use the before or after split logic, that was never intended for the sql lab. Hope this makes more sense now.

mistercrunch: Right. I remember some of the context now. Generally we're trying to bring the SQL-execution code paths back together, and there's more work to be done here. 

I think `SQL_QUERY_MUTATOR` is a super useful config hook, but seems it has grown a bit complex. It's tempting to clarify/redesign, but we have to be cautious around backwards compatibility. Dilemma is probably around change management here.

Let's try to understand the ideal scenario here, and see if we can preserve backwards compatibility.

First, let's start with trying to itemize common use cases for `SQL_QUERY_MUTATOR`:
- #1 is clearly logging, where we add comments to all the SQL executed with information about the query provenance, and the user who triggered it. This allows database admins to get more visibility into who's behind the service account. Many organizations use this to parse user/tool/context information and use it for analytics
- some security/data-access-policy stuff, maybe you have complex data access policy where you want to intercept some things, or double check some rules, and fail queries if something that shouldn't be allowed is submitted, simply `raise` an exception from here
- some redirection / rewiring, maybe redirecting to a similar dataset for things like aggregate awareness, ...
- ?

Thinking about this method and keeping it backwards compatible, one thing we could do is passing in more context information, like a `source` parameter that would inform whether this SQL comes from SQL Lab, Explore, alerts, populating filters' list of values ... Adding net new args to the function should be backwards compatible as long as user catch/disregard *args **kwargs.

About the `is_split` flag, looking at the code, it looks like in explore we call `mutate_sql_based_on_config` both before and after the SQL is broken into chunks for execution. Seems this can be confusing for people authoring `SQL_QUERY_MUTATOR`, where you have to assume something like:
- in explore, this method will be called twice, once with is_split=False first, and once more for each statement with is_split=True
- in SQL Lab, will be called once for the block (could probably be called twice too following the same pattern as for explore)
- throw in some session/env settings, stuff like calling `SET` and things like that. Kind of a hack but who knows what people might be using this for ...

Knowing all this, it's unclear to me what to do, especially given the backwards-compatibility concerns.

Best "mostly backwards"-compatible way forward: add a `source` args for clarity, call twice for SQL Lab for consistency (?) Add some notes in UPGRADING.md about the changes to notify upgraders.

Breaking backwards compatibility? Flag `SQL_QUERY_MUTATOR` for deprecation and redesign the hook(s), maybe something like:
```
SQL_BLOCK_MUTATOR(source)  # always called even if block has a single statement
SQL_STATEMENT_MUTATOR(source) # called for each sub-statement
```

Curious on what @betodealmeida or others might thing is best here. This could be slated along with overdue work around bringing all SQL execution codepaths together, and fix/improve other related issues around code and logic duplication across explore and SQL Lab.

greggailly: We are currently looking into this as our setup requires to use this hook for permissions related issues. Our database connection is setup with a user which by default has no permissions and settings role dynamically at each request. We are trying to achieve this for both charts and sqllab so making it my consistent would definitely be great !
@mistercrunch redesigning the hook does seem like a good idea for long term.
In the meantime is merging this PR considered ou maybe splitting the hook into a specific `SQLLAB_QUERY_MUTATOR`(but means both codepaths continue to diverge for now...)

giacomochiarella: Looking for the same solution here. We want to use one single user database connection configured in Superset and set role before any query is executed, either from sqllab or dataset/chart/dashboard. The role needs to be determined based on the logged in user is executing the query

giacomochiarella: Here my 2 cents on what I found out in Superset 5.0.0.
I’m trying to make MUTATE_AFTER_SPLIT and SQL_QUERY_MUTATOR working. I want to execute set role <rolename> when a query is executed, either in sqllab or from a dashboard/chart/dataset. I’ve setup SQL_QUERY_MUTATOR function in order to add set role to the query:
```
def sql_query_mutator(sql: str, **kwargs: dict) -> str:
    if <some condition>:
        rle_name = <some logic>
        sql = f"set role {role_name}; {sql}"
    logger.info(f"Executing query {sql}...")
    return sql
SQL_QUERY_MUTATOR = sql_query_mutator
```
so far so good, things get weird when I set MUTATE_AFTER_SPLIT:
MUTATE_AFTER_SPLIT = True: everything works fine from sqllab. The role is applied correctly. I get Data error: Error: no results to fetch in charts (I read it is due to set role <> not returning results makes psycopg raising that error)

- MUTATE_AFTER_SPLIT = False: it works from charts and sqllab with a big BUT. In such case, the user needs to add set role <role_name> which means the user can set any role if it knows the role name

- MUTATE_AFTER_SPLIT = False would work perfectly if I could force users in sqllab to not use other roles, for example if I could intercept the sqllab query I could remove any set role and add automatically the set role needed.

The idea is to have one single connection based on a user without any permissions but membership to roles and then add set role automatically based on the username of Superset user which is executing the query. I don't think this is possible currently.

greggailly: @giacomochiarella we are currently manually "forcing" this PR when building the docker image so that sqllab works properly with the query mutator.

We edited the Dockerfile like so:

```
# Edit way sqllab mutates SQL waiting for https://github.com/apache/superset/pull/34111
--
RUN sed -i 's\|sql = database\.mutate_sql_based_on_config(sql)\|sql = database.mutate_sql_based_on_config(\
sql, is_split=config["MUTATE_AFTER_SPLIT"]\
)\|' /app/superset/sql_lab.py
```

giacomochiarella: @greggailly thanks for sharing! I have Superset deployed with docker too. In this way you could change queries coming from charts/datasets and sqllab?
And have you added MUTATE_AFTER_SPLIT = False in superset_config.py? Or True?
Have you noticed any regression?
I think I can add that command in docker-init.sh, isn't it? I'm using the docker image already available apachesuperset.docker.scarf.sh/apache/superset

giacomochiarella: I've just noticed that command will not work if you upgrade, the line changed to 
`query.executed_sql = database.mutate_sql_based_on_config(block)`

greggailly: @giacomochiarella yes this is based on apache/superset:4.1.1. We have MUTATE_AFTER_SPLIT = True as we need it for dashboards and charts.
Yes as long as the PR is not merged this change can break at any time when upgrading. We will like into the new line when we will upgrade.
@mistercrunch do you think a temporary solution could be merged to keep a consistent use of the DB_QUERY_MUTATOR for the moment ?

greggailly: @giacomochiarella depending on your use case you could also use the DB_CONNECTION_MUTATOR. In postgres you can set the role when opening the connection with the -c option (manually changing the connection string): `query_dict.update({"options": "-c role=" + role})`.
We had to move away from this solution as the connection needs to be persistent and this isn't our case with pgbouncer in front of our server.

rusackas:
Haven't seen any motion here in a bit. Thoughts, @mistercrunch ?

giacomochiarella:
is there any news on this issue?

greggailly:
@giacomochiarella or anyone who builds their own image the the sed command should now be:
```
# Edit way sqllab mutates SQL waiting for https://github.com/apache/superset/pull/34111
RUN sed -i 's|query\.executed_sql = database\.mutate_sql_based_on_config(block)|query\.executed_sql = database.mutate_sql_based_on_config(\
    block, is_split=app\.config["MUTATE_AFTER_SPLIT"]\
)|' /app/superset/sql_lab.py
```

Btw we still highly rely on this fix as our SQL_MUTATOR should always be run even in sqllab.

giacomochiarella:
@greggailly thank you! Although it is nice, I don't want to build my own image. Doing such a fix means also maintaining it every time the code changes. If there are huge changes this would be a big issue and it may cause also be stuck on a Superset version.
Honestly, using such a feature would be also a big risk in the case this feature is removed in the future (it would mean change the whole Superset permission setup). Although it would be nice to have a sql editor meanwhile I'm building a dashboard, I find it better to have Superset for just dashboards and having something else to execute queries

rusackas:
Trying to figure out what to do with this one. My robots basically say:

The fix works operationally (mutator now fires) but is semantically wrong, passing `is_split=config["MUTATE_AFTER_SPLIT"]` making the guard always True. 

@mistercrunch identified the deeper design: `is_split` was meant to convey whether SQL is already-split, so the correct fix is to bypass `mutate_sql_based_on_config` entirely in SQLLab and call `SQL_QUERY_MUTATOR` directly.

Behavioral risk: `MUTATE_AFTER_SPLIT=False` users may expect one pre-split mutation but now get per-block calls, an observable difference for multi-statement queries.

Zero new tests. Both MUTATE_AFTER_SPLIT=True and False paths should be covered.

rusackas (changes requested):
See above comment for details, but 

(1) Call SQL_QUERY_MUTATOR directly in SQLLab, bypassing mutate_sql_based_on_config.
(2) Add tests for both MUTATE_AFTER_SPLIT=True and False in SQLLab path.

...or tell me why I'm wrong, and have your agent argue with my agent ;)

giacomochiarella:
@rusackas In short, the idea is to have one database connection in Superset and inject "set role <role_based_on_usename>" before any sql

def sql_query_mutator(sql: str, **kwargs: dict) -> str:
    if kwargs.get("database"):
        logger.info(f"Setting {role_name} role for database connection {database_connection_name}...")
        sql = f"set role {role_name}; {sql}"
    logger.info(f"Executing query {sql}...")
    return sql

SQL_QUERY_MUTATOR = sql_query_mutator
MUTATE_AFTER_SPLIT = True

in any connection to the database. The issue today is that, when MUTATE_AFTER_SPLIT = True then the SQL_QUERY_MUTATOR function is called only on queries executed through dashboards/charts (when you open it) and not through sqllab. In sqllab the user must add itself "set role <>" meaning if it knows another role name it can use it and access to other data. I don't know if this is intentional or not (if not it would be a bug then).
In my opinion, either for consistency and to allow a better data governance, it would be good the function SQL_QUERY_MUTATOR is applied "globally" when MUTATE_AFTER_SPLIT = True. This is not mentioned in the documentation (or at least I have not found it).
Or if you have another documented way is fine as well. The mutator is the closest way I found in the documentation.

rusackas:
Thanks for digging into this one, @LucasWolke - the diagnosis was spot on, and sorry it sat so long. I opened #41127 to carry it across the line: same root cause, but passing `is_split=not run_multiple_statements_as_one` so it stays correct for `MUTATE_AFTER_SPLIT=False` folks too, plus the same fix in the new Celery executor and tests for both paths. Went with a fresh PR rather than pushing to your branch mostly so I could fold in @giacomochiarella's use case and the extra coverage without trampling your history... you're credited as co-author there. That one supersedes this, so it'll close once #41127 lands. Thanks again for sticking with it! :)

rusackas:
PR reviews more than welcome on https://github.com/apache/superset/pull/41127 by the way :D

rusackas:
Closing this in favor of #41127 (to focus attention on that one), which folds your diagnosis in (you're credited as co-author there) plus the correct `is_split` semantics and coverage for both `MUTATE_AFTER_SPLIT` states. Thanks again for chasing this down, @LucasWolke!

rusackas:
https://github.com/apache/superset/pull/41127 merged, closing this.

PR #34118, opened 2025-07-10
fix: frontend translation framework crashes on string errors

mistercrunch:
2 fixes here:
- one was around full on app crashes when a msgid contains a parameter that's not provided on the frontend (is happening on `master` on japanese right now)
- some strings labeled as fuzzy by babel were not making their way to the frontend. I don't fully understand what gets some strings labeled as fuzzy and others not, but without the `--fuzzy` flag the language pack generated by po2json was missing some strings in french
- ~~removed `translateWithNumbers()` and `tn()` functions that are not used anywhere except for unit tests.~~

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/translation/Translator.ts
+++ b/superset-frontend/packages/superset-ui-core/src/translation/Translator.ts
@@ -97,20 +97,17 @@ export default class Translator {
 
   translateWithNumber(key: string, ...args: unknown[]): string {
     try {
-      if (typeof args[0] === 'number') {
-        const [plural, ...rest] = args;
+      const [plural, num, ...rest] = args;
+      if (typeof plural === 'number') {
         return this.i18n
           .translate(key)
           .ifPlural(plural, key)
-          .fetch(plural, ...rest);
+          .fetch(plural, num, ...rest);
       }
-      if (typeof args[1] === 'number') {
-        const [_, plural, ...rest] = args;
-        return this.i18n
-          .translate(key)
-          .ifPlural(plural, key)
-          .fetch(...rest);
-      }
+      return this.i18n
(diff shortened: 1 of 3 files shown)


superset-frontend/packages/superset-ui-core/src/translation/Translator.ts
```diff
@@ -86,20 +86,35 @@ export default class Translator {
...
   translateWithNumber(key: string, ...args: unknown[]): string {
-    const [plural, num, ...rest] = args;
-    if (typeof plural === 'number') {
+    try {
+      const [plural, num, ...rest] = args;
+      if (typeof plural === 'number') {
+        return this.i18n
+          .translate(key)
+          .ifPlural(plural, key)
+          .fetch(plural, num, ...args);
```
rusackas: Should that be `.fetch(plural, num, ...rest)`?

mistercrunch: oh right, good catch, though it looks like it was wrong before too (!?)

Grepping through the repo, it appears `translateWithNumber` is only tied to `tn('...')` calls, which are not used anywhere except in a few unit tests ...

Not sure what to do:
- delete unused code and references to it?
- fix/improve the function and write unit tests for it?
- [not for this PR], start using `tn()` in more places? Guessing it's meant to be used in strings like "You saved {num} dashboards(s)"

mistercrunch: I think i'm going to roll with #1 . I'm guessing it's ok to rely on hard coded pure t like `{num} dashboard(s) were saved.` instead of hoping the gettext pluralization to work (?)

rusackas (approved):
LGTM!

PR #34119, opened 2025-07-10
feat(i18n): load language pack asynchronously

mistercrunch:
Since forever we've been doing this dubious thing of passing the language pack through the dom. The server fetches the json file, and serializes it into a dom node as part of the "bootstrap data". Frontend picks it up deserializes it and hooks things up.

In this PR:
- creating a super simple flask endpoint to serve the json files safely (made sure to only serve the right files...)
- changing the "preamble" to run an async call to fetch the language pack before doing anything

### DIFF
--- a/superset-frontend/src/preamble.ts
+++ b/superset-frontend/src/preamble.ts
@@ -24,6 +24,7 @@ import {
   makeApi,
   initFeatureFlags,
   SupersetClient,
+  LanguagePack,
 } from '@superset-ui/core';
 import setupClient from './setup/setupClient';
 import setupColors from './setup/setupColors';
@@ -50,11 +51,11 @@ setupClient({ appRoot: applicationRoot() });
   const lang = bootstrapData.common.locale || 'en';
   if (lang !== 'en') {
     try {
-      const { json: languagePack } = await SupersetClient.get({
+      // Second call to configure to set the language pack
+      const { json } = await SupersetClient.get({
         endpoint: `/superset/language_pack/${lang}/`,
       });
-      // Second call to configure to set the language pack
-      configure({ languagePack });
+      configure({ languagePack: json as LanguagePack });
       dayjs.locale(lang);
     } catch (err) {
       console.warn(
--- a/superset-frontend/src/preamble.ts
+++ b/superset-frontend/src/preamble.ts
@@ -1,25 +1,31 @@
 /**
  * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
+ * or more contributor license agreements. See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


superset-frontend/src/preamble.ts
```diff
@@ -19,65 +19,81 @@
...
+  LanguagePack,
+} from '@superset-ui/core';
 import setupClient from './setup/setupClient';
 import setupColors from './setup/setupColors';
 import setupFormatters from './setup/setupFormatters';
 import setupDashboardComponents from './setup/setupDashboardComponents';
 import { User } from './types/bootstrapTypes';
 import getBootstrapData, { applicationRoot } from './utils/getBootstrapData';
 
+configure();
```
msyavuz, line 36: What is this for? Is there a loading happening between this and the language pack loading?

msyavuz (approved):
LGTM!

superset-frontend/src/preamble.ts
mistercrunch: I'm unclear what exactly it does, but it prepares a bunch of things around the various "registries" and such. It logs a bunch of things to the console if called too late. Might be worth digging a bit deeper. I'll double check there's no important perf costs in calling it twice.

superset/views/core.py
```diff
@@ -891,6 +903,24 @@ def fetch_datasource_metadata(self) -> FlaskResponse:
         datasource.raise_for_access()
         return json_success(json.dumps(sanitize_datasource_data(datasource.data)))
 
+    @event_logger.log_this
+    @has_access
+    @expose("/language_pack/<lang>/")
```
michael-s-molina, line 908: @mistercrunch It would be nice to put this endpoint under `/api/v1` to be an official endpoint.

mistercrunch: yeah I thought through this wondering if this is a private or public endpoint, and thought ideally should be public, but given the partitioning of libraries in the frontend, and the fact that our i18n framework is monolithic currently (no segmentation or context based on the provenance of the strings), it felt like it would require significantly more work to do this properly.

Not sure whether the extension framework SIP mentions `i18n`, but it may be a bit of a blindspot. 

Without going to deep into it, it seems that if we want for extensions to be localized/localizable, we'd need some way to segment the language pack based on package provenance (backend/frontend, and break down per library on the frontend side). The whole i18n stack might need to offer different ways to fetch and apply translations.

Happy to help think through this, but it's a fairly significant amount of work.

Note that one minor and valuable thing I considered while doing the work here was to find a way to add metadata to translation to identify what is frontend/backend, and make sure the endpoint only returns the strings intended for the frontend ... The gettext-based/flask-babel framework we use does offer ways to do such things, but I encountered some limitations, and decided to punt on these.

This PR here addresses something pretty important - removing the language pack from bootstrap data - but there's much more to do in this area.

PR #34123, opened 2025-07-10
fix(deckgl): Revert "feat(deckgl): add support for OpenStreetMap as our new default and make "tile-providers" more configurable"

DamianPendrak:
Reverts apache/superset#33603
As mentioned in #34017 :
It looks like the PR broke all deck.gl Charts, at least on the local environment. Updating the deck.gl packages break the Charts alone, but it seems like the implemented features from the PR require the packages to be updated.

I couldn't find a fix, but maybe you can @plavacquery?

Screenshot 2025-07-09 at 15 56 44


mistercrunch:
As I mentioned, let's try and push forward if we can. Seems there's another issue around attribution we need to handle as well.

plavacquery:
Hi @DamianPendrak working on it to find out what's wong

DamianPendrak:
@plavacquery I had to revert the feature #33603 on my PR #34017 to allow for QA, and I didn't manage to bring it back before merging to master. Sorry about that! Let me know if I can help

plavacquery:
I have not found a solution yet. It seems that deckgl version changes have broken my feature cause of luma librairy.

Nonetheless I tried to go back to last working deckgl version and I still have the problem, have you any idea or which changes can provok that  ?

mistercrunch:
Not sure what the best place to hold this conversation is, but if indeed the 2 PRs are mutually exclusive, as a maintainer I have to vouch for https://github.com/apache/superset/pull/33603 as it adds a great feature and bumps the `deck-gl` libs forward.

If this is related to colors - I have to admit I know nothing about the core issue - but I wonder if you could use tinycolor2 that's already a lib in the repo (?)

PR #34126, opened 2025-07-10
fix: database model Collapse state 

SBIN2010:
### SUMMARY

The problem is that the Collapse component from the Ant Design library retains its state (such as an open panel) even after closing and reopening the modal window. This is because React does not destroy the component when the modal window is closed, but simply hides it. As a result, the Collapse state is preserved.

To solve this problem, you need to reset the Collapse state every time you open the modal window, which is what this PR solves

continuation of PR #32734

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


msyavuz (approved):
LGTM!

PR #32679, opened 2025-03-14
fix: ensure datasource permission in explore

hxtmdev:
### SUMMARY

Problem: The `/explore` command does not validate if datasource access is permitted.

Source: https://github.com/apache/superset/blob/15cf06699aa05719697f34a16e4e9989dfdc8af1/superset/commands/explore/get.py#L123

The intention seems to be to `raise_for_access` but instead the boolean result is simply discarded.

Result: Metadata about inaccessible datasources is disclosed, including names, can be enumerated through
`/explore/?datasource_type=table&datasource_id=COUNTER`

### TESTING INSTRUCTIONS
open `/explore/?datasource_type=table&datasource_id=X` with(out) permissions and observe datasource name

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34129, opened 2025-07-11
fix(explore): Display missing dataset for denied access

justinpark:
### SUMMARY
Due to patch #32679  in Superset version 5.0, there is an issue where previously, for datasets restricted by permissions, the dataset name would be displayed but the results were not allowed. With the recent changes, these are now shown as “missing dataset,” which leads to the system recognizing that there are no existing charts, even when they actually exist.

|Before (4.1)|After (5.0)|
|--|--|
|Screenshot_2025-07-10_at_4_34_40 PM|Screenshot 2025-07-10 at 3 41 00 PM|

This commit addresses this issue by adding handling for the SupersetSecurityException caused by #32679. Additionally, it improves the UI by displaying both the dataset name and the error message, making it easier to identify the cause when this issue occurs.

After:

Screenshot 2025-07-10 at 3 40 50 PM

### TESTING INSTRUCTIONS

Create a dataset using a table permission doesn't exist
Click a explore link to check the error message

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


justinpark:
cc: @hxtmdev

hxtmdev:
> displaying both the dataset name and the error message

Doesn't this disclose part of the (meta) data that is supposed to be forbidden again? I guess the real question is whether not being able to "access" a datasource should imply not being told any names / table names. The existing code's intent (and my expectation) seems to be that the desired result is not showing them (unless I overlooked them being in the response in some form anyway).

justinpark:
> Doesn't this disclose part of the (meta) data that is supposed to be forbidden again? I guess the real question is whether not being able to "access" a datasource should imply not being told any names / table names. The existing code's intent (and my expectation) seems to be that the desired result is not showing them (unless I overlooked them being in the response in some form anyway).

@hxtmdev The user does not lack permission to access the chart itself, but rather lacks permission to access the data source used in the chart. For example, when a user clicks on a chart from the chart list to view it, but does not have access to the dataset used by the chart, showing a message like "missing dataset" can be misleading. (Since it says missing dataset, it is understood that it was marked as missing and therefore deleted. But it was actually a permission denied issue.)
In other words, the user is unable to access information from the data source used in the chart (such as column information or table records), but being unable to even see the name of the dataset used in the chart makes it difficult to identify which source is causing the permission issue. In fact, within Airbnb, there are a significant number of inquiries about why a chart is not displaying properly and what the root cause is.
To address this issue, it would be very helpful for users if the actual error message encountered when accessing the data source was displayed together with the current data source name. This would greatly assist users in troubleshooting the problem.

michael-s-molina (approved):

michael-s-molina (changes requested):
@justinpark Could we keep the chart title?

justinpark:
> justinpark Could we keep the chart title?

Updated PR to display chart information by fetching the chart api on fallback in explore metadata api.

Screenshot 2025-07-14 at 2 15 41 PM

michael-s-molina (approved):
Thank you!

PR #34144, opened 2025-07-11
feat(theming): Introduce bootstrap-driven Superset theme configurations

gabotorresruiz:
### SUMMARY

This change adds full support for Superset’s bootstrap-provided theme objects (default, dark, and settings) and integrates them into the existing ThemeController API.

### TESTING INSTRUCTIONS

1. Pass a valid `Theme` dict to `THEME_DEFAULT` and or `THEME_DARK` (you can generate one in the Ant Design theme editor: https://ant.design/theme-editor)
2. Reload the frontend app
3. Watch the magic happens 

For `THEME_SETTINGS`:
1. Pass a valid `ThemeSettings` dict
E.g.:
```
THEME_SETTINGS = {
    "enforced": True,            # If True, forces the default theme and ignores user preferences
    "allowSwitching": True,      # Allow user to switch between themes (default and dark)
    "allowOSPreference": True,   # Allow the app to Auto-detect and set system theme preference
}
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
@@ -19,7 +19,11 @@
 /* eslint-disable react-prefer-function-component/react-prefer-function-component */
 // eslint-disable-next-line no-restricted-syntax
 import React from 'react';
-import { theme as antdThemeImport, ConfigProvider } from 'antd';
+import {
+  theme as antdThemeImport,
+  ConfigProvider,
+  type ThemeConfig,
+} from 'antd';
 import tinycolor from 'tinycolor2';
 
 // @fontsource/* v5.1+ doesn't play nice with eslint-import plugin v2.31+
@@ -160,7 +164,7 @@ export class Theme {
    * Dark mode should be specified via the algorithm property in the config
    */
   setConfig(config: AnyThemeConfig): void {
-    const antdConfig = normalizeThemeConfig(config);
+    const antdConfig: ThemeConfig = normalizeThemeConfig(config);
 
     // Apply default tokens to token property
     antdConfig.token = {
@@ -169,7 +173,7 @@ export class Theme {
     };
 
     // First phase: Let Ant Design compute the tokens
-    const tokens = Theme.getFilteredAntdTheme(antdConfig);
+    const tokens: SharedAntdTokens = Theme.getFilteredAntdTheme(antdConfig);
 
     // Set the base theme properties
     this.antdConfig = antdConfig;
--- a/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
@@ -19,11 +19,7 @@
 /* eslint-disable react-prefer-function-component/react-prefer-function-component */
 // eslint-disable-next-line no-restricted-syntax
 import React from 'react';
-import {
-  theme as antdThemeImport,
-  ConfigProvider,
-  type ThemeConfig,
-} from 'antd';
+import { theme as antdThemeImport, ConfigProvider } from 'antd';
 import tinycolor from 'tinycolor2';
 
 // @fontsource/* v5.1+ doesn't play nice with eslint-import plugin v2.31+
@@ -164,7 +160,7 @@ export class Theme {
    * Dark mode should be specified via the algorithm property in the config
    */
   setConfig(config: AnyThemeConfig): void {
-    const antdConfig: ThemeConfig = normalizeThemeConfig(config);
+    const antdConfig = normalizeThemeConfig(config);
 
     // Apply default tokens to token property
     antdConfig.token = {
@@ -173,7 +169,7 @@ export class Theme {
     };
 
     // First phase: Let Ant Design compute the tokens
-    const tokens: SharedAntdTokens = Theme.getFilteredAntdTheme(antdConfig);
+    const tokens = Theme.getFilteredAntdTheme(antdConfig);
 
     // Set the base theme properties
     this.antdConfig = antdConfig;
--- a/superset-frontend/packages/superset-ui-core/src/theme/types.ts
+++ b/superset-frontend/packages/superset-ui-core/src/theme/types.ts
@@ -34,17 +34,32 @@ export type AntdTokens = ReturnType<typeof antdThemeImport.getDesignToken>;
 export type AntdThemeConfig = ThemeConfig;
 
 /**
+ * A combination of theme modes that can be used in theme config.
+ * This is used to define which algorithms are allow to be applied together.
+ */
+export type ThemeAlgorithmCombination = (
+  | ThemeMode.DEFAULT
+  | ThemeMode.DARK
+  | ThemeMode.COMPACT
+)[];
+
+/**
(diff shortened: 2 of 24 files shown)


superset-frontend/packages/superset-ui-core/src/theme/types.ts
```diff
@@ -33,18 +33,33 @@ import { Theme } from '.';
 export type AntdTokens = ReturnType<typeof antdThemeImport.getDesignToken>;
 export type AntdThemeConfig = ThemeConfig;
 
+/**
+ * A combination of theme modes that can be used in theme config.
+ * This is used to define which algorithms are allow to be applied together.
+ */
+export type ThemeAlgorithmCombination = (
```
mistercrunch: I think there's a drier way to write all this, maybe something like:
```ts
export enum ThemeMode {
  DEFAULT = 'default',
  DARK = 'dark',
  COMPACT = 'compact',
}
export type ThemeAlgorithmCombination = ThemeMode[];
export type ThemeAlgorithmOption = ThemeMode | ThemeAlgorithmCombination;
```

mistercrunch: conceptually we're kind of mixing some types that aren't the same, might want to decouple `Algorithms` as supported by antd ('default', 'dark', 'compact') from our `ThemeDarknessSetting` ('default', 'dark', 'system')

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
```diff
@@ -160,7 +164,7 @@ export class Theme {
    * Dark mode should be specified via the algorithm property in the config
    */
   setConfig(config: AnyThemeConfig): void {
-    const antdConfig = normalizeThemeConfig(config);
+    const antdConfig: ThemeConfig = normalizeThemeConfig(config);
```
msyavuz: We don't have a standard for this, but i personally prefer implicit types in places like this where the function's return is already typed

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
```diff
@@ -169,7 +173,7 @@ export class Theme {
     };
 
     // First phase: Let Ant Design compute the tokens
-    const tokens = Theme.getFilteredAntdTheme(antdConfig);
+    const tokens: SharedAntdTokens = Theme.getFilteredAntdTheme(antdConfig);
```
msyavuz: Same discussion here

superset-frontend/packages/superset-ui-core/src/theme/utils.ts
```diff
@@ -38,9 +40,8 @@ export function isSerializableConfig(
 
   if (algorithm === undefined) return true;
 
-  if (Array.isArray(algorithm)) {
+  if (Array.isArray(algorithm))
```
msyavuz, line 42: This is another personal preference but i think multiline if statements are better for these since refactoring and debugging is a bit easier.

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
gabotorresruiz: I think it makes sense

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
gabotorresruiz: Yep, I think it makes sense

superset-frontend/packages/superset-ui-core/src/theme/utils.ts
gabotorresruiz: I understand your point, but having a `single-line if statement` at the top and then a `multiline if statement` for just one condition feels inconsistent.

superset-frontend/packages/superset-ui-core/src/theme/types.ts
gabotorresruiz: Done!

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
gabotorresruiz: Done!

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
gabotorresruiz: Done!

superset-frontend/src/theme/ThemeController.tsx
```diff
@@ -45,202 +81,618 @@ export class ThemeController {
...
+
+    return !enforced && allowSwitching;
+  }
+
+  /**
+   * Normalizes the theme configuration to ensure it has a valid algorithm.
+   * @param theme - The theme configuration to normalize
+   * @returns An object with normalized mode and algorithm.
+   */
+  private normalizeTheme(theme: AnyThemeConfig): {
```
mistercrunch: Isn't this duplicated logic from what's in `utils.ts`?  https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-core/src/theme/utils.ts#L111-L116

Seems ThemeController should probably only deal with `AnyThemeConfig` and assume that the Theme class will always handle things properly or error out if there's a bad theme object (?)

superset-frontend/src/theme/ThemeController.tsx
```diff
@@ -45,202 +81,618 @@ export class ThemeController {
...
+        [...inputSet].every(item => validCombination.has(item)),
     );
   }
 
+  /**
+   * Checks if the algorithm is a valid combination or a simple one.
+   * @param algorithm - The theme mode or combination to convert
+   * @returns A valid ThemeAlgorithm or ThemeAlgorithm[]
+   */
+  private getValidAlgorithm(
```
mistercrunch: Similar to above comment, this looks like dup logic that probably either already exists or belongs directly in Theme or utils

superset/config.py
```diff
@@ -677,18 +678,58 @@ class D3TimeFormat(TypedDict, total=False):
 # This is merely a default
 EXTRA_CATEGORICAL_COLOR_SCHEMES: list[dict[str, Any]] = []
 
-# THEME is used for setting a custom theme to Superset, it follows the ant design
-# theme structure
-# You can use the AntDesign theme editor to generate a theme structure
-# https://ant.design/theme-editor
+# ---------------------------------------------------
```
mistercrunch, line 681: seems we may need to touch up the related comment in `UPDATING.md`. I think I added a comment in there, but we may want to clarify further what people will need to do. The comment that's there might be ok already.

mistercrunch (commented):
Minor comments but overall looks rock solid.

superset-frontend/src/theme/ThemeController.tsx
gabotorresruiz: You're right, let me fix this

superset-frontend/src/theme/ThemeController.tsx
gabotorresruiz: Totally agree, I'll fix this ASAP

mistercrunch (approved):
LGTM!

Issue #34120, opened 2025-07-10
Superset package included twice in v5.0.0 lean image


schoel-bis:
### Bug description

The superset package is included twice in the lean image, once at `/app/superset` and once at `/app/.venv/lib/python3.10/site-packages/superset`. They are almost identical, except that the former contains a lot of `__pycache__` entries that the latter does not have, and the latter does not contain any translation files at all:

```sh
docker run --rm --entrypoint=sh apache/superset:5.0.0 -c 'diff -r /app/.venv/lib/python3.10/site-packages/superset /app/superset'
```

I tried to run an image where the former directory is removed, and that seems to work fine:

```Dockerfile
FROM apache/superset:5.0.0

RUN rm -rf /app/superset
```

Note that the extra directory is ~75MB large, so I would say it is not exactly insignificant.

The cause appears to be in [lines 247ff. of the Dockerfile](https://github.com/apache/superset/blob/17c1a37afb3686f11414cdd7b98d9f238179e204/Dockerfile#L247), where the local superset module is instelled into the venv but then left where it is. Also note that in line 252 the python files inside `/app/superset` are all compiled, but that this happens after installation into the venv, so that should not have any practical effect on execution.

### Superset version

5.0.0

### Python version

3.10

### Node version

Not applicable

### Browser

Not applicable

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [x] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
I usually CC @mistercrunch on these docker image issues, so here I go again ;)

mistercrunch:
Good catch, we should definitely investigate and fix. Would love to get more people deep in there as it's hard to find cycles [and energy] to giving this the attention it deserves.

mistercrunch:
Proposing a solution here -> https://github.com/apache/superset/pull/34146

Has tradeoffs, but overall addresses the issue, is simple, has few/minimal downsides and some good pros attached.

rusackas:
Closed by #34146.

PR #34146, opened 2025-07-12
chore(docker): use editable mode in docker images

mistercrunch:
Trying to address the issue raise here: https://github.com/apache/superset/issues/34120#issuecomment-3058902853

From my analysis, it should address the issue (package installed twice in docker) by using a symlink.

The downsides are pretty minimal, with a micro os-level penalty on resolving symlinks (super minimal from my understanding), and some considerations around image size, where in editable mode we can't/couldn't strip .py files if we wanted.

This seems like a positive step.

Other alternatives would probably require 2 extra layers (an extra build layer for both dev and lean), and it's hard to keep the docker logic DRY and readable in that context. Personally I'm not willing to take this on now... With that approach, we wouldn't have `/app/superset` and would only have superset/ in `.venv`, could compile and strip `.py` to only keep `.pyc` if we wanted there. 

Doing some research, it seems python is still very much lacking around managing lean packages/bundles of bytecode only, where wheels and packages in general tend to be raw .py, and it's tricky to only pacakge .pyc if/when desired. 

---
Some benefits from the `-e` (editable mode). `/app/superset` in docker:
- easy to find, easy to alter (pros/cons), easy to alter/inject files in extra docker layers
- easily docker-mountable, mount /app/superset to your local repo and you're live-running/editing the image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -223,7 +223,7 @@ RUN --mount=type=cache,target=${SUPERSET_HOME}/.cache/uv \
     /app/docker/pip-install.sh --requires-build-essential -r requirements/base.txt
 # Install the superset package
 RUN --mount=type=cache,target=${SUPERSET_HOME}/.cache/uv \
-    uv pip install .
+    uv pip install -e .
 RUN python -m compileall /app/superset
 
 USER superset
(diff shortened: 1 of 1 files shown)


mistercrunch:
@dpgaspar what do you think of this one!?

eschutho (approved):
Looks fine for non-production use.

Issue #23264, opened 2023-03-02
Snowflake connector requires passphrase


mtucker502:
The snowflake connector requires a passphrase.

#### How to reproduce the bug

1. Configure new database connection with the URI `snowflake://{user}:{password}@{account}.{region}/{database}`
2. Add security extra:
```json
{
    "auth_method": "keypair",
    "auth_params": {
        "privatekey_path":"/certs/my_cert.p8"
    }
}
```

3. Click Finish

### Expected results

The connection should work successfully.

### Actual results

Error `An error occurred while creating databases: 'privatekey_pass'` appears.

### Environment

(please complete the following information):

- Firefox 109.0.1 (64-bit)
- superset version: `Superset 0.0.0-dev` (
- python version: `Python 3.8.16`
- any feature flags active:

```Dockerfile
FROM apache/superset:latest

USER root

RUN pip install snowflake-sqlalchemy

USER superset
```

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [✅] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [✅] I have reproduced the issue with at least the latest released version of superset.
- [✅] I have checked the issue tracker for the same issue and I haven't found one similar.

### Additional context

Here are other possible configurations and their errors:

```json
{
    "auth_method": "keypair",
    "auth_params": {
        "privatekey_path":"/certs/my_cert.p8",
        "privatekey_pass": ""
    }
}
```

Error: `An error occurred while creating databases: Password was given but private key is not encrypted.`

```json
{
    "auth_method": "keypair",
    "auth_params": {
        "privatekey_path":"/certs/my_cert.p8",
        "privatekey_pass": null
    }
}
```

Error: `ERROR: 'NoneType' object has no attribute 'encode'`

### Solution(s)

Add logic to accept private key as is if no passphrase is provided.

tonofll:
Hi, is there any update on this or a workaround to use a private key without password?

mvdbosch:
Any update on this issue?

rusackas:
Tempted to close this as stale... and I swear this has been reported elsewhere, but giving a ping to @villebro as well, who might have run into this situation.

emattson:
I hit this same issue this morning

rusackas:
Can anyone validate if this is still an issue we're concerned with, and if it's in 4.1.1/5.0.0?

junyoneyama:
[...]

PR #34156, opened 2025-07-14
feat(snowflake): Support Snowflake private keys w/o passphrase

junyoneyama:
### SUMMARY

Accept Snowflake private keys with no passphrase.

Previously private keys with no passphrase didn't work as described in #23264.

### TESTING INSTRUCTIONS

1. Follow instructions in <https://superset.apache.org/docs/configuration/databases/#snowflake>
  to create a Snowflake database connection.
1. Put `snowflake://{user}@{account}.{region}/{database}?role={role}&warehouse={warehouse}` in
  the SQLAlchemy URI field. Note that no `password` is specified.
1. Put the following JSON in `ADVANCED` > `Security` > `SECURE EXTRA` field.

    ```json
    {
      "auth_method": "keypair",
      "auth_params": {
        "privatekey_body": "-----BEGIN PRIVATE KEY-----\n**********\n-----END PRIVATE KEY-----\n"
      }
    }
    ```

1. Click `Test Connection` and verify that the connection is successful.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #23264
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/snowflake.py
+++ b/superset/db_engine_specs/snowflake.py
@@ -394,9 +394,11 @@ class SnowflakeEngineSpec(PostgresBaseEngineSpec):
             else:
                 with open(auth_params["privatekey_path"], "rb") as key_temp:
                     key = key_temp.read()
+            privatekey_pass = auth_params.get("privatekey_pass", None)
+            password = privatekey_pass.encode() if privatekey_pass is not None else None
             p_key = serialization.load_pem_private_key(
                 key,
-                password=auth_params["privatekey_pass"].encode(),
+                password=password,
                 backend=default_backend(),
             )
             pkb = p_key.private_bytes(


superset/db_engine_specs/snowflake.py
```diff
@@ -394,9 +394,11 @@ def update_params_from_encrypted_extra(
             else:
                 with open(auth_params["privatekey_path"], "rb") as key_temp:
                     key = key_temp.read()
+            privatekey_pass = auth_params.get("privatekey_pass", None)
+            password = privatekey_pass.encode() if privatekey_pass is not None else None
```
mistercrunch, line 398: seems reasonable but tagging @betodealmeida who might have context, and @xiayanzheng (author of https://github.com/apache/superset/pull/21322)

rusackas: Added @villebro as well, who's familiar with the ins and outs of Snowflake auth

mistercrunch: My intuition looking at this if that we may be able to put this whole block (from 398 to 409 inclusively) behind an if statement (if privatekey_pass: {...}), but it could be that snowflake expects the `connect_args["private_key"]` block with an empty password (?) Might be worth trying whether this work, if it does it makes the code more clear ...

junyoneyama: Here `password` is not sent to Snowflake but used to decrypt the private key if it is encrypted. `connect_args["private_key"]` is still necessary even when the private key is unencrypted.

See also: "Generate the private key" section in [Key-pair authentication and key-pair rotation | Snowflake Documentation](https://docs.snowflake.com/en/user-guide/key-pair-auth#generate-the-private-key)

> You can generate either an encrypted version of the private key or an unencrypted version of the private key.

betodealmeida (approved):
Awesome, thanks for this PR!

rusackas:
Thank you for the contribution!

PR #34160, opened 2025-07-14
chore: asked Claude Code to migrate `src/utils` from js -> ts

mistercrunch:
See title, Claude Code seem to be doing a decent job at migrating js->ts. This is a bit of PoC but could really help complete the ts migration.


superset-frontend/src/SqlLab/components/SaveQuery/index.tsx
```diff
@@ -215,7 +214,15 @@ const SaveQuery = ({
         onHide={() => setShowSaveDatasetModal(false)}
         buttonTextOnSave={t('Save & Explore')}
         buttonTextOnOverwrite={t('Overwrite & Explore')}
-        datasource={getDatasourceAsSaveableDataset(query)}
+        datasource={{
```
sadpandajoe, line 217: why did we remove `getDatasourceAsSaveableDataset` when we updated the definition below to be the same? Shouldn't we just keep it?

mistercrunch: seems wrong ... Claude!

PR #20281, opened 2022-06-06
feat: Visualize SqlLab.Query model data in Explore 📈

hughhhh:
### SUMMARY

Explore view now can accept `datasource_type = 'query`, which represents `sql_lab.Query` model. Now users will be able to easily chart data from queries being executed in SQL Lab

https://user-images.githubusercontent.com/27827808/173101938-4c31864d-0c0b-4fb1-aa5a-6bfdb25123bc.mov

### TESTING INSTRUCTIONS

1. Go to Sqllab and run a query
2. Click `Create Chart`
3. Verify that Chart Source is `Query` with the SQL icon

***OR
1. Go to Sqllab and run a query
2. Then run a `select id from query`
3. Take one valid `id` from results returned
4. Enter the following string into the browser `{ephemeral-url}/superset/explore/query/{id}`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34177, opened 2025-07-15
feat: removing dup logic in sqla/models.py and models/helpers.py

mistercrunch:
Huge props to Claude Code for running this whole analysis and fixing everything here.

  Remove duplicate code between SqlaTable and ExploreMixin classes

  This commit eliminates 7 duplicate methods totaling ~146 lines of code that were
  introduced by PR #20281 in July 2022. The duplication occurred when @hughhhh 
  created the `ExploreMixin` class to enable Query objects to be visualized in Explore,
  but instead of properly refactoring shared functionality, he copy-pasted methods
  from SqlaTable into the new mixin.

  ### Root Cause Analysis
  The code duplication was traced to commit e5e886739460c011a885a13b873665410045a19c
  where ExploreMixin was created with methods copied from SqlaTable:
  - 5 methods were exact duplicates with identical implementations
  - 2 methods had diverged over time with different error handling and security practices

  ### Method Resolution Order Impact
  SqlaTable inherits from (Model, BaseDatasource, ExploreMixin), meaning:
  - SqlaTable's implementations override ExploreMixin's versions
  - Query class (inheriting only from ExploreMixin) uses the ExploreMixin versions
  - Both sets of methods were being used, creating genuine duplication

  ### Methods Removed from SqlaTable
  **Exact duplicates:**
  - `filter_values_handler` (60 lines) - Static method for handling filter values
  - `_apply_cte` (12 lines) - Static method for appending CTEs to SQL
  - `make_orderby_compatible` (28 lines) - Instance method for ORDER BY compatibility
  - `text` (2 lines) - Instance method for creating TextClause objects
  - `get_query_str` (4 lines) - Instance method for generating query strings

  **Divergent methods consolidated:**
  - `_normalize_prequery_result_type` (35 lines) - Improved to use ExploreMixin's
    more robust version with better error handling for missing columns
  - `changed_by_name` (5 lines) - Improved to use AuditMixinNullable's more secure
    version with HTML escaping instead of plain string conversion

  ### Technical Improvements
  Beyond eliminating duplication, this refactoring improves the codebase by:
  - Using safer `.get()` method instead of direct dict access for column lookups
  - Implementing proper HTML escaping for user-facing data (security improvement)
  - Maintaining single source of truth for shared functionality between SqlaTable and Query

  ### Testing
  All existing unit tests pass (8/8) and critical integration tests for the
  normalize_prequery_result_type functionality have been verified. The inheritance
  hierarchy ensures SqlaTable now properly inherits the more robust implementations
  from ExploreMixin and AuditMixinNullable.

  Fixes: Code duplication introduced in PR #20281 (feat: Visualize SqlLab.Query model data in Explore)

### DIFF
--- a/superset/connectors/sqla/models.py
+++ b/superset/connectors/sqla/models.py
@@ -26,8 +26,6 @@ from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from typing import Any, Callable, cast, Optional, Union
 
-import dateutil.parser
-import numpy as np
 import pandas as pd
 import sqlalchemy as sa
 from flask_appbuilder import Model
@@ -1168,12 +1166,6 @@ class SqlaTable(
         return self.database.db_engine_spec
 
     @property
-    def changed_by_name(self) -> str:
-        if not self.changed_by:
-            return ""
-        return str(self.changed_by)
-
-    @property
     def connection(self) -> str:
         return str(self.database)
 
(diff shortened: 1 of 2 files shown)


superset/connectors/sqla/models.py
```diff
@@ -2053,6 +1894,14 @@ def load_database(self: SqlaTable) -> None:
             session = inspect(self).session  # pylint: disable=disallowed-name
             self.database = session.query(Database).filter_by(id=self.database_id).one()
 
+    def get_query_str(self, query_obj: QueryObjectDict) -> str:
+        """Returns a query as a string using ExploreMixin implementation"""
+        return ExploreMixin.get_query_str(self, query_obj)
+
+    def text(self, clause: str) -> TextClause:
+        """Returns a text clause using ExploreMixin implementation"""
+        return ExploreMixin.text(self, clause)
```
mistercrunch, line 1903: @betodealmeida these two methods are a bit of a hack, where from my understanding the inheritance scheme is non linear in some specific cases, so we need to cherry pick some methods to skip a level of inheritance. 

PR still achieves the goal of de-duping logic, but doesn't fully fix the inheritance scheme and related patchwork ...

superset/connectors/sqla/models.py
```diff
@@ -1643,43 +1521,6 @@ def _get_series_orderby(
             )
         return ob
 
-    def _normalize_prequery_result_type(
```
betodealmeida, line 1646: The parent method seems more complete, handling cases where `column` is not a dict. This is one of the major problems with code duplication, things get fixed in one place but not another. :(

betodealmeida (approved):

superset/connectors/sqla/models.py
mistercrunch: Confirmed on slack, this comment vouches for deleting this method as opposed to the other one that's slightly different.

PR #34178, opened 2025-07-15
fix: Console errors from various sources

msyavuz:
### SUMMARY

After a few big migrations and theming. Dev console errors increased. I aim to fix whatever i can in this pr.

### TESTING INSTRUCTIONS

Check browser console

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Thanks for tackling this! Console has become unruly, makes it hard to use without active logging/filtering.

mistercrunch:
Great to see the menu refactor happening! Gave it a shot during the theming branch effort and ended up bailing as it was pretty tangled up.

geido (approved):

superset-frontend/src/features/reports/ReportModal/HeaderReportDropdown/index.tsx
```diff
@@ -85,63 +60,59 @@ const DropdownItemExtension = extensionsRegistry.get(
...
+
+  // Select the reports state and specific report with proper reactivity
+  const report = useSelector<any, AlertObject | null>(state => {
+    if (!resourceId) return null;
+    // Select directly from the reports state to ensure reactivity
+    const reportsState = state.reports || {};
+    const resourceTypeReports = reportsState[resourceType] || {};
+    const reportData = resourceTypeReports[resourceId];
+
+    // Debug logging to understand what's happening
```
mistercrunch, line 90: @msyavuz I think this slipped its way into `master`

mistercrunch: The thing that's really weird to me is that the console message appeared in Explore ... Seems this component seems completely unrelated to explore (?) Yet I see a lot of messages, meaning a lot of re-rendering is happening. I'm surprised to see this. Worth an investigation as perf is probably generally affected (?) Also seeing other weird stuff in the Network tab but that's another story and not related to this PR ...

msyavuz: Not sure what’s going on here. @geido I believe you worked on this part. Would you have time to look into it in your follow-up PR?
If not, I can investigate it separately, just let me know what works best for you.

mistercrunch: Also wondering ... linter should have caught this (?) maybe this rule isn't set properly?

msyavuz: Linter catches it but i think its a warning so if claude code doesn't setup todos properly its easy to miss these. Rerenders are more interesting to me, would be nice to improve perf by accidentaly discovering it

mistercrunch: About the linter, we should set to error and fix the ones around the app. I believe we have a legit console logger someplace in the app that we can use and shuts itself off in prod but logs in dev.

PR #34180, opened 2025-07-15
fix(databricks): string escaper

betodealmeida:
### SUMMARY

The Databricks SQLAlchemy dialect we use has a bug when escaping single quotes: it doubles the single quote, instead of using a backslash. For example, a value of `O'Hara` will be rendered in the SQL as `O''Hara`, which is correct for most databases, but for Databricks should be `O\'Hara`.

Upgrading to the latest `databricks-sqlalchemy` package fixes the issue, but requires SQLAlchemy 2.x. The 1.x line of `databricks-sqlalchemy` has the bug as well. Note, we're using the unnoficial `sqlalchemy-databricks` package instead, so we should consider changing to the 1.x version of the official one.

For now, this PR solves the problem by monkeypatching a custom string escaper into the the dialect.

Before:

Screenshot 2025-07-15 at 6 16 23 PM

After:

Screenshot 2025-07-15 at 6 15 11 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/databricks.py
+++ b/superset/db_engine_specs/databricks.py
@@ -17,13 +17,14 @@
 from __future__ import annotations
 
 from datetime import datetime
-from typing import Any, TYPE_CHECKING, TypedDict, Union
+from typing import Any, Callable, TYPE_CHECKING, TypedDict, Union
 
 from apispec import APISpec
 from apispec.ext.marshmallow import MarshmallowPlugin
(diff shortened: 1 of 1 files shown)


Vitor-Avila (approved):
Thanks for porting the fix!

PR #34182, opened 2025-07-16
feat: add a theme CRUD page to manage themes

mistercrunch:
## Summary

  This PR introduces a comprehensive theme management system for Superset with complete CRUD functionality and fixes critical theme scoping issues through a major architecture refactor.

  ## Key Features

  ### 🎨 **Theme CRUD Management**
  - Complete REST API with full CRUD operations for themes
  - Interactive theme editor with live preview and apply-on-demand
  - Import/Export functionality for theme backup and distribution
  - Permission-based access with proper RBAC integration
  - System themes - Protected themes defined in configuration

  ### 📊 **Dashboard Theme Integration**
  - Per-dashboard theme assignment through enhanced dashboard edit modal
  - **FIXED**: Dashboard themes now properly scoped to dashboard content only
  - Theme selector integrated into "Theme & CSS" dashboard submenu
  - Seamless theme switching with automatic cleanup on navigation

  ### 🏗️  **Architecture Refactor** *(Major New Work)*
  **Problem**: Dashboard themes were affecting entire app (including navbar) instead of just dashboard content.

  **Solution**: Complete theme system overhaul:
  - **ThemeController**: Singleton managing all Theme object lifecycles with isolated dashboard themes
  - **Nested Providers**: Global ThemeProvider + Dashboard-specific ThemeProvider for proper isolation
  - **Component Modernization**: Eliminated global `themeObject` dependencies, components now use context-aware `useTheme()` + utilities
  - **Deduplicated Logic**: Removed duplicate Theme class methods, single source of truth in `themeUtils.ts`

  ### 🔧 **System Integration**
  - UUID-based theme references in configuration
  - System theme seeding with automatic startup creation/updates
  - Feature flag update - `THEME_ENABLE_DARK_THEME_SWITCH` defaults to True

  ## Fixed Behaviors

  - ✅ Dashboard themes **only** affect dashboard content (not navbar)
  - ✅ Navbar **always** uses global/system theme
  - ✅ Theme mode picker **only** affects navbar/global theme
  - ✅ Navigation from themed dashboard restores global theme everywhere

  ## Technical Implementation

  **Backend**: Theme model, complete REST API, dashboard integration, system theme seeding
  **Frontend**: CRUD interface, dashboard integration, **proper theme scoping architecture**
  **Component Library**: Refactored to use context-aware theming without global dependencies
  **Testing**: Comprehensive test coverage including new `themeUtils.test.ts`

  ## Migration & Breaking Changes

  - Database migration required for themes table and dashboard `theme_id` column
  - **No breaking changes** for end users
  - **Developer API changes**: Use `useTheme()` + utilities instead of global `themeObject` methods

  ## Configuration Examples

  ```python
  # Traditional approach
  THEME_DEFAULT = {"algorithm": "default", "token": {"colorPrimary": "#2893B3"}}

  # UUID reference to existing theme
  THEME_DEFAULT = {"uuid": "550e8400-e29b-41d4-a716-446655440000"}
```
  This implementation provides enterprise-grade theme management with proper isolation, security, and a solid foundation for advanced theming capabilities.

  🤖 Co-generated with https://claude.ai/code

### CODE REVIEW

File: docs/docs/configuration/theming.mdx [hunk 1]
```diff
@@ -38,7 +38,6 @@ Set the feature flag in your `superset_config`
 ```python
 DEFAULT_FEATURE_FLAGS: dict[str, bool] = {
     {{ ... }}
-    THEME_ALLOW_THEME_EDITOR_BETA = True,
```
mistercrunch, line 41: No need for this feature flag now as we have an "apply" button in the theme editor in the new CRUD

superset/migrations/versions/2025-07-15_16-40_3fd555e76e3d_theme_crud.py
```diff
@@ -0,0 +1,65 @@
...
+from sqlalchemy_utils import UUIDType
+
+# revision identifiers, used by Alembic.
+revision = "3fd555e76e3d"
+down_revision = "363a9b1e8992"
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
sadpandajoe: I think we're supposed to use the helper functions for creating tables and columns going forward

mistercrunch: makes sense, let me fix and see if I can add a linting rule (?) Though would love to spend some cycles improving migrations:
- compact migration before say 3.0 (wondering where to draw the line)
- making sure we use utils consistently
- add linting rules to prevent not using utils

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -236,6 +246,76 @@ export class ThemeController {
...
+
+    this.devThemeOverride = theme;
+    this.storage.setItem(
+      STORAGE_KEYS.DEV_THEME_OVERRIDE,
+      JSON.stringify(theme),
+    );
+
+    const normalizedTheme = this.normalizeTheme(theme);
+    this.updateTheme(normalizedTheme);
+  }
```
msyavuz, line 388: This persists the theme? This is what's called when clicked on apply.

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -236,6 +246,76 @@ export class ThemeController {
...
+        }
+      } catch (error) {
+        console.error('Failed to load CRUD theme:', error);
+        this.fallbackToDefaultMode();
+      }
+    } else {
+      this.storage.removeItem(STORAGE_KEYS.CRUD_THEME_ID);
+      this.resetTheme();
+    }
+  }
```
msyavuz: I don't think this is used anywhere. All of the crud related stuff in this file seems unused, we should probably implement it.

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -43,6 +43,8 @@ const DEFAULT_THEME_SETTINGS = {
 
 const STORAGE_KEYS = {
   THEME_MODE: 'superset-theme-mode',
```
msyavuz, line 45: Is this also irrelevant now? It stays as default in local storage

superset-frontend/src/dashboard/containers/DashboardPage.tsx
```diff
@@ -215,6 +222,34 @@ export const DashboardPage: FC<PageProps> = ({ idOrSlug }: PageProps) => {
...
+  // Apply dashboard theme when dashboard loads
+  useEffect(() => {
+    const applyDashboardTheme = async () => {
+      if (dashboard?.theme?.id) {
+        try {
+          const response = await SupersetClient.get({
+            endpoint: `/api/v1/theme/${dashboard.theme.id}`,
+          });
+          const themeConfig = JSON.parse(response.json.result.json_data);
+          themeContext.setTemporaryTheme(themeConfig);
```
msyavuz: This doesn't differentiate between dashboard themes and appwide themes. Setting a theme from dashboard changes the whole app, was this intended?

superset-frontend/src/dashboard/containers/DashboardPage.tsx
```diff
@@ -215,6 +222,34 @@ export const DashboardPage: FC<PageProps> = ({ idOrSlug }: PageProps) => {
     return () => {};
   }, [css]);
 
+  // Apply dashboard theme when dashboard loads
+  useEffect(() => {
+    const applyDashboardTheme = async () => {
+      if (dashboard?.theme?.id) {
```
geido: When do we clean dead references of deleted themes in the dashboard metadata?

superset-frontend/src/features/themes/ThemeModal.tsx
```diff
@@ -0,0 +1,369 @@
...
+}) => {
+  const supersetTheme = useTheme();
+  const { setTemporaryTheme, clearLocalOverrides, hasDevOverride } =
+    useThemeContext();
+  const [disableSave, setDisableSave] = useState<boolean>(true);
+  const [currentTheme, setCurrentTheme] = useState<ThemeObject | null>(null);
+  const [isHidden, setIsHidden] = useState<boolean>(true);
+  const isEditMode = theme !== null;
+  const user = getBootstrapData()?.user;
+  const canDevelopThemes = user
```
geido: Should we use the permission tied to the `theme` POST/PUT endpoint instead so that non-admins can also develop themes when granted such permissions or is there a different logic behind this choice?

msyavuz (commented):
Right now it looks like we only use backend for storing the Theme config and name. It would be cool to add some kind of tagging system for light/dark to enable quick mode switching for users.

superset-frontend/src/theme/ThemeController.ts
mistercrunch: Right, it applies/persist for the current user using local storage, use can "clear local settings" to fallback on the system default + light/dark preferences

superset-frontend/src/theme/ThemeController.ts
mistercrunch: Yes, this should represent whether the user wants light/dark/os-settings for themselves, I made the related feature flag on by default for discoverability.

superset-frontend/src/dashboard/containers/DashboardPage.tsx
mistercrunch: yes we had a product conversation about this. I think we decided to scope it to only the dashboard areas (navbar shouldn't change). I see to change things here so it only injects the dashboard theme to that section of the page ... TODO

superset-frontend/src/dashboard/containers/DashboardPage.tsx
mistercrunch: Good catch. For now I may just prevent deletion for themes attached to dashboards or dissociate on theme delete, gotta decide what to do .... TODO

superset-frontend/src/theme/ThemeController.ts
mistercrunch: That's right, though we might need it in the future (?) might be attached to user prefs in the future. Can keep for now.

superset-frontend/src/features/themes/ThemeModal.tsx
mistercrunch: ok I fixed that!

superset-frontend/src/dashboard/containers/DashboardPage.tsx
mistercrunch: done!

.pre-commit-config.yaml
```diff
@@ -97,9 +97,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.9.7
     hooks:
+      - id: ruff-format
```
mistercrunch, line 92: bycatch: format before check is better(!)

.gitignore
```diff
@@ -131,3 +131,4 @@ superset/static/stats/statistics.html
 # LLM-related
 CLAUDE.local.md
 .aider*
+CLAUDE.local.md
```
msyavuz:
```suggestion
```
seems like a duplicate

superset-frontend/src/components/ListView/Filters/Search.tsx
```diff
@@ -43,9 +43,7 @@ interface SearchHeaderProps extends BaseFilter {
   toolTipDescription: string | undefined;
 }
 
-const StyledInput = styled(Input)`
-  border-radius: ${({ theme }) => theme.borderRadius}px;
-`;
+// Using vanilla antd Input to avoid styled component circular dependency issue
```
msyavuz:
```suggestion
```

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -406,6 +574,17 @@ export class ThemeController {
...
   private getThemeForMode(mode: ThemeMode): AnyThemeConfig | null {
+    // Priority 1: Dev theme override (highest priority for development)
+    // Dev overrides affect all contexts
+    if (this.devThemeOverride) {
+      return this.devThemeOverride;
+    }
+
+    // Priority 2: CRUD theme (organizational themes)
+    // CRUD themes should NEVER affect the mode-based theme selection
+    // Dashboard themes are handled separately via createDashboardThemeProvider
```
msyavuz: Missing implementation

msyavuz (commented):
From testing:
- When user's apply a theme through the CRUD view, theme mode select doesn't work anymore. Should it?
- Dashboard specific filters are clashing with the app-wide filter 
image

docs/docs/configuration/theming.mdx
```diff
@@ -10,44 +10,92 @@ version: 1
...
+    "allowSwitching": True,     # Allow users to switch between themes
+    "allowOSPreference": True,  # Auto-detect system theme preference
 }
 ```
 
-Restart Superset to apply changes
+### Referencing CRUD Themes
 
-## 3 — Tweak live in the app (beta)
+You can also reference themes created via the CRUD interface by UUID:
```
msyavuz: This also doesn't feel like an ideal workflow to do from config. It means users have to run superset, get the uuid, rerun superset. Maybe we can do this in the backend while the app is running with a SYSTEM tag or something like that?

mistercrunch: What I was thinking here is say you tell a designer: "ok go build an awesome theme for Superset", then the designer goes in and creates a great theme, now how do you set it as your default? I guess you could copy paste the theme definition into the config file manually ...

The whole idea was to allow for a config to reference a non-system-defined theme directly. But if copy/paste JSON blob is fine we can remove some complexity here...

I could remove all that uuid logic/stuff here if we don't think it's helpful. Maybe it's more confusing than helpful overall.

mistercrunch: ok, did more thinking and decided to rip that uuid stuff out, thanks for the feedback here, I was overthinking things.

superset-frontend/packages/superset-ui-core/src/components/ThemeSelect/index.tsx
```diff
@@ -43,36 +50,65 @@ const ThemeSelect: React.FC<ThemeSelectProps> = ({
     [ThemeAlgorithm.COMPACT]: <Icons.CompressOutlined />,
   };
 
+  // Use different icon when local theme is active
+  const triggerIcon = hasLocalOverride ? (
+    <Icons.FormatPainterOutlined style={{ color: theme.colors.error.base }} />
```
EnxDev: Would it be better to use the `iconColor` prop with `theme.colorError` in this case?

superset-frontend/packages/superset-ui-core/src/theme/utils/themeUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+  };
+}
+
+/**
+ * Check if the current theme is dark mode based on background color
+ * @param theme - Theme tokens from useTheme()
+ * @returns true if theme is dark, false if light
+ */
+export function isThemeDark(theme: SupersetTheme): boolean {
+  return tinycolor(theme.colorBgContainer).isDark();
```
EnxDev, line 112: I think it could be a good idea to also consider `colorBgElevated` in addition to `colorBgContainer`

**Note:** There's already a `useIsDark` hook defined in `superset-frontend/plugins/plugin-chart-ag-grid-table/src/utils/useTableTheme.ts` that does the same thing.

superset-frontend/packages/superset-ui-core/src/components/ThemeSelect/index.tsx
EnxDev:
```suggestion
   <Icons.FormatPainterOutlined iconColor={theme.colorError} />
```

superset-frontend/src/dashboard/components/CssEditor/index.tsx
```diff
@@ -151,8 +256,17 @@ class CssEditor extends PureComponent<CssEditorProps, CssEditorState> {
               width="100%"
               editorProps={{ $blockScrolling: true }}
               enableLiveAutocompletion
-              value={this.state.css || ''}
+              value={this.state.pendingCss || ''}
             />
+            <div style={{ marginTop: 16, textAlign: 'right' }}>
```
EnxDev: I think it should be better to use `sizeUnit` and `float: right`

superset-frontend/src/dashboard/components/SaveModal.tsx
```diff
@@ -144,96 +115,87 @@ class SaveModal extends PureComponent<SaveModalProps, SaveModalState> {
...
-      />
-    );
-  }
+              {t('also copy (duplicate) charts')}
+            </Checkbox>
+          </Form.Item>
+        </Form>
+      }
+      modalFooter={
+        <Space style={{ display: 'flex', justifyContent: 'flex-end' }}>
```
EnxDev:
```suggestion
        <Flex justify="flex-end">
```

superset-frontend/src/features/themes/ThemeModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+      }
+    >
+      <Form layout="vertical">
+        {isSystemTheme && (
+          <Typography.Text
+            type="secondary"
+            style={{
+              fontSize: '14px',
+              fontStyle: 'italic',
+              marginBottom: '16px',
```
EnxDev: Do we need these styles here?

superset-frontend/src/features/themes/ThemeModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+            }}
+          >
+            {t('System Theme - Read Only')}
+          </Typography.Text>
+        )}
+
+        <Form.Item
+          label={t('Name')}
+          required={!isReadOnly}
+          style={{ marginBottom: '24px' }}
```
EnxDev: Do you think we should use `css` with `theme.sizeUnit` here?

superset-frontend/src/features/themes/ThemeModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+                {t('Ant Design Theme Editor')}
+              </a>
+              {t(', then paste the JSON below. See our')}{' '}
+              <a
+                href="https://superset.apache.org/docs/configuration/theming/"
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {t('documentation')}
+              </a>
```
EnxDev: is it better to use `Typography.Link` here?

superset-frontend/src/features/themes/ThemeModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+              {t(' for details.')}
+            </span>
+          }
+          style={{ marginBottom: '16px' }}
+        />
+
+        <Form.Item
+          label={t('JSON Configuration')}
+          required={!isReadOnly}
+          style={{ marginBottom: '24px' }}
```
EnxDev: Should we add `Space` component between the items instead of using `margin`?

EnxDev (commented):
LGTM! I just left a couple of comments; great contribution! 
This definitely makes customization and handling of custom themes much easier

superset-frontend/packages/superset-ui-core/src/theme/utils/themeUtils.ts
mistercrunch: I think both/either should return a dark enough color for tinycolor to categorize well.

superset-frontend/src/features/themes/ThemeModal.tsx
mistercrunch: good catch ... let me fix the whole form. AI went to town here, need to align with our styling practices.

superset-frontend/packages/superset-ui-core/src/components/ThemeSelect/index.tsx
```diff
@@ -43,36 +50,65 @@ const ThemeSelect: React.FC<ThemeSelectProps> = ({
...
   };
 
+  // Use different icon when local theme is active
+  const triggerIcon = hasLocalOverride ? (
+    <Icons.FormatPainterOutlined style={{ color: theme.colorErrorText }} />
+  ) : (
+    themeIconMap[themeMode] || <Icons.FormatPainterOutlined />
+  );
+
+  const menuItems: any[] = [
```
msyavuz: There should be a type for menu items

superset-frontend/packages/superset-ui-core/src/components/ThemeSelect/index.tsx
```diff
@@ -43,36 +50,65 @@ const ThemeSelect: React.FC<ThemeSelectProps> = ({
...
+            onClick: () => handleSelect(ThemeMode.SYSTEM),
+          },
+        ]
+      : []),
+  ];
+
+  // Add clear settings option only when there's a local theme active
+  if (onClearLocalSettings && hasLocalOverride) {
+    menuItems.push(
+      { type: 'divider' } as any,
```
msyavuz: These any assertions shouldn't be necessary

superset-frontend/src/features/themes/ThemeModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+    if (
+      currentTheme?.theme_name.length &&
+      currentTheme?.json_data?.length &&
+      isValidJson(currentTheme.json_data)
+    ) {
+      setDisableSave(false);
+    } else {
+      setDisableSave(true);
+    }
+  };
```
msyavuz, line 231: Maybe we can show an error when this fails, i think having incorrect json is common enough.

superset-frontend/src/pages/ThemeList/ThemeList.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+    await waitFor(() => {
+      expect(screen.getByText('Test Theme 1')).toBeInTheDocument();
+    });
+
+    // Should not show add button or bulk select
+    expect(screen.queryByText('Theme')).not.toBeInTheDocument();
+    expect(screen.queryByText('Bulk select')).not.toBeInTheDocument();
+  });
+});
```
msyavuz, line 201: This whole suite is missing the actual testing steps

superset-frontend/src/pages/ThemeList/index.tsx
```diff
@@ -0,0 +1,515 @@
...
+        addSuccessToast={addSuccessToast}
+        onModelImport={handleThemeImport}
+        show={importingTheme}
+        onHide={closeThemeImportModal}
+      />
+      {themeCurrentlyDeleting && (
+        <DeleteModal
+          description={
+            <>
+              <div style={{ marginBottom: 16 }}>
```
msyavuz: Yeah, this component also has some styling issues

tests/integration_tests/dashboards/api_tests.py
```diff
@@ -870,10 +871,13 @@ def test_add_favorite(self):
         """
         Dataset API: Test add dashboard to favorites
         """
+        # Let database auto-generate ID and use unique names
+        import uuid
+
+        unique_id = uuid.uuid4().hex[:8]
```
msyavuz, line 877: Do we need this?

superset-frontend/src/features/themes/ThemeModal.tsx
mistercrunch: ok just added this: 
Screenshot 2025-07-24 at 6 31 45 PM

mistercrunch: Did it in JsonEditor so all other uses of JsonEditor in the app will get this now too!

tests/integration_tests/dashboards/api_tests.py
mistercrunch: might be useful later on, planning on going uuid-first eventually

msyavuz (commented):
Currently we set some default tokens in `packages/superset-ui-core/src/theme/Theme.tsx` But they are not visible in the ui for the theme and i am not sure if they are applied to database either. Should we move them to default config options instead?

mistercrunch:
> Currently we set some default tokens in packages/superset-ui-core/src/theme/Theme.tsx But they are not visible in the ui for the theme and i am not sure if they are applied to database either. Should we move them to default config options instead?

The theme constructor is (AFAIK) deterministic (well kinda), meaning when you set your theme to `{algorithm: "dark"}`, as it gets created, will go through its construction, and defaults get applied, so the defaults here https://github.com/apache/superset/blob/master/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx#L73-L101 will get applied. Now if you were to pass an override like `{algorithm: "dark": token: { brandLogoUrl: "{some_url}" }}`, it'll override the default, and generated tokens will end up the same, except for that one. Now across versions of Superset where defaults might change, the "compiled" theme might change as defaults change, but I'm pretty sure that's what we want in most (all?) cases.

geido (approved):

msyavuz (approved):

eschutho (approved):

EnxDev (approved):

PR #34192, opened 2025-07-16
fix: dataset endpoint `/rowlevelsecurity/related/tables` doesn't apply filters as expected

mistercrunch:
## Summary

  This PR fixes an issue where the filter parameter in the Row Level Security (RLS) related tables endpoint
  (`/api/v1/rowlevelsecurity/related/tables`) was not being applied. When users tried to search for specific tables
  in the RLS CRUD UI, all tables were returned regardless of the filter value.

  ## Problem

  The `/api/v1/rowlevelsecurity/related/tables?q=(filter:mult,page:1,page_size:100)` endpoint was ignoring the
  `filter` parameter because:
  - The RLS API was missing the proper filter configuration for the `tables` field
  - Without this configuration, the `_get_related_filter` method in the base API didn't know which field to filter on

  ## Solution

  1. **Created a new filter class** `FilterRelatedTables` in `superset/views/filters.py` that:
     - Performs case-insensitive substring matching using SQLAlchemy's `ilike` with wildcards
     - Filters tables by their `table_name` field - Follows the same pattern as other "contains" filters in the codebase (e.g., `FilterRelatedOwners`, `DashboardTitleOrSlugFilter`)

  2. **Updated the RLS API configuration** to use this filter: ```python related_field_filters = { "tables": RelatedFieldFilter("table_name", FilterRelatedTables), "changed_by": RelatedFieldFilter("first_name", FilterRelatedOwners), }

  Testing

  Added a new test test_rls_tables_related_api_with_filter that verifies:
  - Filtering with a matching substring returns only tables containing that substring
  - Filtering with a non-existent substring returns no results
  - The filter is case-insensitive

  Changes

  - superset/views/filters.py: Added FilterRelatedTables class
  - superset/row_level_security/api.py: Added filter configuration for tables field
  - tests/integration_tests/security/row_level_security_tests.py: Added test for filter functionality

  Impact

  This fix improves the user experience in the RLS CRUD interface by allowing users to efficiently search and filter
  tables when creating or editing row-level security rules.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/security/row_level_security_tests.py
+++ b/tests/integration_tests/security/row_level_security_tests.py
@@ -555,6 +555,28 @@ class TestRowLevelSecurityWithRelatedAPI(SupersetTestCase):
         assert len(result) == len(db_tables)
         assert db_table_names == received_tables
 
+    def test_rls_tables_related_api_with_filter(self):
+        self.login(ADMIN_USERNAME)
+        # Test with filter that should match 'birth_names'
+        params = prison.dumps({"filter": "birth", "page": 0, "page_size": 100})
+        rv = self.client.get(f"/api/v1/rowlevelsecurity/related/tables?q={params}")
+        assert rv.status_code == 200
+        data = json.loads(rv.data.decode("utf-8"))
+        result = data["result"]
+        received_tables = {table["text"] for table in result}
+        # Should only return tables with 'birth' in the name
+        assert all("birth" in table_name.lower() for table_name in received_tables)
+        assert len(result) >= 1  # At least birth_names should be returned
+
+        # Test with filter that should match nothing
+        params = prison.dumps({"filter": "nonexistent", "page": 0, "page_size": 100})
+        rv = self.client.get(f"/api/v1/rowlevelsecurity/related/tables?q={params}")
+        assert rv.status_code == 200
+        data = json.loads(rv.data.decode("utf-8"))
+        result = data["result"]
+        assert len(result) == 0
+        assert data["count"] == 0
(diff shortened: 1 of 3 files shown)


tests/integration_tests/security/row_level_security_tests.py
```diff
@@ -555,6 +555,28 @@ def test_rls_tables_related_api(self):
...
+        rv = self.client.get(f"/api/v1/rowlevelsecurity/related/tables?q={params}")
+        assert rv.status_code == 200
+        data = json.loads(rv.data.decode("utf-8"))
+        result = data["result"]
+        received_tables = {table["text"] for table in result}
+        # Should only return tables with 'birth' in the name
+        assert all("birth" in table_name.lower() for table_name in received_tables)
+        assert len(result) >= 1  # At least birth_names should be returned
+
+        # Test with filter that should match nothing
```
sadpandajoe, line 573: nit: I think I'd prefer breaking this up into a separate test such that if there is a failure that it's pretty clear which scenario broke instead of having to investigate if it's not returning the right item or empty is broken.

sadpandajoe (approved):

Vitor-Avila (approved):
LGTM!

tests/integration_tests/security/row_level_security_tests.py
mistercrunch: ok, easy enough, done.

PR #34147, opened 2025-07-13
fix: adding and removing tags does not work in control panel properties modal

SBIN2010:
This PR fix #33758

When saving a dashboard after editing a property in a modal window, missing tag transfer has been added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34193, opened 2025-07-16
fix: bug when updating dashboard

SBIN2010:
This PR fix bag after #34147

To reproduce the error
Снимок экрана от 2025-07-16 22-00-34
, you need to try changing the dashboard with the owner assigned to the dashboard, and the dashboard should not contain any tags attached.
When receiving data about the dashboard, information is received about related tags whose type is not custom_tag

To fix the dashboard saving error, a filter for tags has been added

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/components/Header/index.jsx
```diff
@@ -415,7 +415,9 @@ const Header = () => {
       owners: dashboardInfo.owners,
       roles: dashboardInfo.roles,
       slug,
-      tags: dashboardInfo.tags,
+      tags: (dashboardInfo.tags || []).filter(
+        item => item.type === 1 || !item.type,
```
mistercrunch: do we have some sort of enum/constant anywhere on the frontend that properly labels what TagType == 1 means?

mistercrunch: found it on the backend -> https://github.com/apache/superset/blob/master/superset/tags/models.py#L61-L77 but couldn't find it on the frontend. Maybe we need a corresponding enum on the frontend if we don't have one already ... hoping we don't hard-code meaningless values related to TagType in other places in the frontend...

mistercrunch:
Nice, great improvement! You should install `pre-commit` if you haven't -> https://superset.apache.org/docs/contributing/development/#git-hooks , but looks like you need 

Oh looks like npm was down/strugglnig and failed `npm ci` in our CI, I re-triggered the job ...

rusackas:
@mistercrunch CI is passing if this looks good to ya otherwise :D

mistercrunch (approved):
LGTM, thanks for improving the types on the frontend!

PR #34226, opened 2025-07-18
chore(linting): setting no-console to error, and migrating instances to use `logging`

rusackas:
### SUMMARY

Use Claude Code to clean up lint warnings about this particular rule, and enforce it via `error`. Now it's set up so that the actual frontend code MUST use `logging` (which is safer, and extensible, i.e. we can hotwire it to send a copy of the message to the backend logging service, or... whatever). In tests, the opposite is true. We use `console` rather than `logging`. Also, `console` is allowed in the `scripts` folder, since those are an exception, too. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
## Test Failure Analysis

  The CI test failures are due to a pre-existing issue in the test environment setup, not
  directly caused by the console → logging changes in this PR.

  ### Root Cause
  The test suite fails when modules call translation functions (`t()`) at the module level
  during import. Specifically,
  `packages/superset-ui-chart-controls/src/sections/sections.tsx` calls `t('Time')` at
  module initialization, before the translation system is properly initialized in tests.

  ### What Changed
  While investigating, I made some improvements to the test setup files:
  - Fixed empty `data-bootstrap` attribute in `spec/helpers/setup.ts` to use valid JSON
  - These changes only affect test setup, not the production app

  ### Recommendation
  This is a broader test infrastructure issue that should be fixed separately. The console →
   logging changes in this PR are valid and don't cause the test failures - they just
  exposed existing fragility in how translation is initialized during tests.

PR #34233, opened 2025-07-19
fix(chart-download): ensure full table or handlebar chart is captured in image export

### LINKED ISSUE #33904: Using the "Download as Image" option in a table chart or handlebar chart nested in a dashboard only downloads the visible section of the chart and not the full table
### Bug description

When you have a table chart or a handlebar chart (This one simulating the table UI) inside a dashboard, if you use the Download as Image option of that chart, the downloaded image only considers the visual section shown in the dashboard and not the full chart.

1. Create a regular table chart.
2. Add to a dashboard, so that it doesn’t show up entirely (it shows the scroll bar).
3. Download the table chart from the dashboard as an image.

Image

## Expected results
The downloaded image considers all the information that the table chart has

## Actual results
The downloaded image only considers the information shown in the dashboard and not the full table 

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.11

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

fardin-developer:
Fixes: #33904 
### SUMMARY
This PR resolves an issue where using the "Download as Image" option on a table chart or handlebar chart within a dashboard only captures the visible portion of the chart (i.e., the section within the scroll view), instead of the entire chart content. This results in incomplete image exports when the chart content exceeds the visible area.

Screenshot 2025-07-19 at 1 50 57 PM

#### BEFORE
The downloaded image only includes the part of the table that is visible within the dashboard scroll area.
top-15-languages-spoken-at-home-2025-07-19T08-59-11 312Z

#### AFTER
The downloaded image now includes the full chart content, including data outside the visible scroll area.
chatTableWithRoeBar

### CHANGES

- Introduced logic to **detect scrollable elements or descendants** within the chart.
- If scrollable, the element is **cloned**, and styles are adjusted to:
  - Remove height constraints.
  - Make all content visible (including overflow).
  - Preserve the full scrollable area.
- The cloned element is rendered **off-screen** to allow full-content rendering without affecting layout.
- Used the adjusted element to generate the image via **`dom-to-image`**, capturing the **entire scrollable content**.

### DIFF
--- a/superset-frontend/src/utils/downloadAsImage.ts
+++ b/superset-frontend/src/utils/downloadAsImage.ts
@@ -23,6 +23,8 @@ import { kebabCase } from 'lodash';
 import { SupersetTheme, t } from '@superset-ui/core';
 import { addWarningToast } from 'src/components/MessageToasts/actions';
 
+const IMAGE_DOWNLOAD_QUALITY = 0.95;
+
 /**
  * generate a consistent file stem from a description and date
  *
@@ -266,7 +268,7 @@ export default function downloadAsImageOptimized(
       const dataUrl = await domToImage.toJpeg(clone, {
         bgcolor: theme?.colors.grayscale.light4,
         filter,
-        quality: 0.95,
+        quality: IMAGE_DOWNLOAD_QUALITY,
         height: clone.scrollHeight,
         width: clone.scrollWidth,
         cacheBust: true,
--- a/superset-frontend/src/utils/downloadAsImage.ts
+++ b/superset-frontend/src/utils/downloadAsImage.ts
@@ -70,9 +70,41 @@ export default function downloadAsImage(
       return true;
     };
 
+    // Check if element is scrollable
+    const isScrollable = hasScrollableDescendant(elementToPrint as HTMLElement);
+    let targetElement = elementToPrint as HTMLElement;
+    let cloned: HTMLElement | null = null;
+    const extraPadding = theme?.sizeUnit ? theme.sizeUnit * 2 : 16;
+    if (isScrollable) {
+      cloned = elementToPrint.cloneNode(true) as HTMLElement;
+      cloned.style.overflow = 'visible';
+      cloned.style.maxHeight = 'none';
+      cloned.style.height = 'auto';
+      cloned.style.width = `${elementToPrint.scrollWidth}px`;
+
+      cloned.querySelectorAll<HTMLElement>('*').forEach(el => {
+        // eslint-disable-next-line no-param-reassign
+        el.style.overflow = 'visible';
+        // eslint-disable-next-line no-param-reassign
+        el.style.maxHeight = 'none';
+        // eslint-disable-next-line no-param-reassign
+        el.style.height = 'auto';
+      });
+
+      // Render off-screen
+      cloned.style.position = 'absolute';
+      cloned.style.top = '-9999px';
+      cloned.style.left = '-9999px';
+      document.body.appendChild(cloned);
+
+      targetElement = cloned;
+    }
+
     return domToImage
-      .toJpeg(elementToPrint, {
+      .toJpeg(targetElement, {
         bgcolor: theme?.colors.grayscale.light4,
+        height: targetElement.scrollHeight + extraPadding,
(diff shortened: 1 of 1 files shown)


superset-frontend/src/utils/downloadAsImage.ts
```diff
@@ -71,18 +156,35 @@ export default function downloadAsImage(
...
+        quality: 0.95,
+        height: clonedElement.scrollHeight,
+        width: clonedElement.scrollWidth,
       })
       .then((dataUrl: string) => {
+        // Clean up the cloned element
+        const tempContainer = clonedElement.parentElement;
+        if (tempContainer && tempContainer.parentElement) {
+          tempContainer.parentElement.removeChild(tempContainer);
+        }
```
msyavuz: can we return this as a function from `createEnhancedClone` function to both reduce code duplication and remove the possibility of someone using that without a cleanup?

msyavuz (commented):
I have some concerns about the performance of this solution but it works and we might not have many other options than cloning

fardin-developer:
> I have some concerns about the performance of this solution but it works and we might not have many other options than cloning

Thanks for your feedback. I was thinking the same about the performance issue. I have found a solution that significantly improved performance. Although I am still cloning it, I tried a different approach this time and the download is now much faster with a high quality image.

superset-frontend/src/utils/downloadAsImage.ts
fardin-developer: Thank you for your suggestion I have modified createEnhancedClone function signature: Now returns an object with both the clone element and a cleanup function

fardin-developer:
## Performance Enhancements
-  **WeakMap-based style caching**
A WeakMap<Element, CSSStyleDeclaration> is employed to cache the results of getComputedStyle, eliminating redundant computations for elements that have already been processed. This cache provides efficient O(1) lookups and automatically clears entries when elements are removed from the DOM, reducing traversal overhead.
- **Breadth-First Search (BFS) DOM Traversal**
The recursive style-copying logic has been replaced with an iterative Breadth-First Search (BFS) approach using an explicit queue (Array<[Element, Element]>). This change offers several benefits:
       1. Prevents stack overflow in deeply nested DOM structures.
       2. Reduces per-node function call overhead by eliminating recursion.
       3. Ensures consistent and complete traversal across large or complex trees.
- **Critical Style Filtering Using Set**
To optimize cloning further, only a defined set of critical CSS properties is copied using a Set for fast membership checks. This selective copying reduces the size of the cloned DOM and improves rendering performance by excluding unnecessary styles such as transitions and browser defaults.

## See the difference between the previous and optimized approach:
[banner](https://drive.google.com/file/d/1uFUvm8OQb35Fm9mKJvHR81Crdf6X3mgu/view?usp=sharing)



Ckick the banner to watch the video
# [Link](https://drive.google.com/file/d/1uFUvm8OQb35Fm9mKJvHR81Crdf6X3mgu/view?usp=sharing) : Click the banner or link to Watch the Video

superset-frontend/src/utils/downloadAsImage.ts
```diff
@@ -32,57 +32,260 @@ import { addWarningToast } from 'src/components/MessageToasts/actions';
...
-      })
-      .then((dataUrl: string) => {
-        const link = document.createElement('a');
-        link.download = `${generateFileStem(description)}.jpg`;
-        link.href = dataUrl;
-        link.click();
-      })
-      .catch((e: Error) => {
-        console.error('Creating image failed', e);
+        quality: 0.95,
```
msyavuz: Can we have a variable for this instead of a magic number?

msyavuz (approved):

sorren205:
Is it possible to apply the same logic to ‚Export to PDF‘ Section. This would be a great improvement.

fardin-developer:
> Is it possible to apply the same logic to ‚Export to PDF‘ Section. This would be a great improvement.

Yes I think it’s possible to download as pdf also.

sorren205:
Thanks for your answer. Do I have to open a new issues so that someone can work on it to get it implemented? @fardin-developer

Issue #5372, opened 2018-07-11
CSV_EXPORT.encoding config has no effect in Python3


georgexsh:
### Superset version

Superset 0.26.3 and current HEAD 6b15592
Python 3.5.3 (via docker)

### Expected results

after added following config: 
```
CSV_EXPORT = {
    'encoding': 'utf-8-sig',
}
```
expected the exported csv file is encoded in `utf-8-sig`

### Actual results

the exported CSV file is still encoded in `utf-8`, no BOM.

### Steps to reproduce

set this option, export a CSV file in SQL lab.

related with #4506

### Root cause

superset exports csv with following code:

```
csv = df.to_csv(index=False, **config.get('CSV_EXPORT'))
```

`pandas.DataFrame.to_csv` [doc says](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_csv.html#pandas.DataFrame.to_csv):

> path_or_buf : string or file handle, default None
> File path or object, if None is provided the result is returned as a string.

as superset has not set `path_or_buf` parameter, pandas use `StringIO` internally, in Python3, `encoding` parameter is omitted:

```
>>> df = pd.DataFrame(['中文', '测试'])
>>> r1 = df.to_csv()
>>> r2 = df.to_csv(encoding='utf-8-sig')
>>> r1 == r2
True
>>> type(r1), type(r2)
(<class 'str'>, <class 'str'>)
```

superset should do the encoding itself, in python3:

```
--- a/superset/views/core.py
+++ b/superset/views/core.py
@@ -2658,6 +2658,7 @@ class Superset(BaseSupersetView):
             df = query.database.get_df(sql, query.schema)
             # TODO(bkyryliuk): add compression=gzip for big files.
             csv = df.to_csv(index=False, **config.get('CSV_EXPORT'))
+        csv = csv.encode(config.get('CSV_EXPORT')['encoding'])
         response = Response(csv, mimetype='text/csv')
         response.headers['Content-Disposition'] = (
             'attachment; filename={}.csv'.format(unidecode(query.name)))
```

wutengyue:
nice!

PR #34235, opened 2025-07-20
fix(sqllab_export): manually encode CSV output to support utf-8-sig

Habeeb556:
Manually encodes the CSV string before returning it in the response. Since `Werkzeug 3` removed support for automatic encoding.

### SUMMARY

By explicitly encoding the CSV output as `utf-8-sig`, this change ensures proper rendering of non-English characters (e.g., Arabic) in tools like Excel. This is necessary because the CSV file itself isn't interpreted correctly without a BOM, and `Werkzeug 3` no longer handles encoding automatically—so this PR applies the encoding manually to maintain compatibility.

Before:

image

After:

image

### ADDITIONAL INFORMATION

- [X] Has associated issue: #5372, #4506, #29224, #29410, #29506
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch:
Looks like some unit tests are failing... Not sure this won't break anything for anyone out there.

Habeeb556:
> Looks like some unit tests are failing... Not sure this won't break anything for anyone out there.

Due to a **type mismatch**, the test is comparing a **byte string** (`b'foo\n1\n2\n'`) with a **regular string** (`'foo\n1\n2\n'`), which will always fail in Python.

The test assumes the function returns a string, which is fine for displaying data. However, since the actual export returns **bytes** (raw binary data)—as expected for file downloads—the test should be updated to expect bytes instead.

Additionally, as noted in Werkzeug issue https://github.com/pallets/werkzeug/issues/3057, **automatic encoding was disabled in Werkzeug 3**, so this now needs to be handled manually. That’s what I have implemented here.

While downgrading to `Werkzeug==2.3.8` does resolve the issue temporarily, it's not ideal. That version was flagged in issue #28623 due to a **CVE**, so relying on it long-term would be risky.

mistercrunch:
in 2025 utf8 should be supported across the ecosystem really. So about the unit test you should be ok to cast to byte string on the right side of the assertion and pass the test (?)

Oh, about Werkzeug, should we /could we just configure it in `superset/config.py` to do the auto-encoding (?)

Just hoping that nothing is going to break here. Someone somewhere might use an old version of Excel and report an issue at some point. Unclear whether the whole ecosystem is ready for utf8 csv (?) Wondering if we need a checkbox "export as utf8" that people can check/uncheck. Though we could move to utf8 and wait issues to be reported to decide whether we should add a checkbox?

mistercrunch:
My take: assume the ecosystem supports utf8 and merge this PR (after unit tests are fixed) . If/when someone reports issues with utf8 support in antiquated tools, we recommend to upgrade their Excel 95/XP to something from this era.

Habeeb556:
I may have miscommunicated the intent in the PR summary—my concern isn’t about outdated Excel versions, but rather the fact that the CSV file itself isn’t interpreted correctly when it contains non-English characters (like Arabic, Chinese, etc.) even it's encoded with `utf-8-sig`. Since `Werkzeug 3` removed automatic encoding, this needs to be just handled manually from the hard code.

By default, the encoding is set to `utf-8`, but it can be overridden by specifying a custom value in `superset/config.py`

`CSV_EXPORT = {"encoding": "utf-8-sig"}` — this keeps the behavior configurable based on deployment needs.

Tests are now updated to match the byte output.

mistercrunch:
Should we deprecate `CSV_EXPORT = {"encoding": "utf-8-sig"}` then?

mistercrunch:
My take here is that we should just make things work properly with UTF-8 out of the box, but I get why we need the BOM now.

From my understanding, the issue is Excel - without the BOM signature bytes, Excel can't figure out that our UTF-8 CSV files are actually UTF-8, so it mangles any non-English characters. That's why `utf-8-sig` matters - it adds those 3 magic bytes (`\xEF\xBB\xBF`) that tell Excel "hey, this is UTF-8!"

But I still think we should avoid config complexity. A few thoughts:

- Let's just default to `utf-8-sig` for all CSV exports - it's backward compatible and fixes the Excel issue
- The BOM is valid UTF-8 and most modern tools handle it fine (unlike the bad old days)
- No need for a `CSV_EXPORT` config - this should "just work" without users having to know about BOMs

Can we simplify this to:
1. Always use `utf-8-sig` for CSV exports (handles Excel + international chars)  
2. Remove the config option entirely
3. Add tests for various UTF-8 scenarios (Arabic, Chinese, etc.)

What do you think about just making this the default behavior? (while noting the change in UPDATING.md)

Habeeb556:
Awesome! This PR covers two important parts:

1- The first part is about manually encoding the response — which I originally opened this PR for. Even when setting `utf-8-sig`, it wasn’t working correctly due to changes in **Werkzeug 3**, which now requires manual handling of response encoding.

2- The second part, which you suggested (and I agree with), is to use `utf-8-sig` by default to support the BOM and make things work out of the box — especially for Excel compatibility.

As for removing the config option, I would suggest we **keep it** for flexibility, so users can still choose a different encoding if needed. But yes, we should **update the default to `utf-8-sig`** to avoid issues and support international characters.

mistercrunch:
Perfect, we're aligned then! So change the default in `superset/config.py` and add a note in `UPDATING.md` about this minor change-of-default-value?

Habeeb556:
@mistercrunch Verified that the CSV export now correctly includes a UTF-8 BOM (`utf-8-sig`) and handles multilingual data as expected.
- Added a test case to check for BOM presence and correct parsing of Arabic and Chinese characters.
- Also updated `UPDATING.md` to reflect this change.

Everything is working as intended — ready for merge.

mistercrunch (approved):
🔥 🚀

PR #34253, opened 2025-07-21
fix(theming): Theming visual fixes

### LINKED ISSUE #34260: Dark Mode: Text formatting and Visibility Issue in UI
## Screenshot

Markdown Element:
Image

In Edit Mode:
Image

In Chart Creation:
Light Mode:
Image

Dark Mode:
Image

## Description
In Dark Mode, the when the background changes correspondingly the text color doesn't change.

msyavuz:
### SUMMARY

Fixes for various theming visuals

1. Filters overlapping with action buttons:
Before:
image
After:
image

2. UploadDataModal fixes to titles, collapse sizes and remove metadata
Before:
image

After:
image

3. Secondary buttons contrast
Before:
image

After:
image

4. Delete modal spacing
Before:
image
After:
image

5. Async Ace Editor
Before:
image

After:
image

6. Explore columns and metrics
Before:

image

After:
image

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #34260 
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/components/nativeFilters/FilterBar/ActionButtons/index.tsx
```diff
@@ -75,13 +74,9 @@ const verticalStyle = (theme: SupersetTheme, width: number) => css`
 
   background: linear-gradient(
     ${rgba(theme.colors.grayscale.light5, 0)},
-    ${theme.colors.grayscale.light5} 60%
+    ${theme.colors.grayscale.light5} 20%
```
mistercrunch: let's use one of the `colorBg.*` for this. Guessing `colorBgLayout` or `colorBgContainer` (?)

mistercrunch: we still have the task to deprecate all `theme.colors.*` in the repo ...

msyavuz: Right, this file has some more of those. I might as well touch them while at it. Done!

superset-frontend/packages/superset-ui-core/src/components/Button/index.tsx
```diff
@@ -134,6 +133,11 @@ export function Button(props: ButtonProps) {
         '& > span > :first-of-type': {
           marginRight: firstChildMargin,
         },
+        ':not(:hover)': effectiveButtonStyle === 'secondary' && {
+          // Increase contrast for secondary buttons
+          backgroundColor: `${tinycolor(theme.colorPrimaryBg).darken(2).toHexString()} !important`,
```
mistercrunch: ideally here we use `colorPrimary.*` instead of darken/lighten

mistercrunch: can we use these?

Image

mistercrunch (approved):
Minor comment on leveraging `colorPrimary.*` tokens instead of lighten/darken, but approving regadless. Please use the right tokens if they exist or the best one as proxy if they don't

superset-frontend/packages/superset-ui-core/src/components/Button/index.tsx
msyavuz: I looked into this but using colorPrimaryHover or Border for the background of secondary button seemed off semantically. ~Do you think it's better than using lighten/darken?~ Nevermind saw the comment below

mistercrunch: feel free to move forward with this, could extend the framework with extra semantics for `color.*Bg{something}` if there's a clear need for more semantics ....

mistercrunch:
One option would be to add a TODO or NOTE of some kind when we abuse the semantics for a particular shade by using a proxy.

msyavuz:
What i am finding out now is that we don't really have a good option when it comes to contrast for secondary button:

image

I went with the lightest and darkest here, which is slightly better than what we had before. I think we might need to extend the framework for buttons specifically since we are customizing it so much to our needs. Altering the algorithm for our default themes also should work.

mistercrunch:
> I think we might need to extend the framework for buttons specifically since we are customizing it so much to our needs.

Curious what semantic you would introduce? something like `colorPrimaryBgSecondary`? Noting that it might be a bit of a mess since the `darkAlgorithm` is what generates a lot of the shading with builtin logic. Might be better to establish design constraints, telling @kasiazjc she has to work with the shades/tokens in the framework (?)

msyavuz:
If we need a secondary button we can introduce seed tokens for it and extend the algorithm to produce necessary map tokens (in the default themes). `colorSecondary` `colorSecondaryBg` etc. It will mean diversion from Antd but we already do that. 

Also since user's will use Ant Design's theme editor to customize their theme, none of the options seem good enough:

1. Introduce `colorSecondary` as a seed token: Doesn't exist on the theme editor
2. Use semantically incorrect tokens: Unexpected colors that doesn't reflect in theme editor.
3. Lighten/Darken: Same thing as the first one but smaller impact.

So yes, design constraints seem like the easier/most straightforward option out of these

PR #34257, opened 2025-07-21
fix(drill-to-detail): drill to detail by axis with label on axis column

### LINKED ISSUE #33907: Drill to Detail by Axis value does not work if the Axis column has a label
### Bug description

It's possible to Drill to Detail an Echarts chart by an Axis value. This is currently not working properly in case the Axis column has a label defined at the dataset level.

1. Edit your dataset, to add a label to a column.
2. Create Bar Chart, using the column from step 1 as the X-Axis.
3. Add a metric and a dimension, and save this chart to a dashboard.
4. Access the dashboard.
5. Right-click any bar, and use the Drill to Detail by $AXIS_VALUE option.

## Expected results

The D2D table should list only results that match that axis value.

## Actual results

The D2D table shows all values.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.11

### Node version

18 or greater

### Browser

Chrome

### Additional context

_No response_

### Checklist

zaminalishah:
### SUMMARY
When the axis column in a timeseries chart had a label set on the dataset level then the drill to detail by axis value wasnt working. The problem was that we were sending the label to the backend for filtering but the backend expected the column name.

https://github.com/user-attachments/assets/2a15da9c-702d-4e83-8b7c-941ebcf3fffd

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #33907 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/EchartsTimeseries.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/EchartsTimeseries.tsx
@@ -190,7 +190,7 @@ export default function EchartsTimeseries({
               // if the xAxis is '__timestamp', granularity_sqla will be the column of filter
               xAxis.label === DTTM_ALIAS
                 ? formData.granularitySqla
-                : xAxis.label,
+                : formData.xAxis,
             grain: formData.timeGrainSqla,
             op: '==',
             val: data[0],


rusackas:
Thanks for the PR! Approving CI to run 🤞

msyavuz (changes requested):
You are doing drillDetailBy gender in this case. Also the video showing it works is not reflecting the change since it is a temporal column in x axis. Try with anythin else than a temporal column and you will see it shows everything still.

PR #34261, opened 2025-07-21
fix: Matching errorType on superset api error with SupersetError

justinpark:
### SUMMARY

For Airbnb, we display custom errors according to the error_type. However, in the current Superset API, unlike the query result API, SupersetApiError is thrown, and [error_type is mapped to errorType](https://github.com/apache/superset/blob/b24aca0304b53cdf6f92bf600b64152015cafa30/superset-frontend/packages/superset-ui-core/src/query/api/v1/types.ts#L110), which causes custom overrides to not work properly. In this PR, I have resolved the issue by correcting the type of the errorType value.

Before:
Screenshot 2025-07-21 at 1 45 19 PM

After:
Screenshot 2025-07-21 at 1 45 02 PM

### TESTING INSTRUCTIONS
locally

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/query/types/Query.ts
+++ b/superset-frontend/packages/superset-ui-core/src/query/types/Query.ts
@@ -256,14 +256,14 @@ export type ErrorSource = 'dashboard' | 'explore' | 'sqllab' | 'crud';
 export type SupersetError<ExtraType = Record<string, any> | null> =
   | {
       error_type: ErrorType;
-      errorType: undefined;
+      errorType?: undefined;
       extra: ExtraType;
       level: ErrorLevel;
       message: string;
     }
   | {
       errorType: ErrorType;
-      error_type: undefined;
+      error_type?: undefined;
       extra: ExtraType;
       level: ErrorLevel;
       message: string;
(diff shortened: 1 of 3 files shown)


superset-frontend/src/explore/components/controls/DatasourceControl/index.jsx
```diff
@@ -454,16 +455,14 @@ class DatasourceControl extends PureComponent {
         {isMissingDatasource && !isMissingParams && (
           <div className="error-alert">
             {extra?.error ? (
-              <div className="error-alert">
```
justinpark, line 457: removed redundant same error-alert wrapper

superset-frontend/packages/superset-ui-core/src/query/types/Query.ts
```diff
@@ -255,6 +255,7 @@ export type ErrorSource = 'dashboard' | 'explore' | 'sqllab' | 'crud';
 
 export type SupersetError<ExtraType = Record<string, any> | null> = {
   error_type: ErrorType;
+  errorType?: ErrorType;
```
michael-s-molina: @justinpark Can't we fix the API responses to use the same field? Having both in the type definition looks really bad and is an indicative that we should fix the APIs.

michael-s-molina (approved):

PR #34272, opened 2025-07-22
fix: build issues on master with 'npm run dev'

mistercrunch:
## Summary
  - Fixed webpack alias for missing `@icons/material/UnfoldMoreHorizontalIcon` in react-color dependency
  - Changed datamaps import from `datamaps.world.min` to `datamaps.all.min` in WorldMap plugin to resolve build errors
 
## Root cause: The issues likely stemmed from the recent webpack-dev-server
  upgrade from v4.15.2 to v5.2.1 (commit 11215b092a) on July 8th, which also
   included:
  - webpack upgrade from 5.98.0 to 5.99.9
  - webpack-cli upgrade from 4.10.0 to 6.0.1

  Webpack v5 has stricter module resolution and the new version likely
  exposed these existing issues:

  1. react-color issue: The @icons/material/UnfoldMoreHorizontalIcon import
  was always broken, but webpack v4 may have been more lenient about missing
   modules
  2. datamaps issue: The datamaps.world.min vs datamaps.all.min - the
  .world.min version may not exist in the current datamaps package version,
  or webpack v5's stricter resolution revealed this

  The babel plugin cleanup (commit f4754641c8) on July 12th may have
  contributed as well, but the webpack upgrade is the more likely culprit
  for module resolution issues.

  ## Test plan
  - [x] `npm run dev` builds without module resolution errors
  - [x] react-color components render correctly with fallback icon
  - [x] WorldMap plugin loads without import errors

### DIFF
--- a/superset-frontend/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
+++ b/superset-frontend/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
@@ -24,7 +24,7 @@ import {
   getSequentialSchemeRegistry,
   CategoricalColorNamespace,
 } from '@superset-ui/core';
-import Datamap from 'datamaps/dist/datamaps.world.min';
+import Datamap from 'datamaps/dist/datamaps.all.min';
 import { ColorBy } from './utils';
 
 const propTypes = {
--- a/superset-frontend/webpack.config.js
+++ b/superset-frontend/webpack.config.js
@@ -319,6 +319,13 @@ const config = {
           './node_modules/@storybook/react-dom-shim/dist/react-16',
         ),
       ),
+      // Workaround for react-color trying to import non-existent icon
+      '@icons/material/UnfoldMoreHorizontalIcon': path.resolve(
+        path.join(
+          APP_DIR,
+          './node_modules/@icons/material/CubeUnfoldedIcon.js',
+        ),
+      ),
     },
     extensions: ['.ts', '.tsx', '.js', '.jsx', '.yml'],
     fallback: {


superset-frontend/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
```diff
@@ -24,7 +24,7 @@ import {
   getSequentialSchemeRegistry,
   CategoricalColorNamespace,
 } from '@superset-ui/core';
-import Datamap from 'datamaps/dist/datamaps.world.min';
+import Datamap from 'datamaps/dist/datamaps.all.min';
```
eschutho, line 27: This could increase the size of our js bundle. Just FYI in case we see something in the future and are looking for a reason for it.

eschutho (approved):

superset-frontend/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
rusackas: 114KB vs 142KB. You're right, but _probably_ nothing to lose sleep over?

superset-frontend/webpack.config.js
```diff
@@ -319,6 +319,13 @@ const config = {
           './node_modules/@storybook/react-dom-shim/dist/react-16',
         ),
       ),
+      // Workaround for react-color trying to import non-existent icon
+      '@icons/material/UnfoldMoreHorizontalIcon': path.resolve(
+        path.join(
+          APP_DIR,
+          './node_modules/@icons/material/CubeUnfoldedIcon.js',
+        ),
+      ),
```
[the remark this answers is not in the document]
rusackas: @mistercrunch  is our bot friend right? There is indeed a UnfoldMoreHorizontalIcon file in that node module at `./node_modules/@icons/material/UnfoldMoreHorizontalIcon`, so I'm not sure what's up it being considered nonexistent.

superset-frontend/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
mistercrunch: yeah main mission is fixing `master` for now ...

hainenber:
@mistercrunch how did you encounter the build issue? I'm currently running latest `master` with changs in PR being uncommented but unable to reproduce it :(

image
image
image


Edit: I checked out to the commit before this one in `master` branch and rebuild for reproduction but to no avail.

mistercrunch:
> @mistercrunch how did you encounter the build issue? I'm currently running latest master with changs in PR being uncommented but unable to reproduce it :(

oh now you're making me doubt whether my issues could have been local (?) Clearly there's some things that happened recently that didn't play nice with the frontend build cache. Or maybe your environment is ok because of some caching. Note that prior to this fix, I did `npm run prune`, `rm -rf node_modules`, deleted `.temp_cache`, and maybe even nuked npm cache, all of those not successful... So came up with this fix

Now thinking that if I had time, I'd start clean with a new cloned repo and see if my issues can be reproduced prior to this PR.

What I'd do (or might do if I can find the time...):
- start super fresh, git clone in a new folder (no cache!) as of the SHA right before this commit here got merged on `master`
- nuke database
- nuke npm cache, any other cache ...
- `docker compose up`, see if things build or if we see any build-related issues
- if no issues, revert this PR ...
- if there are issues, try as of after this PR to see if my PR indeed fixes issues

Now after this fix or sometime yesterday, I did discover I had a zombie `npm run dev` process that was causing havoc. This whole saga could be related to that, but I'm unclear on that...

PR #34274, opened 2025-07-23
feat: optimize setup-backend to go faster on py 3.12

mistercrunch:
## Summary

  Optimizes CI build times for Python 3.12 by addressing missing wheel availability that was causing 10+ minute longer build times compared to other Python versions.

  ## Changes

  - **Added uv package caching**: Implements caching for uv packages using `~/.cache/uv` to avoid repeated downloads
  - **Enabled `--prefer-binary` flag**: Forces uv to prefer pre-built binary wheels over source compilation for all pip install commands
  - **Removed outdated comment**: Cleaned up comment about Python 3.12 being disabled due to compatibility issues

  ## Performance Impact

  - **Before**: Python 3.12 builds took 10+ minutes longer than other versions due to compiling packages from source
  - **After**: Should match performance of other Python versions by using pre-built wheels when available

  ## Technical Details

  The root cause was Python 3.12's newer release meant fewer packages had pre-built wheels available, forcing compilation from source. The `--prefer-binary` flag tells uv to prioritize binary wheels, and the caching ensures we don't
  re-download packages across CI runs.

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -28,7 +28,6 @@ runs:
         if [ "${{ inputs.python-version }}" = "current" ]; then
           echo "PYTHON_VERSION=3.11" >> $GITHUB_ENV
         elif [ "${{ inputs.python-version }}" = "next" ]; then
-          # currently disabled in GHA matrixes because of library compatibility issues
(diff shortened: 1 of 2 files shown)


mistercrunch:
gave up for now since it's complicated -> https://github.com/astral-sh/uv/issues/1794

PR #34289, opened 2025-07-23
fix: Saved queries list break if one query can't be parsed

michael-s-molina:
### SUMMARY
If a saved query can't be parsed by SQL Glot to extract the table names, the whole listing page breaks with the following error:

```
An error occurred while fetching Saved queries: list index out of range
```

Fixes https://github.com/apache/superset/issues/34019.

### TESTING INSTRUCTIONS
1. Save a query.
2. Edit the query with invalid SQL.
3. Go to the saved queries list.
4. The list should render normally and the problematic SQL should have nothing in the Tables column.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/models/sql_lab.py
+++ b/superset/models/sql_lab.py
@@ -48,7 +48,7 @@ from sqlalchemy.orm import backref, relationship
 from sqlalchemy.sql.elements import ColumnElement, literal_column
 
 from superset import security_manager
-from superset.exceptions import SupersetSecurityException
+from superset.exceptions import SupersetParseError, SupersetSecurityException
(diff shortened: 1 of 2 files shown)


michael-s-molina:
Tagging @betodealmeida @dpgaspar for security review.

betodealmeida (approved):
Looks good! Can we add a test to make sure it works?

michael-s-molina:
> Looks good! Can we add a test to make sure it works?

Done!

PR #34290, opened 2025-07-23
feat: make `SupersetClient` retry on 502-504

betodealmeida:
### SUMMARY

Make `SupersetClient` retry on 502, 503, and 504. Configurable in `superset_config.py`.

Video showing the retries (ignore unrelated overlay error):

https://github.com/user-attachments/assets/c240672d-2def-49ac-839b-bd5cc41a43ef

### TESTING INSTRUCTIONS

Tested by adding a 50% chance of returning a 502 in the `/api/v1/chart/data` endpoint. Loaded a dashboard, saw the retries and that everything loaded fine in the end.

```python
diff --git a/superset/charts/data/api.py b/superset/charts/data/api.py
index efae7de831..239da547e4 100644
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import contextlib
 import logging
+import random
 from typing import Any, TYPE_CHECKING
 
 from flask import current_app, g, make_response, request, Response
@@ -233,6 +234,11 @@ class ChartDataRestApi(ChartRestApi):
         if json_body is None:
             return self.response_400(message=_("Request is not JSON"))
 
+        # TEMPORARY: Add 50% chance of 502 error for retry testing
+        if random.random() < 0.5:  # noqa: S311
+            logger.warning("TEMPORARY: Simulating 502 error for retry testing")
+            return self.response(502, message="Bad Gateway - Temporary testing error")
+
         try:
             query_context = self._create_query_context_from_form(json_body)
             command = ChartDataCommand(query_context)
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/config.py
```diff
@@ -166,6 +166,17 @@ def _try_json_readsha(filepath: str, length: int) -> str | None:
 NATIVE_FILTER_DEFAULT_ROW_LIMIT = 1000
 # max rows retrieved by filter select auto complete
 FILTER_SELECT_ROW_LIMIT = 10000
+
+# SupersetClient HTTP retry configuration
+# Controls retry behavior for all HTTP requests made through SupersetClient
+# This helps handle transient server errors (like 502 Bad Gateway) automatically
+SUPERSET_CLIENT_RETRY_ATTEMPTS = 3  # Maximum number of retry attempts
```
mistercrunch, line 173: could take either side of that debate, but would a nested object a better idea than multiple config keys here?

betodealmeida: Wouldn't a nested object be harder to configure via env vars using that refactoring you started?

mistercrunch: oh I think it might be better, though not sure yet, but was hoping for JSON schema + antd autoform magic

superset-frontend/src/components/Chart/chartAction.js
```diff
@@ -537,7 +537,11 @@ export function postChartFormData(
 
 export function redirectSQLLab(formData, history) {
   return dispatch => {
-    getChartDataRequest({ formData, resultFormat: 'json', resultType: 'query' })
+    getChartDataRequest({
+      formData,
+      resultFormat: 'json',
+      resultType: 'query',
+    })
```
eschutho, line 544: @betodealmeida will this also retry when fetching filters?

betodealmeida: Yes! :)

dpgaspar (approved):

PR #34306, opened 2025-07-24
chore(npm): bump node engine, build fresh package-lock.json

rusackas:
### SUMMARY
Saw some npm installation warnings about the node engine version, so I'm bumping it here. While we're at it, since this is a breaking change window, let's trash the package-lock.json file and build a fresh one, right?!

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/legacy-preset-chart-deckgl/package.json
```diff
@@ -25,7 +25,7 @@
   ],
   "dependencies": {
     "@deck.gl/aggregation-layers": "^9.1.13",
-    "@deck.gl/core": "^9.1.14",
+    "@deck.gl/core": "^9.1.13",
```
rusackas: Not sure why this was downgraded, but... it seems to work?

superset-frontend/src/explore/components/controls/ControlPopover/ControlPopover.test.tsx
```diff
@@ -89,8 +80,8 @@ test('Should lock the vertical scroll when the popover is visible', () => {
   );
 });
 
-test('Should place popover at the top', async () => {
-  const { setStateMock } = setupTest({
+test.skip('Should place popover at the top', async () => {
```
rusackas, line 83: Maybe these skipped tests are a concern... or not... perhaps I'll pull in this file from `master` and see if CI still passes, if we're concerned here.

PR #34308, opened 2025-07-25
feat(timeseries): enhance 'Series Limit' to support grouping the long tail

mistercrunch:
Been meaning to implement this for years.

Screenshot 2025-07-24 at 9 59 54 PM
Screenshot 2025-07-24 at 10 25 49 PM
Screenshot 2025-07-24 at 9 59 42 PM

## Summary

  Adds **"Group remaining as Others"** checkbox to series limit control. When enabled, remaining series beyond the limit are grouped into "Others" instead of being randomly truncated at row limits.

  Includes DRY refactoring that eliminates ~90% code duplication between JOIN and 2-phase query paths.

  ## Changes

  - **Frontend**: New checkbox control next to series limit
  - **Backend**: `group_others_when_limit_reached` parameter with ANSI SQL CASE expressions
  - **Refactor**: Extract `_apply_series_others_grouping()` method with condition factory pattern
  - **Tests**: Unit tests for refactored logic

  ## Technical Details

  **Before**: 63 lines of duplicated CASE expression logic
  **After**: 17 lines using shared method + condition factories

  **Backwards compatible**: Default `false`, generates identical SQL when disabled

  ## Files Changed
  - `superset/common/query_object.py` - Add parameter
  - `superset/models/helpers.py` - Add method + refactor both query paths
  - `superset/charts/schemas.py` - Add API field
  - `superset-frontend/.../sharedControls.tsx` - Add control
  - `superset-frontend/.../sections.tsx` - Update layout
  - `tests/unit_tests/models/helpers_test.py` - Add tests

### on dbs that support subqueries

Screenshot 2025-07-24 at 10 26 58 PM

### two phases on dbs that don't 
Screenshot 2025-07-24 at 10 27 35 PM


superset/models/helpers.py
```diff
@@ -1258,6 +1258,65 @@ def _get_top_groups(
...
+        # Modify SELECT expressions
+        modified_select_exprs = []
+        for expr in select_exprs:
+            if hasattr(expr, "name") and expr.name in groupby_series_columns:
+                # Create condition for this column using the factory function
+                condition = condition_factory(expr.name, expr)
+
+                # Create CASE expression: condition true -> original, else "Others"
+                case_expr = sa.case(
+                    [(condition, expr)], else_=sa.literal_column("'Others'")
```
betodealmeida: Instead of using `literal_column` and passing the single quotes manually (which might not be universal across all DBs), it's better to:

```python
other = sa.literal("Others")
```

Same for line 1311. When compiled it should do the right thing:

```python
>>> str(sa.literal("Others").compile(compile_kwargs={"literal_binds": True}))
"'Others'"
```

betodealmeida (approved):
This is awesome, love the feature and it's something we really needed.

Let a minor change that is needed, but other than that it looks great!

superset/models/helpers.py
mistercrunch: oh nice, I had missed that somehow, found another instance on line 1311

PR #34315, opened 2025-07-25
fix: Charts list is displaying empty dataset names when there's no schema

michael-s-molina:
### SUMMARY
https://github.com/apache/superset/pull/29944 changed the logic to display only the dataset name in the charts list but didn't account for the fact that datasets can have no schemas as defined in [slice.py#L188](https://github.com/apache/superset/blob/9c771fb2baf78488f85dd3acad5af3538fb46077/superset/models/slice.py#L188). This PR fixes the issue where the names of the datasets were empty for datasets with no schemas.

### TESTING INSTRUCTIONS
Check that dataset names are displayed for datasets with or without schemas.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas (approved):
LGTM!

PR #34319, opened 2025-07-25
feat: Enable drilling in embedded

Vitor-Avila:
### SUMMARY
This PR exposes drilling capabilities to embedded users. Highlights:

**New API endpoint**
A new API endpoint was created for drilling: `/api/v1/dataset/{{pk}}/drill_info/`. Previously, the drilling context (columns to drill, dataset info to display in the `useDatasetMetadataBar` etc) were loaded from `/api/v1/dataset/{{pk}}`. Exposing this entire endpoint to embedded users is not ideal, as it exposes a lot of metadata.

**Server-side validations**
* Previously, the backend would return all dataset columns, and then the client would filter out columns that have `groupby=False`. This would allow embedded users to find out names of other columns that should not be exposed, so the list of columns is now filtered in the backend. 
* Specifically for embedded users, there's also server-side validation to prevent a user to manipulate the API call and drill by a column that has `groupby=False`.
* Implemented additional validation to make sure that embedded users can only drill datasets that are associated with a dashboard they have access to.

**Bonus**
* This PR also **_partially_** fixes drilling for `DASHBOARD_RBAC`. Full fix might require migrating `/datasource/samples` to `/api/v1/`.
* Improves performance by reducing repetitive calls. The call to load the drill context was moved up to `ChartContextMenu` and it's passed to `DrillDetail`, `DrillBy` and `useDatasetMetadataBar`.
* The new API endpoint relies on FAB's `select_columns`, to filter the returned columns at the SQL query level (not at the response level) which also improves performance.

**Embedded experience**

https://github.com/user-attachments/assets/d858d6d9-ebbd-45a0-b977-8cfc28de03d3

Highlights:
- API response just includes the list of columns and dataset ID.
- The metadata bar is not exposed to avoid sharing additional context about the underlying data.

**Non-embedded experience**

https://github.com/user-attachments/assets/23948631-1434-4886-b517-4a6edc1fd627

Highlights:
- New endpoint to expose drill-related info.
- Less API calls to load data.

### TESTING INSTRUCTIONS
Test coverage added. For manual testing:
1. Make sure the embedded role has the required permissions to perform drilling operations.
2. Load a dashboard in embedded mode.
3. Validate the right-click menu on a chart exposes the drilling options.
4. Confirm the dialogs don't expose additional information or not useful buttons.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -32,6 +32,7 @@ assists people when migrating to a new version.
 - [32317](https://github.com/apache/superset/pull/32317) The horizontal filter bar feature is now out of testing/beta development and its feature flag `HORIZONTAL_FILTER_BAR` has been removed.
 - [31590](https://github.com/apache/superset/pull/31590) Marks the begining of intricate work around supporting dynamic Theming, and breaks support for [THEME_OVERRIDES](https://github.com/apache/superset/blob/732de4ac7fae88e29b7f123b6cbb2d7cd411b0e4/superset/config.py#L671) in favor of a new theming system based on AntD V5. Likely this will be in disrepair until settling over the 5.x lifecycle.
 - [32432](https://github.com/apache/superset/pull/31260) Moves the List Roles FAB view to the frontend and requires `FAB_ADD_SECURITY_API` to be enabled in the configuration and `superset init` to be executed.
+- [34319](https://github.com/apache/superset/pull/34319) Drill to Detail and Drill By is now supported in Embedded mode, and also with the `DASHBOARD_RBAC` FF. If you don't want to expose these features in Embedded / `DASHBOARD_RBAC`, make sure the roles used for Embedded / `DASHBOARD_RBAC`don't have the required permissions to perform D2D actions.
 
 ## 5.0.0
 
--- a/superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
+++ b/superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
@@ -24,6 +24,7 @@ import {
   useCallback,
   useEffect,
   useImperativeHandle,
+  useMemo,
   useState,
 } from 'react';
 import ReactDOM from 'react-dom';
@@ -184,23 +185,6 @@ const ChartContextMenu = (
         const { json } = response;
         const { result } = json;
 
-        // Filter columns for drill-by if needed
-        if (showDrillBy && result.columns) {
-          const filteredColumns = ensureIsArray(result.columns).filter(
-            column =>
-              // If using extensions, also filter by column.groupby since extensions might not do this
-              (!loadDrillByOptions || column.groupby) &&
-              !ensureIsArray(
-                formData[filters?.drillBy?.groupbyFieldName ?? ''],
-              ).includes(column.column_name) &&
-              column.column_name !== formData.x_axis &&
-              ensureIsArray(additionalConfig?.drillBy?.excludedColumns)?.every(
-                excludedCol => excludedCol.column_name !== column.column_name,
-              ),
-          );
-          result.columns = filteredColumns;
-        }
-
         setDataset(result);
       } catch (error) {
         logging.error('Failed to load dataset:', error);
@@ -216,8 +200,37 @@ const ChartContextMenu = (
     showDrillBy,
     showDrillToDetail,
     formData.datasource,
+    loadDrillByOptions,
+    dashboardId,
+  ]);
+
+  // Compute filtered dataset with drillable_columns whenever dataset or filters change
+  const filteredDataset = useMemo(() => {
+    if (!dataset || !showDrillBy) return dataset;
+
+    const filteredColumns = ensureIsArray(dataset.columns).filter(
+      column =>
+        // If using extensions, also filter by column.groupby since extensions might not do this
+        (!loadDrillByOptions || column.groupby) &&
+        !ensureIsArray(
+          formData[filters?.drillBy?.groupbyFieldName ?? ''],
+        ).includes(column.column_name) &&
+        column.column_name !== formData.x_axis &&
+        ensureIsArray(additionalConfig?.drillBy?.excludedColumns)?.every(
+          excludedCol => excludedCol.column_name !== column.column_name,
+        ),
+    );
+
+    return {
+      ...dataset,
+      drillable_columns: filteredColumns,
+    };
+  }, [
     dataset,
-    filters?.drillBy,
+    showDrillBy,
+    filters?.drillBy?.groupbyFieldName,
+    formData.x_axis,
+    formData[filters?.drillBy?.groupbyFieldName ?? ''],
     additionalConfig?.drillBy?.excludedColumns,
     loadDrillByOptions,
   ]);
@@ -325,7 +338,7 @@ const ChartContextMenu = (
         onSelection={onSelection}
         submenuIndex={showCrossFilters ? 2 : 1}
         setShowModal={setDrillModalIsOpen}
-        dataset={dataset}
+        dataset={filteredDataset}
         isLoadingDataset={isLoadingDataset}
         {...(additionalConfig?.drillToDetail || {})}
       />,
@@ -350,7 +363,7 @@ const ChartContextMenu = (
         open={openKeys.includes('drill-by-submenu')}
         key="drill-by-submenu"
         onDrillBy={handleDrillBy}
-        dataset={dataset}
+        dataset={filteredDataset}
         isLoadingDataset={isLoadingDataset}
         {...(additionalConfig?.drillBy || {})}
       />,
@@ -434,7 +447,7 @@ const ChartContextMenu = (
           onHideModal={() => {
             setDrillModalIsOpen(false);
           }}
-          dataset={dataset}
+          dataset={filteredDataset}
         />
       )}
       {showDrillByModal && drillByColumn && dataset && filters?.drillBy && (
@@ -443,7 +456,7 @@ const ChartContextMenu = (
           drillByConfig={filters?.drillBy}
           formData={formData}
           onHideModal={handleCloseDrillByModal}
-          dataset={{ ...dataset!, verbose_map: verboseMap }}
+          dataset={{ ...filteredDataset!, verbose_map: verboseMap }}
           canDownload={canDownload}
         />
       )}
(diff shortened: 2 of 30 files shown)


michael-s-molina:
@Vitor-Avila Do we really need a new feature flag? Should this be a dashboard property such cross filters?

Vitor-Avila:
> @Vitor-Avila Do we really need a new feature flag? Should this be a dashboard property such cross filters?

Hmm open to suggestions here. We could potentially fully remove the FF and rely solely on validating if the role has the required perms to D2D. Would you say that's better?

michael-s-molina:
> Hmm open to suggestions here. We could potentially fully remove the FF and rely solely on validating if the role has the required perms to D2D. Would you say that's better?

Definitely. Every time we introduce a new feature flag we increase the complexity and maintainability of the codebase. In this case, permissions seem the natural way. @rusackas pointed out that we could also have readonly embedded dashboards, which in this case a configuration like we have for cross-filtering would be more appropriate. Either way works for me and it's better than adding a new feature flag.

Vitor-Avila:
> > Hmm open to suggestions here. We could potentially fully remove the FF and rely solely on validating if the role has the required perms to D2D. Would you say that's better?
> 
> Definitely. Every time we introduce a new feature flag we increase the complexity and maintainability of the codebase. In this case, permissions seem the natural way. @rusackas pointed out that we could also have readonly embedded dashboards, which in this case a configuration like we have for cross-filtering would be more appropriate. Either way works for me and it's better than adding a new feature flag.

Awesome -- I'll work on moving to perm-based validation only and remove the FF. I would prefer on that because then we can potentially scope it by user, as opposed to by dashboard. Users can decide what's the desired perms for their embedded users, and if they need it dynamically they could have two roles and dynamically change the role assigned. In the future, if drilling can be disabled at the dashboard level, we can then respect that too in the embedded context (like a dashboard-level override) that could potentially be controlled via the embedded SDK too.

Vitor-Avila:
@michael-s-molina I've changed the approach here to remove the FF. Could you please take a look? 🙏

superset-frontend/src/components/Chart/DrillBy/DrillByMenuItems.tsx
```diff
@@ -151,56 +123,6 @@ export const DrillByMenuItems = ({
...
-    }
-  }, [
-    addDangerToast,
-    drillByConfig?.groupbyFieldName,
-    excludedColumns,
-    formData,
-    handlesDimensionContextMenu,
-    hasDrillBy,
-  ]);
-
```
Vitor-Avila, line 203: Logic moved to `ChartContextMenu`

superset-frontend/src/features/datasets/metadataBar/useDatasetMetadataBar.tsx
```diff
@@ -16,49 +16,31 @@
...
-      })
-        .then(({ json: { result } }) => {
-          setResult(result);
-          setStatus(ResourceStatus.Complete);
-        })
-        .catch(() => {
-          setStatus(ResourceStatus.Error);
-        });
-    }
-  }, [datasetId, datasetProps]);
```
Vitor-Avila, line 57: Dataset fetched in `ChartContextMenu` and passed through

superset/config.py
```diff
@@ -279,10 +279,11 @@ def _try_json_readsha(filepath: str, length: int) -> str | None:
 
 # Add endpoints that need to be exempt from CSRF protection
 WTF_CSRF_EXEMPT_LIST = [
-    "superset.views.core.log",
-    "superset.views.core.explore_json",
     "superset.charts.data.api.data",
     "superset.dashboards.api.cache_dashboard_screenshot",
+    "superset.views.core.explore_json",
+    "superset.views.core.log",
+    "superset.views.datasource.views.samples",
```
Vitor-Avila, line 286: Just really adding `"superset.views.datasource.views.samples"`. Moved the other two to keep the list alphabetically sorted

tests/integration_tests/base_tests.py
```diff
@@ -180,33 +181,39 @@ def temporary_user(
...
+        pvms = []
+
         if clone_user:
-            temp_user.roles = clone_user.roles
             temp_user.first_name = clone_user.first_name
             temp_user.last_name = clone_user.last_name
             temp_user.password = clone_user.password
+            if clone_user.roles:
+                for role in clone_user.roles:
+                    pvms.extend(role.permissions)
```
Vitor-Avila, line 192: Moving away from doing `temp_user.roles = clone_user.roles` because now we support `pvms_to_remove` and we don't want to remove pvms from the copied role.

tests/integration_tests/base_tests.py
```diff
@@ -180,33 +181,39 @@ def temporary_user(
...
         else:
             temp_user.first_name = temp_user.last_name = username
 
-        if clone_user:
-            temp_user.roles = clone_user.roles
-
```
Vitor-Avila, line 193: this was redundant here, unsure if there was a purpose but tests still work without it

UPDATING.md
```diff
@@ -32,6 +32,7 @@ assists people when migrating to a new version.
 - [32317](https://github.com/apache/superset/pull/32317) The horizontal filter bar feature is now out of testing/beta development and its feature flag `HORIZONTAL_FILTER_BAR` has been removed.
 - [31590](https://github.com/apache/superset/pull/31590) Marks the begining of intricate work around supporting dynamic Theming, and breaks support for [THEME_OVERRIDES](https://github.com/apache/superset/blob/732de4ac7fae88e29b7f123b6cbb2d7cd411b0e4/superset/config.py#L671) in favor of a new theming system based on AntD V5. Likely this will be in disrepair until settling over the 5.x lifecycle.
 - [32432](https://github.com/apache/superset/pull/31260) Moves the List Roles FAB view to the frontend and requires `FAB_ADD_SECURITY_API` to be enabled in the configuration and `superset init` to be executed.
+- [34319](https://github.com/apache/superset/pull/34319) Drill to Detail and Drill By is now supported in Embedded mode, and also with the `DASHBOARD_RBAC` FF. If you don't want to expose these features in Embedded / `DASHBOARD_RBAC`, make sure the roles used for Embedded / `DASHBOARD_RBAC`don't have the required permissions to perform D2D actions.
```
eschutho, line 35: Can we highlight what the roles are to turn this on/off?

superset/security/manager.py
```diff
@@ -570,6 +571,32 @@ def can_access_datasource(self, datasource: "BaseDatasource") -> bool:
 
         return True
 
+    def can_drill_dataset_via_dashboard_access(
+        self, dataset: "BaseDatasource", dashboard: "Dashboard"
+    ) -> bool:
+        """
+        Return True if an embedded user or DASHBOARD_RBAC user can drill a dataset.
```
eschutho, line 578: @Vitor-Avila since drilling exposes more data, is there a way to disable this feature for embedded users if they don't want to turn it on?

Vitor-Avila: @eschutho initially I had a FF to gate this, but was advised against here : https://github.com/apache/superset/pull/34319#issuecomment-3119476516

This will still be controlled by regular role perm. If the embedded role has `can_drill_info` on `Dataset` they'll have access to this endpoint, otherwise they won't have it. Then the UI also checks for the regular permissions to show (or not) the **Drill by** and **Drill to Detail** options.

UPDATING.md
Vitor-Avila: @eschutho I think these are configureable by deployment, so unsure if we could highlight it. IIRC the embedded role is set in `superset_config.py` via the `GUEST_ROLE_NAME` config, then for `DASHBOARD_RBAC` access users can use any role they had created in the Superset UI.
Let me know if you want me to rephrase this in a better way.

superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
```diff
@@ -151,6 +166,75 @@ const ChartContextMenu = (
...
+        logging.error('Failed to load dataset:', error);
+        supersetGetCache.delete(`/api/v1/dataset/${datasetId}/drill_info/`);
+      } finally {
+        setIsLoadingDataset(false);
+      }
+    }
+
+    fetchDataset();
+  }, [
+    visible,
```
betodealmeida, line 199: Don't we need to add `dataset` here as well?

superset/security/manager.py
```diff
@@ -2401,6 +2428,32 @@ def raise_for_access(  # noqa: C901
                             and slc in dashboard_.slices
                             and slc.datasource == datasource
                         )
+                        or (
```
betodealmeida: Oh my, what is this monster expression stating on line 2379?!? 😱

We should break this down into methods that are self-explanatotry, like:

```python
return (
    self.is_d2d_allowed(form_data) or
    self.is_dashbaord_rbac(form_data) or
    ...
)
```

(Not saying _you_ should do it, but you could start by moving the code you're adding into a method and call that instead here.)

betodealmeida (approved):
This is such a great PR, from the summary to the tests, passing through the code! Approved, but left a few minor comments.

superset/datasets/schemas.py
```diff
@@ -19,14 +19,29 @@
...
 from superset.utils import json
 
 get_delete_ids_schema = {"type": "array", "items": {"type": "integer"}}
 get_export_ids_schema = {"type": "array", "items": {"type": "integer"}}
+get_drill_info_schema = {
+    "type": "object",
+    "properties": {
+        "dashboard_id": {"type": "integer"},
+    },
+}
```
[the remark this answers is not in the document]
Vitor-Avila: The parameter is only required for limited roles -- it works for admins or users with proper dataset acces without this param.

superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
Vitor-Avila: I've thought so too and checked with Claude and it advised not to, because this block also does `setDataset(result)` so if `dataset` was a dependency it would re-render one more time as the `dataset` value would change with `setDataset(result)`.

From my manual tests it's working properly.

superset/security/manager.py
Vitor-Avila: Great idea! Will do with this new block and then open a dedicated follow up later

PR #34321, opened 2025-07-25
fix(theming): Superset default theme Fallback

gabotorresruiz:
### SUMMARY

This PR fixes a critical issue where users providing partial Bootstrap theme configurations (e.g., only THEME_DEFAULT defined with empty THEME_DARK) couldn't switch to dark mode (nothing got changed since it wasn't a `THEME_DARK` been set. The solution was to implement a fallback logic that uses Superset's built-in themes with proper Ant Design algorithms when Bootstrap themes are incomplete or missing.

Also I've replaced the `THEME_ENABLE_DARK_THEME_SWITCH` feature flag to start using the `THEME_SETTINGS` approach.

  Changes:
  - Replaced `THEME_ENABLE_DARK_THEME_SWITCH` feature flag with `THEME_SETTINGS.allowSwitching`
  - Updated `RightMenu.tsx` to use theme context's `canSetThemeMode()` instead of feature flag checks
  - Added `canSetThemeMode()` method to theme context interface

### TESTING INSTRUCTIONS

1. Go to `superset/config.py` and only define a `THEME_DEFAULT`.
E.g.:
```
THEME_DEFAULT: Theme = {"token": {"brandLogoUrl": "https://picsum.photos/200"}}
```
2. Since the default for `THEME_SETTINGS.allowSwitching` is `true` if no theme settings are defined
3. Reload the UI and change the theme mode to dark mode
4. You should see how the UI defaults to the superset dark mode theme since we didn't pass a `THEME_DARK` from configs

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/theme/types.ts
+++ b/superset-frontend/packages/superset-ui-core/src/theme/types.ts
@@ -419,4 +419,5 @@ export interface ThemeContextType {
   setTheme: (config: AnyThemeConfig) => void;
   setThemeMode: (newMode: ThemeMode) => void;
   resetTheme: () => void;
+  canSetThemeMode: () => boolean;
 }
--- a/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
+++ b/superset-frontend/packages/superset-ui-core/src/utils/featureFlags.ts
@@ -53,7 +53,6 @@ export enum FeatureFlag {
   SqlValidatorsByEngine = 'SQL_VALIDATORS_BY_ENGINE',
   SshTunneling = 'SSH_TUNNELING',
   TaggingSystem = 'TAGGING_SYSTEM',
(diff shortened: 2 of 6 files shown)


mistercrunch:
If you look on `master` is it fixed? I think I might have handled that in a previous PR (unsure)

gabotorresruiz:
Hey @mistercrunch I think this is 80% already fixed by you, but I would love to fix this conflicts and add the fallback logic that I did here just as a defensive code

mistercrunch:
Might be easier to start fresh than dealing with the merge conflicts here (?) Whatever works best for you.

mistercrunch:
Closing since logic changed quite a bit in this area. We should check whether there's still an issue, and cook a fix on top of `master` if that's the case.

PR #34322, opened 2025-07-25
feat(deckgl): add selected cross-filter indication

DamianPendrak:
### SUMMARY
Added a selected cross-filter indication and improved cross-filters to include all aggregated points, such as in the Grid chart.

Screenshot 2025-07-25 at 21 46 53
Screenshot 2025-07-25 at 21 47 05

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Arc/Arc.tsx
+++ b/superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Arc/Arc.tsx
@@ -18,6 +18,7 @@
  */
 import { ArcLayer } from '@deck.gl/layers';
 import { JsonObject, QueryFormData, t } from '@superset-ui/core';
+import { Color } from '@deck.gl/core';
 import { COLOR_SCHEME_TYPES } from '../../utilities/utils';
 import { commonLayerProps } from '../common';
 import { GetLayerType, createCategoricalDeckGLComponent } from '../../factory';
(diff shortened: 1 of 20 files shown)


superset-frontend/plugins/legacy-preset-chart-deckgl/src/layers/Arc/Arc.tsx
```diff
@@ -98,7 +99,50 @@ export const getLayer: GetLayerType<ArcLayer> = function ({
...
+      targetPosition &&
+      d.sourcePosition[0] === sourcePosition[0] &&
+      d.sourcePosition[1] === sourcePosition[1] &&
+      d.targetPosition[0] === targetPosition[0] &&
+      d.targetPosition[1] === targetPosition[1]
+    ) {
+      return [255, 0, 0, 255] as Color;
+    }
+
+    return [0, 0, 0, 0] as Color;
```
kgabryje: We use [255, 0, 0, 255] and [0, 0, 0, 0] a lot in this PR. Could we replace them with a const? maybe sth like `const redColorArray` and `transparentColorArray`?

kgabryje:
Works very nice! 
1 question - I don't know the requirements so it might be correct, but why do Scatterplot and Screen grid create a range filter (from LON/LAT to LON/LAT), while Grid and Hexagon create an equality filter (LON/LAT == clicked point)?

image

DamianPendrak:
> Works very nice! 1 question - I don't know the requirements so it might be correct, but why do Scatterplot and Screen grid create a range filter (from LON/LAT to LON/LAT), while Grid and Hexagon create an equality filter (LON/LAT == clicked point)?
> 
> image

That's the difference in aggregation - the Grid and Hexagon are using CPU aggregation and Scatterplot and Screengrid are using GPU aggregation. With GPU aggregation the picking event doesn't return the points that are aggregated in the clicked cell, so the area filter is what we came up with

kgabryje (approved):
LGTM

PR #34324, opened 2025-07-25
feat: introducing a docker-compose-light.yml for lighter development

mistercrunch:
## Context

  When working with AI on multiple projects, sometimes I need to run several Superset instances at once. The full docker-compose setup is overkill when you don't need Redis, Celery workers, nginx, etc. This lightweight config includes only the
  essentials: database, app server, and frontend dev server.

AFAIC we could **make this the default** (switcharooski the files, making `docker-compose-light.yml` the main file and current main docker-compose something like `docker-compose-fullstack.yml`). It's lighter on resources, faster to boot up, an how often do you need to develop / test the celery workers!?

NOTE: well aware this isn't DRY, and dup some logic/conf from the main docker, but I think it's better to have that than trying to do inheritence in yaml files or dynamically-generated-yaml...

  ## Changes

  - New `docker-compose-light.yml` with minimal services
  - Uses in-memory caching instead of Redis
  - Parameterized `NODE_PORT` for running multiple instances
  - Database port removed from host (internal Docker network only)
  - Auto-detection in `superset_config_docker.py` for light instances
  - Updated documentation

  ## Usage

  ```bash
  # Single instance
  docker compose -f docker-compose-light.yml up

  # Multiple instances
  NODE_PORT=9002 docker compose -p superset-2 -f docker-compose-light.yml up
  NODE_PORT=9003 docker compose -p superset-3 -f docker-compose-light.yml up

  Each instance is fully isolated with separate volumes and networks.
  ```

### DIFF
--- a/docker-compose-image-tag.yml
+++ b/docker-compose-image-tag.yml
@@ -20,6 +20,9 @@
 # If you choose to use this type of deployment make sure to
 # create you own docker environment file (docker/.env) with your own
 # unique random secure passwords and SECRET_KEY.
+#
+# For verbose logging during development:
+# - Set SUPERSET_LOG_LEVEL=debug in docker/.env-local for detailed Superset logs
 # -----------------------------------------------------------------------
 x-superset-image: &superset-image apachesuperset.docker.scarf.sh/apache/superset:${TAG:-latest-dev}
 x-superset-volumes:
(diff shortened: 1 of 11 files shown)


hainenber:
May I suggest an alternative name? Like `superset-lite` since we have dark and light theme now, I'm afraid folks would believe there's a pure Superset installation with light theme :D

mistercrunch:
Not a bad idea but not sure if it's worth the dev + CI cycle (?)

geido (approved):
Works great. This is gold! Thanks

PR #34327, opened 2025-07-26
feat(helm): Allow multi-database connection support

tadeha:
### SUMMARY

This feature introduces dynamic database connection support for Superset Helm Chart, allowing users to select between different database types like PostgreSQL (default) and MySQL. Previously, the SQLALCHEMY_DATABASE_URI was hardcoded to postgresql+psycopg2. This enhancement enables the database backend to be configured directly through `values.yaml`.

The `supersetNode.connections.db_type` option controls which database type Superset will connect to.

This change is fully backward compatible:

If `supersetNode.connections.db_type` is not explicitly set in `values.yaml` (as would be the case for existing deployments), it defaults to "postgresql", preserving current functionality.

Validation is included to prevent unsupported database_type values from being deployed.

### TESTING INSTRUCTIONS

1.  **To test PostgreSQL (Default Behavior):**
    * Ensure `supersetNode.connections.db_type` is either commented out or set to `"postgresql"` in `values.yaml`.
    * Install/Upgrade Helm chart: `helm upgrade --install superset .`
    * Verify Superset successfully connects to PostgreSQL.

2.  **To test MySQL:**
    * **Prepare your Superset Image:** Build a custom Superset Docker image that includes the `mysqlclient` Python package (e.g., by adding `RUN pip install mysqlclient` after installing build dependencies like `default-libmysqlclient-dev`). Update your `values.yaml` to use this new image.
    * In `values.yaml`, set `supersetNode.connections.db_type: "mysql"`.
    * Configure your MySQL connection details:
        * Set `supersetNode.connections.db_host` to your MySQL server's host.
        * Set `supersetNode.connections.db_user`, `db_pass`, `db_name`, `db_port` for your MySQL database.
    * Install/Upgrade Helm chart: `helm upgrade --install superset .`
    * Verify Superset successfully connects to MySQL.

3.  **To test Validation (Unsupported Type):**
    * In `values.yaml`, set `supersetNode.connections.db_type: "unsupported_db_type"`.
    * Attempt to install/upgrade Helm chart: `helm upgrade --install superset .`
    * Verify that the Helm command **fails** with an error message indicating an unsupported `database_type`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas (approved):

helm/superset/templates/_helpers.tpl
```diff
@@ -96,7 +96,13 @@ CACHE_CONFIG = {
...
 DATA_CACHE_CONFIG = CACHE_CONFIG
 
-SQLALCHEMY_DATABASE_URI = f"postgresql+psycopg2://{env('DB_USER')}:{env('DB_PASS')}@{env('DB_HOST')}:{env('DB_PORT')}/{env('DB_NAME')}"
+{{- if eq .Values.supersetNode.connections.db_type "postgresql" }}
+SQLALCHEMY_DATABASE_URI = f"postgresql+psycopg2://{os.getenv('DB_USER')}:{os.getenv('DB_PASS')}@{os.getenv('DB_HOST')}:{os.getenv('DB_PORT')}/{os.getenv('DB_NAME')}"
+{{- else if eq .Values.supersetNode.connections.db_type "mysql" }}
+SQLALCHEMY_DATABASE_URI = f"mysql+mysqldb://{os.getenv('DB_USER')}:{os.getenv('DB_PASS')}@{os.getenv('DB_HOST')}:{os.getenv('DB_PORT')}/{os.getenv('DB_NAME')}"
+{{- else }}
+{{- fail (printf "Unsupported database type: %s. Please use 'postgresql' or 'mysql'." .Values.supersetNode.connections.db_type) }}
+{{- end }}
```
betodealmeida: Wouldn't it be better to just read `os.getenv('SQLALCHEMY_DATABASE_URI') if it's present, and if not fallback to Postgres?

tadeha: Good idea. I agree it makes sense to check `SQLALCHEMY_DATABASE_URI` first for flexibility. That said, I think it’s still useful to keep the current templated fallback in place for cases where users prefer setting individual `DB_*` env vars. This way, we support both approaches. Wdyt?

tadeha: I pushed a commit with the suggested change.

tadeha:
Can we have one more look, please? @rusackas @betodealmeida

rusackas:
Pinged @betodealmeida on Slack about this, but it looks like the PR is also in need of a rebase, if you don't mind brushing it up a little :D

tadeha:
> Pinged @betodealmeida on Slack about this, but it looks like the PR is also in need of a rebase, if you don't mind brushing it up a little :D

Sure. Thank you, and Done.

tadeha:
Hey there, do we have an update on this? @rusackas

PR #34339, opened 2025-07-27
fix: prevent theme initialization errors during fresh installs

mistercrunch:
Fixes database errors that occur when Superset starts before database migrations are complete, particularly in Docker environments where the app initializes concurrently with database setup.

Changes:
- Add database readiness check before theme seeding and SQLA event registration
- Use 'dashboards' table as proxy for Superset schema being ready (instead of FAB's 'ab_user' table which can exist before Superset tables are created)
- Move database-dependent operations to dedicated method for better organization
- Remove defensive error handling - let initialization fail hard on unexpected issues

The fix prevents stacktraces during 'docker-compose up' while maintaining graceful degradation for fresh installations. Theme seeding and tagging system initialization are deferred until after database schema is confirmed ready.

Future work: Consider broader init system improvements including CLI performance optimizations and better separation of app startup vs configuration phases.

🤖 Generated with [Claude Code](https://claude.ai/code) with lots of TLC from @mistercrunch

### DIFF
--- a/superset/initialization/__init__.py
+++ b/superset/initialization/__init__.py
@@ -34,6 +34,7 @@ from flask_appbuilder.utils.base import get_safe_redirect
 from flask_babel import lazy_gettext as _, refresh
 from flask_compress import Compress
 from flask_session import Session
+from sqlalchemy import inspect
 from werkzeug.middleware.proxy_fix import ProxyFix
 
 from superset.constants import CHANGE_ME_SECRET_KEY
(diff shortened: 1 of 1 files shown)


superset/initialization/__init__.py
```diff
@@ -470,6 +471,31 @@ def init_views(self) -> None:
...
+    def _init_database_dependent_features(self) -> None:
+        """
+        Initialize features that require database tables to exist.
+        This is called during app initialization but checks table existence
+        to handle cases where the app starts before database migration.
+        """
+        inspector = inspect(db.engine)
+
+        # Check if core tables exist (use 'dashboards' as proxy for Superset tables)
+        if not inspector.has_table("dashboards"):
```
geido, line 483: Just double checking. Is it safe to assume that if the ‘dashboards’ table exists, all other required tables have also been created? Could this lead to partial schema assumptions?

mistercrunch: Yes I think so. This will run a fair amount too, so didn't want to bloat. Seems `dashboards` is a fair proxy to confirm database migrations have been executed. The theme-related could double check that the `themes` table is also there, but it's pretty edge-casy that someone would have forgotten to upgrade. Probably a good thing if it crashes too since the app needs to assume the themes table is there...

geido (approved):

mistercrunch:
🙏 thanks for the review @geido

michael-s-molina:
@betodealmeida This PR does not seem to fix https://github.com/apache/superset/issues/34403.

superset/initialization/__init__.py
michael-s-molina, line 483: @mistercrunch @betodealmeida  This will fail for existing databases that don't have the `themes` table yet as the `dashboards` table will be there but not the `themes` one. Check https://github.com/apache/superset/issues/34403.

mistercrunch: right, I wanted a section for "once database has run migration", but you're right that if you have incomplete migration you'll see a warning. Pretty sure the app won't exit but it will `logging.exception` probably ok as it tells you that something is up. If you're running a version with theme-related code, you should run your migration anyways.

michael-s-molina: @mistercrunch The problem is that you're not able to run `superset db upgrade` so you can't startup Superset because of the error. I detailed the problem in https://github.com/apache/superset/issues/34403.

PR #34351, opened 2025-07-28
fix(PivotTable): Render html in cells if allowRenderHtml is true

msyavuz:
### SUMMARY

In pivot tables if the Render columns in html is checked, it should render cells as well as headers.

Before:
image
After:
image

### TESTING INSTRUCTIONS

1. Create a pivot table chart with cleaned_sales_data
2. Add a custom sql metric with:
```
CASE 
    WHEN SUM(sales) < 1000000 
        THEN CONCAT('<span style=color:#00ff00;font-size:12px>', SUM(sales), '</span>') 
    WHEN SUM(sales) < 1100000 
        THEN CONCAT('<span style=color:#FFA500;font-size:12px>', SUM(sales), '</span>') 
    ELSE 
        CONCAT('<span style=color:#808080;font-size:12px>', SUM(sales), '</span>') 
END
```
Now cells should also render with correct styles instead of pure string.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-pivot-table/src/react-pivottable/TableRenderers.jsx
+++ b/superset-frontend/plugins/plugin-chart-pivot-table/src/react-pivottable/TableRenderers.jsx
@@ -34,6 +34,12 @@ const parseLabel = value => {
   return String(value);
 };
 
+function displayCell(value, allowRenderHtml) {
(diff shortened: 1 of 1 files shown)


rusackas (approved):
LGTM!

PR #34375, opened 2025-07-28
feat(timeshift): Add support for date range timeshifts

msyavuz:
## Summary
This PR adds support for date range timeshifts in addition to the existing relative time offsets (like "1 year ago"). Users can now specify explicit date ranges in the timeshift field (e.g., "2015-01-03 : 2015-01-04") to compare data from specific time periods.

Here's a comparison table highlighting the new flag:

  | Aspect                 | Relative Timeshift                              | Date Range Timeshift                           |
  |------------------------|-------------------------------------------------|------------------------------------------------|
  | Offset Format          | '1 year ago', '6 months ago', 'inherit'         | '2021-01-01 : 2021-12-31'                      |
  | Time Range Calculation | Shifts original date range by offset amount     | Uses exact specified date range                |
  | Query Generation       | WHERE ds >= '2021-01-01' AND ds < '2022-01-01'  | WHERE ds >= '2020-06-01' AND ds < '2021-06-01' |
  | Filter Handling        | Existing temporal filters are shifted by offset | Temporal filters replaced with offset range    |
  | Inner Bounds           | Sets inner_from_dttm and inner_to_dttm          | Clears inner bounds (None)                     |
  | Feature Flag           | None required (existing functionality)          | DATE_RANGE_TIMESHIFTS_ENABLED=True             |
  | Join Strategy          | Uses temporal join columns with time grain      | Aggregates metrics, uses cross-join            |
  | Column Names           | SUM(value)__1 year ago                          | SUM(value)__2021-01-01 : 2021-12-31            |
  | Use Case               | Compare to same period in the past              | Compare to specific historical period          |
  | Example Result         | 2022 data vs 2021 data (same months)            | 2022 data vs specific 2020 period              |

### 1. Enhanced Time Offset Processing (`processing_time_offsets`)
- Added `is_valid_date_range()` method to detect date range format timeshifts
- Separate handling logic for date range vs relative offsets
- Fixed temporal filter conflicts that were causing empty results
- Proper management of `inner_from_dttm`/`inner_to_dttm` bounds for different offset types

### 2. Improved Join Logic (`join_offset_dfs`)
- Dynamic temporal column detection using query granularity and datetime series detection
- Special aggregation logic for date range offsets to prevent duplicate rows
- Preserved original behavior for relative time offsets
- Cross-join with aggregation when no non-temporal dimensions exist

### 3. Updated Time Delta Normalization (`normalize_time_delta`)
- Enhanced to handle date range formats gracefully
- Maintains backward compatibility with existing relative time expressions

## Usage Examples

**Relative Offsets (existing functionality):**
```
Time Range: 2024-01-01 to 2024-01-31
Timeshift: "1 year ago"
→ Compares Jan 2024 data with Jan 2023 data
```

**Date Range Offsets (new functionality):**
```
Time Range: 2024-01-01 to 2024-01-31  
Timeshift: "2023-06-15 : 2023-07-15"
→ Compares Jan 2024 data with mid-June to mid-July 2023 data
```

## Testing
- ✅ Existing relative timeshift functionality preserved
- ✅ New date range timeshift functionality working
- ✅ Proper SQL query generation with correct temporal filters
- ✅ Data aggregation and joining working correctly
- ✅ No conflicts between different offset types

## Backward Compatibility
This change is fully backward compatible. All existing timeshift functionality continues to work exactly as before. The new date range format is additive and doesn't affect existing usage patterns.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Antonio-RiveroMartnez:
I like the idea behind this implementation but probably better to have this behind a feature flag? also, adding some tests would be great so we ensure nothing is breaking

superset/common/query_context_processor.py
```diff
@@ -642,15 +782,164 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
 
         if offset_dfs:
+            logger.info("Before joining offset DataFrames:")
+            logger.info(f"Main DataFrame shape: {df.shape}")
+            logger.info(f"Main DataFrame:\n{df.to_string()}")
+            for offset_key, offset_df in offset_dfs.items():
+                logger.info(f"Offset '{offset_key}' DataFrame shape: {offset_df.shape}")
+                logger.info(
+                    f"Offset '{offset_key}' DataFrame:\n{offset_df.to_string()}"
+                )
```
geido: Let's remove all these logs for readability

superset/common/query_context_processor.py
```diff
@@ -476,38 +481,56 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
-                #        {label: 'dttm_col', sqlExpression: 'dttm_col', "columnType": "BASE_AXIS" }  # noqa: E501
-                #      ],
-                #      time_offsets: ['1 year ago'],
-                #      filters: [{col: 'dttm_col', op: 'TEMPORAL_RANGE', val: '2020 : 2021'}],  # noqa: E501
-                #    }
                 original_offset = offset
-                if self.is_valid_date_range(offset):
+                is_date_range_offset = self.is_valid_date_range(offset)
+
+                logger.info(
```
Antonio-RiveroMartnez: This might be a debug log instead

superset/common/query_context_processor.py
```diff
@@ -476,38 +481,56 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
+                if is_date_range_offset and feature_flag_manager.is_feature_enabled(
+                    "DATE_RANGE_TIMESHIFTS_ENABLED"
+                ):
+                    # DATE RANGE OFFSET LOGIC (like "2015-01-03 : 2015-01-04")
                     try:
                         # Parse the specified range
                         offset_from_dttm, offset_to_dttm = (
                             get_since_until_from_time_range(time_range=offset)
                         )
+                        logger.info(
```
Antonio-RiveroMartnez: Same here

superset/common/query_context_processor.py
```diff
@@ -522,34 +545,92 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
                         offset, outer_to_dttm
                     )
 
+                    # For relative offsets, keep the original behavior
```
Antonio-RiveroMartnez: Not quite sure why altering the existing behavior, if we go this path then the FF doesn't guarantee we're not affecting the existing API

superset/common/query_context_processor.py
```diff
@@ -522,34 +545,92 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
+                    # original range
+                    query_object_clone.inner_from_dttm = query_object_clone.from_dttm
+                    query_object_clone.inner_to_dttm = query_object_clone.to_dttm
+
                 x_axis_label = get_x_axis_label(query_object.columns)
                 query_object_clone.granularity = (
                     query_object_clone.granularity or x_axis_label
                 )
+
+                logger.info(
```
Antonio-RiveroMartnez: Same here for debug

superset/common/query_context_processor.py
```diff
@@ -522,34 +545,92 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
+                    f"After setting dates - from_dttm: {query_object_clone.from_dttm}, "
+                    f"to_dttm: {query_object_clone.to_dttm}"
+                )
+                logger.info(
```
Antonio-RiveroMartnez: same

superset/common/query_context_processor.py
```diff
@@ -522,34 +545,92 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
+                        ),
+                    }
+                    query_object_clone.filter.append(new_temporal_filter)
+                    logger.info(
+                        f"Added temporal filter for date range offset: "
+                        f"{new_temporal_filter}"
+                    )
+
+            else:
+                # RELATIVE OFFSETS: Original logic but fixed
```
Antonio-RiveroMartnez: Wondering if the fix should live on a different PR (and probably be shipped before this one?) otherwise your PR is not "just adding feature behind a FF"

superset/common/query_context_processor.py
```diff
@@ -522,34 +545,92 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
+                logger.info("Handling temporal filters for relative offset")
```
Antonio-RiveroMartnez: Same here for debug

superset/common/query_context_processor.py
```diff
@@ -581,14 +689,26 @@ def processing_time_offsets(  # pylint: disable=too-many-locals,too-many-stateme
...
-            # whether hit on the cache
+
             if cache.is_loaded:
                 offset_dfs[offset] = cache.df
                 queries.append(cache.query)
                 cache_keys.append(cache_key)
                 continue
 
             query_object_clone_dct = query_object_clone.to_dict()
+
```
Antonio-RiveroMartnez, line 656: All these should be debugs or removed

superset/common/query_context_processor.py
msyavuz: I am not really altering the existing behaviour just moving it into the else block for feature flag.

superset/common/query_context_processor.py
msyavuz: Same here, I am not really altering the existing behaviour just moving it into the else block for feature flag.

michael-s-molina:
@msyavuz Could you add some screenshots demonstrating the new filters?

msyavuz:
Hey, this is a multi part implementation, frontend might be contributed later

Antonio-RiveroMartnez (commented):
Please let's add the Feature flag here: https://github.com/apache/superset/blob/master/RESOURCES/FEATURE_FLAGS.md too

Antonio-RiveroMartnez (approved):
lgtm -- comments addressed and all changes behind a feature flag so no unexpected changes to existing APIs. Thanks for this

PR #34380, opened 2025-07-28
feat: allow creating dataset without exploring

betodealmeida:
### SUMMARY

Sometimes you create a dataset and you don't want to create a chart from it. This PR adds an option to the button for that, see screenshot below.

Screenshot 2025-07-28 at 7 22 24 PM

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/datasets/AddDataset/Footer/index.tsx
+++ b/superset-frontend/src/features/datasets/AddDataset/Footer/index.tsx
@@ -17,8 +17,14 @@
  * under the License.
  */
 import { useHistory } from 'react-router-dom';
-import { Button } from '@superset-ui/core/components';
-import { t } from '@superset-ui/core';
+import {
+  Button,
+  DropdownButton,
+  Menu,
+  Flex,
+} from '@superset-ui/core/components';
+import { t, useTheme } from '@superset-ui/core';
+import { Icons } from '@superset-ui/core/components/Icons';
 import { useSingleViewResource } from 'src/views/CRUD/hooks';
 import { logEvent } from 'src/logger/actions';
 import withToasts from 'src/components/MessageToasts/withToasts';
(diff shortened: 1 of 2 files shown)


superset-frontend/src/features/datasets/AddDataset/Footer/index.tsx
```diff
@@ -100,32 +107,57 @@ function Footer({
...
       });
     }
   };
 
+  const onSaveOnly = () => {
+    onSave(false);
+  };
+
   const CREATE_DATASET_TEXT = t('Create dataset and create chart');
+  const CREATE_DATASET_ONLY_TEXT = t('Create dataset only');
```
Vitor-Avila, line 125: Not really the scope of this PR, but what do you think about?
**Create and explore dataset**
**Create dataset**
Wondering what are @yousoph and @kasiazjc's thoughts. Just to avoid "create" twice... also technically the chart is not really created until you saved it, so maybe explore is more appropriate?

Vitor-Avila (approved):
LGTM! Left a non-blocking comment

superset-frontend/src/features/datasets/AddDataset/Footer/index.tsx
betodealmeida: Yeah, I didn't want to change the original copy for now, but I'm happy to update this. :)

yousoph: That sounds good to me!

michael-s-molina:
@betodealmeida Maybe "Create dataset and chart" to make the button smaller?

Vitor-Avila:
> Maybe "Create dataset and chart" to make the button smaller?

@michael-s-molina @yousoph I created a follow up for this here: https://github.com/apache/superset/pull/34544

I opted for **Create and explore dataset** because the button does not create a chart, but happy to change if you prefer.

Issue #30330, opened 2024-09-19
Histogram do not support filter by metric


vioao:
### Bug description

I get an error when using the filter by metric in the Histogram chart.

```
Apache Druid Error
'errorClass'


This may be triggered by:
Issue 1002 - The database returned an unexpected error. 
```

### How to reproduce the bug

1. build a histogram chart
2. add a metric filter(`having count(*)`)
3. update the chart to get a view
4. You will get an error

### Screenshots/recordings

![image](https://github.com/user-attachments/assets/fe3f7d1e-008d-426e-9cb4-c75475dd3efc)

### Superset version

4.0.2

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

- [X] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [X] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [X] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

rusackas:
I was able to get this to work... seems more like a druid driver issue? Maybe @betodealmeida will have a clue.

rusackas:
Also, indeed, it seems Dosu is right about the linked discussion looking rather similar.

rusackas:
Is anyone still facing this in 4.1.2 or newer? It's been silent for >200 days and is at risk of being closed as stale.

rusackas:
OK, just went and validated this myself. This is indeed busted. 

On the latest `master` we at least see a more helpful error:

<img width="788" height="733" alt="Image" src="https://github.com/user-attachments/assets/5c62752e-62bb-49ed-a5cd-610f3fe48940" />

Indeed, if I add a "GROUP BY" it works just fine.

betodealmeida:
The problem is that the histogram chart has no aggregations. Without aggregations, you can't have a `HAVING` clause — this is not Druid specific, I was able to test reproduce this in BigQuery. Even if you select one or more dimensions it will still fail, since the aggregation is done on the client for some reason.

@mistercrunch do you know why this is done on the client and not in the SQL?

mistercrunch:
Was community contributed as is. Should really be server-side ... needs 2 passes probably to get min/max, but that's easy now in plugins, probably wasn't easy/possible at the time it was contributed. The best approach is probably to stay in the realm of super-simple SQL, like first phase min/max, second phase `CASE WHEN`

mistercrunch:
[...]

PR #34383, opened 2025-07-29
feat(charts): Enable async buildQuery support for complex chart logic

mistercrunch:
This foundational change allows chart plugins to perform asynchronous operations during the buildQuery phase while properly managing UI loading states.

Key changes:
- Move CHART_UPDATE_STARTED dispatch before buildQuery execution in chartAction.js
- Make buildV1ChartDataPayload async and await buildQuery results
- Ensure loading spinner shows during async buildQuery operations

Benefits:
- Chart plugins can now perform multi-phase queries (e.g., fetch min/max then build histogram bins)
- Plugins can make API calls to determine query structure dynamically
- Loading state is properly shown during longer-running buildQuery operations
- Error handling works correctly with async buildQuery functions

This change is backwards compatible - synchronous buildQuery functions continue to work as before. The async support opens doors for more sophisticated chart implementations like server-side histograms that require preliminary queries to determine bin ranges.

Tested with 5-second delays and error throwing to verify both loading states and error propagation work correctly.

Related to: #30330

### DIFF
--- a/superset-frontend/src/components/Chart/chartAction.js
+++ b/superset-frontend/src/components/Chart/chartAction.js
@@ -164,7 +164,7 @@ const v1ChartDataRequest = async (
   ownState,
   parseMethod,
 ) => {
-  const payload = buildV1ChartDataPayload({
+  const payload = await buildV1ChartDataPayload({
     formData,
     resultType,
     resultFormat,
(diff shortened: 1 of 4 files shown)


betodealmeida (approved):
❤️ this!

PR #34385, opened 2025-07-29
feat: run pre-query cost estimation

### LINKED ISSUE #34384: Implement a "Cost Estimation" Pre-Query Check
Since Superset enables users to run SQL queries directly against large-scale data warehouses like Google BigQuery, Snowflake, and Amazon Redshift. A significant risk in this environment is that a user, particularly one less familiar with SQL optimization, could inadvertently write a query that scans terabytes of data, incurring substantial and unexpected financial costs. Superset currently lacks a built-in "guardrail" to prevent this.

**Proposed Solution:** This feature would integrate Superset with the "dry run" or query cost estimation APIs that are provided by most major cloud data warehouses. Before executing a query from the SQL Lab or a dashboard, Superset would first send the query to the warehouse's estimation endpoint. If the estimated data to be scanned or the computed cost exceeds a user-configurable threshold, Superset would present a clear warning to the user, such as: "This query will process an estimated 5TB of data. Are you sure you wish to continue?".

**Impact:** This contribution would add a critical layer of financial governance and safety to Superset, making it a more robust and trustworthy tool for enterprise environments. It would prevent costly mistakes, encourage more efficient query writing, and provide administrators with greater control over resource usage.

betodealmeida:
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/actions/sqlLab.js
+++ b/superset-frontend/src/SqlLab/actions/sqlLab.js
@@ -97,6 +97,10 @@ export const COST_ESTIMATE_STARTED = 'COST_ESTIMATE_STARTED';
 export const COST_ESTIMATE_RETURNED = 'COST_ESTIMATE_RETURNED';
 export const COST_ESTIMATE_FAILED = 'COST_ESTIMATE_FAILED';
 
+export const COST_THRESHOLD_CHECK_STARTED = 'COST_THRESHOLD_CHECK_STARTED';
+export const COST_THRESHOLD_CHECK_RETURNED = 'COST_THRESHOLD_CHECK_RETURNED';
+export const COST_THRESHOLD_CHECK_FAILED = 'COST_THRESHOLD_CHECK_FAILED';
+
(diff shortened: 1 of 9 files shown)


mistercrunch:
There's probably a need in some cases for more intricate rules that could be budget-related (keeping tabs of what's been used this month for instance). 

Maybe there's a need for something like this in `superset/config.py` to enable more complex budgeting/gatekeeping logic:
```python
def query_budget_enforcement(user, query_cost: float, optimizer_score: float) -> Boolean:
    return True
QUERY_BUDGET_ENFORCEMENT = query_budget_enforcement
```

Then rules can be based on time of day, database load, weather, or externalize to some other service.

Though a lot of this budget-management probably best belongs in the database-provider side. But with a configurable function like this one people can do whatever. 

For message back to the user, that function could probably simply raise too.

rusackas:
Hi @betodealmeida 👋

This draft PR has been inactive for about 4.5 months and is on hold pending SIP approval. Just checking in - is the related SIP still being discussed, or has there been any progress?

Thanks!

PR #34395, opened 2025-07-30
fix(superset-ui-core): Include appRoot in endpoint of SupersetClientClass.postForm action

martyngigg:
### SUMMARY

[SupersetClientClass.request](https://github.com/apache/superset/blob/6006a21378bfaf9d851491e0901327058ae61ccc/superset-frontend/packages/superset-ui-core/src/connection/SupersetClientClass.ts#L185) uses `SupersetClientClass.getUrl` to construct the correct endpoint, including an optional `appRoot` if set.

[SupersetClientClass.postForm](https://github.com/apache/superset/blob/6006a21378bfaf9d851491e0901327058ae61ccc/superset-frontend/packages/superset-ui-core/src/connection/SupersetClientClass.ts#L118) used the `url` as passed to it and subsequently failed if `appRoot` is not empty. 

With this change `postForm` also uses `getUrl` to ensure the appropriate endpoint is created.

### TESTING INSTRUCTIONS

This can be verified using the docker compose development setup. 

Set a non-empty root:

```sh
echo SUPERSET_APP_ROOT=/prefix >> docker/.env-local
```

Bring up the light docker compose config:

```sh
docker compose -f docker-compose-light.yml up -d
```

When the frontend has finished building go to http://localhost:9001/prefix and login with "admin:admin". Before these changes the successful login would redirect to http://localhost:9001/login rather than http://localhost:9001/prefix/login.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/cypress-base/cypress/e2e/auth/login.test.ts
+++ b/superset-frontend/cypress-base/cypress/e2e/auth/login.test.ts
@@ -19,7 +19,7 @@
 import { LOGIN } from 'cypress/utils/urls';
 
 function interceptLogin() {
-  cy.intercept('POST', '/login/').as('login');
+  cy.intercept('POST', '**/login/').as('login');
 }
 
 describe('Login view', () => {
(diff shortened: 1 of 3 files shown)


rusackas (approved):
LGTM (and Claude and RooCode)

PR #33519, opened 2025-05-19
fix: add query identifier to legend items in mixed time series charts

fardin-developer:
fix(chart): improve legend clarity for mixed time series charts

Fixes: #33552  

## SUMMARY  
This PR improves the legend display in Mixed Time Series Charts when multiple queries with overlapping dimensions are used. Previously, the legend did not clearly distinguish which metric belonged to which query, leading to ambiguity in interpretation.

## Changes  
- Updated legend format to display metrics as `Metric Name (Query)`, followed by the dimension value.  
- Example format:  
  - `P75ResponseTime (Query A), EXP UI: 25.66`  
  - `QueryCount (Query B), EXP UI: 49`  
- This ensures each legend entry is clearly associated with its respective query, improving readability in cases of overlapping dimension values.

### Before  
a) Legend entries listed without clarity on their metric name and query source, causing confusion in mixed charts.

image

b) It’s not clear which metric name belongs to which query.

Screenshot 2025-05-19 at 9 04 50 PM

### After  
a) Legend entries now follow this pattern:  
**Metric Name (Query), Dimension Value: OKR Value**, clearly differentiating series per query.

Screenshot 2025-05-22 at 3 28 35 AM

b) The query name is added alongside the metric name to improve clarity and indicate which query it belongs to.

[
Screenshot 2025-05-19 at 9 03 53 PM
](url)

PR #34406, opened 2025-07-30
feat: Add configurable query identifiers for Mixed Timeseries charts

yousoph:
### SUMMARY

Adds a new configuration option "Show query identifiers" to Mixed Timeseries charts, allowing users to control whether "(Query A)" and "(Query B)" identifiers appear in tooltips.
Following the implementation of query identifiers in https://github.com/apache/superset/pull/33519, some customers provided feedback that they prefer not to see these identifiers in their charts. This PR makes the feature configurable.

Changes
Added "Show query identifiers" checkbox control in the Customize tab
Default behavior: identifiers are hidden
When enabled: shows the existing "(Query A)" and "(Query B)" behavior

image
image

### TESTING INSTRUCTIONS

- Create a mixed chart
- Check that the default behavior is that the identifiers are hidden
- Check that the control shows/hides the identifiers

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -366,7 +366,7 @@ const config: ControlPanelConfig = {
               type: 'CheckboxControl',
               label: t('Show query identifiers'),
               description: t(
-                'Add Query A and Query B identifiers to legend and tooltip to help differentiate series',
+                'Add Query A and Query B identifiers to tooltips to help differentiate series',
               ),
               default: false,
               renderTrigger: true,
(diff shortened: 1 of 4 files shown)


mistercrunch (approved):
❤️ ❤️ ❤️

superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
```diff
@@ -358,7 +358,22 @@ const config: ControlPanelConfig = {
...
+              label: t('Show query identifiers'),
+              description: t(
+                'Add Query A and Query B identifiers to tooltips to help differentiate series',
+              ),
+              default: false,
+              renderTrigger: true,
+            },
+          },
+        ],
+        ...richTooltipSection.slice(1), // Skip the tooltip header since we added our own
```
eschutho, line 376: Did claude add a tooltip header and then remove it, too?

mistercrunch: @yousoph can you confirm this works as expected? If not we should follow through with a quick PR to fix the feature

yousoph: It does function as expected but maybe this code needs to be fixed 

My guess is it did something weird with the section header since the new control I added was at the top of the "Tooltip" section

PR #34412, opened 2025-07-30
fix: prevent anonymous code in Postgres

betodealmeida:
### SUMMARY

The current logic that determines if a SQL is read-only doesn't work with anonymous blocks in Postgres.

### TESTING INSTRUCTIONS

Added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql/parse.py
+++ b/superset/sql/parse.py
@@ -654,6 +654,16 @@ class SQLStatement(BaseSQLStatement[exp.Expression]):
         if isinstance(self._parsed, exp.Command) and self._parsed.name == "ALTER":
             return True  # pragma: no cover
 
+        if (
+            self._dialect == Dialects.POSTGRES
+            and isinstance(self._parsed, exp.Command)
+            and self._parsed.name == "DO"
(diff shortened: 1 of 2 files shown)


superset/sql/parse.py
```diff
@@ -654,6 +654,16 @@ def is_mutating(self) -> bool:
         if isinstance(self._parsed, exp.Command) and self._parsed.name == "ALTER":
             return True  # pragma: no cover
 
+        if (
```
mistercrunch, line 657: should this be a super simple method in `db_engine_spec` that's a no-op in base and something different for PG?

betodealmeida: I was hoping to keep all the SQL parsing functionality inside this file, since DB engine specs already deal with too much — connecting and building queries. But yeah, if this starts to have too many special cases then we should consider reorganizing it.

mistercrunch (approved):
LGTM

PR #34413, opened 2025-07-30
fix: sqlparse issue

betodealmeida:
### SUMMARY

Fix for a potential issue when using `sqlparse`. This should be backported to 5.x, since it no longer happens in `master`.

### TESTING INSTRUCTIONS

Unit tests still pass.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -88,9 +88,13 @@ logger = logging.getLogger(__name__)
 # reference: https://sqlparse.readthedocs.io/en/stable/extending/
 lex = Lexer.get_default_instance()
 sqlparser_sql_regex = keywords.SQL_REGEX
-sqlparser_sql_regex.insert(
-    25, (r"'(?:[^'\\]|\\\\?|'')*'", sqlparse.tokens.String.Single)
-)
+# Replace the string pattern at position 25 to fix ReDOS vulnerability
+# Use two patterns: first for strings ending with backslash, then for general escaped
+# strings
+# Pattern 1: Match strings that end with a single backslash (like '\')
+sqlparser_sql_regex[25] = (r"'[^'\\]*\\'(?!\w)", sqlparse.tokens.String.Single)
+# Pattern 2: Match general strings with escape sequences
+sqlparser_sql_regex.insert(26, (r"'(?:[^'\\]|\\.|'')*'", sqlparse.tokens.String.Single))
 lex.set_SQL_REGEX(sqlparser_sql_regex)
 
 
--- a/superset/sql_parse.py
+++ b/superset/sql_parse.py
@@ -88,7 +88,9 @@ logger = logging.getLogger(__name__)
 # reference: https://sqlparse.readthedocs.io/en/stable/extending/
 lex = Lexer.get_default_instance()
 sqlparser_sql_regex = keywords.SQL_REGEX
-sqlparser_sql_regex.insert(25, (r"'(''|\\\\|\\|[^'])*'", sqlparse.tokens.String.Single))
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

mistercrunch:
Not sure if targeting the branch is the proper way or if @sadpandajoe prefers to 🍒

sadpandajoe:
Pinging @michael-s-molina as he's taking care of the patch updates for 5.0

michael-s-molina (changes requested):
According to GPT this change might break valid SQL. It would be nice to add a test case with the example below. If the same logic was applied to master we might also need to open a PR there.

Screenshot 2025-07-31 at 09 12 45

betodealmeida:
> According to GPT this change might break valid SQL.

Yeah, the problem with `sqlparse` is that we were using it to parse "Schrödinger's SQL" — SQL that was neither valid nor invalid until we tried to run it. 😂 Let me give an example, here's a unit test that broke while I was working on this issue (this test addresses https://github.com/andialbrecht/sqlparse/issues/652):

```python
def test_sqlparse_issue_652():
    stmt = sqlparse.parse(r"foo = '\' AND bar = 'baz'")[0]
    assert len(stmt.tokens) == 5
```

Is this SQL expression valid or invalid?

```sql
foo = '\' AND bar = 'baz'
```

The answer is "it depends": Postgres uses the `'` for escaping sinqle quotes, so this works. But for MySQL, this is an unterminated string, since the first single quote matches the third, and the last single quote doesn't have a match.

That being said, I think the example you gave is common enough that we should make sure it can be parsed. I'm adding a new unit test and updating the regex.

michael-s-molina:
Loved the Schrödinger reference 🤣

betodealmeida:
OK, pushed an updated that uses 2 regexes instead of one, to account for the `'\'` edge case. All original tests pass, and I have a test for:

```python
    stmt = sqlparse.parse(r"SELECT 'Don\'t stop' FROM table")[0]
    assert len(stmt.tokens) == 7
    assert str(stmt.tokens[2]) == r"'Don\'t stop'"
```

betodealmeida:
> Not sure if targeting the branch is the proper way

Yeah, I targeted the branch just to make sure the PR showed a single commit. But I don't think we should merge this to the `5.0` branch.

michael-s-molina:
Thanks for adding a test @betodealmeida!

> Yeah, I targeted the branch just to make sure the PR showed a single commit. But I don't think we should merge this to the 5.0 branch.

What's the plan then? Does this needs to be included in 5.0 or can we deliver this in 6.0 (theming)?

betodealmeida:
> What's the plan then? Does this needs to be included in 5.0 or can we deliver this in 6.0 (theming)?

Ideally we'd release this in a 5.0.1.

michael-s-molina:
Ok, more inputs from AI:
Screenshot 2025-07-31 at 16 52 06
Screenshot 2025-07-31 at 16 52 36

betodealmeida:
@michael-s-molina can you ask the AI what pattern we should use?

rusackas:
Hi @betodealmeida 👋

This PR has been inactive for about 4 months. CI is passing and it has approval from @mistercrunch, but there are changes requested by @michael-s-molina that appear to still need addressing.

Are you still planning to complete this? If so, please address the review feedback when you have time.

Thanks!

betodealmeida:
Closing this, since it's going nowhere.

PR #34416, opened 2025-07-31
feat: add runtime custom font loading via configuration

mistercrunch:
Screenshot 2025-07-30 at 5 29 58 PM
Screenshot 2025-07-30 at 5 29 12 PM

  Implements custom font loading that works with stock Docker images without requiring a rebuild.
  Users can now configure external fonts through CUSTOM_FONT_URLS in superset_config.py.

  The implementation:
  - Adds CUSTOM_FONT_URLS configuration option in config.py
  - Loads fonts via Jinja template in spa.html
  - Updates CSP automatically when fonts are configured
  - Documents the feature in the theming guide

  This allows Docker users and organizations to customize fonts without maintaining custom builds.

  Fixes: #34397


betodealmeida (approved):
Comic Sans!

PR #34434, opened 2025-07-31
feat: Add ESLint rule to enforce sentence case in button text

sadpandajoe:
### SUMMARY

Add a new ESLint rule 'sentence-case-buttons' to the i18n-strings plugin that:
- Detects title case button text like "Create Chart", "Delete Dataset"
- Only applies to button contexts (primaryButtonName, Button components, etc.)
- Suggests sentence case alternatives: "Create chart", "Delete dataset"
- Avoids false positives for page titles, menu items, and labels

This helps maintain consistency with the design system preference for sentence case in UI actions while preserving title case where appropriate in navigation and headers.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/.eslintrc.js
+++ b/superset-frontend/.eslintrc.js
@@ -403,6 +403,7 @@ module.exports = {
     'theme-colors/no-literal-colors': 'error',
     'icons/no-fa-icons-usage': 'error',
     'i18n-strings/no-template-vars': ['error', true],
+    'i18n-strings/sentence-case-buttons': 'error',
     camelcase: [
       'error',
       {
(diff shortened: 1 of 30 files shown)


rusackas:
CC @kasiazjc on for good measure. We just want to make sure we're not breaking things like "CSS" or page titles. We have these [Capitalization Guidelines](https://github.com/apache/superset/wiki/Capitalization-Guidelines), so the rule should either follow them, or we can change the rules if warranted. As long as we're consistent :D

sadpandajoe:
> CC @kasiazjc on for good measure. We just want to make sure we're not breaking things like "CSS" or page titles. We have these [Capitalization Guidelines](https://github.com/apache/superset/wiki/Capitalization-Guidelines), so the rule should either follow them, or we can change the rules if warranted. As long as we're consistent :D

We probably want to update the guidelines as i see this line here:
> Exceptions to sentence case:
Input labels, buttons and tabs are all caps

Once that is updated, i'll update my pr to use the guidelines and go from there and fix anything that is needed

eschutho (approved):

PR #34435, opened 2025-07-31
fix(pie chart): Total now positioned correctly with all Legend positions, and respects theming

rusackas:
### SUMMARY

Before, if the Legend position was Right or Left, the Total number was off center. 
Before, if you set Superset to "dark" theme, the total was always black.

Now, the Total is centered in all Label positions, and turns lighter gray in Dark Mode.

After:
image
image
image
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue: Fixes https://github.com/apache/superset/issues/26226
- [ ] Required feature flags: 
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
@@ -220,7 +220,7 @@ export default function transformProps(
         name: otherName,
         value: otherSum,
         itemStyle: {
-          color: theme.colors.grayscale.dark1,
+          color: theme.colorText,
           opacity:
             filterState.selectedValues &&
             !filterState.selectedValues.includes(otherName)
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
```diff
@@ -445,6 +445,7 @@ export default function transformProps(
             text: t('Total: %s', numberFormatter(totalValue)),
             fontSize: 16,
             fontWeight: 'bold',
+            fill: theme.colors.grayscale.dark2,
```
msyavuz: Can we use first level theme colors?

rusackas:
@msyavuz now using top-level theme tokens:

image

msyavuz (approved):
LGTM Thanks!

PR #34509, opened 2025-08-01
fix: Users can't skip column sync when saving virtual datasets

michael-s-molina:
### SUMMARY
This PR fixes an issue where users were forced to always sync the columns of virtual datasets when the SQL changed. Many times, changes to the SQL don't impact in the columns and we shouldn't pay the overhead of syncing columns which is an expensive operation. Now, users will have the option to skip the sync if they wish to do so.

This PR is target at the 5.0 branch. I'll open a PR using the master branch in a follow-up.

https://github.com/user-attachments/assets/e5f056a1-a56a-46cc-a559-0c3e24195fba

### TESTING INSTRUCTIONS
Check the video for instructions.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Datasource/DatasourceModal.tsx
+++ b/superset-frontend/src/components/Datasource/DatasourceModal.tsx
@@ -262,8 +262,8 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
               type="info"
               showIcon={false}
               message={t(`The dataset columns will be automatically synced
-              given the changes in your SQL query. If your changes don't
-              impact columns, you might want to skip this step.`)}
+              based on the changes in your SQL query. If your changes don't
+              impact the column definitions, you might want to skip this step.`)}
             />
             <Checkbox
               checked={syncColumnsRef.current}
--- a/superset-frontend/src/components/Datasource/DatasourceModal.tsx
+++ b/superset-frontend/src/components/Datasource/DatasourceModal.tsx
@@ -16,7 +16,13 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { FunctionComponent, useState, useRef } from 'react';
+import {
+  FunctionComponent,
+  useState,
+  useRef,
+  useEffect,
+  useCallback,
+} from 'react';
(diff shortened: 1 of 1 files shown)


rusackas (approved):
Love it!

superset-frontend/src/components/Datasource/DatasourceModal.tsx
```diff
@@ -228,34 +236,82 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
...
+
+  useEffect(() => {
+    if (datasource.sql !== currentDatasource.sql) {
+      syncColumnsRef.current = true;
+    }
+  }, [datasource.sql, currentDatasource.sql]);
+
   const onClickSave = () => {
-    dialog.current = modal.confirm({
+    const modalInstance = modal.confirm({
```
justinpark, line 305: If I remember correctly, the confirmation result comes asynchronously.

```suggestion
    const modalInstance = await modal.confirm({
```

justinpark: oh.  I think you followed the below example.
```
const modal = Modal.info();

modal.update({
  title: 'Updated title',
  content: 'Updated content',
});

// on 4.8.0 or above, you can pass a function to update modal
modal.update((prevConfig) => ({
  ...prevConfig,
  title: `${prevConfig.title} (New)`,
}));
```

But you should double check this function since you use `modal.confirm()`.

Here is the example of await operation with useModal
```
        <Button
          onClick={async () => {
            const confirmed = await modal.confirm(config);
            console.log('Confirmed: ', confirmed);
          }}
        >
          Confirm
        </Button>
```

michael-s-molina: The promise returns a boolean value according to the user's selection. In this case, I want to monitor the object that is returned from the confirm function:

> modal.confirm return method:
> 
> destroy: Destroy current modal
> update: Update current modal
> then: (Hooks only) Promise chain call, support await operation

PR #34510, opened 2025-08-01
feat(themes): Add UUID reference system for dynamic theme configuration

mistercrunch:
Screenshot 2025-08-01 at 1 42 15 PM

Implements a UUID-based reference system that enables Superset themes to be dynamically configured through external systems like Split.io, overcoming payload size limitations.

## Problem
Split.io has a 1KB limit for JSON payloads, but Superset theme configurations typically exceed 3KB. This prevented teams from managing themes dynamically through feature flags without deployments.

## Solution
Instead of storing full theme objects, store lightweight UUID references:
- Before: `{"algorithm": "dark", "token": {...}, ...}` (3KB+)
- After: `{"uuid": "a7f3c8e2-4d1b-4c7a-9f8e-2b5d6c8a9e1f"}` (<100 bytes)

### Backend Changes
- **ResolveAndUpsertThemeCommand**: New command that resolves UUID references to full theme configurations and upserts them as system themes
- **Enhanced bootstrap data**: Modified `get_theme_bootstrap_data()` to dynamically resolve UUID references on every page load
- **Fallback support**: Graceful degradation to safe defaults if UUID resolution fails (empty object for default theme, dark algorithm for dark theme)

### Frontend Changes
- **UUID display in Theme Modal**: Added read-only UUID field with copy-to-clipboard functionality using existing CopyToClipboard component
- **Minimal styling**: Uses Label component with monospace font, following existing patterns

### How It Works
1. Store themes in Superset's Theme CRUD system (each gets a UUID)
2. Reference themes by UUID in Split.io or other configuration systems
3. On page load, the system:
   - Detects UUID references in theme configuration
   - Resolves them to full theme definitions from the database
   - Upserts as system themes for consistency
   - Falls back to safe defaults on errors

### Testing
Comprehensive test coverage including:
- UUID resolution scenarios (found, not found, invalid JSON)
- System theme upsert behavior (create new, update existing)
- Fallback configurations for different theme types
- UUID precedence over inline configuration

#### Manual test
Changed config to ->
Screenshot 2025-08-01 at 1 56 03 PM
Validated that the content of the theme referenced is replacing the system themes as expected!

This architecture transforms Split.io's constraint into a feature, enabling truly dynamic theme management with instant updates and no deployment required.

🤖 Generated with help from [Claude Code](https://claude.ai/code)

### DIFF
--- a/superset/commands/theme/resolve.py
+++ b/superset/commands/theme/resolve.py
@@ -1,0 +1,126 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import logging
+from typing import Any
+
+from superset.commands.base import BaseCommand
+from superset.daos.theme import ThemeDAO
+from superset.extensions import db
+from superset.models.core import Theme
+from superset.utils import json
+from superset.utils.decorators import transaction
+
+logger = logging.getLogger(__name__)
+
+
+class ResolveAndUpsertThemeCommand(BaseCommand):
+    """Command to resolve theme configuration and upsert to system theme."""
+
+    def __init__(self, theme_config: dict[str, Any], theme_name: str):
+        self._theme_config = theme_config
+        self._theme_name = theme_name
+        self._fallback_config = self._get_fallback_config()
+
+    def run(self) -> dict[str, Any]:
+        """Resolve theme configuration and upsert to system theme."""
+        try:
+            self.validate()
+
+            # First resolve the theme configuration
+            resolved_config = self._resolve_theme_config()
+
+            # Then upsert to system theme
+            self._upsert_system_theme(resolved_config)
+
+            return resolved_config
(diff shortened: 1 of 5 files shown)


superset/commands/theme/resolve.py
```diff
@@ -0,0 +1,126 @@
...
+            db.session.query(Theme)
+            .filter(Theme.theme_name == self._theme_name, Theme.is_system)
+            .first()
+        )
+
+        json_data = json.dumps(theme_config)
+
+        if existing_theme:
+            existing_theme.json_data = json_data
+            logger.info(f"Updated system theme: {self._theme_name}")
```
msyavuz, line 111: Can this cause a problem when there are multiple system themes with the same theme name? I don't think there is a unique constraint on the name field.

mistercrunch: it should always upsert properly, the logic is essentially that the `(theme_name, is_system)` config should be unique. Probably good to check the `Command` that does that, but should be properly wrapped in a `@transation` decorator.

mistercrunch:
Maybe we put this on hold in favor of a pick your system themes UI feature. My analysis on it:
- add a new key in `THEME_SETTINGS`, something like `ALLOW_ADMINS_TO_SET_SYSTEM_THEMES_IN_UI = False`
- add new bool fields to the `ThemeModel` -> `is_system_default` + `is_system_dark`
- add a UI feature to "set as system default(light)", "set as system default(dark)", could be in list view or whatever makes sense here, maybe as sun/moon icon, add a REST endpoint for it. Only Admins can hit it.
- add a `Command` to set theme as system theme `set_as_system_theme_default(theme_uuid)` and `set_as_system_theme_dark(theme_uuid)`, that insures one-and-only-one theme can have the fields `is_system_default` and `is_system_dark` as True
- change the logic in bootstrap data/init to handle theme fetching based on `ALLOW_ADMINS_TO_SET_SYSTEM_THEMES_IN_UI` and falls back on the right `is_system=True` themes if/when a system UI theme is deleted.
- OR, prevent the deletion of themes where is_system_default=True or is_system_dark=True

Something like that.

mistercrunch:
How about this instead?
https://github.com/apache/superset/pull/34560

PR #34512, opened 2025-08-01
fix(sqllab): prevent strings with angle brackets from being hidden

### LINKED ISSUE #25561: Sqllab UI regession in 3.0.0 where it's not displaying certain string values
After we upgraded to 3.0.0, we have received user reporting query results not rendering in sqllab results. I have checked the api call results in network tab and can confirm the correct value has been returned from the backend database. So this seems like an issue with front end.

#### How to reproduce the bug

1. Go to Sqllab, select any database (I tested this on Presto and MySQL)
2. Run `SELECT '<abcdef:12345>' as test_str`
3. Run the query

### Expected results

String values rendered as in 2.0.0

### Actual results

String values not being rendered

#### Screenshots

Screenshot 2023-10-06 at 3 09 52 PM

Screenshot 2023-10-06 at 3 10 06 PM

### Environment

- browser type and version: Chrome latest stable version
- superset version: `superset version` 3.0.0
- python version: `python --version`3.9
- node.js version: `node -v` v20.8.0
- any feature flags active: N/A

### Checklist

rusackas:
### SUMMARY
This PR fixes an issue where strings containing angle brackets (e.g., `<abcdef:12345>`) were not displaying in SQL Lab query results.

The root cause was that the `isProbablyHTML` function was too aggressive in detecting HTML content. It would treat any string that created DOM elements when parsed as HTML, even if those elements were unknown/custom tags. This caused strings like `<abcdef:12345>` to be rendered using `dangerouslySetInnerHTML`, which browsers ignore for unknown tags, making the content disappear.

**Changes:**
- Enhanced `isProbablyHTML` to check against a whitelist of known HTML tags
- Added early return if the string doesn't match common HTML patterns
- Added comprehensive test cases for various edge cases

Fixes #25561

**Before:** Strings like `<abcdef:12345>` would not display in query results
**After:** These strings now display correctly as plain text

### TESTING INSTRUCTIONS
1. Go to SQL Lab
2. Run a query that returns strings with angle brackets:
   ```sql
   SELECT '<abcdef:12345>' as test_str
   UNION ALL
   SELECT '<foo:bar>' as test_str
   UNION ALL
   SELECT 'normal text' as test_str
   UNION ALL
   SELECT '<div>actual html</div>' as test_str
   ```
3. Verify that:
   - `<abcdef:12345>` and `<foo:bar>` display as plain text
   - `normal text` displays normally
   - `<div>actual html</div>` still renders as HTML (maintaining existing functionality)

### ADDITIONAL INFORMATION
- [x] Has associated issue: #25561
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/utils/html.test.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/utils/html.test.tsx
@@ -49,6 +49,21 @@ describe('isProbablyHTML', () => {
     const trickyText = 'a <= 10 and b > 10';
     expect(isProbablyHTML(trickyText)).toBe(false);
   });
+
+  it('should return false for strings with angle brackets that are not HTML', () => {
+    // Test case from issue #25561
+    expect(isProbablyHTML('<abcdef:12345>')).toBe(false);
+
+    // Other similar cases
+    expect(isProbablyHTML('<foo:bar>')).toBe(false);
+    expect(isProbablyHTML('<123>')).toBe(false);
+    expect(isProbablyHTML('<test@example.com>')).toBe(false);
+    expect(isProbablyHTML('<custom-element>')).toBe(false);
+
+    // Mathematical expressions
+    expect(isProbablyHTML('if x < 5 and y > 10')).toBe(false);
+    expect(isProbablyHTML('price < $100')).toBe(false);
+  });
 });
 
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

PR #34513, opened 2025-08-01
fix(charts): handle PostgreSQL INTERVAL type in bar and pie charts

### LINKED ISSUE #30052: Postgres: Charts with INTERVAL metrics fail to render
### Bug description

When using an `INTERVAL` type in postgres as a metric, both bar charts and pie charts fail to render with no obvious errors.

For a pie chart, the legend is rendered but nothing else. For a bar chart, the axes and X-axis labels are rendered, but the bars are not drawn.

No errors are reported, either via toast, in the browser console, or in the application logs.

### How to reproduce the bug

1. Write a query against postgres with an `INTERVAL` column in the response
2. Create a bar chart or pie chart from that query
3. Use the `INTERVAL` as a metric, e.g. with `MAX` aggregation
4. Press the button to render the chart

### Screenshots/recordings

screenshot_2024-08-29_14-00-05_07894c8a
screenshot_2024-08-29_14-00-49_34811ac2

### Superset version

4.0.2

### Python version

3.9

### Node version

16

### Browser

Firefox

### Additional context

Rendering the exact same chart but modifying the query to convert the `INTERVAL` into an integer number of milliseconds worked fine.

### Checklist

rusackas:
### SUMMARY
This PR fixes an issue where bar and pie charts fail to render when using PostgreSQL `INTERVAL` type as a metric.

The root cause was that PostgreSQL returns `INTERVAL` values as Python `timedelta` objects, which couldn't be properly serialized for chart rendering. When these values reached the frontend, they were either missing or in a string format that couldn't be used as numeric metrics.

**Changes:**
- Added `INTERVAL` to PostgreSQL column type mappings, mapping it to `GenericDataType.NUMERIC`
- Implemented a column type mutator that converts `timedelta` objects to total seconds (float)
- Added unit tests to verify the conversion works correctly

Fixes #30052

**Before:** 
- Pie chart: Only legend renders, no chart data
- Bar chart: Axes and X-axis labels appear, but no bars are drawn

**After:** 
- Both charts render correctly with INTERVAL values displayed as numeric seconds

### TESTING INSTRUCTIONS
1. Connect to a PostgreSQL database
2. Create a query that uses an INTERVAL column as a metric:
   ```sql
   SELECT category, MAX(duration_interval) as max_duration
   FROM your_table
   GROUP BY category
   ```
3. Create a bar or pie chart from this query
4. Set the INTERVAL column as the metric
5. Verify that:
   - The chart renders correctly
   - Values are displayed as seconds (numeric)
   - No errors appear in the console

### ADDITIONAL INFORMATION
- [x] Has associated issue: #30052
- [ ] Required feature flags:
- [x] Changes UI

___

## **CodeAnt-AI Description**
Handle PostgreSQL INTERVAL values so bar and pie charts render when used as metrics

### What Changed
- PostgreSQL INTERVAL columns are treated as numeric so charts receive numeric metrics instead of unsupported objects
- INTERVAL values returned as timedeltas are converted to total seconds; numeric inputs become floats; None becomes 0; string or other non-numeric representations are passed through
- Unit tests added to verify timedelta, numeric, None, and string handling for INTERVAL values

### Impact
`✅ Bar and pie charts render when using INTERVAL metrics`
`✅ Fewer chart rendering errors from PostgreSQL INTERVAL columns`
`✅ INTERVAL metric values appear as numeric seconds for aggregation and display`
<details>
<summary><strong>💡 Usage Guide</strong></summary>

### Checking Your Pull Request
Every time you make a pull request, our system automatically looks through it. We check for security issues, mistakes in how you're setting up your infrastructure, and common code problems. We do this to make sure your changes are solid and won't cause any trouble later.

### Talking to CodeAnt AI
Got a question or need a hand with something in your pull request? You can easily get in touch with CodeAnt AI right here. Just type the following in a comment on your pull request, and replace "Your question here" with whatever you want to ask:
<pre>
<code>@codeant-ai ask: Your question here</code>
</pre>
This lets you have a chat with CodeAnt AI about your pull request, making it easier to understand and improve your code.

#### Example
<pre>
<code>@codeant-ai ask: Can you suggest a safer alternative to storing this secret?</code>
</pre>

### Preserve Org Learnings with CodeAnt
You can record team preferences so CodeAnt AI applies them in future reviews. Reply directly to the specific CodeAnt AI suggestion (in the same thread) and replace "Your feedback here" with your input:
<pre>
<code>@codeant-ai: Your feedback here</code>
</pre>
This helps CodeAnt AI learn and adapt to your team's coding style and standards.

#### Example
<pre>
<code>@codeant-ai: Do not flag unused imports.</code>
</pre>

### Retrigger review
Ask CodeAnt AI to review the PR again, by typing:
<pre>
<code>@codeant-ai: review</code>
</pre>

### Check Your Repository Health
To analyze the health of your code repository, visit our dashboard at [https://app.codeant.ai](https://app.codeant.ai). This tool helps you identify potential issues and areas for improvement in your codebase, ensuring your repository maintains high standards of code health.

</details>


tests/unit_tests/db_engine_specs/test_postgres.py
```diff
@@ -280,3 +280,18 @@ def quote_table(table: Table, dialect: Dialect) -> str:
...
+    """
+    # Test timedelta conversion
+    td = timedelta(days=1, hours=2, minutes=30, seconds=45)
+    mutator = spec.column_type_mutators[INTERVAL]
+    assert mutator(td) == 95445.0  # Total seconds: 1*86400 + 2*3600 + 30*60 + 45
+
+    # Test that non-timedelta values pass through unchanged
+    assert mutator("not a timedelta") == "not a timedelta"
+    assert mutator(12345) == 12345
+    assert mutator(None) is None
```
giftig, line 392: This technically tests the code that was added but it's not really clear why the mutator is needed, or if it is at all. This seems to be a rarely-used feature; I only see it used to create `Decimal` values in mysql.

It also just converts it to a number of seconds, but that's not going to display in a very friendly way on the frontend; I think we'll probably want to preserve the `timedelta` and pass it to the frontend in a way it can understand that is an `INTERVAL`, so that it can be displayed as an interval but also be correctly used in aggregations etc.

I'm not sure if there's a concept of an `INTERVAL` for other engines or not. It can probably be reused if there is one, but if not it may need some frontend work as well.

rusackas:
Thank you @korbit-ai and @giftig for the thorough review! I've addressed both concerns:

## Changes made:

### 1. **Enhanced INTERVAL type handling** (addressing @korbit-ai's comment):
- The mutator now robustly handles multiple data types:
  - `timedelta` objects → converted to total seconds (primary use case from psycopg2)
  - Numeric values (int/float) → converted to float for consistency
  - `None` values → converted to 0 (safe for aggregations)
  - String representations → pass through unchanged (preserves PostgreSQL string formats for potential frontend handling)

### 2. **Added documentation and context** (addressing @giftig's comment):
- Added clear comments explaining why the mutator is needed
- The mutator normalizes PostgreSQL INTERVAL values specifically for chart rendering in bar/pie charts
- While converting to seconds isn't the most user-friendly display, it ensures numeric operations work correctly in charts
- String representations are preserved for potential future frontend formatting improvements

### 3. **Comprehensive test coverage**:
- Added tests for all supported data types
- Tests verify proper handling of edge cases
- Ensures backward compatibility while improving robustness

The implementation now handles all PostgreSQL INTERVAL formats more reliably while maintaining backward compatibility. The conversion to seconds ensures charts can perform numeric operations, while preserving string formats leaves room for future frontend improvements to display intervals more user-friendly.

All pre-commit checks pass (except pylint due to local environment issues unrelated to these changes).

rusackas:
From the bot feedback:

1. Regex specificity - ❌ Not worth changing
  The ^interval pattern is consistent with the entire codebase - there are 50+ similar patterns like ^tinyint, ^decimal, ^timestamp, etc. Changing just this one would be inconsistent.

2. Type matching - ❌ Not an issue
  The base class already uses .get() for lookup (see base.py:995), so this works correctly in production.

3. KeyError in test - ❌ Minor style
  Tests that fail with KeyError still clearly indicate the problem. Not worth changing.
  
4. NULL handling - ⚠️ The only one worth considering
  The current behavior (None→0) is fine for bar/pie charts since they need numeric values. If you want to be more conservative, you could change it to pass through None and let the chart layer decide. But this is a minor edge case that probably doesn't warrant blocking the PR. 
    
    Happy to change my opinion on item 4 if anyone is worried about it.

rusackas:
## Updates Made

I've rebased this PR on master and made an improvement to address @giftig's feedback about user-friendly interval display.

### Key Change: Milliseconds Instead of Seconds

INTERVAL values are now converted to **milliseconds** instead of seconds. This enables users to apply Superset's built-in **`DURATION`** number format for human-readable display.

**Before:** Raw numeric display like `95445` (seconds)  
**After with DURATION format:** `1d 2h 30m 45s`

### How to Use

1. Create a chart with a PostgreSQL INTERVAL metric
2. In the chart's "Customize" tab, set the **Number Format** to `DURATION`
3. The interval will display in human-readable format

### Available Duration Formats

| Format | Example Output |
|--------|----------------|
| `DURATION` | `1d 2h 30m 45s` |
| `DURATION_SUB` | `1d 2h 30m 45s 400µs` (with sub-ms) |
| `DURATION_COL` | `26:30:45` (colon notation) |

### Other Improvements

- Fixed type annotation to match MySQL pattern (`types.TypeEngine` instead of `Any`)
- Updated tests and documentation

This makes PostgreSQL INTERVAL values much more user-friendly in charts while maintaining numeric operations for aggregations.

rusackas:
## Addressing Review Feedback

I've gone through all the review comments and here's how each concern has been addressed:

### @korbit-ai - Incomplete INTERVAL type handling
✅ **Addressed** - The mutator now handles multiple formats:
- `timedelta` objects → milliseconds (primary psycopg2 case)
- `int`/`float` values → converted to milliseconds
- `None` → 0 (safe for aggregations)
- Strings → pass through unchanged

### @giftig - Display not user-friendly
✅ **Addressed** - Changed from seconds to **milliseconds**, which enables the built-in `DURATION` formatter. Users can now see `1d 2h 30m 45s` instead of raw numbers by setting Number Format to `DURATION`.

### @giftig - Missing documentation on why mutator is needed  
✅ **Addressed** - Added clear comments explaining:
- Why normalization is needed (psycopg2 returns timedelta)
- Why milliseconds (DURATION formatter compatibility)
- How string representations are handled

### @codeant-ai-for-open-source nitpicks

| Concern | Response |
|---------|----------|
| NULL handling (None→0) | Kept as-is - charts need numeric values for aggregations |
| Regex specificity (`^interval`) | Consistent with 50+ similar patterns in codebase |
| Type matching robustness | Base class uses `.get()` for lookup - not an issue |
| KeyError in test | Tests fail clearly on missing key - minor style preference |

### Additional improvements
- Fixed type annotation to match MySQL pattern (`types.TypeEngine`)
- Rebased on latest master

The PR is ready for another review. Thanks to everyone for the thorough feedback!

rusackas:
@giftig this is an ancient PR at this point, but I let Claude take a run at it. Lemme know if your concerns seem addressed... or if this is just barking up the wrong tree, I'll close it out and let nature take its course.

superset/db_engine_specs/postgres.py
```diff
@@ -489,8 +490,30 @@ class PostgresEngineSpec(BasicParametersMixin, PostgresBaseEngineSpec):
...
+        INTERVAL: lambda v: (
+            v.total_seconds() * 1000
+            if hasattr(v, "total_seconds")
+            else float(v) * 1000
+            if isinstance(v, (int, float))
+            else 0
+            if v is None
+            else v  # Pass through string representations and other types
+        ),
+    }
```
sadpandajoe, line 567: A few things in this lambda worth addressing:

**1. `None` → `0` changes NULL semantics**
NULL INTERVAL means "unknown/missing" — converting to `0` makes it indistinguishable from an actual zero-duration interval. It also means AVG aggregations on the frontend would include false zeros instead of excluding missing data. Returning `None` preserves the correct semantics.

**2. `bool` leaks through the numeric branch**
In Python, `bool` is a subclass of `int`, so `isinstance(True, (int, float))` is `True`. This means `True` → `1000.0` and `False` → `0.0`, which probably isn't intended for an INTERVAL column.

**3. String/other pass-through creates mixed-type NUMERIC columns**
Since the column type mapping above declares INTERVAL as `GenericDataType.NUMERIC`, passing through strings/lists/dicts unchanged could produce mixed-type columns that break chart rendering on the frontend. Might be safer to return `None` for anything that can't be converted to a number.

Here's a version that addresses all three:

```python
def _normalize_interval(v: Any) -> Any:
    """Convert PostgreSQL INTERVAL values to milliseconds for the DURATION formatter."""
    if v is None:
        return None
    if hasattr(v, "total_seconds"):
        return v.total_seconds() * 1000
    if isinstance(v, (int, float)) and not isinstance(v, bool):
        return float(v) * 1000
    return None  # Can't convert to numeric — treat as missing

column_type_mutators: dict[types.TypeEngine, Callable[[Any], Any]] = {
    INTERVAL: _normalize_interval,
}
```

This also makes it easier to test individual branches and add edge cases later.

sadpandajoe (commented):
Nice approach — `column_type_mutators` is the right mechanism and milliseconds is the correct target for the DURATION formatter. A few things in the lambda that could cause subtle data issues, plus some test suggestions.

tests/unit_tests/db_engine_specs/test_postgres.py
```diff
@@ -280,3 +280,34 @@ def quote_table(table: Table, dialect: Dialect) -> str:
...
+    assert mutator(None) == 0
+
+    # Test string values pass through unchanged
+    # (PostgreSQL may return string representations in some cases)
+    assert mutator("1 day 02:30:45") == "1 day 02:30:45"
+    assert mutator("P1DT2H30M45S") == "P1DT2H30M45S"  # ISO 8601 duration
+
+    # Test other types pass through unchanged
+    assert mutator([1, 2, 3]) == [1, 2, 3]
+    assert mutator({"days": 1}) == {"days": 1}
```
sadpandajoe: Good coverage of the lambda itself. A few additions that would help catch integration issues and edge cases:

**1. Integration test via `fetch_data`** (follows the pattern in `test_mysql.py:253`)

This verifies the full wiring: `cursor.description` → type lookup → mutator application.

```python
@pytest.mark.parametrize(
    "data,description,expected_result",
    [
        (
            [(timedelta(hours=1, minutes=30),)],
            [("duration", "interval")],
            [(5400000.0,)],  # 1.5 hours in ms
        ),
        (
            [(None,)],
            [("duration", "interval")],
            [(None,)],  # NULLs preserved
        ),
        (
            [(timedelta(hours=1), "label")],
            [("duration", "interval"), ("name", "varchar(50)")],
            [(3600000.0, "label")],  # Only INTERVAL column mutated
        ),
    ],
)
def test_interval_column_type_mutator(
    data: list[tuple[Any, ...]],
    description: list[tuple[str, str]],
    expected_result: list[tuple[Any, ...]],
) -> None:
    from unittest.mock import Mock

    cursor = Mock()
    cursor.description = description
    cursor.fetchall.return_value = data
    assert spec.fetch_data(cursor) == expected_result
```

**2. Column spec test** — verify INTERVAL is classified as NUMERIC:

```python
# Add to the existing test_get_column_spec parametrize list:
("INTERVAL", INTERVAL, None, GenericDataType.NUMERIC, False),
```

**3. Edge cases for the mutator:**

```python
# Zero duration
assert mutator(timedelta(0)) == 0.0

# Negative interval
assert mutator(timedelta(days=-1)) == -86400000.0

# Bool should not be treated as numeric
assert mutator(True) is None
assert mutator(False) is None
```

superset/db_engine_specs/postgres.py
rusackas: Applied in fab40b6b69. Great catches on all three issues:

1. **NULL semantics** - Now returns `None` instead of `0` to preserve correct NULL behavior
2. **bool handling** - Added explicit `not isinstance(v, bool)` check since bool is a subclass of int
3. **Mixed-type columns** - Unconvertible types now return `None` instead of passing through

Also extracted the lambda to a named `_normalize_interval` method for better testability as suggested.

Note: I skipped the `fetch_data` integration test because PostgreSQL's `get_datatype` uses psycopg2 type codes (integers) rather than string type names, making it difficult to mock without the actual psycopg2 type system. The unit tests provide good coverage of the mutator logic itself.

tests/unit_tests/db_engine_specs/test_postgres.py
rusackas: Applied in fab40b6b69. Added the suggested test improvements:

**Column spec test** - Added INTERVAL to the parametrized `test_get_column_spec` list:
```python
("INTERVAL", INTERVAL, None, GenericDataType.NUMERIC, False),
```

**Edge cases** - Added tests for:
- Zero duration: `timedelta(0)` → `0.0`
- Negative interval: `timedelta(days=-1)` → `-86400000.0`
- Bool handling: `True`/`False` → `None` (not treated as numeric)

**Integration test** - Skipped this one because PostgreSQL's `get_datatype` method uses psycopg2 type codes (integers from `binary_types`/`string_types`) rather than string type names like MySQL. This makes the cursor description format difficult to mock without the actual psycopg2 type system. The unit tests provide comprehensive coverage of the mutator behavior.

aminghadersohi (approved):
Thanks for this fix — INTERVAL columns in bar/pie charts have been a real pain point.

The core approach is correct: hooking into the existing `column_type_mutators` mechanism, converting timedelta→milliseconds, and the DURATION formatter alignment is a nice touch. Test coverage is thorough (including the subtle bool-subclasses-int edge case).

Two medium suggestions and a nit, none blocking:

**MEDIUM — Dead code branch** (~L551 in `postgres.py`)
```python
if isinstance(v, (int, float)) and not isinstance(v, bool):
    return float(v) * 1000
```
psycopg2 (and psycopg3) always returns `datetime.timedelta` for INTERVAL — never int/float. The docstring confirms this. This branch can't be reached in practice, so it adds complexity and the tests covering it create false confidence. Suggest removing it; the final `return None` already handles any non-timedelta non-None value.

**MEDIUM — `.__func__` + `# type: ignore`** (~L557)
```python
INTERVAL: _normalize_interval.__func__,  # type: ignore[attr-defined]
```
Accessing `.__func__` on a `staticmethod` descriptor in the class body is unusual and mypy flags it. The MySQL spec uses an inline lambda to avoid this. Cleanest fix: define `_normalize_interval` as a module-level function and reference it directly — no suppress needed.

**NIT — Silent `None` should log a warning**
The `return None  # Can't convert to numeric` at the end silently drops unexpected-type values. A `logger.warning()` there would help diagnose future driver surprises.

Happy to merge as-is — the functional fix is solid.

rusackas:
Thanks Amin — all three suggestions applied in b3604af4ce:

1. **Dead int/float branch removed.** You're right, psycopg2/3 always hand back `datetime.timedelta` for INTERVAL — the branch was unreachable in practice. Dropped it; the defensive `return None` at the end still covers any unexpected non-timedelta value.

2. **`.__func__` + `# type: ignore` gone.** Moved `_normalize_interval` to a module-level function (above the class). `column_type_mutators` now references it directly. Cleaner, mypy-clean, and matches the MySQL pattern's spirit (just module-level rather than inline lambda, since the body's too long for a lambda).

3. **`logger.warning` added.** The defensive return-None branch now logs `Cannot normalize PostgreSQL INTERVAL value of type %s to numeric; returning None.` so a future driver surfacing something other than timedelta gets diagnosed rather than silently dropped.

Test was tightened in the same shot — the numeric/bool assertions that exercised the removed branch now exercise the defensive return-None path instead. Same expectations, but for the right reason.

Pre-commit clean (including mypy).

PR #34514, opened 2025-08-01
fix(charts): handle MSSQL datetime values outside JavaScript's safe range

### LINKED ISSUE #27790: Datetime values are returned as 0NaN-NaN-NaN	outside of a certain range
### Bug description

Datetime values outside of the range '1938-04-24 22:13:20.003' to '2286-11-20 17:46:39.997' are returned as 0NaN-NaN-NaN when creating a chart. SQL Lab returns values correctly.
Microsoft SQL Server allows a range of 1753-01-01 through 9999-12-31 for datetime values.

Using the current latest commit 5b1d6b2850cfca3d3fbf5be0c09cb4bcebbee7e5
Microsoft SQL Server database

### How to reproduce the bug

Create the following dataset using a mssql DB:
```
SELECT CAST('1938-04-24 22:13:20.002' AS DATETIME) AS datetimeA, CAST('2286-11-20 17:46:39.998' AS DATETIME) AS datetimeB
```
Check the return values in chart creation. If you shorten the range, they will be returned normally.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

rusackas:
### SUMMARY
This PR fixes an issue where datetime values outside JavaScript's safe date range were returning "0NaN-NaN-NaN" in charts when using Microsoft SQL Server.

The root cause was that MSSQL supports a much wider datetime range (1753-9999) than JavaScript's Date object can handle (approximately 1938-2286). When dates outside this range were converted to epoch milliseconds, they would overflow JavaScript's Number type, resulting in NaN values that displayed as "0NaN-NaN-NaN" in charts.

**Changes:**
- Added constants defining JavaScript's safe date range in `superset/utils/dates.py`
- Modified `json_int_dttm_ser` to return ISO strings for dates outside the safe range
- Added comprehensive tests for edge cases
- Maintains backward compatibility for dates within the normal range

Fixes #27790

**Before:** Charts display "0NaN-NaN-NaN" for datetime values outside JavaScript's safe range
**After:** Charts display the actual datetime values correctly as ISO strings

### TESTING INSTRUCTIONS
1. Connect to a Microsoft SQL Server database
2. Create a query with datetime values outside JavaScript's safe range:
   ```sql
   SELECT 
       CAST('1938-04-24 22:13:20.002' AS DATETIME) AS early_date,
       CAST('2286-11-20 17:46:39.998' AS DATETIME) AS late_date,
       CAST('2000-01-01 12:00:00' AS DATETIME) AS normal_date
   ```
3. Create a chart using these datetime columns
4. Verify that:
   - Early and late dates display as proper datetime strings (not "0NaN-NaN-NaN")
   - Normal dates continue to work as before
   - No JavaScript console errors appear

### ADDITIONAL INFORMATION
- [x] Has associated issue: #27790
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset/utils/json.py
+++ b/superset/utils/json.py
@@ -29,7 +29,12 @@ from jsonpath_ng import parse
 from simplejson import JSONDecodeError
 
 from superset.constants import PASSWORD_MASK
-from superset.utils.dates import datetime_to_epoch, EPOCH
+from superset.utils.dates import (
+    datetime_to_epoch,
+    EPOCH,
+    JS_DATE_RANGE_MAX,
+    JS_DATE_RANGE_MIN,
+)
 
 logging.getLogger("MARKDOWN").setLevel(logging.INFO)
 logger = logging.getLogger(__name__)
@@ -155,9 +160,27 @@ def json_int_dttm_ser(obj: Any) -> Any:
     """
 
     if isinstance(obj, (datetime, pd.Timestamp)):
+        # Check if datetime is within JavaScript's safe date range
+        # If not, return ISO string instead of epoch milliseconds
+        if isinstance(obj, pd.Timestamp):
+            dttm = obj.to_pydatetime()
+        else:
+            dttm = obj
+
+        # Remove timezone info for comparison
+        dttm_no_tz = dttm.replace(tzinfo=None) if dttm.tzinfo else dttm
+
+        if dttm_no_tz < JS_DATE_RANGE_MIN or dttm_no_tz > JS_DATE_RANGE_MAX:
+            # Return ISO string for dates outside JavaScript's safe range
+            return obj.isoformat()
+
         return datetime_to_epoch(obj)
 
     if isinstance(obj, date):
+        # Check if date is within JavaScript's safe date range
+        date_as_datetime = datetime.combine(obj, datetime.min.time())
+        if date_as_datetime < JS_DATE_RANGE_MIN or date_as_datetime > JS_DATE_RANGE_MAX:
+            return obj.isoformat()
         return (obj - EPOCH.date()).total_seconds() * 1000
 
     return base_json_conv(obj)
--- a/superset/utils/dates.py
+++ b/superset/utils/dates.py
@@ -19,6 +19,11 @@ from datetime import datetime
 import pytz
 
 EPOCH = datetime(1970, 1, 1)
+# JavaScript's Date object can safely represent dates within this range
+# These are the min/max values that can be represented as milliseconds since epoch
+# without overflowing JavaScript's Number type (±2^53)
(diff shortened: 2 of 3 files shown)


superset/utils/json.py
```diff
@@ -155,9 +160,27 @@ def json_int_dttm_ser(obj: Any) -> Any:
...
+        dttm_no_tz = dttm.replace(tzinfo=None) if dttm.tzinfo else dttm
+
+        if dttm_no_tz < JS_DATE_RANGE_MIN or dttm_no_tz > JS_DATE_RANGE_MAX:
+            # Return ISO string for dates outside JavaScript's safe range
+            return obj.isoformat()
+
         return datetime_to_epoch(obj)
 
     if isinstance(obj, date):
+        # Check if date is within JavaScript's safe date range
```
mistercrunch, line 180: Oh interesting, it would have been an issue before this PR, but does `date.date` not have a tzinfo like `date.datetime`? Seem we should apply this consistently ...

superset/utils/json.py
```diff
@@ -155,9 +160,27 @@ def json_int_dttm_ser(obj: Any) -> Any:
...
+        # If not, return ISO string instead of epoch milliseconds
+        if isinstance(obj, pd.Timestamp):
+            dttm = obj.to_pydatetime()
+        else:
+            dttm = obj
+
+        # Remove timezone info for comparison
+        dttm_no_tz = dttm.replace(tzinfo=None) if dttm.tzinfo else dttm
+
+        if dttm_no_tz < JS_DATE_RANGE_MIN or dttm_no_tz > JS_DATE_RANGE_MAX:
```
mistercrunch, line 173: So this means our JSON serializer tries to convert to epoch, and will return mixed types for dates (string or int depending on whether it fits in an int?) Wondering if it's handled properly in the frontend...

mistercrunch: Anyhow we're mostly good until 2286 :/

mistercrunch:
Would have to dig deeper, but seems dangerous to introduce mixed types for date field in json across the board (?) Might need to be tested across plugins / explore/ sql lab. Feels like the right solution might be to use ISO consistently all the time? Would/could that break external plugins for instance? Almost feel like we'd need a new `api/v3/` endpoints that switch to ISO, and retrofit the plugins to use it?

superset/utils/json.py
rusackas: Here's what _le bot_ says:
image

sfirke:
Maybe I didn't understand https://github.com/apache/superset/issues/27790 but I just tried to reproduce and couldn't, so I closed that issue. This PR was linked to that issue so I'm mentioning it here.

PR #34521, opened 2025-08-02
chore: use logger on all migrations

villebro:
### SUMMARY
I noticed we're using many different variants of logging in migrations. This PR unifies the following:
- fix a few instances, where we were logging without the import
- always use a `logger` instead of `logging` directly
- replace deprecated calls, like `logger.warn()` with `logger.warning()`
- always use the `alembic.env` logger which is defined in the `env.py` file: https://github.com/apache/superset/blob/1a7a381bd574665d4694be4a7554962f8edd98ad/superset/migrations/env.py#L37

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/migrations/shared/catalogs.py
+++ b/superset/migrations/shared/catalogs.py
@@ -37,7 +37,7 @@ from superset.migrations.shared.security_converge import (
 )
 from superset.models.core import Database
 
-logger = logging.getLogger("alembic")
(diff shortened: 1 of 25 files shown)


betodealmeida (approved):
Nice!

mistercrunch:
Love it!

PR #34523, opened 2025-08-03
fix(table): hide conditional formatting color options without time comparison

### LINKED ISSUE #34141: Green for increase, Y for decrease color scheme option on the Table Chart does not work
### Bug description

Users can choose the X for increase, Y for decrease color scheme option on the Table Chart. However, it does not work.

## Reproduction
1. Create a virtual dataset with query
```sql
SELECT
  1 AS X,
  2 AS Y,
  1 - 2 AS "X-Y"
UNION ALL
SELECT
  5 AS X,
  2 AS Y,
  5 - 2 AS "X-Y"
UNION ALL
SELECT
  4 AS X,
  2 AS Y,
  4 - 2 AS "X-Y"
UNION ALL
SELECT
  1 AS X,
  5 AS Y,
  1 - 5 AS "X-Y"
UNION ALL
SELECT
  1 AS X,
  1 AS Y,
  1 - 1 AS "X-Y";
```
2. Create a table chart with this dataset
3. Go to the Customize tab, and add Conditional Formatting to the third column.
4. Choose the Green for increase, red for decrease option.

## Expected results
The "Green for increase, red for decrease" and "Red for increase, green for decrease" conditional formatting options shouldn't appear unless there is a time shift applied to the table
## Actual results
The Conditional Formatting for the X for increase, Y for decrease does not work.

### Screenshots/recordings

Image

### Superset version

master / latest-dev

### Python version

3.11

### Node version

18 or greater

### Browser

Firefox

### Additional context

_No response_

### Checklist

rusackas:
### SUMMARY
Fixes issue #34141 where conditional formatting color scheme options for "Green for increase, red for decrease" and "Red for increase, green for decrease" were showing in table charts even when no time comparison was active. These color options only work with time comparison data, so they should be hidden when time_compare is empty.

The fix modifies both table chart control panels to dynamically show/hide these color options based on whether time comparison is enabled.

**Before:** Color scheme options "Green for increase, red for decrease" and "Red for increase, green for decrease" appeared in conditional formatting dropdowns even without time comparison enabled.

**After:** These color scheme options only appear when time comparison is active (time_compare has a value).

### TESTING INSTRUCTIONS
1. Create a new table chart without time comparison enabled
2. Go to Visual formatting -> Custom conditional formatting  
3. Create a new conditional formatting rule
4. Check the color scheme dropdown - should NOT see "Green for increase, red for decrease" options
5. Enable time comparison in the chart configuration
6. Return to conditional formatting - should now see the increase/decrease color options

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #34141
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-ag-grid-table/src/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-ag-grid-table/src/controlPanel.tsx
@@ -673,16 +673,6 @@ const config: ControlPanelConfig = {
               type: 'ConditionalFormattingControl',
               renderTrigger: true,
               label: t('Custom conditional formatting'),
-              extraColorChoices: [
-                {
-                  value: ColorSchemeEnum.Green,
-                  label: t('Green for increase, red for decrease'),
-                },
-                {
-                  value: ColorSchemeEnum.Red,
-                  label: t('Red for increase, green for decrease'),
-                },
-              ],
               description: t(
                 'Apply conditional color formatting to numeric columns',
               ),
@@ -695,6 +685,23 @@ const config: ControlPanelConfig = {
                 )
                   ? (explore?.datasource as Dataset)?.verbose_map
                   : (explore?.datasource?.columns ?? {});
+
+                // Only show increase/decrease color options when time comparison is active
+                const hasTimeComparison = !isEmpty(
+                  explore?.form_data?.time_compare,
+                );
+                const extraColorChoices = hasTimeComparison
+                  ? [
+                      {
+                        value: ColorSchemeEnum.Green,
+                        label: t('Green for increase, red for decrease'),
+                      },
+                      {
+                        value: ColorSchemeEnum.Red,
+                        label: t('Red for increase, green for decrease'),
+                      },
+                    ]
+                  : [];
                 const chartStatus = chart?.chartStatus;
                 const { colnames, coltypes } =
                   chart?.queriesResponse?.[0] ?? {};
@@ -725,6 +732,7 @@ const config: ControlPanelConfig = {
                   removeIrrelevantConditions: chartStatus === 'success',
                   columnOptions,
                   verboseMap,
+                  extraColorChoices,
                 };
               },
             },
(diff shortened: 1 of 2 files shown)


superset-frontend/plugins/plugin-chart-ag-grid-table/src/controlPanel.tsx
```diff
@@ -695,6 +685,23 @@ const config: ControlPanelConfig = {
                 )
                   ? (explore?.datasource as Dataset)?.verbose_map
                   : (explore?.datasource?.columns ?? {});
+
+                // Only show increase/decrease color options when time comparison is active
+                const hasTimeComparison = !isEmpty(
```
Antonio-RiveroMartnez, line 690: Should this check for `explore?.controls?.time_compare?.value` like [columnOptions](https://github.com/apache/superset/blob/2029fd013105611e965015268c4f212a9441e59e/superset-frontend/plugins/plugin-chart-ag-grid-table/src/controlPanel.tsx#L722) does? and since both are checking probably just use the same const for both conditionals?

Antonio-RiveroMartnez: nit: I know is not fault of this PR, but I see we are telling whether it has time comparison or not in several parts of these files and using each its own approach or duplicating, probably good to have a helper method we can invoke and reuse?

SBIN2010:
@rusackas @Antonio-RiveroMartnez 
 Maybe it's better to do this: 
const hasTimeComparison = explore?.controls?.time_compare?.value;
and replace all 
explore?.controls?.time_compare?.value
with hasTimeComparison

jesperct:
Hi @rusackas, I noticed that this PR has been open for a while. Thank you for the work so far. I’m currently looking at a ticket related to this issue and would like to help move it forward. Would you be okay with me continuing on top of your branch and pushing the remaining changes? Happy to coordinate so we don’t overlap.

rusackas:
Closing in favor of #37362 which was merged with the same fix. Thanks @JCelento for picking this up!

PR #34528, opened 2025-08-03
fix(explore): Fix missing await for async buildV1ChartDataPayload calls

mistercrunch:
PR #34383 made buildV1ChartDataPayload async to support complex chart logic, but missed updating two critical call sites to await the Promise:

1. In saveModalActions.ts - getSlicePayload was calling buildV1ChartDataPayload without await, causing query_context to be saved as "{}" (stringified Promise)

2. In ExploreChartPanel/index.jsx - updateQueryContext was also missing await

This caused all charts created after PR #34383 to have invalid query_context, leading to "Error loading chart datasources" warnings in dashboards.

Fixes:
- Made getSlicePayload async and added await for buildV1ChartDataPayload
- Updated all callers of getSlicePayload to await it
- Added await in ExploreChartPanel's updateQueryContext
- Updated all related tests to handle async behavior

This ensures charts are saved with proper query_context instead of a stringified Promise.

Fixes regression from #34383

### DIFF
--- a/superset-frontend/src/explore/actions/saveModalActions.test.ts
+++ b/superset-frontend/src/explore/actions/saveModalActions.test.ts
@@ -491,8 +491,8 @@ describe('getSlicePayload', () => {
     dashboards: [],
   };
 
-  test('should return the correct payload when no adhoc_filters are present in formDataWithNativeFilters', () => {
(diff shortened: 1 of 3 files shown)


rusackas (approved):
LGTM!

Issue #33105, opened 2025-04-11
Heatmap renders of 0 and False values as <NULL>


fzhengTT:
**EDIT: This was a two-part issue. The original description remains but title has been changed. See issue XXX for second part of issue related to y-axis sorting.**

### Bug description

Observed on Superset 5.0.0, the deployment relies on a [fork](https://github.com/tenstorrent/apache-superset-tt) of Superset OSS version [5.0.0rc1](https://github.com/apache/superset/releases/tag/5.0.0rc1). It's deployed in a EC2 instance using docker compose.

The Heatmap chart seems to sort the Y-axis through some unknown logic.

A similar issue was raised before for Heatmap chart at https://github.com/apache/superset/issues/31318 and was stated to be fixed in PR https://github.com/apache/superset/pull/31752. 
Another similar issue: https://github.com/apache/superset/issues/32591 still open, but the issue reporter has a different Superset version. 
Regardless of whether there is a string or numeric value on the Y-axis, sorting does not seem to make sense as it does not seem to be in alphabetical or numeric order.

In addition, it renders all the 0 values as <NULL> on both the x and y axis, and possibly also on the heatmap grids. 

### Screenshots/recordings

<img width="1385" alt="Image" src="https://github.com/user-attachments/assets/4edd682e-1976-41e8-b864-c4925e8a4c20" />

In this screenshot:

- 0 values rendered as `<NULL>` on axes:
  - 0 in `ret_col` and `ret_row` is displayed as `<NULL>` on the X and Y axes.
  - The raw data clearly includes 0 (see Results on the bottom right corner), so it should not be shown as null.

- Y-axis sorting is inconsistent:
  - Even when “Sort Y Axis” is set to “Axis ascending”, the Y-axis values are not ordered numerically.
  - Y-axis order appears to depend on the order of values in the X-axis or underlying data groupings, this can be observed in the Results on the bottom right corner: `ret_row` is what on the y-axis, and it's order in the Results is the same as the order on the heatmap.

<img width="1047" alt="Image" src="https://github.com/user-attachments/assets/9b21f4fa-56ac-49c2-9b5b-983259631d43" />

In this screenshot:

- When `test_value` is 0, it's not being displayed on the plot. It doesn't have a grid with 0 in it, as the grids for test_value=1. I also don't see anything when hover over it.

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Checklist

[...]

PR #34547, opened 2025-08-05
fix(heatmap): preserve axis order when data combinations are missing

### LINKED ISSUE #33245: Heatmap lost ability to sort both y-axis and x-axis starting in 4.1
### Bug description

Originally filed as https://github.com/apache/superset/issues/33105 but as a two-part issue which made it hard to track. We dealt with one part of the issue and closed that one. See that thread for some discussion about this sorting problem.

In Superset 4.0.2 you could sort both the x-axis and y-axis of the heatmap chart, allowing charts like this where day and hour are plotted against each other, both in logical order:

Image

Starting in Superset 4.1 this doesn't work if some combinations lack data. For instance (images taken from #33105):

Image

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

rusackas:
## Summary

This PR fixes issue #33245 where the heatmap chart's y-axis would display in random order when some data combinations were missing. Starting in version 4.1, the chart lost the ability to maintain proper axis ordering in these cases.

### Root Cause
The new heatmap chart uses server-side ordering via SQL ORDER BY clauses. However, ECharts automatically creates axis categories from the data it receives, which doesn't preserve the intended order when data is sparse.

### Solution
The fix explicitly sets axis categories in the correct order by:
1. Extracting unique values for each axis from the data
2. Sorting them according to the user's sorting configuration (alphabetical or by metric value)
3. Explicitly setting the axis data in the ECharts configuration

This ensures both axes maintain their proper order regardless of missing data combinations.

## Test plan

Added comprehensive unit tests that verify:
- Axes maintain alphabetical order when data is missing
- Descending order works correctly
- Sorting by metric value works as expected
- Numeric values are sorted numerically (not as strings)
- Original order is preserved when no sorting is specified
- Null/undefined values are handled correctly

### Manual Testing
1. Create a heatmap chart with sparse data (e.g., day of week vs hour with some combinations missing)
2. Set both X and Y axis sorting to "Axis ascending"
3. Verify both axes display in the correct order
4. Test with different sorting options (descending, by metric value)

## Related Issues
- Fixes #33245
- Related to #33105 (original issue that was split)

### DIFF
--- a/.claude_rc
+++ b/.claude_rc
@@ -1,0 +1,40 @@
+# Claude Code RC for issue34531
+
+This is a claudette-managed Apache Superset development environment.
+
+## Project: issue34531
+- Worktree Path: /Users/evan_1/.claudette/worktrees/issue34531
+- Frontend Port: 9003
+- Frontend URL: http://localhost:9003
+
+## Quick Commands
+
+Start services:
+```bash
+claudette docker up
+```
+
+Access frontend:
+```bash
+open http://localhost:9003
+```
+
+Run tests:
+```bash
+# Backend
+pytest tests/unit_tests/
+
+# Frontend
+cd superset-frontend && npm test
+```
+
+## Environment Details
+- Python venv: `.venv/` (auto-activated in claudette shell)
+- Node modules: `superset-frontend/node_modules/`
+- Docker prefix: `issue34531_`
+
+## Development Tips
+- Always use `claudette shell` to work in this project
+- Run `pre-commit run --all-files` before committing
+- Use `claudette docker` instead of docker-compose directly
+- The frontend dev server runs on port 9003 to avoid conflicts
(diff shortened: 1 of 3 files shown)


sfirke:
Thanks for trying to work on this! I deployed this branch into my test environment. The problematic heatmaps now don't render any data at all, everything on the x-axis is NULL (though you can see in the Results below the chart that underlying data is present):

image

I then went back to 6.0.0rc2 and confirmed (a) that the heatmap renders there, so this is a failure on this fix branch (b) that it renders with the y-axis categories in the wrong order = the original sorting bug persists.

sfirke (changes requested):
Right now it's not rendering the heatmap in question - makes me wonder if the tests are invalid or just missing something about my case.

rusackas:
Superseded by #36302

PR #34559, opened 2025-08-05
fix(saved_query_url): restore the legacy link for backup

justinpark:
### SUMMARY

In #31421 the existing saved query link ID has been replaced with a new permalink API. This change is currently impacting Airbnb services that reference the previous saved query links.

To address this, in 5.0, we restored the original query link functionality to a "share link" action, while keeping the updated functionality for "copy query URL."
Additionally, please note that a saved query ID always reflects the latest information when a saved query is edited. In contrast, the "copy query URL" uses the permalink method, which means it does not capture any changes made to the query after the URL is generated. To clarify this behavior, this commit updates the label to "copy current query URL."

After:

Screenshot 2025-08-05 at 10 10 16 AM

Screenshot 2025-08-05 at 10 10 20 AM

### TESTING INSTRUCTIONS

Go to Saved queries and then check the actions

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/pages/SavedQueryList/SavedQueryList.test.jsx
+++ b/superset-frontend/src/pages/SavedQueryList/SavedQueryList.test.jsx
@@ -376,9 +376,9 @@ describe('RTL', () => {
     // Wait for the tooltip to pop up
     await screen.findByRole('tooltip');
 
-    // Grab and assert that "Copy query URl" tooltip is in the document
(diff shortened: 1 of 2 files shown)


justinpark:
cc: @LevisNgigi

rusackas:
I think this makes sense, but curious if @kasiazjc or @mistercrunch think having two links that copy different things is too confusing.

PR #34561, opened 2025-08-05
feat: Tiled screenshots in Playwright reports

### LINKED ISSUE #29326: Superset 3.1.3: Long Dashboards render blank thumbnails unless Dashboard length is reduced
### Bug description

We have several dashboards that render their thumbnails properly. For some, Playwright would only render the background of the dashboard and not the charts (we have color backgrounds using CSS for some)

When I reduce the height of charts to fit them in less than 3 page down scrolls, the thumbnail works fine, but if the length exceeds a certain amount, a blank thumbnail is rendered.

### How to reproduce the bug

```python

FEATURE_FLAGS = {
    "ENABLE_TEMPLATE_PROCESSING": True,
    "DRILL_BY": True,
    "GLOBAL_ASYNC_QUERIES": True,
    "HORIZONTAL_FILTER_BAR": True,
    "THUMBNAILS": True,
    "THUMBNAILS_SQLA_LISTENERS": True,
    "LISTVIEWS_DEFAULT_CARD_VIEW": True,
    "ALERT_REPORTS": True,
    "PLAYWRIGHT_REPORTS_AND_THUMBNAILS": True,
    "USE_ANALAGOUS_COLORS": True,
    "TAGGING_SYSTEM": False,
}
WEBDRIVER_BASEURL = f"https://subdomain.domain.com"
WEBDRIVER_BASEURL_USER_FRIENDLY = "https://subdomain.domain.com"
SCREENSHOT_LOCATE_WAIT = 1000
SCREENSHOT_LOAD_WAIT = 5000
SCREENSHOT_SELENIUM_RETRIES = 1

SCREENSHOT_PLAYWRIGHT_WAIT_EVENT = "load"
SCREENSHOT_PLAYWRIGHT_DEFAULT_TIMEOUT = int(
    timedelta(seconds = 60).total_seconds() * 1000
)
```

Dockerfile:
```dockerfile
FROM apache/superset:3.1.3
RUN pip install --no-cache-dir \
    playwright \
    && playwright install-deps \
    && playwright install
```

1. Create a dashboard with several charts and tables that are at least 4-5 pages (on desktop) long and has 5-6 filters: one with a default value, and others that depend on a couple of the filters
2. Set filters to be horizontal

The dashboard should render a blank thumbnail:
notfine

Now remove rows from the dashboard to reduce its height or resize elements to fit them in a shorter height, the thumbnail is rendered fine:
fine

### Screenshots/recordings

_No response_

### Superset version

3.1.3

### Python version

Not applicable

### Node version

Not applicable

### Browser

Not applicable

### Additional context

_No response_

### Checklist

kgabryje:
### SUMMARY
Playwright reports used to return a blank screenshots when run on large dashboards. Fixes: https://github.com/apache/superset/issues/29326.

After some experiments I found out that the problem appears to actually come from Chromium itself, not Playwright. After enabling CPU 4x throttling and taking a screenshot using Chrome's dev tools, I noticed that before taking the screenshot, it unmounts and redraws a lot of UI elements, leaving just the dashboard's background, and it doesn't finish redrawing until after the screenshot is taken, resulting in blank image (see video).
The solution that this PR implements was to take multiple small screenshots/tiles and glue them together into the final report before sending it to the user.
I tested the solution on a large dashboard (~50 charts, deck.gl and Echarts) which failed consistently before. After my changes, it consistently sends expected screenshot.

There are also some config flags for fine tuning the tiling feature:
- `SCREENSHOT_TILED_ENABLED` (True by default, disabling not recommended)
- `SCREENSHOT_TILED_CHART_THRESHOLD` (threshold of dashboard height at which tiling gets triggered)
- `SCREENSHOT_TILED_HEIGHT_THRESHOLD` (number of charts at which tiling gets triggered)
- `SCREENSHOT_TILED_VIEWPORT_HEIGHT` (single tile viewport height)
If the blank screenshot problem persists, you can try decreasing the value of each of those flags

The changes described above, including new config flags, take effect only when `PLAYWRIGHT_REPORTS_AND_THUMBNAILS` feature flag is enabled.

Before:

image

After:

image

### TESTING INSTRUCTIONS
1. Enable `PLAYWRIGHT_REPORTS_AND_THUMBNAILS`, install playwright, run `playwright install`.
2. Create a very large dashboard (~50 charts should be enough)
3. Set up a report with this dashboard
4. Verify that the screenshot is generated correctly
5. Also test on smaller dashboards were the tiling effect should not trigger

### ADDITIONAL INFORMATION

- [x] Has associated issue: closes https://github.com/apache/superset/issues/29326
- [ ] Required feature flags:
- [ ] Changes UI


kgabryje:
Side note - the "After" screenshot looks normal on my email, Github must have squished it for some reason 🙂

superset/config.py
```diff
@@ -865,6 +865,14 @@ class D3TimeFormat(TypedDict, total=False):
     timedelta(seconds=60).total_seconds() * 1000
 )
 
+# Tiled screenshot configuration for large dashboards
+SCREENSHOT_TILED_ENABLED = True  # Enable tiled screenshots for large dashboards
+SCREENSHOT_TILED_CHART_THRESHOLD = 20  # Minimum charts to trigger tiled screenshots
+SCREENSHOT_TILED_HEIGHT_THRESHOLD = (
```
sadpandajoe, line 871: does this affect the screenshots for alerts/reports? I believe there is a way to trigger a min/max height and width.

pyproject.toml
```diff
@@ -71,6 +71,7 @@ dependencies = [
     "nh3>=0.2.11, <0.3",
     "numpy>1.23.5, <2.3",
     "packaging",
+    "Pillow==10.3.0",
```
sadpandajoe: nit: it's not alphabetical, but should this be sorted alphabetically?

kgabryje: Oh well spotted

superset/config.py
kgabryje: Yes it's for alerts/reports. We can set viewport width in report creation modal, and height can be set in config.py.
This however is an optimization for large dashboards - if dashboard is longer than `SCREENSHOT_TILED_HEIGHT_THRESHOLD`, then we use tiled screenshots. Tiled screenshots mean that instead of screenshotting the entire dashboard (i.e. the `.standalone` element), we screenshot only a fragment (`SCREENSHOT_TILED_VIEWPORT_HEIGHT`) multiple times.
For example, if the dashboard has height 10000px, and `SCREENSHOT_TILED_VIEWPORT_HEIGHT` is 3000, we take 3 3000px screenshots and 1 1000px screenshot and glue them together into 1 10000px image.

sfirke:
Outstanding! This will close #29326

pyproject.toml
```diff
@@ -79,6 +79,7 @@ dependencies = [
     "parsedatetime",
     "paramiko>=3.4.0",
     "pgsanity",
+    "Pillow==11.3.0",
```
villebro: Is there some reason we're pinning to 11.3.0, and not a range/lower bound?

pyproject.toml
```diff
@@ -181,7 +182,7 @@ tdengine = [
     "taos-ws-py>=0.3.8"
 ]
 teradata = ["teradatasql>=16.20.0.23"]
-thumbnails = ["Pillow>=10.0.1, <11"]
+thumbnails = ["Pillow>=10.0.1, <12"]
```
villebro: If `Pillow` is already a hard dependency as defined above, do we really need this option any more?

pyproject.toml
```diff
@@ -399,6 +400,7 @@ authorized_licenses = [
     "isc license (iscl)",
     "isc license",
     "mit",
+    "mit-cmu",
```
villebro, line 403: Did we check that this license is ok by ASF?

villebro (commented):
Great work! Let's try push this through asap! A few dep/license nags, other than that LGTM.

pyproject.toml
kgabryje: Yes, according to Claude it's basically an MIT license:

image

pyproject.toml
kgabryje: We can use range >=11, <12. We should use at least 11, as in versions <=10 they used a different license

pyproject.toml
kgabryje: Good point, I'll delete it

pyproject.toml
villebro: I think `>=11, <12` sounds good. I think there was some discussion a while back to try to avoid having upper bounds, as it restricts dependabot. So maybe just `>=11.3.0` is ok (people should be using lock files in prod anyway..)

pyproject.toml
```diff
@@ -181,7 +182,6 @@ tdengine = [
     "taos-ws-py>=0.3.8"
 ]
 teradata = ["teradatasql>=16.20.0.23"]
-thumbnails = ["Pillow>=10.0.1, <11"]
```
villebro, line 184: @michael-s-molina is the removal of an optional dependency considered a breaking change? As `Pillow` is hereby promoted to a hard dependency, do we need to leave the `thumbnails` optional dependency empty until the breaking window?

michael-s-molina: I think it is a breaking change but we kind of are in a breaking window based on @rusackas' [email](https://lists.apache.org/thread/pmb6z6lspt4322nonm8vrc1h8gxfdpb1)?. I'm not the release manager for 6.0 so @sadpandajoe @rusackas will probably know if we still are in a breaking window or not.

michael-s-molina: By the way, this PR might need something in `UPDATING.md`.

kgabryje: @michael-s-molina @villebro According to @rusackas email we're not accepting new breaking changes now, so I set `thumbnails = []` in `pyproject.toml`.
Also added an entry in `UPDATING.md`

rusackas:
> Side note - the "After" screenshot looks normal on my email, Github must have squished it for some reason 🙂

Yep... there's some max height. It's annoying.

Thanks for this PR, this is a very important one, and closes quite a few issues!

pyproject.toml
villebro: @kgabryje one last request: can we add a deprecation notice here so we can remove this in the next breaking window? I assume the removal will happen for 7.0.

kgabryje: @villebro Done 🙂

villebro (approved):
LGTM, thanks for the fix, and happy for you to finally get closure on this issue 😆

kgabryje:
> LGTM, thanks for the fix, and happy for you to finally get closure on this issue 😆

Now I won’t go bald from scratching my head about this 😄

udayv2000:
is this feature  available in the latest docker image 6.0.0 or 6.0.1rc

Syam9290:
Hi @kgabryje ,
 thanks for the update on this PR — I’m now able to schedule reports without any issues.
However, I’m facing a problem with the tiling screenshot method. In the generated reports, the tab titles are not showing up in the screenshots.
Could you please let me know if this is a known issue or if there’s a way to fix or configure this?
For context, I’m currently using Superset 6.
Thanks in advance for your help!

PR #34583, opened 2025-08-06
fix(drilling): drill by pagination works with MSSQL data source via db_engine_spec

### LINKED ISSUE #24072: MSSQL requires an order_by when using an OFFSET or a non-simple LIMIT clause

sfirke:
### SUMMARY
This is a new attempt at fixing #24072 . It takes a different approach than my first attempt, https://github.com/apache/superset/pull/27442. This time it makes the ordering change in the Microsoft SQL Server (mssql) spec rather than for all engines.

I checked that this solves the problem in my environment. I would like input on whether my change to `compile_sqla_query` is correct.

### TESTING INSTRUCTIONS
Deploy to an environment with SQL Server data warehouse, try drilling and flipping through the pages. It works for me, if needed I can produce a video.

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #24072
- [ ] Required feature flags:
- [ ] Changes UI


superset/db_engine_specs/mssql.py
```diff
@@ -163,6 +164,33 @@ def extract_error_message(cls, ex: Exception) -> str:
...
+        try:
+            # Try to access the internal query structure safely
+            has_offset = getattr(qry, "_offset", None) is not None
+            has_order_by = bool(getattr(qry, "_order_by_clauses", ()))
+
+            if has_offset and not has_order_by:
+                if qry.selected_columns:
+                    first_column = list(qry.selected_columns)[0]
+                    qry = qry.order_by(first_column)
+                else:
```
sfirke, line 185: This was AI's idea. If someone feels confident that columns will always be available in this context - seems likely to me - we could remove the if/else and this backup casting of 1.

sfirke:
Closed in favor of #34724 which just merged

PR #34589, opened 2025-08-07
feat(Timeseries & MixedTimeseries): Added configurability to value labels in Timeseries and Mixed Timeseries charts. Can rotate and set render distance of value labels

Dev10-34:
### SUMMARY
The value label once you check the "Show value" checkbox for Time Series and Mixed Timeseries charts, is not configurable. Added the capability to customize it as seen fit. One can now rotate the labels or even set the render distance of the value labels. The former is helpful in both Timeseries and Mixed Timeseries, but the latter change is very useful for Mixed Timeseries type charts. In situations where you have very close values, there is a lot of visual clutter and it is hard to distinguish the labels. In that case, using Mixed timeseries, anyone can just set the distance of type A query by a positive value and type B query by a negative value to increase readability.

Before:
image

After:
image
image
image
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


rusackas:
This PR hasn't been updated in a long time and has merge conflicts that need to be resolved. Testing instructions are also missing from the PR description. We'll mark it as draft for now — please rebase against `master`, add testing instructions, and set it back to 'ready for review' when it's ready.

rusackas (commented):
Still in draft here, and the things I flagged last time — rebase on master and testing instructions — are still open, so flip it back to ready when those are in. 

One thing I want to understand before that though: in transformers.ts you're now setting verticalAlign: 'middle' on the label unconditionally, so it applies to every timeseries chart even when nobody touches the new rotate/distance controls. Doesn't that shift label positioning for existing charts? Curious if that's intentional, or if it should be gated behind showValue/the new props like rotate and distance are. 

Also the new transformer behavior isn't covered by a test — the test change just backfills the new fields into existing fixtures. Worth an assertion that rotateValue/distanceValue actually land on the series label. 

Minor: control order is distance then rotate in controls.tsx but rotate then distance in the Area and MixedTimeseries panels — would be nice to keep them consistent.

This has been dormant for a long time, so we might close it in our next pass at stale PRs.

rusackas:
@Dev10-34 since this has gone quiet for a while and is conflicting with master, with testing instructions still missing. Closing as inactive, but ping me if you want to pick it back up and I'll happily reopen.

Issue #28053, opened 2024-04-16
Time Series Bar Chart is Showing only Alternate Months in the X axis despite having data


hk-the-analyst:
## Screenshot
![image](https://github.com/apache/superset/assets/123436992/df52c36a-3fde-4127-b667-21588a609ea5)

## Time Series Bar Chart is Showing only Alternate Months in the X axis despite having data

[describe the issue here!]

## Design input
I want to see every month being displayed on the x-axis

shantanukhond:
For the above chart can you try changing X Axis time format from adaptive to 
 %b %y 
as shown in screenshot below if it works?
![image](https://github.com/apache/superset/assets/24731253/29bf93dd-7772-4126-a43a-83900f61310c)

magickspell:
It looks like this can be handled in code:
in file `/superset/superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts`
add `splitNumber: data.length`, somewhere in xAxis object `let xAxis: any = {`

```
let xAxis: any = {
    splitNumber: data.length,
    type: xAxisType,
    name: xAxisTitle,
    nameGap: convertInteger(xAxisTitleMargin),
    nameLocation: 'middle',
    axisLabel: {
      hideOverlap: true,
      formatter: xAxisFormatter,
      rotate: xAxisLabelRotation,
    },
    minorTick: { show: minorTicks },
    minInterval:
      xAxisType === AxisType.Time && timeGrainSqla
        ? TIMEGRAIN_TO_TIMESTAMP[
            timeGrainSqla as keyof typeof TIMEGRAIN_TO_TIMESTAMP
          ]
        : 0,
    ...getMinAndMaxFromBounds(
      xAxisType,
      truncateXAxis,
      xAxisMin,
      xAxisMax,
      seriesType,
    ),
  };
```

sfirke:
Is this still an issue after this merged: https://github.com/apache/superset/pull/33729 ?

rusackas:
Can anyone confirm whether PR #33729 resolved the alternating-month display issue on time series bar charts? If confirmed fixed, we can close this out.

rusackas:
#34595 added the force-timegrain option for timestamp x-axes, which was the agreed fix here, and it merged - closing, but ping me to reopen if the months still skip.

PR #34595, opened 2025-08-07
feat(Timeseries & MixedTimeseries): Force selected timegrain on timeseries intervals when the x-axis is of timestamp type

Dev10-34:
### SUMMARY

Currently, the X-Axis interval doesn't work when the axis type is taking a temporal type column. To fix that, added a check box which will force the time grain as selected interval for labels on the x axis.

Before:
image

After:
image

image

### ADDITIONAL INFORMATION

- [x] Has associated issue:https://github.com/apache/superset/issues/28053
 https://github.com/apache/superset/pull/33191

- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -42,6 +42,7 @@ import {
   xAxisBounds,
   xAxisLabelRotation,
   xAxisLabelInterval,
+  forceMaxInterval,
 } from '../controls';
 
 const {
@@ -358,6 +359,7 @@ const config: ControlPanelConfig = {
         ['x_axis_time_format'],
         [xAxisLabelRotation],
         [xAxisLabelInterval],
+        [forceMaxInterval],
         [<ControlSubSectionHeader>{t('Tooltip')}</ControlSubSectionHeader>],
         [
           {
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -211,6 +211,7 @@ export default function transformProps(
     sortSeriesAscending,
     sortSeriesAscendingB,
     timeGrainSqla,
+    forceMaxInterval,
     percentageThreshold,
     showQueryIdentifiers = false,
     metrics = [],
@@ -579,6 +580,12 @@ export default function transformProps(
               timeGrainSqla as keyof typeof TIMEGRAIN_TO_TIMESTAMP
             ]
           : 0,
+      maxInterval:
+        xAxisType === 'time' && timeGrainSqla && forceMaxInterval
+          ? TIMEGRAIN_TO_TIMESTAMP[
+              timeGrainSqla as keyof typeof TIMEGRAIN_TO_TIMESTAMP
+            ]
+          : undefined,
       ...getMinAndMaxFromBounds(
         xAxisType,
         truncateXAxis,
--- a/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -575,7 +575,7 @@ export default function transformProps(
       },
       minorTick: { show: minorTicks },
       minInterval:
-        xAxisType === AxisType.Time && timeGrainSqla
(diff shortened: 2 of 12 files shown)


mistercrunch:
Seems like a bit of a band-aid, any reason why we wouldn't fix the core issue/functionality?

Dev10-34:
> Seems like a bit of a band-aid, any reason why we wouldn't fix the core issue/functionality?

Short answer: The interval config doesn't work for xAxis values of type "time".
Long answer:
If you look into heatmaps, we are able to see that configuring the XScaleInterval option is able to add the intervals properly irrespective of the type of data going into the xaxis (in terms of echarts, the categories are time, value and category). This is because the type for x-axis labels is hard set to 'category' ( line 251 in [superset](https://github.com/apache/superset/tree/master)/[superset-frontend](https://github.com/apache/superset/tree/master/superset-frontend)/[plugins](https://github.com/apache/superset/tree/master/superset-frontend/plugins)/[plugin-chart-echarts](https://github.com/apache/superset/tree/master/superset-frontend/plugins/plugin-chart-echarts)/[src](https://github.com/apache/superset/tree/master/superset-frontend/plugins/plugin-chart-echarts/src)/[Heatmap](https://github.com/apache/superset/tree/master/superset-frontend/plugins/plugin-chart-echarts/src/Heatmap)/transformProps.ts).  
Not sure why this is the way it happens, but it allows all the timestamp intervals in Heatmaps to become a configurable property. Under Timeseries type of charts, however, the type of xAxis is not hard coded. This results in the xAxisInterval configuration to not function when the type of selected column is of "time".
Hence, taking a bit of an idea from this : https://github.com/apache/superset/pull/33191/files#r2055905885, I have made it so that  intervals of timestamp type of xAxis are able to be set properly as well. 
One way to do it was to set the max intervals between each timestamp as the selected timegrain itself.
The only downside to it is that timgrainsqla property doesn't have all the Timegrain options. I am wondering if I should include all the options in a seperate PR as well

mistercrunch:
I'd like to better understand the core issue, is it that you want the ticks on the x axis to match exactly the time grain (?) 
So say you have days as granularity, you want one tick per day?

It's hard to come up with a perfect name for the checkbox, but it's something like "Force one tick label per time grain interval". If that's the case, it's essentially working around echart trying to be smart and compute which ticks to show. In theory this feature can/could lead to lots more labels than should be displayed (overlapping and such) in some cases. 

Working on Superset it's hard to strike the balance between magic (defaults that generally work) and opening the full configurability of the underlying library (say exposing all the dozens of props that echarts has behind the scene). In theory just the Xaxis could expose dozens of props. We generally side on exposing the long tail where needed with smart defaults.

Dev10-34:
The goal is to show all possible labels if the axis type is of time type.   
I think the core issue is that the interval property does not allow the user to make the intervals configurable IF the xaxis is of a type other than "category". Basically, if you put categorical values in x axis and then set the intervals, it will work as expected. However, if you are assigning a timestamp to the x axis, the interval configuration doesn't work as intended.

Dev10-34:
For example, lets say we want to show monthly data (time grain is set to be Month). To ensure no overlapping, echarts will omit a few months from the xaxis, even if the labels are rotated by 90 degress. So, instead of all 12 months coming up, only quarterly months come up (Jan, April, July, October). Using the X Axis Interval configuration doesn't change the frequency at all either. As a result, this checkbox will allow the user to set the interval based on the selected timegrain itself. 
 No config state (initial state):
image

Setting label intervals:
image

Checking the checkbox:
image

mistercrunch:
How wow what's happening in those scrollbars in your screenshots? Is this FF?

Dev10-34:
> How wow what's happening in those scrollbars in your screenshots? Is this FF?

I am not sure what you mean here. Could you please explain it again?

mistercrunch:
[it's unrealted to your pr, but] In the screenshots you provided, the scrollbars don't seem to be styled properly (white background instead of black). Wondering if it's a browser-specific issue. What browser/OS are you using?

Dev10-34:
I am running it in wsl using docker compose in brave browser

Dev10-34:
Does the PR make sense now though? Is there anything I can add onto this?

Dev10-34:
Hey @mistercrunch, just checking in on the PR. What can I do to improve this?

mistercrunch (approved):
Ok! Sorry took a moment to get to it, but I spent a moment analyzing deeply whether the new checkbox is a true "no-op" when unchecked and I'm pretty sure that it's the case. Only difference might be the presence of `maxInterval: undefined` which is, well, what it was implicitly before the change.

mistercrunch:
Thanks for the contrib btw!

Dev10-34:
Thanks for the review and merge!

Dev10-34:
Hi @mistercrunch,
Any chances this gets added on a new release?

PR #34598, opened 2025-08-07
fix(db_engine_specs): generate correct boolean filter SQL syntax for Athena compatibility

oscep:
### SUMMARY

Fixes boolean filter SQL generation that uses invalid `IS TRUE/IS FALSE` syntax instead of `= true/= false`. 
When boolean filters are applied, Superset generates SQL like `column IS TRUE` which is invalid syntax for Athena which expects `column = true`.

- Before: Boolean filters generate `test_col IS TRUE` (invalid Athena syntax)
- After: Boolean filters generate `test_col = true` (valid Athena syntax)

### TESTING INSTRUCTIONS

1. Create a chart with a boolean column filter
2. Set the filter value to `true` or `false`
3. Click "Update chart" to generate and execute SQL
4. Observe the error message `Data error: Error: An error occurred (InvalidRequestException) when calling the StartQueryExecution operation: line 6:15: mismatched input 'true'. Expecting: 'DISTINCT', 'NOT', 'NULL'`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration

- additional unit test fails as expected prior to applying fix (see [here](https://github.com/apache/superset/actions/runs/16805947172/job/47598304001#step:5:281)).


msyavuz (approved):
LGTM! Thanks

PR #34607, opened 2025-08-08
feat: Add Rison URL filters for simplified filter syntax

mistercrunch:
- Add RisonFilterParser to parse f parameter with Rison syntax
- Integrate with explore endpoint to merge filters with form_data
- Support equality, IN, NOT, OR, comparison operators, BETWEEN, LIKE
- Add comprehensive unit tests and documentation
- Enable human-readable filter URLs like f=(country:USA,year:2024)

Examples:
- Simple: `?f=(country:USA)`
- Multiple: `?f=(country:USA,year:2024)`
- IN operator: `?f=(country:\!(USA,Canada))`
- NOT: `?f=(NOT:(status:inactive))`
- OR: `?f=(OR:\!(status:active,priority:high))`
- Comparisons: `?f=(sales:(gt:100000))`
- BETWEEN: `?f=(date:(between:\!(2024-01-01,2024-12-31)))`
- LIKE: `?f=(name:(like:'%smith%'))`
- Special chars: `?f=(region:'North America')`

Behavior

- in Explore, filters are added in the Filter dropdown box
- in Dashboard, filters are added as a new section in the filter bar as seen here:

Screenshot 2025-08-11 at 8 31 10 AM
Screenshot 2025-08-11 at 8 30 55 AM

### DIFF
--- a/docs/docs/using-superset/url-filters.mdx
+++ b/docs/docs/using-superset/url-filters.mdx
@@ -1,0 +1,283 @@
+---
+title: URL Filters
+hide_title: false
+sidebar_position: 4
+version: 1
+---
+
+# URL Filters
+
+Apply filters to dashboards and charts directly through the URL using a simple, human-readable syntax.
+
+Superset URL filters use [Rison](https://github.com/Nanonid/rison), a data serialization format that's JSON-compatible but optimized for URLs - it's expressive, compact, and looks as great as URLs can look without all the percent-encoding clutter.
+
+## Why URL Filters?
+
+URL filters allow you to:
+- Share specific views of data with colleagues
+- Bookmark frequently used filter combinations
+- Create dynamic links in external applications
+- Override saved dashboard states temporarily
+- Build data-driven workflows
+
+## Quick Start
+
+Add the `f` parameter to any dashboard or explore URL:
+
+```
+/dashboard/123?f=(country:USA)
+/explore?f=(year:2024)
+```
+
+## Basic Syntax
+
+### Single Filter
+
+Filter by a single value:
+
+```
+f=(country:USA)
+```
+
+### Multiple Filters (AND)
+
+Combine multiple filters with commas (AND logic):
+
+```
+f=(country:USA,year:2024)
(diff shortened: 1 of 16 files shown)


yousoph:
Bugs: 
1. On dashboards, when the dashboard filter has a space in it, setting `?f=('Country Code'=USA)` with the quotes and the space didn't successfully match the dashboard filter that already exists 
2. Unable to set two values for a filter on dashboards. I tried `?f=(country:USA,Canada)` and `?f=(country:USA,country:Canada)` and neither set the country filter to have both USA and Canada filter values 
3. Unable to set dashboard filters for a numerical range filter. I tried using this syntax: `?f=(msrp:(between:\!(35,200)))`

Improvements: 
1. On dashboards, when the filter parameter in the URL matches an existing dashboard setting, the "URL Filters" does not need to be displayed in the Dashboard Filter Bar 
2. On both charts and dashboards, it would be best if the filters were already applied when the user loads the chart. Currently, the filters are set but the "Apply filters" or "Update chart" button needs to be clicked before the filters are visible in the chart or dashboard 
4. On dashboards, when a filter doesn't exist as a dashboard filter and is applied as a URL filter, we should use the same visual pattern as we do for cross filters. This means: 
   a. The filter bar section for URL filters should match the cross filters section visually. The section should be collapsible and each filter should have a chip that can be removed from the UI 
   b. When there are URL filters applied, the filter icon on the charts should have a section for URL filters 
   c. The order of the sections in the filter bar and the filter icon hover tooltip should be: URL filters, cross filters, dashboard filters 
5. If the user doesn’t have permission to set the URL filters, show them an error on the dashboard so they are aware that the filters were not set properly. This should probably show in the URL filters section

rusackas:
@mistercrunch do we still want to keep this one open?

eschutho:
@rusackas we're going to have someone pick this up and carry it forward.

rusackas:
Happy to adopt this if needed :)

rusackas:
Looks like @hughhhh picked it up and merged it (thank you, and hi!) so I'll close this one out now.

PR #34625, opened 2025-08-08
fix: Remove deprecated @types/classnames package

rusackas:
## Summary
Removes the deprecated `@types/classnames` npm package to fix npm warning.

## Details
The classnames library provides its own TypeScript definitions, so the separate `@types/classnames` package is unnecessary and deprecated. This PR removes it from package.json to eliminate the following npm warning:

```
npm warn deprecated @types/classnames@2.3.4: This is a stub types definition. classnames provides its own type definitions, so you do not need this installed.
```

## Test plan
- [x] npm install runs without the @types/classnames deprecation warning
- [x] TypeScript compilation still works correctly (classnames provides its own types)


sadpandajoe (approved):
LGTM

Issue #23466, opened 2023-03-23
Is there a way to beautify login page?


k6407399:
## Screenshot
<img width="956" alt="Capture" src="https://user-images.githubusercontent.com/97595567/227122313-2fe50098-ba66-492b-b35f-cebdaddb8291.PNG">

## Description
Dear Team,

Is it possible to beautify the login page? with nice background and with some transitions/animations.

We are currently using Apache Superset for 2 different versions for 2 different Projects.

For a Project we are using Superset Version : 0.38.1
For the other Project we are using Superset Version : 2.0.1

Which Version is feasible for beautifying the Superset Login Page. 

## Design input
Want to have a nice look and feel as soon as we launch the Supersets Login Page.

EugeneTorap:
Hi @k6407399! I like this idea!
@michael-s-molina @kgabryje @justinpark Is it possible to move login page to ReactJS?

nytai:
While the login page could be converted to react, it would require breaking from FAB and having to reimplement all the different auth flavors FAB supports. You can override the html templates via the `BLUEPRINTS` config

rusackas:
Closing this as stale since it's been silent for so long, and we're trying to steer toward a more actionable Issues backlog. If people are still encountering this in current versions (currently 3.x) please re-open this issue, open a new Issue with updated context, or raise a PR to address the problem. Thanks!

PR #34640, opened 2025-08-11
feat(theming): add dynamic theme-aware login page background and login form branding

VasilijeBursac:
### SUMMARY
This PR introduces dynamic, theme-driven customization/branding of Apache Superset's login page.  The goal was to provide organizations with the ability to personalize their Superset login screen with a custom background and login form logo, similar to how Microsoft 365 allows company-specific branding on its sign-in pages.

**Key Changes:**
- Added support for theme tokens to control login background and login form logo
- Introduced new optional tokens: `loginPageBackgroundImageUrl`, `loginFormBrandLogoUrl`, `loginPageBackgroundOverlayColor`, `loginPageBackgroundBlur` 
- When tokens are omitted, the login page has Superset’s default styling
- `loginPageBackgroundBlur`: CSS blur radius applied to the login page background image (e.g., `4px`, `0.25rem`, `1em`, `10vw`)
- `loginPageBackgroundOverlayColor` is an optional theme token that specifies the overlay color applied on top of the login page background image. It accepts any valid CSS color value, including:
	- Hex without alpha (e.g. `#000000`)
	- Hex with alpha (e.g. `#00000080`)
	- RGB / RGBA (e.g. rgba(`0, 0, 0, 0.5`))
	- HSL / HSLA (e.g. hsla(`0, 0%, 0%, 0.5`))
	- Named colors (e.g. `black`)
- If `loginPageBackgroundImageUrl` is set, but loginPageBackgroundOverlayColor is omitted, the overlay color defaults to `rgba(0, 0, 0, 0.5)`, i.e., a half-transparent black overlay. The default value for `loginPageBackgroundBlur` is `0px` (no blur effect on the background image).

**Notes:**
One visible difference in the login form’s appearance is that the "Sign in" title text is larger than in the default styling. Additionally, I manually handled the margins for this section, which may make the code feel somewhat cumbersome and spacing inconsistent with the rest of the Superset's UI. Both of these choices stem from the need to inject both the logo and the title into the card header to accommodate the new branding elements. If anyone has suggestions on how to implement this more consistently with Superset’s standards, I’d be happy to update the implementation accordingly.

I’ve aimed to follow Superset’s existing styling and component conventions, but please let me know if anything feels inconsistent. This is my first contribution to the project, and I’m still getting familiar with Superset, so it probably isn't done perfectly. Looking forward to your feedback!

For demonstration purposes, I’ve used Superset’s default logo and a placeholder background image sourced from the internet. These are not intended as final assets, but they should give a clear idea of how organizations can apply their own branding using theme tokens.

**Before:**

Light:
chrome_F4W1ZNxjHZ

Dark:
chrome_jdl83pBwet

**After:**

Light:
chrome_CcXqgbbuiB

Dark:
chrome_KMsjrGoAJZ

Dark + Blur Effect:
chrome_E6nHKMcgpC

Reference - Microsoft 365’s company-branded sign-in screen:
Microsoft365 branded login

### TESTING INSTRUCTIONS

1. Place your background and logo images in the following folder: `/superset-frontend/src/assets/images`
2. Reference Images in Theme JSON:
   - Superset's Theme CRUD UI → Settings > Theme > JSON Editor
   - `superset_config.py` → using the THEME config
3. Example theme JSON:
```
{
  "algorithm": "dark",
  "token": {
    "loginPageBackgroundImageUrl": "/static/assets/images/login-background.jpg",
    "loginFormBrandLogoUrl": "/static/assets/images/superset-logo-horiz.png",
    "loginPageBackgroundOverlayColor": "rgba(0, 0, 0, 0.6)",
    "loginPageBackgroundBlur": "3px"
  }
}
```

### ADDITIONAL INFORMATION

- [ x] Has associated issue: https://github.com/apache/superset/issues/23466
- [ ] Required feature flags:
- [x ] Changes UI

### DIFF
--- a/superset-frontend/src/pages/Login/index.tsx
+++ b/superset-frontend/src/pages/Login/index.tsx
@@ -19,7 +19,7 @@
 
 import { t } from '@apache-superset/core/translation';
 import { SupersetClient } from '@superset-ui/core';
-import { styled, css } from '@apache-superset/core/theme';
+import { styled, css, useTheme } from '@apache-superset/core/theme';
 import {
   Button,
   Card,
@@ -80,7 +80,46 @@ const StyledLabel = styled(Typography.Text)`
   `}
 `;
 
+const StyledBackground = styled.div`
+  ${({ theme }) => {
+    const bgImageUrl = theme.loginPageBackgroundImageUrl;
+    const overlayColor =
+      theme.loginPageBackgroundOverlayColor || 'rgba(0, 0, 0, 0.5)';
+    return bgImageUrl
+      ? css`
+          background-image: url(${bgImageUrl});
+          background-size: cover;
+          background-position: center;
+          position: fixed;
+          top: 0;
+          left: 0;
+          width: 100vw;
+          height: 100vh;
+          &::after {
+            content: '';
+            position: absolute;
+            top: 0;
+            left: 0;
+            width: 100%;
+            height: 100%;
+            background-color: ${overlayColor};
+          }
+        `
+      : '';
+  }}
+`;
+
+const StyledBrandLogo = styled.img`
+  display: block;
+  max-width: 160px;
+  max-height: 64px;
+  margin: 0;
+  object-fit: contain;
+`;
+
 export default function Login() {
+  const theme = useTheme();
+
   const [form] = Form.useForm<LoginForm>();
   const [loading, setLoading] = useState(false);
   const dispatch = useDispatch();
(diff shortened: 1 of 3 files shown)


superset-frontend/src/pages/Login/index.tsx
```diff
@@ -74,7 +74,62 @@ const StyledLabel = styled(Typography.Text)`
...
+          &::after {
+            content: '';
+            position: absolute;
+            top: 0;
+            left: 0;
+            width: 100%;
+            height: 100%;
+            background-color: ${overlayColor};
+            z-index: 1;
+          }
```
msyavuz, line 122: Are these psuedo element styles necessary? I think we can simplify this a bit further. Something like this with more tweaks:

```suggestion
					background: ${ bgImageUrl ? url(bgImageUrl) : overlayColor }
```

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -74,7 +74,62 @@ const StyledLabel = styled(Typography.Text)`
...
+  }}
+`;
+
+const StyledBrandLogo = styled.img`
+  display: block;
+  max-width: 160px;
+  max-height: 64px;
+  margin: 0;
+  object-fit: contain;
+`;
```
msyavuz, line 134: We do a very similar thing with Main navbar brand logo. Maybe we can extract that out to a component and use that here as well? Also enable us to unify existing brand logo tokens and this new one. (add a config option for displaying brand logo in login for example)

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -74,7 +74,62 @@ const StyledLabel = styled(Typography.Text)`
   `}
 `;
 
+const StyledBackground = styled.div`
+  ${({ theme }) => {
+    const bgImageUrl = theme.loginPageBackgroundImageUrl;
+    const overlayColor =
+      theme.loginPageBackgroundOverlayColor || 'rgba(0, 0, 0, 0.5)';
```
msyavuz, line 87: How does this default work with dark mode now?

msyavuz (commented):
Thank you for the PR! This is definitely a nice customization option. Also, you have a ci failure. I recommend installing pre-commit to catch these nice and early

superset-frontend/src/pages/Login/index.tsx
VasilijeBursac: > How does this default work with dark mode now?

The default overlay color (`rgba(0, 0, 0, 0.5)`) currently applies to both light and dark themes. You can see this in the first two "after" screenshots. Both use the default and show the semi-transparent black overlay regardless of theme mode.

I did consider setting different defaults based on the theme algorithm (e.g. lighter overlay for default, darker for dark), but decided it might be overkill since the overlay color and transparency are fully customizable via the `loginPageBackgroundOverlayColor` token. That gives users precise control over branding and contrast if they want to override the default.

Do you have any suggestions or preferences for how we should handle defaults across themes (different approach, different default value, etc.)? I’m happy to adjust if there’s a clearer UX or branding benefit.

msyavuz: If possible, i prefer we don't set a custom default at all and let Ant Design handle it. Now that i think about it i am not sure if login page specific color token makes sense. One of the antd tokens already controls this color right? cc @mistercrunch, @kasiazjc

superset-frontend/src/pages/Login/index.tsx
VasilijeBursac: Thanks for the suggestion! I started with pseudo-elements to keep the background image and overlay color visually separate, especially to support blur effects on the image without affecting the overlay. That way, the image can be blurred independently without affecting the overlay, and the overlay can be styled, themed, or even animated without interference.

My idea was to always render both layers, but visually separate:
- The background image + optional blur effect (`::before`)
- The overlay (`::after`), which can be fully transparent via the `loginPageBackgroundOverlayColor` token if users want to "turn off/hide" the overlay.

The proposed shorthand `background: ${ bgImageUrl ? url(bgImageUrl) : overlayColor` seems to assume we’re choosing between the image and the overlay, whereas I intended to layer both. Did I understand that part correctly? Happy to tweak the implementation if we can preserve that layering behavior. 
I am also open to any ideas and suggestions!

superset-frontend/src/pages/Login/index.tsx
VasilijeBursac: Thanks, that’s a great idea! I agree it would be useful to extract the logo logic into a shared `BrandLogo `component used in both the main navigation bar and the login form.

One small consideration: some organizations might prefer different logos in each context (e.g., a simplified icon in the navbar and a full logo in the login form), depending on available space or branding guidelines. That might be nitpicking, but it could be worth hearing from someone on Superset’s design/theme team about how much flexibility we want to support here.

If we go ahead with a unified approach, did you mean we’d keep a single `brandLogoUrl` token and add something like `showBrandLogoInLoginForm` token to toggle visibility in the login form, instead of having two separate url tokens?

Either way, I think extracting the logo into a component is a solid move. I’m happy to take that on once we’ve settled on the direction.

superset-frontend/src/pages/Login/index.tsx
mistercrunch: yes ... one thing I've been meaning to add to the theming framework is `component` support for `useTheme()`. Meaning wherever we do `const theme = useTheme()`, we could instead/optionally call `useTheme('Login')`, and it would return a full theme object with the component overrides applied.

Would be fairly easy to add, but we also need to tackle `css` and `styled` and got tangled up last time I tried to do this as they're kind of like js "decorators" (forgot the official name) but was a mind-bend.

rusackas:
@VasilijeBursac thanks for sticking with this — it's drifted into conflict with `master`, so a rebase is the first step. I'll tackle that now. You flagged the card-header spacing yourself; let's get it green and then sort out making that consistent with the rest of the UI.

rusackas:
@VasilijeBursac I'll rebase this. You flagged the card-header spacing yourself, so let's get it green first, then you/I/we can make that consistent with the rest of the UI.

rusackas:
Rebased this onto master and pushed to your branch — resolved the login-page markup conflict (kept master's `ensureAppRoot` fix on the register link) and merged in the new theme tokens cleanly. CI's running now; once it's green we can revisit the card-header spacing.

rusackas:
Pre-commit is failing now that it's rebased: custom-rules-frontend flags the hardcoded `rgba(0, 0, 0, 0.5)` overlay default in Login/index.tsx, needs a theme token instead. Also still open: the two `providers.map()` blocks need `key` props, and @msyavuz's threads on the pseudo-element background, extracting a shared BrandLogo component, and the overlay-color default in dark mode. Let's get those sorted before merging, please.

PR #34654, opened 2025-08-12
fix: centralize cache timeout -1 logic to prevent caching

dpgaspar:
### SUMMARY
 This PR fixes a critical bug where cache timeout value of -1 was being incorrectly interpreted as "cache forever" instead of "bypass cache" as documented in the UI. The issue stems from a semantic
   mismatch between Superset's intended behavior and the underlying cachelib/Redis implementation.

  The Semantic Mismatch:
  - Superset UI documentation: "Set to -1 to bypass the cache"
  - cachelib implementation (https://github.com/pallets-eco/cachelib/blob/main/src/cachelib/redis.py#L89-L93): -1 means cache forever (no expiration)

  When Superset passes -1 to Flask-Caching/cachelib, the underlying code explicitly checks for this value:
  #### In cachelib's RedisCache.set() method
``` python
  if timeout == -1:
      result = self._write_client.set(key, value)  # No expiration
  else:
      result = self._write_client.setex(key, value, time=timeout)  # With expiration
```

  This means users who set cache timeout to -1 expecting to bypass caching were actually creating permanent cache entries that never expire.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/constants.py
+++ b/superset/constants.py
@@ -243,3 +243,7 @@ class CacheRegion(StrEnum):
     DEFAULT = "default"
     DATA = "data"
     THUMBNAIL = "thumbnail"
+
+
+# Cache timeout constants
+CACHE_NO_TIMEOUT = -1  # Special value indicating no caching should occur
--- a/superset/constants.py
+++ b/superset/constants.py
@@ -246,4 +246,4 @@ class CacheRegion(StrEnum):
 
 
 # Cache timeout constants
-CACHE_NO_TIMEOUT = -1  # Special value indicating no caching should occur
+CACHE_DISABLED_TIMEOUT = -1  # Special value indicating no caching should occur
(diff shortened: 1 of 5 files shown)


Antonio-RiveroMartnez (approved):
lgtm

superset/constants.py
```diff
@@ -243,3 +243,7 @@ class CacheRegion(StrEnum):
     DEFAULT = "default"
     DATA = "data"
     THUMBNAIL = "thumbnail"
+
+
+# Cache timeout constants
+CACHE_NO_TIMEOUT = -1  # Special value indicating no caching should occur
```
Antonio-RiveroMartnez: nit: rename to `CACHE_DISABLED_TIMEOUT` because the `NO_TIMEOUT` in my mind reads like "it's cached forever because it has no timeout"

Vitor-Avila (approved):
LGTM!

PR #34620, opened 2025-08-08
fix: Use labels in Drill to Detail

Vitor-Avila:
### SUMMARY
This PR improves the Drill to Detail / Drill to Detail By flows to use the column labels.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
**Before**

image

**After**

image

### TESTING INSTRUCTIONS
1. Define labels to columns in a dataset.
2. Create a chart using it.
3. Confirm that drill to detail render using the column labels.
Fixes https://github.com/apache/superset/issues/34402

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34665, opened 2025-08-13
fix: Avoid dataset drill request if no perm

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/34620. Users might not have drilling access, and in this case the API request would return a `403`. While this is expected, having failed requests is suboptimal for embedded use-cases, so I've added a `skip` flag to the method (since we can't conditionally call it).

Unsure if it's the best way to solve this -- but open to feedback.

No UI changes.

### TESTING INSTRUCTIONS
Test coverage added.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
+++ b/superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
@@ -184,6 +184,7 @@ const SliceHeaderControls = (
     props.slice.datasource,
     props.dashboardId,
     props.formData,
+    !canDrillToDetail,
   );
 
   const datasetWithVerboseMap =
--- a/superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
+++ b/superset-frontend/src/components/Chart/ChartContextMenu/ChartContextMenu.tsx
@@ -161,6 +161,7 @@ const ChartContextMenu = (
     formData.datasource,
     dashboardId,
     formData,
+    !canDrillToDetail && !canDrillBy,
   );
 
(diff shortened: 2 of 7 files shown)


superset-frontend/src/dashboard/components/SliceHeaderControls/index.tsx
```diff
@@ -184,6 +184,7 @@ const SliceHeaderControls = (
     props.slice.datasource,
     props.dashboardId,
     props.formData,
+    !canDrillToDetail,
```
msyavuz, line 187: Only improvement i could think of at first is to move `useDatasetDrillInfo` inside the actual components but not sure if it's an improvement at all. I think having a skip argument is fine.

michael-s-molina: I think this should be handled by callers of this hook. The hook's functionality is to make a fetch request and return the drill info. You shouldn't call the hook if the required permissions are not there.

```
const datasetResource = canDrillToDetail ? useDatasetDrillInfo(...) : undefined;
```

Then you can handle the `undefined` `datasetResource`.

Vitor-Avila: @michael-s-molina I agree your suggestion looks a lot better, and also that the caller should decide if the request is needed or not. However, that would be a conditional hook, which then raises on `pre-commit`. 

I believe the ideal/long-term solution would be to refactor this whole flow, but for now hopefully this `skip` param is fine.

betodealmeida (approved):

PR #34671, opened 2025-08-13
fix: Bar chart crash when switching from Big Number

kgabryje:
### SUMMARY
When you try to switch viz types from a chart that doesn't have a groupby control (like Big Number), Bar Chart would crash because of undefined error. This PR adds a null check.

Before:

https://github.com/user-attachments/assets/d2c38e65-4fab-436a-a283-5fce11e3f383

After:

https://github.com/user-attachments/assets/f225ddc4-8d93-419a-82f9-991044e6f99a

### TESTING INSTRUCTIONS
1. Create and save a bar chart
2. Switch to big number
3. Switch back to bar chart
4. Verify that it doesn't crash

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Antonio-RiveroMartnez (approved):
lgtm

PR #34674, opened 2025-08-13
feat(db): custom database error messages

DamianPendrak:
### SUMMARY
Introduce `CUSTOM_DATABASE_ERRORS` in the config to replace raw database exceptions with a custom message and optional documentation links.

`custom_doc_links` - contains links that will show after the error message
`show_issue_info` - if set to False, hides the "See more" button that contains the issue code and link to Superset documentation

#### Before
Screenshot 2025-08-12 at 09 40 07

#### After
Screenshot 2025-08-12 at 09 43 44

### TESTING INSTRUCTIONS
1. In SQL Lab, run a query that results in a database error, ex. "SELECT * FROM non_existing_table"
2. Find a string that will match this error with a regex, ex. "TrinoUserError"
3. Add the custom errors in the `superset/custom_database_errors.py` file or add `CUSTOM_DATABASE_ERRORS` to `superset_config.py` or other config file with this structure:
```Python
CUSTOM_DATABASE_ERRORS = {
    "examples": {
        re.compile('permission denied for view'): (
            __(
                'Permission denied'
            ),
            SupersetErrorType.GENERIC_DB_ENGINE_ERROR,
            {
                "custom_doc_links": [
                    {
                        "url": "https://example.com/docs/1",
                        "label": "Check documentation"
                    },
                ],
                "show_issue_info": False,
            }
        )
    },
}
```
`examples` - the database connection name.
`permission denied for view` - part of the original message.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/ErrorMessage/CustomDocLink.tsx
+++ b/superset-frontend/src/components/ErrorMessage/CustomDocLink.tsx
@@ -1,0 +1,16 @@
+import { useTheme } from '@superset-ui/core';
+import { Icons } from '@superset-ui/core/components';
+
+export type CustomDocLinkProps = {
+  url: string;
+  label: string;
+};
+
+export const CustomDocLink = ({ url, label }: CustomDocLinkProps) => {
+  const theme = useTheme();
+  return (
+    <a href={url} target="_blank" rel="noopener noreferrer">
+      {label} <Icons.Full iconSize="m" iconColor={theme.colorPrimary} />
+    </a>
+  );
+};
--- a/superset-frontend/src/components/ErrorMessage/CustomDocLink.tsx
+++ b/superset-frontend/src/components/ErrorMessage/CustomDocLink.tsx
@@ -17,19 +17,17 @@
  * under the License.
  */
 
-import { useTheme } from '@superset-ui/core';
-import { Icons } from '@superset-ui/core/components';
+import { Flex, Icons } from '@superset-ui/core/components';
 
 export type CustomDocLinkProps = {
   url: string;
   label: string;
 };
 
-export const CustomDocLink = ({ url, label }: CustomDocLinkProps) => {
-  const theme = useTheme();
-  return (
-    <a href={url} target="_blank" rel="noopener noreferrer">
-      {label} <Icons.Full iconSize="m" iconColor={theme.colorPrimary} />
-    </a>
-  );
-};
+export const CustomDocLink = ({ url, label }: CustomDocLinkProps) => (
+  <a href={url} target="_blank" rel="noopener noreferrer">
+    <Flex align="center" gap={4}>
+      {label} <Icons.Full iconSize="m" />
+    </Flex>
+  </a>
+);
--- a/superset/custom_database_errors.py
+++ b/superset/custom_database_errors.py
@@ -1,0 +1,49 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import re
+from typing import Any
+
+from superset.errors import SupersetErrorType
+
+# CUSTOM_DATABASE_ERRORS: Configure custom error messages for database exceptions.
+# Transform raw database errors into user-friendly messages with optional documentation
+# links using custom_doc_links. Set show_issue_info=False to hide default error codes.
+# Example:
+# CUSTOM_DATABASE_ERRORS = {
+#     "trino": {
+#         re.compile(r'message="(?P<message>[^"]*)"'): (
+#             __(
+#                 'Unexpected error: "%(message)s"'
+#             ),
+#             SupersetErrorType.GENERIC_DB_ENGINE_ERROR,
+#             {
+#                 "custom_doc_links": [
+#                     {
+#                         "url": "https://example.com/docs/1",
+#                         "label": "Check documentation"
+#                     },
+#                 ],
+#                 "show_issue_info": False,
+#             }
+#         )
+#     }
+# }
+
+CUSTOM_DATABASE_ERRORS: dict[
+    str, dict[re.Pattern[str], tuple[str, SupersetErrorType, dict[str, Any]]]
+] = {}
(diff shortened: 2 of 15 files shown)


sadpandajoe:
@DamianPendrak so would you need files per database as different databases could have different links and different messages? Would it be helpful to also have a link or a collapsed box that will show the original error?

DamianPendrak:
@sadpandajoe My idea was to separate them based on the regex. But now I see that it might be an issue if there is the same error message from different databases, and you can't separate them using regex. I pushed an update with a changed structure of the `CUSTOM_DATABASE_ERRORS` so we can separate them by database engine names, and apply them accordingly:
```Python
CUSTOM_DATABASE_ERRORS = {
  "trino": {
    re.compile(r"TrinoUserError"): (
        "Custom error message",
        SupersetErrorType.GENERIC_DB_ENGINE_ERROR,
        {
            "custom_doc_links": [
                {
                    "url": "https://example.com/docs",
                    "label": "Check documentation"
                },
            ],
        }
    ),
  },
  "postgres": {}
}
```

About the original error - there is a similar way to overwrite the error messages in the db engine file in `superset/db_engine_specs`. Currently, it only shows an issue code and a link to the documentation in a collapsed component. The original error might be useful, but then I think it should be available in both custom error messages (db_engine_specs file and `superset_config.py`)
Screenshot 2025-08-06 at 13 23 28

sadpandajoe:
@DamianPendrak one other question. If you want to override multiple error messages and you have multiple databases, won't the config file get huge? Wondering if that is fine or not.

DamianPendrak:
@sadpandajoe, good point. The config file could become huge, so it's a good idea to put it into a separate file. I pushed the change that imports the custom errors, but you can still overwrite them in the config file

superset/custom_database_errors.py
```diff
@@ -0,0 +1,49 @@
...
+
+from superset.errors import SupersetErrorType
+
+# CUSTOM_DATABASE_ERRORS: Configure custom error messages for database exceptions.
+# Transform raw database errors into user-friendly messages with optional documentation
+# links using custom_doc_links. Set show_issue_info=False to hide default error codes.
+# Example:
+# CUSTOM_DATABASE_ERRORS = {
+#     "trino": {
+#         re.compile(r'message="(?P<message>[^"]*)"'): (
```
kgabryje: Could we perhaps make it easier to append a custom message or just the doc links to the original message? I know that this regex does that, but who likes writing regexes 😄

superset-frontend/src/components/ErrorMessage/CustomDocLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+export type CustomDocLinkProps = {
+  url: string;
+  label: string;
+};
+
+export const CustomDocLink = ({ url, label }: CustomDocLinkProps) => {
+  const theme = useTheme();
+  return (
+    <a href={url} target="_blank" rel="noopener noreferrer">
+      {label} <Icons.Full iconSize="m" iconColor={theme.colorPrimary} />
```
kgabryje: Nit: I think the icon is not centered vertically with the text. Also, the text's color changes on hover, but icon's color doesn't

Image

superset/db_engine_specs/base.py
```diff
@@ -1331,7 +1331,18 @@ def extract_errors(
         raw_message = cls._extract_error_message(ex)
 
         context = context or {}
-        for regex, (message, error_type, extra) in cls.custom_errors.items():
+
+        config_custom_errors = app.config.get("CUSTOM_DATABASE_ERRORS", {})
+        if not isinstance(config_custom_errors, dict):
+            config_custom_errors = {}
+        db_engine_custom_errors = config_custom_errors.get(cls.engine_name, {})
```
kgabryje: Should we also check for `cls.engine`? I tested on sqlite, and when I used `sqlite` as key in CUSTOM_DATABASE_ERRORS dict, it didn't work because `cls.engine_name` returns SQLite. Perhaps we could allow for both "human readable" and "normal" engine names?

Vitor-Avila:
hey @DamianPendrak could it be a concern if I have 2 `trino` connections (maybe with different auth methods each) and I just want to show a custom error message on one of them?

superset/custom_database_errors.py
DamianPendrak: Right, that's probably too advanced example. I added a simpler one that matches with a part of the error message "permission denied for view". The regex is also used in the custom errors in the database engine files, so it keeps the same approach

superset-frontend/src/components/ErrorMessage/CustomDocLink.tsx
DamianPendrak: Fixed

superset/db_engine_specs/base.py
DamianPendrak: Done. Works with both now

DamianPendrak:
@Vitor-Avila it could be an issue if the error message is the same and cannot be distinguished from another error message using regex. If one of the auth methods returned a different error message, then it can be achieved.

kgabryje (approved):
LGTM

kgabryje:
Added `hold!` label per contributor's request

DamianPendrak:
Thanks @Vitor-Avila for pointing that out. I changed the implementation to match errors with the database unique name instead of the database engine name. It allows users to create more precise custom errors. @kgabryje are you okay with reviewing the changes?

superset/db_engine_specs/base.py
```diff
@@ -1342,11 +1342,10 @@ def extract_errors(
...
-            if engine_key and engine_key in config_custom_errors:
-                engine_errors = config_custom_errors[engine_key]
-                if isinstance(engine_errors, dict):
-                    db_engine_custom_errors.update(engine_errors)
+        if database_name and database_name in config_custom_errors:
+            database_errors = config_custom_errors[database_name]
+            if isinstance(database_errors, dict):
+                db_engine_custom_errors.update(database_errors)
 
         if not isinstance(db_engine_custom_errors, dict):
```
kgabryje: Is there any scenario where db_engine_custom_errors is not a dict? We initialize it as dict in line 1343, and then update it only with other dicts

superset/db_engine_specs/databricks.py
```diff
@@ -279,19 +280,40 @@ def get_table_names(
...
         # for reference if their error message changes
 
         for key, value in cls.context_key_mapping.items():
             context[key] = context.get(value)
 
-        for regex, (message, error_type, extra) in cls.custom_errors.items():
+        if not isinstance(config_custom_errors, dict):
+            config_custom_errors = {}
+
+        db_engine_custom_errors = {}
```
kgabryje: This part is very similar to base.py. Does it make sense to DRY it?

superset/db_engine_specs/base.py
DamianPendrak: That's right. Removed the unnecessary check

superset/db_engine_specs/databricks.py
DamianPendrak: Good point. Moved it to a method

kgabryje (approved):
Re-approving after latest changes

PR #34679, opened 2025-08-13
fix: Check migration status before initializing database-dependent features

mistercrunch:
This fixes a startup issue where SeedSystemThemesCommand tried to access theme columns (is_system, is_system_default, is_system_dark) before database migrations were applied, causing initialization failures.

The solution adds an Alembic-based migration check to ensure the database schema is up-to-date before running any database-dependent initialization.

Changes:
- Added _is_database_up_to_date() method that compares current DB revision with the head revision from migration files
- Simplified _init_database_dependent_features() by removing redundant table existence checks and consolidating all DB-dependent features under the migration check
- Removed unused sqlalchemy imports (inspect, OperationalError)

This prevents startup failures when:
1. The database hasn't been migrated yet
2. New columns are added but migrations haven't run
3. The app starts before 'superset db upgrade' is executed

The fix ensures theme seeding and other DB-dependent features only run when the database schema is fully up-to-date.

### DIFF
--- a/superset/app.py
+++ b/superset/app.py
@@ -147,18 +147,6 @@ class SupersetApp(Flask):
             # Import here to avoid circular import issues
             from superset.extensions import feature_flag_manager
 
-            # Check if database URI is a fallback value before trying to connect
-            db_uri = self.config.get("SQLALCHEMY_DATABASE_URI", "")
-            if not db_uri or any(
-                fallback in db_uri.lower()
-                for fallback in ["nouser", "nopassword", "nohost", "nodb"]
-            ):
-                logger.warning(
-                    "Database URI appears to be a fallback value. "
-                    "Skipping database-dependent sync."
-                )
-                return
-
             # Check if database is up-to-date with migrations
             if not self._is_database_up_to_date():
                 logger.info("Pending database migrations: run 'superset db upgrade'")
--- a/superset/app.py
+++ b/superset/app.py
@@ -21,6 +21,10 @@ import os
 import sys
 from typing import cast, Iterable, Optional
 
+from alembic.config import Config
+from alembic.runtime.migration import MigrationContext
+from alembic.script import ScriptDirectory
+
 if sys.version_info >= (3, 11):
     from wsgiref.types import StartResponse, WSGIApplication, WSGIEnvironment
 else:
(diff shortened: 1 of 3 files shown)


superset/initialization/__init__.py
```diff
@@ -503,11 +504,40 @@ def init_views(self) -> None:
             icon="fa-lock",
         )
 
+    def _is_database_up_to_date(self) -> bool:
```
dpgaspar: nit: we could move this to `utils` if you think it could be useful for other code paths in the future

superset/initialization/__init__.py
```diff
@@ -517,39 +547,24 @@ def _init_database_dependent_features(self) -> None:
...
         # Register SQLA event listeners for tagging system
         if feature_flag_manager.is_feature_enabled("TAGGING_SYSTEM"):
             register_sqla_event_listeners()
 
         # Seed system themes from configuration
         from superset.commands.theme.seed import SeedSystemThemesCommand
 
-        if inspector.has_table("themes"):
-            SeedSystemThemesCommand().run()
+        SeedSystemThemesCommand().run()
```
dpgaspar: We do `AppBuilder(update_perms=False)` to avoid needing the db on startup and use the cli to update permissions. We could also use `FAB_UPDATE_PERMS` to control this behaviour also.
Following the same pattern we could create a new config flag to avoid seeding the themes at app startup, and move this to `superset init` for example.

I'm also thinking about `register_sqla_event_listeners` that won't register any listeners if we don't have a request context, and having a request context is not expected on app initialization, so these listeners won't be registered if the db uri contains `["nouser", "nopassword", "nohost", "nodb"]` and `TAGGING_SYSTEM` is `True`

superset/app.py
```diff
@@ -96,6 +100,90 @@ def send_static_file(self, filename: str) -> Response:
...
+        """
+        try:
+            # Import here to avoid circular import issues
+            from superset.extensions import feature_flag_manager
+
+            # Check if database URI is a fallback value before trying to connect
+            db_uri = self.config.get("SQLALCHEMY_DATABASE_URI", "")
+            if not db_uri or any(
+                fallback in db_uri.lower()
+                for fallback in ["nouser", "nopassword", "nohost", "nodb"]
```
dpgaspar: do we still need this?

superset/app.py
```diff
@@ -96,6 +100,90 @@ def send_static_file(self, filename: str) -> Response:
...
+            ):
+                logger.warning(
+                    "Database URI appears to be a fallback value. "
+                    "Skipping database-dependent sync."
+                )
+                return
+
+            # Check if database is up-to-date with migrations
+            if not self._is_database_up_to_date():
```
dpgaspar, line 151: is it still necessary? if superset init's without having the database up to date will it be silently in a broken state?

superset/app.py
mistercrunch: idk! @eschutho I think you might have added this logic

superset/app.py
mistercrunch: I think it's necessary, the issue that prompted all this was that if you were on an old migration (pre Theme model creation), everything would fail, including `superset db upgrade` ...

In general, I think that `sync_config_to_db` requires a check to make sure that the db is up-to-date. If anything, maybe we should even fail harder and only allow you to run `superset db upgrade` or things that really don't depend on the database.

I think this check here makes things better, there's now a clear `sync_to_db()` step that requires an up-to-date database, pretty important assumption before syncing anything...

superset/app.py
mistercrunch: Spoke with @eschutho and it's not needed anymore

dpgaspar (approved):

PR #34680, opened 2025-08-13
fix(bootstrapData): Missing application_root data throws an error

justinpark:
### SUMMARY
When adding the `application_root` bootstrap data in #30134, an issue arose where the frontend would always throw an error if this data was not previously included. This commit fixes the error by referencing the default value when the bootstrap data is not provided, just as it does for other default values.

Before:

_DEV__Superset

After:

_DEV__Superset

### TESTING INSTRUCTIONS

Run superset without configure the `SUPERSET_APP_ROOT`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/utils/getBootstrapData.test.ts
+++ b/superset-frontend/src/utils/getBootstrapData.test.ts
@@ -106,5 +106,32 @@ describe('getBootstrapData and helpers', () => {
       expect(applicationRoot()).toEqual(expectedAppRoot);
       expect(staticAssetsPrefix()).toEqual(expectedStaticPrefix);
     });
+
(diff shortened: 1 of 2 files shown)


justinpark:
cc: @martyngigg

michael-s-molina (approved):

bxter:
when can release a version with support application_root??

@martyngigg

martyngigg:
> when can release a version with support application_root??
> 
> @martyngigg

@bxter While I contributed the feature I'm not a maintainer so I can't say when it will be released I'm afraid. I think v6.0 is about to be cut and it looks like it is in the RC: https://github.com/apache/superset/blob/6.0.0rc1/CHANGELOG/6.0.0.md

Issue #34681, opened 2025-08-13
[Deck.gl aggregated charts] error: ReferenceError: module is not defined


richardfogaca:
### Description
  All deck.gl aggregated chart types (Grid, Contour, Hexagon, Screengrid, Heatmap) fail to render with a ReferenceError in a fresh Superset installation.
  The error message indicates that `module` is not defined, suggesting a potential module loading or bundling issue.

  ### Expected Behavior
  deck.gl aggregated charts should render properly when selected from the chart list.

  ### Actual Behavior
  When opening any deck.gl aggregated chart, the visualization fails with: 
Data error 
An error occurred while rendering the visualization: ReferenceError: module is not defined

  ### Steps to Reproduce
  1. Fresh install of Superset using Docker Compose
  2. Run `docker compose up`
  3. Navigate to Charts list
  4. Select any of the following chart types:
     - Grid
     - Contour
     - Hexagon
     - Screengrid
     - Heatmap
  5. Observe the error message

  ### Screenshots
<img width="1488" height="309" alt="Image" src="https://github.com/user-attachments/assets/076a1362-4586-4a96-86cc-4664d3f1becd" />

  ### Additional Context
  - The issue appears to be specific to deck.gl aggregated charts
  - The error suggests a JavaScript module loading issue, possibly related to webpack configuration or deck.gl library integration
  - This occurs on a clean installation with no custom configurations

  ### Possible Root Cause
  The error "module is not defined" typically indicates:
  - Missing webpack configuration for deck.gl modules
  - CommonJS/ES6 module compatibility issues
  - Missing polyfills or module loaders
  - Incorrect bundling of deck.gl dependencies

### Superset version

master / latest-dev

### Python version

3.10

### Node version

18 or greater

### Browser

Chrome

### Checklist

- [x] I have searched Superset docs and Slack and didn't find a solution to my problem.
- [x] I have searched the GitHub issue tracker and didn't find a similar bug report.
- [x] I have checked Superset's logs for errors and if I found a relevant Python stacktrace, I included it here as text in the "additional context" section.

msyavuz:
This is also happening in Calendar Heatmap charts

amaannawab923:
ReferenceError: module is not defined
    at eval (utils.js:4:151)
    at eval (utils.js:4:161)
[...]

PR #34690, opened 2025-08-14
fix(deck.gl): add webpack rule to define module global for deck.gl charts

richardfogaca:
### Problem
  All deck.gl aggregated chart types (Grid, Contour, Hexagon, Screengrid, Heatmap, Calendar Heatmap) were failing to render with a "module is not defined" error. This occurred because deck.gl's distribution files contain code that references the Node.js `module` global, which doesn't exist in browser environments. This PR fixes issue #34681 

  ### Root Cause
  The deck.gl library's bundled JavaScript files in `node_modules/@deck.gl/*/dist/*.js` contain references to `module.exports`, expecting a CommonJS environment. When webpack processes these files for the browser, the `module` global is undefined, causing a ReferenceError.

  ### Solution
  Added a webpack rule using `imports-loader` to inject `var module = module || {exports: {}};` at the top of all deck.gl and luma.gl JavaScript files. This ensures the `module` object exists before any code tries to access it, preventing the runtime error.

  ### Testing
  - Verified all affected deck.gl aggregated charts (Grid, Contour, Hexagon, Screengrid, Heatmap, Calendar Heatmap) now render correctly
  - Confirmed non-aggregated deck.gl charts (Arc, Path, etc.) continue to work as expected
  - No impact on build performance or bundle size

### DIFF
--- a/superset-frontend/webpack.config.js
+++ b/superset-frontend/webpack.config.js
@@ -363,6 +363,13 @@ const config = {
         },
       },
       {
+        test: /node_modules\/(@deck\.gl|@luma\.gl).*\.js$/,
+        loader: 'imports-loader',
+        options: {
+          additionalCode: 'var module = module || {exports: {}};',
+        },
+      },
+      {
         test: /\.tsx?$/,
         exclude: [/\.test.tsx?$/],
         use: [


amaannawab923:
This is cool , Works smoothly , Nice work @richardfogaca 

Screenshot 2025-08-14 at 5 03 51 PM

msyavuz (approved):
LGTM

PR #34692, opened 2025-08-14
chore(deps): upgrade databricks connectors

drummerwolli:
### SUMMARY

upgrade databricks connectors

sqlalchemy support for databricks was first moved to databricks-sql-connector and then outsourced again into a new package name, hence the rename: 

https://github.com/crflynn/sqlalchemy-databricks?tab=readme-ov-file#%EF%B8%8F-this-project-is-archived-a-new-sqlalchemy-dialect-is-available-directly-from-databricks

image

https://github.com/databricks/databricks-sql-python?tab=readme-ov-file#sqlalchemy

image

#### DEPENDENCY ISSUE

`databricks-sqlalchemy` is available for SQLAlchemy 1 and 2. I pinpointed this version now to 1.0.5 which is the version which is for SQLAlchemy 1.

This version 1.0.5 needs though `pyarrow<17`. pyarrow in Superset was upgrade in #31476 to require at least v18. So we either upgrade SQLAlchemy to v2 (tricky and big, is someone working on this?) or downgrade pyarrow to 16 (which is still an upgrade to v14 which is included in 5.0.0, and the feature required by the PR creator vor v18 is already included in v16). The latter is proposed here: #34693

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF

n/a

### TESTING INSTRUCTIONS
1. install databricks drivers
2. set up a databricks connection and check that it works

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #31476, opened 2024-12-16
chore(deps): Upgrade pyarrow to 18.1.0

phillipleblanc:
### SUMMARY
Upgrade pyarrow to 18.1.0, to support the StringView arrow type.

### TESTING INSTRUCTIONS
Start Superset and verify that the SQL Lab queries are still working.

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #31475
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34693, opened 2025-08-14
chore(deps): downgrade pyarrow to v16

drummerwolli:
### SUMMARY

some database connectors are not supporting the newer pyarrow version (introduced in https://github.com/apache/superset/pull/31476) yet, or require first an upgrade to SQLAlchemy 2.0. see https://github.com/apache/superset/pull/34692

fyi @phillipleblanc

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ dependencies = [
     "python-dateutil",
     "python-dotenv", # optional dependencies for Flask but required for Superset, see https://flask.palletsprojects.com/en/stable/installation/#optional-dependencies
     "python-geohash",
-    "pyarrow>=18.1.0, <19",
+    "pyarrow>=16.1.0, <17",
     "pyyaml>=6.0.0, <7.0.0",
     "PyJWT>=2.4.0, <3.0",
     "redis>=4.6.0, <5.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ dependencies = [
     "python-dateutil",
     "python-dotenv", # optional dependencies for Flask but required for Superset, see https://flask.palletsprojects.com/en/stable/installation/#optional-dependencies
     "python-geohash",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -84,7 +84,7 @@ dependencies = [
     "python-dateutil",
     "python-dotenv", # optional dependencies for Flask but required for Superset, see https://flask.palletsprojects.com/en/stable/installation/#optional-dependencies
     "python-geohash",
-    "pyarrow>=18.1.0, <19",
+    "pyarrow>=16.1.0, <17",
```
sadpandajoe: If we cannot upgrade, can we add a comment on why and what is the condition to lift this range?

sadpandajoe:
By downgrading this, will it initially reintroduce the issue that #31476 was trying to fix?

phillipleblanc:
> By downgrading this, will it initially reintroduce the issue that #31476 was trying to fix?

pyarrow 16 has support for the StringView types that I wanted to upgrade pyarrow for, so this PR shouldn't break that. Thanks @drummerwolli for catching this.

pyproject.toml
drummerwolli: i added a comment. let me know if it is clear.

sadpandajoe (approved):

nigzak:
Hi together @phillipleblanc @sadpandajoe @drummerwolli 

With this downgrade a critical CVE will come back to superset with CVSS V3 CISA-APP score 9.8

It should be may be considered to update to the fixed version V17.0.0 (or newer) and not again downgrade to an affected version?

Hint: I expecially don't name the CVE details here, all actual scanner find them in all current superset versions 4.x (
I did not check V5 yet)

AWS ECR scan marks this a critical finding
JFROG marks it as critical finding
(like this is a well-known finding in pyarrow version < 17.0.0)

PR #34694, opened 2025-08-14
fix(drill-to-detail): ensure axis label filters map to original column names

### LINKED ISSUE #33907: Drill to Detail by Axis value does not work if the Axis column has a label

LisaHusband:
fix(drill-to-detail): map axis labels to original column names in filters

### SUMMARY
This PR fixes an issue where the **Drill to Detail** feature in ECharts charts fails when the X-axis column has a `verbose_name` defined at the dataset level.  
Previously, the frontend would send the column's label as the `filters.col` value in the `/samples` request. The backend did not translate the label back to the original column name, causing the filter condition to be ignored and the query to return all rows.  

Changes in this PR:
- Added `reverse_verbose_map` to the `Datasource` object to map labels (`verbose_name`) back to their original column names.
- Updated `get_samples` to preprocess `payload.filters`, replacing label-based column names with their original names.
- Removed the `grain` parameter for non-temporal columns in Drill to Detail payloads to ensure correct SQL filtering.
- Kept the existing `verbose_map` behavior unchanged so UI display is not affected.

**Before:**  
When drilling to detail on a bar chart, the resulting table displayed all rows instead of filtering by the selected axis value.  

**After:**  
Drill to detail returns only rows matching the selected axis value.

### TESTING INSTRUCTIONS
1. In a dataset, add a `verbose_name(Label)` to a column.
2. Create a bar chart using that column as the X-axis and add a metric.
3. Save the chart to a dashboard.
4. In the dashboard, right-click a bar and choose "Drill to Detail by $AXIS_VALUE".
5. Verify that only rows matching the selected axis value are returned.

### ADDITIONAL INFORMATION
- [x] Has associated issue: Fixes #33907 
- [ ] Required feature flags:
- [ ] Changes UI


superset/views/datasource/utils.py
```diff
@@ -44,6 +44,35 @@ def get_limit_clause(page: Optional[int], per_page: Optional[int]) -> dict[str,
...
+                getattr(col, column_attr)
+                for col in columns
+                if getattr(col, verbose_attr) == f["col"]
+            ),
+            None,
+        )
+        if match:
+            f["col"] = match
+
+
```
msyavuz: Can we get a test for this as well?

LisaHusband: of course

LisaHusband:
@msyavuz  it ready for review.

tests/views/datasource/test_replace_verbose_with_column.py
msyavuz, line 1: Is this a good place for this file? `/tests/unit_tests/datasource/utils/` seem like a better place to put this

msyavuz (commented):
LGTM Other than a small nit!

tests/views/datasource/test_replace_verbose_with_column.py
LisaHusband: really？

LisaHusband: because other func in that util.py  have not tests, but i will change its localtion  later.

msyavuz (approved):
LGTM!

PR #34698, opened 2025-08-14
fix: Invalid error tooltip if control label is function

kgabryje:
### SUMMARY
If a control's label is a function instead of a string (like for example x axis control), the error tooltip at the top of control panel didn't display it properly.

Before:

image

After:

image

### TESTING INSTRUCTIONS
1. Create a new chart with x axis control, like Line Chart
2. Before adding any controls, hover over the error icon next to Data tab and verify that the message is correct

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/ExploreViewContainer/index.jsx
+++ b/superset-frontend/src/explore/components/ExploreViewContainer/index.jsx
@@ -541,7 +541,11 @@ function ExploreViewContainer(props) {
       .map(message => {
         const matchingLabels = controlsWithErrors
           .filter(control => control.validationErrors?.includes(message))
-          .map(control => control.label);
+          .map(control =>
+            typeof control.label === 'function'
+              ? control.label(props.exploreState)
+              : control.label,
+          );
         return [matchingLabels, message];
       })
       .map(([labels, message]) => (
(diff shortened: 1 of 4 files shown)


Antonio-RiveroMartnez (approved):
lgtm, my comment below can be done in a follow up in case we want to pursue it, this PR is fine as a fix from my POV

superset-frontend/src/explore/components/ExploreViewContainer/index.jsx
```diff
@@ -541,7 +541,11 @@ function ExploreViewContainer(props) {
       .map(message => {
         const matchingLabels = controlsWithErrors
           .filter(control => control.validationErrors?.includes(message))
-          .map(control => control.label);
+          .map(control =>
+            typeof control.label === 'function'
```
Antonio-RiveroMartnez, line 545: Not sure how feasible, but if a function, we should call the func before setting the label that way the component doesn't need to be aware of this or any other type and instead "just render", but not sure how worthy that approach would be assuming it would require bigger changes

kgabryje: That's true, it would be better to store the control values after resolution, not before, but that requires some careful planning for the refactor. The current flow makes it difficult to run the function fields earlier (label is not the only one that can be either a function or a simple value)

PR #34711, opened 2025-08-15
feat: refactor modals to use consistent design patterns

mistercrunch:
### before
Screenshot 2025-08-19 at 9 09 12 AM
Screenshot 2025-08-19 at 9 09 04 AM

### after
Screenshot 2025-08-15 at 7 37 28 AM
Screenshot 2025-08-15 at 7 44 28 AM

- Create shared modal components (StandardModal, ModalFormField, useModalValidation)
- Refactor Chart Properties modal to use collapsible sections with validation
- Refactor Dashboard Properties modal to use shared components
- Refactor AlertReportModal to use StandardModal
- Replace 2-column layouts with single-column collapsible sections
- Add real-time validation with visual feedback
- Standardize modal width to 500px for consistent UX
- Improve accessibility and keyboard navigation

Key improvements:
- ~40% code reduction through component reuse
- Consistent styling across all property modals
- Better organization with logical collapsible sections
- Type-safe validation system with error tooltips

Feature flags for testing the ephemeral environment: 
FEATURE_ALERT_REPORTS=True


yousoph:
1. We don't always use the green check mark in the modals - we do that in Alerts & Reports but not in DB connection advanced settings (where everything is not optional). @kasiazjc in the like dashboard properties where some sections have required fields and others don't, should we only show the green check in sections with required sections? 
2. In alerts/reports, the checkmark is replaced with a red asterisk when there is an incomplete required field in the section 
3. The Save button in Alerts/Reports modal is also disabled when a required field is missing. This looks like it's partially working here - it goes disabled after you click it instead of being disabled before you try to click. 
4. Dashboard Name hint text needs updating, or just remove it to match the Chart name field
image

5. The modal is shrinking in height when all sections are collapsed - it should stay the same height. Example: 
image

kasiazjc:
> 1. We don't always use the green check mark in the modals - we do that in Alerts & Reports but not in DB connection advanced settings (where everything is not optional). @kasiazjc in the like dashboard properties where some sections have required fields and others don't, should we only show the green check in sections with required sections?

I would use check/x for sections, it will be easier to spot what is missing. It would be great if we could add a rule, that all required fields should have red `*` next to the label across the app - not sure if it is something we can do easily @mistercrunch ? 

> 4. Dashboard Name hint text needs updating, or just remove it to match the Chart name field
> 
> image

In general I would avoid using metadata for basic hints - can we move it to the field as placeholder? @mistercrunch 

One more note from me (for now) is that errors should be shown in metadata in red under the field. This warning should show up next to the label. 
image

eschutho (approved):

PR #34712, opened 2025-08-15
feat: replace react-color with AntD ColorPicker for theming support

mistercrunch:
Screenshot 2025-08-14 at 6 36 25 PM

- Replace react-color SketchPicker with AntD ColorPicker component
- Create ColorPicker wrapper in @superset-ui/core/components
- Convert ColorPickerControl from JSX to TypeScript
- Maintain RGB color format for backward compatibility
- Preserve color presets from categorical scheme registry
- Update AnnotationLayer to use new ColorPicker
- Remove react-color dependency and webpack workarounds
- Fix TypeScript issues in ContourControl


superset-frontend/src/explore/components/controls/AnnotationLayerControl/AnnotationLayer.jsx
```diff
@@ -835,23 +839,38 @@ class AnnotationLayer extends PureComponent {
...
+          />
+          {color !== AUTOMATIC_COLOR && (
+            <div style={{ marginTop: this.props.theme.sizeUnit * 2 }}>
+              <ControlHeader label={t('Color')} />
+              <ColorPicker
+                value={color}
+                presets={[{ label: 'Theme colors', colors: colorScheme }]}
+                onChange={colorValue =>
+                  this.setState({ color: colorValue.toHexString() })
+                }
```
msyavuz: Should we use onChangeComplete here? I noticed a lag that doesn't exist on the ant design default.
https://ant.design/components/color-picker#color-picker-demo-controlled

https://github.com/user-attachments/assets/18ffdcf2-89a8-4b80-ba3b-7a4963060ac3

mistercrunch: yeah it's got a fair amount to re-render, good catch on finding `onChangeComplete`, adapting to use it!

msyavuz (approved):
LGTM!

PR #34719, opened 2025-08-15
fix: default value in run-server.sh

prochac:
With `set -u`, it was reporting "unbound variable". I believe there is typo, but I don't know bash well

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/docker/entrypoints/run-server.sh
+++ b/docker/entrypoints/run-server.sh
@@ -26,7 +26,7 @@ gunicorn \
     --workers ${SERVER_WORKER_AMOUNT:-1} \
     --worker-class ${SERVER_WORKER_CLASS:-gthread} \
     --threads ${SERVER_THREADS_AMOUNT:-20} \
-    --log-level "${GUNICORN_LOGLEVEL:info}" \
+    --log-level "${GUNICORN_LOGLEVEL:-info}" \
     --timeout ${GUNICORN_TIMEOUT:-60} \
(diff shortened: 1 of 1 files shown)


prochac:
> ## Pull Request Overview
> 
> This PR fixes a shell scripting syntax error in the Gunicorn server configuration that was causing "unbound variable" errors when running with `set -u`. The fix corrects the parameter expansion syntax for the GUNICORN_LOGLEVEL environment variable.
> 
> - Corrects bash parameter expansion syntax from `:info` to `:-info`
> 
> 
> 
> 
Duh

docker/entrypoints/run-server.sh
```diff
@@ -26,7 +26,7 @@ gunicorn \
     --workers ${SERVER_WORKER_AMOUNT:-1} \
     --worker-class ${SERVER_WORKER_CLASS:-gthread} \
     --threads ${SERVER_THREADS_AMOUNT:-20} \
-    --log-level "${GUNICORN_LOGLEVEL:info}" \
+    --log-level "${GUNICORN_LOGLEVEL:-info}" \
```
msyavuz, line 29: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

Here's a reference for this

msyavuz (approved):
LGTM!

PR #34724, opened 2025-08-16
fix(drilling): drill by pagination works with MSSQL data source, cont.

### LINKED ISSUE #24072: MSSQL requires an order_by when using an OFFSET or a non-simple LIMIT clause

sfirke:
Continuing https://github.com/apache/superset/pull/27442 but I force-pushed so can't reopen it.

This is a simpler alternative to https://github.com/apache/superset/pull/34583. My sense is this one is better. Comparing the two:

**This PR #34724:** adds a sort-by the first column to drill-to-detail queries. Changes only drilling queries.
**The other PR #34583**: kept current behavior for all database types that aren't MSSQL - but it adds a dialect check that will affect many more queries.

This fixes #24072. See discussion of #27442 for full context.


sfirke:
@rusackas this is ready for review. If folks agree this is better than #34583, we can close that PR and also close an issue when we merge this.

rusackas (approved):
LGTM (and my friend Claude, after a convo about the issue and these changes)

PR #34725, opened 2025-08-16
fix: standardize menu by removing the caret from dropdown menus

mistercrunch:
Simplify menus by removing carets as it's not well-supported by antd-v5. Decided to maintain the custom animations.


rusackas (approved):

rusackas:
Haven't reviewed the ephemeral yet, but the code looks OK to me!

yousoph:
Not sure if this is coming from this PR or a previous change, but the underline when hovering over a menu item is in the wrong spot for the + and paint roller icons and the menus seem like they're in the wrong spots also 

https://github.com/user-attachments/assets/8cc4d063-50f2-40da-bd86-b252110760cc


cc @kasiazjc

yousoph:
image

msyavuz:
Yeah, i was going to test those usages on eph, but you seemed catch most of them.

There is another issue with the main nav:
Plus icon hover is smaller than the actual item, same goes for other submenus

yousoph:
Another one: 

image

mistercrunch:
It's officially beyond my CSS skills to fix these menus. I may never touch this component again. I think I give up. CSS won this battle.

mistercrunch:
Please someone take over before I throw my computer out the window!

mistercrunch:
css-html

msyavuz:
Ok, i am taking this on 🤞🏾

PR #34759, opened 2025-08-19
fix(echarts): Show full labels in bar chart tooltips

### LINKED ISSUE #31864: Bar Charts - Showing a full name on hover - How?
## Screenshot

Image

## Description

In Bar-Charts we use a short name for the axis / label so 12 characters in the query -  so it the length of the label does not impact the bar charts UX

We want to be able to show the full name for each bar though when they hoover instead so the user can understand what they're looking at. I'm unable to determine how to do this. We've tried multiple solutions and none have worked.

Anyone know how we can show an alternative column or the non shortened name in the hover over .

rusackas:
## Summary
Fixes #31864

When bar chart axis labels are visually truncated by ECharts due to space constraints, the tooltip now shows the full label text by preferring the `axisValue`/`axisValueLabel` properties which contain the complete text.

This enhancement allows users to hover over data points and see the full name even when the axis label is shortened for display purposes.

## Changes
- Modified the tooltip formatter in `transformProps.ts` to use `axisValue` or `axisValueLabel` when available (for axis tooltips)
- Falls back to the original value extraction for compatibility
- Added comprehensive tests for the tooltip behavior with long labels

## Test plan
1. Create a bar chart with long category names (> 12 characters)
2. Observe that axis labels are visually truncated
3. Hover over the bars to see tooltips
4. Verify that tooltips show the full label text

### Unit tests
Added three new test cases:
- Test that tooltips use `axisValue` when available (richTooltip mode)
- Test fallback behavior when `axisValue` is not present
- Test item tooltips (non-richTooltip mode)

All tests pass successfully.

## Screenshots
N/A - This is a tooltip behavior change that doesn't affect the visual appearance of the chart itself.

### DIFF
--- a/superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
@@ -1520,3 +1520,99 @@ test('should assign distinct dash patterns for multiple time offsets consistentl
   // must be different patterns
   expect(symbol1).not.toEqual(symbol2);
 });
+
+describe('Tooltip with long labels', () => {
+  test('should use axisValue for tooltip when available (richTooltip)', () => {
+    const longLabelData: ChartDataResponseResult[] = [
+      createTestQueryData([
+        {
+          'This is a very long category name that would normally be truncated': 100,
+          __timestamp: 599616000000,
+        },
+        {
+          'Another extremely long category name for testing purposes': 200,
+          __timestamp: 599916000000,
+        },
+      ]),
+    ];
+
+    const chartProps = createTestChartProps({
+      formData: {
+        richTooltip: true,
+      },
+      queriesData: longLabelData,
+    });
+
+    const transformedProps = transformProps(chartProps);
+
+    // Get the tooltip formatter function
+    const tooltipFormatter = (transformedProps.echartOptions as any).tooltip
+      .formatter;
+
+    // Simulate params from ECharts with axisValue containing full label
+    // Use distinct values for axisValue and seriesName to verify axisValue is used
+    const mockParams = [
+      {
+        axisValue:
+          'This is a very long category name that would normally be truncated',
+        value: [599616000000, 100],
+        seriesName: 'Some Series Name',
+      },
+    ];
+
+    // Call the formatter and check it uses the full label from axisValue
+    const result = tooltipFormatter(mockParams);
+    expect(result).toContain(
+      'This is a very long category name that would normally be truncated',
+    );
+  });
+
+  test('should fallback to value when axisValue is not available', () => {
+    const chartProps = createTestChartProps({
+      formData: {
+        richTooltip: true,
+      },
+    });
+
+    const transformedProps = transformProps(chartProps);
+
+    const tooltipFormatter = (transformedProps.echartOptions as any).tooltip
+      .formatter;
+
+    // Simulate params without axisValue
+    const mockParams = [
+      {
+        value: [599616000000, 1],
+        seriesName: 'San Francisco',
+      },
+    ];
+
+    // Should still work with fallback to value
+    const result = tooltipFormatter(mockParams);
+    expect(result).toBeDefined();
+    expect(typeof result).toBe('string');
+  });
+
+  test('should handle item tooltips correctly', () => {
+    const chartProps = createTestChartProps({
+      formData: {
+        richTooltip: false,
+      },
+    });
+
+    const transformedProps = transformProps(chartProps);
+
+    const tooltipFormatter = (transformedProps.echartOptions as any).tooltip
+      .formatter;
+
+    // For item tooltips, params is a single object
+    const mockParams = {
+      value: [599616000000, 1],
+      seriesName: 'San Francisco',
+    };
+
+    const result = tooltipFormatter(mockParams);
+    expect(result).toBeDefined();
+    expect(typeof result).toBe('string');
+  });
+});
--- a/superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
+++ b/superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
@@ -1724,10 +1724,10 @@ describe('Tooltip with long labels', () => {
       },
     ];
 
-    // Should still work with fallback to value
+    // Should fall back to the x-value (value[xIndex]) and render it in the title
     const result = tooltipFormatter(mockParams);
-    expect(result).toBeDefined();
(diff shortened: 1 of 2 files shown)


EROGE-lab:
Is this fixed with this request?
Screenshot 2025-09-12 172201
Screenshot 2025-09-12 172038

The tooltip display is unprofessional. Also, a percentage with two decimal places is not appealing. One decimal place would suffice. This cannot be configured. Superset 5.0

rusackas:
## PR Assessment and Rebase

### Issue Status
The linked issue #31864 was **closed as "NOT_PLANNED"** on September 9, 2025, because:
- The team couldn't reproduce the issue
- They were focusing on 6.0 development

### However, This PR May Still Be Valuable

**The fix is minimal and low-risk:**
```typescript
// Before: Only uses value array
const xValue = params[0].value[xIndex]

// After: Prefers full label from axisValue/axisValueLabel
const xValue = params[0].axisValue ?? params[0].axisValueLabel ?? params[0].value[xIndex]
```

**Why this matters:**
- When ECharts truncates axis labels due to space constraints, the `value` array contains the truncated text
- The `axisValue`/`axisValueLabel` properties contain the **full, untruncated text**
- This ensures tooltips show complete labels even when axis labels are shortened

**Evidence of continued need:**
- @EROGE-lab commented on this PR (Sept 12, 2025) asking about tooltip display issues
- The issue had multiple users reporting similar problems before it was closed

### Recommendation

Consider merging this PR despite the closed issue because:
1. **Low risk** - Simple fallback logic with null coalescing
2. **Good UX** - Users can see full labels on hover when axis labels are truncated  
3. **Well tested** - Includes comprehensive test coverage
4. **No breaking changes** - Falls back to existing behavior if `axisValue` is unavailable

### Technical Notes
- Pre-existing type errors in `transformProps.ts` exist on master (not introduced by this PR)
- The PR rebases cleanly on current master
- Tests are comprehensive and cover the new behavior

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
```diff
@@ -642,6 +642,116 @@ describe('Does transformProps transform series correctly', () => {
...
+            'This is a very long category name that would normally be truncated',
+        },
+      ];
+
+      // Call the formatter and check it uses the full label
+      const result = tooltipFormatter(mockParams);
+      expect(result).toContain(
+        'This is a very long category name that would normally be truncated',
+      );
+    });
```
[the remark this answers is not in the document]
rusackas: Addressed in 5f4e772. The test now uses distinct values for `axisValue` (the full category label) and `seriesName` ('Series A') to properly verify that `axisValue` is used for the tooltip title. The assertion specifically checks for the `fullCategoryLabel` which only comes from `axisValue`, not from `seriesName`.

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
```diff
@@ -808,6 +808,114 @@ describe('Does transformProps transform series correctly', () => {
       'foo2, bar2': ['foo2', 'bar2'],
     });
   });
+
+  describe('Tooltip with long labels', () => {
```
EnxDev: Would it make sense to avoid using describe here?

EnxDev (commented):
Do you think some tests should better reflect real user behavior, for example: 

`await userEvent.hover(bar);

expect(
  screen.getByText('This is a very long category name...')
).toBeInTheDocument();`

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
rusackas: fixed!

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
```diff
@@ -1520,3 +1520,99 @@ test('should assign distinct dash patterns for multiple time offsets consistentl
   // must be different patterns
   expect(symbol1).not.toEqual(symbol2);
 });
+
+describe('Tooltip with long labels', () => {
```
EnxDev, line 1661: Could it make sense to avoid using `describe` here?

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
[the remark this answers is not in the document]
rusackas: Thanks — no action needed.

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
```diff
@@ -1520,3 +1520,99 @@ test('should assign distinct dash patterns for multiple time offsets consistentl
...
+      {
+        value: [599616000000, 1],
+        seriesName: 'San Francisco',
+      },
+    ];
+
+    // Should still work with fallback to value
+    const result = tooltipFormatter(mockParams);
+    expect(result).toBeDefined();
+    expect(typeof result).toBe('string');
```
aminghadersohi, line 1729: These assertions only check that the result is a string — they never verify the formatted x-value actually appears, so a regression in the fallback path would still pass. Assert the value is present (as the `axisValue` test does with `toContain`). Same applies to `should handle item tooltips correctly` (L1615-1616).

aminghadersohi (approved):

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
rusackas: Good catch — fixed in 9ff0ed8. Both fallback tests now assert that the x-value (`599616000000`) actually appears in the rendered tooltip title, so a regression in the fallback path would be caught rather than silently passing on the `typeof === 'string'` check alone.

superset-frontend/plugins/plugin-chart-echarts/test/Timeseries/transformProps.test.ts
rusackas: I kept the `describe` block here intentionally for consistency — this test file already groups its suites with `describe` throughout (e.g. `EchartsTimeseries transformProps`, `legend sorting`, etc.), and the `Tooltip with long labels` group keeps these related cases together. Converting the whole file to flat `test()` calls would be a broader, file-wide refactor outside the scope of this tooltip fix, so I will leave it matching the surrounding convention.

PR #34760, opened 2025-08-19
fix(charts): Handle virtual dataset names without schema prefix correctly

### LINKED ISSUE #33335: The dataset is not showing up in the chart list if it's virtual and not in the correct . format
## Screenshot
WHAT IS HAPPENING:
Image
Image
EXPECTED:

Image
Image

## Description
The dataset column is giving weird displays of names if it's virtual and has a weird format.
I think it's a mix of these 2 playing weird. 
Image
Image

rusackas:
## Summary

Fixes #33335

Virtual datasets with names that don't contain a period (no schema prefix) were not displaying correctly in the chart list. The dataset name would disappear entirely instead of showing the full name.

### Root Cause
PR #29944 introduced logic to show only the dataset name (without schema) in the chart list for better readability. However, the implementation assumed all dataset names would follow the `schema.name` format and used `split('.')[1]` to extract the name. When a dataset had no schema (no period in the name), this would return `undefined`, causing the name to disappear.

### Solution
Updated the dataset name extraction logic to:
- Check if the name contains a period
- If yes: extract everything after the first period (handles names with dots correctly)
- If no: use the full name as-is

## Test Plan

✅ Added comprehensive unit tests covering:
- Dataset names with schema prefix (`public.test_dataset`)
- Dataset names without schema (`Jinja 5`)  
- Dataset names containing dots (`schema.table.with.dots`)

✅ All existing tests pass
✅ Pre-commit checks pass

## Manual Testing
1. Create a virtual dataset with a name like "Jinja 5" (no period)
2. Navigate to the Charts list
3. Verify the dataset name displays correctly in the Dataset column
4. Verify the tooltip shows the full name on hover

### DIFF
--- a/superset-frontend/src/pages/ChartList/ChartList.listview.test.tsx
+++ b/superset-frontend/src/pages/ChartList/ChartList.listview.test.tsx
@@ -93,6 +93,84 @@ describe('ChartList - List View Tests', () => {
     });
   });
 
+  it('correctly displays dataset names with and without schema', async () => {
+    // Create custom mock data with different datasource_name_text formats
+    const customMockCharts = [
+      {
+        ...mockCharts[0],
+        id: 100,
+        slice_name: 'Chart with Schema',
+        datasource_name_text: 'public.test_dataset',
+      },
+      {
+        ...mockCharts[0],
+        id: 101,
+        slice_name: 'Chart without Schema',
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
LGTM

PR #34330, opened 2025-07-26
feat: conditional formatting improvements in tables

SBIN2010:
This PR extends conditional formatting for string values for a table  #22790 

Снимок экрана от 2025-07-26 23-19-06

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34762, opened 2025-08-19
feat: conditional formatting improvements add flag toAllRow and  toTextColor in tables

SBIN2010:
This PR is a continuation #34330 of 
to expand the possibilities of conditional formatting

### SUMMARY

Added flags :
toAllRow - to apply formatting to the entire row
toTextColor - to apply formatting to the text color

Added migration to add flags to conditional formatting values config

Снимок экрана от 2025-08-19 22-09-09

### TESTING INSTRUCTIONS

run test  FormattingPopoverContent.test.tsx and TableChart.test.tsx

Migrations were tested in batches of 100:
upgrade:
Снимок экрана от 2025-08-21 22-57-14

downgrade:
Снимок экрана от 2025-08-21 22-56-01

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/migrations/versions/2025-08-13_19-09_da125679ebb8_fix_table_chart_conditional_formatting_.py
```diff
@@ -0,0 +1,107 @@
...
+    id = Column(Integer, primary_key=True)
+    viz_type = Column(String(250))
+    params = Column(Text)
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = db.Session(bind=bind)
+
+    for slc in session.query(Slice).filter(Slice.viz_type == "table"):
```
mistercrunch: we're always a little scared of mass-updates to potentially large table. I think we have utilities in other database migrations to make that safer (batching/failure-handling, ...)

mistercrunch: haven't looked at the details, but It's good if there's a way to avoid having updating previous chart, like handling the change in the plugin for instance.

SBIN2010: [mistercrunch](https://github.com/mistercrunch) 
I made improvements in migration:
1) Processing records in small batches (1000 records) instead of loading all data at once
2) Detailed logging of the migration process
3) Guaranteed session closure using the finally block

SBIN2010:
@mistercrunch @rusackas 
Do you have any further comments on this PR?

rusackas (approved):
Looks good to me!

superset/migrations/versions/2025-08-13_19-09_da125679ebb8_fix_table_chart_conditional_formatting_.py
```diff
@@ -52,56 +54,156 @@ def upgrade():
...
+                    break
+
+                for slc in slices_batch:
+                    try:
+                        params = json.loads(slc.params)
+                        conditional_formatting = params.get(
+                            "conditional_formatting", []
+                        )
+
+                        if conditional_formatting:
```
mistercrunch, line 164: when getting to >8 indents in it usually time means it's to write a method

superset/migrations/versions/2025-08-13_19-09_da125679ebb8_fix_table_chart_conditional_formatting_.py
```diff
@@ -52,56 +54,156 @@ def upgrade():
...
+                            new_conditional_formatting = []
+                            for formatter in conditional_formatting:
+                                if (
+                                    "toAllRow" in formatter
+                                    or "toTextColor" in formatter
+                                ):
+                                    new_formatter = formatter.copy()
+                                    new_formatter.pop("toAllRow", None)
+                                    new_formatter.pop("toTextColor", None)
```
mistercrunch: when getting to >8 indents in it usually time to write a method

SBIN2010: Corrected

rusackas:
I think people are just a bit nervous about the need for a mass migration here. Is there a way to handle it gracefully in a forward-compatible way on the frontend only?

SBIN2010:
> I think people are just a bit nervous about the need for a mass migration here. Is there a way to handle it gracefully in a forward-compatible way on the frontend only?

Hi @rusackas !

In my opinion, there are three possible solutions:
1) The method presented in this PR using migrations
2) Create a stub in plugin-chart-table/src/controlPanel.tsx - check if the object already has these flags and, if not, add them
```
                const value = _?.value ?? [];
                if (value && Array.isArray(value)) {
                  value.forEach((item: ConditionalFormattingConfig) => {
                    if (
                      item.colorScheme &&
                      !['Green', 'Red'].includes(item.colorScheme)
                    ) {
                      if (!item.toAllRow) {
                        item.toAllRow = false;
                      }
                      if (!item.toTextColor) {
                        item.toTextColor = false;
                      }
                    }
                  });
                }
```
3) Leave existing conditional formatting objects alone, and add the flags only to new objects

rusackas:
I asked Claude about it, as well as a fourth option - doing it in transformProps. I think that makes more sense than adding the logic to the control panel, but I'm curious your opinion. Here's the [LLM thread](https://claude.ai/share/be55d469-32eb-41fd-938a-ebae71795ed0), for what it's worth.

SBIN2010:
> I asked Claude about it, as well as a fourth option - doing it in transformProps. I think that makes more sense than adding the logic to the control panel, but I'm curious your opinion. Here's the [LLM thread](https://claude.ai/share/be55d469-32eb-41fd-938a-ebae71795ed0), for what it's worth.

I agree with you that the best option without migrations would be to add processing to transformProps.
However, this option doesn't solve the current problem, which can be formulated as follows: the missing flags need to be added when opening the conditional formatting "card."
That's why I suggested adding flag handling to the controlPanel, which ensures that new flags are transferred to the conditional formatting "cards".

For clarity, I'll attach screenshots of the processing in transformProps.
Снимок экрана от 2025-09-29 22-18-13

and controlPanel
Снимок экрана от 2025-09-29 11-15-31

SBIN2010:
Hi @mistercrunch @rusackas !
I've removed migrations and added change processing to the plugin.

superset-frontend/src/explore/components/controls/ConditionalFormattingControl/FormattingPopoverContent.tsx
```diff
@@ -339,6 +370,47 @@ export const FormattingPopoverContent = ({
           </Row>
         )}
       </FormItem>
+      <Row style={{ display: 'flex', height: '40px' }}>
```
rusackas: Do we need the CSS here? We're hoping that using "vanilla" AntD components means we don't need to provide style overrides... but if we do, maybe they should be props for the Row component so that they're used systematically wherever needed?

superset-frontend/src/explore/components/controls/ConditionalFormattingControl/FormattingPopoverContent.tsx
```diff
@@ -246,6 +255,28 @@ export const FormattingPopoverContent = ({
...
+  );
+
+  const showToColorText = useMemo(
+    () =>
+      conditionalFormattingFlag && conditionalFormattingFlag.toColorTextCheck
+        ? config?.toTextColor === undefined
+        : config?.toTextColor !== undefined,
+    [conditionalFormattingFlag, config],
+  );
+
```
rusackas, line 284: These two blocks are nearly identical... should we combine theminto a single method?  ¯\\\_(ツ)_/¯

superset-frontend/src/explore/components/controls/ConditionalFormattingControl/FormattingPopoverContent.tsx
```diff
@@ -339,6 +370,47 @@ export const FormattingPopoverContent = ({
...
+                valuePropName="checked"
+                initialValue={toAllRow}
+              >
+                <Checkbox
+                  onChange={event => setToAllRow(event.target.checked)}
+                  checked={toAllRow}
+                />
+              </FormItem>
+            </Col>
+            <Col style={{ padding: '5px' }}>
```
rusackas: Same here... we shouldn't need to be changing the padding of Col components in multiple places. If it doesn't work as "vanilla" then we should see if the Col component itself need adjustment (or a new prop for certain contexts).

rusackas:
@mistercrunch the `showtime-blocked` tag is new to me... what's that indicate, just so I/we know?

rusackas:
Looking much better! Only one bug left to squash, as far as I can find :)

You can test on this table here, if it helps:
http://35.87.154.148:8080/explore/?slice_id=122

The issue is that the "Show cell bars" option does not show the bars at all _unless_ you also check the "Add colors to cell bars" option. 

I would expect you could still see the original gray bars if that were the only thing checked.

SBIN2010:
> Looking much better! Only one bug left to squash, as far as I can find :)
> 
> You can test on this table here, if it helps: http://35.87.154.148:8080/explore/?slice_id=122
> 
> The issue is that the "Show cell bars" option does not show the bars at all _unless_ you also check the "Add colors to cell bars" option.
> 
> I would expect you could still see the original gray bars if that were the only thing checked.

Thanks for the feedback, but perhaps it would be worthwhile to make corrections to the display of bars in the table in a separate PR?

rusackas:
Agreed... that issue is on `master` it seems.... let's tackle it separately.

PR #34763, opened 2025-08-19
fix: Upload CSV as Dataset

### LINKED ISSUE #31813: Upload forms have non-actionable error message
### Bug description

When trying to upload a CSV file to Superset, if there's a problem with the data such as a numeric column containing an empty value or a data column with wrong formatting, the modal shows `Database upload file failed` or `Table already exists` without any context about the error. This message does not help users to find what's the problem with their CSV file resulting in a very poor user experience and frequent on-call requests.

The error should contain more information about what the problem was such as the column, row and type of problem.

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

luizotavio32:
### SUMMARY
This PR adds improvements to the error messages shown on the Import CSV feature

 🐛  When uploading csv with mismatching types in one of the columns, the error shown is not actionable at all:

image

image
<br>
<br>

🔧 Solution: Error provides column type information and highlights the exact location of the invalid value. It will show the first 5 errors, the remaining will be hidden to not affect the payload size:

image

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #31813 
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/features/databases/UploadDataModel/index.tsx
```diff
@@ -222,6 +222,7 @@ const UploadDataModal: FunctionComponent<UploadDataModalProps> = ({
 }) => {
   const [form] = Form.useForm();
   const [currentDatabaseId, setCurrentDatabaseId] = useState<number>(0);
+  const [databases, setDatabases] = useState<any[]>([]);
```
michael-s-molina: You don't need this additional state. Just change `currentDatabaseId` to `currentDatabase` instead.

luizotavio32: Created an interface/object for currentDatabase, containing the id (value) and label (database_name)

superset-frontend/src/features/databases/UploadDataModel/UploadDataModal.test.tsx
```diff
@@ -661,3 +661,91 @@ describe('UploadDataModal Collapse Tabs', () => {
...
+        },
+        {
+          id: 2,
+          database_name: 'Regular Database',
+          engine_name: 'postgresql',
+        },
+      ],
+    },
+    { overwriteRoutes: true },
+  );
```
justinpark: if you want the same mock each case, you could do with 
```
beforeEach(() => {
   fetchMock.get(...);
});

afterEach(() => {
  fetchMock.reset();
});
```

superset-frontend/src/features/databases/UploadDataModel/index.tsx
```diff
@@ -329,6 +337,18 @@ const UploadDataModal: FunctionComponent<UploadDataModalProps> = ({
...
+  const validateTableName =
+    (currentDatabase: CurrentDatabase) => (_: any, value: string) => {
+      const isPrestoTrino = currentDatabase?.label
+        ?.toLowerCase()
+        .includes('trino');
+
+      if (isPrestoTrino && value.includes(' ')) {
+        return Promise.reject(t('Trino table names must not contain spaces'));
+      }
+      return Promise.resolve();
```
justinpark: It doesn't seem ideal to explicitly check for specific cases. Instead, I think it's better to catch this issue during the backend process that executes the 'CREATE TABLE' statement and deliver it as an error message.

luizotavio32: So do you think it's better to run the name validation after the upload button is pressed ?

superset/commands/database/uploaders/base.py
```diff
@@ -124,6 +125,8 @@ def _dataframe_to_database(
                 df,
                 to_sql_kwargs=to_sql_kwargs,
             )
+        except SupersetException:
+            raise
```
luizotavio32: I'm adding this raise here because some code overrides raises errors using SupersetException class rather than ValueError.

superset-frontend/src/features/databases/UploadDataModel/index.tsx
justinpark: that's right

superset/commands/database/uploaders/base.py
```diff
@@ -133,7 +133,8 @@ def _dataframe_to_database(
                 )
             ) from ex
         except Exception as ex:
-            raise DatabaseUploadFailed(exception=ex) from ex
+            message = ex.message if hasattr(ex, "message") and ex.message else str(ex)
+            raise DatabaseUploadFailed(message=message, exception=ex) from ex
```
luizotavio32, line 137: Handling the exception here will help us catch errors raised inside of databases code overrides `df_to_sql()` method. It will return a more actionable error rather than `Database upload failed`

superset/commands/database/uploaders/csv_reader.py
```diff
@@ -123,6 +123,96 @@ def _select_optimal_engine() -> str:
...
+    ) -> pd.Series:
+        """Find invalid values for non-numeric type conversion."""
+        invalid_mask = pd.Series([False] * len(df), index=df.index)
+        for idx, value in df[column].items():
+            if pd.notna(value):
+                try:
+                    pd.Series([value]).astype(dtype)
+                except (ValueError, TypeError):
+                    invalid_mask[idx] = True
+                    break
```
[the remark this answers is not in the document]
luizotavio32: The idea here is exactly finding the first invalid value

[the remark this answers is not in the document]
michael-s-molina: @luizotavio32 Would be possible to show multiple errors to improve the user experience? I would like to avoid a scenario where the user needs to try to upload the file multiple times because there are multiple errors.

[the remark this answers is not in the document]
luizotavio32: It's possible. In this case I would suggest not to show the exact line and value that has a typing error, only something like `"There are xx non int values on column yy" `. What do you think?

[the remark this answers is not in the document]
michael-s-molina: The exact line and value is really important to help the user fix the issue. It can be a list of error messages.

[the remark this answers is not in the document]
luizotavio32: Maybe we should define a limit of how many errors to output to the user, imagine a column with hundreds of wrong type values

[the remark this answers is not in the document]
michael-s-molina: Yep. You can use something like
- Error 1
- Error 2
- Error 3
- and more 53 errors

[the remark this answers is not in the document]
luizotavio32: Is that what you're thinking ?
image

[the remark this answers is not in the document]
michael-s-molina: Yes!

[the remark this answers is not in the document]
luizotavio32: Ok. Just committed the changes

superset/config.py
```diff
@@ -1328,6 +1328,13 @@ def allowed_schemas_for_csv_upload(  # pylint: disable=unused-argument
 # Smaller values use less memory but may be slower for large files
 READ_CSV_CHUNK_SIZE = 1000
 
+# Maximum number of detailed errors to display in CSV upload error messages
+# When there are more errors than this limit, the system will show the first
+# CSV_UPLOAD_MAX_ERRORS_DISPLAYED errors with details and indicate how many
+# additional errors exist (e.g., "and 5 more error(s)")
+CSV_UPLOAD_MAX_ERRORS_DISPLAYED = 3
```
michael-s-molina: @luizotavio32 I think the number of max errors is based on UI real state and should be a static value to avoid adding another configuration option.

michael-s-molina: @luizotavio32 Could you also update the screenshot in the PR description with an example of the multi error message?

michael-s-molina (approved):

PR #34767, opened 2025-08-19
feat(sqllab): introduce splitter for adjusting sidebar and query panel

justinpark:
### SUMMARY
Previously, the resizable sidebar and react-splitter were managed separately, but now panel resizing is handled through the splitter provided by Ant Design.
Additionally, this commit updates the functionality so that, besides the existing menu, user can collapse or expand the panel using the buttons provided by the splitter.

Furthermore, since the results height is already calculated using [autosizer](https://github.com/apache/superset/pull/34683/files#diff-407b1e72d7f4462c85eca33d215cd9981b18db050f649dfaaac5748a262d09b8R701-R715), this commit removes related height calculation logic, making the size management simpler.

After:

https://github.com/user-attachments/assets/57301174-489b-47ce-bac4-d9dd8626cdac

### TESTING INSTRUCTIONS
Open SQL Lab and then adjust side panel and south panel sizes

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/constants.ts
+++ b/superset-frontend/src/SqlLab/constants.ts
@@ -66,13 +66,9 @@ export const TIME_OPTIONS = [
 ];
 
 // SqlEditor layout constants
-export const SQL_EDITOR_GUTTER_HEIGHT = 5;
-export const SQL_EDITOR_GUTTER_MARGIN = 3;
-export const SQL_TOOLBAR_HEIGHT = 51;
+export const SQL_EDITOR_GUTTER_HEIGHT = 4;
 export const SQL_EDITOR_LEFTBAR_WIDTH = 400;
-export const SQL_EDITOR_PADDING = 10;
 export const INITIAL_NORTH_PERCENT = 30;
-export const INITIAL_SOUTH_PERCENT = 70;
 export const SET_QUERY_EDITOR_SQL_DEBOUNCE_MS = 2000;
 export const VALIDATION_DEBOUNCE_MS = 600;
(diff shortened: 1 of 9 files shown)


rusackas:
This _looks_ a little different from the results/query splitter. is it using the same components?

justinpark:
> This _looks_ a little different from the results/query splitter. is it using the same components?

It's the same component, but it customizes the styling to match the previous divider(handler) UI.

superset-frontend/src/SqlLab/constants.ts
```diff
@@ -66,13 +66,9 @@ export const TIME_OPTIONS = [
 ];
 
 // SqlEditor layout constants
-export const SQL_EDITOR_GUTTER_HEIGHT = 5;
-export const SQL_EDITOR_GUTTER_MARGIN = 3;
-export const SQL_TOOLBAR_HEIGHT = 51;
+export const SQL_EDITOR_GUTTER_HEIGHT = 4;
```
rusackas, line 69: @msyavuz I wonder if (in another PR) some of these consts should be consider for promotion to theming tokens (or just... using tokens we have available instead of setting them here at all).

rusackas (approved):
LGTM!

superset-frontend/src/SqlLab/constants.ts
msyavuz: I would go for using the existing tokens here, but not opposed to extending Superset specific tokens either.

PR #34772, opened 2025-08-20
feat(color picker): enhance color picker to easier alpha changes

Abhirup-99:
### SUMMARY
This pull request updates the `ColorPicker` component to match the previous color picker UI.

Old Design
Screenshot from 2025-08-20 16-44-52

New Design
Screenshot from 2025-08-20 16-44-08

### TESTING INSTRUCTIONS

Check the screenshot difference

### ADDITIONAL INFORMATION

- [x] Has associated issue: fixes https://github.com/apache/superset/issues/34721
- [ ] Required feature flags:
- [X] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/components/ColorPicker/index.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/components/ColorPicker/index.tsx
@@ -20,10 +20,33 @@ import {
   ColorPicker as AntdColorPicker,
   type ColorPickerProps as AntdColorPickerProps,
 } from 'antd';
+import { css, Global } from '@emotion/react';
 
 // Re-export the AntD ColorPicker as-is for themeable usage
 export type ColorPickerProps = AntdColorPickerProps;
-export const ColorPicker = AntdColorPicker;
+export const ColorPicker: React.FC<ColorPickerProps> = props => (
+  <>
+    <Global
+      styles={css`
+        .ant-slider.ant-color-picker-slider.ant-slider-horizontal {
+          margin: 11px 5px;
+        }
+        .ant-color-picker
+          .ant-color-picker-input-container
+          .ant-color-picker-slider-alpha,
+        .ant-color-picker
+          .ant-color-picker-input-container
+          .ant-color-picker-alpha-input,
+        .ant-color-picker
+          .ant-color-picker-input-container
+          .ant-color-picker-format-select {
+          display: inline-flex;
+        }
+      `}
+    />
+    <AntdColorPicker {...props} />
+  </>
+);
 
 // Export RGB color type for backward compatibility
 export type RGBColor = {


Abhirup-99:
fixes #34721

superset-frontend/packages/superset-ui-core/src/components/ColorPicker/index.tsx
```diff
@@ -20,10 +20,33 @@ import {
...
+        .ant-color-picker
+          .ant-color-picker-input-container
+          .ant-color-picker-format-select {
+          display: inline-flex;
+        }
+      `}
+    />
+    <AntdColorPicker {...props} />
+  </>
+);
```
msyavuz, line 49: Global styles for this is not the right approach, you can use css prop but i would suggest inspecting why it doesn't behave like this in antd itself without customization

msyavuz (changes requested):

mistercrunch:
we love to maintain vanilla antd as much as possible, wondering why antd's component isn't what we want it here, seemed you're mostly spacing out the 2 siders? Can this be addressed in core antd? or a Superset-specific styles interfering here?

Mmmh just looked at antd and it looks solid, wondering if it's a version thing or other types interfering, if so we should remove/delete the interference. Less styles is better!

Abhirup-99:
@msyavuz @mistercrunch the styles are coming from the geostyler library inside node modules.
```
.ant-color-picker .ant-color-picker-slider-container {
  padding-bottom: 6px;
}
.ant-color-picker .ant-color-picker-slider-container .ant-color-picker-slider-group {
  align-self: end;
}
.ant-color-picker .ant-color-picker-slider-alpha,
.ant-color-picker .ant-color-picker-alpha-input,
.ant-color-picker .ant-color-picker-format-select {
  display: none;
}
```
these are unscoped styles inside https://github.com/geostyler/geostyler/blob/a5960db5592d1661fdcf1bbc6dc234bf97c383ee/src/Component/Symbolizer/Field/ColorField/ColorField.css#L26

Antd doesn't allow adding a classname to the popover which is created when the color picker is clicked. https://ant.design/components/color-picker

An alternative could be to scope the entire panel under a div using panelrender  and style the div with the changes with we are looking for.

```
panelRender	Custom Render Panel	(panel: React.ReactNode, extra: { components: { Picker: FC; Presets: FC } }) => React.ReactNode	-	5.7.0
```

or ask the geostyles team to do the same for their colorpicker CSS properties.

mistercrunch:
> these are unscoped styles 

oh wow, we should report that on their repo ... 

I'm not good at CSS puzzles, maybe @rusackas can work some magic?

Abhirup-99:
@rusackas can you check this out?

msyavuz:
That is certainly an oversight in geostyler, [reported this there.](https://github.com/geostyler/geostyler/issues/2631). If we can we should probably fix it there and update our geostyler version here. Also just realized the really problematic styles don't exist so updating geostyler should fix some issues already

msyavuz:
Thanks to helpful people over at geostyler, i fixed this upstream. We should just update our version now. @Abhirup-99

PR #34777, opened 2025-08-20
fix(dashboard): enable undo/redo buttons for layout changes

gabotorresruiz:
### SUMMARY

Fixed a bug where undo/redo buttons in dashboard edit mode weren't getting enabled when adding or removing dashboard components. The issue was caused by a known bug in the redux-undo library where the `includeAction` filter incorrectly filters out actions that should be tracked.

**Root Cause**: Redux-undo's filter functionality is broken in multiple versions (including the beta version used by Superset), causing layout actions like `DELETE_COMPONENT` to be incorrectly excluded from undo history.

**Solution**: Implemented reducer-level filtering by wrapping the base `dashboardLayout` reducer before passing it to redux-undo. This approach:
- Prevents non-layout actions from reaching the reducer (no state change = no undo entry)
- Allows redux-undo to work without filtering (which works correctly)
- Maintains proper action filtering while avoiding the redux-undo bug
- Is well-documented with references to upstream bug reports for future removal

Before:
undo_redo_before

After:
undo_redo_after

### TESTING INSTRUCTIONS

1. Open a dashboard in edit mode
2. Add a new chart component to the dashboard
3. Verify the undo button becomes enabled
4. Delete a chart component from the dashboard
5. Verify the undo button becomes enabled
6. Click undo to verify the action is reversed
7. Verify the redo button becomes enabled after undoing
8. Click redo to verify the action is re-applied

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/Header/Header.test.tsx
+++ b/superset-frontend/src/dashboard/components/Header/Header.test.tsx
@@ -18,19 +18,17 @@
  */
 import * as redux from 'redux';
 import { useUnsavedChangesPrompt } from 'src/hooks/useUnsavedChangesPrompt';
-import {
-  render,
-  screen,
-  fireEvent,
-  userEvent,
-  within,
-} from 'spec/helpers/testing-library';
+import { screen, userEvent, within, waitFor } from '@superset-ui/core/spec';
+import { ActionCreators as UndoActionCreators } from 'redux-undo';
 import fetchMock from 'fetch-mock';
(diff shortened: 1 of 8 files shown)


rusackas:
Can we add tests? If it broke once, it might break again :)

gabotorresruiz:
@rusackas yep, I'm on it

eschutho (approved):

PR #34782, opened 2025-08-21
fix: Add dataset ID to file name on exports

Vitor-Avila:
### SUMMARY
When generating an export, Superset would add the asset ID to the chart and dashboard file names, however for datasets it would use only the dataset name. This is problematic because it's possible to have multiple datasets with the same name as long as they have a different schema, catalog, or are from a different DB connection.

As a result, if two datasets with the same name are exported, only one (likely the last one to be generated) is included in the ZIP file. 

This PR fixes this issue by adding the dataset ID to the file name (which is already the case for dashboards and charts).

No UI changes.

### TESTING INSTRUCTIONS
1. Make sure you have 2 DB connections.
2. Access SQL lab and create two virtual datasets using `select 1 as one` with the same name (one from each DB connection).
3. Access the **Datasets** menu and export both datasets via the bulk export. 
4. Confirm both datasets are present in the ZIP file.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/commands/dataset/export.py
+++ b/superset/commands/dataset/export.py
@@ -46,7 +46,7 @@ class ExportDatasetsCommand(ExportModelsCommand):
         db_file_name = get_filename(
             model.database.database_name, model.database.id, skip_id=True
         )
-        ds_file_name = get_filename(model.table_name, model.id, skip_id=True)
+        ds_file_name = get_filename(model.table_name, model.id)
         return f"datasets/{db_file_name}/{ds_file_name}.yaml"
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


superset/commands/dataset/export.py
```diff
@@ -46,7 +46,7 @@ def _file_name(model: SqlaTable) -> str:
         db_file_name = get_filename(
             model.database.database_name, model.database.id, skip_id=True
         )
-        ds_file_name = get_filename(model.table_name, model.id, skip_id=True)
```
betodealmeida, line 49: Nice!

betodealmeida (approved):

Issue #32793, opened 2025-03-21
[SIP-160] Proposal for Customizable Drill-to-Detail Chart


EMachad0:
*Please make sure you are familiar with the SIP process documented*
[here](https://github.com/apache/superset/issues/5602). The SIP will be numbered by a committer upon acceptance.

### Motivation

Currently, the Drill-to-Detail chart in Superset displays the full dataset when a drilldown is performed. In my experience users have identified three key pain points while using superset:

- **Selective Data Display**: As superset users, we use a single dataset for multiple charts, currently the entire dataset is displayed in the drilldown, we would like the ability to filter and display only the columns that are most relevant to each chart.

- **Custom Labeling**: Custom labels can improve readability and contextual understanding, a feature that already exists in both the table chart and the per dataset but is missing here.

- **CSV Export**: End users often need to export the exact view presented in the drilldown to CSV for further analysis or reporting.

Enhancing the drill-to-detail functionality to mirror the customization available in the table chart will create a more consistent and user-friendly experience.

### Proposed Change

I would be glad for sugestions here.

- Leverage the existing custom labels defined at the dataset level for display in the Drill-to-Detail view.
- Introduce a configuration option within the chart control panel allowing users to specify which dataset columns should appear in the Drill-to-Detail view.
- Incorporate an optional export button within the Drill-to-Detail modal, enabling users to download the currently displayed data as a CSV file.

These configurations could be integrated into a new "Drilldown Options" customize section within the chart control panel.​

(Mockup of the desired Drill-to-Detail interface)
![Image](https://github.com/user-attachments/assets/a2ed01d7-c8b7-4082-a228-8b5abb5e6e07)

### New or Changed Public Interfaces

As far as I know, no change is needed.

### New dependencies

As these enhancements build upon existing functionality, no new dependencies are required.​

### Migration Plan and Compatibility

Maintain backward compatibility with existing charts by ensuring that if the customization options are not configured, the drilldown behavior remains unchanged.

### Rejected Alternatives

An alternative considered was the ability to embed any chart within the Drill-to-Detail view.
However, this approach was deemed overly complex due to the unnecessary complexity related to embeeding charts.

rusackas:
Sounds interesting! Might be useful to see a mockup of the control panel and/or dataset editor mods that would be relevant.

CC @kasiazjc since I know she's had some thoughts about this as well.

mistercrunch:
[...]

PR #23454, opened 2023-03-22
feat: Implement context menu for drill by

kgabryje:
### SUMMARY
Charts that support drill by: all Echarts, World Map, Table, Pivot Table.

This PR implements right-click context menu with drill by option for charts that support the feature. When user right-click on series, the chart plugin constructs a filter object (similar to drill to detail filter), which in the result chart will be appended to original chart's adhoc filters. Then, user can select a column from drill by submenu that later will be used as a dimension of the result chart.
The columns are fetched when user opens the context menu. The result is cached, so that the request is not repeated for charts that use the same dataset. Only columns that are marked as dimensions in the dataset and are not used as dimensions of current chart can be selected.
If there are more than 10 available columns, a search input in the submenu is displayed.

### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

PR #34785, opened 2025-08-21
feat: allow drill-to-details to a specific chart (optionaly) configured at the dataset level

mistercrunch:
This PR introduces customizable drill-to-details functionality that allows dataset owners to configure which chart should be used when users drill through from any chart using their dataset.

Screenshot 2025-08-20 at 11 41 36 PM

Screenshot 2025-08-20 at 11 36 27 PM

Problem: Current drill-to-details uses a one-size-fits-all approach showing all columns in a generic table with no customization options.

Solution: Dataset-level drill-to-details customization allowing association of any existing chart as the drill-to-details target.

Key Benefits:
- Take advantage of all the features of the Table viz in drill-to-details (column selection, ordering, sorting, aggregation, conditional formatting, link/url support, ...)
- Flexible chart types: Any chart type or visualization supported
- Backwards compatible: No-op unless configured, falls back gracefully
- Cross-dashboard consistency: Same experience across all dashboards

Implementation:
- Backend: Added drill_through_chart_id column with FK relationship
- Frontend: ChartSelect component with Dataset Editor integration
- Enhanced drill modal to render StatefulChart when configured
- Full TypeScript support with proper type definitions

Walkthrough:
- create a chart you'd like to see when drilling-to-details for a given dataset
- edit the dataset properties, settings, pick that chart as your drill-to-details chart
- open a dashboard with that dataset
- drill-to-details

### DIFF
--- a/superset-frontend/package-lock.json
+++ b/superset-frontend/package-lock.json
@@ -6781,6 +6781,16 @@
         "node": "^18.14.0 || ^20.0.0 || ^22.0.0 || >=24.0.0"
       }
     },
+    "node_modules/@jest/source-map/node_modules/callsites": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/callsites/-/callsites-3.1.0.tgz",
+      "integrity": "sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=6"
+      }
+    },
     "node_modules/@jest/test-result": {
       "version": "30.0.2",
       "resolved": "https://registry.npmjs.org/@jest/test-result/-/test-result-30.0.2.tgz",
@@ -20793,15 +20803,6 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
-      }
-    },
-    "node_modules/callsites": {
-      "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/callsites/-/callsites-3.1.0.tgz",
-      "integrity": "sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==",
-      "license": "MIT",
-      "engines": {
-        "node": ">=6"
       }
     },
     "node_modules/camel-case": {
@@ -44941,6 +44942,15 @@
       "dependencies": {
         "callsites": "^3.0.0"
       },
+      "engines": {
+        "node": ">=6"
+      }
+    },
+    "node_modules/parent-module/node_modules/callsites": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/callsites/-/callsites-3.1.0.tgz",
+      "integrity": "sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==",
+      "license": "MIT",
       "engines": {
         "node": ">=6"
       }
--- a/superset-frontend/package-lock.json
+++ b/superset-frontend/package-lock.json
@@ -6781,16 +6781,6 @@
         "node": "^18.14.0 || ^20.0.0 || ^22.0.0 || >=24.0.0"
       }
     },
-    "node_modules/@jest/source-map/node_modules/callsites": {
-      "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/callsites/-/callsites-3.1.0.tgz",
-      "integrity": "sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">=6"
-      }
-    },
     "node_modules/@jest/test-result": {
       "version": "30.0.2",
       "resolved": "https://registry.npmjs.org/@jest/test-result/-/test-result-30.0.2.tgz",
@@ -20803,6 +20793,15 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
+      }
+    },
+    "node_modules/callsites": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/callsites/-/callsites-3.1.0.tgz",
+      "integrity": "sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=6"
       }
     },
     "node_modules/camel-case": {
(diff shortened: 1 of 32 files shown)


mistercrunch:
While testing, you might hit a weird known issues related to this comment: https://github.com/apache/superset/pull/23454#issuecomment-3212866369

Refreshing the page fixes it, so you may need to refresh the page between saving the dataset with the new drill-to-chart for it to work in the dashboard.

mistercrunch:
🎪 @unknown Creating ephemeral environment for commit `8148a7c`

**Action:** [View workflow](https://github.com/repo/actions/runs/run)
**Environment:** `8148a7c`
**Powered by:** [Superset Showtime](https://github.com/mistercrunch/superset-showtime)

*Building and deploying... Watch the labels for progress updates.*

EMachad0:
Hey, I really liked your approach and I believe it addresses all the needs of SIP-160!

superset-frontend/package-lock.json
msyavuz, line 1: Is this change intentional?

mistercrunch: wasn't, now I'm wondering what's not deterministic in our package.json causing these side effects ...

superset/datasets/api.py
EMachad0, line 1: I just realized the drill trough chart id is on the dataset level

I understand this does not addresses one problem described on SIP-160,
> As a superset user, we use a single dataset for multiple charts, currently the entire dataset is displayed in the drilldown, we would like the ability to filter and display only the columns that are most relevant to each chart.

I really like the idea of being able to use any chart on the drilldown, do you see putting the drill through chart id on chart level as an option?

mistercrunch: Yeah thought about this and I'm thinking we need another similar feature for:
- drill to chart
- drill to dashboard

Now the great thing is that it's now fairly easy to do drill-to-chart, largely similar feature building on this PR. For drill-to-dashboard, it's just a matter of opening another dashboard with the right filters applied, also fairly easy I think.

The challenge, and where we'd need product/design input @yousoph @kasiazjc , is how does one configure this (?)

Seems it's the dashboard's owner responsibility do define how they want to enable users to drill. So likely a "Dashboard Properties" thing. Guessing there could be a "Drilling" section in dashboard props where the dashboad owner would pick things like:
- Drill to charts:
  - Select a specific set of charts
  - Any chart present on the dashboard (does this make sense, why not cross-filtering instead? maybe can maintain the state of the dashboard while seeing more details...)
  - OR any chart on the associated dataset (?) beyond the dashboard
- Drill to dashboard
  - select a set of related dashboard, all filters from the current state would try to carry through, assuming dashboards share filter configurations (?)

That's without touching the complexity of:
- which specific chart or specific dimension would allow drilling to which set of charts? Say you might want to drill into a specific customer to the "customer scorecard" dashboard, like only customer_name or customer_id would be ones that make sense to target for that specific use case...
- charts that share dataset are naturally "drillable", charts on different dataset that share some columns are partially drillable, ...

Seems there's quite a bit of complexity here. Could go with simpler hard rules like "we'll allow you to drill only on charts that share the same dataset", OR we'll do our best effort to pass down filters that align across dataset.

Seems like "configurability" here is slippery, there are lots of things people might want to open and/or restrict...

rusackas:
Any hope of getting this one through? It's a pretty awesome feature :D

Gunnnn:
What can i do to make this PR happen? Please help me to begin, i need it dreadfully for my work!

syafiqmuda:
how to enable this function? I got issue with this drill-to-detail table where I can sort the table or resize the row and column width

https://github.com/apache/superset/issues/32793#issuecomment-4242171581

rusackas:
@mistercrunch there's real demand for this one (a few folks pinging the thread) but it's conflicting and cypress is red. I'll rebase it and see how it does...

rusackas:
Rebase is still needed here (mergeable shows conflicting) and CI's red across the board (cypress, playwright, lint-frontend). Also worth resolving before another look: bito's flag that `getExploreUrl` needs `formData.datasource`, the duplicate `drill_through_chart_id` in `datasets/schemas.py`'s select columns, and the GHAS DOM-text-as-HTML finding on `DrillDetailModal.tsx`.

Pushing some commits here to see if I can nudge it forward..

rusackas:
Conflicts are back after last month's rebase, and CI never got past queued on the last push here — cypress, playwright, python-unit, sonarqube, codecov all stuck, only docs-preview and labeler ran. Anyone seen that on other branches, or is this one specifically stuck?

rusackas:
LMK if you want help getting this one patched up and reviewed.

superset-frontend/src/components/Chart/DrillDetail/DrillDetailModal.tsx
```diff
@@ -109,14 +120,65 @@ export default function DrillDetailModal({
...
-  }, [exploreUrl, history]);
+  const drillThroughFormData = useMemo(() => {
+    if (!dataset?.drill_through_chart_id || !dataset?.id) {
+      return null;
+    }
+
+    const drillThroughBaseFormData = {
+      slice_id: dataset.drill_through_chart_id,
+      datasource: `${dataset.id}__table`,
+      viz_type: 'table',
```
sadpandajoe, line 136: This overrides the saved chart's `viz_type`, so configuring any non-table chart for drill-through still renders a table. Could the override contain only the drill context and leave the selected chart's visualization type intact?

superset/datasets/schemas.py
```diff
@@ -322,6 +323,7 @@ def fix_extra(self, data: dict[str, Any], **kwargs: Any) -> dict[str, Any]:
     fetch_values_predicate = fields.String(allow_none=True)
     extra = fields.Dict(allow_none=True)
     uuid = fields.UUID(required=True)
+    drill_through_chart_id = fields.Integer(allow_none=True)
```
sadpandajoe, line 326: This exports an environment-local `slices.id` as dataset configuration. Importing a dataset into another instance can either fail the foreign key before that chart exists or bind the dataset to an unrelated chart with the same ID. Could this use a portable chart reference and resolve it after chart import instead?

superset-frontend/src/components/Chart/DrillDetail/DrillDetailModal.tsx
```diff
@@ -33,37 +33,46 @@ import { isEmbedded } from 'src/dashboard/util/isEmbedded';
...
+  showEditButton,
+  onExploreClick,
+  isGeneratingUrl,
 }: ModalFooterProps) => {
   const theme = useTheme();
 
   return (
     <>
-      {!isEmbedded() && (
+      {!isEmbedded() && showEditButton && (
```
sadpandajoe, line 62: This hides the existing Explore/Edit action for every dataset without the new configuration, so the claimed no-op fallback removes the only shortcut back to the source chart after upgrade. Should the unconfigured path keep the previous source-chart action?

superset-frontend/src/components/Select/ChartSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+      },
+      {
+        col: 'datasource_type',
+        opr: 'eq',
+        value: 'table',
+      },
+    ];
+
+    return {
+      q: rison.encode({
```
sadpandajoe, line 74: This request gets only Flask-AppBuilder’s default 20-row page, and `SelectAsyncControl` neither paginates nor sends a server-side search term. A valid chart that sorts after the first page can never be selected; can this use the paginated async selector instead?

superset/datasets/schemas.py
```diff
@@ -174,6 +174,7 @@ class DatasetPutSchema(Schema):
     columns = fields.List(fields.Nested(DatasetColumnsPutSchema))
     metrics = fields.List(fields.Nested(DatasetMetricsPutSchema))
     folders = fields.List(fields.Nested(FolderSchema), required=False)
+    drill_through_chart_id = fields.Integer(allow_none=True)
```
sadpandajoe, line 177: The runtime dataset API now accepts and returns this field, but the checked-in OpenAPI document has no `drill_through_chart_id` property. Generated clients therefore cannot configure the feature and may reject the response; could the API specification be regenerated with this field?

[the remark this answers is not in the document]
sadpandajoe: Agreed—the UI restricts this relationship to charts from the dataset, but the PUT contract accepts any slice ID; the renderer then keeps that chart’s controls while replacing its datasource, so incompatible columns or metrics fail the drill query. Should the update command verify that the target chart uses this dataset?

superset-frontend/src/dashboard/hooks/useDashboardFormData.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    // Find which filters apply to this specific chart
+    const filterIdsAppliedOnChart = getFilterIdsAppliedOnChart(
+      activeFilters,
+      chartId,
+    );
+
+    // If no filters apply, return just the base context
+    if (filterIdsAppliedOnChart.length === 0) {
+      return baseContext;
```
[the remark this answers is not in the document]
sadpandajoe: Agreed—the target chart is normally off-dashboard, so this lookup finds no native-filter scope and the drill result can ignore filters applied to the source chart. Should the context be derived from the source chart ID instead?

PR #34789, opened 2025-08-21
Fea jiazhang/time filter column binding

shirley26107plus:
### 📝 Description
Add the ability to bind time range filters to specific time columns, allowing users to filter charts based on custom time columns instead of default ones.
image

Multiple time range filters can be applied to specific columns of one dataset
image
### ✨ Changes Made
- **Time Filter Configuration**: Add time column binding option in filter configuration UI
- **Column Filtering**: Time filters now only show temporal columns in dropdown
- **Data Flow**: Ensure time filter data is correctly passed to chart queries

### �� Technical Details
- Modified `TimeFilterPlugin.tsx` to support column binding
- Updated `FiltersConfigForm.tsx` to show dataset/column options for time filters
- Enhanced filter data merging in `getFormDataWithDashboardContext.ts`
- Fixed default time filter injection logic in `mixins.tsx`

### 🧪 Testing
- ✅ Time filter column binding works correctly
- ✅ Multiple time filters with different columns work properly
- ✅ Value filters remain unaffected
- ✅ Dashboard to chart filter passing works
- ✅ UI styling is consistent

### 📋 Checklist
- [x] Code follows project style guidelines
- [x] Tests pass
- [x] Documentation updated
- [x] No console errors


msyavuz:
There is something wrong with the pr branch, i don't think you are intending to change 320 files. Marking this as draft as it seems it needs more work

michael-s-molina:
Giving the option to choose a specific column is valid but you shouldn't break the existing functionality of using the default time column. One way to implement this would be to have an All option in the Dataset select and only show the Column select if the Dataset is not All. cc @kasiazjc

sadpandajoe:
@shirley26107plus did you open this by mistake against apache superset? Seeing multiple conflicts as well as 320 file changes on 100+ commits. If this isn't opened by accident, feel free to rebase against master so others can review.

PR #34801, opened 2025-08-21
fix: make `get_image()` always return `BytesIO`

betodealmeida:
### SUMMARY

Currently the method `ScreenshotCachePayload.get_image()` returns `BytesIO | None`, but we pass the result to a file wrapper that doesn't take `None`:

```python
return Response(
    FileWrapper(cache_payload.get_image()),  # <= raises error if get_image() is None
    mimetype="image/png",
    direct_passthrough=True,
)
```

This is causing 5xx errors that should be 4xx. This PR fixes it by (1) making sure the method either returns bytes or raises an exception, and (2) catching the exception and returning a 404.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


mistercrunch (approved):
LGTM

PR #34802, opened 2025-08-21
fix: User-provided Jinja template parameters causing SQL parsing errors

michael-s-molina:
### SUMMARY
Fixes a bug where user-provided Jinja template parameters were not being properly threaded through the SQL processing pipeline, causing parsing errors and security validation failures.

🐛 Problem

When users provided custom Jinja template parameters (e.g., {"table_name": "my_table"}), these parameters were not being passed to the Jinja template processor during SQL parsing and security validation. This caused:
- SQL parsing errors when templates couldn't resolve variables
- Security validation failures when table extraction failed
-Inconsistent behavior between SQL execution and validation phases

🔧 Solution

- Enhanced parameter threading: Added `template_params` parameter throughout the call stack from SQL execution to security validation
- API improvements: Refactored `extract_tables_from_jinja_sql` → `process_jinja_sql` with structured return type for better maintainability

### TESTING INSTRUCTIONS

1. Open SQL Lab
2. Create a query with Jinja template variables:
```
SELECT * FROM {{ table_name }} 
WHERE status = '{{ status_filter }}'
```

3. Add template parameters in the "Template Parameters" field:
```
{
  "table_name": "your_table_name",
  "status_filter": "active"
}
```
4. Expected: Query should execute successfully without parsing errors
5. Before fix: Would fail with parsing errors

cc @jamra 

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/postgres.py
+++ b/superset/db_engine_specs/postgres.py
@@ -21,7 +21,7 @@ import logging
 import re
 from datetime import datetime
 from re import Pattern
-from typing import Any, TYPE_CHECKING
+from typing import Any, Optional, TYPE_CHECKING
 
 from flask_babel import gettext as __
 from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION, ENUM, JSON
@@ -35,7 +35,7 @@ from superset.db_engine_specs.base import BaseEngineSpec, BasicParametersMixin
 from superset.errors import ErrorLevel, SupersetError, SupersetErrorType
 from superset.exceptions import SupersetException, SupersetSecurityException
 from superset.models.sql_lab import Query
-from superset.sql.parse import SQLScript
+from superset.sql.parse import process_jinja_sql
 from superset.utils import core as utils, json
 from superset.utils.core import GenericDataType, QuerySource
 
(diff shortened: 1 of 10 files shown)


justinpark (approved):
LGTM

superset/db_engine_specs/postgres.py
```diff
@@ -301,7 +302,7 @@ def get_default_schema_for_query(
                 )
             )
 
-        return super().get_default_schema_for_query(database, query)
+        return super().get_default_schema_for_query(database, query, template_params)
```
jamra, line 305: Why do we need to propagate template_params if we already do the jinja processing here? Is it for the table extraction method?

superset/models/sql_lab.py
```diff
@@ -80,10 +84,10 @@ class SqlTablesMixin:  # pylint: disable=too-few-public-methods
     def sql_tables(self) -> list[Table]:
         try:
             return list(
-                extract_tables_from_jinja_sql(
+                process_jinja_sql(
```
jamra, line 87: <3 this approach. Process the Jinja then extract tables. Doesn't this also need the template params though?

michael-s-molina: It would but this function looks to be only invoked from dead code that should be removed.

superset/db_engine_specs/postgres.py
michael-s-molina: The super API defines a parameter for `template_params` which is not used currently in the base implementation but could be used in the future.

PR #34803, opened 2025-08-21
fix(error-handling): jinja2 error handling improvements

gabotorresruiz:
### SUMMARY

Adds backward compatibility layer for `Jinja2` template error handling in virtual dataset queries. Maintains existing HTTP status codes (400) while ensuring template syntax errors are properly caught and converted to QueryObjectValidationError for consistent API behavior.

#### CHANGES:
- Added exception handling in `models/helpers.py` to catch both `TemplateError` and `SupersetSyntaxErrorException`
- Created template validation utilities for future use
- Maintained full backward compatibility with existing tests
- No HTTP status code changes (stays 400 for template errors)
- Added comprehensive unit tests for template validation

### TESTING INSTRUCTIONS

1. Test that virtual datasets with invalid `Jinja2` templates return consistent `400` errors
2. Verify error messages are user-friendly rather than raw `Jinja2` exceptions
3. Confirm valid `Jinja2` templates continue to work as expected

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset/utils/jinja_template_validator.py
```diff
@@ -0,0 +1,164 @@
...
+
+def validate_jinja_template(template_str: str) -> Tuple[bool, Optional[str]]:
+    """
+    Validates Jinja2 template syntax.
+
+    Returns:
+        Tuple of (is_valid, error_message)
+        If valid: (True, None)
+        If invalid: (False, "Error description")
+    """
```
betodealmeida: A more idiomatic pattern here would be to raise a custom exception with the error message, instead of returning a tuple (which reminds me of Go and Rust):

```python
class JinjaValidationError(Exception):
    def __init__(self, message: str):
        self.message = message


def validate_jinja_template(template_str: str) -> None:
    # if something goes bad
    raise JinjaValidationError(message)
```

Then you simply call:

```python
try:
    validate_jinja_template(my_template)
except JinjaValidationError as ex:
    message = ex.message
    # do something with message
```

betodealmeida (commented):
Can you give more context on «The "such as" mistake'»? I'm a bit worried that we're trying to fix invalid Jinja2 for the user, what would be the use case here?

superset/utils/jinja_template_validator.py
gabotorresruiz: You're absolutely right to be concerned @betodealmeida! I've removed the `sanitize_jinja_template()` function entirely (I was using it just to test a few things, and I forgot to remove it before)

Let me give you some more context on «The "such as" mistake». This refers to users mistakenly writing:
```
SELECT * WHERE date = {{ start_date such as '2023-01-01' }}
```
Instead of correct Jinja2: 
```
{{ start_date | default('2023-01-01') }}
```

Also, thanks for this suggestion, it totally makes sense!

betodealmeida: Oh, is the "such as" syntax common? I've never seen it before.

superset/connectors/sqla/models.py
```diff
@@ -686,11 +687,12 @@ def get_sqla_row_level_filters(
             grouped_filters = [or_(*clauses) for clauses in filter_groups.values()]
             all_filters.extend(grouped_filters)
             return all_filters
-        except TemplateError as ex:
+        except (TemplateError, SupersetSyntaxErrorException) as ex:
+            msg = ex.message if hasattr(ex, "message") else str(ex)
```
betodealmeida: Small nit:

```suggestion
            msg = getattr(ex, "message", str(ex))
```

superset/datasets/api.py
```diff
@@ -1319,9 +1322,12 @@ def render_item_list(item_list: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
                 data[new_key] = func(data[key])
-            except TemplateSyntaxError as ex:
-                raise SupersetTemplateException(
+            except (TemplateSyntaxError, SupersetSyntaxErrorException) as ex:
+                template_exception = SupersetTemplateException(
                     f"Unable to render expression from dataset {item_type}.",
-                ) from ex
+                )
+
+                template_exception.status = 400
```
betodealmeida: I wonder if the status for `SupersetTemplateException` should be changed to 400, since the template is always provided by the user?

betodealmeida (approved):

superset/datasets/api.py
gabotorresruiz: Hey @betodealmeida not sure if you meant `422`. Right now is a `400` error

superset/connectors/sqla/models.py
gabotorresruiz: Done! Thanks

superset/datasets/api.py
gabotorresruiz: Oh @betodealmeida! Now I see what you mean. Right now `SupersetTemplateException` raises a 500 error. Do you want me to make the change so `SupersetTemplateException` raises an error status 400 instead of 500? Let me know

betodealmeida: Yeah, let's change this to a 422, since it's always caused by user input. I asked Claude to take a look as well, here's what it said:

<hr/>

  1. The exception is raised for user-provided template issues:
    - Invalid Jinja templates in SQL queries
    - Unsafe template values/functions
    - Missing or incorrect template parameters
    - Recursion errors in user templates
  2. There's already a related exception with 400 status: SupersetTemplateParamsErrorException (line 150-165 in exceptions.py) has status = 400, showing the pattern for template-related user errors.
  3. Current usage shows it's treated as a client error: In datasets/api.py:1177-1178, when SupersetTemplateException is caught, it returns a 400 response (self.response_400()), not a 500.
  4. All the error scenarios are user mistakes:
    - Using unsafe functions in templates
    - Infinite recursion in templates
    - Invalid partition specifications
    - Missing metrics in templates

The fix would be to add status = 400 or status = 422 to SupersetTemplateException. Given that SupersetTemplateParamsErrorException uses 400 and the API already returns 400 when catching this
  exception, 400 would be the most consistent choice.

<hr/>

gabotorresruiz: Sounds good! I'll do this change ASAP

PR #34805, opened 2025-08-22
docs: Add local configuration guidance for LLMs

eschutho:
## Summary
- Add Local Configuration section to LLM documentation 
- Provide clear guidance for Docker vs Flask deployments
- Specify git-ignored config files for each deployment method
- Include restart instructions appropriate for each environment

## Test plan
- [x] Verify Docker config approach works with feature flags
- [x] Validate git-ignore patterns work correctly  
- [x] Test config file auto-import behavior

### DIFF
--- a/LLMS.md
+++ b/LLMS.md
@@ -187,6 +187,26 @@ pre-commit run eslint            # Frontend linting
 - **[GPT.md](GPT.md)** - For OpenAI/ChatGPT tools
 - **[.cursor/rules/dev-standard.mdc](.cursor/rules/dev-standard.mdc)** - For Cursor editor
 
+## Local Configuration
+
+### Feature Flags & Settings
+Always use git-ignored config files instead of modifying core files:
+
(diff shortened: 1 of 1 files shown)


mistercrunch:
would love to wrap this in tooling so we can tell the LLM and users something like: "to review or alter Superset configurations, use `superset config --help`". Would lead to more **context density** and a bulletproof way to do this properly.

`superset config` ideally would be able to expose "effective configuration" by exposing what's get assigned to `flask.current_app.conf`, and guide the user/llm to alter the right entrypoint based on their current setup. Maybe `superset config --help` would provide docs and/or would be like "do you want me to alter your `docker/.env-local` for you?". In theory could ask:

```
$ superset config set
How is your environment currently setup?
1. docker with docker-compose
2. Makefile
3. custom
4. ...
``

sadpandajoe (approved):

PR #34806, opened 2025-08-22
fix(query_context_processor): apply CSV encoding to follow the encoding settings of 'CSV_EXPORT'

haje01:
### SUMMARY
The encoding issue when exporting CSV from sql_lab was resolved by https://github.com/apache/superset/pull/34235, but the encoding issue when exporting CSV from dashboard graphs still remains. This PR fixes that issue.

### TESTING INSTRUCTIONS
1. Click the three-dot icon on any dashboard graph and select Download / Export to .CSV
2. Open the downloaded file in vim's hex mode to check if BOM (Byte Order Mark) is attached

### DIFF
--- a/superset/charts/data/api.py
+++ b/superset/charts/data/api.py
@@ -524,8 +524,11 @@ class ChartDataRestApi(ChartRestApi):
             # return multi-query results bundled as a zip file
             def _process_data(query_data: Any) -> Any:
                 if result_format == ChartDataResultFormat.CSV:
-                    encoding = app.config["CSV_EXPORT"].get("encoding", "utf-8")
-                    return query_data.encode(encoding)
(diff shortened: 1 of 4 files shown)


rusackas:
This PR hasn't been updated in a long time and has several CI failures (pre-commit, test-mysql, test-postgres, test-sqlite). The test failures are likely caused by the `.encode()` call changing the result type from `str` to `bytes`, which may break downstream callers. We'll mark it as draft for now — please investigate the test failures, fix the type issue, rebase against `master`, and set it back to 'ready for review' when it's ready.

rusackas (approved):
Made some touchups myself to turn this green, but I'm OK with this getting across the finish line. Please review/block if you feel otherwise

Issue #8574, opened 2019-11-14
[SIP-26] Proposal for Implementing Connection Pooling for Analytics Database Connections


willbarrett:
### Motivation

Currently, Superset’s connections to analytics databases do not have long-lived connection pools. In most instances, a database connection is spawned immediately before a query is executed and discarded after a single use. This introduces a small amount of latency into every query. While most queries run against data warehouses are expected to be longer-running than a typical web application query, this latency will be noticeable when performing operations such as loading schema and table lists for display in the UI, or loading table definitions and previews. 
A more serious concern is that the number of open database connections to analytics databases is only bounded by the number of threads available to the application across all processes. Under peak load, this can lead to hammering databases with a large number of connection requests and queries simultaneously. This does not allow us to provide meaningful upper bounds for the number of available database connections. Implementing connection pooling at the process level will allow us to provide a configurable maximum number of connections that Superset is able to leverage.

### Proposed Change

I recommend we add a singleton object to hold a SQLAlchemy Engine instance for each configured database in the application. I believe that engines should not be instantiated on startup, but instead instantiated on first use to avoid unnecessary connection negotiation.

I further recommend that we use the [SQLAlchemy QueuePool](https://docs.sqlalchemy.org/en/13/core/pooling.html#sqlalchemy.pool.QueuePool) as the default pool implementation while retaining the ability to configure Superset to use a NullPool, configurable via the Database setup system. I would like to make the `pool_size` and `max_overflow` properties configurable, as well as whether to treat the queue as FIFO or LIFO and the `pool_pre_ping` option, and customization of the `connect_args` passed on engine instantiation (which controls things like connection timeouts). I believe that LIFO queues will be preferable for infrequently-accessed database connections, as they will generally maintain a lower number of connections in the pool, and thus should be the default. I would also recommend that for LIFO queues we default to the `pool_pre_ping` option to trigger pool member invalidation when necessary, as stale connections are more likely under the LIFO configuration.

As part of this work, I recommend moving engine instantiation code out of the Database model and into its own module, probably as part of the singleton that will maintain an in-memory list of database pools. We will need to update the code that alters database records to reinitialize the processes’ engine after Database record creation and update.

[...]

PR #34826, opened 2025-08-22
feat: a manager for SQLAlchemy engines

betodealmeida:
### SUMMARY

This PR centralizes engine creation in a single `EngineManager`, consolidating logic previously scattered in `Database.get_sqla_engine()`

**Code Changes:**
- New `superset/engines/manager.py` with `EngineManager` class (~680 lines)
- New `superset/extensions/engine_manager.py` Flask extension
- Simplified `Database.get_sqla_engine()` to delegate to `EngineManager`
- Removed `superset/extensions/ssh.py` (functionality merged into EngineManager)
- Added type aliases for `DBConnectionMutator` and `EngineContextManager`
- Comprehensive unit tests

Note that this introduces breaking changes -- see `UPDATING.md`. It should only be merged during the breaking window for 7.0.

N/A - Backend infrastructure change with no UI impact.

### TESTING INSTRUCTIONS

Added unit tests. Nothing should change for the user or admin.

### ADDITIONAL INFORMATION

- [X] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI
- [x] Removes existing feature or API (removes `SSHManager` in favor of integrated `EngineManager`)


superset/models/core.py
```diff
@@ -421,130 +416,31 @@ def get_effective_user(self, object_url: URL) -> str | None:
         )
 
     @contextmanager
-    def get_sqla_engine(  # pylint: disable=too-many-arguments
+    def get_sqla_engine(
         self,
         catalog: str | None = None,
         schema: str | None = None,
-        nullpool: bool = True,
```
aminghadersohi, line 431: dont need this?

superset/extensions/engine_manager.py
```diff
@@ -0,0 +1,100 @@
...
+        self.engine_manager: EngineManager | None = None
+
+    def init_app(self, app: Flask) -> None:
+        """
+        Initialize the EngineManager with Flask app configuration.
+        """
+        engine_context_manager = app.config["ENGINE_CONTEXT_MANAGER"]
+        db_connection_mutator = app.config["DB_CONNECTION_MUTATOR"]
+        mode = app.config["ENGINE_MANAGER_MODE"]
+        cleanup_interval = app.config["ENGINE_MANAGER_CLEANUP_INTERVAL"]
```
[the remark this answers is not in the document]
betodealmeida: I updated the comments in `superset/config.py`, we want to use a `timedelta`.

superset/models/core.py
betodealmeida: No, that's the default now, and if the admin wants something different they need to specify pool options.

villebro:
@betodealmeida I believe the original SIP has been abandoned, and before it was closed I left a comment about the design not considering typical distributed worker deployments, for which the singleton doesn't really work: https://github.com/apache/superset/issues/8574#issuecomment-1749752198 With this design connections will be grow linearly with the number of pods, which makes it more difficult to manage multi-pod deployments. So I think we need to revisit the architecture to come up with a design where we can coordinate connection pooling across the whole worker fleet.

michael-s-molina:
> So I think we need to revisit the architecture to come up with a design where we can coordinate connection pooling across the whole worker fleet.

Agree with @villebro. To illustrate the problem in more detail:

                                                                           
                                                                                                                 
####  1. Connection Pool Multiplication                                                                              
                                                                                                                 
  In `SINGLETON` mode, each pod/Celery worker has its own `EngineManager` instance:                                  
                                                                                                                 
  Pod 1: EngineManager → Pool(size=5) → 5 connections                                                            
  Pod 2: EngineManager → Pool(size=5) → 5 connections                                                            
  Pod 3: EngineManager → Pool(size=5) → 5 connections                                                            
  ...                                                                                                            
                                                                                                                 
  With 10 pods and pool size 5, you could hit 50 connections to the database, not 5. This can easily exhaust     
  database connection limits, especially for managed databases with strict limits.                               
                                                                                                                 
  #### 2. SSH Tunnel Multiplication                                                                                   
                                                                                                                 
  Same problem with SSH tunnels:                                                                                 
                                                                                                                 
```
# Each pod creates its own tunnel cache   
self._tunnels: dict[TunnelKey, SSHTunnelForwarder] = {}   
```                                                     
                                                                                                                 
  If you have 20 Celery workers across multiple pods, you could have 20 separate SSH tunnels to the same bastion 
  host. Many SSH servers limit concurrent connections per user (often 10 by default).                            
                                                                                                                 
  #### 3. Celery Prefork Workers Make It Worse                                                                        
                                                                                                                 
  Celery's default prefork model spawns multiple worker processes:                                               
                                                                                                                 
  Pod 1:                                                                                                         
    └── Celery Worker (prefork, concurrency=4)                                                                   
         ├── Process 1 → EngineManager → Pool                                                                    
         ├── Process 2 → EngineManager → Pool                                                                    
         ├── Process 3 → EngineManager → Pool                                                                    
         └── Process 4 → EngineManager → Pool                                                                    
                                                                                                                 
  Each forked process gets its own memory space, so connection pools aren't shared even within a single pod.     
                                                                                                                 
 #### 4. No Cluster-Wide Coordination                                                                                
                                                                                                                 
  The cache keys are deterministic but there's no coordination:                                                  
                                                                                                                 
  ```
# manager.py:258
  engine_key = self._get_engine_key(database, catalog, schema, source, user_id)
```                              
                                                                                                                 
  Each process independently decides to create connections without knowing cluster-wide state.                   
                                                                                                                 
 #### What's Missing for True Connection Pooling                                                                     
                                                                                                                 
  For distributed connection pooling, you'd typically need:                                                      
                                                                                                                 
  1. External connection pooler like PgBouncer, ProxySQL, or cloud-native poolers (RDS Proxy, Cloud SQL Proxy)   
  2. Centralized state via Redis or similar for tunnel coordination                                              
  3. Connection limits per-pod that account for cluster size

File: superset/models/core.py [hunk 3]
```diff
@@ -424,130 +419,31 @@ def get_effective_user(self, object_url: URL) -> str | None:
...
-        if nullpool:
-            engine_kwargs["poolclass"] = NullPool
-        connect_args = engine_kwargs.setdefault("connect_args", {})
-
-        # modify URL/args for a specific catalog/schema
-        sqlalchemy_url, connect_args = self.db_engine_spec.adjust_engine_params(
-            uri=sqlalchemy_url,
-            connect_args=connect_args,
+        # Use the engine manager to get the engine
+        engine_manager = engine_manager_extension.manager
```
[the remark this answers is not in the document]
betodealmeida: `EngineManager.get_engine` does this:

```python
    @contextmanager
    def get_engine(...) -> Iterator[Engine]:
        with self.engine_context_manager(database, catalog, schema):
            with check_for_oauth2(database):
                yield self._get_engine(database, catalog, schema, source)
```

betodealmeida:
@villebro @michael-s-molina I understand this is not a solution for every deployment out there, but it's still incredibly valuable for other architectures (like Preset, but also smaller Superset deployments). For distributed workers you can continue using `EngineModes.NEW`, or potentially implement a third engine mode that manages the pool between pods.

Regardless, IMHO this PR is a big win just for the fact that it centralizes the engine creation in a single place in a Flask extension.

dpgaspar:
> > So I think we need to revisit the architecture to come up with a design where we can coordinate connection pooling across the whole worker fleet.
> 
> Agree with @villebro. To illustrate the problem in more detail:
> 
> #### 1. Connection Pool Multiplication
> In `SINGLETON` mode, each pod/Celery worker has its own `EngineManager` instance:
> 
> Pod 1: EngineManager → Pool(size=5) → 5 connections Pod 2: EngineManager → Pool(size=5) → 5 connections Pod 3: EngineManager → Pool(size=5) → 5 connections ...
> 
> With 10 pods and pool size 5, you could hit 50 connections to the database, not 5. This can easily exhaust database connection limits, especially for managed databases with strict limits.
> 
> #### 2. SSH Tunnel Multiplication
> Same problem with SSH tunnels:
> 
> ```
> # Each pod creates its own tunnel cache   
> self._tunnels: dict[TunnelKey, SSHTunnelForwarder] = {}   
> ```
> 
> If you have 20 Celery workers across multiple pods, you could have 20 separate SSH tunnels to the same bastion host. Many SSH servers limit concurrent connections per user (often 10 by default).
> 
> #### 3. Celery Prefork Workers Make It Worse
> Celery's default prefork model spawns multiple worker processes:
> 
> Pod 1: └── Celery Worker (prefork, concurrency=4) ├── Process 1 → EngineManager → Pool ├── Process 2 → EngineManager → Pool ├── Process 3 → EngineManager → Pool └── Process 4 → EngineManager → Pool
> 
> Each forked process gets its own memory space, so connection pools aren't shared even within a single pod.
> 
> #### 4. No Cluster-Wide Coordination
> The cache keys are deterministic but there's no coordination:
> 
> ```
> # manager.py:258
> engine_key = self._get_engine_key(database, catalog, schema, source, user_id)
> ```
> 
> Each process independently decides to create connections without knowing cluster-wide state.
> 
> #### What's Missing for True Connection Pooling
> For distributed connection pooling, you'd typically need:
> 
> 1. External connection pooler like PgBouncer, ProxySQL, or cloud-native poolers (RDS Proxy, Cloud SQL Proxy)
> 2. Centralized state via Redis or similar for tunnel coordination
> 3. Connection limits per-pod that account for cluster size

I generally agree, these are valid concerns. But, on a generic web app, connection pooling is highly desirable, same logic applies up to a certain extent here.

Connection pooling provides measurable performance and stability benefits as well as reduced connection churn improves database health. A small pool smooths that behavior and stabilizes connection usage. 

Also given Superset's nature, longer connections, pooling avoids the connection storm effect. 
Dashboard refresh, burst of queries
→ burst of connection creations
→ burst of TLS/auth
→ DB CPU spike

This can cause an unpredictable upper bound

Yes, pools multiply across pods and prefork workers. This doesn’t mean pooling is wrong; it means we must size pools intentionally and cap pod HPA scaling. Predicting a worst case scenario.

On Preset's case and probably many others, PgBouncer or ProxySQL are not viable options.

Would be great if we could load test this solution under concurrency at the web layer, that I think we benefit the most. Multiple concurrent calls to `/api/v1/chart/data` with NullPool and with Pool and measuring the db connections

michael-s-molina:
> I generally agree, these are valid concerns. But, on a generic web app, connection pooling is highly desirable, same logic applies up to a certain extent here.

@dpgaspar I completely agree that connection pooling is a valuable and highly desirable feature. The points raised by @villebro and myself are more about how we should introduce this feature into Superset. Currently, Superset operates in at least two modes:

- A basic mode where Celery and Redis are not available
- An advanced mode where Celery and Redis are available, enabling distributed operation

In recent efforts—such as [GTF](https://github.com/apache/superset/pull/36368)—we’ve ensured support for both modes, so organizations of any size or setup can benefit from new features.

To illustrate the alternative: @villebro could have designed GTF to function only in distributed environments, making Celery and Redis mandatory.

Perhaps we need a broader conversation about which modes of operation we officially support, and whether new features should be compatible with all modes. I propose we discuss this further on Friday's Town Hall.

betodealmeida:
@michael-s-molina @villebro I rewrote this PR keeping the non-controversial part: centralizing the engine creation in a single class. I think there's a lot of value there already, since today the creation is scattered in the database model. This keeps the existing behavior, but opens the possibility of more complex engine managers in the future, addressing both use cases that we discussed.

superset-frontend/src/embedded/EmbeddedContextProviders.tsx
```diff
@@ -64,7 +65,7 @@ export const EmbeddedContextProviders: React.FC<{
 }> = ({ children }) => {
   const RootContextProviderExtension = extensionsRegistry.get(
     'root.context.provider',
-  );
+  ) as ComponentType<{ children?: ReactNode }> | undefined;
```
betodealmeida, line 68: I had to touch this and a few other unrelated files to fix lint.

richardfogaca (commented):
_Posting on Richard's behalf - this is his PR reviewer agent. Forward any pushback to him and he'll loop me back in._

Left a few notes below. The feature-flag lookup looks like the main functional regression to address before merge; the other two are compatibility/extension-surface questions. All line numbers verified against HEAD `9c5e6d1`.

### Functional - worth fixing before merge

- **`superset/engines/manager.py:118`**

  This creates a fresh `FeatureFlagManager()` inside the engine path. Since that instance is never initialized with the Flask app, `is_feature_enabled("IMPERSONATE_WITH_EMAIL_PREFIX")` will always fall back to false, so deployments relying on email-prefix impersonation silently stop mapping the Superset username to the database principal they expect.

  WDYT - could this reuse the initialized feature flag manager or existing `is_feature_enabled` helper, and maybe restore the removed email-prefix impersonation test coverage?

### Compatibility questions

- **`superset/extensions/engine_manager.py:47`**

  The extension now always constructs the built-in `EngineManager`, while the previous `SSH_TUNNEL_MANAGER_CLASS` hook is removed from config. That may be intentional as part of centralizing engine creation, but any deployment using a custom SSH manager for bastion routing, audit, host-key policy, or credential handling would lose that behavior during upgrade.

  Would it be worth preserving an equivalent configurable hook, or documenting this as a breaking change with the migration path?

- **`superset/models/core.py:422`**

  `Database.get_sqla_engine()` no longer accepts the `nullpool` keyword. In-repo callers look updated, but this method has been a long-standing extension/test/plugin surface, so external code passing `nullpool=False` will now fail with `TypeError` before it reaches the manager.

  Could we keep the keyword temporarily, even if deprecated, or call this out explicitly as a breaking API change?

### Praise

- **`superset/engines/manager.py:161`**

  Nice improvement moving URI validation after the mutator path as well. That makes the centralized flow a bit easier to reason about because the final URI is checked immediately before engine creation.

betodealmeida:
Addressed all the comments in the latest commit. Let's wait to merge this during a breaking window.

richardfogaca (commented):
_Posting on Richard's behalf — this is his PR reviewer agent. Forward any pushback to him and he'll loop me back in._

Left a few notes below — the main things I’d look at before merge are the engine prequery regression and a couple of compatibility/config-ordering edges from moving engine creation into the new manager. All line numbers verified against HEAD `75a64b3`.

### Functional — worth investigating before merge

- **`superset/engines/manager.py:86`**

  `get_engine()` now yields the newly-created engine directly, but I don’t see the previous `get_prequeries()` connect listener being reattached on the `get_sqla_engine()` path. `get_raw_connection()` still runs prequeries manually, but normal engine callers go through this manager path; for engines like Postgres and Databricks those prequeries set `search_path`, `USE CATALOG`, and `USE SCHEMA`, and the Postgres docstring calls out SQL Lab access checks for unqualified table names.

  WDYT — could we move the old prequery event registration into the manager before yielding the engine, and add a focused regression test for `get_sqla_engine(schema=...)`?

- **`superset/extensions/engine_manager.py:41`**

  `EngineManagerExtension.init_app()` snapshots `ENGINE_CONTEXT_MANAGER` and `DB_CONNECTION_MUTATOR` into the manager at startup. Since `configure_engine_manager()` now runs before `FLASK_APP_MUTATOR` in `superset/initialization/__init__.py:619-627`, deployments that install or replace those hooks from the app mutator look like they’ll silently keep the pre-mutator values; previously `Database._get_sqla_engine()` read the mutator/context manager from `app.config` at engine creation time.

  Could we either initialize the engine manager after `FLASK_APP_MUTATOR`, or keep these hooks late-bound from `current_app.config` inside the manager?

### Compatibility suggestions

- **`superset/models/core.py:424-427`**

  The deprecated `nullpool` argument is still accepted as a keyword, but moving `source` before it changes the old positional contract. A caller using the previous positional form `get_sqla_engine(catalog, schema, False, source)` would now pass `False` as `source` and the old `source` value as `nullpool`, which means the compatibility shim doesn’t cover positional callers.

  Small suggestion: could we keep the old order as `catalog, schema, nullpool=None, source=None` while still warning and ignoring `nullpool`?

- **`superset/models/core.py:481-486`**

  `get_raw_connection()` also used to accept and forward `nullpool`, but the new signature drops it entirely. Existing callers of `get_raw_connection(..., nullpool=False)` will now fail with `TypeError`, while `UPDATING.md` only documents the `get_sqla_engine(nullpool=...)` deprecation path.

  Could we keep a deprecated `nullpool: bool | None = None` keyword here too, even if it’s ignored, so the compatibility story is consistent?

### Praise

- **`superset/engines/manager.py:58`**

  Centralizing the engine lifecycle behind one context manager is a nice direction. Having SSH, OAuth2, impersonation, encrypted extras, and connection mutators in one path makes this much easier to audit than the previous split between `Database` and the SSH extension.

aminghadersohi (commented):
PR #34826 — feat: a manager for SQLAlchemy engines
HEAD SHA: 75a64b3062dabdb18e8beb05df53029e33b9cc8d
Author: betodealmeida
Re-review: new commits pushed 2026-05-11, prior review 2025-12-04

Second opinion: not spawned — user requested immediate post. PR qualifies (>10 files, new Flask extension) but skipping per operator instruction.

---

## Scan coverage

- Scan 1 (inline imports): 5 violations in manager.py, 1 in models/core.py
- Scan 2 (config.get): 0 violations
- Scan 3 (f-string logging): 0 violations
- Scan 4 (broad exception): 1 in manager.py (manager-mapped, see MEDIUM)
- Scan 5 (sensitive info in errors): 0 violations
- Scan 6 (module-level mutable globals): 0 violations
- Scan 7 (print statements): 0 violations
- Scan 8 (db.session.commit): 0 violations
- Scan 9 (JWT decode without verify): 0 violations
- Scan 10 (hardcoded timeouts): 0 violations
- Scan 11 (test methods missing -> None): 0 violations
- Scan 12 (Any type usage): 0 violations
- Scan 13 (module-level app context refs): 0 violations
- Scan 14 (SQL injection): 0 violations
- Scan 15 (private symbol imports): 0 violations
- Scan 16 (or-default collapsing falsy): 0 violations
- Scan 17 (compound boolean without parens): 0 violations
- Scan 18 (migration boolean filter): not applicable
- Scan 19 (async client without timeout): 0 violations
- Scan 20 (asyncio.get_event_loop in async): 0 violations

---

## BLOCKER (must fix before merge)

- [superset/engines/manager.py:67] Inline import without circular-import justification — `from superset.utils.oauth2 import check_for_oauth2`
  Suggestion: If a real cycle exists (likely: `utils.oauth2` → `models` → `extensions` → `engines.manager`), add a `# circular import` comment naming the cycle. Otherwise move to module top. Same pattern applies to lines 101, 102, 125, 179 below.

- [superset/engines/manager.py:101] Inline import without circular-import justification — `from superset import is_feature_enabled`
  Suggestion: `superset/__init__.py` imports `models.core`, and `models.core` loading triggers `extensions/__init__.py` → `engines.manager`, making this a real cycle. Add: `# circular import: superset/__init__ → models.core → extensions → engines.manager`

- [superset/engines/manager.py:102] Inline import without circular-import justification — `from superset.extensions import security_manager`
  Suggestion: `extensions/__init__.py` imports `EngineManagerExtension` from this package, making this circular at module load. Add: `# circular import: extensions/__init__ → engine_manager → engines.manager`

- [superset/engines/manager.py:125] Inline import without circular-import justification — `from superset.utils.oauth2 import get_oauth2_access_token`
  Suggestion: Add `# circular import` comment naming the cycle. Also consider consolidating with the `check_for_oauth2` import at line 67 since both are from `superset.utils.oauth2`.

- [superset/engines/manager.py:179] Inline import without circular-import justification — `from superset.utils.ssh_tunnel import get_default_port`
  Suggestion: If `utils.ssh_tunnel` doesn't transitively depend on `engines.manager`, this can move to module top. Note: moving it also requires updating the test patch target from `superset.utils.ssh_tunnel.get_default_port` to `superset.engines.manager.get_default_port`.

- [superset/models/core.py:441] Inline `import warnings` with no justification — `warnings` is a stdlib module with no circular import risk.
  Suggestion: Move `import warnings` to the module-level import block at the top of `core.py`.

---

## HIGH (should fix)

- [superset/engines/manager.py:55-56] `sshtunnel.TUNNEL_TIMEOUT` and `sshtunnel.SSH_TIMEOUT` are module-level globals on a third-party library. Setting them in `__init__` means the last-constructed `EngineManager` wins process-wide. If two instances are ever created with different timeouts (in tests, or when the class is subclassed and reinstantiated), they silently stomp each other. The old `SSHManager` had the same footgun; this PR carries it forward. Consider passing these values directly to `sshtunnel.open_tunnel` via `ssh_timeout` and `ssh_tunnel_timeout` kwargs instead of mutating module globals.

- [superset/engines/manager.py:172-176] Resource leak if `tunnel.start()` raises inside `_create_tunnel`. The `SSHTunnelForwarder` is allocated by `sshtunnel.open_tunnel()` at line 174 and started at line 175. If `.start()` throws, the forwarder is neither returned to the caller nor cleaned up — the TCP socket/thread may be leaked. The caller's `try/finally` at lines 81-88 only protects after the tunnel is successfully returned, not during construction.
  Suggestion:
  ```python
  def _create_tunnel(self, ssh_tunnel, uri):
      kwargs = self._get_tunnel_kwargs(ssh_tunnel, uri)
      tunnel = sshtunnel.open_tunnel(**kwargs)
      try:
          tunnel.start()
      except Exception:
          tunnel.stop()
          raise
      return tunnel
  ```

- [superset/extensions/engine_manager.py:47-48] No validation that `engine_manager_class` is a subclass of `EngineManager`. A misconfigured `ENGINE_MANAGER_CLASS` (valid dotted path, wrong type) will instantiate without error at `init_app` time and fail with an opaque `AttributeError` at the first `get_engine()` call under user traffic. Per Rule 29, operator config must fail fast at startup.
  Suggestion: Add after the `load_class_from_name` call:
  ```python
  if not issubclass(engine_manager_class, EngineManager):
      raise TypeError(
          f"ENGINE_MANAGER_CLASS must be a subclass of EngineManager, got {engine_manager_class}"
      )
  ```

---

## MEDIUM (nice to fix)

- [superset/engines/manager.py:43-49] Constructor default `tunnel_timeout=timedelta(seconds=30)` doesn't match the config default `SSH_TUNNEL_TIMEOUT_SEC = 10.0` in `config.py`. Production is fine since `EngineManagerExtension.init_app` always passes the config value. But direct instantiation (tests, subclasses, tooling) gets 30s — 3× what operators configure. Align the constructor default with the config default, or remove the default entirely and require the caller to pass it explicitly.

- [superset/engines/manager.py:167-170] `except Exception as ex` in `_create_engine` catches programming errors (e.g., `TypeError` from bad kwargs) alongside DBAPI errors. Both are forwarded to `get_dbapi_mapped_exception`, which may not handle non-DBAPI exceptions gracefully. This pattern exists elsewhere in the codebase, but a comment explaining why broad catch is intentional here would prevent future narrowing that breaks the mapping contract.

- [superset/engines/manager.py:178] UPDATING.md documents an intentional behavioral change: the old `SSHManager.create_tunnel` passed `debug_level=logging.getLogger("flask_appbuilder").level` to `sshtunnel.open_tunnel`; the new `_get_tunnel_kwargs` intentionally omits it. This context lives in UPDATING.md but not in the code. Per Rule 27, design decisions that disappear from view during future maintenance need an inline comment.
  Suggestion: Add above line 178:
  ```python
  # debug_level intentionally omitted vs old SSHManager.create_tunnel.
  # Subclasses that need the old behavior should override _get_tunnel_kwargs.
  ```

- [tests/unit_tests/engines/manager_test.py] Missing test coverage for the `get_engine()` public path and `_create_engine()` exception mapping. The three existing tests exercise `_get_engine_args` (NullPool, impersonation) and `_get_tunnel_kwargs` (missing port), but there are no tests for: (a) the happy path without SSH tunnel, (b) the happy path with SSH tunnel, (c) `_create_engine` DBAPI exception mapping, or (d) the `engine_context_manager` and `check_for_oauth2` context managers being entered. The deleted test in `superset/engines/__init__.py` verified the sshtunnel global state set in `__init__`; there is no equivalent for the new `EngineManager.__init__` behavior.

---

## NIT (non-blocking)

- [tests/unit_tests/engines/manager_test.py:32-37] The nested `dummy_context_manager` function is missing a return type annotation. Add `-> Iterator[None]` (import `Iterator` from `collections.abc`).

- [superset/engines/manager.py:101-102] The two inline imports `is_feature_enabled` and `security_manager` could be grouped as a single comment block at the top of `_get_engine_args` rather than scattered mid-function. Minor readability.

---

## PRAISE

- `EngineManagerExtension` follows the Flask `init_app()` pattern correctly: no module-level state until `init_app` is called, `RuntimeError` guard on the `manager` property before init. This is the right pattern.
- `_get_tunnel_kwargs` separated from `_create_tunnel` specifically for testability — `test_get_tunnel_kwargs_requires_database_port` demonstrates this pays off immediately.
- `models/core.py:450` has the `# Import here to avoid circular imports` comment justifying the deferred import of `engine_manager_extension`. Exactly what the other inline imports in `manager.py` are missing.
- Lazy `%s` log formatting throughout `engine_manager.py` lines 58-62. No f-strings in logging calls.
- Comprehensive UPDATING.md section with a migration table mapping old `SSHManager` methods to new `EngineManager` override points. Genuinely useful for operators who subclassed `SSHManager`.
- `DBConnectionMutator` and `EngineContextManager` type aliases in `superset_typing.py` are well-typed and improve the interface over the untyped callables they replace.

aminghadersohi (commented):
## PR #34826 — `feat: a manager for SQLAlchemy engines`
**HEAD SHA:** `75a64b3062dabdb18e8beb05df53029e33b9cc8d`

**Mode:** code review (21 files, 563 additions / 437 deletions — complex PR threshold triggered)

---

### PRAISE

- Clean, well-scoped extraction. `EngineManager` has a clear interface and the Flask extension pattern (`EngineManagerExtension`) matches every other Superset extension.
- The UPDATING.md migration guide is thorough — the method-to-method mapping table for `SSHManager` subclassers is genuinely useful.
- `NullPool` is now always enforced regardless of `engine_params`, removing the old footgun where passing `poolclass` in extras could accidentally enable connection pooling.
- Proper `try/finally` in `get_engine()` guarantees the SSH tunnel is always stopped even if engine creation or usage raises.
- Good use of `TYPE_CHECKING` guard for `SSHTunnel` and `Database` — avoids circular imports at module load time.
- `DBConnectionMutator` and `EngineContextManager` type aliases in `superset_typing.py` codify previously implicit callable contracts.
- OAuth2 access token is now only fetched inside `if database.impersonate_user:`, avoiding an unnecessary `get_oauth2_config()` call on every engine creation when impersonation is off.
- `get_user_id()` replaces the old `hasattr(g, "user") and hasattr(g.user, "id")` guard — cleaner and centralised.
- `"@" in user.email` guard (manager.py:121) before splitting — minor defensive improvement over the old code.

---

### Scan coverage (all 19 run)

1. Inline imports without circular-import comment: **FOUND** (HIGH-3)
2. SQL injection vectors: 0 matches
3. Hardcoded credentials/secrets: 0 matches
4. Bare/broad `except`: 1 match (manager.py:169) — immediately re-raises mapped exception; acceptable
5. Unvalidated external input: 0 new API routes
6. `Any` types in new code: 1 match — `DBConnectionMutator` 4th param (NIT-5)
7. `# type: ignore` without explanation: 0 matches
8. TODO/FIXME/HACK comments: 0 matches
9. Missing `stacklevel` in `warnings.warn`: `stacklevel=2` present — OK
10. Mutable default arguments: 0 matches
11. Missing auth checks on new routes: no new routes
12. Logging of sensitive data: 0 matches
13. `eval()` / `exec()` usage: 0 matches
14. String formatting in logging: all `%s` style — OK
15. Unused imports: 0 matches
16. Missing type annotations: all new methods have return types — OK
17. `os.system()` / shell injection: 0 matches
18. Broad exception swallowing (catch-without-reraise): 0 matches
19. Missing tests for new code paths: **FOUND** (HIGH-1)

---

### HIGH

**HIGH-1 — Sparse test coverage for a critical security component**

`tests/unit_tests/engines/manager_test.py` has 3 tests covering a ~200-line class that owns SSH credential handling, OAuth2 token passing, user impersonation, and engine creation. Missing paths:

- `get_engine` happy path end-to-end (no SSH tunnel)
- `get_engine` with SSH tunnel: verify tunnel `start()` called and `stop()` called in `finally`
- `get_engine` with SSH tunnel that raises during `start()` — cleanup behaviour
- `_get_tunnel_kwargs` with password auth vs. private key auth (only port-error path covered)
- `_create_engine` exception mapping (`get_dbapi_mapped_exception` re-raise)
- `db_connection_mutator` code path
- `IMPERSONATE_WITH_EMAIL_PREFIX` feature flag path

A component that wraps SSH key material and OAuth2 tokens with no test for those code paths is a gap worth closing before merge.

---

**HIGH-2 — `debug_level` silently dropped from `sshtunnel.open_tunnel()`**

`manager.py:186-191` (new `_get_tunnel_kwargs`) omits the `debug_level` argument that the old `SSHManager.create_tunnel()` forwarded:

```python
# OLD superset/extensions/ssh.py (deleted by this PR)
params = {
    ...
    "debug_level": logging.getLogger("flask_appbuilder").level,
}
return sshtunnel.open_tunnel(**params)
```

Without it, `sshtunnel` uses its own default log level. Operators who relied on SSH tunnel debug output for troubleshooting broken tunnels see nothing. If intentionally removed, please add a comment explaining why; otherwise pass through the logger level.

---

**HIGH-3 — SSH tunnel success log removed with no replacement**

Old `get_sqla_engine` logged a structured message on every successful tunnel connection:

```python
logger.info(
    "[SSH] Successfully created tunnel w/ %s tunnel_timeout + %s ssh_timeout at %s",
    sshtunnel.TUNNEL_TIMEOUT,
    sshtunnel.SSH_TIMEOUT,
    ssh_context.local_bind_address,
)
```

New `get_engine` (manager.py:79-90) has no per-connection tunnel log. `EngineManagerExtension.init_app` logs timeout config once at startup but not at connection time. Operators who tail logs to confirm SSH connectivity after a certificate rotation or tunnel change lose that signal. Suggest a `logger.debug`/`logger.info` inside `_create_tunnel` after `tunnel.start()`.

---

**HIGH-4 — Masked-URL debug log removed**

Old `_get_sqla_engine` logged:
```python
masked_url = self.get_password_masked_url(sqlalchemy_url)
logger.debug("Database._get_sqla_engine(). Masked URL: %s", str(masked_url))
```

This is gone in the new code. It is `DEBUG`-level, so not high-severity, but it was the primary way operators with `DEBUG` logging confirmed which URI was actually used for a connection. If intentionally removed, worth noting in the PR description.

---

### MEDIUM

**MEDIUM-1 — No `issubclass` validation after `load_class_from_name`**

`superset/extensions/engine_manager.py:47-53`:
```python
engine_manager_class: type[EngineManager] = load_class_from_name(
    app.config["ENGINE_MANAGER_CLASS"]
)
self.engine_manager = engine_manager_class(...)
```

If `ENGINE_MANAGER_CLASS` points to a class that doesn't subclass `EngineManager`, the failure happens at instantiation with a generic `TypeError` deep in `init_app`. Adding a guard makes misconfiguration immediately obvious:

```python
if not issubclass(engine_manager_class, EngineManager):
    raise ImproperlyConfigured(
        f"ENGINE_MANAGER_CLASS must be a subclass of EngineManager, got {engine_manager_class}"
    )
```

---

**MEDIUM-2 — Process-wide `sshtunnel` global mutation in `__init__`**

`manager.py:55-56`:
```python
sshtunnel.TUNNEL_TIMEOUT = tunnel_timeout.total_seconds()
sshtunnel.SSH_TIMEOUT = ssh_timeout.total_seconds()
```

This sets process-wide globals on the `sshtunnel` module. The old `SSHManager` did the same and it was fine because it was a singleton. `EngineManager` is also a singleton in practice, so this probably works in production, but:

1. Any test that constructs multiple `EngineManager` instances with different timeouts will silently clobber the globals — the last one constructed wins.
2. Third-party code in the same process that uses `sshtunnel` will have its timeouts overridden without warning.

At minimum, add a comment noting the singleton assumption. Better: pass these as per-call kwargs to `sshtunnel.open_tunnel()` rather than setting the module globals.

---

**MEDIUM-3 — Inline imports in `manager.py` lack circular-import comments**

Five inline imports in `manager.py` have no explanation:

- Line 67: `from superset.utils.oauth2 import check_for_oauth2`
- Line 101: `from superset import is_feature_enabled`
- Line 102: `from superset.extensions import security_manager`
- Line 125: `from superset.utils.oauth2 import get_oauth2_access_token`
- Line 179: `from superset.utils.ssh_tunnel import get_default_port`

Compare with `models/core.py:451` in this same PR: `# Import here to avoid circular imports`. The new file should follow the same convention. Without the comment, reviewers have no signal that these are intentional.

---

### NIT

**NIT-1 — `_`-prefixed methods advertised as the public extension API**

UPDATING.md documents `_get_tunnel_kwargs(ssh_tunnel, uri)` and `_create_tunnel(ssh_tunnel, uri)` as the intended override points for subclassers. By Python convention, a leading `_` signals "not part of the public API." This creates a confusing contract.

Either remove the leading `_` (and add docstrings marking them as override points), or explicitly note in UPDATING.md and the method docstrings that "single underscore by convention but intended as a stable override point."

---

**NIT-2 — RSA-only private key support**

`manager.py:197-200`:
```python
private_key = RSAKey.from_private_key(
    private_key_file,
    ssh_tunnel.private_key_password,
)
```

Only RSA keys are handled. An Ed25519 or ECDSA private key fails here with a cryptic `paramiko.ssh_exception.SSHException: not a valid RSA private key file`. This is carry-over from the old `SSHManager`, but since `_get_tunnel_kwargs` is now the documented subclassing hook, a comment acknowledging the limitation (or a pointer to overriding with `Ed25519Key`) would save the next person who tries a modern key type.

---

**NIT-3 — Deprecation warning has no removal version**

The `nullpool` deprecation `warnings.warn` in `models/core.py` says "deprecated and ignored" but doesn't indicate when it will be removed. Callers need to know their urgency. Suggest:

```
"deprecated and ignored; the engine manager always uses NullPool. "
"Will be removed in Superset 8.0."
```

---

**NIT-4 — Frontend changes appear unrelated**

The four TypeScript files (`EmbeddedContextProviders.tsx`, `RootContextProviders.tsx`, `SubMenu.tsx`, `CollectionControl/index.tsx`) contain only minor type-assertion fixes and a one-line reformat — nothing to do with the EngineManager. They add noise to the diff of an already complex infrastructure change. Were these intentionally included or accidentally rebased in?

---

**NIT-5 — `Any` for `security_manager` in `DBConnectionMutator`**

`superset_typing.py`, `DBConnectionMutator`:
```python
DBConnectionMutator: TypeAlias = Callable[
    [URL, dict[str, Any], str | None, Any, "QuerySource | None"],  # 4th param = security_manager
    tuple[URL, dict[str, Any]],
]
```

The 4th param is `security_manager`, typed `Any`. This loses static type checking for all `DB_CONNECTION_MUTATOR` implementations. `SupersetSecurityManager` (or a minimal `Protocol`) would give better IDE support. Acceptable to defer, but worth tracking.

aminghadersohi (commented):
## PR #34826 — `feat: a manager for SQLAlchemy engines`

**HEAD SHA:** `75a64b3062dabdb18e8beb05df53029e33b9cc8d`
**Author:** betodealmeida
**Review round:** 3 (prior Amingor reviews on 2025-12-04 and 2026-05-11 ×2 at same HEAD SHA)
**Code change since last review:** None — HEAD SHA unchanged from 2026-05-11 reviews.

> Per REVIEW_WORKFLOW.md this PR would normally be skipped (author hasn't pushed new commits since last review), but explicitly requested here. Full 20-scan re-run below; all prior findings confirmed.

---

### Scan coverage (all 20, HEAD 75a64b3)

1. Inline imports without circular-import comment: **FOUND** (BLOCKER-1)
2. SQL injection vectors: 0
3. Hardcoded credentials: 0
4. Broad `except Exception` without narrowing: 1 — `manager.py:169`, immediately re-raises via `get_dbapi_mapped_exception`; acceptable
5. Sensitive info in error messages: 0
6. Module-level mutable globals: 0
7. `print` in production code: 0
8. `db.session.commit` without transaction guard: 0
9. JWT decode without verification: N/A
10. Hardcoded timeouts: 0 — all from config
11. Test methods missing `-> None`: 0
12. `Any` type in new typed aliases: 1 — `DBConnectionMutator` 4th param (NIT-5 in prior review)
13. Module-level app-context refs: 0 — `LocalProxy` used correctly
14. SQL injection in new queries: 0
15. Private symbol imports from external modules: 0
16. `or`-default collapsing falsy: 0
17. Compound boolean without parens: 0
18. Migration boolean filter: N/A
19. Async client without timeout: N/A
20. `asyncio.get_event_loop` in async context: N/A

---

### BLOCKER (unaddressed since 2026-05-11)

**BLOCKER-1 — Five inline imports in `manager.py` with no circular-import comment**

```python
# line 67 (inside get_engine):
from superset.utils.oauth2 import check_for_oauth2

# lines 101-102 (inside _get_engine_args):
from superset import is_feature_enabled
from superset.extensions import security_manager

# line 125 (inside _get_engine_args):
from superset.utils.oauth2 import get_oauth2_access_token

# line 179 (inside _get_tunnel_kwargs):
from superset.utils.ssh_tunnel import get_default_port
```

Each needs a `# circular import: <cycle chain>` comment. Example for lines 101-102: `extensions/__init__.py` imports `EngineManagerExtension` → `engines/manager.py`, so importing `extensions` at module level would cycle back. The cycle is real; the comments are just missing.

Compare with `models/core.py:450` in this same PR — that import correctly has:
```python
# Import here to avoid circular imports
from superset.extensions import engine_manager_extension
```

**BLOCKER-2 — `import warnings` inlined in `models/core.py:441`**

`warnings` is a stdlib module with zero circular import risk. Move it to the module-level import block at the top of `core.py`.

*Prior inline reply (betodealmeida, 2026-02-04):* The reply on line 431 discusses `poolclass` defaults — it does not address the `import warnings` concern.

---

### HIGH (unaddressed since 2026-05-11)

**HIGH-1 — Resource leak if `tunnel.start()` raises inside `_create_tunnel`**

`manager.py:174-176`: `sshtunnel.open_tunnel()` allocates the forwarder; if `tunnel.start()` then throws, the forwarder is never cleaned up. The caller's `try/finally` at lines 81-88 only covers after `_create_tunnel` returns:

```python
# Fix:
tunnel = sshtunnel.open_tunnel(**kwargs)
try:
    tunnel.start()
except Exception:
    tunnel.stop()
    raise
return tunnel
```

**HIGH-2 — Process-wide `sshtunnel` global mutation in `__init__`**

`manager.py:55-56` sets `sshtunnel.TUNNEL_TIMEOUT` and `sshtunnel.SSH_TIMEOUT` as module-level globals. Fine in production (singleton), but tests constructing multiple `EngineManager` instances with different timeouts silently clobber each other. `sshtunnel.open_tunnel()` accepts per-call `ssh_timeout` and `ssh_tunnel_timeout` kwargs — prefer those over module-global mutation.

**HIGH-3 — Sparse test coverage for a credential-handling class**

`manager_test.py` has 3 tests for a ~200-line class that owns SSH key material, OAuth2 tokens, and user impersonation. Missing paths:
- `get_engine` happy path (no tunnel)
- `get_engine` with tunnel: `start()` called; `stop()` called in `finally`
- `_create_tunnel` failure path: tunnel cleaned up
- `_get_tunnel_kwargs` with password auth and private-key auth (only port-error case covered)
- `_create_engine` DBAPI exception mapping
- `db_connection_mutator` path
- `IMPERSONATE_WITH_EMAIL_PREFIX` feature-flag branch

**HIGH-4 — `debug_level` silently dropped from `sshtunnel.open_tunnel()`**

Old `SSHManager.create_tunnel` forwarded `debug_level=logging.getLogger("flask_appbuilder").level`. New `_get_tunnel_kwargs` omits it. UPDATING.md notes the behavioral change, but there is no inline comment in `_get_tunnel_kwargs` for future maintainers. Add:
```python
# debug_level intentionally omitted (vs old SSHManager.create_tunnel).
# Subclassers that need the old behavior should add it in _get_tunnel_kwargs.
```

**HIGH-5 — SSH tunnel connection log removed**

Old `get_sqla_engine` logged `[SSH] Successfully created tunnel w/ ...` on every successful tunnel. New `_create_tunnel` has no per-connection log. Add a `logger.debug` after `tunnel.start()` so operators can confirm SSH connectivity at runtime.

---

### MEDIUM (unaddressed since 2026-05-11)

**MEDIUM-1 — No `issubclass` guard after `load_class_from_name`**

`engine_manager.py:47-53`: A misconfigured `ENGINE_MANAGER_CLASS` (valid path, wrong type) fails at the first `get_engine()` call under user traffic with an opaque `AttributeError`. Add at startup:
```python
if not issubclass(engine_manager_class, EngineManager):
    raise TypeError(f"ENGINE_MANAGER_CLASS must be a subclass of EngineManager, got {engine_manager_class}")
```

**MEDIUM-2 — Constructor `tunnel_timeout` default mismatches config default**

`manager.py:249`: `tunnel_timeout=timedelta(seconds=30)`, but `config.py` has `SSH_TUNNEL_TIMEOUT_SEC = 10.0`. Direct instantiation (tests, subclasses, tooling) gets 3× the configured default. Align or make the parameter required.

---

### NIT (unaddressed since 2026-05-11)

**NIT-1 — Underscore-prefixed override points in public API docs**

UPDATING.md documents `_get_tunnel_kwargs` and `_create_tunnel` as subclass override points. By Python convention `_` signals "not public". Either drop the underscore or add a caveat in UPDATING.md.

**NIT-2 — RSA-only private key support**

`_get_tunnel_kwargs` uses `RSAKey.from_private_key` — Ed25519/ECDSA keys fail with a cryptic SSHException. Note the limitation with a comment. (Also: PR #40139 adds polymorphic key support to the paramiko layer; if merged into master before this PR, confirm it's not regressed by the `RSAKey`-only path here.)

**NIT-3 — Deprecation warning has no removal version**

`models/core.py`: `"deprecated and ignored"` with no target version. Suggest `"Will be removed in Superset X.0."`.

**NIT-4 — Frontend changes explained**

betodealmeida confirmed the 4 TypeScript files were lint fixes only. No functional concerns.

---

### Notes on prior comment responses

- **codeant-ai re `check_for_oauth2` not wrapping engine creation:** Author correctly showed `EngineManager.get_engine` does wrap with `check_for_oauth2`. That bot comment was a false positive.
- **`connect_args` rebinding after `adjust_engine_params`:** The pattern is identical to the deleted `_get_sqla_engine`. Not a regression introduced by this PR.
- **GitHub Advanced Security bot re SHA256 password hashing:** False positive — `ssh_tunnel.password` is passed directly to `sshtunnel.open_tunnel`, not hashed.

---

### Verdict

Two full reviews were posted against this exact commit on 2026-05-11 with identical scope. The code is well-structured and the architecture is sound — `EngineManager` is a clean extraction. Outstanding blockers before merge:

1. **BLOCKER-1**: Add `# circular import` comments to the 5 inline imports in `manager.py` (one-line fix each).
2. **BLOCKER-2**: Move `import warnings` to module-level in `models/core.py`.
3. **HIGH-1**: Guard `_create_tunnel` against `tunnel.start()` failure with `try/except/stop()`.

HIGH-3 (test coverage) and HIGH-4/HIGH-5 (logging) are worth addressing before merge but not strictly blocking.

rusackas (commented):
Clearing my review queue, so here are some bot comments for ya as an official review comment ;)

> `test_get_sqla_engine_user_impersonation_email` got dropped and the new `manager_test.py` only asserts `impersonate_user` was called, thus the email-prefix username rewrite under `IMPERSONATE_WITH_EMAIL_PREFIX` isn't covered anymore. Can we port that case over so the prefix behavior stays pinned?

PR #34828, opened 2025-08-22
fix(theming): Icons in ExecutionLogList and Country map chart tooltip theme consistency

rebenitez1802:
### SUMMARY

County Map chart tooltip fixed to match theme
County Map chart tooltip fixed to match theme
This PR fixes two styling issues:

  1. ExecutionLogList Running Icon Size: Fixed size inconsistency of the custom Running icon compared to other Antd status icons
  2. Country Map Tooltip Theme Colors: Fixed tooltip text colors to use theme variables instead of hardcoded values

Alert and Reports:
ReportAndAlerts

Country Map Chart:
CountryMapLight
CountryMapDark

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/alerts/components/AlertStatusIcon.tsx
+++ b/superset-frontend/src/features/alerts/components/AlertStatusIcon.tsx
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { t, SupersetTheme, useTheme } from '@superset-ui/core';
+import { t, SupersetTheme, useTheme, css } from '@superset-ui/core';
 import { Tooltip } from '@superset-ui/core/components';
 import { Icons } from '@superset-ui/core/components/Icons';
 import { AlertState } from '../types';
@@ -32,7 +32,7 @@ function getStatusColor(
     case AlertState.Error:
       return theme.colorErrorText;
     case AlertState.Success:
-      return isReportEnabled ? theme.colorSuccessText : theme.colorErrorText;
+      return theme.colorSuccessText;
     case AlertState.Noop:
       return theme.colorSuccessText;
     case AlertState.Grace:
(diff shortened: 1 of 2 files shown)


superset-frontend/src/features/alerts/components/AlertStatusIcon.tsx
```diff
@@ -95,16 +93,30 @@ export default function AlertStatusIcon({
...
     <Tooltip title={lastStateConfig.label} placement="bottomLeft">
-      <Icon
-        iconSize="m"
-        iconColor={getStatusColor(
-          lastStateConfig.status,
-          isReportEnabled,
-          theme,
-        )}
-      />
+      <span
```
rebenitez1802, line 99: this is hack needed to fix the size issue with this icon, the main reason is that the original SVG uses only the center portion of their viewBox, changing the SVG could fix this without this hack, but this Icon is being used also in SQLlab and Query history so probably changing the SVG can break those.

michael-s-molina:
Thanks @rebenitez1802 for the PR. Could you add a better PR title?

superset-frontend/src/features/alerts/components/AlertStatusIcon.tsx
```diff
@@ -95,16 +93,30 @@ export default function AlertStatusIcon({
...
+        css={
+          isRunningIcon
+            ? css`
+                display: inline-flex;
+                align-items: center;
+                justify-content: center;
+                transform: scale(1.8);
```
sadpandajoe, line 106: Is there a reason why we are hardcoding 1.8 and not using what we do with unitSize or some other variable?

rebenitez1802: this Icon SVG is smaller than the others, using standard sizes (S,M,L,XL) doesn't fit the size needed. so in order to look like the other ones in the screen scaling is easiest solution. I also considered modifying the SVG itself but this icon is used in other components. The other posible solution could be duplicating this icon but bigger, but to be honest i think is better to change size using styles.

mistercrunch (approved):
LGTM

mistercrunch:
@rusackas needs a CODEOWNER

rusackas (approved):

rusackas:
LGTM, though we ought to fix the positioning too, one of these days, so it shows up in the right place :)

PR #34836, opened 2025-08-24
feat(api): dataset read API uuid support

dankor:
### SUMMARY

It's a followup of https://github.com/apache/superset/pull/29573

Added uuid support:
- GET dataset/`uuid_or_id`
- GET dataset/`uuid_or_id`/related_objects
- POST chart/data {"datasource": { "id": `uuid_or_id`, "type": "table"}}

image

image

image

### TESTING INSTRUCTIONS

Run the same request and replace id with it uuid. 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


dankor:
@mistercrunch @michael-s-molina @rusackas 
As promised, I’m continuing to add **id_or_uuid** support bit by bit. Please take a look.

superset/charts/schemas.py
```diff
@@ -1130,8 +1130,9 @@ class AnnotationLayerSchema(Schema):
 
 class ChartDataDatasourceSchema(Schema):
     description = "Chart datasource"
-    id = fields.Integer(
-        metadata={"description": "Datasource id"},
+    # Coudn't find a better way to mix uuid and id
+    id = fields.Raw(
```
mistercrunch: seems like this would work -> https://chatgpt.com/share/68ad1737-e5a8-8010-86f1-f4879f014f8c , do it once and reuse where needed.

dankor: @mistercrunch Thanks! I reworked.

Please take another look, included `marshmallow-union` dependency.

superset/daos/base.py
```diff
@@ -47,12 +47,41 @@ class BaseDAO(Generic[T]):
...
+
+        if model_id_or_uuid.isdigit():
+            filter = id_column == int(model_id_or_uuid)
+        else:
+            filter = uuid_column == model_id_or_uuid
+        try:
+            return query.filter(filter).one_or_none()
+        except StatementError:
+            # can happen if int is passed instead of a string or similar
+            return None
```
mistercrunch, line 83: mmmh, shouldn't we either handle int or let it raise here? Not sure about what the implications are or might be...

mistercrunch: i guess mypy should complain if using an `int` anywhere in the codebase. Since this is a base function going to get reused it could be good to just support `int` here. Might even be a tiny bit cheaper to validate the type than `isdigit()` is, but would run both (pseudo: `type(v) == int or isdigit()`)

mistercrunch (commented):
added a minor comment but otherwise LGTM

superset/daos/base.py
dankor: Well, you're right, the comment is a little bit irrelevant. Fixed. 

Not sure about `int` optimization thought. Don't we plan to deprecate `id` eventually?

mistercrunch: I think you're right. Doesn't matter and mypy should catch error (would be good to confirm). While we're in here and setting the pattern, should we support short-shas in here too? Would make for nicer looking urls, pretty standard nowadays (git/github/...), not sure if it's worth the complexity, collision risks, ...

dankor: Why is this useful? We don’t currently expose these URLs in the UI. Maybe in the next iteration we should reintroduce the slug + short id URL style (similar to Medium), for example:

- dashboard/unicode-test-91c65d59d37/
- chart/unicode-cloud-b86ba33fc5cd/

That said, this looks like a complex and potentially breaking change. It could be useful for dashboards, especially as a replacement for custom slugs. But charts and datasets don’t follow the same pattern — they just redirect you to the `/explore` endpoint.

I’d prefer to discuss this in a different setting, since it will require some brainstorming.

mistercrunch (approved):
LGTM

PR #34843, opened 2025-08-25
fix(dashboard): Anchor link positions

kgabryje:
### SUMMARY
1. Fixes anchor link in tabs always visible - it should only display on hover
2. Fixes anchor link in headers always visible - it should only display on hover
3. Fixes incorrect position of anchor link in headers - it was displayed in a separate line, now it's displayed next to text
4. Adds a small spacing between the url text and copy icon (visible when anchor popover is open)

Before:

https://github.com/user-attachments/assets/ff600f73-c7e3-4c4c-933c-584b7040bdaa

After:

https://github.com/user-attachments/assets/e363fde0-c603-41f4-ba77-9431b70ee6fe

### TESTING INSTRUCTIONS
1. Open a dashboard
2. Add some tabs and headers components
3. Verify that anchor links appear only on hover

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/CopyToClipboard/index.tsx
+++ b/superset-frontend/src/components/CopyToClipboard/index.tsx
@@ -17,7 +17,7 @@
  * under the License.
  */
 import { Component, cloneElement, ReactElement } from 'react';
-import { t } from '@superset-ui/core';
+import { t, css, SupersetTheme } from '@superset-ui/core';
 import copyTextToClipboard from 'src/utils/copy';
 import { Tooltip } from '@superset-ui/core/components';
 import withToasts from '../MessageToasts/withToasts';
@@ -104,7 +104,14 @@ class CopyToClip extends Component<CopyToClipboardProps> {
     return (
       <span css={{ display: 'inline-flex', alignItems: 'center' }}>
         {this.props.shouldShowText && this.props.text && (
-          <span data-test="short-url">{this.props.text}</span>
+          <span
+            data-test="short-url"
+            css={(theme: SupersetTheme) => css`
+              margin-right: ${theme.sizeUnit}px;
+            `}
+          >
+            {this.props.text}
+          </span>
         )}
         {this.renderTooltip('pointer')}
       </span>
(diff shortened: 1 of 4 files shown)


rusackas (approved):

superset-frontend/src/components/CopyToClipboard/index.tsx
```diff
@@ -104,7 +104,14 @@ class CopyToClip extends Component<CopyToClipboardProps> {
     return (
       <span css={{ display: 'inline-flex', alignItems: 'center' }}>
         {this.props.shouldShowText && this.props.text && (
-          <span data-test="short-url">{this.props.text}</span>
+          <span
```
mistercrunch, line 107: different ways to do this, but throughout the code base the emerging pattern is to `const theme = useTheme()` and then you can `style={{ marginRigth: ``${theme.sizeUnit}px`` }}`. That way types comes directly.

Would be cool to have a `css` and `styled` in @superset-ui/core but couldn't make it work easily when I tried

superset-frontend/src/dashboard/components/gridComponents/Header.jsx
```diff
@@ -75,6 +75,18 @@ const HeaderStyles = styled.div`
       font-size: ${theme.fontSizeXXL}px;
     }
 
+    .anchor-link-container {
+      display: inline;
+      line-height: 0;
+      vertical-align: bottom; /* trick to align the anchor with text */
+      opacity: 0;
+      transition: opacity 0.2s ease-in-out;
```
mistercrunch: nit: can use `theme.motionDurationMid` here, I believe it's a string -> `0.2s`

mistercrunch (approved):
small nits but otherwise LGTM

superset-frontend/src/components/CopyToClipboard/index.tsx
kgabryje: We can't `useTheme` here because it's a class component. We could use a withTheme HOC but I think `css={theme => ...` might be the cleanest here

PR #34848, opened 2025-08-25
fix(docker): configure gunicorn for DuckDB thread safety in docker-compose-light

mistercrunch:
DuckDB is not fully thread-safe in all contexts, causing database locking issues when used with Gunicorn's default gthread worker class. This change configures docker-compose-light.yml to use multiple sync workers (8) with single threading (1 thread per worker) instead of the default threaded configuration to prevent database conflicts.

### DIFF
--- a/docker-compose-light.yml
+++ b/docker-compose-light.yml
@@ -116,14 +116,13 @@ services:
       DATABASE_HOST: db-light
       DATABASE_DB: superset_light
       POSTGRES_DB: superset_light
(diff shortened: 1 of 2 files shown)


eschutho (approved):

rusackas (approved):
LGTM, but does this need docs (or a note in UPDATING.md at least) in case people set this differently for whatever reasons, or may be negatively impacted?

PR #34831, opened 2025-08-23
feat: Use DuckDB for examples data in Docker development environment

mistercrunch:
This change migrates the Docker development setup to use DuckDB for examples data storage while keeping PostgreSQL for Superset metadata. This significantly improves development environment startup time, resource usage and example-related query performance in dev envs.

Key changes:
- Add duckdb-engine to development dependencies and upgrade to v0.17.0
- Configure Docker to download pre-built examples.duckdb during image build
- Update docker-compose files to use DuckDB URIs for examples data
- Add environment variable override support for SQLALCHEMY_EXAMPLES_URI
- Fix DuckDB engine spec default_driver (was empty, now "duckdb_engine")
- Work around cursor.description bug in duckdb-engine driver

The examples.duckdb file (~37MB) is downloaded from apache-superset/examples-data during Docker build using the LOAD_EXAMPLES_DUCKDB build arg. We considered alternative approaches like downloading during container startup, but injecting the file deep in the Docker layer stack provides better Docker layer caching and faster subsequent builds.

This hybrid approach uses PostgreSQL for reliable metadata storage and DuckDB for fast analytical examples data, eliminating examples loading time while maintaining full functionality.

Technical notes:
- Includes workaround for duckdb-engine cursor.description bug I posted while working on this: https://github.com/Mause/duckdb_engine/issues/1322
- Examples data baked into container image, no runtime loading needed
- Compatible with existing Claudette workflows and testing
- Build arg defaults to true for development environments

PR #34851, opened 2025-08-26
fix(docker): remove --with-threads from remaining dev workflows for DuckDB compatibility

mistercrunch:
Follow-up to #34831 and #34848 - completing the threading removal for DuckDB compatibility. Removes --with-threads from Makefile flask-app target and docker-bootstrap.sh, and cleans up unused gunicorn env vars in docker-compose-light.yml since it uses flask dev server, not gunicorn.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ js-format:
 	cd superset-frontend; npm run prettier
 
 flask-app:
-	flask run -p 8088 --with-threads --reload --debugger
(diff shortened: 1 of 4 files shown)


betodealmeida (approved):
> Some people, when confronted with a problem, think, “I know, I’ll use threads,” and then two they hav erpoblesms.

https://nedbatchelder.com/blog/201204/two_problems.html

PR #34858, opened 2025-08-26
fix: SelectControl default sort numeric choices by value

### LINKED ISSUE #35008: Sorting in filter values
### Bug description

The version is 6.0.0rc1 (but it was the same in 5)
The order of int values is incorrect:
Image

as far as it's Int type, I expect: 1,2,3...

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

kgabryje:
### SUMMARY
The numeric choices passed to SelectControl were being sorted alphabetically, leading to confusing order of the items. This PR implements sorting by value if ALL values passed through `props.choices` or `props.options` are numeric.
We can still pass `sortComparator` prop which overrides the default logic, so it's possible to create a control with custom order of items.

Before:

image

After:

image

### TESTING INSTRUCTIONS
1. Open Table chart
2. Verify that `Server page length` and `Row limit` controls have their options sorted by value, not alphabetically

### ADDITIONAL INFORMATION

- [ ] Has associated issue: Fixes #35008
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/explore/components/controls/SelectControl.jsx
+++ b/superset-frontend/src/explore/components/controls/SelectControl.jsx
@@ -87,6 +87,56 @@ const defaultProps = {
   valueKey: 'value',
 };
 
+const numberComparator = (a, b) => {
+  const aVal = typeof a.value === 'number' ? a.value : parseFloat(a.value);
+  const bVal = typeof b.value === 'number' ? b.value : parseFloat(b.value);
+  return aVal - bVal;
+};
+
+export const areAllChoiceValuesNumbers = (choices, valueKey = 'value') => {
+  if (!choices || choices.length === 0) {
+    return false;
+  }
+  return choices.every(c => {
+    if (Array.isArray(c)) {
+      const [value] = c;
+      return typeof value === 'number';
+    }
+    if (typeof c === 'object' && c !== null) {
+      return typeof c[valueKey] === 'number';
+    }
+    return typeof c === 'number';
+  });
+};
+
+export const areAllOptionValuesNumbers = (options, valueKey = 'value') => {
+  if (!options || options.length === 0) {
+    return false;
+  }
+  return options.every(o => typeof o[valueKey] === 'number');
+};
+
+export const getSortComparator = (
+  choices,
+  options,
+  valueKey,
+  explicitComparator,
+) => {
+  if (explicitComparator) {
+    return explicitComparator;
+  }
+
+  if (options && areAllOptionValuesNumbers(options, valueKey)) {
+    return numberComparator;
+  }
+
+  if (choices && areAllChoiceValuesNumbers(choices, valueKey)) {
+    return numberComparator;
+  }
+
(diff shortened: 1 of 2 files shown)


superset-frontend/src/explore/components/controls/SelectControl.jsx
```diff
@@ -87,6 +87,56 @@ const defaultProps = {
   valueKey: 'value',
 };
 
+const numberComparator = (a, b) => {
+  const aVal = typeof a.value === 'number' ? a.value : parseFloat(a.value);
```
Antonio-RiveroMartnez: Isn't the type already checked in `areAllChoiceValuesNumbers` and `areAllOptionValuesNumbers`?

superset-frontend/src/explore/components/controls/SelectControl.jsx
```diff
@@ -87,6 +87,56 @@ const defaultProps = {
...
+  const bVal = typeof b.value === 'number' ? b.value : parseFloat(b.value);
+  return aVal - bVal;
+};
+
+export const areAllChoiceValuesNumbers = (choices, valueKey = 'value') => {
```
Antonio-RiveroMartnez: this method and `areAllOptionValuesNumbers` seems very similar, and seems like this could work for both cases? so might be good to just have one `areAllValuesNumbers(...)`?

superset-frontend/src/explore/components/controls/SelectControl.jsx
```diff
@@ -87,6 +87,56 @@ const defaultProps = {
...
+
+export const areAllOptionValuesNumbers = (options, valueKey = 'value') => {
+  if (!options || options.length === 0) {
+    return false;
+  }
+  return options.every(o => typeof o[valueKey] === 'number');
+};
+
+export const getSortComparator = (
+  choices,
```
Antonio-RiveroMartnez, line 109: nit: Could have add some types here?

kgabryje: Nope, it's a jsx file 🙂

superset-frontend/src/explore/components/controls/SelectControl.jsx
kgabryje: done!

superset-frontend/src/explore/components/controls/SelectControl.jsx
kgabryje: done!

superset-frontend/src/explore/components/controls/SelectControl.jsx
Antonio-RiveroMartnez: Oh my bad, didn't look at that 😓

Antonio-RiveroMartnez (approved):
lgtm

PR #34859, opened 2025-08-26
refactor(deckgl): update deck.gl charts to use new api

DamianPendrak:
### SUMMARY
This PR implements a refactor of deck.gl charts from legacy API for all of the charts, except for Multi Layer, resolving multiple issues around them. I was not able to refactor the Multi Layer Chart as it doesn't create any query, and refactoring to the new API will change the chart's behavior - it would require all included layers to be from the same dataset.
The PR also includes a fix for the Multi Layer Chart to display "No results" in the results tab in Explore instead of an error message.

Screenshot 2025-09-04 at 11 05 07
Screenshot 2025-09-04 at 11 04 41

### TESTING INSTRUCTIONS
All of the deck.gl charts functionality should be preserved, and the filters badge should now be consistent.
The Multi Layer Chart still doesn't have the filters badge, but now it should display "No results" in the results tab in Explore

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/legacy-preset-chart-deckgl/src/CategoricalDeckGLContainer.tsx
+++ b/superset-frontend/plugins/legacy-preset-chart-deckgl/src/CategoricalDeckGLContainer.tsx
@@ -169,12 +169,12 @@ const CategoricalDeckGLContainer = (props: CategoricalDeckGLContainerProps) => {
           }));
         }
         case COLOR_SCHEME_TYPES.color_breakpoints: {
-          const defaultBreakpointColor = fd.deafult_breakpoint_color
+          const defaultBreakpointColor = fd.default_breakpoint_color
             ? [
-                fd.deafult_breakpoint_color.r,
-                fd.deafult_breakpoint_color.g,
-                fd.deafult_breakpoint_color.b,
-                fd.deafult_breakpoint_color.a * 255,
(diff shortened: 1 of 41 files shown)


rusackas:
That's awesome! Let us know when it's ready for testing/review :)

rusackas (approved):

kgabryje:
Amazing job! 2 things:
1. You should do the honors of removing the "legacy-" prefix from the deckgl plugins 😄
2. The deckgl classes from viz.py are unused now, right? Can we remove them?

kgabryje (approved):
Code looks good! Also I did some testing and couldn't break it, but we should do more extensive testing before merging, given that this PR refactors multiple popular viz plugins

DamianPendrak:
@kgabryje, the viz.py is still used by the Multi Layer Chart. I am looking for ways to deprecate this chart and implement a new version. The main problem is that it allows the use of layers from different datasets.

kgabryje:
> @kgabryje, the viz.py is still used by the Multi Layer Chart. I am looking for ways to deprecate this chart and implement a new version. The main problem is that it allows the use of layers from different datasets.

I see, thanks for explanation 👍

richardfogaca:
Great work @DamianPendrak! 💯

PR #34868, opened 2025-08-27
fix(theming): fix TimeTable chart issues

gabotorresruiz:
### SUMMARY

This PR addresses theming and visual issues in Time-series charts while significantly improving the TimeTable component architecture through comprehensive refactoring:

**Theming Fixes:**
- Fixed visual styling issues in Time-series charts
- Resolved functional bugs affecting chart rendering
- Improved consistency with Ant Design theming tokens

**Component Structure Improvements:**
- Converted JavaScript files to TypeScript files
- Split monolithic TimeTable component into smaller, testable components
- Extracted inline components into separate modules with dedicated test files
- Reorganized utility functions into dedicated folders with comprehensive tests

### BEFORE/AFTER SCREENSHOTS

Before:
wrong_icon

After:
correct_icon

-----------------------------------------------------
Before:
time_series_wrong

After:
time_series_correct

-----------------------------------------------------
Before:
wrong_info_icon

After:
correct_info_icon

-----------------------------------------------------
Before:
wrong_visual_drag

After:
correct_visual_drag

-----------------------------------------------------
Before:
image

After:
image

-----------------------------------------------------
Before:
bad_visuals_dark_mode_period_pivot

After:
image

### TESTING INSTRUCTIONS

1. Navigate to a `Time-series Table` chart (or create a new one)
2. Verify charts render correctly with proper theming
3. Test value cell formatting and color coding functionality

- You can also test that the `Time-series Period Pivot` chart is also rendering as expected with theming

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


superset-frontend/packages/superset-ui-core/src/validator/validateNonEmpty.ts
```diff
@@ -26,7 +26,7 @@ export default function validateNonEmpty(v: unknown) {
     v === '' ||
     (Array.isArray(v) && v.length === 0)
   ) {
-    return t('cannot be empty');
+    return t('Cannot be empty');
```
sadpandajoe: Did we confirm that this isn't in any translation files that need to be updated?

superset-frontend/packages/superset-ui-core/src/components/Tooltip/index.tsx
```diff
@@ -23,7 +23,11 @@ import type { TooltipProps, TooltipPlacement } from './types';
 export const Tooltip = ({ overlayStyle, ...props }: TooltipProps) => (
   <AntdTooltip
     styles={{
-      body: { overflow: 'hidden', textOverflow: 'ellipsis' },
+      body: {
+        overflow: 'hidden',
+        textOverflow: 'ellipsis',
+        textAlign: 'center',
```
sadpandajoe: Doesn't this affect all tooltips? Do we want everything to be center aligned?

sadpandajoe:
@kasiazjc @yousoph i've spun up an ephemeral if you'd like to take a look at it

superset-frontend/packages/superset-ui-core/src/components/Tooltip/index.tsx
mistercrunch: yup careful on touching core components as will ripple through the app...

gabotorresruiz: Sorry, my bad @sadpandajoe @mistercrunch. I was just testing something. I still think tooltip texts looks better center-aligned lol, but that’s not my call to make. Reverting this change

superset-frontend/packages/superset-ui-core/src/validator/validateNonEmpty.ts
gabotorresruiz: You’re right, but I don’t think we should handle this here. I’ll revert the change, and we can address it later in a separate ticket along with the necessary translation updates.

superset-frontend/packages/superset-ui-core/src/components/IconTooltip/index.tsx
```diff
@@ -43,13 +43,7 @@ export const IconTooltip = ({
   );
   if (tooltip) {
     return (
-      <Tooltip
-        id="tooltip"
-        title={tooltip}
-        placement={placement}
-        mouseEnterDelay={0.3}
-        mouseLeaveDelay={0.15}
```
msyavuz, line 51: Do we not need these delays anymore?

gabotorresruiz: Hey @msyavuz! If we want this to behave the same as the other tooltips, we don’t need those delays. If we prefer to keep two different behaviors, we can leave it as is.

Looping in @kasiazjc and @yousoph to hear their thoughts

gabotorresruiz: Hey @msyavuz, I went a step further, I added support for configuring those delays when needed, while keeping the same default values if nothing is passed. This way, the problem should be solved without changing the existing behavior. Please take a look at my latest commit --> https://github.com/apache/superset/pull/34868/commits/6281e5f825576c38a8e83af944f32d199ff19077

msyavuz (approved):
LGTM! Tested and works fine.

PR #34874, opened 2025-08-27
fix(ui-core): Invalid postTransform process

justinpark:
### SUMMARY
Previously, the processChartProps selector had an inefficient cache logic: it always repeated the entire process of `post(transform(pre(props)))` even if only one of pre, transform, or post was changed.
For example, if only `postChartProps` was updated, it still redundantly repeated the unchanged `pre` and `transform` steps, resulting in performance degradation.

To address this, this commit refactors the process by separating each step and implementing cache selectors. This change allows unnecessary transformation steps to be handled by the cache, improving performance by avoiding redundant computations.
Additionally, we added a missing equality check for `postTransformProps` in ChartRenderer for the correct re-rendering logic.

### TESTING INSTRUCTIONS

specs are added

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/chart/components/SuperChartCore.test.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/chart/components/SuperChartCore.test.tsx
@@ -1,0 +1,75 @@
+import { render, waitFor } from '@testing-library/react';
+
+import {
+  ChartPlugin,
+  ChartMetadata,
+  ChartProps,
+  DatasourceType,
+  getChartComponentRegistry,
+} from '@superset-ui/core';
+
+import SuperChartCore from './SuperChartCore';
+
+describe('processChartProps', () => {
+  const props = {
+    chartType: 'line',
+  };
+  const FakeChart = () => <span>test</span>;
+
+  beforeEach(() => {
+    const metadata = new ChartMetadata({
+      name: 'test-chart',
+      thumbnail: '',
+    });
+    const buildQuery = () => ({
+      datasource: { id: 1, type: DatasourceType.Table },
+      queries: [{ granularity: 'day' }],
+      force: false,
+      result_format: 'json',
+      result_type: 'full',
+    });
+    const controlPanel = { abc: 1 };
+    const plugin = new ChartPlugin({
+      metadata,
+      Chart: FakeChart,
+      buildQuery,
+      controlPanel,
+    });
+    plugin.configure({ key: props.chartType }).register();
+  });
+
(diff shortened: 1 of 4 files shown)


superset-frontend/packages/superset-ui-core/src/chart/components/SuperChartCore.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {
+  ChartPlugin,
+  ChartMetadata,
+  DatasourceType,
+  getChartComponentRegistry,
+} from '@superset-ui/core';
+
+import SuperChartCore from './SuperChartCore';
+
+describe('processChartProps', () => {
```
michael-s-molina: Could you use `test` instead of `describe/it`?

superset-frontend/packages/superset-ui-core/src/chart/components/SuperChartCore.tsx
```diff
@@ -89,27 +89,76 @@ export default class SuperChartCore extends PureComponent<Props, {}> {
...
         preTransformProps?: PreTransformProps;
-        transformProps?: TransformProps;
-        postTransformProps?: PostTransformProps;
       }) => input.chartProps,
       input => input.preTransformProps,
+    ],
+    (chartProps, pre = IDENTITY) => pre(chartProps),
+  );
+
+  /**
```
michael-s-molina, line 105: Could you adjust the comments for `transformSelector` and `postSelector` to:

> Memoized the function so it will not recompute and return the previous value unless one of the input arguments have changed.

superset-frontend/packages/superset-ui-core/src/chart/components/SuperChartCore.tsx
```diff
@@ -89,27 +89,76 @@ export default class SuperChartCore extends PureComponent<Props, {}> {
...
+        postTransformProps?: PostTransformProps;
+      }) => input.chartProps,
       input => input.postTransformProps,
     ],
-    (chartProps, pre = IDENTITY, transform = IDENTITY, post = IDENTITY) =>
-      post(transform(pre(chartProps))),
+    (transformedChartProps, post = IDENTITY) => post(transformedChartProps),
   );
 
+  processChartProps = ({
```
michael-s-molina, line 137: Could you add a docstring for this function?

michael-s-molina (approved):

PR #34875, opened 2025-08-28
fix: use local db model in migration

eschutho:
### SUMMARY

Using a local version of the db model so that this migration can be run in the future when there are changes to the model in future versions and this older migration needs to run before the newer one. 
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


betodealmeida (approved):
Ah, thanks!!

sadpandajoe (approved):

mistercrunch (approved):
LGTM

rusackas:
@eschutho do we still want to keep this one open?

rusackas:
This hit my stale draft radar again... I'll probably politely close it out next time it hits another 90+ days ;)

rusackas:
@eschutho I keep threatening to close this, but it's actually the right migration pattern (local model so it survives future schema changes) and it has a test. I'll rebase it since I'm here to see if we can get it green, rather than close it. Holler if you'd rather I let it go.

PR #34876, opened 2025-08-28
feat: Add ECharts options overrides to theme system

mistercrunch:
## Summary

Implements fine-grained theming control for ECharts visualizations through the Superset theme system. Users can now override ECharts configuration options globally or per chart type without modifying individual chart configurations.

## Key Features

- **Global overrides**: Apply settings to all ECharts visualizations via `echartsOptionsOverrides`
- **Chart-specific overrides**: Target specific chart types via `echartsOptionsOverridesByChartType`
- **Smart merging**: Deep object merging with array replacement using lodash `mergeWith`
- **Proper precedence**: Chart-specific > Global > Plugin defaults > Base theme

## Technical Implementation

- Added theme interface properties to `SupersetTheme` with TypeScript support
- Enhanced core `Echart.tsx` component with intelligent merge logic at single injection point
- Updated chart components to pass `vizType` for chart-specific targeting
- Implemented custom array replacement logic to ensure proper ECharts configuration behavior

## Testing & Validation

- Comprehensive unit tests covering merge behavior and edge cases
- All pre-commit hooks passing (ESLint, TypeScript, formatting)
- Documentation updated with usage examples and best practices

## Usage Example

```javascript
// Theme configuration
{
  echartsOptionsOverrides: {
    grid: { left: "10%", right: "10%" },
    tooltip: { backgroundColor: "rgba(0,0,0,0.8)" }
  },
  echartsOptionsOverridesByChartType: {
    echarts_pie: { legend: { orient: "vertical" } },
    echarts_timeseries: { xAxis: { axisLabel: { rotate: 45 } } }
  }
}
```

This enables powerful corporate branding and custom styling capabilities while maintaining backward compatibility.


msyavuz (commented):
I might be missing something but this theme config in crud view doesn't change anything
```
{
  "token": {
    "colorPrimary": "#2893B3"
  },
  "echartsOptionsOverrides": {
    "grid": {
      "left": "10%",
      "right": "10%"
    },
    "tooltip": {
      "backgroundColor": "rgba(0, 0, 0, 0.8)"
    }
  },
  "echartsOptionsOverridesByChartType": {
    "echarts_pie": {
      "legend": {
        "orient": "vertical",
        "right": 10
      }
    }
  }
}
```

mistercrunch:
Oh as in the `mergeReplaceArrays` function doesn't merge as specified? I have to admit I didn't run manual tests [yet]

mistercrunch:
Tested with:
```
{
  "token": {
    "colorPrimary": "#55AA22",
    "echartsOptionsOverrides": {
      "color": [
        "#FF6B6B",
        "#4ECDC4", 
        "#45B7D1",
        "#96CEB4",
        "#FFEAA7"
      ],
      "tooltip": {
        "backgroundColor": "#0F0",
        "borderColor": "#F00"
      }
    },
    "echartsOptionsOverridesByChartType": {
      "pie": {
        "tooltip": {
          "backgroundColor": "#FF0",
          "borderColor": "#A0A"
        }
      },
      "echarts_timeseries": {
        "tooltip": {
          "backgroundColor": "#FF0",
          "borderColor": "#0A0"
        }
      }
    }
  }
}
```
Screenshot 2025-09-03 at 3 03 54 PM
Screenshot 2025-09-03 at 3 03 49 PM

superset-frontend/packages/superset-ui-core/src/utils/lodash.ts
```diff
@@ -0,0 +1,52 @@
...
+export function mergeReplaceArrays<T = any>(...sources: any[]): T {
+  const replaceArrays = (objValue: any, srcValue: any) => {
+    if (Array.isArray(srcValue)) {
+      return srcValue; // Replace arrays entirely
+    }
+    return undefined; // Let lodash handle object merging for non-arrays
+  };
+
+  return mergeWith({}, ...sources, replaceArrays);
+}
```
villebro: I think calling this utility `lodash` is confusing, as it collides with the actual `lodash` package name. Furthermore, this appears to be a very thin wrapper around `loads.mergeWith`. If we really need a util for this, I propose calling it something more generic that decouples it from the implementation. If we ever swap out `lodash` for something else, this name will no longer be accurate.

superset-frontend/packages/superset-ui-core/src/theme/types.ts
```diff
@@ -127,6 +127,15 @@ export interface SupersetSpecificTokens {
   // Spinner-related
   brandSpinnerUrl?: string;
   brandSpinnerSvg?: string;
+
+  // ECharts-related
+  /** Global ECharts configuration overrides applied to all chart types */
+  echartsOptionsOverrides?: any;
```
villebro, line 133: While ECharts is our main viz library, we do have others that may also need library-level overriding. Also, in the future we expect other viz extensions to pop up (e.g. D3 based), and for those this approach of overriding "the option" will not be the correct solution. So I wonder if we should instead leave this more generic, so that you can pass some type of mutator per library type, and that then applies the necessary changes?

villebro (commented):
Some quick thoughts.

superset-frontend/packages/superset-ui-core/src/theme/types.ts
rusackas: Agreed. The more I think about this, it kind of couples ECharts to the core codebase fairly tightly, where we're also making an active effort to _decouple_ these libraries/visualizations from the core, to have a more extensible/agnostic Superset. I'm a little worried about the proliferation of themes within a theme... it might not be long until we have the Superset Theme containing sets of overrides for any number of viz libraries that get loaded in as extensions.

A couple of alternative options have come up before - though they don't seem quite sufficient:
1) Having an "advanced" control that lets you override this on a per-chart basis
2) Adopting the tokens we want to apply to viz components _in general_ into the core Superset theme, and then adding a means of mapping them in the chart implementations (e.g. coding ECharts extensions to map Superset's `borderRadius` to Echart's `barBorderRadius` for the bar chart, `radius` for Pie, etc.)

Another way to handle that *might* be to brainstorm a way to make Themes themselves be extendable. I.e. if you add one or more extensions for ECharts, you could unlock an additional panel to tweak ECharts options... and get even more mapping control surfaces if you add new things to the mix like Observable Plot / Keppler / Plotly / Visx / Whatever :)

msyavuz: I gave this some thought as well but I don't think this couples echarts with core that much. Having options for each viz library in the theme config without abstractions is simple enough and if we make the theme extendible it seems to solve our issues.

Suppose we make the theme itself extendible and use keys like echartsOptionsOverrides etc. If user wants to create a viz plugin they just need to extend the theme with their own keys like d3ChartsOptions and merge the options there just like Max did here with Echarts. That looks like a simple and straightforward flow to me.

superset-frontend/packages/superset-ui-core/src/utils/lodash.ts
msyavuz: Agreed, maybe we can specify the usage and call it something like `mergeThemeConfigs`?

msyavuz:
Tested manually and works great now!

superset-frontend/packages/superset-ui-core/src/utils/lodash.ts
mistercrunch: intention was to extend lodash with things that are on the edge of what it can or should do. As in "lodash-related utilities". Could also file it in something like `utils/merge.ts` (?)

mistercrunch: btw for `echartsOptions` merging, maybe merging/stacking arrays is fine (?) Felt wrong to me. Maybe it's right in some cases, wrong in others.

villebro: I much prefer `utils/merge.ts` over `utils/lodash.ts`.

superset-frontend/packages/superset-ui-core/src/theme/types.ts
villebro: I like @rusackas 's thinking here. I suspect we may need to be open to doing some exploration here, as finding the ultimate solution upfront may be challenging. So as a compromise, maybe we could let this logic in, with the caveat that this is highly experimental, and will likely break in future iterations.

> Suppose we make the theme itself extendible and use keys like echartsOptionsOverrides etc. If user wants to create a viz plugin they just need to extend the theme with their own keys like d3ChartsOptions and merge the options there just like Max did here with Echarts. That looks like a simple and straightforward flow to me.

@msyavuz I'm not convinced this will work. Different viz libraries apply overrides differently. For instance, ECharts is declarative with a single huge "option" object, while eg. D3 has a much more imperative API, with lots of chained function calls to arrive at the final chart. So applying similar changes on D3 will likely require a very different approach to ECharts.

mistercrunch: Yes it's going to be hard to expose something perfect here. Some bulk thoughts:
- short of exposing a way for people to use code ( think `echartOptionsMutator(vizType, theme, echartOptions)` ) there's only so much you can do with static objects like a json blob. Maybe extensions could expose this kind of hook, idk
- `echartOptions` very much the theming interface for echarts, lots you can do here, especially if you're builidng your theme object as code (think embedded-type use case)
- namespacing overrides could be powerful, by that I mean registering arbitrary targets in Superset as antd "components", meaning you could go something like this:
```
const theme = {
  "component": {
    "anyComponentInSuperset": {
      "colorText": "#F00";
      "componentSpecificToken": "whatever",
    },
  }
};
```

mistercrunch: I looked into doing this early on but bailed, but any component could define its namespace and read from it with something like:
```
# in say EchartPieChart
const theme = useThemeWithComponentOverrides("EchartPieChart");
# ^^^ returns a full theme object with EchartPieChart overrides applied.
```

superset-frontend/packages/superset-ui-core/src/utils/lodash.ts
mistercrunch: done!

superset-frontend/plugins/plugin-chart-echarts/src/Bubble/EchartsBubble.tsx
```diff
@@ -20,13 +20,14 @@ import { BubbleChartTransformedProps } from './types';
...
 export default function EchartsBubble(props: BubbleChartTransformedProps) {
-  const { height, width, echartOptions, refs } = props;
+  const { height, width, echartOptions, refs, formData } = props;
   return (
     <Echart
       height={height}
       width={width}
       echartOptions={echartOptions}
       refs={refs}
+      vizType={formData.vizType}
```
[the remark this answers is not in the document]
mistercrunch: Wrong, shenanigans switch the case upstream, we should be able to trust tsc here too

msyavuz (approved):
LGTM!

PR #34892, opened 2025-08-28
chore(react18): Migrate legacy react methods

justinpark:
### SUMMARY
This commit migrates legacy lifecycle methods to their newer versions as part of the upgrade process to React 18.

N/A

### TESTING INSTRUCTIONS
CI and specs

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/.eslintrc.js
+++ b/superset-frontend/.eslintrc.js
@@ -413,13 +413,6 @@ module.exports = {
     'icons/no-fa-icons-usage': 'error',
     'i18n-strings/no-template-vars': ['error', true],
     'i18n-strings/sentence-case-buttons': 'error',
-    camelcase: [
-      'error',
-      {
-        allow: ['^UNSAFE_'],
-        properties: 'never',
-      },
-    ],
     'class-methods-use-this': 0,
     curly: 2,
     'func-names': 0,
--- a/superset-frontend/src/components/Datasource/components/CollectionTable/index.tsx
+++ b/superset-frontend/src/components/Datasource/components/CollectionTable/index.tsx
@@ -93,6 +93,7 @@ export default class CRUDCollection extends PureComponent<
       collectionArray,
       sortColumn: '',
       sort: 0,
+      prevCollection: props.collection,
     };
     this.onAddItem = this.onAddItem.bind(this);
     this.renderExpandableSection = this.renderExpandableSection.bind(this);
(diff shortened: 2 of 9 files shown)


michael-s-molina (commented):
## Potential Improvements
### A. getDerivedStateFromProps Usage
getDerivedStateFromProps must return either a new state object or null. The code mostly does this; however, there are cases where it always returns an object, even if nothing changed (e.g., { prevCollection: nextProps.collection }).
Best Practice: Only return the minimal state update if something actually changed.
Impact: This can cause unnecessary re-renders but is technically correct. It’s a minor efficiency concern.

### B. State Naming
Adding keys like prevCollection, prevLastUpdated, etc., to state is a common workaround. However, this can make the state object larger and more complex.
Best Practice: If the component only needs to react to changes (not keep a historical value), consider comparing props directly in componentDidUpdate.
Impact: Not a bug, but can increase cognitive load for maintainers.

### C. Possible Missed Edge Cases
In some components (e.g., AdhocFilterControl), derived state is always returned, even when not necessary. This may cause extra renders.
In WithPopoverMenu, the derived state always returns { needsClickListner: undefined } if neither condition matches, which may also cause extra renders.

justinpark:
> ## Potential Improvements
> ### A. getDerivedStateFromProps Usage
> getDerivedStateFromProps must return either a new state object or null. The code mostly does this; however, there are cases where it always returns an object, even if nothing changed (e.g., { prevCollection: nextProps.collection }). Best Practice: Only return the minimal state update if something actually changed. Impact: This can cause unnecessary re-renders but is technically correct. It’s a minor efficiency concern.
> 
> ### B. State Naming
> Adding keys like prevCollection, prevLastUpdated, etc., to state is a common workaround. However, this can make the state object larger and more complex. Best Practice: If the component only needs to react to changes (not keep a historical value), consider comparing props directly in componentDidUpdate. Impact: Not a bug, but can increase cognitive load for maintainers.
> 
> ### C. Possible Missed Edge Cases
> In some components (e.g., AdhocFilterControl), derived state is always returned, even when not necessary. This may cause extra renders. In WithPopoverMenu, the derived state always returns { needsClickListner: undefined } if neither condition matches, which may also cause extra renders.

Sounds good. Minimized the getDerievedStateFromProps and use componentDidUpdate instead.

michael-s-molina (approved):

PR #34895, opened 2025-08-28
fix: Athena quoting

### LINKED ISSUE #34856: Superset 6.0.0rc1: charts cannot be displayed with Pinot DB (incorrect SQL dialect used)
### Bug description

We've been testing the 6.0.0rc1 release and have run into a problem. Our prior setup (Superset 5.0.0 and Pinot 1.0.0) operated as expected, but after the update, we are experiencing an error on all charts (see screenshot below).
While investigating, we came across ticket https://github.com/apache/superset/issues/34700. Ironically, this ticket is supposed to correct the exact problem we are now facing. This bug did not affect us in version 5.0.0, but has been introduced with the 6.0.0rc1 release, despite the inclusion of the supposed fix.

### Screenshots/recordings

Image

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

_No response_

### Checklist

betodealmeida:
### SUMMARY

https://github.com/apache/superset/pull/34180 fixed how single quotes are escaped in Databricks by monkeypatching the dialect:

```python
DatabricksDialect.colspecs[types.String] = DatabricksStringType
```

Unfortunately, it seems like `colspecs` is an attribute of the base class, and `DatabricksDialect` does not override it. Which means that:

```python
>>> AthenaPandasDialect.colspecs is DatabricksDialect.colspecs
True
```

So as a side-effect, the same escaping was being applied to Athena (and potentially [Pinot](https://github.com/apache/superset/issues/34856) and other dialects).

This PR fixes the problem with:

```python
DatabricksDialect.colspecs = DatabricksDialect.colspecs.copy()
DatabricksDialect.colspecs[types.String] = DatabricksStringType
```

### TESTING INSTRUCTIONS

Tested with Athena, with the fix it generates the correct SQL.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/databricks.py
+++ b/superset/db_engine_specs/databricks.py
@@ -81,6 +81,9 @@ def monkeypatch_dialect() -> None:
     try:
         from sqlalchemy_databricks._dialect import DatabricksDialect
 
+        # copy because the dictionary is shared with other subclasses if it hasn't been
+        # overwritten
+        DatabricksDialect.colspecs = DatabricksDialect.colspecs.copy()
         DatabricksDialect.colspecs[types.String] = DatabricksStringType
     except ImportError:
         pass


Vitor-Avila:
@betodealmeida the fix actually broke Databricks, unfortunately. 

Before this PR:
image

With this commit:
image

It does fix the issue for Athena tho!

betodealmeida:
@Vitor-Avila thanks, let me take a look why it fails on Databricks...

betodealmeida:
That's weird, it works with this script:

```python
#!/usr/bin/env python
"""Test how Databricks monkeypatch affects other dialects"""

from typing import Any, Callable

from databricks.sql.utils import ParamEscaper
from pyathena.sqlalchemy_athena import AthenaPandasDialect
from sqlalchemy import literal_column, types
from sqlalchemy_databricks._dialect import DatabricksDialect


class DatabricksStringType(types.TypeDecorator):
    impl = types.String
    cache_ok = True
    pe = ParamEscaper()

    def process_literal_param(self, value: Any, dialect: Any) -> str:
        return self.pe.escape_string(value)

    def literal_processor(self, dialect: Any) -> Callable[[Any], str]:
        def process(value: Any) -> str:
            _step1 = self.process_literal_param(value, dialect="databricks")
            if dialect.identifier_preparer._double_percents:
                _step2 = _step1.replace("%", "%%")
            else:
                _step2 = _step1

            return "%s" % _step2

        return process


# monkeypatch
DatabricksDialect.colspecs = DatabricksDialect.colspecs.copy()
DatabricksDialect.colspecs[types.String] = DatabricksStringType

# Test after monkeypatch
query = literal_column("name").in_(["Anna's Decorations"])
compiled = query.compile(
    dialect=AthenaPandasDialect(), compile_kwargs={"literal_binds": True}
)
print(f"  PyAthena result after: {compiled}")
compiled = query.compile(
    dialect=DatabricksDialect(), compile_kwargs={"literal_binds": True}
)
print(f"  Databricks result after: {compiled}")
```

```bash
% python test_monkeypatch_effect.py
  PyAthena result after: name IN ('Anna''s Decorations')
  Databricks result after: name IN ('Anna\'s Decorations')
```

I need to do more investigation.

rusackas (approved):

PR #34900, opened 2025-08-29
fix: Filter bar orientation submenu should not be highlighted

kgabryje:
### SUMMARY
In dashboard filters config menu, the "Orientation of filter bar" submenu items was always highlighted. That's the default antd behavior when any of the submenu's children are selected.
This PR adds some styles to set the text color to the same color as other menu items

Before:

image

After:

image

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterBarSettings/index.tsx
+++ b/superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterBarSettings/index.tsx
@@ -190,6 +190,7 @@ const FilterBarSettings = () => {
       items.push({
         key: 'placement',
         label: t('Orientation of filter bar'),
+        className: 'filter-bar-orientation-submenu',
         children: [
           {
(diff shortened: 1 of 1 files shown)


superset-frontend/src/dashboard/components/nativeFilters/FilterBar/FilterBarSettings/index.tsx
```diff
@@ -200,9 +201,6 @@ const FilterBarSettings = () => {
                   FilterBarOrientation.Vertical && (
                   <Icons.CheckOutlined
                     iconColor={theme.colorPrimary}
-                    css={css`
-                      vertical-align: -${theme.sizeUnit * 0.03125}em;
```
kgabryje, line 204: this style is already a default for antd icons so it's not necessary here

Antonio-RiveroMartnez (approved):

PR #34908, opened 2025-08-29
docs(alerts & reports): update how to get headless browser

sfirke:
### SUMMARY
- People should use Playwright + Chrome by default, we already have that install documented 
- Don't advise to run the -dev image as it's no longer bundled with everything needed
- Update docker-build docs to reflect that we are now past 5.0.0 (so always installing with `uv`)


sfirke:
I added the historical note in the 2nd commit as we have so much old discussion floating around that it might help people understand when they're seeing something stale.

rusackas (approved):

PR #34935, opened 2025-09-01
fix(menu): Use app prefix in Menu brand icon url

martyngigg:
### SUMMARY

Since the theming changes, if an `app_root` is set to a non-empty value, the brand logo failed to render as the `src` link was incorrect. I think this is a bug against 6.0?

These changes include the app_root prefix similarly to other places across the frontend.

### TESTING INSTRUCTIONS

Add `SUPERSET_APP_ROOT="/workspace/example"` to `docker/.env-local`.

Spin up the `docker-compose-light` stack and use the web developer tools to inspect the logo image `src` attribute. Before the change the asset link was missing the "/workspace/example" prefix but it is included after this change.

In development, using the webpack server, the image actually still appears as the image is in the `superset/static/assets/images` directory and the webpack server can still find it. I'm not sure how this can be fixed...

### ADDITIONAL INFORMATION

- [x] Has associated issue: #35027
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/home/Menu.tsx
+++ b/superset-frontend/src/features/home/Menu.tsx
@@ -19,6 +19,7 @@
 import { useState, useEffect } from 'react';
 import { styled, css, useTheme } from '@superset-ui/core';
 import { debounce } from 'lodash';
+import { assetUrl } from 'src/utils/assetUrl';
 import { getUrlParam } from 'src/utils/urlUtils';
 import { MainNav, MenuMode } from '@superset-ui/core/components/Menu';
 import { Tooltip, Grid, Row, Col, Image } from '@superset-ui/core/components';
@@ -285,7 +286,7 @@ export function Menu({
         >
           <Image
             preview={false}
-            src={theme.brandLogoUrl}
+            src={assetUrl(theme.brandLogoUrl)}
             alt={theme.brandLogoAlt || 'Apache Superset'}
           />
         </Typography.Link>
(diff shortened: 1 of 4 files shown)


superset-frontend/src/features/home/Menu.tsx
```diff
@@ -285,7 +286,7 @@ export function Menu({
         >
           <Image
             preview={false}
-            src={theme.brandLogoUrl}
+            src={assetUrl(theme.brandLogoUrl)}
```
msyavuz: Won't this cause issues when the brandLogoUrl is hosted outside Superset?

martyngigg: Maybe I have misunderstood how this now works with the new theming?

When `assetUrl` was originally implemented in #30134 it was assumed that [`STATIC_ASSETS_PREFIX`](https://github.com/apache/superset/blob/0fce5ecfa5295f9978f873f2930ca972b868cfe5/superset/config.py#L1940) would be a prefix to all static assets and if using a CDN for example then this would be set as the `STATIC_ASSETS_PREFIX` and this would then become a prefix to everything.

Similarly we have [`ensureAppRoot`](https://github.com/apache/superset/blob/0fce5ecfa5295f9978f873f2930ca972b868cfe5/superset-frontend/src/features/home/RightMenu.tsx#L319) for prefixing non-asset paths.

I think I saw that the default `theme.brandLogoUrl` value was a relative path much like `superset_config[APP_ICON]` so I assumed the same rules applied. Maybe `assetUrl()` could also check the value passed to it is a relative path and only add the prefix then?

msyavuz: > Maybe assetUrl() could also check the value passed to it is a relative path and only add the prefix then?

Yes, that would be my solution as well.

martyngigg: I've gone with adding a second `assetUrlf` that checks for the relative path as in the other places it is used the check is not necessary so it avoids wasting time in those places.

martyngigg:
Closing in favour of #36058.

PR #27409, opened 2024-03-06
fix(excel/csv filename on chart download): downloads zip file and excel/csv file with chart name.

nandwalritik:
fix(excel/csv filename on chart download): downloads zip file and excel/csv file with chart name.
### SUMMARY

When we download any chart as excel/csv file, the file doesn't have appropriate name. So for filename its like
*  if the report is paginated it downloads zip folder, whose name is just timestamp for that we can make it like reportName_timstamp.zip.
     * Inside that zip comes 2 files, one which contains the data that can be named as reportname_timestamp.csv and other one reportname_rowcount_timestamp.csv
* In case report is not paginated we can directly name the file as reportname_timestamp.csv.
### BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
Before
image
After
image
image

### TESTING INSTRUCTIONS

* Clone the repo, checkout to this branch.
* Run superset > navigate to any dashboard > click on three dots > download excel / download csv > View the filename
* The filename should be same as chart name.
### ADDITIONAL INFORMATION

- [x] Has associated issue: #21859 
- [ ] Required feature flags:
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

Hi @sfirke as we discussed over [slack](https://apache-superset.slack.com/archives/C0170U650CQ/p1708620763042799?thread_ts=1708406517.357759&cid=C0170U650CQ),I have made the changes and tested manually on my local.
* For formatting I ran below command
``` 
tox -e pre-commit
tox -e pylint
```
* But while running test cases using pytest as below
```pytest  tests/integration_tests/charts/data/api_tests.py```
I am getting below error and unable to find root cause for it
```ERROR tests/integration_tests/charts/data/api_tests.py::test_chart_data_subquery_allowed[200-extras3] - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: dbs```

PR #34988, opened 2025-09-02
feat: Use dashboard name for screenshot download

Vitor-Avila:
### SUMMARY
The client-side generated dashboard PDF/PNG download would use the dashboard title as the filename (with the timestamp). The backend generation would always produce a file named `screenshot.${format}` which is not ideal.

This PR improves the backend generation to use the dashboard title.

image

### TESTING INSTRUCTIONS
Integration test added. For manual testing:
1. Make sure you have caching, celery and Selenium/Playwright running.
2. Enable all required FFs to generate screenshots via the backend.
3. Trigger a PDF/PNG download of a dashboard, and confirm the suggested file name.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/dashboard/hooks/useDownloadScreenshot.ts
```diff
@@ -98,12 +98,25 @@ export const useDownloadScreenshot = (
...
+            if (disposition && disposition.includes('filename=')) {
+              const filenameRegex = /filename="([^"]+)"/;
+              const matches = filenameRegex.exec(disposition);
+              if (matches && matches[1]) {
+                filename = matches[1];
+              }
+            }
+
+            return response.blob().then(blob => ({ blob, filename }));
+          })
```
betodealmeida, line 121: It looks like we have `content-disposition` as a dependency already, so you can use it here:

```typescript
// import contentDisposition from 'content-disposition';
.then((response: Response) => {
  let filename = `screenshot.${format}`; // default filename
  const disposition = response.headers.get('Content-Disposition');
  
  if (disposition) {
    try {
      const parsed = contentDisposition.parse(disposition);
      filename = parsed.parameters.filename || filename;
    } catch (error) {
      // Fallback to default filename if parsing fails
      console.warn('Failed to parse Content-Disposition header:', error);
    }
  }

  return response.blob().then(blob => ({ blob, filename }));
})
```

sfirke:
Just FYI on a related effort that stalled out - I'm making the link in case folks land on that other PR tracking progress: https://github.com/apache/superset/pull/27409

superset-frontend/src/dashboard/hooks/useDownloadScreenshot.ts
Vitor-Avila: Great suggestion! Let me work on this

betodealmeida (approved):

superset/dashboards/api.py
```diff
@@ -1204,6 +1205,10 @@ def screenshot(self, pk: int, digest: str) -> WerkzeugResponse:
                 image = cache_payload.get_image()
             except ScreenshotImageNotAvailableException:
                 return self.response_404()
+
+            filename = get_filename(
+                dashboard.dashboard_title, dashboard.id, skip_id=True
```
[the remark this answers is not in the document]
Vitor-Avila: ok, I manually tested this and it turns out that this is correct. Technically you can create a dashboard with a `POST api/v1/dashboard` with payload set to `{}` and that would break `get_filename()`

PR #34991, opened 2025-09-02
fix(databricks): string escaper v2

Vitor-Avila:
### SUMMARY
This is a follow up to https://github.com/apache/superset/pull/34895. That approach didn't work for Databricks because the query execution never accesses `DatabricksDialect.colspecs` directly (it goes through the parent class). This PR implements a context-aware approach by overriding the parent specifically for the `DatabricksDialect`.

**Before**
image

**After**
image

Confirmed that Athena still works after this:
image

### TESTING INSTRUCTIONS
1. Create a Databricks connection.
2. Create a dataset from a table that has values with `'` (like `O'hara`).
3. Filter the column by this value.
4. Make sure that this same flow also works for Athena

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/databricks.py
+++ b/superset/db_engine_specs/databricks.py
@@ -25,6 +25,7 @@ from flask_babel import gettext as __
 from marshmallow import fields, Schema
 from marshmallow.validate import Range
 from sqlalchemy import types
+from sqlalchemy.engine.default import DefaultDialect
 from sqlalchemy.engine.reflection import Inspector
 from sqlalchemy.engine.url import URL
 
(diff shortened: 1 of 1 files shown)


superset/db_engine_specs/databricks.py
```diff
@@ -72,19 +72,42 @@ def process(value: Any) -> str:
...
+        from sqlalchemy.engine.default import DefaultDialect
+
+        class ContextAwareStringType(types.TypeDecorator):
+            impl = types.String
+            cache_ok = True
+
+            def literal_processor(
+                self, dialect: DefaultDialect
+            ) -> Callable[[Any], str]:
+                if dialect.__class__.__name__ == "DatabricksDialect":
```
betodealmeida, line 96: Small suggestion, leave the import from line 82 and change this to:

```suggestion
                if dialect.__class__ is DatabricksDialect:
```

superset/db_engine_specs/databricks.py
```diff
@@ -72,19 +72,42 @@ def process(value: Any) -> str:
...
+    fixed version requires SQLAlchemy>=2.0, which is not yet compatible with Superset.
+
+    The challenge: DatabricksDialect inherits from HiveDialect and accesses the
+    `colspecs` type mapping through its parent class rather than its own. This means:
+    - Direct assignment (`DatabricksDialect.colspecs[types.String] = ...`) works because
+    DatabricksDialect.colspecs points to the same dict object as HiveDialect.colspecs,
+    but then it affects all other dialects that inherit from HiveDialect, causing
+    incorrect escaping for them.
+    - Shallow copying then modifying `DatabricksDialect.colspecs` doesn't work
+    because it's never accessed.
```
betodealmeida: This seems more appropriate in the PR summary than as a comment in the code, since it mentions the previous attempt at fixing it. I'd rephrase it to something like:

> Since the `DatabricksDialect.colspecs` points to the base class (`HiveDialect.colspecs`) we can't patch it without affecting other Hive-based dialects. The solution is to introduce a dialect-aware string type so that the change applies only to Databricks.

superset/db_engine_specs/databricks.py
```diff
@@ -72,19 +72,42 @@ def process(value: Any) -> str:
...
+
+    The solution: A context-aware approach by creating a custom String type
+    that checks which dialect is being used during SQL compilation, allowing
+    to apply Databricks-specific escaping only for Databricks connections.
```
betodealmeida: And I think we can remove this.

betodealmeida (approved):

superset/db_engine_specs/databricks.py
Vitor-Avila: Tried that (also with `isclass()` and `issubclass()`) but couldn't make it work. Will keep as is for now

superset/db_engine_specs/databricks.py
Vitor-Avila: Thanks! done

superset/db_engine_specs/databricks.py
Vitor-Avila: Done too!

PR #34999, opened 2025-09-03
fix: Bump pandas to 2.1.4 for python 3.12

### LINKED ISSUE #34910: Bump pandas to 2.2.* to support Python 3.12
### Bug description

Building with Python 3.12, pandas is not supported at 2.0.3 and is built from source. Support for Python 3.12 was added at version 2.2.*

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

I don't know

### Node version

16

### Browser

Chrome

### Additional context

Check out the duration of Python unit tests for python 3.12 vs 3.11 etc https://github.com/apache/superset/actions/runs/17336877127/job/49224547091

### Checklist

rad-pat:
### SUMMARY
Pandas at 2.0.3 does not support Python 3.12 and it is currently building pandas from source in that case. Bumps to a supported version that has been tested and does not include breaking changes.
N.B. Cannot go >=2.2.* until on SQLAlchemy 2.0

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #34910
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ dependencies = [
     "packaging",
     # --------------------------
     # pandas and related (wanting pandas[performance] without numba as it's 100+MB and not needed)
-    "pandas[excel]>=2.0.3, <2.1",
+    "pandas[excel]>=2.0.3, <2.2",
     "bottleneck", # recommended performance dependency for pandas, see https://pandas.pydata.org/docs/getting_started/install.html#performance-dependencies-recommended
     # --------------------------
     "parsedatetime",
--- a/superset/commands/database/uploaders/excel_reader.py
+++ b/superset/commands/database/uploaders/excel_reader.py
@@ -72,7 +72,7 @@ class ExcelReader(BaseDataReader):
             "na_values": self._options.get("null_values")
             if self._options.get("null_values")  # None if an empty list
             else None,
-            "parse_dates": self._options.get("column_dates"),
+            "parse_dates": self._options.get("column_dates") or False,
             "skiprows": self._options.get("skip_rows", 0),
             "sheet_name": self._options.get("sheet_name", 0),
             "nrows": self._options.get("rows_to_read"),
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -158,6 +158,7 @@ greenlet==3.1.1
     # via
     #   apache-superset (pyproject.toml)
     #   shillelagh
+    #   sqlalchemy
 gunicorn==23.0.0
     # via apache-superset (pyproject.toml)
 h11==0.16.0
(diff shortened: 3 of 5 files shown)


rad-pat:
Oh wow, this is a bust - I'll see what I can do to rectify, marking draft for now.

pyproject.toml
```diff
@@ -76,7 +76,7 @@ dependencies = [
     "packaging",
     # --------------------------
     # pandas and related (wanting pandas[performance] without numba as it's 100+MB and not needed)
-    "pandas[excel]>=2.0.3, <2.1",
+    "pandas[excel]>=2.1.4, <2.2",
```
sadpandajoe: Any reason to not just bump just the upper range? Since the lower bound is now higher than our old upper bound.

rad-pat: Nope, I'll change it. Just ended up here after starting with 2.2.3 and working backwards 🤦

rad-pat:
It looks like to fix the presto test, it requires an update to the presto dialect to implement the `get_view_names` method.

rad-pat:
> It looks like to fix the presto test, it requires an update to the presto dialect to implement the `get_view_names` method.

So it looks like the lack of `get_view_names` is worked around in the db engine spec for presto, but with the update to pandas, pandas is trying to inspect the view names and the implementation is not available in the dialect directly. I will try to open a PR in that repo to fix it. https://github.com/apache/superset/blob/8406a827dd927a6230c90ae17f8c56d0cb9a7048/superset/db_engine_specs/presto.py#L1009

rad-pat:
Ref - https://github.com/apache/kyuubi/pull/7190

superset/commands/database/uploaders/excel_reader.py
```diff
@@ -72,7 +72,7 @@ def file_to_dataframe(self, file: FileStorage) -> pd.DataFrame:
             "na_values": self._options.get("null_values")
             if self._options.get("null_values")  # None if an empty list
             else None,
-            "parse_dates": self._options.get("column_dates"),
+            "parse_dates": self._options.get("column_dates") or False,
```
sadpandajoe, line 75: Is this backwards compatible? I'm guessing this was done for the upper end of pandas updates

rad-pat: This was done because sending None was doing something incorrectly in later Pandas. I can't remember exactly what now, but I traced it through to a change in Pandas. False is the default param value and has been for as far back as I can see in Pandas docs, so probably should be what is sent in the case no "column_dates" are provided. Hmmm, maybe should be `self._options.get("column_dates") if self._options.get("column_dates") is not None else False`  What do you think?

sadpandajoe (approved):

PR #35001, opened 2025-09-03
fix(dashboard): normalize spacings and background colors

gabotorresruiz:
### SUMMARY

Normalized dashboard layout spacings and improved background colors consistency. Key changes include:

- **Simplified margin logic**: Changed dashboard content margin calculation from complex conditional logic to a simpler editMode-based approach
- **Consistent background colors**: Added proper background colors (`colorBgContainer` for filters panel, `colorBgLayout` for main content)
- **Fixed tab navigation padding**: Made tab navigation padding responsive to dashboard filter state
- **Unified grid container margins**: Simplified grid container margins from separate top/right/bottom/left values to a unified margin approach

### BEFORE/AFTER SCREENSHOTS

Before:
Screenshot 2025-09-03 at 13 29 22
Screenshot 2025-09-03 at 13 28 35

After:
Screenshot 2025-09-03 at 14 01 35
Screenshot 2025-09-03 at 13 21 56

### TESTING INSTRUCTIONS

1. Open any dashboard in Superset
2. Verify consistent background colors across the dashboard interface
3. Toggle between edit mode and view mode - observe proper spacing adjustments
4. Open/close the native filters panel - verify tab navigation padding adjusts correctly
5. Test with different filter bar orientations (horizontal/vertical)
6. Ensure no visual regressions in grid layout and component spacing

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI


superset-frontend/src/dashboard/components/DashboardBuilder/DashboardBuilder.tsx
```diff
@@ -286,14 +288,16 @@ const StyledDashboardContent = styled.div<{
       padding-left: 0;
     }
 
+    .grid-container .ant-tabs-nav {
+      padding-left: 0 !important;
```
rusackas: Is there a better selector so we can avoid `!important`?

superset-frontend/src/dashboard/components/DashboardBuilder/DashboardWrapper.tsx
```diff
@@ -84,6 +84,12 @@ const StyledDiv = styled.div`
       background: transparent;
     }
 
+    .dashboard-component-tabs .ant-tabs-nav {
+      padding-left: ${dashboardFiltersOpen
+        ? '0 !important'
```
rusackas: Same here, would love to avoid `!important` wherever possible.

gabotorresruiz:
Hey @rusackas, as you mentioned above, I kept looking at the code and found a more elegant way to handle this without using `!important`. I also took the opportunity to refactor the `gridComponents` folder into a structure that should be clearer for anyone who needs to make changes in the future. This makes the PR a bit more challenging to review, but I believe the effort is worthwhile.

I added some tests to make sure that I didn't break anything important. Also we can CC @kasiazjc and @sadpandajoe, just to make sure that everything looks and works as expected.

rusackas:
I think we can merge this, since it's already a large PR handling a lot of different concerns (filter bar toggling, gutter adjustment, file re-org, etc.). I think @mistercrunch was also wondering if/how we should handle gutters for charts within tabs (or charts within tabs within tabs). Do you think that's something we should discuss/address here, or just merge all this and then tackle another design/implementation question around gutters within tabs?

mistercrunch (approved):
I agree with @rusackas , let's merge this as a step in the right direction, addressing important issues and increasing test coverage. Still some thinking needed to make dashboard cards look the same in-and-out of tabs

gabotorresruiz:
@rusackas @mistercrunch totally agree, that’s why when we talked earlier today I called this like a “phase 1”. Still a few things to do here for sure, but I believe this is a nice way to start

PR #35007, opened 2025-09-04
fix(tests): one of integration test in TestSqlaTableModel  does not support MySQL "concat" 

catpineapple:
### SUMMARY

The test: tests/integration_tests/model_tests.py::TestSqlaTableModel::test_query_with_expr_groupby_timeseries uses MySQL as the database backend (SQLALCHEMY_EXAMPLES_URI). However, I noticed that the SQL sent to MySQL does not conform to expectations. It contains the PG syntax "select state || gender || '_test'" (or "or" in MySQL). This behavior is intended to implement string concatenation, but it is not logically valid for MySQL. Although this case passes, its logic is flawed.

The mysql || (or "or") operation on multiple strings evaluates to 0. This 0 is used as the result of a subquery in the join SQL statement, resulting in a query result. However, this test only checks whether a result was obtained, not whether the result is correct, so it passes.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/tests/integration_tests/model_tests.py
+++ b/tests/integration_tests/model_tests.py
@@ -450,12 +450,11 @@ class TestSqlaTableModel(SupersetTestCase):
         spec.allows_joins = inner_join
 
         # Use database-specific string concatenation syntax
-        database = self.get_database_by_id(tbl.database_id)
-        if database.backend == "mysql":
-            arbitrary_gby = "CONCAT(state, gender, '_test')"
-        else:
-            # Use PostgreSQL/SQLite style for other databases
-            arbitrary_gby = "state || gender || '_test'"
+        arbitrary_gby = (
+            "CONCAT(state, gender, '_test')"
+            if get_example_database().backend == "mysql"
+            else "state || gender || '_test'"
+        )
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/model_tests.py
```diff
@@ -448,11 +448,15 @@ def query_with_expr_helper(self, is_timeseries, inner_join=True):
...
-            else "state || gender || '_test'"
-        )
+
+        # Use database-specific string concatenation syntax
+        database = self.get_database_by_id(tbl.database_id)
+        if database.backend == "mysql":
+            arbitrary_gby = "CONCAT(state, gender, '_test')"
+        else:
+            # Use PostgreSQL/SQLite style for other databases
+            arbitrary_gby = "state || gender || '_test'"
```
msyavuz: Can't we just do:
```suggestion
        arbitrary_gby = (
            "CONCAT(state, gender, '_test')"
            if get_example_database().backend == "mysql"
            else "state || gender || '_test'"
        )
```

Also great catch! How did you find this out?

msyavuz (approved):
LGTM!

PR #33055, opened 2025-04-09
fix: Bump FAB to 5.X

dpgaspar:
## Summary
This PR upgrades Flask-AppBuilder from 4.5.1 to 5.0.0, unlocking the path for Superset to adopt SQLAlchemy 2.x and Flask 3.x. While this upgrade alone doesn't enable these newer versions (additional Superset changes will be required), Flask-AppBuilder is no longer a blocker for these critical upgrades.

## Key Changes
- Upgraded Flask-AppBuilder to 5.0.0
- Updated dependencies to support FAB 5.x requirements
- Modified code to comply with FAB 5.0.0 breaking changes

## Breaking Changes
Please review the Flask-AppBuilder migration guides:
- [FAB 5.0.0 Migration Guide](https://flask-appbuilder.readthedocs.io/en/latest/versionmigration.html#migrating-to-5-0-0)
- [FAB 5.0.0 PR with detailed changes](https://github.com/dpgaspar/Flask-AppBuilder/pull/2241)

## Performance Impact
Performance testing shows a slight improvement of approximately **~2%** in API response times:
- Tested with 200 requests each to `/api/v1/chart`, `/api/v1/dataset`, and `/api/v1/dashboard`
- Overall average response time improved from 80.78ms to 78.61ms

## Future Work
This PR sets the foundation for:
- **SQLAlchemy 2.x support**: I will submit follow-up PRs to enable SQLAlchemy 2.x compatibility
- **Flask 3.x support**: Additional PRs will address Flask 3.x compatibility requirements

Note: While this PR makes FAB 5.0.0 compatible, actual migration to SQLAlchemy 2.x and Flask 3.x will require additional changes to Superset's codebase, which will be addressed in separate PRs.

## Testing Instructions
- All existing tests pass
- API performance validated
- No regressions in core functionality

## UPDATING.md
- [ ] Add entry documenting this upgrade for users

---

This upgrade is a critical step in modernizing Superset's dependency stack and ensuring long-term maintainability with current Python ecosystem standards.

PR #35044, opened 2025-09-07
docs: fixed typo in sso docs, fixed code style in sso docs

winterka:
### SUMMARY
Just changed misleading part of SSO documentation (<myClientId>) and formatted example code in the same doc to PEP8

### TESTING INSTRUCTIONS
Check docs diff :)


winterka:
Could anyone review this one please ?

rusackas:
I don't know a whole lot about keycloak, so I'm just running CI and pinging a couple additional reviewers here :D

Vitor-Avila:
I'm not familiar with SSO integrations directly with Superset, so can't help much here, unfortunately. Perhaps @dpgaspar has any thoughts?

rusackas:
Thanks for the contribution! I think even if this isn't perfect, it's probably good, so long as we add some "this is just an example" kind of disclaimer - i.e. "We offer this as a starting point, and provide no warranties. Your deployment may have different requirements or security concerns to contend with."

That said, I ran this through AI, and got some feedback. Some of it sounds pretty critical, but I'd love your feedback on whether or not we should take these issues into consideration.

- Syntax errors in the example superset_config snippet
  - You used colons instead of equals for Python assignments (e.g. SECRET_KEY: '...' and OIDC_OPENID_REALM: '...'). Change these to =.
- Incorrect / fragile calls to AppBuilder / Flask-OIDC APIs
  - In OIDCSecurityManager.init you do OpenIDConnect(self.appbuilder.get_app). Usually you must call the getter: self.appbuilder.get_app() (or pass the Flask app instance). Verify AppBuilder API in your Superset/FAB version.
  - Check calls like self.appbuilder.get_url_for_index and self.appbuilder.get_url_for_login — verify whether these are properties or methods in your FAB version (you may need parentheses).
- Inconsistency in role names
  - In the login handler you create a user with sm.find_role("Gamma") but in the config you set AUTH_USER_REGISTRATION_ROLE = 'Public'. Make these consistent or explain why they differ.
- OIDC client_secret.json structure and Flask-OIDC expectations
  - The example client_secret.json structure in the doc may not match what Flask-OIDC expects by default (Flask-OIDC/OAuth2 libraries commonly use a "web" or "client" key layout). Add a note to verify the file format against Flask-OIDC docs and Keycloak client export format.
- Logout flow and redirect
  - The logout implementation is fragile: calling oidc.logout(), calling super().logout(), then building a redirect using oidc.client_secrets.get("issuer") may not work reliably. client_secrets shape and key lookup may differ; also oidc.client_secrets may not exist in the form you expect. Test and pin the exact behavior against the Flask-OIDC release you intend to support.
- Security hardening & production guidance
  - OIDC_ID_TOKEN_COOKIE_SECURE = False is unsafe for production; set True if using HTTPS in production (and document that).
  - Don’t hard-code SECRET_KEY or client secrets in repo — recommend environment variables or a vault, and warn not to commit client_secret.json to source control.
  - Recommend setting appropriate cookie, session, and token lifetimes and validating issuer/audience claims.
- Package maintenance / recommendation
  - Flask-OIDC has not been actively maintained compared to newer libraries (e.g. Authlib). Consider recommending Authlib or documenting that Flask-OIDC may be legacy, and note which versions of Flask and Flask-OIDC you tested against. You already mention 2.2.0 which is good; also call out Flask version compatibility.
- User identity mapping & uniqueness
  - You use email (or preferred_username) to create users. Document assumptions about which attribute you use as the unique identifier and what to do if Keycloak doesn't provide email or if emails aren't unique.
- Example readability and usability
  - Small improvements: use consistent quoting, show exact file placement (e.g. path of keycloak_security_manager.py relative to superset_config.py), show how to enable/disable self-registration clearly, and provide an example Keycloak client configuration (redirect URIs, access type, service accounts, etc.).

sfirke:
One other thing @winterka - the general guidance directs users to the Flask-AppBuilder security config docs. I see there's an example of Keycloak there: https://flask-appbuilder.readthedocs.io/en/latest/security.html#authentication-oauth

Did you try that and did it not work? Just making sure we've covered all of our bases. Ideally we first update the Flask-AppBuilder docs as the source of record, then add anything that is beyond that / Superset-specific to this repo.

rusackas:
Converting to draft for now — please bring this back to ready once the outstanding items above are addressed. If there is no movement in a future triage sweep this may get closed; reopen as ready any time.

rusackas:
@winterka thanks - this is more than a typo fix, it's a handy Keycloak/Flask-OIDC example :) One thing: it's landing in a versioned-docs file. Can this go in the current docs so it propagates in future versions? A quick rebase too and we can take another look. Holler if you want help.

rusackas:
@winterka just checking in. Did you / will you get a chance to move this out of the versioned-docs snapshot into current docs and rebase? Would love to land the Keycloak example.

docs/versioned_docs/version-6.0.0/configuration/configuring-superset.mdx
```diff
@@ -363,6 +363,121 @@ CUSTOM_SECURITY_MANAGER = CustomSsoSecurityManager
...
+
+Then add to your `superset_config.py` file:
+
+```python
+from keycloak_security_manager import OIDCSecurityManager
+from flask_appbuilder.security.manager import AUTH_OID, AUTH_REMOTE_USER, AUTH_DB, AUTH_LDAP, AUTH_OAUTH
+import os
+
+AUTH_TYPE = AUTH_OID
+SECRET_KEY: 'SomethingNotEntirelySecret'
```
sadpandajoe, line 448: These configuration examples use annotations instead of assignments (`SECRET_KEY: ...`, and likewise the two OIDC settings below), so copying them leaves all three values unset. Could these use `=` so the documented configuration actually takes effect?

rusackas:
This has been quiet since October despite a couple of check-ins from me, Mind if I take it forward as a fresh PR that moves the example into the current docs and clears up the syntax issues?

winterka:
@rusackas Hi. I'd move it into current docs in 1-2 days. Did not got a chance to hop into PR for a long time.

rusackas:
Thanks for the update. No rush, just flag me here once it's moved into current docs and rebased and I'll re-review

rusackas:
Curious if you're still interested in revisiting this, @winterka. Would love to see it happen.

rusackas:
Since you're (hopefully?) planning to move this over, worth flagging a couple standing items. Bito's right that the config examples use `:` instead of `=`, which won't actually work as Python. I'm also not totally sure about a few of the FAB/Flask-OIDC calls, the role names, or the shape of `client_secret.json`. Still don't know Keycloak well enough to say for certain, so if anyone more familiar with that stack wants to weigh in, that'd help. Appreciate you sticking with it if you can, but holler if that's changed.

rusackas:
@winterka thanks for the patience on this one, and sorry it dragged. Going back to the original commit, the change itself was tiny and right: the `client_id` placeholder was wrong and the example wanted a black-style cleanup. 

The catch is that the whole Keycloak / Flask-OIDC section got removed from the docs when we bumped to Flask-AppBuilder 5 (#33055), since that Flask-OpenID based path no longer works there, so the text this fixes isn't in the tree anymore and the merge from `master` is now re-adding the deleted section wholesale. Closing this out rather than course-correcting it. If you're running Keycloak on a current release and have a working Authlib-based setup, a short doc PR for that would be very welcome.

PR #35062, opened 2025-09-08
feat(embedded): Change function signature of setupExtensions

msyavuz:
### SUMMARY

We have a function called setupExtensions which is used for code overrides. This pr renames it to setupCodeOverrides and add options to enable it in embedded context. Other options can be added in the future to enable extensions for different contexts

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -23,7 +23,7 @@ This file documents any backwards-incompatible changes in Superset and
 assists people when migrating to a new version.
 
 ## Next
--  [35062](https://github.com/apache/superset/pull/34871): Changed the function signature of `setupExtensions` to `setupCodeOverrides` with options as arguments.
+-  [35062](https://github.com/apache/superset/pull/35062): Changed the function signature of `setupExtensions` to `setupCodeOverrides` with options as arguments.
 - [34871](https://github.com/apache/superset/pull/34871): Fixed Jest test hanging issue from Ant Design v5 upgrade. MessageChannel is now mocked in test environment to prevent rc-overflow from causing Jest to hang. Test environment only - no production impact.
 - [34782](https://github.com/apache/superset/pull/34782): Dataset exports now include the dataset ID in their file name (similar to charts and dashboards). If managing assets as code, make sure to rename existing dataset YAMLs to include the ID (and avoid duplicated files).
 - [34536](https://github.com/apache/superset/pull/34536): The `ENVIRONMENT_TAG_CONFIG` color values have changed to support only Ant Design semantic colors. Update your `superset_config.py`:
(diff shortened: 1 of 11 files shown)


michael-s-molina (changes requested):
@msyavuz I believe this is a breaking change given that you changed the hook API and would require a major version. If we're ok with including this as part of 6.0 breaking changes, I would ask you to rename the function to `setupCodeOverrides` to not conflict with the new extensions feature. 

We also need to include instructions in `UPDATING.md` about the API change.

cc @sadpandajoe @villebro

msyavuz:
@michael-s-molina ~I don't think this is a breaking change though? The modifications here are entirely optional. Only thing changing here is that extensions now apply to embedded as well. Not sure how this relates to the extension work you guys are doing but happy to sync up if this breaks anything there~ API change comes from overriding this function not using it. I changed the function signature and included the change in the `UPDATING.md`. Not sure if we can include this in 6.0 now @sadpandajoe or is it too late for that?

UPDATING.md
```diff
@@ -23,6 +23,7 @@ This file documents any backwards-incompatible changes in Superset and
 assists people when migrating to a new version.
 
 ## Next
+-  35062    : Change the function signature of `setupExtensions` to `setupCodeOverrides` with options as arguments.
```
michael-s-molina:
```suggestion
-  [35062](https://github.com/apache/superset/pull/34871): Changed the function signature of `setupExtensions` to `setupCodeOverrides` with options as arguments.
```

superset-frontend/src/setup/setupCodeOverrides.ts
```diff
@@ -17,5 +17,20 @@
  * under the License.
  */
 
-// For individual deployments to add custom overrides
-export default function setupExtensions() {}
+/**
+ * Configuration options for setting up code overrides in different contexts.
+ */
+interface setupCodeOverridesOptions {
```
michael-s-molina:
```suggestion
interface CodeOverrideOptions {
```

superset-frontend/src/setup/setupCodeOverrides.ts
```diff
@@ -17,5 +17,20 @@
...
+  embedded?: boolean;
+}
+
+/**
+ * Hook for individual deployments to add custom overrides
+ * @param options - Configuration options for the setup process
+ */
+export default function setupCodeOverrides(
+  options: setupCodeOverridesOptions = {},
+) {}
```
michael-s-molina:
```suggestion
export default function setupCodeOverrides(
  options: CodeOverrideOptions = {},
) {}
```

michael-s-molina (approved):
@msyavuz Thank you for the changes. I'm approving the PR but let's wait for @sadpandajoe approval as well, as the release manager for 6.0, before merging the PR.

sadpandajoe (approved):

PR #35066, opened 2025-09-09
chore: standardize webpack dev server host configuration and enable LAN access

mistercrunch:
Standardize webpack dev server host configuration and enable LAN access

    - Adds configurable devserverHost parameter with CLI and env var support
    - Supports WEBPACK_DEVSERVER_HOST and WEBPACK_DEVSERVER_PORT environment variables
    - Maintains CLI argument precedence: --devserverHost > WEBPACK_DEVSERVER_HOST > 127.0.0.1
    - Updates allowedHosts to support .local domains for LAN development
    - Removes docker-compose localhost binding restriction for external access
    - Follows 2025 framework standards: secure by default, easy opt-in for network access

### DIFF
--- a/docker-compose-light.yml
+++ b/docker-compose-light.yml
@@ -162,8 +162,11 @@ services:
       SCARF_ANALYTICS: "${SCARF_ANALYTICS:-}"
       # configuring the dev-server to use the host.docker.internal to connect to the backend
       superset: "http://superset-light:8088"
+      # Webpack dev server configuration
+      WEBPACK_DEVSERVER_HOST: "${WEBPACK_DEVSERVER_HOST:-127.0.0.1}"
+      WEBPACK_DEVSERVER_PORT: "${WEBPACK_DEVSERVER_PORT:-9000}"
     ports:
-      - "127.0.0.1:${NODE_PORT:-9001}:9000"  # Parameterized port
+      - "${NODE_PORT:-9001}:9000"  # Parameterized port, accessible on all interfaces
     command: ["/app/docker/docker-frontend.sh"]
     env_file:
       - path: docker/.env # default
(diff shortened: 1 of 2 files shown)


docker-compose-light.yml
```diff
@@ -162,8 +162,11 @@ services:
       SCARF_ANALYTICS: "${SCARF_ANALYTICS:-}"
       # configuring the dev-server to use the host.docker.internal to connect to the backend
       superset: "http://superset-light:8088"
+      # Webpack dev server configuration
+      WEBPACK_DEVSERVER_HOST: "${WEBPACK_DEVSERVER_HOST:-127.0.0.1}"
```
rusackas, line 166: I wonder how well this works with IPV6? Does it work for you with `::1`? Since we're coming into the modern age and all, anyway.

mistercrunch: ✅ IPv6 Support Already Included:

  1. allowedHosts: We already include '::1' (IPv6 localhost) in the allowed hosts list
  2. Host binding: When devserverHost is set to '0.0.0.0', webpack-dev-server typically binds to both IPv4 and IPv6
  by default
  3. Docker: Docker containers support IPv6 networking

  Potential Considerations:

  The '0.0.0.0' binding should work for IPv6, but if someone specifically wants IPv6-only binding, they could set
  WEBPACK_DEVSERVER_HOST='::' (IPv6 equivalent of 0.0.0.0).

  We could enhance the allowedHosts to be more IPv6-friendly by adding IPv6 local network patterns, but the current
  setup with ::1 should handle most cases.

  Should work fine with IPv6! The ::1 in allowedHosts covers IPv6 localhost, and 0.0.0.0 binding typically handles
  both IPv4 and IPv6 traffic in modern Node.js/webpack setups.

  Is the reviewer asking about any specific IPv6 scenario?

aminghadersohi (approved):

eschutho (approved):

PR #34917, opened 2025-08-31
feat: Add orval.dev integration POC with TanStack Query 

mistercrunch:
This POC demonstrates complete orval + TanStack Query integration for Superset, replacing manual SupersetClient calls with auto-generated, type-safe hooks.

### Orval + TanStack Query Foundation
- Complete orval setup with TanStack Query hook generation
- Charts API client with 7 hooks (useGetCharts, useCreateChart, etc.)
- 14 TypeScript interfaces with full type safety
- Custom mutator handling URL prefixes and CSRF token integration
- Clean API exports via @superset-ui/core/api import path

### Real Production Code Migrations
- ChartList: Migrated from useListViewResource to useGetCharts hook
- PropertiesModal: Migrated from SupersetClient.put to useUpdateChart mutation
- Demonstrates actual transformation from manual → auto-generated patterns

### Technical Architecture
- Build tooling at superset-frontend level (orval config, npm scripts)
- Generated code in packages/superset-ui-core/src/orval-generated-code/
- Crystal-clear organization (generated code gets bulldozed, custom code preserved)
- TanStack Query dependency added for modern React patterns
- Sectional generation approach for immediate value delivery

## Generated Features
- Automatic caching and background refresh
- Optimistic updates for mutations
- Type-safe query parameters and responses
- Smart cache invalidation patterns
- Reduced boilerplate (15+ lines of state management → 1 hook call)

## Strategic Value
This POC proves orval can transform Superset's frontend development experience while providing a clear path for gradual adoption without disrupting existing code. Foundation established for expanding to all 176 Superset API endpoints.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #35077, opened 2025-09-10
feat(sqllab extension): core query interface api

justinpark:
### SUMMARY
This commit implements the following SQL Lab extension apis.

- onDidQueryRun
- onDidQueryStop
- onDidQueryFail
- onDidQuerySuccess

### TESTING INSTRUCTIONS
The event has been triggered during the query execution

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/SqlLab/components/QueryAutoRefresh/index.tsx
```diff
@@ -119,7 +119,14 @@ function QueryAutoRefresh({
                   !failedQueries.current.has(id) &&
                   state === QueryState.Failed
                 ) {
-                  dispatch(logFailedQuery(query, query.extra?.errors));
+                  dispatch(
+                    queryFailed(
```
justinpark, line 123: queryFailed action includes `logFailedQuery`

superset-frontend/src/core/sqlLab.ts
```diff
@@ -66,34 +213,168 @@ const predicate = (actionType: string): AnyListenerPredicate<RootState> => {
...
   );
 
 export const onDidQueryFail: typeof sqlLabType.onDidQueryFail = (
-  listener: (e: string) => void,
+  listener: (query: sqlLabType.QueryErrorResultContext) => void,
   thisArgs?: any,
 ): Disposable =>
   createActionListener(
-    predicate(QUERY_FAILED),
+    action => action.type === QUERY_FAILED,
```
michael-s-molina: Why not use `predicate` which also checks for the active editor?

superset-frontend/packages/superset-core/src/api/core.ts
```diff
@@ -76,6 +76,35 @@ export declare interface Database {
...
 }
 
+// Keep in sync with superset/errors.py
+export type ErrorLevel = 'info' | 'warning' | 'error';
+
+/**
+ * Superset error object structure.
+ * Contains details about an error that occurred within Superset.
+ */
+export type SupersetError<ExtraType = Record<string, any> | null> = {
```
michael-s-molina:
```suggestion
export type SupersetError = {
```

superset-frontend/packages/superset-core/src/api/core.ts
```diff
@@ -76,6 +76,35 @@ export declare interface Database {
...
+  /**
+   * Error types, see enum of SupersetErrorType in superset/errors.py
+   */
+  error_type: string;
+
+  /**
+   * Extra properties based on the error types
+   */
+  extra: ExtraType;
```
michael-s-molina:
```suggestion
  extra?: Record<string, any>;
```

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
   panels: Panel[];
 }
 
+/**
+ * Generic data types, see enum of the same name in superset/utils/core.py.
+ */
+enum GenericDataType {
```
michael-s-molina: `GenericDataType` should be moved to `api/core` as it's not an exclusive concept of SQL Lab.

I'll open a follow-up PR to replace all references to `GenericDataType` from `@superset-ui/core`.

superset-frontend/packages/superset-core/src/api/core.ts
```diff
@@ -76,6 +76,35 @@ export declare interface Database {
   schemas: Schema[];
 }
 
+// Keep in sync with superset/errors.py
+export type ErrorLevel = 'info' | 'warning' | 'error';
```
michael-s-molina, line 100: @villebro An example of why a fully typed REST API client for Superset is necessary. Ideally, we would have a common definition like Thrift and export both TS and Python classes.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+  Numeric = 0,
+  String = 1,
+  Temporal = 2,
+  Boolean = 3,
+}
+
+/**
+ * Column metadata returned in query results.
+ */
+export type QueryColumn = {
```
michael-s-molina: Can we use the `Column` definition in `api/core`? It looks like all properties would be valid.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+
+  /**
+   * Contains CTAS if the query requests table creation
+   */
+  ctas: CTAS | null;
+
+  /**
+   * The SQL editor instance associated with the query
+   */
+  editor: Editor;
```
michael-s-molina: The `editor` is already available in `tab`.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+
+  /**
+   * The tab instance associated with the request query
+   */
+  tab: Tab;
+
+  /**
+   * A key-value JSON formatted string associated with Jinja template variables
+   */
+  templateParams: string;
```
michael-s-molina: Shouldn't `templateParams` be a `Record<str, any>`? The serialization problem needs to be handled elsewhere. Extensions and built-in features might want to interact with `templateParams`.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+   */
+  method: CTASMethod;
+
+  /**
+   * Temporary table name for creation using a CTAS query
+   */
+  tempTable: string | null;
+}
+
+export interface QueryRequestContext {
```
michael-s-molina: Given that the properties defined here are used both when requesting a query but also for query results and errors, I think we should remove the `Request` word from the name and just call it `QueryContext`.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+
+  /**
+   * Executed sql after parsing the jinja templates
+   */
+  executedSql: string;
+
+  /**
+   * Actual number of rows returned by the query
+   */
+  limit: number;
```
michael-s-molina: This name needs to be more distinguishable given that there's a `queryLimit` property inherited from `QueryContext`. How about changing both properties to `requestedLimit` and `appliedLimit`? We would also need to rename `limitingFactor`.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+
+  /**
+   * Major factor that is determining the row limit of the query results
+   */
+  limitingFactor: string;
+
+  /**
+   * Remote query id stored in backend
+   */
+  queryId: number;
```
michael-s-molina: `QueryContext` has an `id` property which is a client-side ID. Maybe we should rename both to `clientId` and `remoteId` to make things clearer.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+   */
+  errorMessage: string;
+
+  /**
+   * Error details in a SupersetError structure
+   */
+  errors?: SupersetError[];
+
+  /**
+   * Executed sql after parsing the jinja templates
```
michael-s-molina:
```suggestion
   * Executed SQL after parsing Jinja templates.
```

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+}
+
+export interface QueryResultContext extends QueryRequestContext {
+  /**
+   * Finished datetime for the query in a numerical timestamp
+   */
+  endDttm: number;
+
+  /**
+   * Executed sql after parsing the jinja templates
```
michael-s-molina:
```suggestion
   * Executed SQL after parsing Jinja templates.
```

superset-frontend/packages/superset-core/src/api/sqlLab.ts
```diff
@@ -111,6 +111,171 @@ export interface Tab {
...
+export type QueryColumn = {
+  /**
+   * Label of the column
+   */
+  name?: string;
+
+  /**
+   * Column name defined
+   */
+  column_name: string;
```
michael-s-molina: If we had something like Thrift configured to generate both Typescript and Python classes, we would configure it so that Typescript classes would follow the camelCase convention and Python classes would follow the snake_case convention. Something like:

```
// Define in Thrift with consistent naming
struct QueryRequest {
  1: i32 query_limit
  2: string template_params
}

// Configure generators:
// - TypeScript: --gen js:ts,snake_case_to_camel_case
// - Python: --gen py (keeps snake_case naturally)
```

Given that we're defining Typescript APIs, let's make them all camelCase and handle the conversion manually for now. We can introduce an automatic conversion layer later.

michael-s-molina (commented):
Great work @justinpark! I left some first-pass comments.

superset-frontend/packages/superset-core/src/api/core.ts
villebro: Yeah, now/soon may be the time to start considering something like this, as keeping these types synced between FE/BE is pretty error prone.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
villebro: I suspect we'll soonish want to start replacing the JSON-based data blob with a more performant binary format that has built-in data types. I'm thinking we would want to default to using the binary transport protocol, but let charts/extensions opt-in for the legacy JSON format, in which case they get the generic data types. Just something to keep in mind for the migration of `GenericDataType` to core..

superset-frontend/packages/superset-core/src/api/sqlLab.ts
villebro: Automatic translation between snake/camel case can be tricky as has been seen in the current charting implementation, in which there's `formData` with converted camel to snake and `rawFormData` with the unconverted properties. This has lead to inconsistent use of `formData` and `rawFormData`, with the risk of collisions. If we want to pass along SQLA model instances via the REST API, then it feels awkward to convert from the native format (snake_case) to the local format (camelCase) just for the sake of aesthetics. So let's make sure we think hard about language specific conventions vs having a single case definition throughout the platform.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
villebro: We actually already have a `QueryContext` class in the backend that's used by the chart data API. So let's make sure we disambiguate those.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
villebro: I agree, while these are stored in string format in the SQLA model, these should be deserialized by the backend when returning the result. However, since this in the current implementation this is indeed a string, we will need to wait for a breaking window to change this. Maybe in the interim introducing `templateParameters: Record<string, any>` that is deserialized and deprecating `templateParams`?

superset-frontend/src/core/sqlLab.ts
justinpark: The query can be executed asynchronously, so it may fail while you have switched to another tab.
You can manually check if the failed query belongs to the current tab by verifying like request.tab.id === activeTabId.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
villebro: Same here - changing this would be a breaking change.

villebro (commented):
First pass comments. It seems it's going to be tricky to move over elements from `@superset-ui` and the host to `apache-superset/core` while simultaneously cleaning them up. Maybe a topic for the next planning meeting, i.e. how to balance cleanup vs velocity vs maintaining SemVer outside breaking windows..

superset-frontend/packages/superset-core/src/api/sqlLab.ts
michael-s-molina: @villebro I don't mean to change the backend side of things. Just the frontend part which is not a breaking change as we're still defining the API.

michael-s-molina: In other words, the `queryId` returned by the backend is translated to `remoteId`.

superset-frontend/src/core/sqlLab.ts
michael-s-molina: You're totally correct. We need a better way to manage this.

superset-frontend/src/core/sqlLab.ts
```diff
@@ -16,19 +16,175 @@
...
+    this.runAsync = runAsync;
+    this.startDttm = startDttm;
+    this.requestedLimit = options.requestedLimit ?? null;
+    this.ctas = options.tempTable
+      ? new CTAS(options.ctasMethod === CTASMethod.View, options.tempTable)
+      : null;
+    this.templateParams = options.templateParams ?? '';
+  }
+
+  get templateParameters() {
```
michael-s-molina, line 96: It would be good to add a comment explaining that we want to do the parsing only when the property is accessed. That's why we didn't add to the constructor.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
michael-s-molina: @villebro I think this concept is defined on the backend but we don't use in the frontend. I'll add a topic for our meeting to discuss this as well.

superset-frontend/packages/superset-core/src/api/sqlLab.ts
michael-s-molina: @villebro Added a topic for our next meeting. For now, we'll keep snake case.

michael-s-molina (approved):
Thank you so much for the great work and welcome to the Extensions project! We're lucky to have you!

superset-frontend/packages/superset-core/src/api/core.ts
msyavuz: There's also [this POC](https://github.com/apache/superset/pull/34917) for typed api client which has the idea of generating the client from openapi spec

villebro (approved):
LGTM! With these changes I think it highlights the need to refactor the core package. For instance, breaking it up into sub-packages that are then exported on the top `index.ts`, similar to how `@superset-ui/core` does it. Let's maybe discuss that too in the next meeting..

michael-s-molina:
> Let's maybe discuss that too in the next meeting..

@villebro I'll add a topic about the package structure.

PR #35105, opened 2025-09-11
fix: SQL Lab tab events

michael-s-molina:
### SUMMARY
This adds a unique `immutableId` field to the `QueryEditor` type that persists throughout the tab's lifecycle, enabling reliable event filtering and tab identification for SQL Lab extensions while maintaining backward compatibility. Previously, SQL Lab tabs could only be identified by their mutable id field, which made it difficult for extensions to reliably track events and maintain stable references to specific editor instances across state changes and persistence operations.

The addition of `immutableId` for tabs might help improving the current logic that uses mutable `id` and `tabViewId` introduced in https://github.com/apache/superset/pull/34720 to reduce re-renders. I'll let @justinpark analyze the code and check if things can be improved in a follow-up PR.

### TESTING INSTRUCTIONS
Check that the events are triggered for the correct tabs.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/SqlLab/fixtures.ts
+++ b/superset-frontend/src/SqlLab/fixtures.ts
@@ -188,6 +188,7 @@ export const table = {
 export const defaultQueryEditor = {
   version: LatestQueryEditorVersion,
   id: 'dfsadfs',
+  immutableId: 'dfsadfs',
   autorun: false,
   dbId: 1,
   latestQueryId: null,
@@ -204,6 +205,7 @@ export const defaultQueryEditor = {
 export const extraQueryEditor1 = {
   ...defaultQueryEditor,
   id: 'diekd23',
+  immutableId: 'diekd23',
   sql: 'SELECT *\nFROM\nWHERE\nLIMIT',
   name: 'Untitled Query 2',
   selectedText: 'SELECT',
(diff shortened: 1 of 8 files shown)


superset-frontend/src/SqlLab/fixtures.ts
```diff
@@ -188,6 +188,7 @@ export const table = {
 export const defaultQueryEditor = {
   version: LatestQueryEditorVersion,
   id: 'dfsadfs',
+  immutableId: 'dfsadfs',
```
villebro: Can we make these different from `id` to make sure these don't get mixed up?

villebro (commented):
LGTM with a minor test fixture comment. I'm curious, do you know wht the ids are mutable? I would have thought the main id would stay unchanged throughout the tab's lifecycle.

superset-frontend/src/SqlLab/fixtures.ts
michael-s-molina: Done.

michael-s-molina:
> I'm curious, do you know why the ids are mutable?

I think it's because they are updated when the tab is saved on the server. @justinpark will probably know better but I'm hoping that we can improve the logic now that an immutable id is present and maybe keep the remote id in a separate variable.

villebro (approved):
Approving but hoping for remaining fixtures to also be updated!

Issue #34529, opened 2025-08-04
Embedded Dashboard continues to call fetchGuestToken after unmount() in Angular applications


NewtonFernandis:
### Bug description

The embedded dashboard continues to make `fetchGuestToken` calls even after the component containing it has been destroyed and `unmount()` has been called. This leads to unnecessary network requests and potential memory leaks in Single Page Applications (SPAs).

@superset-ui/embedded-sdk version: 0.1.3
Framework: Angular 16
Browser: 136.0.7103.114

## Steps to Reproduce

1. Create an Angular component that embeds a Superset dashboard using embedDashboard()
2. Configure the dashboard with fetchGuestToken callback
3. Navigate to the component and let the dashboard load successfully
4. Navigate away from the component (triggering ngOnDestroy)
5. Call unmount() on the dashboard instance in the cleanup method
6. Wait approximately 5 minutes
7. Observe network requests in browser DevTools

## Expected Behavior
After calling `unmount()` on the dashboard instance, no further `fetchGuestToken` calls should be made, and all internal timers/intervals should be cleared.

## Actual Behavior
The fetchGuestToken callback continues to be invoked approximately 5 minutes after navigation, even though:

- The component has been destroyed
- unmount() was called on the dashboard instance
- The DOM container has been cleared

## Code Example
```js
export class AnalyticsComponent implements OnInit, OnDestroy {
  private currentDashboard: EmbeddedDashboard | null = null;

  async ngOnInit() {
    this.currentDashboard = await embedDashboard({
      id: 'dashboard-id',
      supersetDomain: 'https://superset.example.com',
      mountPoint: this.containerElement,
      fetchGuestToken: () => this.fetchGuestToken(), // This continues to be called!
      dashboardUiConfig: {
        hideTitle: true,
        filters: { expanded: false }
      }
    });
  }

  ngOnDestroy() {
    this.cleanup();
  }

  private cleanup() {
    if (this.currentDashboard) {
      try {
        this.currentDashboard.unmount(); // Called but doesn't stop token fetching
        this.currentDashboard = null;
      } catch (error) {
        console.error('Error unmounting dashboard:', error);
      }
    }
    
    // Clear container
    if (this.containerElement) {
      this.containerElement.innerHTML = '';
    }
  }

  private async fetchGuestToken(): Promise<string> {
    console.log('Token fetch called - this should not happen after unmount!');
    // ... token fetching logic
  }
}
```
## Potential Root Cause
[...]

PR #35143, opened 2025-09-15
fix(embedded-sdk): clear fetchGuestToken timers on unmount

NewtonFernandis:
### SUMMARY
Fixes memory leak in @superset-ui/embedded-sdk where fetchGuestToken refresh timers continue running after unmount().

N/A

### TESTING INSTRUCTIONS
1. Create a test application that uses embedDashboard()
2. Call unmount() after embedding
3. Verify no setTimeout timers continue running
4. Check browser dev tools for continued network requests (should be none)

### ADDITIONAL INFORMATION

- [x] Has associated issue: #34529 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-embedded-sdk/src/index.ts
+++ b/superset-embedded-sdk/src/index.ts
@@ -254,7 +254,7 @@ export async function embedDashboard({
       log("error refreshing guest token:", error);
       // Still schedule retry if not unmounted
       if (!isUnmounted) {
-        refreshTimerId = setTimeout(refreshGuestToken, 30000); // retry in 30 seconds
+        refreshTimerId = setTimeout(refreshGuestToken, getGuestTokenRefreshTiming(newGuestToken));
       }
     }
   }
--- a/superset-embedded-sdk/src/index.ts
+++ b/superset-embedded-sdk/src/index.ts
@@ -252,7 +252,7 @@ export async function embedDashboard({
       }
     } catch (error) {
       log("error refreshing guest token:", error);
-      // Still schedule retry if not unmounted (you might want to add exponential backoff here)
+      // Still schedule retry if not unmounted
       if (!isUnmounted) {
         refreshTimerId = setTimeout(refreshGuestToken, 30000); // retry in 30 seconds
       }
--- a/superset-embedded-sdk/src/index.ts
+++ b/superset-embedded-sdk/src/index.ts
@@ -107,6 +107,9 @@ export async function embedDashboard({
   iframeSandboxExtras = [],
   referrerPolicy,
 }: EmbedDashboardParams): Promise<EmbeddedDashboard> {
+  let refreshTimerId: NodeJS.Timeout | null = null;
+  let isUnmounted = false;
+  
   function log(...info: unknown[]) {
     if (debug) {
       console.debug(`[superset-embedded-sdk][dashboard ${id}]`, ...info);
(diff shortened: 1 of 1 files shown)


superset-embedded-sdk/src/index.ts
```diff
@@ -107,6 +107,9 @@ export async function embedDashboard({
   iframeSandboxExtras = [],
   referrerPolicy,
 }: EmbedDashboardParams): Promise<EmbeddedDashboard> {
+  let refreshTimerId: NodeJS.Timeout | null = null;
+  let isUnmounted = false;
```
msyavuz, line 111: This unmounted and following checks seem unnneccessary to me. When the component using this gets unmounted this already won't run again. Am i missing something?

superset-embedded-sdk/src/index.ts
```diff
@@ -226,15 +229,51 @@ export async function embedDashboard({
...
+
+      ourPort.emit("guestToken", { guestToken: newGuestToken });
+      
+      // Schedule next refresh only if not unmounted
+      if (!isUnmounted) {
+        refreshTimerId = setTimeout(refreshGuestToken, getGuestTokenRefreshTiming(newGuestToken));
+      }
+    } catch (error) {
+      log("error refreshing guest token:", error);
+      // Still schedule retry if not unmounted (you might want to add exponential backoff here)
```
msyavuz:
```suggestion
      // Still schedule retry if not unmounted
```

superset-embedded-sdk/src/index.ts
```diff
@@ -226,15 +229,51 @@ export async function embedDashboard({
...
+      if (!isUnmounted) {
+        refreshTimerId = setTimeout(refreshGuestToken, 30000); // retry in 30 seconds
```
msyavuz:
```suggestion
        refreshTimerId = setTimeout(refreshGuestToken, getGuestTokenRefreshTiming(newGuestToken));
```

msyavuz (commented):
Thanks for opening the PR!

msyavuz:
I see you have some failing ci. You can use pre-commit to avoid ci failures on commit https://superset.apache.org/docs/contributing/development/#git-hooks

superset-embedded-sdk/src/index.ts
NewtonFernandis: Thanks for the question! You're right to question this, The problem is that setTimeout creates global browser timers that run independently of Angular's component lifecycle. Currently, refreshGuestToken creates a recursive chain of timers:
setTimeout(refreshGuestToken, timing);
// Inside refreshGuestToken, it schedules another:
setTimeout(refreshGuestToken, timing);
// This continues indefinitely...
When the Angular component is destroyed and unmount() is called, these global timers keep running in the background. You can verify this by navigating away from the component and checking the browser's network tab - you'll see fetchGuestToken requests continuing.

NewtonFernandis:
> I see you have some failing ci. You can use pre-commit to avoid ci failures on commit https://superset.apache.org/docs/contributing/development/#git-hooks

Thank you, I will check this ✨️

rusackas:
Would love to see this one through if you can get those precommit fixes taken care of and address the comments. Thanks in advance!

rusackas:
Thanks @NewtonFernandis — the cleanup itself looks right. One snag: in the `catch` you reschedule with `getGuestTokenRefreshTiming(newGuestToken)`, but `newGuestToken` is scoped to the `try`, so it's undefined there. Mind fixing that, running pre-commit, and rebasing? Thanks!

NewtonFernandis:
> Thanks @NewtonFernandis — the cleanup itself looks right. One snag: in the `catch` you reschedule with `getGuestTokenRefreshTiming(newGuestToken)`, but `newGuestToken` is scoped to the `try`, so it's undefined there. Mind fixing that, running pre-commit, and rebasing? Thanks!

Yes will do that ✌️

NewtonFernandis:
> Thanks @NewtonFernandis — the cleanup itself looks right. One snag: in the `catch` you reschedule with `getGuestTokenRefreshTiming(newGuestToken)`, but `newGuestToken` is scoped to the `try`, so it's undefined there. Mind fixing that, running pre-commit, and rebasing? Thanks!


  @rusackas  Before pushing a fix, I went to rebase this branch onto `master` and hit a conflict that turned out to be informative: **master has already landed a more complete version of this work.** Its `refreshGuestToken` tracks the pending timer and cancels it on unmount (the cleanup this branch was for), and its `catch` schedules the retry with a `DEFAULT_TOKEN_REFRESH_RETRY_MS` constant rather than a token-derived delay, so the out-of-scope `newGuestToken` reference doesn't exist there at all. The bug is already gone upstream.

  Master also went a bit further:
  - a `withTimeout` wrapper around `fetchGuestToken`, and
  - iframe teardown if the initial fetch rejects after mount.

  So this branch looks superseded, I couldn't find anything here that master is missing. I'd suggest we close this in favor of what's on `master`. Happy to do a focused diff of the two implementations first if you want to confirm
  nothing's lost.

rusackas:
OK, thanks for looking into that. Closing this out, but appreciative of this and any future contribution!

PR #35151, opened 2025-09-15
fix(viz): resolve dark mode compatibility issues in BigNumber and Heatmap

mistercrunch:
- BigNumber with Trendline: Replace hardcoded supersetTheme.colorBgContainer with 'transparent' in sparkline gradient
- Heatmap: Use live theme.colorText with alpha transparency instead of static supersetTheme colors for emphasis styles

Both visualizations now properly support theme switching without visual artifacts in dark mode.


Antonio-RiveroMartnez (approved):
lgtm

msyavuz (approved):

PR #35161, opened 2025-09-16
fix: Oauth2 connection failures with postional arguments

sadpandajoe:
### SUMMARY

**🐛 Problem**

Snowflake OAuth2 connections were failing with the error:
```
SnowflakeOAuth2Override.get_extra_params() takes 2 positional arguments but 3 were given
```
This error occurred when attempting to connect to Snowflake databases with OAuth2 authentication enabled, preventing users from establishing database connections.

**🔍 Root Cause**

The issue was a method signature mismatch when user agent customization was added:

- Problem: `get_extra_params` methods were declared as `@staticmethod` but called as if they were instance methods
- Call Pattern: `Database.get_extra()` calls `self.db_engine_spec.get_extra_params(self, source)`
- Mismatch: Static methods expect (database, source) but receive (class_reference, database, source)

### TESTING INSTRUCTIONS

Added comprehensive test suite in `tests/unit_tests/models/test_core_database_methods.py`:
  - Integration tests: Verify Database.get_extra() works with all 5 database engine types
  - Contract tests: Ensure Database ↔ EngineSpec interface compatibility
  - Regression test: Specific test for the reported Snowflake OAuth2 scenario
  - Coverage: Tests the actual production call pattern that was broken

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/base.py
+++ b/superset/db_engine_specs/base.py
@@ -2027,9 +2027,9 @@ class BaseEngineSpec:  # pylint: disable=too-many-public-methods
         """
         return None
 
-    @staticmethod
+    @classmethod
     def get_extra_params(
-        database: Database, source: QuerySource | None = None
+        cls, database: Database, source: QuerySource | None = None
     ) -> dict[str, Any]:
         """
         Some databases require adding elements to connection parameters,
--- a/superset/db_engine_specs/databricks.py
+++ b/superset/db_engine_specs/databricks.py
@@ -261,9 +261,9 @@ class DatabricksDynamicBaseEngineSpec(BasicParametersMixin, DatabricksBaseEngine
         "port": "port",
     }
 
-    @staticmethod
+    @classmethod
     def get_extra_params(
-        database: Database, source: QuerySource | None = None
+        cls, database: Database, source: QuerySource | None = None
     ) -> dict[str, Any]:
         """
         Add a user agent to be used in the requests.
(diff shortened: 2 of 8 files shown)


betodealmeida (approved):
I'm not sure I understand the problem here, and I'm not convinced this would fix it. :)

It seems to me the problem is that in Superset `get_extra_params` is a `staticmethod`, but in the Preset-specific DB engine-spec it's a `classmethod`. This shouldn't be a problem:

```python
class Parent:
    @staticmethod
    def method(a: int) -> None:
        print(a)


class Child:
    @classmethod
    def method(cls, a: int) -> None:
        print(a, cls)


Parent.method(42)  # 42
Child.method(42)  # 42 <class '__main__.Child'>
```

But OTOH I don't think the changes in this PR are bad, making `get_extra_params` be a `classmethod` by default gives more flexibility to DB engine specs.

sadpandajoe:
@betodealmeida the problem is that snowflake can't connect for 6.0 right now. When trying to do so, there is an error saying there we should have 2 arguments but were given 3. If this isn't the right solution, i can dive in a bit more

betodealmeida:
> @betodealmeida the problem is that snowflake can't connect for 6.0 right now. When trying to do so, there is an error saying there we should have 2 arguments but were given 3. If this isn't the right solution, i can dive in a bit more

@sadpandajoe I think the problem is that we currently have in Preset this:

```python
    @classmethod
    def get_extra_params(cls, database: "Database") -> dict[str, Any]:
        """
        Add a user agent to be used in the requests.
        """
```

So it's missing the `source` argument; instead it should be:

```python
    @classmethod
    def get_extra_params(cls, database: "Database", source: QuerySource | None = None) -> dict[str, Any]:
        """
        Add a user agent to be used in the requests.
        """
```

sadpandajoe:
Closing in favor of doing this in the Preset repo as it affects preset only

PR #35170, opened 2025-09-17
fix: Bump pandas to 2.3.2

rad-pat:
### SUMMARY
I opened PR to bump pandas to 2.1.4, but couldn't remember the reason why we could not go further, so this is just to test out which tests fail.


rad-pat:
Reason is that SQLAlchemy 2.0 is required for pandas >= 2.2.0

PR #35179, opened 2025-09-17
fix: bug in tooltip timeseries chart in calculated total with annotation layer

### LINKED ISSUE #31735: Annotation layer using formula is included in the sum of the rich tool tip
### Bug description

1. Create a bar chart
2. Add a Annotation Layer using Formula 
3. Ensure that rich tool tip is enabled 
4. When hovering over the bar chart the value from the annotation layer is included in the rich tooltip and is included in the total value. 

### Screenshots/recordings

image

### Superset version

4.1.1

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

Assuming that some users are using the annotation layers to create baselines and/or goals it would be great if either:
1) The annotation layer value was not included in the tooltip, or;
2) Annotation layers has an option to not include the values in the rich tooltip. 

### Checklist

SBIN2010:
This PR fixed #31735 

### SUMMARY

Annotation layers are no longer counted towards total and percentage calculations.

BEFORE
Снимок экрана от 2025-09-18 00-46-24

AFTER
Снимок экрана от 2025-09-18 00-18-47

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -582,9 +582,21 @@ export default function transformProps(
...
+        const filteredForecastValue = forecastValue.filter(
+          (item: any) =>
+            !annotationLayers.some(
+              (annotation: any) => item.seriesName === annotation.name,
+            ),
+        );
         const forecastValues: Record<string, ForecastValue> =
           extractForecastValuesFromTooltipParams(forecastValue, isHorizontal);
 
+        const filteredForecastValues: Record<string, ForecastValue> =
```
mistercrunch, line 600: No biggy but I was surprised to see types re-stated here, looked up the function definition and types are defined there.

mistercrunch (approved):
minor nit, otherwise LGTM

superset-frontend/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
SBIN2010: I agree, any should be replaced with the correct type., 👍

rusackas (approved):

sfirke:
Thank you for fixing this! My users will be excited. And thanks Joe for tagging it into 6.0.

PR #35190, opened 2025-09-18
docs: Add Production Security Hardening Guide

sha174n:
### SUMMARY

This guide provides comprehensive best practices for securing a Superset instance in production, covering session management, SECRET_KEY security, and database connection hardening. Addresses community feedback on session security clarity.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


sfirke:
Thanks @sha174n, I can take a look at this later. At a glance it looks like this was written with the assistance of AI. If so, can you please clarify the extent of that - what role did AI play and in particular, have you verified the correctness and applicability of all of this to Superset?

sha174n:
> Thanks @sha174n, I can take a look at this later. At a glance it looks like this was written with the assistance of AI. If so, can you please clarify the extent of that - what role did AI play and in particular, have you verified the correctness and applicability of all of this to Superset?

Hi @sfirke! I've definitely used AI for formatting and proper use of the English language! Everything else was taken from existing documentation, issues and security reports,  but I'd be very happy if you could provide an extra pair of eyes and identify any issue or recommend any changes you might find needed!

This "production guide" would help address some vulnerability reports we're dealing with that mostly result from misconfigurations, so I feel it's deeply needed. Again, all help is appreciated!

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+### **Critical Prerequisites: HTTPS/TLS Configuration**
+
+Running Superset without HTTPS (TLS) is not secure. Without it, all network traffic—including user credentials, session tokens, and sensitive data—is sent in cleartext and can be easily intercepted.
+
+* **Use a Reverse Proxy:** Your Superset instance should always be deployed behind a reverse proxy (e.g., Nginx, Traefik) or a load balancer (e.g., AWS ALB, Google Cloud Load Balancer) that is configured to handle HTTPS termination.
+* **Enforce Modern TLS:** Configure your proxy to enforce TLS 1.2 or higher with strong, industry-standard cipher suites.
+* **Implement HSTS:** Use the HTTP Strict Transport Security (HSTS) header to ensure browsers only connect to your Superset instance over HTTPS. This can be configured in your reverse proxy or within Superset's Talisman settings.
+
+> #### ⚠️ Warning: Never Run Superset on HTTP in Production
```
sfirke: Feels redundant with the section header.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+>
+> Exposing Superset directly over HTTP is a critical security risk that exposes your entire instance to network interception attacks.
```
sfirke: Feels redundant with the first sentence under the header, the one that ends "... can easily be intercepted". In general, the less often we can repeat things, the shorter and more readable this guide is.

I do really like how you've organized it into individual security topics.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+### **Session Management Security (CRITICAL)**
+
+Properly configuring user sessions is essential to prevent session hijacking and ensure that sessions are terminated correctly.
+
+#### **Use a Server-Side Session Backend (Strongly Recommended for Production)**
+
+The default stateless cookie-based session handling presents challenges for immediate session invalidation upon logout. For all production deployments, we strongly recommend configuring a server-side session backend like Redis, Memcached, or a database. This ensures that session data is stored securely on the server and can be instantly destroyed upon logout, rendering any copied session cookies immediately useless.
+
+**Example `superset_config.py` for Redis:**
```
sfirke, line 48: This is great that we are providing a full config here. I wonder if you can remove what's here https://superset.apache.org/docs/security/#switching-to-server-side-sessions and instead point to this guide.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
+---
+title: Securing Your Superset Installation for Production
```
sfirke, line 2: In general I'd like to avoid duplicating any content on https://superset.apache.org/docs/security/. It makes it harder to keep it maintained. I like how practical and focused this guide is and would suggest keeping it as is, but taking a pass through that Security page to find content that overlaps. Where that exists, you could remove it and replace it with pointers to sections on this page.

I'd also suggest adding a prominent link from the top of that page to this page. If this page is going to be "practical guide to securing Superset", what does that mean for that page? I expect it's "everything else"?

Same goes for this page: https://superset.apache.org/docs/configuration/networking-settings/

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+from datetime import timedelta
+
+# Set a short absolute session timeout
+# The default is 31 days, which is NOT recommended for production.
+PERMANENT_SESSION_LIFETIME = timedelta(hours=8)
+
+# Enforce secure cookie flags to prevent browser-based attacks
+SESSION_COOKIE_SECURE = True      # Transmit cookie only over HTTPS
+SESSION_COOKIE_HTTPONLY = True    # Prevent client-side JS from accessing the cookie
+SESSION_COOKIE_SAMESITE = 'Lax'   # Provide protection against CSRF attacks
```
sfirke, line 91: What do you think of `SESSION_COOKIE_SAMESITE = 'None'` ? This is required to embed an iFrame of Superset and have it work on a Chrome browser, https://stackoverflow.com/a/79754758/4470365

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+            'https://api.mapbox.com',
+            'https://events.mapbox.com',
+            'https://tile.openstreetmap.org',
+            'https://tile.osm.ch',
+        ],
+        'object-src': "'none'",
+        'style-src': ["'self'", "'unsafe-inline'"],
+        'script-src': ["'self'"],
+    },
+    'force_https': False, # HTTPS should be handled by a reverse proxy
```
sfirke:
```suggestion
    'force_https': False, # HTTPS should be handled by a reverse proxy and workers may need the HTTP route for Alerts & Reports
```

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+  * **Use a Dedicated Database User:** The database connection configured in Superset should use a dedicated, limited-privilege database user. This user should only have the minimum required permissions (e.g., `SELECT` on specific schemas) for the data sources it needs to query. It should **not** have `INSERT`, `UPDATE`, `DELETE`, or administrative privileges.
+  * **Restrict Dangerous SQL Functions:** To mitigate potential SQL injection risks, configure the `DISALLOWED_SQL_FUNCTIONS` list in your `superset_config.py`. Be aware that this is a defense-in-depth measure, not a substitute for proper database permissions.
+
+### **Additional Security Layers**
+
+  * **Web Application Firewall (WAF):** Deploying Superset behind a WAF (e.g., Cloudflare, AWS WAF) is strongly recommended. A WAF with a standard ruleset (like the OWASP Core Rule Set) provides a critical layer of defense against common attacks like SQL Injection, XSS, and remote code execution.
+
+### **Monitoring and Logging**
+
+  * **Configure Structured Logging:** Set up a robust logging configuration to capture important security events.
```
sfirke, line 134: This section makes sense to me, but do we have any guidance we could point people toward that shows how to do it? I love that above we are providing specific config settings, this part seems more vague. If we don't, that shouldn't block this PR, though.

I think it would be fine to link to external resources about this kind of log monitoring, since it's hardly a problem unique to Superset.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+  - [ ] Talisman is explicitly enabled and configured with a strict Content Security Policy.
+  - [ ] Database connections use dedicated, limited-privilege accounts.
+  - [ ] Authentication is integrated with an enterprise identity provider (OAuth/LDAP/SAML).
+  - [ ] A Web Application Firewall (WAF) is deployed in front of Superset.
+  - [ ] Logging is configured and logs are shipped to a central monitoring system.
+
+#### **Ongoing Maintenance:**
+
+  - [ ] Regularly apply Superset security updates.
+  - [ ] Rotate the `SUPERSET_SECRET_KEY` periodically (e.g., quarterly) and after any potential security incident.
```
sfirke, line 163: Why should one rotate the secret key regularly?

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+  - [ ] `DEBUG` mode is set to `False`.
+  - [ ] Talisman is explicitly enabled and configured with a strict Content Security Policy.
+  - [ ] Database connections use dedicated, limited-privilege accounts.
+  - [ ] Authentication is integrated with an enterprise identity provider (OAuth/LDAP/SAML).
+  - [ ] A Web Application Firewall (WAF) is deployed in front of Superset.
+  - [ ] Logging is configured and logs are shipped to a central monitoring system.
+
+#### **Ongoing Maintenance:**
+
+  - [ ] Regularly apply Superset security updates.
```
sfirke:
```suggestion
  - [ ] Regularly update to the latest major or minor versions of Superset. Those versions receive up-to-date security patches.
```

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+  - [ ] Rotate the `SUPERSET_SECRET_KEY` periodically (e.g., quarterly) and after any potential security incident.
+  - [ ] Conduct quarterly access reviews for all users.
+  - [ ] Review security monitoring alerts weekly.
```
sfirke: I don't think "security monitoring alerts" exist? I would remove this last bullet. Or if they exist and I don't know about them, explain more.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+If you suspect your `SUPERSET_SECRET_KEY` has been compromised, follow these steps immediately to contain the incident:
+
+1.  **Generate a new `SUPERSET_SECRET_KEY`** using `openssl rand -base64 42`.
+2.  **Update the Key:** Update the environment variable or secrets vault with the new key.
+3.  **Re-encrypt Existing Secrets:** Run the Superset CLI command to re-encrypt all credentials and other secrets stored in the metadata database with the new key.
+    ```bash
+    superset re-encrypt-secrets
+    ```
+4.  **Restart All Services:** Perform a full restart of all Superset web servers and workers to ensure they load the new key.
+5.  **Invalidate All Sessions:** The key rotation will automatically invalidate all existing user sessions, forcing all users to log in again.
```
sfirke: If this happens automatically, I don't think it should be a step.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,205 @@
...
+  - [ ] Rotate the `SUPERSET_SECRET_KEY` periodically (e.g., quarterly) and after any potential security incident.
+  - [ ] Conduct quarterly access reviews for all users.
+  - [ ] Review security monitoring alerts weekly.
+
+### **Appendix B: Emergency Procedure for a Compromised `SECRET_KEY`**
+
+If you suspect your `SUPERSET_SECRET_KEY` has been compromised, follow these steps immediately to contain the incident:
+
+1.  **Generate a new `SUPERSET_SECRET_KEY`** using `openssl rand -base64 42`.
+2.  **Update the Key:** Update the environment variable or secrets vault with the new key.
```
sfirke: I believe this is not enough and could cause a catastrophic problem for an unprepared admin. If I recall correctly, don't you need the prior key in order to rotate to a new key? I might be mistaken. If we have guidance on rotating to a new key (I think we do from after the vulnerability about default secret keys was published), let's link to that from this doc.

sfirke (commented):
I think this is going to be a really useful page, thank you for drafting it! I left some precise comments, some broader ones - let me know what you think.

docs/docs/security/securing_superset.mdx
sfirke: A lot of people faced trouble with this in 2023: https://github.com/apache/superset/discussions/23598 for example

docs/docs/security/securing_superset.mdx
sha174n: Yeah, you're right. That warning section was specifically created to address a report in our security list, but it's definitely overkill, given all that's written above it. I'll remove it.

docs/docs/security/securing_superset.mdx
sha174n: Same, will be removed.

docs/docs/security/securing_superset.mdx
sha174n: Good suggestion! I'll put it in the backlog

docs/docs/security/securing_superset.mdx
sha174n: I agree with your points, I also feel the Security documentation needs some restructuring. @dpgaspar and @rusackas , how do you feel about this?

docs/docs/security/securing_superset.mdx
sha174n: I don't think we should remove the 'Lax' recommendation, but we could definitely add a note about embedded dashboards right after it. Something like this:

```
Note on iFrame Embedding and SESSION_COOKIE_SAMESITE
The recommended default setting 'Lax' provides good CSRF protection for most use cases. However, if you need to embed Superset dashboards into other applications using an <iframe>, you will need to change this setting to 'None'.

SESSION_COOKIE_SAMESITE = 'None'

Setting SameSite to 'None' requires that SESSION_COOKIE_SECURE is also set to True. Be aware that this configuration disables some of the browser's built-in CSRF protections to allow for cross-domain functionality, so it should only be used when iFrame embedding is necessary.
```

docs/docs/security/securing_superset.mdx
sha174n: It's a security best practice (from NIST, for instance), ideally automated, to limit the window of exposure for a key that might have been unknowingly compromised. In the real world, it's a matter of risk appetite. We're recommending the best practice, but we can definitely add some notes about a balanced approach, especially if you can't automate key rotation and/or it's a costly process. Something like what's in OWASP:

```
Key Lifetimes and Rotation
Encryption keys should be changed (or rotated) based on a number of different criteria:

If the previous key is known (or suspected) to have been compromised.
- This could also be caused by a someone who had access to the key leaving the organisation.

After a specified period of time has elapsed (known as the cryptoperiod).
- There are many factors that could affect what an appropriate cryptoperiod is, including the size of the key, the sensitivity of the data, and the threat model of the system. See section 5.3 of [NIST SP 800-57](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r4.pdf) for further guidance.

After the key has been used to encrypt a specific amount of data.
- This would typically be 2^35 bytes (~34GB) for 64-bit keys and 2^68 bytes (~295 exabytes) for 128-bit block size.

If there is a significant change to the security provided by the algorithm (such as a new attack being announced).

Once one of these criteria have been met, a new key should be generated and used for encrypting any new data.
```

docs/docs/security/securing_superset.mdx
sha174n: You're right, good point.

docs/docs/security/securing_superset.mdx
sha174n: That's another best practice: implementing security monitoring by forwarding logs from various sources - not only Superset, but also Identity Providers, WAF, DB Audit, etc - to a centralized SIEM configured to detect suspicious activity in Superset. That's outside the scope of a Superset itself, besides providing logs, but it is strongly recommended as part of a mature security program. The "Monitoring and Logging" topic touches this topic.

docs/docs/security/securing_superset.mdx
sha174n: Nice catch @sfirke! I was (dangerously) missing a step, and instead of trying to re-invent the wheel, I'll link to the current docs.

docs/docs/security/securing_superset.mdx
sha174n: I'm actually planning on writing something specific on incident response for Superset (focusing primarily on detection and initial containment) - I can link it here, once done.

docs/docs/security/securing_superset.mdx
sha174n: This also assumes the last point of the previous topic:
 `Logging is configured and logs are shipped to a central monitoring system.`
 
 Changed the text to clarify the requirements.

sha174n:
Thank you for your help @sfirke! I've implemented the changes.

docs/docs/security/securing_superset.mdx
sfirke: Ah this is great, helps me address the extent to which the advice applies to my own situation 🙏

docs/docs/security/securing_superset.mdx
sfirke: Ah okay, I thought this meant like email bulletins from the Apache Superset project. Explaining that it means monitoring logging alerts you have set up helps.

sfirke:
Thanks @sha174n, this looks great! I might be missing it but it looks like you marked one suggestion resolved but didn't implement it. Instead of `Regularly apply Superset security updates.`, make it
```
Regularly update to the latest major or minor versions of Superset. Those versions receive up-to-date security patches.
```

I think it's better to more specifically state the instruction here.

Otherwise this looks good to me!

sha174n:
> Thanks @sha174n, this looks great! I might be missing it but it looks like you marked one suggestion resolved but didn't implement it. Instead of `Regularly apply Superset security updates.`, make it
> 
> ```
> Regularly update to the latest major or minor versions of Superset. Those versions receive up-to-date security patches.
> ```
> 
> I think it's better to more specifically state the instruction here.
> 
> Otherwise this looks good to me!

I agree, should be fixed now. And thanks!

sfirke (approved):
Looks great! Thanks for incorporating the feedback.

sfirke:
I am good with merging this, it's a really nice addition to the docs. I won't click merge just yet in case @rusackas or @dpgaspar want to weigh in. I think any subsequent steps like reorganizing the main Security docs page could be a separate future addition.

docs/docs/security/securing_superset.mdx
dpgaspar: We can remove CORS, CSRF and talisman ref's from networking to security

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,203 @@
...
+    ```bash
+    # Example using openssl to generate a strong key
+    openssl rand -base64 42
+    ```
+* **Store the Key Securely:** The key must be kept confidential. The recommended approach is to store it as an environment variable or in a secrets management system (e.g., AWS Secrets Manager, HashiCorp Vault). **Do not hardcode the key in `superset_config.py` or commit it to version control.**
+    ```python
+    # In superset_config.py
+    import os
+    SECRET_KEY = os.environ.get('SUPERSET_SECRET_KEY')
+    ```
```
dpgaspar, line 34: There's probably some overlap here also, https://superset.apache.org/docs/configuration/configuring-superset/#specifying-a-secret_key
There's some overlap with the security chapter itself:
https://superset.apache.org/docs/security/#user-sessions

We can improve the origin (if needed) and add a link here?
Or the other way around?

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+SESSION_COOKIE_SAMESITE = 'None'
+
+Setting SameSite to 'None' requires that SESSION_COOKIE_SECURE is also set to True. Be aware that this configuration disables some of the browser's built-in CSRF protections to allow for cross-domain functionality, so it should only be used when iFrame embedding is necessary.
+
+### **Authentication and Authorization**
+
+While Superset's built-in database authentication is convenient, for production it's highly recommended to integrate with an enterprise-grade identity provider (IdP).
+
+  * **Use an Enterprise IdP:** Configure authentication via OAuth, OIDC, SAML, or LDAP to leverage your organization's existing identity management system. This provides benefits like Single Sign-On (SSO), Multi-Factor Authentication (MFA), and centralized user provisioning/deprovisioning.
```
dpgaspar: Don't think SAML or OIDC is possible

docs/docs/security/securing_superset.mdx
rusackas: Agreed on this - this page can link to the other docs, where we mention all the instructions and setting (including key size). DRY principle FTW :)

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+#### **Use a Server-Side Session Backend (Strongly Recommended for Production)**
+
+The default stateless cookie-based session handling presents challenges for immediate session invalidation upon logout. For all production deployments, we strongly recommend configuring a server-side session backend like Redis, Memcached, or a database. This ensures that session data is stored securely on the server and can be instantly destroyed upon logout, rendering any copied session cookies immediately useless.
+
+**Example `superset_config.py` for Redis:**
+
+```python
+# superset_config.py
+from redis import Redis
```
rusackas, line 52: Redis is optional, so we should frame all of this as such.

docs/docs/security/securing_superset.mdx
```diff
@@ -0,0 +1,203 @@
...
+### **Authentication and Authorization**
+
+While Superset's built-in database authentication is convenient, for production it's highly recommended to integrate with an enterprise-grade identity provider (IdP).
+
+  * **Use an Enterprise IdP:** Configure authentication via OAuth, OIDC, SAML, or LDAP to leverage your organization's existing identity management system. This provides benefits like Single Sign-On (SSO), Multi-Factor Authentication (MFA), and centralized user provisioning/deprovisioning.
+  * **Principle of Least Privilege:** Assign users to the most restrictive roles necessary for their jobs. Avoid over-provisioning users with Admin or Alpha roles, and ensure row-level security is applied where appropriate.
+
+### **Content Security Policy (CSP) and Other Headers**
+
+Superset can use Flask-Talisman to set security headers. However, it must be explicitly enabled.
```
rusackas, line 111: There's a link/page we can use about all of this, to keep this file shorter.

rusackas:
Some questions that came up during group review:

1) Are we SURE this is comprehensive? Maybe we should add a disclaimer that this is an evolving set of best practices

2) Deployment setups vary wildly... some of these steps may apply to some deployments and not others, and other deployments may have safety checks that are not included here. 

3) Should this even be a doc this long, or should we take all the sections/details you've written, and add details to the existing sections, and use the "checklist" as the main entry here to link to all of those detailed pages/instructions. 

4) Should we add a step to delete the default admin user?

5) For consideration, we could enhance the installation/start scripts to be a more interactive CLI asking about these options, and warning about more of them at the end with "Warning: X / Y / Z. Proceed? (y/N)"

sha174n:
> Some questions that came up during group review:
> 
> 1. Are we SURE this is comprehensive? Maybe we should add a disclaimer that this is an evolving set of best practices
> 2. Deployment setups vary wildly... some of these steps may apply to some deployments and not others, and other deployments may have safety checks that are not included here.
> 3. Should this even be a doc this long, or should we take all the sections/details you've written, and add details to the existing sections, and use the "checklist" as the main entry here to link to all of those detailed pages/instructions.
> 4. Should we add a step to delete the default admin user?
> 5. For consideration, we could enhance the installation/start scripts to be a more interactive CLI asking about these options, and warning about more of them at the end with "Warning: X / Y / Z. Proceed? (y/N)"

Thanks, @rusackas! This is great feedback, and I appreciate you  taking the time!

To clarify: the goal was to create a centralized, comprehensive, production hardening guide that could serve as a living document for admins, and that could evolve as Superset improves and the threat landscape changes. I see real value in having a single place for this, especially in a checklist-like format.

That being said, you've raised several valid points that will definitely improve it:

Disclaimers (Points 1 & 2): Fully agree. I can add a disclaimer at the top stating that this is an evolving set of best practices and that administrators should adapt the guide to their specific deployment architecture.

Default Admin User (Point 4): That's a good point! I will add a step to the "Authentication and Authorization" section recommending the deletion or disabling of the default admin user after a new administrative account has been configured.

Structure (Point 3): I'm happy to discuss the structure further. My initial thought was that a single, long-form document is easier for an admin to follow from start to finish than navigating between multiple smaller pages, but I'm open to what the consensus is.

Interactive CLI (Point 5): I love this idea! How can we propose this?

I'll make the changes for points 1, 2 and 4. Thanks again for the constructive feedback!

docs/docs/security/securing_superset.mdx
sha174n: I can propose a restructuring of the security docs after this guide. Would that work?

docs/docs/security/securing_superset.mdx
sha174n: Thanks @rusackas. We do mention Redis is one of the options, and the code is the Redis example, but I agree we should make it clearer. I'll change it.

docs/docs/security/securing_superset.mdx
dpgaspar: for sure!

docs/docs/security/securing_superset.mdx
sha174n: Corrected

docs/docs/security/securing_superset.mdx
sha174n: Corrected

dpgaspar (approved):

PR #35211, opened 2025-09-19
fix(table-chart): fix cell bar visibility in dark theme

LevisNgigi:
### SUMMARY
fix(table-chart): fix cell bar visibility in dark theme. Cell bars were not visible in dark theme.

BEFORE
Table_before

AFTER

Table_after

### ADDITIONAL INFORMATION

- [ x] Has associated issue: Fixes #35145
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
@@ -53,6 +53,7 @@ import {
   tn,
   useTheme,
   SupersetTheme,
+  isThemeDark,
 } from '@superset-ui/core';
 import {
   Input,
@@ -170,12 +171,24 @@ function cellOffset({
 function cellBackground({
   value,
   colorPositiveNegative = false,
+  theme,
 }: {
   value: number;
   colorPositiveNegative: boolean;
+  theme: SupersetTheme;
 }) {
-  const r = colorPositiveNegative && value < 0 ? 150 : 0;
-  return `rgba(${r},0,0,0.2)`;
+  if (!colorPositiveNegative) {
+    const color = isThemeDark(theme)
+      ? 'rgba(255,255,255,0.3)'
+      : 'rgba(0,0,0,0.2)';
+    return color;
+  }
+
+  const r = value < 0 ? 150 : 0;
+  const color = isThemeDark(theme)
+    ? `rgba(255,255,255,0.3)`
+    : `rgba(${r},0,0,0.2)`;
+  return color;
 }
 
 function SortIcon<D extends object>({ column }: { column: ColumnInstance<D> }) {
@@ -881,6 +894,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
                 background-color: ${cellBackground({
                   value: value as number,
                   colorPositiveNegative,
+                  theme,
                 })};
               `}
           `;
@@ -1089,6 +1103,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
       totals,
       columnColorFormatters,
       columnOrderToggle,
+      theme,
     ],
   );
 
--- a/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
@@ -184,10 +184,11 @@ function cellBackground({
     return color;
   }
 
-  const r = value < 0 ? 150 : 0;
   const color = isThemeDark(theme)
-    ? `rgba(255,255,255,0.3)`
-    : `rgba(${r},0,0,0.2)`;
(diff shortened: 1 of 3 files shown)


superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -170,12 +171,25 @@ function cellOffset({
...
   colorPositiveNegative = false,
+  theme,
 }: {
   value: number;
   colorPositiveNegative: boolean;
+  theme: SupersetTheme;
 }) {
-  const r = colorPositiveNegative && value < 0 ? 150 : 0;
-  return `rgba(${r},0,0,0.2)`;
+  if (!colorPositiveNegative) {
```
mistercrunch, line 179: We're missing out on writing a nifty utility function here, maybe like `themeAdaptiveAlpha(0.3)` or similar... Not sure what the ideal name for it is but seems it'd be useful/helpful.

Either that or reuse some of these?
Image

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -170,12 +171,25 @@ function cellOffset({
...
+    const color = isThemeDark(theme)
+      ? 'rgba(255,255,255,0.3)'
+      : 'rgba(0,0,0,0.2)';
+    return color;
+  }
+
+  const color = isThemeDark(theme)
+    ? value < 0
+      ? `rgba(255,100,100,0.4)`
```
mistercrunch: no likey the hard-coded red here, would it make sense to use one of those? `colorErrorBg*` is alpha-transparency required here or not

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
amaannawab923: True ... We can use these theming tokens .... only problem was the alpha value is a bit low at anywhere between 0.04 to 0.15 

function cellBackground({
  value,
  colorPositiveNegative = false,
  theme,
}: {
  value: number;
  colorPositiveNegative: boolean;
  theme: SupersetTheme;
}) {
  if (!colorPositiveNegative) {
    const fillColor = theme.colorFillSecondary; 
    const { r, g, b } = tinycolor(fillColor).toRgb();
    return `rgba(${r},${g},${b},0.2)`; 
  }

  if (value < 0) {
    const errorColor = theme.colorError; // Use the main error color (red) instead of bg
    const { r, g, b } = tinycolor(errorColor).toRgb();
    return `rgba(${r},${g},${b},0.2)`;
  }

  const successColor = theme.colorSuccess; // Use proper green success color
  const { r, g, b } = tinycolor(successColor).toRgb();
  return `rgba(${r},${g},${b},0.2)`; 
}

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
amaannawab923: colorErrorBg is a bit too pink we can use , theme.colorError .... the alpha value might be required otherwise the cellbars end up covering the content of the cell

superset-frontend/plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -170,12 +171,25 @@ function cellOffset({
...
 }: {
   value: number;
   colorPositiveNegative: boolean;
+  theme: SupersetTheme;
 }) {
-  const r = colorPositiveNegative && value < 0 ? 150 : 0;
-  return `rgba(${r},0,0,0.2)`;
+  if (!colorPositiveNegative) {
+    const color = isThemeDark(theme)
+      ? 'rgba(255,255,255,0.3)'
```
amaannawab923: Do not check for isThemeDark ... Directly use this 

function cellBackground({
  value,
  colorPositiveNegative = false,
  theme,
}: {
  value: number;
  colorPositiveNegative: boolean;
  theme: SupersetTheme;
}) {
  if (!colorPositiveNegative) {
    const fillColor = theme.colorFillSecondary; 
    const { r, g, b } = tinycolor(fillColor).toRgb();
    return `rgba(${r},${g},${b},0.2)`; 
  }

  if (value < 0) {
    const errorColor = theme.colorError; // Use the main error color (red) instead of bg
    const { r, g, b } = tinycolor(errorColor).toRgb();
    return `rgba(${r},${g},${b},0.2)`; 
  }

  const successColor = theme.colorSuccess; // Use proper green success color
  const { r, g, b } = tinycolor(successColor).toRgb();
  return `rgba(${r},${g},${b},0.2)`;
}

it is better to rely on theming tokens directly

mistercrunch (approved):
LGTM

PR #35215, opened 2025-09-19
fix(PropertiesModal): do not show validation errors while loading

msyavuz:
### SUMMARY

Dashboard properties modal was showing validation errors when in a loading state. This pr fixes that.

Before:

https://github.com/user-attachments/assets/95fec6a6-cb7f-483d-a4b9-4c6565947ec7

After:

https://github.com/user-attachments/assets/6ea06a81-42f3-43d1-afba-267071e82794

### TESTING INSTRUCTIONS

Visual testing

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/components/Modal/StandardModal.tsx
+++ b/superset-frontend/src/components/Modal/StandardModal.tsx
@@ -18,7 +18,7 @@
  */
 import { ReactNode } from 'react';
 import { styled, t } from '@superset-ui/core';
-import { Modal } from '@superset-ui/core/components';
+import { Modal, Loading, Flex } from '@superset-ui/core/components';
 import { ModalTitleWithIcon } from 'src/components/ModalTitleWithIcon';
 
 interface StandardModalProps {
@@ -39,6 +39,7 @@ interface StandardModalProps {
   destroyOnClose?: boolean;
   maskClosable?: boolean;
   wrapProps?: object;
+  contentLoading?: boolean;
 }
 
 // Standard modal widths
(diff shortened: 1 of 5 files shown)


mistercrunch:
Oh thanks for fixing this. Wondering if it would make more sense to add a loading state for the reusable modal instead. I think I shifted things so all properties modal use the same component, so maybe we could handle there instead of handling in the first section/form elements. Modal would need the context though as to whether it's loading, but should be doable no? 

Thinking further, wondering if we pop-fast with a spinner in-modal, or whether we spinner the trigger (trigger becomes a spinner, wait until it's loaded to pop the modal). Fine with either approach, spinner-in-modal is probably easier/best (?)

msyavuz:
I think 'time to be interactive' is too low to warrant a loading spinner in modals, i will play with this some more and try to get rid of the loading state alltogether, if not yes could probably add loading to the underlying component.

mistercrunch:
My main thing would be to handle this through a common solution/UX for all properties modal, so it's DRY and consistent across modals. About loading time, you never really know what could be happening, backend hiccup, super slow connection, ... 

I hate loading-related flickering, let's find a solution that that minimizes it. The loading-state-in-trigger might be decent (antd has loading states for buttons). Otherwise was thinking a spinner inside the full-size modal might be ok.

mistercrunch (approved):
LGTM

PR #35220, opened 2025-09-20
feat(theming): add base theme config

gabotorresruiz:
### SUMMARY

This PR implements a base theme configuration system that provides consistent foundation themes across the application. This change introduces `THEME_DEFAULT` and `THEME_DARK` as complete base themes that serve as the foundation for all theme customization, ensuring consistent theming while maintaining flexibility.

### TESTING INSTRUCTIONS

1. **Test base theme configuration**:
     - Start Superset with the default configuration
     - Verify the base themes are applied correctly (check colorPrimary, borders, etc.)
     - Toggle between light/dark/system modes
     - Confirm mode persistence after page reload

  2. **Test with custom theme tokens**:
     - Modify `THEME_DEFAULT` in `config.py`:
       ```python
       THEME_DEFAULT = {
           "token": {
               "colorPrimary": "#ff0000",
               "borderRadius": 8,
               # ... other custom tokens
           }
       }
       THEME_DARK = {
           **THEME_DEFAULT,
           "algorithm": "dark"
       }
       ```
     - Restart and verify custom tokens are applied

  3. **Test UI theme administration**:
     - Navigate to Settings > Themes
     - Create custom themes and set as system default/dark
     - Verify database themes properly overlay on base configuration
     - Test "Remove as system theme"
     - Restart and verify, should fallback to config themes

  4. **Test confirmation modals theming**:
     - In the Themes list, click "Set as system default theme" on any theme
     - Verify the confirmation modal follows current theme colors
     - Test in both light and dark modes
     - Confirm modal buttons and styling inherit properly

  5. **Test dashboard-specific themes**:
     - Apply a theme to a dashboard using the bolt button
     - Verify dashboard theme properly merges with base theme
     - Navigate between dashboards to confirm proper theme switching

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/components/Typography/Typography.test.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/components/Typography/Typography.test.tsx
@@ -50,7 +50,7 @@ describe('Typography Component', () => {
 
   it('renders strong text', () => {
     render(<Typography.Text strong>Strong Text</Typography.Text>);
-    expect(screen.getByText('Strong Text')).toHaveStyle('font-weight: 500');
+    expect(screen.getByText('Strong Text')).toHaveStyle('font-weight: 600');
   });
 
   it('renders underlined text', () => {
--- a/superset-frontend/packages/superset-ui-core/src/theme/GlobalStyles.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/theme/GlobalStyles.tsx
@@ -16,6 +16,18 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
+// @fontsource/* v5.1+ doesn't play nice with eslint-import plugin v2.31+
+/* eslint-disable import/extensions */
+import '@fontsource/inter/200.css';
+import '@fontsource/inter/400.css';
+import '@fontsource/inter/500.css';
+import '@fontsource/inter/600.css';
+import '@fontsource/fira-code/400.css';
+import '@fontsource/fira-code/500.css';
+import '@fontsource/fira-code/600.css';
+/* eslint-enable import/extensions */
+
 import { css, useTheme, Global } from '@emotion/react';
 
 export const GlobalStyles = () => {
--- a/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
@@ -20,23 +20,6 @@
 // eslint-disable-next-line no-restricted-syntax
 import React from 'react';
 import { theme as antdThemeImport, ConfigProvider } from 'antd';
-
-// @fontsource/* v5.1+ doesn't play nice with eslint-import plugin v2.31+
-/* eslint-disable import/extensions */
-import '@fontsource/inter/200.css';
-/* eslint-disable import/extensions */
-import '@fontsource/inter/400.css';
-/* eslint-disable import/extensions */
-import '@fontsource/inter/500.css';
-/* eslint-disable import/extensions */
-import '@fontsource/inter/600.css';
-/* eslint-disable import/extensions */
-import '@fontsource/fira-code/400.css';
-/* eslint-disable import/extensions */
-import '@fontsource/fira-code/500.css';
-/* eslint-disable import/extensions */
-import '@fontsource/fira-code/600.css';
-
 import {
   ThemeProvider,
   CacheProvider as EmotionCacheProvider,
@@ -44,7 +27,6 @@ import {
 import createCache from '@emotion/cache';
 import { noop } from 'lodash';
 import { GlobalStyles } from './GlobalStyles';
-
 import {
   AntdThemeConfig,
   AnyThemeConfig,
@@ -53,63 +35,16 @@ import {
   allowedAntdTokens,
   SharedAntdTokens,
 } from './types';
-
 import { normalizeThemeConfig, serializeThemeConfig } from './utils';
-
-/* eslint-disable theme-colors/no-literal-colors */
 
 export class Theme {
   theme: SupersetTheme;
-
-  private static readonly defaultTokens = {
-    // Brand
-    brandLogoAlt: 'Apache Superset',
-    brandLogoUrl: '/static/assets/images/superset-logo-horiz.png',
-    brandLogoMargin: '18px',
-    brandLogoHref: '/',
-    brandLogoHeight: '24px',
-
-    // Spinner
-    brandSpinnerUrl: undefined,
-    brandSpinnerSvg: undefined,
-
-    // Default colors
-    colorPrimary: '#2893B3', // NOTE: previous lighter primary color was #20a7c9
-    colorLink: '#2893B3',
-    colorError: '#e04355',
-    colorWarning: '#fcc700',
-    colorSuccess: '#5ac189',
-    colorInfo: '#66bcfe',
-
-    // Forcing some default tokens
-    fontFamily: `'Inter', Helvetica, Arial`,
-    fontFamilyCode: `'Fira Code', 'Courier New', monospace`,
-
-    // Extra tokens
-    transitionTiming: 0.3,
-    brandIconMaxWidth: 37,
-    fontSizeXS: '8',
-    fontSizeXXL: '28',
-    fontWeightNormal: '400',
-    fontWeightLight: '300',
-    fontWeightStrong: 500,
-  };
 
   private antdConfig: AntdThemeConfig;
 
   private constructor({ config }: { config?: AnyThemeConfig }) {
     this.SupersetThemeProvider = this.SupersetThemeProvider.bind(this);
-
-    // Create a new config object with default tokens
-    const newConfig: AnyThemeConfig = config ? { ...config } : {};
-
-    // Ensure token property exists with defaults
-    newConfig.token = {
-      ...Theme.defaultTokens,
-      ...(config?.token || {}),
-    };
-
-    this.setConfig(newConfig);
+    this.setConfig(config || {});
   }
 
   /**
@@ -118,9 +53,48 @@ export class Theme {
    * If simple tokens are provided as { token: {...} }, they will be applied with defaults
    * If no config is provided, uses default tokens
    * Dark mode can be set via the algorithm property in the config
+   * @param config - The theme configuration
+   * @param baseTheme - Optional base theme to apply under the config
    */
-  static fromConfig(config?: AnyThemeConfig): Theme {
-    return new Theme({ config });
+  static fromConfig(
+    config?: AnyThemeConfig,
+    baseTheme?: AnyThemeConfig,
+  ): Theme {
+    let mergedConfig: AnyThemeConfig | undefined = config;
+
+    if (baseTheme && config) {
+      mergedConfig = {
+        ...baseTheme,
+        ...config,
+      } as AnyThemeConfig;
+
+      if (baseTheme.token || config.token) {
+        (mergedConfig as any).token = {
+          ...(baseTheme.token || {}),
+          ...(config.token || {}),
+        };
+      }
+
+      if (baseTheme.components || config.components) {
+        const mergedComponents: any = { ...(baseTheme.components || {}) };
+
+        if (config.components) {
+          const configComponents = config.components as any;
+          Object.keys(configComponents).forEach(componentName => {
+            mergedComponents[componentName] = {
+              ...(mergedComponents[componentName] || {}),
+              ...configComponents[componentName],
+            };
+          });
+        }
+
+        (mergedConfig as any).components = mergedComponents;
+      }
+    } else if (baseTheme && !config) {
+      mergedConfig = baseTheme;
+    }
+
+    return new Theme({ config: mergedConfig });
   }
 
   private static getFilteredAntdTheme(
@@ -148,22 +122,15 @@ export class Theme {
   setConfig(config: AnyThemeConfig): void {
     const antdConfig = normalizeThemeConfig(config);
 
-    // Apply default tokens to token property
-    antdConfig.token = {
-      ...Theme.defaultTokens,
-      ...(antdConfig.token || {}),
-    };
-
     // First phase: Let Ant Design compute the tokens
     const tokens = Theme.getFilteredAntdTheme(antdConfig);
 
     // Set the base theme properties
     this.antdConfig = antdConfig;
     this.theme = {
-      ...Theme.defaultTokens,
-      ...antdConfig.token, // Passing through the extra, superset-specific tokens
-      ...tokens,
-    };
+      ...tokens, // First apply Ant Design computed tokens
+      ...(antdConfig.token || {}), // Then override with our custom tokens
+    } as SupersetTheme;
 
     // Update the providers with the fully formed theme
     this.updateProviders(
@@ -250,5 +217,3 @@ export class Theme {
     );
   }
 }
-
-/* eslint-enable theme-colors/no-literal-colors */
--- a/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
@@ -25,7 +25,7 @@ import {
   CacheProvider as EmotionCacheProvider,
 } from '@emotion/react';
 import createCache from '@emotion/cache';
-import { noop } from 'lodash';
+import { noop, mergeWith } from 'lodash';
 import { GlobalStyles } from './GlobalStyles';
 import {
   AntdThemeConfig,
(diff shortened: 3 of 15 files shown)


superset-frontend/src/types/bootstrapTypes.ts
```diff
@@ -148,6 +148,8 @@ export interface MenuData {
 export interface BootstrapThemeDataConfig {
   default: SerializableThemeConfig | {};
   dark: SerializableThemeConfig | {};
+  baseThemeDefault?: SerializableThemeConfig | null;
```
mistercrunch: I have to admit I didn't think about this when ideating. It's kind of getting out-of-control that we're sending 4 theme configs over from the backend on every page load. Wondering if there's a way that the THEME_DEFAULT/THEME_DARK could be treated as BASE?

mistercrunch: What if we just treated themes in CRUD as overrides on top of THEME_DEFAULT/THEME_DARK?

superset-frontend/packages/superset-ui-core/src/theme/GlobalStyles.tsx
```diff
@@ -16,6 +16,18 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
+// @fontsource/* v5.1+ doesn't play nice with eslint-import plugin v2.31+
+/* eslint-disable import/extensions */
+import '@fontsource/inter/200.css';
```
mistercrunch, line 22: These fonts got carried over from the previous implementation, and they do look nice, but I'm wondering whether they're kind of just bloating the build in some ways. I'm wondering if we could either:
- drop them to and use stock fonts in default build
- dogfood the custom font system in default configs

Probably beyond the scope of this PR, but wanted to bring it up.

mistercrunch: @kasiazjc what do you think? Could we find some of the best-looking browser stock fonts for vanilla Superset?

superset-frontend/src/types/bootstrapTypes.ts
gabotorresruiz: That's a great point! You're absolutely right that sending 4 theme configs on every page load is getting excessive. I think your suggestion to treat `THEME_DEFAULT/THEME_DARK` as the base themes makes a lot of sense.

Here's what I'm thinking:
**Current architecture:**
- **BASE_THEME_DEFAULT/BASE_THEME_DARK:** organization-wide base tokens
- **THEME_DEFAULT/THEME_DARK:** system themes (merged with base)
- **User CRUD themes:** also merged with base

**Simplified architecture:**
- **THEME_DEFAULT/THEME_DARK:** act as both system defaults and base themes
- **User CRUD themes:** simple overrides on top of the system themes

Let me know your thoughts on this @mistercrunch

rusackas:
Does the PR description need an update/reboot after the latest changes?

superset-frontend/packages/superset-ui-core/src/theme/GlobalStyles.tsx
rusackas: Inter is nice because not only is it a pretty font, but it includes a lot of OpenType featues, e.g. fixed-spacing for numerals, and a bunch of other goodies.

rusackas: https://rsms.me/inter/

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
```diff
@@ -118,9 +53,48 @@ export class Theme {
...
+      } as AnyThemeConfig;
+
+      if (baseTheme.token || config.token) {
+        (mergedConfig as any).token = {
+          ...(baseTheme.token || {}),
+          ...(config.token || {}),
+        };
+      }
+
+      if (baseTheme.components || config.components) {
```
mistercrunch: didn't look deep into the logic, but is this whole section a use case for `lodash`'s merge function? Maybe can save ~30 LoC

superset-frontend/src/pages/ThemeList/index.tsx
```diff
@@ -340,14 +360,28 @@ function ThemesList({
...
+                  >
                     <Icons.SunOutlined /> {t('Default')}
                   </Tag>
                 </Tooltip>
               )}
               {original.is_system_dark && (
                 <Tooltip title={t('This is the system dark theme')}>
-                  <Tag color="default">
+                  <Tag
+                    color="default"
```
mistercrunch: NIT: repeating the lines above, wondering the elegant way to do this. Maybe either `marginRight` on that first one or [maybe better] wrapping both into `<Space size="small" style={{ display: 'inline-flex' }}>` might be most antd-aligned, style-free. I think it'll generate and empty div but that's probably ok.

mistercrunch: Somehow came to hate most `css` and even `style` (to a slightly lesser extend) while working on theming for months. I think I'm getting a little OCD about it...

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -698,18 +743,38 @@ export class ThemeController {
...
       this.fallbackToDefaultMode();
     }
   }
 
+  /**
+   * Determines if a theme configuration is for dark mode.
+   * @param theme - The theme configuration to check
+   * @returns True if the theme is dark mode, false otherwise
+   */
+  private isThemeDark(theme: AnyThemeConfig): boolean {
```
mistercrunch: already exists in packages/superset-ui-core/src/theme/utils/themeUtils.ts

superset/daos/theme.py
```diff
@@ -47,25 +47,20 @@ def find_system_default(cls) -> Optional[Theme]:
 
         if len(system_defaults) > 1:
             logger.warning(
-                "Multiple system default themes found (%s), "
-                "falling back to config theme",
+                "Multiple system default themes found (%s), none will be used",
                 len(system_defaults),
             )
+            return None
```
mistercrunch: this line probably not needed

superset/daos/theme.py
```diff
@@ -76,13 +71,10 @@ def find_system_dark(cls) -> Optional[Theme]:
 
         if len(system_darks) > 1:
             logger.warning(
-                "Multiple system dark themes found (%s), falling back to config theme",
+                "Multiple system dark themes found (%s), none will be used",
                 len(system_darks),
             )
+            return None
```
mistercrunch: line not needed as there's a catchall bellow

superset/views/base.py
```diff
@@ -311,66 +311,76 @@ def menu_data(user: User) -> dict[str, Any]:
...
+    """Check if a theme should be replaced with base theme entirely"""
+    if theme is None or theme == {}:
+        return True
+
+    return False
+
+
+def _load_theme_from_model(
+    theme_model: Any | None, fallback_theme: dict[str, Any] | None, theme_type: str
+) -> dict[str, Any] | None:
```
mistercrunch: there's a type for it here superset/themes/types.py

superset/views/base.py
```diff
@@ -311,66 +311,76 @@ def menu_data(user: User) -> dict[str, Any]:
...
+            return json.loads(theme_model.json_data)
+        except json.JSONDecodeError:
+            logger.error(
+                "Invalid JSON in system %s theme %s", theme_type, theme_model.id
+            )
+            return fallback_theme
+    return fallback_theme
+
+
+def _process_theme(theme: dict[str, Any] | None, theme_type: str) -> dict[str, Any]:
```
mistercrunch: superset/themes/types.py

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -572,6 +589,30 @@ export class ThemeController {
...
+
+    // Check if theme has any tokens defined
+    if (theme.token && Object.keys(theme.token).length > 0) return false;
+
+    // Check if theme has components defined
+    if (theme.components && Object.keys(theme.components).length > 0)
+      return false;
+
+    return true;
+  }
```
msyavuz: This doesn't look right to me, can we add proper types to the argument? First two checks might not be needed

superset-frontend/packages/superset-ui-core/src/theme/types.ts
```diff
@@ -78,6 +78,7 @@ export type SerializableThemeConfig = {
   algorithm?: ThemeAlgorithmOption;
   hashed?: boolean;
   inherit?: boolean;
+  cssVar?: boolean | { key?: string; prefix?: string };
```
msyavuz, line 81: I am not sure what this is

gabotorresruiz: The `cssVar` property is an `Ant Design` theme configuration option that controls whether CSS variables are generated for the theme tokens.

According to the official Ant Design documentation at --> https://ant.design/docs/react/css-variables:
- **cssVar:** true - Enables CSS variables for all theme tokens (e.g., --ant-color-primary)
- **cssVar:** false - Disables CSS variables (default)
- **cssVar:** { key?: string; prefix?: string } - Enables CSS variables with custom configuration:
- **key:** The key for the CSS variables container (helps with theme isolation)
- **prefix:** The prefix for CSS variable names (default: 'ant')

This allows themes to expose their design tokens as CSS variables that can be used in custom CSS/styled-components. For example, when enabled, you could use `var(--ant-color-primary)` in your CSS.

I wanted to preserve this in our `SerializableThemeConfig` type to ensure full compatibility with Ant Design's theming system --> https://ant.design/docs/react/customize-theme#theme

superset-frontend/src/theme/ThemeController.ts
gabotorresruiz: You're right, I was overcomplicating things here. Thanks, nice catch!

superset/views/base.py
gabotorresruiz: Sure! I'll change this

superset/daos/theme.py
gabotorresruiz: Agree!

superset/daos/theme.py
gabotorresruiz: Agree as well!

superset-frontend/src/theme/ThemeController.ts
gabotorresruiz: Yes but that function serves a different purpose, it checks if a computed theme is `dark` by examining the `background color`.

The `ThemeController` needed something different, it needs to check if a configuration uses dark algorithm before the theme is computed. So I added a new function `isThemeConfigDark` alongside the `isThemeDark` in `packages/superset-ui-core/src/theme/utils/themeUtils.ts`

superset-frontend/src/pages/ThemeList/index.tsx
gabotorresruiz: You're right! I think I found a "better/elegant" way of doing this

superset-frontend/packages/superset-ui-core/src/theme/Theme.tsx
gabotorresruiz: Yep, `mergeWith` will do the trick. Nice catch!

superset-frontend/src/theme/ThemeController.ts
mistercrunch: oh right. Though in theory could potentially have a dark theme without the dark algorithm, though highly unlikely/impractical. I think the main Theme class has or used to have an isDark method too. `theme.fromConfig(themeConfig).isDark` or similar might work too but require a bit more computation. 

there's a subtle difference between a theme config and a computed theme, probably the typing system should be aware of the difference (maybe it is already, I'd have to check) ... The Theme class knows how to generate one from the other. There might be two `isThemeConfigDark` and `isThemeDark` .... If we need both maybe they both live in utils.

I'm not sure what makes sense to do in the context of this PR, but might be good to sort this out.

gabotorresruiz: @mistercrunch please take a look into this --> https://github.com/apache/superset/commit/d4eb79965843dc2dfcb524ad5d3eadd856169c6f#diff-b48e95953d3af3d6d542dd931e3269b3f74031f32a292dd4b77fb11ddba31097R128

Let me know your thoughts

superset-frontend/packages/superset-ui-core/src/components/Typography/Typography.test.tsx
```diff
@@ -50,7 +50,7 @@ describe('Typography Component', () => {
 
   it('renders strong text', () => {
     render(<Typography.Text strong>Strong Text</Typography.Text>);
-    expect(screen.getByText('Strong Text')).toHaveStyle('font-weight: 500');
+    expect(screen.getByText('Strong Text')).toHaveStyle('font-weight: 600');
```
mistercrunch, line 53: I don't understand why this test needed to change, from my understanding we just moved the default base theme from the frontend to the backend.

gabotorresruiz: The test changed because `fontWeightStrong: 500` theme token was moved from the frontend (Theme.tsx) to the backend (config.py). During unit tests, the backend config isn't available, so Ant Design's default (600) is used instead. The test now reflects this fallback behavior.

mistercrunch: Oh, then I'm surprised we don't have more unit tests failing. Noting this test doesn't seem to use the spec/helpers/ render (which renders with a theme provider), it does `import '@testing-library/jest-dom';` instead of `@superset-ui/core/spec`, wondering if it has something to do with it.

Sidetrack (outside the scope of this PR): It's a bit of a mess around this, might be good to not allow importing `@testing-library` directly so we're always providing a theme context. It's a bit confusing since each library might have slightly different helpers (useRedux:true in main app for instance).

Not sure what to do with all this, maybe it's just out-of-scope of this PR and this change is ok.

superset-frontend/src/pages/ThemeList/index.tsx
```diff
@@ -112,6 +117,16 @@ function ThemesList({
...
+  // State for confirmation modal
+  const [confirmModalConfig, setConfirmModalConfig] = useState<{
+    visible: boolean;
+    title: string;
+    message: string;
+    onConfirm: () => Promise<any>;
+    successMessage: string;
+    errorMessage: string;
+  } | null>(null);
+
```
alexandrusoare, line 132: Maybe it would be better to have a separate type for this state, something like `confirmModalConfigType`

superset-frontend/src/pages/ThemeList/index.tsx
```diff
@@ -235,60 +253,83 @@ function ThemesList({
...
-  };
+  const showThemeConfirmation = useCallback(
+    (config: {
+      title: string;
+      content: string;
+      onConfirm: () => Promise<any>;
+      successMessage: string;
+      errorMessage: string;
+    }) => {
+      setConfirmModalConfig({
```
alexandrusoare, line 267: ditto, having a type for the config would be a nice nit

superset/config.py
```diff
@@ -723,46 +723,69 @@ class D3TimeFormat(TypedDict, total=False):
...
+        "colorSuccess": "#5ac189",
+        "colorInfo": "#66bcfe",
+        # Fonts
+        "fontFamily": "'Inter', Helvetica, Arial",
+        "fontFamilyCode": "'Fira Code', 'Courier New', monospace",
+        # Extra tokens
+        "transitionTiming": 0.3,
+        "brandIconMaxWidth": 37,
+        "fontSizeXS": "8",
+        "fontSizeXXL": "28",
```
alexandrusoare, line 774: Since ant design has changed the fontSizeXS from 12 to 14, maybe we should use 14 as well, just a nit

superset/config.py
```diff
@@ -723,46 +723,69 @@ class D3TimeFormat(TypedDict, total=False):
...
+        # Default colors
+        "colorPrimary": "#2893B3",  # NOTE: previous lighter primary color was #20a7c9 # noqa: E501
+        "colorLink": "#2893B3",
+        "colorError": "#e04355",
+        "colorWarning": "#fcc700",
+        "colorSuccess": "#5ac189",
+        "colorInfo": "#66bcfe",
+        # Fonts
+        "fontFamily": "'Inter', Helvetica, Arial",
+        "fontFamilyCode": "'Fira Code', 'Courier New', monospace",
```
alexandrusoare, line 769: Why does Inter needs ' '?

superset/config.py
```diff
@@ -723,46 +723,69 @@ class D3TimeFormat(TypedDict, total=False):
...
+        # Extra tokens
+        "transitionTiming": 0.3,
+        "brandIconMaxWidth": 37,
+        "fontSizeXS": "8",
+        "fontSizeXXL": "28",
+        "fontWeightNormal": "400",
+        "fontWeightLight": "300",
+        "fontWeightStrong": 500,
```
alexandrusoare:
```suggestion
        "fontWeightStrong": "500",
```

superset/daos/theme.py
```diff
@@ -45,27 +45,14 @@ def find_system_default(cls) -> Optional[Theme]:
...
-                len(system_defaults),
-            )
-
-        # Fallback to is_system=True theme with name 'THEME_DEFAULT'
-        return (
-            db.session.query(Theme)
-            .filter(Theme.is_system.is_(True), Theme.theme_name == "THEME_DEFAULT")
-            .first()
-        )
+        return None
```
alexandrusoare, line 49: Why did we drop the logic if there are multiple themes found for either default or dark?

superset/views/base.py
```diff
@@ -311,66 +313,86 @@ def menu_data(user: User) -> dict[str, Any]:
...
 
-    if not is_valid_theme(dark_theme):
-        logger.warning(
-            "Invalid dark theme configuration: %s, using empty theme",
-            dark_theme,
-        )
-        dark_theme = {}
+    # Process and validate themes
+    default_theme = _process_theme(default_theme, "default")
+    dark_theme = _process_theme(dark_theme, "dark")
```
alexandrusoare: Maybe we can use enums for `dark`, `light`, `default` etc

alexandrusoare (approved):
LGTM, just a few nits

superset-frontend/src/pages/ThemeList/index.tsx
gabotorresruiz: Sounds good!

superset/daos/theme.py
gabotorresruiz: it was intentionally simplified. Returning `None` causes the system to use the `config.py` themes as fallback, which is the desired behavior.

superset/views/base.py
gabotorresruiz: Yes, we already have one defined in `superset/themes/types.py` so I'll just use that one. Nice catch!

mistercrunch (approved):
LGTM!

PR #35222, opened 2025-09-20
refactor(Notification): use Ant Design notifications for toasts

msyavuz:
### SUMMARY

Replace the toast implementation with Ant Design notifications.

Before:
image

After:
image

### TESTING INSTRUCTIONS

Visual testing

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


rusackas:
Can we get rid of some of this extra padding at the bottom?
Pasted_Image_9_22_25__10_49 AM

mistercrunch:
Would be surprised if it looked this way in vanilla antd ... styles spilling over?

msyavuz:
I don't think its styles spilling over, ~we are only using the message but we should be using description as well~ it was fixed in recent antd versions. I will update the screenshots to better reflect proper usage afterfixing it

msyavuz:
Recent antd version causes a lot of test failures, i say we merge with the bottom padding now, it will be fixed as a byproduct when ant design gets upgraded. @rusackas

mistercrunch:
> Recent antd version causes a lot of test failures

Oh no! is it a major version? What seems to be the pattern behind these errors?

msyavuz:
Oops, since i force pushed all those are not visible on the pr now. https://github.com/apache/superset/actions/runs/17945544416?pr=35222

> > Recent antd version causes a lot of test failures
> 
> Oh no! is it a major version? What seems to be the pattern behind these errors?

It's a minor version but that minor bump seems to render a bunch of things twice for some reason

mistercrunch:
That's a bit concerning, I guess we'll have to deal with it at some point ...

mistercrunch:
Whatever temporary fix we set for style we should TODO / comment around referencing the issue if there's one or PR comment above directly

msyavuz:
I will give upgrading ant design a go on [this pr](https://github.com/apache/superset/pull/35252). You’re right that it’s something we’ll need to tackle sooner or later, so I might as well make life easier for future me

sadpandajoe:
@msyavuz should we rebase or should we close this PR if it's no longer needed?

msyavuz:
@sadpandajoe This is still a needed change. Enzo is upgrading antd version in another pr, after that i will come back to this and do a rebase/cleanup

msyavuz:
This might in fact not be needed, closing it!

PR #35223, opened 2025-09-20
chore(effect): add eslint plugin to reduce rerenders

msyavuz:
### SUMMARY

useEffect is widely used in the frontend of superset.
https://react.dev/learn/you-might-not-need-an-effect#

There is an eslint plugin for catching these: 
https://github.com/NickvanDyke/eslint-plugin-react-you-might-not-need-an-effect

PR also includes a sample performance improvement for select component that gets rid of useEffects that are used for derived state. https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/.eslintrc.js
+++ b/superset-frontend/.eslintrc.js
@@ -83,6 +83,7 @@ module.exports = {
     'plugin:react-hooks/recommended',
     'plugin:react-prefer-function-component/recommended',
     'plugin:storybook/recommended',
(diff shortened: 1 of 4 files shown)


rusackas (approved):
Looking foward to all the followup PRs :)

For others; This raises lots of warnings that will be followe up on separately.

PR #35227, opened 2025-09-22
ci-cd yaml file added to branch

balaji-a11y:
I have uploaded the ci-cd file to baseline/upstream-5.0.0 branch.


msyavuz:
I don't think this was supposed to be opened against upstream. If you intended to open this can you add some description?

Issue #34162, opened 2025-07-14
[SIP-177] Proposal for SQL Lab Extensions


michael-s-molina:
### Motivation

Superset’s evolution and adoption have highlighted the limitations of a monolithic architecture. As the platform’s user base has grown, so have the demands for advanced features, custom workflows, and integrations. Many organizations require capabilities such as query explainers, optimizers, or natural language support, but the absence of a robust extension model has often forced teams to fork the codebase or make invasive modifications. This leads to significant maintenance overhead, fragmentation, and a barrier to broader community contributions.

This SIP is a direct follow-up to [SIP-151](https://github.com/apache/superset/issues/31932) and builds on the direction established in previous discussions/proposals. The primary motivation is to establish a foundation for extensibility in SQL Lab by defining clear patterns and constructs for extensions. The objective at this stage is not to fully implement all possible APIs or extension points, but rather to continue toward a flexible architecture (using SQL lab as a starting point) and a set of guiding principles that the community can iterate on and expand over time. This approach encourages experimentation and shared ownership, allowing extensions to evolve organically as new needs and use cases emerge. While the initial focus is on SQL Lab, the intention is to reuse these patterns and APIs across other modules in Superset, such as dashboards, charts, and connectors, to promote consistency and extensibility throughout the platform.

During the course of this work, we decided to change the terminology from "Plugins" to "Extensions." This change reflects a more accurate description of the intended architecture and philosophy. While "plugin" often implies a narrowly scoped add-on or isolated feature, "extension" better conveys the idea of enhancing and expanding the core capabilities of Superset in a modular and integrated way. The term "extension" also aligns with broader industry conventions, emphasizing the goal of providing well-defined, versioned points for the community to build upon and extend the platform’s functionality.

The design of this extension architecture draws inspiration from the successful model established by [VS Code Extensions](https://code.visualstudio.com/api). VS Code has demonstrated how a thoughtfully designed extension architecture can empower a broad community to innovate, share, and maintain enhancements without burdening the core maintainers or fragmenting the platform. By adopting similar principles—such as clear contribution points, versioned APIs, and a modular approach—this SIP aims to create an environment where the Superset community can collaboratively define and expand the platform’s capabilities over time.

### The vision

[...]

PR #35235, opened 2025-09-22
fix: docker-compose file and superset-core

betodealmeida:
### SUMMARY

`docker compose up` stopped working because of the introduction of `superset-core`.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


docker/docker-bootstrap.sh
```diff
@@ -18,9 +18,16 @@
 
 set -eo pipefail
 
-# Make python interactive
+# Install critical dependencies first in dev mode to fix volume mount issues
 if [ "$DEV_MODE" == "true" ]; then
     if [ "$(whoami)" = "root" ] && command -v uv > /dev/null 2>&1; then
+      echo "Installing critical dependencies for dev mode..."
+      # Install pydantic first as it's needed for imports
+      uv pip install --no-cache-dir "pydantic>=2.8.0"
```
mistercrunch: This should be in requirements/development.txt already (?)

docker/docker-bootstrap.sh
```diff
@@ -18,9 +18,16 @@
...
+      echo "Installing base requirements"
+      uv pip install -r /app/requirements/base.txt
```
mistercrunch: Should we do development.txt here?

docker/docker-bootstrap.sh
```diff
@@ -18,9 +18,16 @@
...
+      echo "Reinstalling superset-core in editable mode"
+      uv pip install -e /app/superset-core
```
mistercrunch: wondering what this is about (?) I have to admit I don't understand the new structure

docker/docker-bootstrap.sh
michael-s-molina: @mistercrunch I think there are folks that use it for production even if not recommended 🤷🏼‍♂️

docker/docker-bootstrap.sh
michael-s-molina: @mistercrunch `superset-core` is similar to `superseet-ui-core`. I recommend reading [[SIP-177] Proposal for SQL Lab Extensions](https://github.com/apache/superset/issues/34162) to understand the structure better.

docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+if [ "$DEV_MODE" == "true" ]; then
+    if [ "$(whoami)" = "root" ] && command -v uv > /dev/null 2>&1; then
+        echo "Dev mode detected. Ensuring critical dependencies are installed..."
+        # Quick install of critical packages that are needed for imports
+        uv pip install --no-cache-dir "pydantic>=2.8.0" "marshmallow-union" 2>/dev/null || true
+    fi
+fi
+
+# Now run the actual bootstrap script
+exec "$@"
```
michael-s-molina: @betodealmeida You need a blank line at the end here to fix CI 😉 
```suggestion
exec "$@"

```

michael-s-molina (approved):
Thank you for the fix @betodealmeida!

dpgaspar (approved):
Nice! thks for the fix @betodealmeida. LGTM, yet would be great if we could avoid the hardcoded pydantic install

rusackas (approved):
LGTM!

michael-s-molina:
@betodealmeida It seems there's an error yet: https://github.com/apache/superset/actions/runs/17953994393/job/51060324919

betodealmeida:
> @betodealmeida It seems there's an error yet: https://github.com/apache/superset/actions/runs/17953994393/job/51060324919

I'll fix that in a separate PR.

PR #35237, opened 2025-09-23
refactor: remove obsolete Flask flash messaging system

mistercrunch:
## Summary

The Flask flash message system is a legacy feature from when Superset was a server-rendered application. In the current SPA architecture, these messages are never displayed to users because:
- Flash messages only render on page load via FlashProvider's componentDidMount
- Modern UI interactions use async API calls that don't trigger page reloads
- The main consumer (/explore/ POST endpoint) is already marked @deprecated

All user-facing notifications are already handled by the frontend toast system, including chart save operations (see saveModalActions.ts) which dispatches success toasts for:
- Chart saved/overwritten
- Chart added to dashboard
- New dashboard created with chart

### Backend
- Removed all flash() calls from views (14 occurrences in 4 files)
- Converted error flashes to JSON responses or logging
- Removed redirect_with_flash utility function
- Fixed open redirect vulnerability in dashboard access denial

### Frontend
- Deleted FlashProvider component and tests
- Removed flash_messages from CommonBootstrapData type
- Cleaned up context providers and test fixtures
- Removed unused getBootstrapData imports

### Security Fixes
- Fixed open redirect vulnerability by using url_for() instead of request.url
- Dashboard access denial now uses safe URL construction

### Code Cleanup
- Removed unnecessary pass statements and comments
- Converted permalink errors to JSON responses for consistency
- Verified no tests depend on flash functionality

## BREAKING CHANGE
Removes flask.flash() messaging infrastructure. However, no actual functionality is lost as the frontend already handles all notifications through its own toast system.

## TESTING INSTRUCTIONS
1. Save a chart from Explore - verify toast notifications appear
2. Add chart to dashboard - verify success message
3. Try accessing a dashboard without permissions - verify proper redirect
4. Run frontend tests: `npm test`
5. Run backend tests: `pytest`

## ADDITIONAL INFORMATION
The application now relies entirely on client-side toast notifications for user feedback, which properly support async operations and provide a consistent UX.


superset-frontend/src/dashboard/actions/hydrate.js
```diff
@@ -274,7 +274,6 @@ export const hydrateDashboard =
           superset_can_csv: findPermission('can_csv', 'Superset', roles),
           common: {
             // legacy, please use state.common instead
-            flash_messages: common?.flash_messages,
```
rusackas, line 277: Question from review session: Where are server-side messages being processed/summoned now that they're not provided as part of the bootstrap data?

superset/views/auth.py
```diff
@@ -66,7 +65,7 @@ def activation(self, activation_hash: str) -> WerkzeugResponse:
         reg = self.appbuilder.sm.find_register_user(activation_hash)
         if not reg:
             logger.error(LOGMSG_ERR_SEC_NO_REGISTER_HASH, activation_hash)
-            flash(as_unicode(self.false_error_message), "danger")
+            logger.error("Registration activation failed: %s", self.false_error_message)
```
rusackas, line 68: @mistercrunch this is one of the errors that used to be (rightly) surfaced to users, but now might silent in the UI. These (and other?) server-side messages might no longer appear on the frontend.

rusackas: Relayed: We just need to make sure we have a frontend counterpart for all server side messages that were deleted.

superset-frontend/src/dashboard/actions/hydrate.js
mistercrunch: Answer is case by case, but short answer is "as they should" from async calls. Checked that most had been re-implemented the right way since they were used. Also confirmed that old mechanics are to surface "on the next page load", which has very few pre-determined triggers, stuff like logging in maybe.

superset/views/auth.py
mistercrunch: Let me confirm that logging in is a SPA operation as it should be as opposed to a backend-round-trip.

mistercrunch: Login screen is the only place where this seems to actually work (round-trip + timely flash). Let me refactor this nonsense!

mistercrunch: Mitigated with a toast, but had to keep the round trip (same as now) at least for now

superset/views/base.py
```diff
@@ -596,7 +591,7 @@ def _delete(self: BaseView, primary_key: int) -> None:
         try:
             self.pre_delete(item)
         except Exception as ex:  # pylint: disable=broad-except
-            flash(str(ex), "danger")
+            logger.error("Pre-delete error: %s", str(ex))
```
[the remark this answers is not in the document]
mistercrunch: this probably never should have gone through to user

superset/views/base.py
```diff
@@ -629,7 +623,7 @@ def muldelete(self: BaseView, items: list[Model]) -> FlaskResponse:
             try:
                 self.pre_delete(item)
             except Exception as ex:  # pylint: disable=broad-except
-                flash(str(ex), "danger")
+                logger.error("Pre-delete error: %s", str(ex))
```
mistercrunch, line 626: this probably never should have gone through to user

superset/views/core.py
```diff
@@ -428,9 +425,10 @@ def explore(  # noqa: C901
                         _("Error: permalink state not found"), status=404
                     )
             except (ChartNotFoundError, ExplorePermalinkGetFailedError) as ex:
-                flash(__("Error: %(msg)s", msg=ex.message), "danger")
```
mistercrunch, line 431: function flagged as @deprecated

superset/views/core.py
```diff
@@ -440,18 +438,8 @@ def explore(  # noqa: C901
             dataset_id = request.args.get("dataset_id")
             if slice_id:
                 initial_form_data["slice_id"] = slice_id
-                if form_data_key:
-                    flash(
-                        _("Form data not found in cache, reverting to chart metadata.")
```
mistercrunch, line 445: part of a function flagged as @deprecated

superset/views/core.py
```diff
@@ -440,18 +438,8 @@ def explore(  # noqa: C901
...
-                    )
             elif dataset_id:
                 initial_form_data["datasource"] = f"{dataset_id}__table"
-                if form_data_key:
-                    flash(
-                        _(
-                            "Form data not found in cache, reverting to dataset metadata."  # noqa: E501
```
mistercrunch, line 452: part of a function flagged as @deprecated

superset/views/core.py
```diff
@@ -652,13 +636,6 @@ def save_or_overwrite_slice(  # noqa: C901
                     _("You don't have the rights to alter this dashboard"),
                     status=403,
                 )
-
-            flash(
-                _("Chart [{}] was added to dashboard [{}]").format(
```
mistercrunch, line 657: tested that chart adding to dashboard is handled frontend now and flashed properly without these obsolete flash messages

superset/views/core.py
```diff
@@ -673,12 +650,6 @@ def save_or_overwrite_slice(  # noqa: C901
                 dashboard_title=request.args.get("new_dashboard_name"),
                 owners=[g.user] if g.user else [],
             )
-            flash(
-                _(
-                    "Dashboard [{}] just got created and chart [{}] was added to it"
```
mistercrunch, line 678: tested that chart adding to dashboard is handled frontend now and flashed properly without these obsolete flash messages

superset/views/utils.py
```diff
@@ -551,8 +550,3 @@ def get_cta_schema_name(
     if not func:
         return None
     return func(database, user, schema, sql)
-
-
-def redirect_with_flash(url: str, message: str, category: str) -> Response:
```
mistercrunch, line 556: not used anymore

superset/views/core.py
```diff
@@ -796,19 +767,9 @@ def dashboard(
...
-                redirect_url = url_for("DashboardModelView.list")
-                warn_msg = utils.error_msg_from_exception(ex)
-            return redirect_with_flash(
-                url=redirect_url,
-                message=warn_msg,
-                category="danger",
-            )
+        except SupersetSecurityException:
+            # Return 404 to avoid revealing dashboard existence
+            abort(404)
```
mistercrunch: 404 is most accurate when you don't have access

rusackas: Yep... it's even in the official [spec](https://www.rfc-editor.org/rfc/rfc9110.html) for 403 (I got curious):
> An origin server that wishes to "hide" the current existence of a forbidden [target resource](https://www.rfc-editor.org/rfc/rfc9110.html#target.resource) MAY instead respond with a status code of [404 (Not Found)](https://www.rfc-editor.org/rfc/rfc9110.html#status.404).

rusackas (approved):
LGTM!

PR #35159, opened 2025-09-16
chore: TypeScript Configuration Modernization and Cleanup

michael-s-molina:
### SUMMARY

Modernized and streamlined TypeScript configuration across Superset's monorepo, consolidating redundant configurations and standardizing project references for better maintainability and consistency.

#### 1. Root Configuration Modernization

- Consolidated `tsconfig.base.json` into main `tsconfig.json` to eliminate redundancy
- Updated module resolution from "node" to "bundler" (modern recommended approach)
- Updated target from "esnext" to "es2020" for better compatibility
- Enhanced include patterns to cover packages and plugins test files
- Added ECharts internal types path mapping to resolve bundler resolution conflicts
- Added `@apache-superset/core` path mappings for proper module resolution

#### 2. Package/Plugin tsconfig Cleanup (27+ files affected)

- Standardized project references - All packages now properly reference `superset-core`
- Simplified configurations using extends pattern from root `tsconfig.json`
- Eliminated redundant compiler options already defined in root config
- Consistent `baseUrl` and `outDir` settings across all packages
- Updated generator templates to use new standardized configuration structure

#### 3. Configuration Standardization

- All packages extend root `tsconfig.json` for consistency
- Proper project references to `superset-core`, `superset-ui-core`, and `chart-controls`
- Centralized path mappings in root `tsconfig.json` for better maintainability
- Fixed ECharts internal type imports blocked by bundler resolution
- Added `@apache-superset/core` dependency to packages that need it

#### 4. Additional Fixes

- CI workflow improvement - Added package build step before type checking
- Day.js plugin type resolution - Centralized imports in `preamble.ts`
- Asset type declarations - Added missing image format support

#### Technical Benefits

✅ Faster type checking - Proper project references and inheritance
✅ Simplified configuration - Clean extends-based inheritance pattern
✅ Better IDE support - Modern bundler resolution works better with IDEs
✅ Consistent build process - Standardized across all packages and plugins
✅ Eliminated build conflicts - No more TypeScript declaration file issues
✅ Maintained compatibility - All existing imports continue working

### TESTING INSTRUCTIONS

1. Type checking: `npm run type` passes without errors
2. Linting: `npm run eslint` continues working
3. Build: `npm run build` completes successfully
4. Plugin build: `npm run plugins:build` generates declaration files
5. Module resolution: All internal imports resolve correctly
6. External libraries: ECharts types resolved properly

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
[...]

PR #35240, opened 2025-09-23
feat(build): auto-rebuild/check TypeScript types for packages/plugins in webpack

mistercrunch:
After PR #35159 changes, developers had to manually run `npm run plugins:build` whenever types changed in packages or plugins. This adds automatic type checking and .d.ts generation to the webpack dev process using ForkTsCheckerWebpackPlugin.

Key improvements:
- Automatic type rebuilding when packages/plugins change
- Inline type error reporting in webpack console
- Incremental compilation for better performance
- skipLibCheck optimization for faster type checking
- Can be disabled with DISABLE_TYPE_CHECK=true npm run dev
- Fixes PropTypes compatibility for JavaScript files

This eliminates the need for manual type rebuilding and makes the development experience smoother while maintaining type safety.

### DIFF
--- a/superset-frontend/webpack.config.js
+++ b/superset-frontend/webpack.config.js
@@ -184,21 +184,64 @@ if (!isDevMode) {
       chunkFilename: '[name].[chunkhash].chunk.css',
     }),
   );
+}
 
-  // Runs type checking on a separate process to speed up the build
+// Type checking for both dev and production
+// In dev mode, this provides real-time type checking and builds .d.ts files for plugins
+// Can be disabled with DISABLE_TYPE_CHECK=true npm run dev
+if (isDevMode) {
+  if (process.env.DISABLE_TYPE_CHECK) {
+    console.log('⚡ Type checking disabled (DISABLE_TYPE_CHECK=true)');
+  } else {
+    console.log(
+      '✅ Type checking enabled (disable with DISABLE_TYPE_CHECK=true npm run dev)',
+    );
+    // Optimized configuration for development - much faster type checking
+    plugins.push(
+      new ForkTsCheckerWebpackPlugin({
+        typescript: {
+          memoryLimit: 8192,
+          build: true, // Generate .d.ts files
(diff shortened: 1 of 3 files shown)


superset-frontend/webpack.config.js
```diff
@@ -344,7 +390,12 @@ const config = {
   },
   resolve: {
     // resolve modules from `/superset_frontend/node_modules` and `/superset_frontend`
-    modules: ['node_modules', APP_DIR],
+    modules: [
+      'node_modules',
+      APP_DIR,
+      path.resolve(APP_DIR, 'packages'),
```
mistercrunch, line 396: I think these two lines here are the key, meaning now it watches plugins/packages and rebuilds.

superset-frontend/webpack.config.js
```diff
@@ -184,21 +184,64 @@ if (!isDevMode) {
...
+    console.log('⚡ Type checking disabled (DISABLE_TYPE_CHECK=true)');
+  } else {
+    console.log(
+      '✅ Type checking enabled (disable with DISABLE_TYPE_CHECK=true npm run dev)',
+    );
+    // Optimized configuration for development - much faster type checking
+    plugins.push(
+      new ForkTsCheckerWebpackPlugin({
+        typescript: {
+          memoryLimit: 8192,
```
geido: Does it need this much memory? The default seems to be 2048. This a pretty big jump

mistercrunch: Oh right, other places in the file we use 4096, maybe i'll stick to that

mistercrunch: Factored out a constant and using/reusing it places.

mistercrunch:
Everyone cool with the default of running type-checked in `npm run dev`? Seems like a better default to support my personal workflow and desirable for everyone else. Only reason to not default might be cpu/memory usage of that process, but I'm exposing DISABLE_TYPE_CHECK if people need that ...

eschutho (approved):

PR #35248, opened 2025-09-23
feat(themes): Adding SupersetText support to Themes Modal

rebenitez1802:
### SUMMARY

In order to be able to easily customize URL's in Theme Modal, `SupersetText` is added with propper fallback 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/src/features/themes/ThemeModal.tsx
+++ b/superset-frontend/src/features/themes/ThemeModal.tsx
@@ -21,6 +21,7 @@ import { FunctionComponent, useState, useEffect, ChangeEvent } from 'react';
 import { css, styled, t, useTheme } from '@superset-ui/core';
 import { useSingleViewResource } from 'src/views/CRUD/hooks';
 import { useThemeContext } from 'src/theme/ThemeProvider';
+import SupersetText from 'src/utils/textUtils';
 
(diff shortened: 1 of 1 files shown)


mistercrunch (approved):
LGTM

Antonio-RiveroMartnez (approved):
lgtm

rusackas:
Should we be adding these kinds of settings to an optional file? It seems like we should add this to the stock/default file, if we're going to have the frontend depend on it as a reference... but then if we're starting to do that, shouldn't it just be set in the config instead?

I see the text file as a bit of a "junk drawer" of things we can use for overrides, but it probably shouldn't be where we look up default values/overrides for configurable UI.

michael-s-molina:
> Should we be adding these kinds of settings to an optional file? It seems like we should add this to the stock/default file, if we're going to have the frontend depend on it as a reference... but then if we're starting to do that, shouldn't it just be set in the config instead?

+1. We shouldn't add configurations to `SupersetText`.

mistercrunch:
Oh my bad, I missed that during review. I [wrongly] assumed those constants already existed from prior work.

eschutho:
I'm not sure that we want to add small text changes to a config file, or am I missing something?

PR #35253, opened 2025-09-23
feat(ci): GitHub Models - Part 1 of `n`

rusackas:
### SUMMARY

Taking a first stab at adding GHA using the new GitHub Models feature(s). This is just copping a couple examples from this blog post:
https://github.blog/open-source/maintainers/how-github-models-can-help-open-source-maintainers-focus-on-what-matters/

Apparently the org has to pay for this model access and set it up on GitHub. It's unclear if we can validate that it's currently enabled without just... trying it. So here we go!

If it fails (fairly likely), then I'll follow up with ASF Infra to see a) if they can turn this on at all, and b) if/how we need to account for token usage/fees

### TESTING INSTRUCTIONS

1) Merge it
2) Open an issue!
3) 🙏 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/.github/workflows/github_model_detect_duplicate_issues.yml
+++ b/.github/workflows/github_model_detect_duplicate_issues.yml
@@ -1,0 +1,26 @@
+name: Detect Duplicate Issues
+
+on:
+  issues:
+    types: [opened, reopened]
+
+permissions:
+  models: read
+  issues: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+jobs:
+  continuous-triage-dedup:
+    if: ${{ github.event.issue.user.type != 'Bot' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: pelikhan/action-genai-issue-dedup@v0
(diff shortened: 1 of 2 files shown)


mistercrunch (approved):
new toys!

geido (approved):
Curious to see these in action!

msyavuz (approved):

rusackas:
Opened an Infra ticket here: https://issues.apache.org/jira/browse/INFRA-27273

rusackas:
Got the results of the INFRA ticket back:

> I'm afraid you are going to have to make do with the free version with the API limits:
> https://docs.github.com/en/github-models/use-github-models/prototyping-with-ai-models#rate-limits
> We wont be enabling the paid version sorry.

Seems plausible to do, but I'm not sure if it'd be enough bandwidth for us, and I'm not sure if we'd be pooling these tokens with all other ASF projects... it would seem so, which would compound the problem.

rusackas:
This isn't likely to work for us given the limitations. Closing for now.

PR #35259, opened 2025-09-24
chore: abstract models and daos into superset-core

villebro:
### SUMMARY
This introduces SQLAlchemy models and DAOs for entity access in the `superset_core.api` package as abstract classes. These in turn are implemented in the host package, which in turn are dependency injected back into the `superset_core.api` package, exposing them to extension developers. Furthermore, the old class based APIs are replaced with functional equivalents, providing a more modern and pythonic pattern.

The hierarchy goes as follows:
1. Abstract classes defined in `superset_core.api.models` and `superset_core.api.daos`. These contain a subset of the current functionality in the host. By starting out small and expanding the scope as needed, we can adhere to SemVer by incrementing minors when new features are added, rather than having to do major releases in case we need to break the current host API.
2. Models and DAOs are implemented in the host application by extending the abstract classes in 1. At a minimum they implement the minimum contract as defined by the abstract classes, but may add additional functionality.
3. Finally, the implementations are exposed in place of the abstract classes defined in step 1 via dependency injection. By doing so, extension developers get full access to the implementations from a single import. Overriding the abstracts is done with the intention of keeping the DevEx as simple as possible for developers, so they don't have to face the confusion of dual representations of all classes in the core package (one for the abstract class, the other for the injected implementation).

For now, the abstract DAOs only expose the following methods for CRUD operations (these read mathods are pretty inconsistent, so we may want to consider cleaning them up, too):
- `find_all() -> list[T]`
- `find_one_or_none(**filter_by: Any) -> T | None`
- `filter_by(**filter_by: Any) -> list[T]` 
- `query(query: SQLAQuery) -> list[T]`
- `create()`
- `update()`
- `delete()`

The abstract vs dependency injected classes can be see in action as follows:

Starting the Python REPL directly:
```python
$ python
Python 3.10.18+ (heads/3.10:1df5d00, Aug 15 2025, 07:55:42) [Clang 17.0.0 (clang-1700.0.13.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from superset_core.api.models import Dataset
>>> Dataset
<class 'superset_core.api.models.Dataset'>
```

Starting Flask shell (notice the class is now `SqlaTable` from the host):
```python
$ flask shell
Loaded your LOCAL configuration at [/Users/ville/pythonpath/superset_config.py]
2025-11-11 17:23:43,077:DEBUG:superset.utils.logging_configurator:logging was configured successfully
2025-11-11 17:23:43,085:DEBUG:root:Configured event logger of type <class 'superset.utils.log.DBEventLogger'>
2025-11-11 17:23:43,354:INFO:alembic.runtime.migration:Context impl SQLiteImpl.
2025-11-11 17:23:43,354:INFO:alembic.runtime.migration:Will assume non-transactional DDL.
2025-11-11 17:23:43,513:INFO:superset.app:Syncing configuration to database...
2025-11-11 17:23:43,516:DEBUG:superset.commands.theme.seed:Updated system theme: THEME_DEFAULT
2025-11-11 17:23:43,517:DEBUG:superset.commands.theme.seed:Updated system theme: THEME_DARK
2025-11-11 17:23:43,517:INFO:superset.app:Configuration sync to database completed successfully
Python 3.10.18+ (heads/3.10:1df5d00, Aug 15 2025, 07:55:42) [Clang 17.0.0 (clang-1700.0.13.5)] on darwin
App: superset.app
Instance: /Users/ville/projects/superset/instance
>>> from superset_core.api.models import Dataset
>>> Dataset
<class 'superset.connectors.sqla.models.SqlaTable'>
```

The decision to override the abstract/type definition in step 3 is similar to the pattern in the frontend core package, with the notable difference of these being actual classes, rather than type definitions. We considered taking a similar approach in the backend, where we would define pure Protocols rather than classes/functions, but quickly realized that this would not be feasible, as that would have required exposing Prototypes that have the full scope of the base SQLAlchemy models. In addition, using abstract classes gives us the freedom to add functionality to the abstract classes that carry over to the implementations.

### Old pattern
```python
# import class based APIs
from superset_core.api import rest_api, models

# fetch database entities
databases = models.get_databases(id=database_id)

# add extension API
rest_api.add_extension_api(DatasetReferencesAPI)
```

### New pattern

```python
# import functional APIs
from superset_core.api.models import Database, get_session
from superset_core.api.daos import DatabaseDAO
from superset_core.api.rest_api import add_extension_api

# fetch database entities
databases = DatabaseDAO.filter_by(uuid=database.uuid)

# add extension API
add_extension_api(DatasetReferencesAPI)
```

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-core/src/superset_core/dao/types.py
```diff
@@ -0,0 +1,127 @@
...
+        @classmethod
+        def find_by_id(cls, model_id: str | int) -> MyCustomModel | None:
+            # Implementation here
+            pass
+    ```
+    """
+
+    # Class attributes that implementations should define
+    model_cls: Optional[type[T_Model]]
+    base_filter: Optional[BaseFilter]
```
betodealmeida: Can we use something more generic here instead of `BaseFilter`, which is tightly coupled with FAB? Maybe we could rething how base filter works — it could be something passed to the instance, eg:

```python
dao = DatasetDAO()
datasets = dao.find_by_ids([1, 2])  # equivalent to `skip_base_filter=True`

dao_for_user = dao.filtered(User.id == current_user.id)
datasets = dao_for_user.find_by_ids([1, 2])  # equivalent to `skip_base_filter=False`
```

In general I think the less dependencies `superset-core` has, the better.

superset-core/src/superset_core/dao/types.py
```diff
@@ -0,0 +1,127 @@
...
+        model_id_or_uuid: str,
+        skip_base_filter: bool = False,
+    ) -> Optional[T_Model]:
+        """Find a model by ID or UUID."""
+        ...
+
+    @abstractmethod
+    def find_by_ids(
+        self,
+        model_ids: Union[list[str], list[int]],
```
betodealmeida: Can we use `Collection` instead of `list`? Here and in other places where order doesn't matter.

Also, can we use this opportunity to just use UUIDs everywhere, instead of having int IDs, string IDs, and UUIDs?

mistercrunch:
A tiny bit concerned about having `superset-core` treated (explicitely or implicitely) as a wide public API and the change management concerns that come with that. To help manage that, I'd propose that we:
- use python convention for private/public (`_method`= you probably shouldn't use that if care about easy upgrade, this is meant for the internal ecosystem) and `__method` -> definitely don't want to use that, even in the ecosystem/monorepo
- actively reduce the API surface, go through all the methods and "privatize" what's not essential in the public API
- make superset-core pre-1.0 and allow ourselves to break API and privatize things until 1.0 is officially out. From that point onwards we do proper semver/change-management on public methods/objects/properties

mistercrunch:
About Commands more specifically, I can see how some are **useful** here, but thinking most should either live in the app, or be treated as private.

mistercrunch:
Note: another option I mentioned before would be to treat superset-core as 100% private, and have something like a `superset-extension-sdk` as the thing that's public/semvered, it would import/re-export the bits of superset-core we want to actively expose (and maintain) as a public API.

superset-core/src/superset_core/dao/types.py
dpgaspar: Fluent Interfaces are cool! but like that makes me think why not just use `Query` from `SQLAlchemy`?
Have a somewhat static base filter is a plus for security, we enforce that all DAO operations take into account a certain filter that enforces a security constraint

superset-core/src/superset_core/api/types/models.py
```diff
@@ -43,48 +45,48 @@ def get_session() -> scoped_session:
...
 
-        :returns: The Database SQLAlchemy model class.
+        :returns: The Database implementation class.
         """
         ...
 
     @staticmethod
     @abstractmethod
-    def get_datasets(query: BaseQuery | None = None, **kwargs: Any) -> list[Any]:
+    def get_datasets(query: BaseQuery | None = None, **kwargs: Any) -> list[Dataset]:
```
dpgaspar: nit: We can use `Query` from SQLAlchemy

superset-core/src/superset_core/dao/types.py
```diff
@@ -0,0 +1,127 @@
...
+    # Class attributes that implementations should define
+    model_cls: Optional[type[T_Model]]
+    base_filter: Optional[BaseFilter]
+    id_column_name: str
+    uuid_column_name: str
+
+    @abstractmethod
+    def find_by_id(
+        self, model_id: Union[str, int], skip_base_filter: bool = False
+    ) -> Optional[T_Model]:
```
dpgaspar: nit: In the future we could have models with composite keys

superset-core/src/superset_core/models/base.py
```diff
@@ -0,0 +1,74 @@
...
+    Abstract base class that extends Flask-AppBuilder's Model.
+
+    This class provides the interface contract for all Superset models.
+    The host package provides concrete implementations.
+    """
+
+    __abstract__ = True
+
+
+class Database(CoreModel):
```
dpgaspar: Should we call this `AbstractDatabase` or `BaseDatabase` ?

michael-s-molina:
> A tiny bit concerned about having superset-core treated (explicitely or implicitely) as a wide public API and the change management concerns that come with that. 

@mistercrunch I share the same concerns as you and this is one of the motivations for exposing only the API definitions in `superset-core` without any implementation or private methods. In essence, the APIs should only contain public methods that we will support with semver. Any private method will not be part of `superset-core` and will be defined at the host implementation. For example:

```
ISomeDAO: (superset-core)
     @abstractmethod
     def public_function():
          ...

SomeDAO(ISomeDAO):  (host)
    def public_function():
         ... implementation logic
    
    def private_function():
         ... implementation logic
```

> make superset-core pre-1.0 and allow ourselves to break API and privatize things until 1.0 is officially out. From that point onwards we do proper semver/change-management on public methods/objects/properties

That's exactly what we defined in the SIP and the reason we publish the packages as 0.0.1-rcx in both NPM and PyPI.

michael-s-molina:
> Note: another option I mentioned before would be to treat superset-core as 100% private, and have something like a superset-extension-sdk as the thing that's public/semvered, it would import/re-export the bits of superset-core we want to actively expose (and maintain) as a public API.

@mistercrunch I think we're achieving the same objective but through different means. The `superset-core` contains only the bits we want to actively expose and the host application contain the private parts. The private parts don't need to be a separate package, they are already part of the `apache-superset` package.

Correlating with your example, `superset-core` is what you're calling as `superset-extension-sdk` but it's not only for extensions. It can be used by embedded and MCP tools.

mistercrunch:
> the APIs should only contain public methods

but clearly there are useful bits/methods/objects that we want to use/reuse within `superset-core` but don't want to or don't need to expose as part of the public API. I say we just put an underscore in their name to signal they're not part of the public api. 

One example might be something like `_is_uuid_or_int(any_id: int | str)`, clearly it's something we need to use/reuse, but really no need to make that public...

Python really lacking here, though doing it by convention with `_` and `__` prefix is fine.

Also restating that I think there are many pieces of logic that logically belong in `superset-core/` (especially if we want to keep the very-related logic together, in one place) that are logically private or only meant for the main app. 

Two main approaches here:
- have those bits of logic live in `superset-core`, close to the code they are highly related-to and marked as private
- crazy inheritance schemes (some highly-related DAOs/Commands live in superset-core, and superset-app inherits-and-augments them) -> not my favorite approach! ... Some commands are in one module/app, others are in another, calling `super()` or methods from the parent class, .... = inheritance confusion

superset-core/src/superset_core/dao/types.py
aminghadersohi: Regarding `Optional[T_Model]`, consider: many ORMs provide `get()` (raises) and `get_or_none()` (returns None) to handle different use cases. Do we want/need both of these or are we good with what is here?

When debugging missing entities, would we rather see `AttributeError: 'NoneType' object has no attribute 'name' on line  47`, or `EntityNotFoundError: Dashboard with id=123 not found on line 23`?

superset-core/src/superset_core/dao/types.py
```diff
@@ -0,0 +1,127 @@
...
+from flask_appbuilder.models.filters import BaseFilter
+from flask_sqlalchemy import BaseQuery
+
+from superset_core.models.base import CoreModel
+
+# Type variable bound to our CoreModel
+T_Model = TypeVar("T_Model", bound=CoreModel)
+
+
+class BaseDAO(Generic[T_Model], ABC):
```
aminghadersohi: I think this fits in perfectly with the architecture of mcp service. Currently the POC adds a set of MCP Core classes that need a BaseDAO like this passed so they can stay generic, but because there isn't one yet, i had to add this: 
```
class DAO(Protocol):
    """Protocol for Data Access Objects used in model tools.
    To be replaced with one from superset core
    """

    model_cls: Type[Any]

    @classmethod
    def list(
        cls,
        column_operators: Optional[List[Any]] = None,
        order_column: str = "changed_on",
        order_direction: str = "desc",
        page: int = 0,
        page_size: int = 10,
        search: Optional[str] = None,
        search_columns: Optional[List[str]] = None,
        custom_filters: Optional[Dict[str, Any]] = None,
        columns: Optional[List[str]] = None,
    ) -> tuple[List[Any], int]:
        """List method that all DAOs should implement."""
        ...

    @classmethod
    def find_by_id(cls, id: int) -> T | None:
        """Find by ID method that all DAOs should implement."""
        ...

    @classmethod
    def get_filterable_columns_and_operators(cls) -> Dict[str, Any]:
        """Get filterable columns and operators."""
        ...

    @classmethod
    def count(cls) -> int:
        """Count total number of records. Required for instance info."""
        ...

```

We can replace it later when this PR is decided upon super easily

superset-core/src/superset_core/api/types/models.py
villebro: Nice catch! Fixed

superset-core/src/superset_core/dao/types.py
villebro: I've also always been a fan of generic collection types when possible 👍 I agree we should also consider just dumping the old ids in favor of UUIDs now, as it's been a long time coming. Changing this now will be fairly simple, but will be a major breaking change going forward. Thoughts @michael-s-molina ? The main concern here is this would constitute a pretty major breaking change for the internal API.

superset-core/src/superset_core/models/base.py
villebro: Yeah, I was actually originally leaning towards `BaseDatabase`, but I feel this may be the exception to the rule, as these will be used kind of like instances by extension developers (they get actual instances that implement the abstract class). But I'm not super opinionated on this. @michael-s-molina any thoughts here?

superset-core/src/superset_core/dao/types.py
villebro: For the `find_*` methods, here's what I propose: On the public API, we only expose UUID based lookups, and keep the `int`, `str` etc ones private for now. We can promote them later if needed, but we'll try to get by with just the UUIDs to steer extension devs in a better direction from the beginning.

superset/security/manager.py
```diff
@@ -2372,7 +2372,7 @@ def raise_for_access(  # noqa: C901
                     continue
 
                 schema_perm = self.get_schema_perm(
-                    database,
+                    database.database_name,
```
villebro, line 2375: Bycatch: it expects the name (str), not the instance

superset/security/manager.py
```diff
@@ -2388,7 +2388,7 @@ def raise_for_access(  # noqa: C901
                 for datasource_ in datasources:
                     if self.can_access(
                         "datasource_access",
-                        datasource_.perm,
+                        datasource_.perm or "",
```
villebro, line 2391: Bycatch: `perm` can be `None`

superset-core/src/superset_core/dao/types.py
villebro: We've actually removed many of the base methods provided by the current host implementation to keep the signature as slim as possible. This of course doesn't limit the host from implementing additional methods, which it now does. In the future we may expose more methods as they become needed. In this case it will be a simple case of a minor bump on the core version.

rusackas:
This is quite the thread! Here's a Claude generated recap, if it helps anyone take stock of things. If it's not... just delete this comment! 😅

Committers can edit this comment and replace `[ ]` with `[X]` to track decisions.

| Item | To Do | Hold Off | Done |
|------|-------|----------|------|
| **Remove FAB dependency**: Replace `BaseFilter` with a more generic filtering approach to avoid Flask-AppBuilder coupling in superset-core | [ ] | [ ] | [ ] |
| **Type improvements**: Change `list` to `Collection` in method signatures where order doesn't matter | [ ] | [ ] | [ ] |
| **UUID migration**: Decide whether to migrate to UUID-only identifiers now (breaking change) or support mixed int/str/UUID | [ ] | [ ] | [ ] |
| **UUID-only for public API**: Expose only UUID-based lookups in public API, keep int/str lookups private (@villebro's proposal) | [ ] | [ ] | [ ] |
| **Composite key support**: Consider future support for composite primary keys in DAO methods | [ ] | [ ] | [ ] |
| **Error handling**: Decide between returning `Optional[T]` vs raising exceptions (implement both `get()` and `get_or_none()` patterns?) | [ ] | [ ] | [ ] |
| **Naming convention**: Standardize abstract class names (e.g., `Database` vs `AbstractDatabase` vs `BaseDatabase`) | [ ] | [ ] | [ ] |
| **Commands scope**: Determine which commands belong in superset-core vs treating most as private/app-specific | [ ] | [ ] | [ ] |
| **API exposure strategy**: Decide if superset-core should be 100% private with a separate superset-api package for public APIs | [ ] | [ ] | [ ] |
| **Private underscore convention**: Consider using `_` prefix convention for private modules/functions in superset-core | [ ] | [ ] | [ ] |
| **Security filtering approach**: Resolve @dpgaspar's concern about maintaining static base filters for security | [ ] | [ ] | [ ] |
| **Fix bycatch issues**: Address the two type issues found in security manager (`database.database_name` and `datasource_.perm or ""`) | [ ] | [ ] | [ ] |

superset/core/api/core_api_injection.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def inject_model_implementations() -> None:
+    """
+    Replace abstract model classes in superset_core.api.models with concrete
+    implementations from Superset.
+
+    Uses in-place replacement to maintain single import location for extensions.
+    """
+    import superset_core.api.models as core_models_module
```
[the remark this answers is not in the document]
villebro: What's your suggestion?

superset/core/api/core_api_injection.py
```diff
@@ -0,0 +1,180 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""
+Core API Dependency Injection
```
aminghadersohi, line 19: What is our best reply to: "Runtime injection breaks static analysis. IDEs can't "go to definition", type checkers can't follow the indirection, debugging is harder."

superset-core/src/superset_core/api/daos.py
```diff
@@ -0,0 +1,262 @@
...
+    SavedQuery,
+    Tag,
+    User,
+)
+
+# Type variable bound to our CoreModel
+T = TypeVar("T", bound=CoreModel)
+
+
+class BaseDAO(Generic[T], ABC):
```
aminghadersohi, line 56: The abstract DAOs still expose SQLAlchemy (BaseFilter, query(), etc.)?

aminghadersohi:
Left a couple of inline comments. the big one on my mind is who needs this? and other one is do we alternative solutions?

superset/core/api/core_api_injection.py
villebro: Well, I think static analysis should be fine, as the injected objects just replace the abstract definitions. Here's a screenshot from VS Code when I'm using the abstract core definitions:
image

If you do need the richer implementations, just import those from where they're defined. In the case of the DAOs and MCP tools in the `superset` package, you can just import them directly, in which case you'll see all the extra methods they make available that aren't featured on the abstract classes.

superset-core/src/superset_core/api/daos.py
villebro: For now those aren't exposed on the abstract classes. If needed, they can be explicitly called out there, too. But from the perspective of an extension developer, they don't need to think about BaseFilters, as it's just something that the host uses.

villebro:
> Left a couple of inline comments. the big one on my mind is who needs this? and other one is do we alternative solutions?

@aminghadersohi without these abstractions, extension developers would not be able to do the following:
- Expose REST API endpoints
- Do CRUD operations on Superset entities (databases, charts, dashboards etc)

Once this lands, I'm planning on moving over the MCP tool constructs here and startup registration of extension tools. After they land, it will be possible to create an extension that exposes additional tools on the MCP server without needing to modify the main `superset` package/codebase.

superset-core/src/superset_core/dao/types.py
michael-s-molina: @villebro Your suggested approach looks good 👍🏼

superset-core/src/superset_core/models/base.py
michael-s-molina: `Database` seems fine now that we dynamically replace the instances.

superset/core/api/core_api_injection.py
michael-s-molina: Ideally we would always interact with the public API. @aminghadersohi is there a use case where you would need an implementation detail?

villebro: IIRC, the main MCP implementation needs some of the DAO features we're not exposing here. But that's fine - the `superset` package is free to extend the public API and adhere to that internally.

michael-s-molina:
> Left a couple of inline comments. the big one on my mind is who needs this?

@aminghadersohi These APIs have 2 primary consumers:
- Backend MCP tools
- Backend extensions

Screenshot 2025-11-14 at 11 29 43

superset-core/src/superset_core/__init__.py
```diff
@@ -14,3 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+"""
+Apache Superset Core - Public API for Extension Development
```
michael-s-molina: It's not exclusive for extension development as it will be used by MCP tools as well.

```suggestion
Apache Superset Core - Public API with core functions of Superset
```

superset-core/src/superset_core/api/models.py
```diff
@@ -0,0 +1,295 @@
...
+    This abstract class defines the contract that dataset models should implement,
+    providing consistent data source operations and metadata.
+
+    It provides the public API for Datasets implemented by the host application.
+    """
+
+    __abstract__ = True
+
+    # Type hints for expected attributes (no actual field definitions)
+    id: int
```
michael-s-molina, line 92: Should we reduce the scope of what's exposed in these models? In other words, should all these attributes be part of the public API initially?

superset-core/src/superset_core/api/daos.py
```diff
@@ -0,0 +1,262 @@
...
+
+    @classmethod
+    @abstractmethod
+    def find_one_or_none(cls, **filter_by: Any) -> T | None:
+        """Get the first entity that fits the base_filter."""
+        ...
+
+    @classmethod
+    @abstractmethod
+    def create(
```
michael-s-molina, line 84: @villebro I wonder if we're exposing the correct layer as a public API... For example, when we expose create/update commands of the DAO layer, we allow MCP tools and extensions to modify these objects bypassing the logic defined at the commands layer. Using DELETE as an example, the `DeleteChart` command enforces a restriction where you can't delete a chart if there are associated alerts or reports. By exposing the delete method from the DAO layer, extensions or MCP tools could bypass this restriction. I wonder if the commands layer is the one that should be exposed instead. WDYT?

superset-core/src/superset_core/api/models.py
villebro: Yes - I initially considered reducing it, but it hit me that the C and U in CRUD will require having full property coverage, otherwise the objects may be incompatible with the metastore data model.

superset-core/src/superset_core/api/daos.py
villebro: Good question. As commands will invariably need to use these C, U and D methods, not exposing them in the base class will be slightly weird. But as we don't yet support custom models, this isn't a problem yet. So I'm ok removing these for now, and introducing them later when/if we start supporting custom models (which will require adding support for extension-specific db migrations).

michael-s-molina: I wonder if we need another API abstraction which is not commands or DAOs but that use them for the implementation. For example, the API would expose CRUD operations and delegate R operations to DAOs but CUD operations to commands. The method signatures could be the same as their implementations or slightly better (command APIs are bad).

In this case, the exposed APIs will always be a subset of the DAOs and commands APIs.

michael-s-molina: We discussed this topic and agreed that it's hard to determine the best approach now without evolving the work more and having more consumption examples. Let's revisit the design if needed once we have more pieces in place.

michael-s-molina (approved):

PR #35264, opened 2025-09-24
feat(frontend): add dataset cache clearing utilities and integration

msyavuz:
### SUMMARY
Implements dataset cache clearing functionality to ensure fresh data is fetched after dataset modifications. This addresses potential stale data issues where cached API responses were not being invalidated when datasets were updated.

**Key Changes:**
- Added `datasetCache.ts` utility with functions to clear cached API responses for specific datasets
- Integrated cache clearing in `SaveDatasetModal` after dataset updates  
- Integrated cache clearing in `DatasourceModal` and `DatasourceEditor` components
- Ensures that any subsequent API calls fetch fresh data from the server

N/A - This is a backend caching improvement with no visible UI changes

### TESTING INSTRUCTIONS
1. Navigate to SQL Lab and create/modify a dataset
2. Open the dataset in the datasource editor and make changes
3. Verify that the changes are immediately reflected without requiring a page refresh
4. Test saving datasets from SQL Lab and confirm fresh data is loaded
5. Check browser network tab to verify cache invalidation is working (no stale cached responses)

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/src/utils/datasetCache.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { supersetGetCache } from './cachedSupersetGet';
```
mistercrunch: Would keep all this stuff within `cachedSupersetGet` to keep things contained. I think the goal is ultimately to deprecate `cachedSupersetGet`. When debugging around this before there was some clear nonsense around this. Plus the dataset is already in redux, and there are subtle variations around the schema/endpoint to try and work around datasets with lots of columns. 

Client-side cache fighting with redux is super confusing.

superset-frontend/src/components/Datasource/components/DatasourceEditor/DatasourceEditor.jsx
```diff
@@ -837,6 +838,11 @@ class DatasourceEditor extends PureComponent {
           col => !col.expression, // remove calculated columns
         ),
       });
+
+      // Clear the dataset cache to ensure fresh data when fetching columns for filters
+      // This ensures that newly synced columns are immediately available in filter creation
```
geido:
```suggestion
```

superset-frontend/src/components/Datasource/DatasourceModal/index.tsx
```diff
@@ -203,6 +204,11 @@ const DatasourceModal: FunctionComponent<DatasourceModalProps> = ({
       const { json } = await SupersetClient.get({
         endpoint: `/api/v1/dataset/${currentDatasource?.id}`,
       });
+
+      // Clear the dataset cache to ensure fresh data when fetching columns for filters
+      // This ensures that changes to the dataset are immediately reflected
```
geido:
```suggestion
```

superset-frontend/src/SqlLab/components/SaveDatasetModal/index.tsx
```diff
@@ -169,6 +170,10 @@ const updateDataset = async (
     headers,
     body,
   });
+
+  // Clear the dataset cache after updating to ensure fresh data
```
geido:
```suggestion
```

superset-frontend/src/SqlLab/components/SaveDatasetModal/index.tsx
```diff
@@ -335,15 +340,18 @@ export const SaveDatasetModal = ({
         datasourceName: datasetName,
       }),
     )
-      .then((data: { id: number }) =>
-        postFormData(data.id, 'table', {
+      .then((data: { id: number }) => {
+        // Clear cache for the newly created dataset
```
geido:
```suggestion
```

superset-frontend/src/components/Datasource/components/DatasourceEditor/DatasourceEditor.jsx
geido: About this one in particular? Shouldn't we provide a way for the user to clear the cache when fetching columns rather than doing it automatically, so that we can actually leverage the cache?

msyavuz: Hey, this is to prevent stale data by clearing the cache when syncing metadata. Isn't this the expected behaviour? I think it would be confusing for users to have that option when it should be done by default

superset-frontend/src/utils/cachedSupersetGet.ts
```diff
@@ -27,3 +27,66 @@ export const cachedSupersetGet = cacheWrapper(
...
+export function clearAllDatasetCache(): void {
+  const keysToDelete: string[] = [];
+
+  supersetGetCache.forEach((_value, key) => {
+    if (key.includes('/api/v1/dataset/')) {
+      keysToDelete.push(key);
+    }
+  });
+
+  keysToDelete.forEach(key => supersetGetCache.delete(key));
```
kgabryje: nit: I think we could do it in 1 loop? According to Claude it's safe to delete map entries while iterating over it.

```
  for (const key of supersetGetCache.keys()) {
    if (key.includes('/api/v1/dataset/')) {
      supersetGetCache.delete(key);
    }
  }
```

msyavuz: Good catch! Fixed

kgabryje (approved):

PR #35265, opened 2025-09-24
fix(dashboard): refresh tabs as they load when dashboard is refreshed

msyavuz:
## SUMMARY

This PR fixes an issue where refreshing a dashboard does not update charts in tabs that haven't been loaded yet. Previously, only charts in tabs that were already visited would refresh, leaving unvisited tabs with stale data.

The fix implements lazy refresh behavior for tabs - when a dashboard refresh is triggered, tabs that haven't been loaded yet will automatically refresh their charts when the user switches to them.

### Before
- When refreshing a dashboard with multiple tabs, only the charts from tabs that were already loaded would be refreshed
- Charts in unvisited tabs would remain stale even after a dashboard refresh

### After  
- Dashboard refresh marks the refresh timestamp
- When switching to a previously unloaded tab after a refresh, that tab's charts are automatically refreshed
- This matches the initial loading behavior where tabs load lazily as the user navigates to them

## TESTING INSTRUCTIONS

1. Create a dashboard with 3 tabs, and add one chart on each tab
2. Wait a couple of minutes so data gets cached on all charts
3. Access the dashboard (loads on first tab) and navigate to the second tab
4. Trigger a dashboard refresh using the refresh button
5. Navigate to the third tab
6. Verify that the third tab's charts are refreshed when you switch to it (you should see loading indicators)

## ADDITIONAL INFORMATION

- Added tracking of tab activation times to determine when tabs need refreshing
- Added `lastRefreshTime` to dashboard state to track when refreshes occur
- Tab component now checks if it needs to refresh when becoming visible
- Created utility function `getChartIdsFromComponent` to get all chart IDs within a tab

Fixes the issue where dashboard refresh doesn't affect unloaded tabs.


superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
```diff
@@ -113,6 +114,49 @@ const renderDraggableContent = dropProps =>
...
+      }
+    }
+  }, [
+    props.isComponentVisible,
+    props.renderType,
+    props.id,
+    lastRefreshTime,
+    tabActivationTimes,
+    dashboardLayout,
+    dashboardInfo.id,
```
amaannawab923, line 155: A bit worried about the dependancy array, have seen that dependancy array might sometimes cause an infinite number of calls , especially , dispatch and dashboardlayout which can be a huge object

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
```diff
@@ -113,6 +114,49 @@ const renderDraggableContent = dropProps =>
...
+        tabActivationTime &&
+        lastRefreshTime > tabActivationTime
+      ) {
+        const chartIds = getChartIdsFromComponent(tabId, dashboardLayout);
+        if (chartIds.length > 0) {
+          // Small delay to ensure charts are fully mounted
+          setTimeout(() => {
+            // Refresh charts in this tab
+            dispatch(onRefresh(chartIds, true, 0, dashboardInfo.id));
+          }, 100);
```
amaannawab923: Should we keep this 100ms in a constnat something like Chart_Mount_Delay

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
```diff
@@ -113,6 +117,49 @@ const renderDraggableContent = dropProps =>
...
+  const lastRefreshTime = useSelector(
+    state => state.dashboardState.lastRefreshTime,
+  );
+  const tabActivationTimes = useSelector(
+    state => state.dashboardState.tabActivationTimes || {},
+  );
+  const dashboardInfo = useSelector(state => state.dashboardInfo);
+
+  // Check if tab needs refresh when it becomes visible
+  useEffect(() => {
```
geido, line 131: Can we add tests to capture this new behavior?

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
```diff
@@ -113,6 +117,49 @@ const renderDraggableContent = dropProps =>
...
+        tabActivationTime &&
+        lastRefreshTime > tabActivationTime
+      ) {
+        const chartIds = getChartIdsFromComponent(tabId, dashboardLayout);
+        if (chartIds.length > 0) {
+          // Small delay to ensure charts are fully mounted
+          setTimeout(() => {
+            // Refresh charts in this tab
+            dispatch(onRefresh(chartIds, true, 0, dashboardInfo.id));
+          }, CHART_MOUNT_DELAY);
```
geido: What's the guarantee that charts have been mounted at this point? What happens for dashboards that lazy-load their charts? Would they be affected?

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
```diff
@@ -114,6 +119,43 @@ const renderDraggableContent = dropProps =>
...
+        }
+      }
+    }
+  }, [
+    props.isComponentVisible,
+    props.renderType,
+    props.id,
+    lastRefreshTime,
+    tabActivationTime,
+    dashboardLayout,
```
EnxDev, line 154: Could we include only the tabChartIds instead of the entire dashboardLayout object?

const tabChartIds = useMemo(
  () => getChartIdsFromComponent(props.id, dashboardLayout),
  [props.id, dashboardLayout]
);

superset-frontend/src/dashboard/reducers/dashboardState.js
```diff
@@ -216,10 +227,17 @@ export default function dashboardStateReducer(state = {}, action) {
         .difference(new Set(action.activeTabs))
         .union(new Set(action.inactiveTabs));
 
+      // Track when each tab was last activated
+      const tabActivationTimes = { ...state.tabActivationTimes };
```
EnxDev, line 231: Should we add a cleanup here to clear old tab times and prevent a memory leak?

msyavuz: We are updating already existing tabActivationTimes, i don't think there is a cleanup to be done here. Can you give an example code snippet?

EnxDev: After that action.activeTabs.forEach(tabId => {  tabActivationTimes[tabId] = Date.now(); });
 (line 234),  but it probably doesn’t have much impact in our case.

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
msyavuz: This would not help. It still depends  on the dashboardLayout in your example. It's just moving the computation from inside the effect to outside, but with no performance gain.

EnxDev: You're right; this version still depends on `dashboardLayout`, so it wouldn't reduce the number of re-renders.
My intention was to make the dependency more focused on the actual chart IDs.
Maybe we could extract only the needed data from `dashboardLayout` (e.g., the chart IDs for the current tab) and make that part more stable instead?

msyavuz: We need chart id's of all tabs though

EnxDev: So in this case, the dependency optimization doesn’t really help.

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
msyavuz: fixed!

superset-frontend/src/dashboard/components/gridComponents/Tab/Tab.jsx
msyavuz: I don't see any issues from manual testing

EnxDev (approved):
LGTM, thanks!

PR #35274, opened 2025-09-24
fix(table): New ad-hoc columns retain the name of previous columns

### LINKED ISSUE #25592: Table chart : Name in calculated dimensions
When we create several calculated dimensions in a table chart aggregated, Superset keep the last name given, and not the default value 

#### How to reproduce the bug

1. Create table chart 'aggregated'
2. Create a dimension with custom_sql and Name it
3. Create another one 
4. By default it keep the name of the previous column, and not the value by default 
5. So the new column is not taken in account in the table, because they have the same name

### Expected results

Always have the default value when we create a calculated dimension 

### Actual results

By default it keep the name of the previous column, and not the value by default 

#### Screenshots

New calculated dimensions : 
image

### Environment

- browser type and version: Google Chrome Version 117.0.5938.149
- superset version: 2.1.0
- node.js version: `node -v`
- any feature flags active: Table Chart

### Checklist

SBIN2010:
This PR fix #25592

### SUMMARY

Reset only if the pop-up window is closed or has just opened.

BEFORE
Снимок экрана от 2025-09-25 01-14-27

AFTER
Снимок экрана от 2025-09-25 01-12-42

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
Can we add a test to prevent the bug from happening again?

SBIN2010:
> Can we add a test to prevent the bug from happening again?

Hi @michael-s-molina!
Yes, I've already thought about tests. Right now, I'm still in the process of writing them.

SBIN2010:
@michael-s-molina Tests added

superset-frontend/src/explore/components/controls/DndColumnSelectControl/ColumnSelectPopoverTrigger.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+  expressionType: 'SQL',
+};
+
+const secondAdhocColumn: AdhocColumn = {
+  label: 'Second Custom Column',
+  sqlExpression: 'SUM(amount)',
+  expressionType: 'SQL',
+};
+
+describe('ColumnSelectPopoverTriggerWrapper - Multiple Columns', () => {
```
michael-s-molina: Could you avoid using `describe` and `it` and use `test` instead to follow our [testing guidelines](https://github.com/apache/superset/wiki/Testing-Guidelines-and-Best-Practices#avoid-nesting-when-youre-testing)?

michael-s-molina (approved):
Thank you for the contribution and for adding tests!

PR #35277, opened 2025-09-25
fix(dataset): sort by database in Dataset and Saved queries Issue

rebenitez1802:
### SUMMARY

Fixes database column sorting in Dataset and SavedQuery list views. After the Ant Design v5 migration, nested columns (like database.database_name) were split into arrays for dataIndex, but when sorting occurs, AntD returns the array format instead of the original dot-notation string, causing backend validation errors.

  The fix converts array fields back to dot notation in the handleTableChange callback while preserving existing string fields unchanged.

### TESTING INSTRUCTIONS

  1. Navigate to Datasets page (/dataset/list/)
  2. Click on the "Database" column header to sort
  3. Verify sorting works without API errors
  4. Repeat for Saved Queries page (/savedqueryview/list/)
  5. Test sorting other columns to ensure no regressions
  
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { render, screen } from '@superset-ui/core/spec';
+import { render, screen, fireEvent } from '@superset-ui/core/spec';
 import { renderHook } from '@testing-library/react-hooks';
 import { TableInstance, useTable } from 'react-table';
 import TableCollection from '.';
@@ -207,7 +207,7 @@ test('Bulk selection should work with pagination', () => {
   expect(checkboxes.length).toBeGreaterThan(0);
 });
 
-test('handleTableChange should convert array field to dot notation for nested fields', () => {
+test('should call setSortBy when clicking sortable column header', () => {
   const setSortBy = jest.fn();
   const sortingProps = {
     ...defaultProps,
@@ -216,55 +216,25 @@ test('handleTableChange should convert array field to dot notation for nested fi
 
   render(<TableCollection {...sortingProps} />);
 
-  expect(screen.getByRole('table')).toBeInTheDocument();
+  const columnHeaders = screen.getAllByRole('columnheader');
+  expect(columnHeaders.length).toBeGreaterThan(0);
 
-  const mockSorter = {
-    field: ['database', 'database_name'], // Array format from AntD
-    order: 'descend' as const,
-  };
+  const firstColumnHeader = columnHeaders[0];
+  expect(firstColumnHeader).toBeInTheDocument();
 
-  const handleTableChange = (
-    _pagination: any,
-    _filters: any,
-    sorter: { field: string | string[]; order: 'ascend' | 'descend' },
-  ) => {
-    if (sorter && sorter.field) {
-      const fieldId = Array.isArray(sorter.field)
-        ? sorter.field.join('.')
-        : sorter.field;
+  // Click on the column header to trigger sorting
+  fireEvent.click(firstColumnHeader);
 
-      setSortBy([
-        {
-          id: fieldId,
-          desc: sorter.order === 'descend',
-        },
-      ]);
-    }
-  };
+  // Verify setSortBy was called immediately
+  expect(setSortBy).toHaveBeenCalled();
 
-  // Test the callback logic with array field
-  handleTableChange(null, null, mockSorter);
+  const sortCallArgs = setSortBy.mock.calls[0][0];
+  expect(Array.isArray(sortCallArgs)).toBe(true);
+  expect(sortCallArgs[0]).toHaveProperty('id');
+  expect(sortCallArgs[0]).toHaveProperty('desc');
 
-  expect(setSortBy).toHaveBeenCalledWith([
-    {
-      id: 'database.database_name', // Should be converted to dot notation
-      desc: true,
-    },
-  ]);
-
-  // Test with string field (should pass through unchanged)
-  setSortBy.mockClear();
-  const mockStringSorter = {
-    field: 'table_name',
-    order: 'ascend' as const,
-  };
-
-  handleTableChange(null, null, mockStringSorter);
-
-  expect(setSortBy).toHaveBeenCalledWith([
-    {
-      id: 'table_name',
-      desc: false,
-    },
-  ]);
+  // Verify it was called with a valid column ID (any string) and boolean desc
+  expect(typeof sortCallArgs[0].id).toBe('string');
+  expect(sortCallArgs[0].id.length).toBeGreaterThan(0);
+  expect(typeof sortCallArgs[0].desc).toBe('boolean');
 });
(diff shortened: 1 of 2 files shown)


superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +206,67 @@ test('Bulk selection should work with pagination', () => {
...
+  expect(table).toBeInTheDocument();
+
+  // Simulate the handleTableChange function behavior directly
+  // This tests the logic without requiring actual table interaction
+  const mockSorter = {
+    field: ['database', 'database_name'], // Array format from AntD
+    order: 'descend',
+  };
+
+  // Get the component instance to test the callback logic
```
msyavuz:
```suggestion
```

msyavuz (approved):
LGTM with a small nit. Thanks for fixing this!

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
msyavuz: Also we might be mocking too much

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +206,67 @@ test('Bulk selection should work with pagination', () => {
...
+  const setSortBy = jest.fn();
+  const sortingProps = {
+    ...defaultProps,
+    setSortBy,
+  };
+
+  const component = render(<TableCollection {...sortingProps} />);
+
+  // Get the Table component instance
+  const table = component.container.querySelector('.ant-table');
```
geido: Can we use roles wherever possible? See [wiki](https://github.com/apache/superset/wiki/Testing-Guidelines-and-Best-Practices#be-specific-when-using-byrole)

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +206,67 @@ test('Bulk selection should work with pagination', () => {
...
+
+  // Simulate the handleTableChange function behavior directly
+  // This tests the logic without requiring actual table interaction
+  const mockSorter = {
+    field: ['database', 'database_name'], // Array format from AntD
+    order: 'descend',
+  };
+
+  // Get the component instance to test the callback logic
+  const handleTableChange = (_pagination: any, _filters: any, sorter: any) => {
```
geido:
```suggestion
  const handleTableChange = (_pagination: any, _filters: any, sorter: SorterResult) => {
```

rebenitez1802:
Test refactored please review again @geido or @msyavuz . Thanks for the comments.

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +206,38 @@ test('Bulk selection should work with pagination', () => {
...
+    : mockArraySorter.field;
+
+  expect(fieldIdArray).toBe('database.database_name');
+
+  // Test string field passthrough
+  const fieldIdString = Array.isArray(mockStringSorter.field)
+    ? mockStringSorter.field.join('.')
+    : mockStringSorter.field;
+
+  expect(fieldIdString).toBe('table_name');
```
geido: I am confused about this change. Isn't this just testing the `fieldIdString` logic instead of testing the component behavior?

geido (changes requested):

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
msyavuz: Yeah this just implements the logic in the component again. If the component logic changes for any reason this test will still pass as it does not interact with the component.

msyavuz (changes requested):

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +206,65 @@ test('Bulk selection should work with pagination', () => {
...
+        : sorter.field;
+
+      setSortBy([
+        {
+          id: fieldId,
+          desc: sorter.order === 'descend',
+        },
+      ]);
+    }
+  };
```
msyavuz: This does the `implement same thing in test and test that` again (just one level above) and still doesn't interact with the component

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +206,35 @@ test('Bulk selection should work with pagination', () => {
...
+  render(<TableCollection {...sortingProps} />);
+
+  const columnHeaders = screen.getAllByRole('columnheader');
+  expect(columnHeaders.length).toBeGreaterThan(0);
+
+  const firstColumnHeader = columnHeaders[0];
+  expect(firstColumnHeader).toBeInTheDocument();
+
+  // Click on the column header to trigger sorting
+  fireEvent.click(firstColumnHeader);
```
msyavuz: I remember a discussion about using `userEvent` vs `fireEvent` but i think this is fine.

msyavuz (commented):
Test structure is correct now but it tests the wrong column header. The pr fixes sort by database.

rebenitez1802:
@msyavuz I added a generic value to make sure we cover the case scenario that this PR is fixing, I wouldn't use database column because this is a shared component not  reserved for Dataset List, but the value I added follows the same pattern.

superset-frontend/packages/superset-ui-core/src/components/TableCollection/TableCollection.test.tsx
```diff
@@ -206,3 +215,32 @@ test('Bulk selection should work with pagination', () => {
...
+  expect(setSortBy).toHaveBeenCalled();
+
+  const sortCallArgs = setSortBy.mock.calls[0][0];
+  expect(Array.isArray(sortCallArgs)).toBe(true);
+  expect(sortCallArgs[0]).toHaveProperty('id');
+  expect(sortCallArgs[0]).toHaveProperty('desc');
+
+  // Verify the nested field array was converted to dot notation
+  expect(sortCallArgs[0].id).toBe('parent.child');
+  expect(typeof sortCallArgs[0].desc).toBe('boolean');
```
msyavuz: Can we use `toHaveBeenCalledWith` here? This is mostly a nit though and we are on the same page about tests now

rebenitez1802: fixed

geido (approved):

geido:
@msyavuz does this look good to you now?

msyavuz (approved):
Yes, LGTM!

PR #35294, opened 2025-09-25
fix: Enable DuckDB examples loading in showtime ephemeral environments

eschutho:
## Summary

This PR fixes two critical issues preventing DuckDB examples from loading in showtime ephemeral environments:

1. **Docker ARG scope issue**: `LOAD_EXAMPLES_DUCKDB` build argument was being ignored
2. **DuckDB version compatibility**: Newer DuckDB versions (1.x) cause `unhashable type: '_duckdb.typing.DuckDBPyType'` errors

### 1. Docker Build Argument Scope
**Issue**: The `LOAD_EXAMPLES_DUCKDB` ARG was declared inside the `python-common` stage, making it unavailable when passed via `--build-arg` to the `showtime` target in multi-stage builds.

**Fix**: Move ARG declaration to global scope and re-declare in the stage where it's used.

### 2. DuckDB Version Compatibility  
**Issue**: Current DuckDB version `1.3.2` has type system changes incompatible with duckdb-engine, causing connection failures.

**Fix**: Constrain DuckDB core to `>=0.10.2,<0.11` while keeping duckdb-engine at current `>=0.17.0`. Testing confirmed the newer driver works fine with DuckDB 0.10.x.

## Version Constraints

```python
# DuckDB 1.x has type system incompatibilities with duckdb-engine.
duckdb = ["duckdb>=0.10.2,<0.11", "duckdb-engine>=0.17.0"]
```

**Rationale**:
- Uses **version ranges** (not pinned) to allow patch updates
- Only constrains what's broken (duckdb core 1.x)
- Keeps duckdb-engine at current version (tested and working)
- Matches approach used in motherduck configuration

## Testing

✅ **Local testing confirms both fixes work**:
- Docker ARG is properly passed through build stages
- DuckDB 0.10.2 + duckdb-engine 0.17.0 connections work without errors
- File-based DuckDB connections work correctly

```bash
# Tested combination
pip install duckdb==0.10.2 duckdb-engine==0.17.0
# Result: ✅ All connection types work
```

## Before/After

**Before**: 
- Showtime builds ignored `LOAD_EXAMPLES_DUCKDB=true` → No examples file downloaded
- DuckDB connections failed with type errors → No examples accessible

**After**:
- Docker properly downloads examples.duckdb during build
- DuckDB connections work correctly  
- Showtime environments can load examples successfully

## Impact

This enables showtime ephemeral environments to successfully load and display examples, making them much more useful for testing and demonstration purposes.


pyproject.toml
```diff
@@ -133,7 +133,7 @@ denodo = ["denodo-sqlalchemy~=1.0.6"]
 dremio = ["sqlalchemy-dremio>=1.2.1, <4"]
 drill = ["sqlalchemy-drill>=1.1.4, <2"]
 druid = ["pydruid>=0.6.5,<0.7"]
-duckdb = ["duckdb-engine>=0.17.0"]
+duckdb = ["duckdb>=0.10.2,<0.11", "duckdb-engine>=0.12.1,<0.13"]
```
eschutho: I'm still testing out these versions.

eschutho: latest change from Claude:
│   Testing confirmed that duckdb-engine>=0.17.0 works correctly with                                                                                 │
│   duckdb 0.10.2. The issue is only with newer duckdb versions (1.3.x),                                                                              │
│   not with the SQLAlchemy driver.                                                                                                                   │
│                                                                                                                                                     │
│   This more targeted fix:                                                                                                                           │
│   - Constrains duckdb to >=0.10.2,<0.11 (avoids 1.3.x compatibility issues)                                                                         │
│   - Keeps duckdb-engine at >=0.17.0 (current version, works fine)

eschutho:
### Why Pin DuckDB to <0.11?

During testing of showtime ephemeral environments, we discovered that newer DuckDB versions (1.3.x) cause connection failures with the error:

```
[SQL: unhashable type: '_duckdb.typing.DuckDBPyType']
```

### Root Cause

The DuckDB core library versions ≥1.0 introduced type system changes that are incompatible with Superset's DuckDB integration code. The error occurs when trying to establish any DuckDB connection (even basic `:memory:` connections).

### Testing

✅ **Confirmed working**: `duckdb==0.10.2` + `duckdb-engine>=0.17.0`
❌ **Confirmed broken**: `duckdb==1.3.2` + `duckdb-engine>=0.17.0`

### Version Constraint Rationale

- **`duckdb>=0.10.2,<0.11`**: Stay within the 0.10.x series that we've tested and confirmed works
- **`duckdb-engine>=0.17.0`**: Keep the current driver version (no issues with newer driver)

This matches the approach used in the `motherduck` configuration which also pins `duckdb==0.10.2` to avoid these compatibility issues.

mistercrunch (approved):
LGTM

eschutho:
Examples are showing up in the ephemeral now. 🙌

PR #35297, opened 2025-09-25
fix(mysql): render TINYINT(1) columns as boolean in SQL Lab and charts

sadpandajoe:
### SUMMARY
  **Root Cause**

  MySQL boolean fields were incorrectly displaying as numeric icons with raw 0/1 values instead of readable True/False text in SQL Lab and Explore views. The issue had two layers:

  1. Type Mapping: TINYINT(1) columns were mapped to GenericDataType.NUMERIC instead of GenericDataType.BOOLEAN
  2. Value Conversion Bug: bool("0") and bool(b"0") return True in Python instead of the expected False

  **Solution**

  Two-layer fix targeting MySQL's TINYINT(1) boolean representation:

  1. Schema-level: Added precise pattern matching for TINYINT(1) → GenericDataType.BOOLEAN
  2. Runtime-level: Enhanced fetch_data() with normalization logic for strings/bytes/Decimal before boolean conversion

  **Changes**

  - superset/db_engine_specs/mysql.py: Added boolean type mapping and conversion logic
  - tests/unit_tests/db_engine_specs/test_mysql.py: Comprehensive test coverage for edge cases

### TESTING INSTRUCTIONS

  - ✅ All existing MySQL tests pass
  - ✅ New parametrized tests cover string/bytes/Decimal conversion scenarios
  - ✅ Preserves behavior for non-boolean TINYINT variants (TINYINT(2), etc.)
  - ✅ End-to-end pipeline validation with pandas boolean inference

### ADDITIONAL INFORMATION

- [X] Has associated issue: Fixes: #35166
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/db_engine_specs/mysql.py
+++ b/superset/db_engine_specs/mysql.py
@@ -320,7 +320,14 @@ class MySQLEngineSpec(BasicParametersMixin, BaseEngineSpec):
                 new_row = list(row)
                 for col_idx in bool_column_indices:
                     if new_row[col_idx] is not None:
-                        new_row[col_idx] = bool(new_row[col_idx])
+                        # Normalize different value types before boolean conversion
+                        # bool("0") returns True, but we need False for MySQL boolean
+                        value = new_row[col_idx]
+                        if isinstance(value, (str, bytes)):
+                            value = int(value)
+                        elif isinstance(value, Decimal):
+                            value = int(value)
+                        new_row[col_idx] = bool(value)
                 converted_data.append(tuple(new_row))
             return converted_data
 
--- a/superset/db_engine_specs/mysql.py
+++ b/superset/db_engine_specs/mysql.py
@@ -77,6 +77,18 @@ class MySQLEngineSpec(BasicParametersMixin, BaseEngineSpec):
     supports_multivalues_insert = True
 
     column_type_mappings = (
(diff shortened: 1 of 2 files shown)


superset/db_engine_specs/mysql.py
```diff
@@ -260,6 +272,67 @@ def get_datatype(cls, type_code: Any) -> Optional[str]:
...
+                for col_idx in bool_column_indices:
+                    if new_row[col_idx] is not None:
+                        # Normalize different value types before boolean conversion
+                        # bool("0") returns True, but we need False for MySQL boolean
+                        value = new_row[col_idx]
+                        if isinstance(value, (str, bytes)):
+                            value = int(value)
+                        elif isinstance(value, Decimal):
+                            value = int(value)
+                        new_row[col_idx] = bool(value)
```
mistercrunch, line 330: I think this is a perfect case for `column_type_mutators`, see https://github.com/apache/superset/blob/master/superset/db_engine_specs/base.py#L342-L344 and https://github.com/apache/superset/blob/master/superset/db_engine_specs/base.py#L993-L998

mistercrunch: Unless @betodealmeida can think of a better way

sadpandajoe:
closing for now as it's pretty stale. will work on it again later if priorities change.

PR #35308, opened 2025-09-26
chore(superset-core): move ui to new core

msyavuz:
### SUMMARY

As part of the extensions work, ui components are moved to the new core package. In this pr i am moving the required parts for ui components to work in new package and moving Alert component as a proof of concept.

### TESTING INSTRUCTIONS

All Alert usages should work correctly.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/packages/superset-core/src/spec/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * under the License.
+ */
+import userEvent from '@testing-library/user-event';
+import { ReactElement } from 'react';
+import { render, RenderOptions } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { themeObject } from '../ui/theme';
+
+// Define the wrapper component outside
+const AllTheProviders = ({ children }: { children: React.ReactNode }) => (
```
michael-s-molina: I think we can rename this to Providers.

superset-frontend/packages/superset-core/src/spec/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import userEvent from '@testing-library/user-event';
+import { ReactElement } from 'react';
+import { render, RenderOptions } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { themeObject } from '../ui/theme';
```
michael-s-molina: Given that this utility is used for testing UI-related components, I suggest moving it inside the `ui` package and renaming it.

```
src/ui/testing.ts
```

superset-frontend/packages/superset-ui-core/src/chart/components/Matrixify/MatrixifyGridCell.tsx
```diff
@@ -18,7 +18,7 @@
  */
 
 import { memo, useMemo } from 'react';
-import { styled, useTheme } from '../../../theme';
+import { styled, useTheme } from '@apache-superset/core/src/ui/themeet/core/src/ui/theme';
```
michael-s-molina: Let's think about exposing these objects at the root level of the `ui` folder. Something like:

```
import { styled, useTheme } from '@apache-superset/core/ui';
```

superset-frontend/packages/superset-core/src/ui/components/Alert/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ...rest
+  } = props;
+  return (
+    <AntdAlert
+      role="alert"
+      aria-live={type === 'error' ? 'assertive' : 'polite'}
+      type={type}
+      showIcon={showIcon}
+      closable={closable}
+      message={children || 'Default message'}
```
michael-s-molina, line 77: We'll need the `t` function because probably the components are using it.
```suggestion
      message={children || t('Default message')}
```

superset-frontend/packages/superset-core/src/ui/components/Alert/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Alert component for displaying important messages to users.
+ *
+ * Wraps Ant Design's Alert component with sensible defaults and improved accessibility.
+ * Supports four severity levels: success, info, warning, and error.
+ *
+ * @param props - Alert configuration props
+ *
+ * @example
```
michael-s-molina, line 37: Great example section!

superset-frontend/packages/superset-ui-core/src/chart/components/Matrixify/MatrixifyGridCell.tsx
msyavuz: They should be already exposed, i will fix the imports from typescript language server action

superset-frontend/packages/superset-ui-core/src/components/index.ts
```diff
@@ -28,7 +28,7 @@ export {
   ConfirmStatusChange,
   type ConfirmStatusChangeProps,
 } from './ConfirmStatusChange';
-export { Alert, type AlertProps } from './Alert';
+export { Alert, type AlertProps } from '@apache-superset/core/ui';
```
msyavuz: The question here is do we do this for now or remove this entirely and import from new core everywhere else

michael-s-molina: I think if we move one component at a time, you could update the imports everywhere. If you prefer to keep re-exporting the component from `@superset-ui/core` and later open a PR just to fix the imports, that also works.

superset-frontend/packages/superset-core/package.json
```diff
@@ -10,6 +10,15 @@
   ],
   "author": "",
   "license": "ISC",
+  "dependencies": {
+    "@emotion/cache": "^11.4.0",
```
michael-s-molina: Shouldn't these be peer dependencies?

superset-frontend/src/components/Datasource/ChangeDatasourceModal/index.tsx
```diff
@@ -285,7 +285,6 @@ const ChangeDatasourceModal: FunctionComponent<ChangeDatasourceModalProps> = ({
         {!confirmChange && (
           <>
             <Alert
-              roomBelow
```
michael-s-molina, line 288: Is this change intended?

superset-frontend/src/theme/ThemeController.ts
```diff
@@ -26,7 +26,7 @@ import {
   ThemeMode,
   themeObject as supersetThemeObject,
 } from '@superset-ui/core';
-import { normalizeThemeConfig } from '@superset-ui/core/theme/utils';
+import { normalizeThemeConfig } from '@apache-superset/core/ui/theme';
```
michael-s-molina: It seems this function could be exported from `@apache-superset/core/ui` as it seems generic enough.

superset-frontend/src/components/Datasource/ChangeDatasourceModal/index.tsx
msyavuz: Yes, this was used in a single component and styling like this did not seem right for a single usage

msyavuz: Also it wasn't even used in the component at all, just for stories

superset-frontend/packages/superset-core/package.json
```diff
@@ -19,11 +24,29 @@
...
+    "@emotion/styled": "^11.14.1",
+    "@types/lodash": "^4.17.20"
   },
   "peerDependencies": {
     "antd": "^5.24.9",
-    "react": "^17.0.2"
+    "@emotion/cache": "^11.4.0",
+    "@emotion/react": "^11.4.1",
+    "@emotion/styled": "^11.14.1",
+    "@testing-library/dom": "^8.20.1",
```
michael-s-molina, line 24: We should move `@testing-library` and `@types` packages to  `devDependencies`.

superset-frontend/packages/superset-core/package.json
```diff
@@ -10,6 +10,11 @@
   ],
   "author": "",
   "license": "ISC",
+  "dependencies": {
```
michael-s-molina: Let's move these to `peerDependencies`.

michael-s-molina (commented):
@msyavuz Can you use AI to remove `describe` in favor of `test` in test files inside `superset-core`?

Also, we have `src/ui/theme/utils.ts` but also a `src/ui/theme/utils` folder. Can we merge these?

michael-s-molina (approved):
Thank you for the PR @msyavuz and for addressing all comments!

kgabryje (approved):
LGTM

villebro (approved):
FANTASTIC WORK! Super excited about this moving into `@apache-superset/core`! ❤️

PR #35321, opened 2025-09-29
fix(theming): CRUD view padding

msyavuz:
## SUMMARY

Added consistent padding for a cleaner appearance.

## BEFORE/AFTER

**Before:**
- Inconsistent spacing around table content
charts-crud-view-BEFORE-fix

**After:**  
- Consistent padding using theme spacing tokens
image

## TESTING INSTRUCTIONS

1. Navigate to any CRUD view (Charts, Dashboards, Datasets, etc.)
2. Check that spacing around the table content is consistent
3. Test with different themes to ensure the fix works across all theme variants

## ADDITIONAL INFORMATION

This change aligns with the ongoing Ant Design theming migration by:
- Applying consistent spacing using theme padding tokens

No breaking changes - purely visual improvement.


geido:
I am not sure about this change. @kasiazjc have you looked at this? Now the between the page title and the filters there is a gap of a different background color. Also, I tend to think the separation was actually good. This is not pretty plain and harder to focus on

geido:
Thanks for reverting that change!

geido (approved):

PR #35333, opened 2025-09-29
fix: sqlglot dialect for Pinot

### LINKED ISSUE #35270: [6.0.0-RC2] Superset+Pinot: Error with escape field names
### Bug description

I have an github project with superset + pinot, where I configure the dashboard USA Births Names to use with Pinot:
https://github.com/jacob-roldan/issues-superset6-pinot

One error is in the chart "Girl Name Cloud PINOT"
This is the query:
```sql
SELECT "name", SUM(num) 
FROM birth_names 
WHERE ds1 >= '1925-09-24 00:00:00.000000' AND ds1 < '2025-09-24 17:30:50.000000' AND gender = 'girl' GROUP BY "name"
 LIMIT 50000;
```
 

Where the field _name_ is escaped, but the result is wrong:

Image

That query executed with dBeaver is OK:

Image

The step to reproduce it using that git-repo:

```sh
//Start containers
docker-compose up -d

//Install driver for Pinot
./superset-config-pinot.sh

//Load data and import dashboard
./example_birthnames.sh

```

### Screenshots/recordings

_No response_

### Superset version

master / latest-dev

### Python version

3.9

### Node version

16

### Browser

Chrome

### Additional context

With version 5.0.0, there is a problem with Pinot driver. It's not possible to configure a connection to Pinot.
With version 4.1.4 all charts of USA Births Names dashboard are OK except one

Image

### Checklist

betodealmeida:
### SUMMARY

Pinot uses Calcite to parse SQL queries, specifically the `MYSQL_ANSI` dialect. Because of this, in Superset we are currently using the "mysql" dialect from `sqlglot`. The problem is that `MYSQL_ANSI` uses double quotes for identifiers, while the default MySQL dialect (in `sqlglot` and in the actual database) uses double quotes for string literals.

This PR introduces a small `sqlglot` dialect for Pinot, which is essentially the MySQL dialect but using double quotes for identifiers. That seems to be the only difference between Calcite's `MYSQL` and `MYSQL_ANSI`: https://calcite.apache.org/javadocAggregate/org/apache/calcite/config/Lex.html#MYSQL_ANSI

### ADDITIONAL INFORMATION

- [X] Has associated issue: Fixes #35270 
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/sql/dialects/pinot.py
+++ b/superset/sql/dialects/pinot.py
@@ -33,6 +33,8 @@ class Pinot(MySQL):
 
     The main difference from standard MySQL is that double quotes (") are used for
     identifiers instead of string literals, following ANSI SQL conventions.
+
+    See: https://calcite.apache.org/javadocAggregate/org/apache/calcite/config/Lex.html#MYSQL_ANSI
     """
 
     class Tokenizer(MySQL.Tokenizer):
--- a/superset/sql/dialects/pinot.py
+++ b/superset/sql/dialects/pinot.py
@@ -15,8 +15,27 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from .dremio import Dremio
-from .firebolt import Firebolt, FireboltOld
-from .pinot import Pinot
+"""
+MySQL ANSI dialect for Apache Pinot.
 
-__all__ = ["Dremio", "Firebolt", "FireboltOld", "Pinot"]
+This dialect is based on MySQL but follows ANSI SQL quoting conventions where
+double quotes are used for identifiers instead of string literals.
+"""
+
+from __future__ import annotations
+
+from sqlglot.dialects.mysql import MySQL
+
+
+class Pinot(MySQL):
+    """
+    MySQL ANSI dialect used by Apache Pinot.
+
+    The main difference from standard MySQL is that double quotes (") are used for
+    identifiers instead of string literals, following ANSI SQL conventions.
+    """
+
+    class Tokenizer(MySQL.Tokenizer):
+        QUOTES = ["'"]  # Only single quotes for strings
+        IDENTIFIERS = ['"', "`"]  # Backticks and double quotes for identifiers
+        STRING_ESCAPES = ["'", "\\"]  # Remove double quote from string escapes
--- a/tests/unit_tests/sql/dialects/pinot_tests.py
+++ b/tests/unit_tests/sql/dialects/pinot_tests.py
@@ -21,6 +21,14 @@ import sqlglot
 from superset.sql.dialects.pinot import Pinot
 
 
+def test_pinot_dialect_registered() -> None:
+    """Test that Pinot dialect is properly registered."""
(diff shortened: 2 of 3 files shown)


superset/sql/dialects/pinot.py
```diff
@@ -0,0 +1,41 @@
...
+    MySQL ANSI dialect used by Apache Pinot.
+
+    The main difference from standard MySQL is that double quotes (") are used for
+    identifiers instead of string literals, following ANSI SQL conventions.
+    """
+
+    class Tokenizer(MySQL.Tokenizer):
+        QUOTES = ["'"]  # Only single quotes for strings
+        IDENTIFIERS = ['"', "`"]  # Backticks and double quotes for identifiers
+        STRING_ESCAPES = ["'", "\\"]  # Remove double quote from string escapes
```
aminghadersohi, line 43: nit: should these have type hints? 
```
class Tokenizer(MySQL.Tokenizer):
      QUOTES: list[str] = ["'"]
      IDENTIFIERS: list[str] = ['"', "`"]
      STRING_ESCAPES: list[str] = ["'", "\\"]
```

tests/unit_tests/sql/dialects/pinot_tests.py
```diff
@@ -0,0 +1,340 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+import sqlglot
+
+from superset.sql.dialects.pinot import Pinot
+
+
```
aminghadersohi, line 23: do we cover the test that the dialect properly registered? 
```
  def test_pinot_dialect_registered() -> None:
      """Test that Pinot dialect is properly registered."""
      from superset.sql.parse import SQLGLOT_DIALECTS
      assert "pinot" in SQLGLOT_DIALECTS
      assert SQLGLOT_DIALECTS["pinot"] == Pinot
```

superset/sql/dialects/pinot.py
```diff
@@ -0,0 +1,41 @@
...
+
+from sqlglot.dialects.mysql import MySQL
+
+
+class Pinot(MySQL):
+    """
+    MySQL ANSI dialect used by Apache Pinot.
+
+    The main difference from standard MySQL is that double quotes (") are used for
+    identifiers instead of string literals, following ANSI SQL conventions.
```
aminghadersohi, line 35: should we add a link?

  See: https://calcite.apache.org/javadocAggregate/org/apache/calcite/config/Lex.html#MYSQL_ANSI

aminghadersohi (approved):
approved with some non blocking comments

superset/sql/dialects/pinot.py
betodealmeida: It's actually more idiomatic to not annotate local variables when the type is obvious; so, for example, this is OK:

```python
a: list[str] = []
```

But not this:

```python
a: list[str] = ["b"]
```

eschutho (approved):

PR #35338, opened 2025-09-30
fix(tables): Dark mode scrollbar styles for webkit

msyavuz:
### SUMMARY

Scrollbar styles for chart components did not apply on chrome on dark mode. This pr fixes that by specifying scrollbar styles for webkit

Before:
image

After:
image

### TESTING INSTRUCTIONS

Visual testing of table charts in dark mode

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [x] Changes UI

### DIFF
--- a/superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
@@ -170,8 +170,8 @@ const StyledTable = styled(AntTable as FC<AntTableProps>)<{ height?: number }>(
       
       /* Chrome/Safari/Edge webkit scrollbar styling */
       &::-webkit-scrollbar {
-        width: 8px;
-        height: 8px;
+        width: ${theme.sizeUnit * 2}px;
+        height: ${theme.sizeUnit * 2}px;
       }
       
       &::-webkit-scrollbar-track {
@@ -180,7 +180,7 @@ const StyledTable = styled(AntTable as FC<AntTableProps>)<{ height?: number }>(
       
       &::-webkit-scrollbar-thumb {
         background: ${theme.colorFillSecondary};
-        border-radius: 4px;
+        border-radius: ${theme.sizeUnit}px;
         
         &:hover {
           background: ${theme.colorFillTertiary};
--- a/superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
+++ b/superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
@@ -167,6 +167,33 @@ const StyledTable = styled(AntTable as FC<AntTableProps>)<{ height?: number }>(
     .ant-table-body {
       overflow: auto;
       height: ${height ? `${height}px` : undefined};
+      
+      /* Chrome/Safari/Edge webkit scrollbar styling */
+      &::-webkit-scrollbar {
+        width: 8px;
+        height: 8px;
+      }
+      
+      &::-webkit-scrollbar-track {
+        background: ${theme.colorFillQuaternary};
+      }
+      
+      &::-webkit-scrollbar-thumb {
+        background: ${theme.colorFillSecondary};
+        border-radius: 4px;
+        
+        &:hover {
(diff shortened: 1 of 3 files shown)


superset-frontend/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -55,9 +55,38 @@ const Styles = styled.div<PivotTableStylesProps>`
...
-  overflow: auto;
+  ${({ theme }) => `
+    height: 100%;
+    max-width: inherit;
+    overflow: auto;
+
+    /* Chrome/Safari/Edge webkit scrollbar styling */
+    &::-webkit-scrollbar {
+      width: ${theme.sizeUnit * 2}px;
+      height: ${theme.sizeUnit * 2}px;
```
geido: I think we should not be using `sizeUnit` for widths and heights but only for spacing.

superset-frontend/plugins/plugin-chart-table/src/DataTable/hooks/useSticky.tsx
```diff
@@ -221,6 +224,26 @@ function StickyWrap({
...
+    &::-webkit-scrollbar {
+      width: 8px;
+      height: 8px;
+    }
+    &::-webkit-scrollbar-track {
+      background: ${theme.colorFillQuaternary};
+    }
+    &::-webkit-scrollbar-thumb {
+      background: ${theme.colorFillSecondary};
+      border-radius: 4px;
```
geido: Can we use the theme provided border radius here?

superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
```diff
@@ -167,6 +167,33 @@ const StyledTable = styled(AntTable as FC<AntTableProps>)<{ height?: number }>(
     .ant-table-body {
       overflow: auto;
       height: ${height ? `${height}px` : undefined};
+
+      /* Chrome/Safari/Edge webkit scrollbar styling */
```
geido, line 171: Sorry to be a pain here but these styles look very similar across the implementation. Can we put them in a re-usable place? That should help with consistency as well.

msyavuz: I thought about it but the components are too far apart so the only logical place that makes sense is `GlobalStyles` inside `superset-ui-core` but i didn't want to do that for tables.

superset-frontend/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -55,9 +55,38 @@ const Styles = styled.div<PivotTableStylesProps>`
...
-  max-width: inherit;
-  overflow: auto;
+  ${({ theme }) => `
+    height: 100%;
+    max-width: inherit;
+    overflow: auto;
+
+    /* Chrome/Safari/Edge webkit scrollbar styling */
+    &::-webkit-scrollbar {
+      width: 8px;
```
alexandrusoare, line 65: Can we use theme.sizeUnit instead?

msyavuz: Suggestion from Diego on doing this https://github.com/apache/superset/pull/35338#discussion_r2394904315

superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
kgabryje: Maybe we could export those styles from superset-ui-core? Like `export const webkitTableScrollbarStyles = css...` and then reuse it?

geido (approved):

superset-frontend/packages/superset-ui-core/src/components/Table/index.tsx
geido: Let's talk more about how we should have re-usable pieces for plugins

PR #35345, opened 2025-09-30
fix(auth): redirect anonymous attempts to view dashboard with next

### LINKED ISSUE #35344: Redirect flow for not-logged-in users attempting to view a dashboard has broken
### Bug description

The enhancement previously merged in #30380 was broken in two places, as far as I can tell:
1. A [PR last week](https://github.com/apache/superset/pull/35237/files#diff-76c6a0b5c14a297c62ac3207416870471affdb153c5ae12b96d7b7da9304e9ddL797-R772) now just throws a 404 if a user doesn't have access to a dashboard. If not logged in, a login should be attempted first, passing the target URL as a `next` parameter.
2. The massive theming PR #31590 moved Login CRUD functionality into Superset, instead of relying on FAB. That may be what caused the `next` URL parameter to be lost during the roundtrip to the SSO provider.

The prior behavior should be restored: if an anonymous user tries to view a dashboard (whether or not it exists), they should be prompted to log in. Upon completing login, then the dashboard access check is performed and they are granted or denied as normal.

### Superset version

master / latest-dev, I branched off it this morning

### Browser

Firefox

### Checklist

## Screenshot

This is from 6.0.0rc2:

Image

I've moused-over the Login button when taking this screenshot. You can see that the Login button link appears in the bottom-left of the screenshot. Note that it's missing the `/?next=https:// ...` that was present in the URL, and thus that redirect path will be lost in the process of logging in.

sfirke:
### SUMMARY
Fixes #35344. Restores functionality from #30380 while retaining the security improvement made in [#35237](https://github.com/apache/superset/pull/35237): not leaking the existence of a dashboard the user doesn't have access to.

This may be a minor improvement on #35237, as that uses `Response(status=404)` and returns a bare 404 page. Now it uniformly uses `abort(404)` to return a Superset-formatted 404 page in both cases.

I tested these workflows:
1. Public dashboards remain visible to unauthenticated users
2. Anonymous (unauthenticated) users are always redirected to login when entering a dashboard URL, regardless of whether it exists or they have access. The `next` parameter is used to pass this redirect through the auth process.
3. Authenticated users receive the same Superset-branded 404 page whether they try to visit a dashboard they don't have access to or a dashboard that doesn't exist.

### TESTING INSTRUCTIONS
I added tests, I also deployed to my test environment and manually verified each scenario above. I'm using Microsoft Azure SSO. Was not able to test other auth types like OIDC.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #35344


superset/views/utils.py
```diff
@@ -58,6 +59,33 @@
...
+        if request.query_string:
+            target = request.full_path.rstrip("?")
+        else:
+            target = request.path
+
+    if target:
+        query["next"] = [target]
+
+    encoded_query = parse.urlencode(query, doseq=True)
+    redirect_url = parse.urlunparse(parsed._replace(query=encoded_query))
```
msyavuz, line 85: Can we use anything else than the private method?

superset-frontend/src/pages/Login/index.tsx
```diff
@@ -82,6 +82,26 @@ export default function Login() {
...
+    () => (nextUrl ? `/login/?next=${encodeURIComponent(nextUrl)}` : '/login/'),
+    [nextUrl],
+  );
+
+  const buildProviderLoginUrl = (providerName: string) => {
+    const base = `/login/${providerName}`;
+    return nextUrl
+      ? `${base}${base.includes('?') ? '&' : '?'}next=${encodeURIComponent(nextUrl)}`
+      : base;
+  };
```
msyavuz, line 104: It looks like this might be a bit more DRY but it's just a nit

sadpandajoe:
How does this work with `APP_ROOT`? Are we going to see any issues?

sfirke:
> How does this work with `APP_ROOT`? Are we going to see any issues?

I *think* it works, because we start with `appbuilder.get_url_for_login` which has the `APP_ROOT` prefix. Then we only swap the query string, leaving the base URL alone. Do we have any CI tests that would check that? I'm not deploying with that variable set.

superset/views/utils.py
sfirke: I think `_replace` is just the canonical, documented way to do this. E.g., search for `_replace` here to find examples: https://docs.python.org/3/library/urllib.parse.html.

superset/views/core.py
```diff
@@ -762,16 +762,28 @@ def dashboard(
...
         """
 
+        def redirect_to_login() -> FlaskResponse:
+            login_url = appbuilder.get_url_for_login
+            parsed = parse.urlparse(login_url)
+            query = parse.parse_qs(parsed.query, keep_blank_values=True)
+            query["next"] = [request.url]
+            encoded_query = parse.urlencode(query, doseq=True)
+            redirect_url = parse.urlunparse(parsed._replace(query=encoded_query))
+            return redirect(redirect_url)
```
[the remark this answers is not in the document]
sfirke: I did this!

sfirke:
@msyavuz @sadpandajoe @rusackas this is ready for review. Sorry to take so many tries to fix it, abstracting the redirect logic led to some knock-on side effects I had to resolve.

I feel confident that this is substantial improvement in application behavior, to have the `next` value persist through login. My users clamor for this and having deployed this in my test environment, I love it. But please do kick the tires on my code, I got considerable assistance from AI helpers.

superset/views/core.py
```diff
@@ -765,13 +765,21 @@ def dashboard(
         dashboard = Dashboard.get(dashboard_id_or_slug)
 
         if not dashboard:
+            if g.user is None or g.user.is_anonymous:
```
dpgaspar: We could use `get_current_user` instead on `superset.utils`

superset/views/core.py
```diff
@@ -765,13 +765,21 @@ def dashboard(
...
+                return redirect_to_login()
             abort(404)
 
+        # Redirect anonymous users to login for unpublished dashboards,
+        # in the edge case where a dataset has been shared with public
+        if (g.user is None or g.user.is_anonymous) and not dashboard.published:
```
dpgaspar: use `get_current_user` ?

superset/views/core.py
```diff
@@ -765,13 +765,21 @@ def dashboard(
...
+            return redirect_to_login()
+
         try:
             dashboard.raise_for_access()
         except SupersetSecurityException:
-            # Return 404 to avoid revealing dashboard existence
-            return Response(status=404)
+            if g.user is None or g.user.is_anonymous:
```
dpgaspar: use `get_current_user` ?

dpgaspar (commented):
LGTM, left a couple of comments.

dpgaspar:
> > How does this work with `APP_ROOT`? Are we going to see any issues?
> 
> I _think_ it works, because we start with `appbuilder.get_url_for_login` which has the `APP_ROOT` prefix. Then we only swap the query string, leaving the base URL alone. Do we have any CI tests that would check that? I'm not deploying with that variable set.

yes `appbuilder.get_url_for_login` uses Flask's `url_for` that respects Flask's `APPLICATION_ROOT`. If we want to be extra sure we could add a new test for this

superset/views/core.py
sfirke: Good suggestion! I made the change in each place.

sfirke:
@dpgaspar thank you for the review! I made the change you suggested, can you approve this for merging if nothing else is needed?

rusackas (approved):
LGTM!

PR #35346, opened 2025-09-30
feat(helm): migrate bitnami helm charts

mihaiiftode:
### SUMMARY

If you were not aware bitnami made changes to their helm charts and images, and images and charts were removed: https://github.com/bitnami/charts

image

I wasn't sure if maybe it should be using charts that bitnami migrated out of their main repo or use another open source alternative. I went with https://github.com/CloudPirates-io/helm-charts as the configuration parity is almost the same. 

### TESTING INSTRUCTIONS

You can manually deploy this on a k8s cluster, I tested it on a cluster and the application seems to be spawning, and migrations working correctly

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


Antonio-RiveroMartnez (commented):
Hey! First, thanks for being productive and taking care of this! ❤️ My two cents here, we should invest time evaluating, consulting others and researching before committing to yet another provider like `Cloudpirates` -- which might still be the one we want moving forward but probably best to explore first --? Otherwise we might get into a similar issue soon or start using images that are not safe etc.

PS: I'm not actively using these images so I don't want to block anything if others feel this is the right path 🫡

mihaiiftode:
> Hey! First, thanks for being productive and taking care of this! ❤️ My two cents here, we should invest time evaluating, consulting others and researching before committing to yet another provider like `Cloudpirates` -- which might still be the one we want moving forward but probably best to explore first --? Otherwise we might get into a similar issue soon or start using images that are not safe etc.
> 
> PS: I'm not actively using these images so I don't want to block anything if others feel this is the right path 🫡

Thanks, I understand your concerns, and I don't think there are charts that are on par with bitnami images/charts 😬 and I found that provider on artifacthub and was mentioned on a reddit post. The only problem is that the bitnami images are yanked or gone from their main repos, which can impact current deployments. 

I can ofc take the time to migrate to other chart providers, but I am not the best source to research what alternative providers you could use, I don't have that time to do proper research.

The alternative would be to use bitnami legacy images but those will no longer be updated.

rusackas:
Maybe this should be discussed as a SIP to either:
• Remove bitnami and recommend people use their own service of preference
• Remove official support for helm charts

rusackas:
CC @villebro

chgl:
As a quick workaround, we can also switch to the bitnamilegacy repository. That's what I did for a quick solution over at: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/861/files.

Just as an addition from a user perspective:

> • Remove bitnami and recommend people use their own service of preference

For CI and testing the helm chart as part of this repo, having local dependencies as sub-charts works best. Also for quick testing and debugging. But agreed that for production usage, we should recommend using external dependencies, e.g. via operators.

I personally used the cloudpirate charts as well, the main benefit is that they use official upstream images, which makes migrating away from the charts easier. But the risk of a Bitnami 2.0 still exists.

rusowyler:
Any news on this? 
Probably using bitnamilegacy until a better solution is found is a good idea. I mean, right now it doesn't work!

mihaiiftode:
My focus was taken away from this, I'd recommend you use helmfile and deploy an external redis helm chart of your preference, than relying on the provided ones.

PR #35419, opened 2025-10-01
fix(cache): ensure SQL is sanitized before cache key generation

betodealmeida:
### SUMMARY

Fixes an isssue where we get a cache miss because we sanitize SQL clauses (where/having) when caching a query object, but not when reading it from the cache. With the migration to `sqlglot` this introduces additional whitespace to the clauses, which significantly alters them, causing the cache miss. But the root of the problem is the inconsistent validation when genereating the cache key on write vs. read.

I also changed the `sanitize_clause` method to strip whitespace and comments, making the key more compact and consistent.

### TESTING INSTRUCTIONS

Updated and added tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI

### DIFF
--- a/superset/common/query_context_processor.py
+++ b/superset/common/query_context_processor.py
@@ -129,6 +129,11 @@ class QueryContextProcessor:
         self, query_obj: QueryObject, force_cached: bool | None = False
     ) -> dict[str, Any]:
         """Handles caching around the df payload retrieval"""
+        if query_obj:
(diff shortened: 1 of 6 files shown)


aminghadersohi (approved):

Vitor-Avila (approved):
LGTM!

PR #35425, opened 2025-10-01
feat(tree chart): add initial tree depth to tree chart

SBIN2010:
This PR Tree chart improvement

### SUMMARY

idea #18520 implemented

Снимок экрана от 2025-10-01 23-26-28

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


superset-frontend/plugins/plugin-chart-echarts/src/Tree/constants.ts
```diff
@@ -42,4 +42,5 @@ export const DEFAULT_FORM_DATA: Partial<EchartsTreeFormData> = {
   nodeLabelPosition: 'left',
   childLabelPosition: 'bottom',
   emphasis: 'descendant',
+  initialTreeDepth: 2,
```
rusackas, line 45: Should we default to -1? I'm assuming that's the default behavior. I don't want eveyone's charts out there in the wild to suddenly be smaller/simpler than expected.

SBIN2010: On the contrary, this is the default value.
https://echarts.apache.org/en/option.html#series-tree.initialTreeDepth

rusackas (approved):
LGTM!

PR #35453, opened 2025-10-02
fix(ag-grid-table): remove enterprise features to use community version

amaannawab923:
## SUMMARY
  🔧 Remove AG Grid Enterprise features to use community version only. AG Grid has strict enforcement where using any enterprise-only
  options without the corresponding enterprise modules registered will cause the table to render blank. This PR fixes those issues by
  removing all enterprise features.

  ## BEFORE/AFTER SCREENSHOTS OR ANIMATED GIF
  **Before:** ❌
  - Blank table rendering due to enterprise options being used without enterprise modules
  - Runtime error: "AG Grid: error  200 Unable to use rowGroupPanelShow as RowGroupingPanelModule is not registered"
  - Console errors for missing enterprise  modules (enableRowGroup, enableValue)

  
Screenshot 2025-10-02 at 9 55 30 PM

  **After:** ✅
  - AG Grid table renders correctly using only community features
  - No enterprise module errors
  - Table displays data properly with all community features functional
  
Screenshot 2025-10-02 at 9 56 12 PM

  

  ## TESTING INSTRUCTIONS
  1. ✨ Create a new chart using the AG Grid Table visualization
  2. ✅ Verify the table renders correctly with data (not blank)
  3. 🔍 Test column sorting, filtering, and resizing functionality
  4. 🐛 Verify no console errors related to missing AG Grid modules
  5. 📄 Confirm pagination works as expected
  6. 🔎 Test search and column rearrangement features


superset-frontend/plugins/plugin-chart-ag-grid-table/src/AgGridTable/index.tsx
```diff
@@ -131,10 +131,7 @@ const AgGridDataTable: FunctionComponent<AgGridTableProps> = memo(
 
     const defaultColDef = useMemo<ColDef>(
       () => ({
-        flex: 1,
```
kgabryje, line 134: does it work better without flex: 1? does it resize nicely when you resize the container?

alexandrusoare (approved):
LGTM - there is one CI failing but seems unrelated

superset-frontend/plugins/plugin-chart-ag-grid-table/src/AgGridTable/index.tsx
amaannawab923: Thanks for pointing this out , The Flex 1 , seems to be conflicting with sizeColumnsToFit again causing the table to render blank but we need sizeColumnsToFit , so  sizeColumnsToFit  takes care of this now and we dont need flex 1

kgabryje (approved):

msyavuz:
A rebase would fix the ci failures

msyavuz (approved):

amaannawab923:
The ci is green now @msyavuz

PR #35459, opened 2025-10-02
feat(pie): add the ability to make half a donut

SBIN2010:
### SUMMARY

This PR adds the ability to control the initial position of the sweep and control the display of the sweep angle.

Снимок экрана от 2026-01-10 15-26-42

Снимок экрана от 2026-01-10 15-26-59

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Required feature flags:
- [ ] Changes UI


michael-s-molina:
@rusackas @SBIN2010 In my opinion, this would be a great use case for that control that accepts additional ECharts configurations and let users further customize their dashboards. The problem of creating controls for these use cases is that the Explore panel keeps getting more complex.

SBIN2010:
Hi @michael-s-molina @rusackas!
Are there any comments or suggestions regarding this PR?
This PR has been gathering dust for a long time without your attention.

michael-s-molina:
> Hi @michael-s-molina @rusackas!
> Are there any comments or suggestions regarding this PR?

> @rusackas @SBIN2010 In my opinion, this would be a great use case for that control that accepts additional ECharts configurations and let users further customize their dashboards. The problem of creating controls for these use cases is that the Explore panel keeps getting more complex.

rusackas:
Agreed with Michael on the issue of control panel complexity as we edge closer and closer to supporting all the ECharts features. 

I also wonder if this thing starts to collide with the intent of the Gauge chart (which is hard to use, in my opinion). Maybe some day we can consolidate the two, if we design it well.

What @michael-s-molina is referring to, @SBIN2010, is that I have this "rainy day" plan that you'd be more than welcome to adopt on a PR. I was hoping to make a reusable Code Editor control panel input (promoting the one used by the Handlebars Chart) in one PR, and then give it a JSON editor mode. Once it has that, one could simply type up a JSON object of ECharts options... then you could add spacing between pie wedges, round the corners of wedges, or do whatever you want, really.

SBIN2010:
> Agreed with Michael on the issue of control panel complexity as we edge closer and closer to supporting all the ECharts features.
> 
> I also wonder if this thing starts to collide with the intent of the Gauge chart (which is hard to use, in my opinion). Maybe some day we can consolidate the two, if we design it well.
> 
> What @michael-s-molina is referring to, @SBIN2010, is that I have this "rainy day" plan that you'd be more than welcome to adopt on a PR. I was hoping to make a reusable Code Editor control panel input (promoting the one used by the Handlebars Chart) in one PR, and then give it a JSON editor mode. Once it has that, one could simply type up a JSON object of ECharts options... then you could add spacing between pie wedges, round the corners of wedges, or do whatever you want, really.

Hi @michael-s-molina  @rusackas! 
@rusackas , thank you very much for the clarification, now everything is clear to me. However, I don't entirely agree with you. You can put settings in JSON that wouldn't have dependencies on other settings. Take this PR as an example: if we are making a setting to display the donut percentage, we need to consider that the total display will no longer be centered, and also if the legend display changes, that must be taken into account as well.

If we consider other settings, like those @rusackas  mentioned, which don't have dependent settings, it would be more reasonable to wrap them in JSON.

rusackas:
Agreed things that are more dynamic would make sense to have as controls... especially where one parameter change requires another. 

I'd still prefer a "degrees" input that lets you set anything (defaulting to 360) than hard-coding a half-circle. Gets us closer to killing the Gauge ;)

rusackas:
Thanks for iterating on this @SBIN2010. A few things still need addressing before I can approve:

**Design question**: now that `start_angle` and `swept_angle` controls exist, is the separate `half` boolean still needed? It feels redundant — half-donut is just `start_angle: 180, swept_angle: 180`. I'd lean toward dropping `half` and relying on the angle controls alone (with sensible defaults), which avoids the dual-codepath padding logic below.

**Code issues flagged by review bots that look legitimate**:

1. `transformProps.ts:107` — codeant-ai flagged a division-by-zero: `chartPadding` divided by `height`/`width` with no guard. Worth an early return / safe default when dimensions are non-positive.
2. `transformProps.ts:111` — bito flagged that the `if (half)` block sets `padding.top`, then the subsequent `if (chartPadding.top)` block overwrites it. Need to consolidate so only one path sets the final value.
3. Control names `start_angle`/`swept_angle` (snake_case) don't match `DEFAULT_FORM_DATA.startAngle`/`sweptAngle` (camelCase). Defaults likely won't apply correctly — please align the casing.
4. The new control descriptions ("Angle at which to start progress axis", "axis of progress rotates") look copy-pasted from the gauge chart. Could you reword them for pie/donut semantics?
5. Worth adding clamping/validation on the angles (negatives, >360, swept ≤ 0) before passing to ECharts.

A quick test for the new transform logic would also help.

EnxDev:
Hey @SBIN2010, thanks for your effort here! Are you planning to continue working on this? Feel free to ping me if you need any help.

rusackas:
@SBIN2010 picking this back up — the points from my last pass are still open (the redundant `half` boolean + dual padding path, snake/camel control-name mismatch, copy-pasted gauge descriptions, angle clamping, a test). On top of that `lint-frontend` is now red on a trivial one: `transformProps.ts:430` `startAngle: startAngle` trips `object-shorthand`, just write `startAngle`. Want to push a rebase + those fixes and re-ping me?

EnxDev:
## EnxDev's Review Agent — apache/superset#35459 · HEAD 8ea2986
request changes — the padding refactor is behavior-preserving, but the half-donut re-centering only fires for one exact angle combo (`startAngle===180`), so most configurations reachable from the two new controls render off-center; and the new logic ships with no tests.

Feature PR: adds `Start angle` / `Swept angle` controls to the Pie chart and derives a "half donut" layout. I verified the `getTotalValuePadding` rewrite against `master`: for existing (non-half) charts it is behaviorally identical — `getChartPadding` (non-horizontal branch, the one Pie uses) only ever sets one of top/bottom/left/right, so the reordered branches produce the same result, and the new `safeHeight/safeWidth` guards only change the previously-NaN zero-dimension case. Defaults `startAngle:90 / sweptAngle:360` reproduce ECharts' own defaults (`startAngle 90`, `endAngle -270`), so saved charts are unchanged. No blocking regression there.

### 🔴 Functional
- **`transformProps.ts:428` (`isHalfDonut`) & `:431` (`center`)** · _Medium_ — Re-centering keys off `startAngle === HALF_DONUT_START_ANGLE` (exactly 180) AND `sweptAngle <= 180`. The controls advertise any angle 0–360. So the common case — user sets `Swept angle = 180` but leaves `Start angle` at its default 90 (or picks 0/270) — is *not* recognized as half: `center` stays `['50%','50%']` and the Total text uses the full-donut path, so the half arc hangs off-center and the total is mispositioned. Key the layout off the actual geometry (e.g. `sweptAngle <= 180`, deriving the center shift from `startAngle`) rather than a hardcoded start angle — or constrain the controls so only supported combos are selectable. **regression test:** `transformProps` with `{startAngle:90, sweptAngle:180}` should still re-center (`center[1] !== '50%'`), and `{startAngle:180, sweptAngle:180}` → `center:['50%','70%']`, `startAngle:180`, `endAngle:0`.

### 🟡 Should-fix
- **`transformProps.ts` / `test/Pie/transformProps.test.ts`** — New exported logic (`isHalfDonut`, the rewritten `getTotalValuePadding`, `startAngle`/`endAngle`/`center`) has zero test coverage; the functions were even exported but no test was added. Superset requires a regression test per behavioral change. Add cases for: the `isHalfDonut` threshold boundaries (`sweptAngle` 180 vs 181; `startAngle` 180 vs 90), `endAngle = startAngle - sweptAngle`, and the donut/non-donut + legend-position padding branches.
- **`controlPanel.tsx:123` / `:138`** — `min:0` lets `Swept angle = 0`, which renders an empty chart with no feedback. Consider `min:10` (matching `step`) or a sensible floor so the control can't produce an invisible viz.

### 🔵 Nits
- `transformProps.ts:76-77` — `HALF_DONUT_BASE_POSITION = 69` vs `HALF_DONUT_TOP_OFFSET = 68.5` are used for the same "half donut center" concept with a 0.5% gap that looks accidental; collapse to one constant unless the difference is intentional (add a comment if so).
- `transformProps.ts:82` — `isHalfDonut(sweptAngle, startAngle)` takes `(swept, start)` — reversed from the natural `(start, swept)` reading. Both call sites are correct, but the order is easy to mis-call later; consider `(startAngle, sweptAngle)`.


_Reviewed by EnxDev's Review Agent — @EnxDev · HEAD 8ea2986._

superset-frontend/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
```diff
@@ -411,7 +425,9 @@ export default function transformProps(
       animation: false,
       roseType: roseType || undefined,
       radius: [`${donut ? innerRadius : 0}%`, `${outerRadius}%`],
-      center: ['50%', '50%'],
+      center: ['50%', isHalfDonut(startAngle, sweptAngle) ? '70%' : '50%'],
+      startAngle,
+      endAngle: startAngle - sweptAngle,
```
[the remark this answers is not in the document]
SBIN2010: Default values ​​are set when clearing the field.

EnxDev:
## EnxDev's Review Agent — apache/superset#35459 · HEAD 939a16a
request changes — most prior feedback is addressed (tests added, `Swept angle` floor, constants + arg order cleaned up), but the one blocking item stands: half-donut re-centering still fires only when `startAngle === 180`, so the obvious "set Swept angle = 180" path renders off-center.

Re-review of [my prior review](https://github.com/apache/superset/pull/35459#issuecomment-4918890990) (HEAD `8ea2986`). Fixed since: ✅ test coverage added (`isHalfDonut` boundaries, `center`, `endAngle`, padding branches); ✅ `sweptAngle` control now `min: 10`; ✅ constants collapsed to `HALF_DONUT_OFFSET = 68.5` (no more stray `69`); ✅ `isHalfDonut(startAngle, sweptAngle)` reads in natural order. The `getTotalValuePadding` rewrite re-verified against `master`: still behaviorally identical for existing (non-half) charts — `calculateTop`/`calculateLeft` preserve the old bottom-over-top and right-over-left precedence with the same formulas — and defaults (`startAngle 90 / sweptAngle 360 → endAngle −270`) reproduce ECharts' own defaults, so saved charts are unchanged.

The widely-flagged "NaN when the angle field is cleared" is a non-issue: both controls declare non-null `default`s, and the control reducer re-applies `default` whenever a `NumberControl` is cleared to `undefined` (`getControlState.ts:124-128` — `if (state.default != null && value == null) value = state.default`). So `startAngle`/`sweptAngle` never reach `transformProps` as `undefined`; `endAngle` can't become `NaN`. The author's reply on that thread is correct.

### 🔴 Functional
- **`transformProps.ts` · `isHalfDonut` (`sweptAngle <= 180 && startAngle === 180`)** · _Medium_ — Unchanged from the prior review, and the new test `isHalfDonut(90, 180) → false` now entrenches it. The two controls accept any angle, so the natural way to make a half donut — set `Swept angle = 180`, leave `Start angle` at its default 90 (or pick 0 / 270) — is not detected as half: `center` stays `['50%','50%']` and the Total text takes the full-donut path, so the arc hangs off-center and the total is mispositioned, with no in-UI hint that `Start angle` must also be exactly 180. Either derive the layout from geometry (re-center whenever `sweptAngle < 360`, shifting the center based on `startAngle`) or replace the two free-form controls with a supported-configuration control (e.g. a `Half donut` toggle) so only renderable combos are reachable. **regression test:** `transformProps({ startAngle: 90, sweptAngle: 180 })` should still re-center (`center[1] !== '50%'`) and place the total off the geometric middle.

### 🙌 Praise
- `test/Pie/transformProps.test.ts` — the added suite asserts real values across the new branches (`isHalfDonut` boundaries, `center`, `endAngle`, donut/non-donut + legend-position padding); genuine regression guards, not smoke tests.


_Reviewed by EnxDev's Review Agent — @EnxDev · HEAD 939a16a._

superset-frontend/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
```diff
@@ -73,52 +76,84 @@ export function parseParams({
...
+  top: { centerOffset: 20, totalTopBase: 68.5 },
+  bottom: { centerOffset: -20, totalTopBase: 30 },
+  none: { centerOffset: 0, totalTopBase: 50 },
+};
+
+export const getHalfDonut = (
+  startAngle: number,
+  sweptAngle: number,
+): HalfDonut => {
+  if (sweptAngle > HALF_DONUT_SWEEP_LIMIT) return 'none';
```
EnxDev, line 125: `getHalfDonut` only recognizes top (startAngle % 360 === 180) and bottom (=== 0) halves. A lateral half-donut  startAngle=90 or 270 with sweptAngle=180, fully reachable via the new controls  gets 'none', so center stays ['50%', '50%'] (transformProps.ts:451) and the arc renders pushed into one lateral half of the canvas with the other half blank. 
Same for any typed start angle that isn't exactly 0/180/360. Worth either adding left/right layouts with a horizontal centerOffset, or documenting top/bottom-only in the control description.

rusackas:
Thanks for this, @SBIN2010! One thing before we can merge: @EnxDev's comment about the lateral half-donut case (startAngle 90/270 with sweptAngle ≤180) is still open, the arc ends up pushed to one side since `getHalfDonut` only handles top and bottom. Might be worth adding left/right centering too, or just restricting the control description to top/bottom for now.

Also there are a bunch of bot comments still open to assess/address/resolve, if you don't mind. Let me know if you want help with that part. Thanks again!

superset-frontend/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
rusackas: Left/right layouts have landed since this comment, and I've updated the control descriptions to match the behavior. Fully geometric recentering for arbitrary angles can come in a follow-up PR. Thanks @EnxDev!

rusackas (approved):
Thanks for sticking with this one @SBIN2010, and for the quick turnaround on the left/right layouts. I pushed a cleanup commit to get it over the line... mostly rewording the two control descriptions to match the new behavior, moving the angle controls into the Pie shape section, and syncing the translation catalogs since the strings changed. Also a couple more tests and a small nudge to the Total label in the right-side layout.

Smarter geometric recentering (so non-cardinal start angles don't jump off-center) can come in a follow-up.

Approving, and hopefully we can merge once CI's green. Thanks again!

